PR #1360, opened 2021-06-18
Fix path for extensions loaded using runbook

dsrivastavv:
All the extensions mentioned recursive runbooks are being loaded wrt to `constants.RUNBOOK_PATH`. This limits specifying relative path in parent runbooks.

### DIFF
--- a/lisa/parameter_parser/runbook.py
+++ b/lisa/parameter_parser/runbook.py
@@ -238,7 +238,7 @@ class RunbookBuilder:
         data_from_current: Dict[str, Any],
     ) -> List[Any]:
         old_extensions = self._load_extensions(merged_path, data_from_parent)
-        extensions = self._load_extensions(constants.RUNBOOK_PATH, data_from_current)
+        extensions = self._load_extensions(merged_path, data_from_current)
         # remove duplicate paths
         for old_extension in old_extensions:
             for extension in extensions:


lisa/parameter_parser/runbook.py
```diff
@@ -238,7 +238,7 @@ def _merge_extensions(
         data_from_current: Dict[str, Any],
     ) -> List[Any]:
         old_extensions = self._load_extensions(merged_path, data_from_parent)
-        extensions = self._load_extensions(constants.RUNBOOK_PATH, data_from_current)
+        extensions = self._load_extensions(merged_path, data_from_current)
```
squirrelsc, line 241: How about previous line? The old_extensions path are using current folder too. Is it tested?

dsrivastavv: That should be correct as it is basically resolving parent runbooks data using current path. Do you have specific situation in mind that I can test here?

squirrelsc: I think we need some UT to cover the `_merge_extensions` and `_merge_variables`. Would you add a task about it?

Please try below scenario in this change,
root.yml
```yaml
parent:
  - path: ./path2/tier.yml
extension:
  - ./path5
```
./path2/tier.yml
```yaml
parent:
  - path: ./path1/t$(tier).yml
extension:
  - ./path3
```
./path2/path1/t1.yml
```yaml
extension:
  - ./path4
```

squirrelsc (approved):

lisa/parameter_parser/runbook.py
dsrivastavv: The runbook works fine. And I will add a task to cover Unit test for this scenario.

PR #1361, opened 2021-06-18
Make clarification for users to read test results

WenyiQ:
Add content on test results (status) for users to get a hold of how tests went


WenyiQ:
Something went wrong

PR #1363, opened 2021-06-21
Make clarification for users to read test results

WenyiQ:
Add understand_test.md for users to understand test results.


squirrelsc (dismissed):
Looks good to me, and make sure it organized well with other information, so it can help people understand LISA smoothly.

squirrelsc:
BTW, we can add a picture to describe how the test results transit.

docs/run_test/understand_results.md
```diff
@@ -0,0 +1,71 @@
...
+
+- **ASSIGNED**
+
+  ASSIGNED tests are tests that are assigned to an environment, and will start
+  to run once the environment is deployed. They suggest some environmental
+  setting up is going on. 
+
+  ASSIGNED tests will end with FAILED if the environment fails to deploy. They
+  will also move backward to QUEUED again if the environment does not fit the
+  requirement of the test.
```
glima: Do we ever need that change of status in the first place, given env reqs don't change dynamically, AFAIK, and the check is simple enough? (I mean back from ASSIGNED)

squirrelsc: Yes, The reason of case move back to Queued is due to all cases need to be reselected. the described condition wont' happen (the env doesn't fit the test).

docs/run_test/understand_results.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Overview
+
+It's essential to understand the results after running tests. LISA has 7 kinds
+of test results in total: 3 of which are intermediate results, and 4 of which
+are final results, as explained in sections below. Each test case can and will
+be moved from one result to another but can never have two or more results at
+the same time. 
+
+![test_results](../img/test_results.png)
```
squirrelsc, line 15: Between the queued and assigned, it should have two lines with opposite direction.

docs/run_test/understand_results.md
```diff
@@ -0,0 +1,73 @@
...
+- **PASSED**
+
+  PASSED tests are tests that passed, or at least partially passed, with a
+  special `PASSException` that warns there are minor errors in the run but they
+  do not affect the test result.
+
+- **SKIPPED**
+
+  SKIPPPED tests are tests that did not start and would no longer run. They
+  suggest failure to meet some requirements in the environment involved with the
```
squirrelsc: "the environment" may be multiple one. One of skipping reason is, a queued case will try to match every created environment. If no env is matched, it will be skipped.

docs/run_test/understand_results.md
glima: We should/could also state that queued also means a pre-selection has already taken place (modulo environment), and the test has passed it (runbook criteria), no?

squirrelsc: @WenyiQ 
Yes, we can specify the queued case is selected by criteria. If a case is not queued, check the log if the case is included by extension paths, and criteria conditions.

squirrelsc (dismissed):

docs/run_test/understand_results.md
squirrelsc: The reasons is below, the explanation should be shorter to clear.

queued -> assigned: assign a test case to an environment. (the underlying reason may be to run or to attach deployment failure.
assigned -> queued: environment deployment or initialization is success, no need to use the test case to attach a failure.

WenyiQ: So, in other words, for queued -> assigned, it means a test in queue gets assigned to an env, either in order to run or to attach (handle?) a deployment failure, and for queued -> assigned, the env gets deployed/initialized successfully. Is my understanding correct?

squirrelsc (approved):

PR #1364, opened 2021-06-21
Part1 - Add NVMe test suite.

LiliDeng:
1. Add nvme feature.
2. Add nvmecli tool.
3. Add case nvme_basic_validation.

### DIFF
--- a/lisa/features/nvme.py
+++ b/lisa/features/nvme.py
@@ -1,0 +1,87 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from typing import Any, List
+
+from lisa.feature import Feature
+from lisa.tools import Lspci, Nvmecli
+from lisa.tools.lspci import PciDevice
+from lisa.util.logger import get_logger
+
+FEATURE_NAME_NVME = "NVME"
+
+
+class Nvme(Feature):
+    # crw------- 1 root root 251, 0 Jun 21 03:08 /dev/nvme0
+    _device_pattern = re.compile(r".*(?P<device_name>/dev/nvme[0-9]$)", re.MULTILINE)
+    # brw-rw---- 1 root disk 259, 0 Jun 21 03:08 /dev/nvme0n1
+    _namespace_pattern = re.compile(
+        r".*(?P<namespace>/dev/nvme[0-9]n[0-9]$)", re.MULTILINE
+    )
+    # '/dev/nvme0n1          351f1f720e5a00000001 Microsoft NVMe Direct Disk               1           0.00   B /   1.92  TB    512   B +  0 B   NVMDV001' # noqa: E501
+    _namespace_cli_pattern = re.compile(
+        r"(?P<namespace>/dev/nvme[0-9]n[0-9])", re.MULTILINE
+    )
+    _pci_device_name = "Non-Volatile memory controller"
+    _ls_devices: str = ""
+
+    def __init__(self, node: Any, platform: Any) -> None:
+        super().__init__(node, platform)
+        self._log = get_logger("feature", self.name(), self._node.log)
+
+    @classmethod
+    def name(cls) -> str:
+        return FEATURE_NAME_NVME
+
+    @classmethod
+    def enabled(cls) -> bool:
+        return True
+
+    @classmethod
+    def can_disable(cls) -> bool:
+        return True
+
+    def get_devices(self) -> List[str]:
+        devices_list = []
+        self._get_device_from_ls()
+        for row in self._ls_devices.splitlines():
+            matched_result = self._device_pattern.match(row)
+            if matched_result:
+                devices_list.append(matched_result.group("device_name"))
+        return devices_list
+
+    def get_namespaces(self) -> List[str]:
+        namespaces = []
+        self._get_device_from_ls()
+        for row in self._ls_devices.splitlines():
+            matched_result = self._namespace_pattern.match(row)
+            if matched_result:
+                namespaces.append(matched_result.group("namespace"))
+        return namespaces
+
+    def get_namespaces_from_cli(self) -> List[str]:
+        namespaces_cli = []
+        nvme_cli = self._node.tools[Nvmecli]
+        nvme_list = nvme_cli.run("list", shell=True, sudo=True)
+        for row in nvme_list.stdout.splitlines():
+            matched_result = self._namespace_cli_pattern.match(row)
+            if matched_result:
+                namespaces_cli.append(matched_result.group("namespace"))
+        return namespaces_cli
+
+    def get_devices_from_lspci(self) -> List[PciDevice]:
+        devices_from_lspci = []
+        lspci_tool = self._node.tools[Lspci]
+        device_list = lspci_tool.get_device_list()
+        devices_from_lspci = [
+            x for x in device_list if self._pci_device_name == x.device_class
+        ]
+        return devices_from_lspci
+
+    def _get_device_from_ls(self, force_run: bool = False) -> None:
+        if (not self._ls_devices) or force_run:
+            execute_results = self._node.execute(
+                "ls -l /dev/nvme*", shell=True, sudo=True
+            )
+            self._ls_devices = execute_results.stdout
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -266,6 +266,7 @@ class AzurePlatform(Platform):
     def supported_features(cls) -> List[Type[Feature]]:
         return [
             features.Gpu,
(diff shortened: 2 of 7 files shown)


lisa/executable.py
```diff
@@ -154,7 +154,7 @@ def _check_exists(self) -> bool:
         else:
             where_command = "where"
         result = self.node.execute(
-            f"{where_command} {self.command}", shell=True, no_info_log=True
+            f"{where_command} {self.command}", shell=True, sudo=True, no_info_log=True
```
squirrelsc: When a command run with sudo, the path may be different. It's not safe to add sudo here. If it's needed for main scenario, we can check twice. The second time uses sudo, if the command can be found in this step, the command will be executed with sudo always. Please give me a task, I will update it.

lisa/features/nvme.py
```diff
@@ -0,0 +1,74 @@
...
+class Nvme(Feature):
+    _nvme_device_pattern = re.compile(r".*(/dev/nvme[0-9]$)", re.MULTILINE)
+    _nvme_namespace_pattern = re.compile(r".*(/dev/nvme[0-9]n[0-9]$)", re.MULTILINE)
+    _nvme_namespace_cli_pattern = re.compile(r"(/dev/nvme[0-9]n[0-9])", re.MULTILINE)
+    _nvme_pci_device_name = "Non-Volatile memory controller"
+
+    def __init__(self, node: Any, platform: Any) -> None:
+        super().__init__(node, platform)
+        self._log = get_logger("feature", self.name(), self._node.log)
+        self._nvme_devices = self._node.execute(
```
squirrelsc: Don't call command or other long time operations in `__init__`.

lisa/features/nvme.py
```diff
@@ -0,0 +1,74 @@
...
+from lisa.feature import Feature
+from lisa.tools import Lspci, Nvmecli
+from lisa.tools.lspci import PciDevice
+from lisa.util.logger import get_logger
+
+FEATURE_NAME_NVME = "Nvme"
+
+
+class Nvme(Feature):
+    _nvme_device_pattern = re.compile(r".*(/dev/nvme[0-9]$)", re.MULTILINE)
```
squirrelsc: Add examples for the patterns.

lisa/features/nvme.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @classmethod
+    def enabled(cls) -> bool:
+        return True
+
+    @classmethod
+    def can_disable(cls) -> bool:
+        return True
+
+    def get_nvme_device(self) -> List[str]:
```
squirrelsc: 1. The method returns multiple items, 
2. It's nvme feature already. 
So, the name should be like `get_devices`.

lisa/features/nvme.py
```diff
@@ -0,0 +1,74 @@
...
+
+import re
+from typing import Any, List
+
+from lisa.feature import Feature
+from lisa.tools import Lspci, Nvmecli
+from lisa.tools.lspci import PciDevice
+from lisa.util.logger import get_logger
+
+FEATURE_NAME_NVME = "Nvme"
```
squirrelsc: The name doesn't need to follow programing name convention, and it should follow custom usage. So it should be "NVME".

lisa/features/nvme.py
```diff
@@ -0,0 +1,74 @@
...
+            if matched_result:
+                nvme_devices.append(matched_result.group(1))
+        return nvme_devices
+
+    def get_nvme_namespace(self) -> List[str]:
+        self.nvme_namespaces = []
+        for row in self._nvme_devices.stdout.splitlines():
+            matched_result = self._nvme_namespace_pattern.match(row)
+            if matched_result:
+                self.nvme_namespaces.append(matched_result.group(1))
```
squirrelsc: use named group, it make the code more readable.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1182,6 +1183,8 @@ def _resource_sku_to_capability(
         node_space.name = f"{location}_{resource_sku.name}"
         node_space.features = search_space.SetSpace[str](is_allow_set=True)
         for sku_capability in resource_sku.capabilities:
+            if "standardLSv2Family" == resource_sku.family:
```
squirrelsc: write like below, it's extendable.
`if resource_sku.family in ["standardLSv2Family"]:`

lisa/tools/nvmecli.py
```diff
@@ -0,0 +1,44 @@
...
+        self.node.shell.mkdir(tool_path, exist_ok=True)
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
+        make = self.node.tools[Make]
+        code_path = tool_path.joinpath("nvme-cli")
+        posix_os.install_packages("pkg-config")
+        make.make_and_install(cwd=code_path)
+
+    def install(self) -> bool:
+        if not self._check_exists():
```
squirrelsc, line 38: Does it have to be installed from source? If the logic is tested in LISAv2, it's ok. , but it's better to try to install from package firstly.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -0,0 +1,68 @@
...
+            supported_features=[Nvme],
+        ),
+    )
+    def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
+        # 1. Get nvme devices and nvme namespaces from /dev/ folder,
+        #  compare the count of nvme namespaces and nvme device.
+        nvme = node.features[Nvme]
+        nvme_device = nvme.get_nvme_device()
+        nvme_namespace = nvme.get_nvme_namespace()
+        assert_that(len(nvme_device)).is_equal_to(len(nvme_namespace))
```
squirrelsc: add description.

lisa/features/nvme.py
```diff
@@ -0,0 +1,79 @@
...
+    @classmethod
+    def enabled(cls) -> bool:
+        return True
+
+    @classmethod
+    def can_disable(cls) -> bool:
+        return True
+
+    def get_devices(self) -> List[str]:
+        devices = self._node.execute("ls -l /dev/nvme*", shell=True, sudo=True)
```
squirrelsc: Put the execute in a shared method, it can parse both device and namespaces together.

lisa/tools/nvmecli.py
```diff
@@ -0,0 +1,44 @@
...
+from lisa.operating_system import Posix
+from lisa.tools import Git, Make
+
+
+class Nvmecli(Tool):
+    repo = "https://github.com/linux-nvme/nvme-cli"
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git, Make]
```
squirrelsc: Because the dependencies is not necessary, when installing form package, let's call `install_packages` in `_install_from_src`, and not declare dependencies here.

lisa/features/nvme.py
```diff
@@ -0,0 +1,86 @@
...
+
+    @classmethod
+    def enabled(cls) -> bool:
+        return True
+
+    @classmethod
+    def can_disable(cls) -> bool:
+        return True
+
+    def get_device_from_ls(self, force_run: bool = False) -> None:
```
squirrelsc: Mark it private, and put under other public methods.

lisa/tools/nvmecli.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install_from_src(self) -> None:
+        posix_os: Posix = cast(Posix, self.node.os)
+        tool_path = self.get_tool_path()
+        self.node.shell.mkdir(tool_path, exist_ok=True)
+        git = self.node.tools[Git]
```
squirrelsc, line 27: Call `os.install_packages([Git, Make])`, it would be faster than call them one by one.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -0,0 +1,79 @@
...
+        nvme_namespace = nvme.get_namespaces()
+        assert_that(len(nvme_device)).described_as(
+            "nvme devices count should be equal to namespace count by listing devices "
+            "under folder /dev."
+        ).is_equal_to(len(nvme_namespace))
+
+        # 2. Compare the count of nvme namespaces return from `nvme list`
+        #  and list nvme namespaces under /dev/.
+        nvme_namespace_cli = nvme.get_namespaces_from_cli()
+        assert_that(len(nvme_namespace_cli)).described_as(
```
squirrelsc: use is_length as possible, it will show actual values, when it's failed.

squirrelsc (dismissed):

squirrelsc (approved):

PR #1365, opened 2021-06-21
Add support for new distro AlmaLinux

lubaihua33:


Testscripts/Linux/CPUOfflineKernelBuild.sh
```diff
@@ -101,7 +101,7 @@ function Main() {
 		# Start kernel compilation
 		# RHEL & CentOS need extra space for kernel repo due to different disk partitioning.
 		LogMsg "Clone and compile new kernel from $repo_url"
-		if [[ $DISTRO == *"redhat"* ]] || [[ $DISTRO == *"centos"* ]]; then
+		if [[ $DISTRO == *"redhat"* ]] || [[ $DISTRO == *"centos"* ]] || [[ $DISTRO == *"almalinux"* ]]; then
```
lizzha, line 104: better if we can use a common function instead of such conditions, e.g. isRhelFamily, so that we don't need to change so many places for adding a new distro support.

Not asking to address it in this PR. I guess LISAv3 will handle this better?

lizzha (approved):

Testscripts/Linux/CPUOfflineKernelBuild.sh
squirrelsc: In v3, there is only one place to detect the distro type. the special behaviors of RedHat will be inherited. If the distro has it's own behavior, it can be customized, but also inherit others of RedHat.

squirrelsc (approved):

PR #1369, opened 2021-06-22
Part2 - Migrate NVMe test cases.

LiliDeng:
Add below cases -

- nvme_function_validation
- nvme_fstrim_validation
- nvme_blkdiscard_validation
- nvme_manage_nm_validation
- nvme_rescind_validation

### DIFF
--- a/lisa/tools/fdisk.py
+++ b/lisa/tools/fdisk.py
@@ -1,0 +1,68 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from enum import Enum
+from typing import cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.util import LisaException
+
+from .mkfs import Mkfsext, Mkfsxfs
+
+FileSystem = Enum(
+    "mkfs",
+    ["xfs", "ext2", "ext3", "ext4"],
+)
+
+
+class Fdisk(Tool):
+    @property
+    def command(self) -> str:
+        return "fdisk"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def make_partition(self, disk_name: str, file_system: FileSystem) -> None:
+        """
+        disk_name: make a partition against the disk.
+        file_system: making the file system type against the partition
+        Make a partition and a filesystem against the disk.
+        """
+        # n => new a partition
+        # p => primary partition
+        # 1 => Partition number
+        # "" => Use default 2048 for 'First sector'
+        # "" => Use default 2147483647 as 'Last sector'
+        # w => write table to disk and exit
+        self.node.execute(
+            f"(echo n; echo p; echo 1; echo ; echo; echo ; echo w) | "
+            f"{self.command} {disk_name}",
+            shell=True,
+            sudo=True,
+        )
+        if file_system == FileSystem.xfs:
+            mkfs_xfs = self.node.tools[Mkfsxfs]
+            mkfs_xfs.mkfs(f"{disk_name}p1", str(file_system))
+        elif file_system in [FileSystem.ext2, FileSystem.ext3, FileSystem.ext4]:
+            mkfs_ext = self.node.tools[Mkfsext]
+            mkfs_ext.mkfs(f"{disk_name}p1", str(file_system))
+        else:
+            raise LisaException(f"Unrecognized file system {file_system}.")
+
+    def delete_partition(self, disk_name: str) -> None:
+        """
+        disk: delete one partition against the disk.
+        Delete the only partition of this disk.
+        """
+        # d => delete a partition
+        # w => write table to disk and exit
+        self.node.execute(
+            f"(echo d; echo w) | {self.command} {disk_name}", shell=True, sudo=True
+        )
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("util-linux")
+        return self._check_exists()
--- a/lisa/tools/lspci.py
+++ b/lisa/tools/lspci.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 import re
-from typing import Any, List
+from typing import Any, Dict, List
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
@@ -29,6 +29,12 @@ PATTERN_PCI_DEVICE = re.compile(
     r"(?P<vendor>[^\"\']+)[\"\']\s+[\"\'](?P<device>.*?)[\"\']?$",
     re.MULTILINE,
 )
+
+DEVICE_TYPE_DICT: Dict[str, str] = {
+    "SRIOV": "Ethernet controller",
+    "NVME": "Non-Volatile memory controller",
+    "GPU": "3D controller",
+}
 
 
 class PciDevice:
@@ -64,8 +70,16 @@ class Lspci(Tool):
             self.node.os.install_packages("pciutils")
         return self._check_exists()
 
+    def _get_devices_slots_by_class_name(
+        self, class_name: str, force_run: bool = False
+    ) -> List[str]:
+        devices_list = self.get_device_list(force_run)
+        devices_slots = [x.slot for x in devices_list if class_name == x.device_class]
+        return devices_slots
+
     def get_device_list(self, force_run: bool = False) -> List[PciDevice]:
         if (not self._pci_devices) or force_run:
+            self._pci_devices = []
             result = self.run("-m", force_run=force_run, shell=True)
             if result.exit_code != 0:
                 result = self.run("-m", force_run=force_run, shell=True, sudo=True)
@@ -79,3 +93,27 @@ class Lspci(Tool):
                 self._pci_devices.append(pci_device)
 
         return self._pci_devices
+
+    def disable_devices(self, device_type: str) -> None:
+        if device_type.upper() not in DEVICE_TYPE_DICT.keys():
+            raise LisaException(f"pci_type {device_type} is not supported to disable.")
+        device_type_name = DEVICE_TYPE_DICT[device_type.upper()]
+        devices_slot = self._get_devices_slots_by_class_name(device_type_name)
+        if 0 == len(devices_slot):
+            self._log.debug("No matched devices found.")
+            return
+        for device_slot in devices_slot:
+            cmd_result = self.node.execute(
+                f"echo 1 > /sys/bus/pci/devices/{device_slot}/remove",
+                shell=True,
+                sudo=True,
+            )
+            cmd_result.assert_exit_code()
+        if len(self._get_devices_slots_by_class_name(device_type_name, True)) > 0:
+            raise LisaException(f"Fail to disable {device_type_name} devices.")
+
+    def enable_devices(self) -> None:
+        cmd_result = self.node.execute(
+            "echo 1 > /sys/bus/pci/rescan", shell=True, sudo=True
+        )
+        cmd_result.assert_exit_code()
--- a/lisa/tools/mkfs.py
+++ b/lisa/tools/mkfs.py
@@ -1,0 +1,46 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import re
+from typing import cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+
+
+class Mkfs(Tool):
+    __EXIST_FILE_SYSTEM_PATTERN = re.compile(
+        r".*appears to contain an existing filesystem", re.MULTILINE
+    )
+
+    @property
+    def command(self) -> str:
+        return "mkfs.xfs"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    # command - mkfs.xfs, mkfs.ext2, mkfs.ext3, mkfs.ext4
+    def mkfs(self, disk: str, command: str) -> None:
+        cmd_result = self.node.execute(
+            f"echo y | {command} {disk}", shell=True, sudo=True
+        )
+        if self.__EXIST_FILE_SYSTEM_PATTERN.match(cmd_result.stdout):
+            cmd_result = self.node.execute(
+                f"echo y | {command} -f {disk}", shell=True, sudo=True
+            )
+        cmd_result.assert_exit_code()
+
+
+class Mkfsxfs(Mkfs):
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("xfsprogs")
+        return self._check_exists()
+
+
+class Mkfsext(Mkfs):
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("e2fsprogs")
+        return self._check_exists()
--- a/lisa/tools/mount.py
+++ b/lisa/tools/mount.py
@@ -1,0 +1,46 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import re
+from pathlib import PurePosixPath
+from typing import cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.tools import Fdisk
+from lisa.util import LisaException
+
+
+class Mount(Tool):
+    __UMOUNT_ERROR_PATTERN = re.compile(r".*mountpoint not found", re.MULTILINE)
+
+    @property
+    def command(self) -> str:
+        return "mount"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def mount(self, disk_name: str, point: str) -> None:
+        self.node.shell.mkdir(PurePosixPath(point), exist_ok=True)
+        cmd_result = self.node.execute(
+            f"mount {disk_name} {point}", shell=True, sudo=True
+        )
+        cmd_result.assert_exit_code()
+
+    def umount(self, disk_name: str, point: str, erase: bool = True) -> None:
+        cmd_result = self.node.execute(f"umount {point}", shell=True, sudo=True)
+        if erase:
+            fdisk = self.node.tools[Fdisk]
+            fdisk.delete_partition(disk_name)
+            self.node.execute(f"rm -r {point}", shell=True, sudo=True)
+        if (
+            not self.__UMOUNT_ERROR_PATTERN.match(cmd_result.stdout)
+            and 0 != cmd_result.exit_code
+        ):
+            raise LisaException(f"Fail to run umount {point}.")
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("util-linux")
+        return self._check_exists()
--- a/lisa/tools/nvmecli.py
+++ b/lisa/tools/nvmecli.py
@@ -1,15 +1,23 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-
+import re
 from typing import cast
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
 from lisa.tools import Git, Make
+from lisa.util import find_patterns_in_lines
+from lisa.util.process import ExecutableResult
 
 
 class Nvmecli(Tool):
     repo = "https://github.com/linux-nvme/nvme-cli"
+    # error_count\t: 0
+    __error_count_pattern = re.compile(r"^error_count.*:[ ]+([\d]+)\r?$", re.M)
+    # [3:3] : 0     NS Management and Attachment Supported
+    __ns_management_attachement_support = "NS Management and Attachment Supported"
+    # [1:1] : 0x1   Format NVM Supported
+    __format_device_support = "Format NVM Supported"
 
     @property
     def command(self) -> str:
@@ -30,6 +38,18 @@ class Nvmecli(Tool):
         code_path = tool_path.joinpath("nvme-cli")
         make.make_and_install(cwd=code_path)
 
+    def create_namespace(self, namespace: str) -> ExecutableResult:
+        return self.run(f"create-ns {namespace}", shell=True, sudo=True)
+
+    def delete_namespace(self, namespace: str, id: int) -> ExecutableResult:
+        return self.run(f"delete-ns -n {id} {namespace}", shell=True, sudo=True)
+
+    def detach_namespace(self, namespace: str, id: int) -> ExecutableResult:
+        return self.run(f"detach-ns -n {id} {namespace}", shell=True, sudo=True)
+
+    def format_namespace(self, namespace: str) -> ExecutableResult:
+        return self.run(f"format {namespace}", shell=True, sudo=True)
+
     def install(self) -> bool:
         if not self._check_exists():
             posix_os: Posix = cast(Posix, self.node.os)
@@ -38,3 +58,22 @@ class Nvmecli(Tool):
             if not self._check_exists():
                 self._install_from_src()
         return self._check_exists()
+
+    def get_error_count(self, namespace: str) -> int:
+        error_log = self.run(f"error-log {namespace}", shell=True, sudo=True)
+        error_count = 0
+        # for row in error_log.stdout.splitlines():
+        errors = find_patterns_in_lines(error_log.stdout, [self.__error_count_pattern])
+        if errors[0]:
+            error_count = sum([int(element) for element in errors[0]])
+        return error_count
+
+    def support_ns_manage_attach(self, device_name: str) -> bool:
+        cmd_result = self.run(f"id-ctrl -H {device_name}", shell=True, sudo=True)
+        cmd_result.assert_exit_code()
+        return self.__ns_management_attachement_support in cmd_result.stdout
+
+    def support_device_format(self, device_name: str) -> bool:
+        cmd_result = self.run(f"id-ctrl -H {device_name}", shell=True, sudo=True)
+        cmd_result.assert_exit_code()
+        return self.__format_device_support in cmd_result.stdout
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -8,9 +8,10 @@ import signal
 import subprocess
 import time
 from dataclasses import dataclass
-from typing import Dict, Optional
+from typing import Dict, List, Optional, Union
 
 import spur  # type: ignore
+from assertpy.assertpy import AssertionBuilder, assert_that
 from spur.errors import NoSuchCommandError  # type: ignore
 
 from lisa.util.logger import Logger, LogWriter, get_logger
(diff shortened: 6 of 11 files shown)


microsoft/testsuites/nvme/nvme.py
```diff
@@ -1,14 +1,45 @@
...
+        node.execute(f"rm -rf {mount_point}", shell=True, sudo=True)
+
+
+def _format_mount_disk(
+    node: Node,
+    namespace: str,
+    file_system: str,
+) -> None:
+    posix_os: Posix = cast(Posix, node.os)
+    posix_os.install_packages("xfsprogs")
```
squirrelsc: If the package is installed, it shouldn't be installed again. If it's a subclass of Tool, the installed or not can be tracked easily. Please create a tool, instead of calling it like a method. If the tool is used by nvme test cases only, it can be in this file.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -1,14 +1,45 @@
...
+    _unmount_disk(node, namespace)
+
+    node.execute(
+        f"(echo n; echo p; echo 1; echo ; echo; echo ; echo w) | " f"fdisk {namespace}",
+        shell=True,
+        sudo=True,
+    )
+    node.execute(f"echo 'y' | mkfs.{file_system} {namespace}p1", shell=True, sudo=True)
+    node.execute(f"mkdir {mount_point}", shell=True, sudo=True)
+    node.execute(f"mount {namespace}p1 {mount_point}", shell=True, sudo=True)
```
squirrelsc: Mount may be used by other storage test cases, it's worth to be a tool. the mount/umount can be in same tool.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -1,14 +1,45 @@
...
+    node: Node,
+    namespace: str,
+    file_system: str,
+) -> None:
+    posix_os: Posix = cast(Posix, node.os)
+    posix_os.install_packages("xfsprogs")
+    mount_point = namespace.rpartition("/")[-1]
+    _unmount_disk(node, namespace)
+
+    node.execute(
```
squirrelsc: add comments to explain what happens in the key sequence.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +108,230 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+        ),
+    )
+    def nvme_function_validation(self, environment: Environment, node: Node) -> None:
+        nvme = node.features[Nvme]
+        nvme_namespaces = nvme.get_namespaces()
+        nvme_cli = node.tools[Nvmecli]
+        cat = node.tools[Cat]
+        for namespace in nvme_namespaces:
+            # 1. Check the the number of errors from nvme-cli before operations.
+            error_count_before_operations = nvme_cli.get_error_count(namespace)
```
squirrelsc, line 120: Rather than error count, is it possible to retrieve last error?

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +108,230 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+        This test case will do following things for each NVMe device.
+        1. Check the the number of errors from nvme-cli before operations.
+        2. Create a partition, filesystem and mount it.
+        3. Create a txt file on the partition, content is 'TestContent'.
+        4. Create a file 'data' on the partition, get the md5sum value.
+        5. Umount and remount the partition.
+        6. Get the txt file content, compare the value.
+        7. Check the the number of errors from nvme-cli after operations.
+        """,
+        priority=1,
```
squirrelsc: By default, the priority from v2 should plus 1. There are too many p1 test cases for nvme, we need to think about that. If a test case fail, what's the impact, and what's different coverage of two test cases.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +108,230 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+                f"fstrim {mount_point} -v", shell=True, sudo=True
+            )
+            assert_that(
+                intermediate_fstrim.exit_code,
+                f"command `fstrim {mount_point} -v` should execute successfully "
+                "after dd command.",
+            ).is_equal_to(0)
+
+            # 5. Delete the file 'data'.
+            node.execute(f"rm -rf {mount_point}/data", shell=True, sudo=True)
```
sharsonia: Should we add the tool/ shell method for remove dir?

microsoft/testsuites/nvme/nvme.py
```diff
@@ -1,14 +1,45 @@
...
+    mount_point = namespace.rpartition("/")[-1]
+    _unmount_disk(node, namespace)
+
+    node.execute(
+        f"(echo n; echo p; echo 1; echo ; echo; echo ; echo w) | " f"fdisk {namespace}",
+        shell=True,
+        sudo=True,
+    )
+    node.execute(f"echo 'y' | mkfs.{file_system} {namespace}p1", shell=True, sudo=True)
+    node.execute(f"mkdir {mount_point}", shell=True, sudo=True)
```
sharsonia: Is there a reason not to use shell mkdir method instead?

lisa/tools/nvmecli.py
```diff
@@ -38,3 +40,12 @@ def install(self) -> bool:
...
                 self._install_from_src()
         return self._check_exists()
+
+    def get_error_count(self, namespace: str) -> int:
+        error_log = self.run(f"error-log {namespace}", shell=True, sudo=True)
+        error_count = 0
+        for row in error_log.stdout.splitlines():
+            matched_result = self.__error_count_pattern.findall(row)
+            if matched_result:
+                error_count += int(matched_result[0])
```
glima: This can use the find_patterns_in_lines() helper

microsoft/testsuites/nvme/nvme.py
```diff
@@ -1,14 +1,45 @@
...
 from lisa.testsuite import simple_requirement
-from lisa.tools import Lscpu
+from lisa.tools import Cat, Lscpu, Nvmecli
+from lisa.util import SkippedException
+
+
+def _unmount_disk(node: Node, namespace: str, destory_mountpoint: bool = True) -> None:
+    mount_point = namespace.rpartition("/")[-1]
+    node.execute(f"umount {mount_point}", shell=True, sudo=True)
+    if destory_mountpoint:
```
glima: typo

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +108,230 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+            )
+            assert_that(
+                mount_result.exit_code,
+                f"command `mount {namespace}p1 {mount_point}` should fail to execute.",
+            ).is_not_equal_to(0)
+
+    @TestCaseMetadata(
+        description="""
+        This test case will check below commands are expected to fail to execute.
+        1. `nvme create-ns namespace` - create a namespace.
```
glima: I guess the reader deserves to know why they should fail?

lisa/tools/fdisk.py
```diff
@@ -0,0 +1,42 @@
...
+class Fdisk(Tool):
+    @property
+    def command(self) -> str:
+        return "fdisk"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def make_partition(self, disk: str) -> None:
```
squirrelsc: Add doc string to introduce what this method do, and what the argument is for. Is it the name of new partition?

lisa/tools/mkfsext.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @property
+    def command(self) -> str:
+        return self._command
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def mkfs(self, disk: str, command: str = "mkfs.ext4") -> None:
```
squirrelsc: add doc string to introduce what the method do, what can be filled into the command? Why the command need to be an argument?

lisa/tools/mkfsext.py
```diff
@@ -0,0 +1,42 @@
...
+        self._command = command
+        cmd_result = self.node.execute(
+            f"echo y | {self.command} {disk}", shell=True, sudo=True
+        )
+        if self.__EXIST_FILE_SYSTEM_PATTERN.match(cmd_result.stdout):
+            cmd_result = self.node.execute(
```
squirrelsc: If user run this case on their disks, will it damage the data? How to prevent it?

lisa/tools/mount.py
```diff
@@ -0,0 +1,54 @@
...
+    @property
+    def command(self) -> str:
+        return self._command
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def operations(
+        self, disk: str, point: str, command: str = "mount", erase: bool = True
```
squirrelsc: It can be two methods, `mount` and `umount`. It's not a good practice to pass the command in. The logic is different, and the str may have typo, and others don't know what's the right value.

lisa/tools/mount.py
```diff
@@ -0,0 +1,54 @@
...
+            if 0 != cmd_result.exit_code:
+                raise LisaException(f"Fail to run {self._command} {disk} {point}.")
+        elif "umount" == command:
+            self._command = "umount"
+            cmd_result = self.run(f"{point}", shell=True, sudo=True, force_run=True)
+            if erase:
+                self.node.execute(
+                    f"(echo d; echo w) | fdisk {disk}", shell=True, sudo=True
+                )
+                self.node.execute(f"rm -rf {point}", shell=True, sudo=True)
```
squirrelsc: The `-rf` is very dangerous. It doesn't need to use `-r` for a mount point. The mount point should be a single folder.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -8,7 +8,27 @@
...
+    node: Node,
+    namespace: str,
+    file_system: str,
+) -> None:
+    mount_point = namespace.rpartition("/")[-1]
+    fdisk = node.tools[Fdisk]
+    mount = node.tools[Mount]
+    mount.operations(namespace, mount_point, command="umount")
+    fdisk.make_partition(namespace)
+    if "xfs" == file_system:
```
squirrelsc: If they are always used together with fdisk, the logic can be in fdisk, not in test case. The two tools can be used by fdisk.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +97,237 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+            _format_mount_disk(node, namespace, "ext4")
+
+            # 3. Create a txt file on the partition, content is 'TestContent'.
+            mount_point = namespace.rpartition("/")[-1]
+            node.execute(
+                f"echo TestContent > {mount_point}/testfile.txt", shell=True, sudo=True
+            )
+
+            # 4. Create a file 'data' on the partition, get the md5sum value.
+            node.execute(
```
squirrelsc: not to add sudo, if it's not required.

microsoft/testsuites/nvme/nvme.py
squirrelsc: Avoid to use str on very limited candidates. the `file_system` can be an enum.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +97,237 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            supported_features=[Nvme],
+        ),
+    )
+    def nvme_function_negative_validation(
+        self, environment: Environment, node: Node
+    ) -> None:
+        negative_operations = ["create-ns", "delete-ns -n 1", "detach-ns -n 1"]
```
squirrelsc: They can be methods, because it's easier to use.

microsoft/testsuites/nvme/nvme.py
LiliDeng: To be frank, I am also lack of the context here, but I will try to figure it out.

lisa/tools/mkfsext.py
LiliDeng: for nvme, the disk name is easy to be distinguished from the OS disk. I will add a check to skip this step if user uses an OS disk.

lisa/tools/mkfs.py
```diff
@@ -0,0 +1,54 @@
...
+    def mkfs(self, disk: str, command: str) -> None:
+        self._command = command
+        cmd_result = self.node.execute(
+            f"echo y | {self.command} {disk}", shell=True, sudo=True
+        )
+        if self.__EXIST_FILE_SYSTEM_PATTERN.match(cmd_result.stdout):
+            cmd_result = self.node.execute(
+                f"echo y | {self.command} -f {disk}", shell=True, sudo=True
+            )
+        if 0 != cmd_result.exit_code:
```
squirrelsc: Any logic is used to validate the exit_code, it should log the exit code too.

lisa/tools/mount.py
```diff
@@ -0,0 +1,48 @@
...
+    @property
+    def command(self) -> str:
+        return self._command
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def mount(self, disk: str, point: str) -> None:
+        self._command = "mount"
```
squirrelsc: It doesn't need to save the self._command, because this value is only effective in this method. In the following code, it can call `self.node.execute`

lisa/tools/fdisk.py
```diff
@@ -0,0 +1,65 @@
...
+        self.node.execute(
+            f"(echo n; echo p; echo 1; echo ; echo; echo ; echo w) | "
+            f"{self.command} {disk}",
+            shell=True,
+            sudo=True,
+        )
+        if file_system == FileSystem.xfs:
+            mkfs_xfs = self.node.tools[Mkfsxfs]
+            mkfs_xfs.mkfs(f"{disk}p1", str(file_system))
+        if file_system in [FileSystem.ext2, FileSystem.ext3, FileSystem.ext4]:
```
squirrelsc: use elif, and raise Exception, if it's unrecognized.

lisa/tools/fdisk.py
```diff
@@ -0,0 +1,65 @@
...
+            mkfs_ext = self.node.tools[Mkfsext]
+            mkfs_ext.mkfs(f"{disk}p1", str(file_system))
+
+    def del_partition(self, disk: str) -> None:
```
squirrelsc: `delete_partition` is better, is not too long.

lisa/tools/fdisk.py
```diff
@@ -0,0 +1,65 @@
...
+class Fdisk(Tool):
+    @property
+    def command(self) -> str:
+        return "fdisk"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def make_partition(self, disk: str, file_system: FileSystem) -> None:
```
squirrelsc: The `disk_name` is better to understand.

lisa/tools/fdisk.py
```diff
@@ -0,0 +1,65 @@
...
+        # w => write table to disk and exit
+        self.node.execute(
+            f"(echo n; echo p; echo 1; echo ; echo; echo ; echo w) | "
+            f"{self.command} {disk}",
+            shell=True,
+            sudo=True,
+        )
+        if file_system == FileSystem.xfs:
+            mkfs_xfs = self.node.tools[Mkfsxfs]
+            mkfs_xfs.mkfs(f"{disk}p1", str(file_system))
```
squirrelsc: This line is the same. It can be out of the if statement.

lisa/tools/fdisk.py
glima: Which partition? The first it encounters?

lisa/tools/mkfs.py
glima: We should at least log what is going on here, since this are destructive ops going on.

lisa/tools/mount.py
```diff
@@ -0,0 +1,48 @@
...
+        self.node.shell.mkdir(PurePosixPath(point), exist_ok=True)
+        cmd_result = self.run(f"{disk} {point}", shell=True, sudo=True, force_run=True)
+        if 0 != cmd_result.exit_code:
+            raise LisaException(f"Fail to run {self._command} {disk} {point}.")
+
+    def umount(self, disk: str, point: str, erase: bool = True) -> None:
+        self._command = "umount"
+        cmd_result = self.run(f"{point}", shell=True, sudo=True, force_run=True)
+        if erase:
+            self.node.execute(f"(echo d; echo w) | fdisk {disk}", shell=True, sudo=True)
```
glima: Ditto

lisa/tools/nvmecli.py
```diff
@@ -38,3 +53,12 @@ def install(self) -> bool:
             if not self._check_exists():
                 self._install_from_src()
         return self._check_exists()
+
+    def get_error_count(self, namespace: str) -> int:
+        error_log = self.run(f"error-log {namespace}", shell=True, sudo=True)
+        error_count = 0
+        for row in error_log.stdout.splitlines():
+            errors = find_patterns_in_lines(row, [self.__error_count_pattern])
```
glima: This already spans lines, so no need for the outer for.

microsoft/testsuites/nvme/nvme.py
LiliDeng: there are many entries, we need sum up them together.

microsoft/testsuites/nvme/nvme.py
LiliDeng: without sudo, no permission

lisa/tools/mount.py
LiliDeng: without -r, can't delete folder

squirrelsc: I mean, use `-r` only, no `-f`.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +92,218 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+        This test case will
+        1. Create a partition, xfs filesystem and mount it.
+        2. Check how much the mountpoint is trimmed before operations.
+        3. Create a 300 gb file 'data' using dd command in the partition.
+        4. Check how much the mountpoint is trimmed after creating the file.
+        5. Delete the file 'data'.
+        6. Check how much the mountpoint is trimmed after deleting the file,
+         and compare the final fstrim status with initial fstrim status.
+        """,
+        priority=1,
```
squirrelsc: leave only 1 P1 case, and others can be p2 or p3.

squirrelsc: There is a P0 test case `nvme_basic_validation`, may be it should be P1, and downgrade others.

lisa/tools/lspci.py
```diff
@@ -79,3 +93,27 @@ def get_device_list(self, force_run: bool = False) -> List[PciDevice]:
                 self._pci_devices.append(pci_device)
 
         return self._pci_devices
+
+    def disable_devices(self, device_type: str) -> None:
+        if device_type.upper() not in DEVICE_TYPE_DICT.keys():
+            raise LisaException(f"pci_type {device_type} is not supported to disable.")
+        device_type_name = DEVICE_TYPE_DICT[f"{device_type.upper()}"]
```
squirrelsc: Write like this way is enough: `device_type.upper()`

lisa/tools/mkfs.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @property
+    def command(self) -> str:
+        return self._command
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def mkfs(self, disk: str, command: str) -> None:
```
squirrelsc, line 24: why the command need to be passed in? Which values support? If it have to do like this way, the comments is needed to explain.

lisa/tools/mkfs.py
```diff
@@ -0,0 +1,54 @@
...
+        self._command = command
+        cmd_result = self.node.execute(
+            f"echo y | {self.command} {disk}", shell=True, sudo=True
+        )
+        if self.__EXIST_FILE_SYSTEM_PATTERN.match(cmd_result.stdout):
+            cmd_result = self.node.execute(
+                f"echo y | {self.command} -f {disk}", shell=True, sudo=True
+            )
+        if 0 != cmd_result.exit_code:
+            raise LisaException(f"Fail to run 'echo y | {self.command} {disk}'.")
```
squirrelsc: Change this to use assert_exit_code too.

lisa/tools/nvmecli.py
```diff
@@ -30,6 +34,18 @@ def _install_from_src(self) -> None:
...
+    def create_namespace(self, namespace: str) -> ExecutableResult:
+        cmd_result = self.run(f"create-ns {namespace}", shell=True, sudo=True)
+        return cmd_result
+
+    def delete_namespace(self, namespace: str, id: int) -> ExecutableResult:
+        cmd_result = self.run(f"delete-ns -n {id} {namespace}", shell=True, sudo=True)
+        return cmd_result
+
+    def detach_namespace(self, namespace: str, id: int) -> ExecutableResult:
+        cmd_result = self.run(f"detach-ns -n {id} {namespace}", shell=True, sudo=True)
```
squirrelsc: It can be like below. BTW, why not parse the result, and return success or fail?
`return self.run(f"detach-ns -n {id} {namespace}", shell=True, sudo=True)`

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +92,234 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+            supported_features=[Nvme],
+        ),
+    )
+    def nvme_function_negative_validation(self, node: Node) -> None:
+        nvme = node.features[Nvme]
+        nvme_namespaces = nvme.get_namespaces()
+        nvme_cli = node.tools[Nvmecli]
+        for namespace in nvme_namespaces:
+            create_namespace = nvme_cli.create_namespace(namespace)
+            create_namespace.assert_exit_code(expected_exit_code=1)
```
squirrelsc: It needs description to explain why the `1` is expected, once it's failed.

lisa/tools/mkfs.py
```diff
@@ -0,0 +1,53 @@
...
+    def command(self) -> str:
+        return self._command
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    # command - mkfs.xfs, mkfs.ext2, mkfs.ext3, mkfs.ext4
+    def mkfs(self, disk: str, command: str) -> None:
+        self._command = command
```
squirrelsc: Avoid to save it to `_command`. it's not useful.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +92,252 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+        description="""
+        This test case will do run commands 2-5, the commands are expected fail or not
+         based on device supports or not.
+        1. Use `nvme id-ctrl device` command list the capabilities of the device.
+        2. `nvme format namespace` - format a namespace.
+        3. `nvme create-ns namespace` - create a namespace.
+        4. `nvme delete-ns -n 1 namespace` - delete a namespace.
+        5. `nvme detach-ns -n 1 namespace` - detach a namespace.
+        """,
+        priority=2,
```
squirrelsc: Maybe some P2 test cases can be P3? We need to think about what's the impact, if a test case fails, and the cost to run the case.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +92,252 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+        nvme = node.features[Nvme]
+        nvme_namespaces = nvme.get_namespaces()
+        nvme_devices = nvme.get_devices()
+        nvme_cli = node.tools[Nvmecli]
+        nm_format_exit_code = 0
+        nm_management_exit_code = 0
+        # 1. Use `nvme id-ctrl device` command list the capabilities of the device.
+        if not nvme_cli.support_nm_format(nvme_devices[0]):
+            nm_format_exit_code = 1
+        if not nvme_cli.support_nm_manage_attach(nvme_devices[0]):
```
squirrelsc: Add more comments to explain the behavior.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +92,252 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+            format_namespace.assert_exit_code(nm_format_exit_code)
+            # 3. `nvme create-ns namespace` - create a namespace.
+            create_namespace = nvme_cli.create_namespace(namespace)
+            create_namespace.assert_exit_code(nm_management_exit_code)
+            # 4. `nvme delete-ns -n 1 namespace` - delete a namespace.
+            delete_namespace = nvme_cli.delete_namespace(namespace, 1)
+            delete_namespace.assert_exit_code(nm_management_exit_code)
+            # 5. `nvme detach-ns -n 1 namespace` - detach a namespace.
+            detach_namespace = nvme_cli.detach_namespace(namespace, 1)
+            detach_namespace.assert_exit_code(nm_management_exit_code)
```
squirrelsc: Update the `assert_exit_code`, to make it accept a customized message string, and join it with other information.

lisa/util/process.py
```diff
@@ -23,11 +24,30 @@ class ExecutableResult:
...
 
     def __str__(self) -> str:
         return self.stdout
 
+    def assert_exit_code(
+        self, expected_exit_code: int = 0, message: str = ""
+    ) -> AssertionBuilder:
+        display_message = f"get unexpected exit code on cmd {self.cmd}"
+        if message:
+            display_message += f", {message}"
```
squirrelsc: The user inputted message should be more important. So it should be put before the generated message.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -77,3 +92,262 @@ def nvme_basic_validation(self, environment: Environment, node: Node) -> None:
...
+            cmd_result = node.execute(
+                f"dd if=/dev/zero of={mount_point}/data bs=10M count=100",
+                shell=True,
+                sudo=True,
+            )
+            cmd_result.assert_exit_code()
+            initial_md5 = node.execute(
+                f"md5sum {mount_point}/data", shell=True, sudo=True
+            )
+            initial_md5.assert_exit_code()
```
squirrelsc: Add descriptions on the errors. Why it happens, what's wrong, how to fix it.

squirrelsc (approved):

PR #1370, opened 2021-06-22
variable, fix metadata may be lost during merging

squirrelsc:


lisa/variable.py
```diff
@@ -66,26 +66,42 @@ def load_variables(
...
+) -> None:
+    """
+    inplace update variables. If variables don't exist, will create them
+    """
+    for name, new_variable in new_variables.items():
+        variable = variables.get(name, None)
+        if variable:
+            variable.update(new_variable)
+        else:
+            variable = new_variable
```
sharsonia: why is this step needed?

squirrelsc: Good catch, it's not necessary.

sharsonia (approved):

PR #1376, opened 2021-06-23
Add more ethtool functionality

sharsonia:


lisa/tools/ethtool.py
```diff
@@ -78,26 +97,123 @@ class DeviceFeatures:
...
+
+        if not self.link_settings:
+            raise LisaException(
+                f"Could not get any link settings for device {interface}"
+                " in the defined pattern"
+            )
+
+
+class DeviceRingBufferSettings:
+    __ring_buffer_settings_pattern = re.compile(
```
squirrelsc: Add examples above.

lisa/tools/ethtool.py
```diff
@@ -23,26 +46,20 @@ class DeviceChannel:
...
     #   Other:          0
     #   Combined:       1
-    __max_settings_pattern = re.compile(
-        r"Pre-set maximums:(\s+)(?P<settings>.*?)Current hardware settings:",
-        re.DOTALL,
-    )
-    __current_settings_pattern = re.compile(
-        r"Current hardware settings:(\s+)(?P<settings>.*?)$", re.DOTALL
-    )
     __channel_count_param_pattern = re.compile(
```
squirrelsc: Because it has no subclass, so one underscore is enough to a private member.

squirrelsc (approved):

PR #1378, opened 2021-06-24
Use TIMESTAMP in log path structure

pupacha:
Removed VMRoleName as, for Tests with multiple VMs can make the path long and used timestamp instead.

Ex: C:\Users\lizzha\lisa\TestResults\20210623235547-YN69\PERF-NETWORK-TCP-SINGLE-CONNECTION-THROUGHPUT-SYNTHETIC\LISAv2-penguinator-test-YN69-0623235803-role-1 LISAv2-penguinator-test-YN69-0623235803-role-0

This change will replace `LISAv2-penguinator-test-YN69-0623235803-role-1 LISAv2-penguinator-test-YN69-0623235803-role-0` with a timestamp.

### DIFF
--- a/TestControllers/TestController.psm1
+++ b/TestControllers/TestController.psm1
@@ -469,7 +469,8 @@ Class TestController {
 		try {
 			# Prepare test case log folder
 			$currentTestName = $($CurrentTestData.testName)
-			$CurrentTestLogDir = "$global:LogDir\$currentTestName\$($VmData.RoleName)"
+			$uniqueId = New-TimeBasedUniqueId
+			$CurrentTestLogDir = "$global:LogDir\$currentTestName\$uniqueId"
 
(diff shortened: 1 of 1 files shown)


lizzha (approved):

TestControllers/TestController.psm1
```diff
@@ -469,7 +469,8 @@ Class TestController {
 		try {
 			# Prepare test case log folder
 			$currentTestName = $($CurrentTestData.testName)
-			$CurrentTestLogDir = "$global:LogDir\$currentTestName\$($VmData.RoleName)"
+			$uniqueId = New-TimeBasedUniqueId
+			$CurrentTestLogDir = "$global:LogDir\$currentTestName\$uniqueId"
```
squirrelsc, line 473: Please check if it works with v3

pupacha: @squirrelsc It works with v3.

PR #1379, opened 2021-06-24
Update GetDistro function

KsenijaS:
Update GetDisto function so it recognizes CBLD as part od Debian OS family.


squirrelsc (approved):

KsenijaS:
LGTM!

PR #1380, opened 2021-06-24
Add a utility script to scan Azure subscription for VM sizes and quota needed by LISA

lesscodingmorehappiness:
…ipeline to scan subscription and record unavailabe vm or insufficient vm quota

Tested in cloned pipeline

### DIFF
--- a/Utilities/ScanSubscription.ps1
+++ b/Utilities/ScanSubscription.ps1
@@ -1,0 +1,362 @@
+# Description: This script is used for chercking enabled VM sizes and VM core quota in the given azure subscription.
+# It will upload subscription id and unavailable VM/insufficient VM core to db
+# This script will be executed by 'Penguinator-Validate-Azure-Subscription-Info' pipeline
+# This script should be under .\lisav2\Utilities directory
+
+$clientSecret = "$(CLIENT_SECRET)"
+$clientID = $env:CLIENT_ID
+$tenantID = $env:TENANT_ID
+$subID = $env:SUBSCRIPTION_ID
+$lisaTestCasesXMLPath = ".\lisav2\XML\TestCases"
+$lisaVMConfigsPath = ".\lisav2\XML\VMConfigurations"
+$penguinatorSecureFilePath = $env:PENGUINATOR_SECUREFILEPATH
+$AllLisaTests = [System.Collections.ArrayList]::new()
+$script:AllTestVMSizes = @{}
+$script:InsufficientVMSizes = [System.Collections.ArrayList]::new()
+
+Import-Module .\lisav2\Libraries\TestLogs.psm1
+
+# Merge VM config files for simpler checking
+function Join-VMConfigFiles ($filePath) {
+    $files = Get-ChildItem $filePath
+    $finalXml = "<TestSetup>"
+    foreach ($file in $files) {
+        $xml = [xml](Get-Content $file.FullName)    
+        $finalXml += $xml.TestSetup.InnerXml
+    }
+    $finalXml += "</TestSetup>"
+    ([xml]$finalXml).Save(".\lisav2\mergedVMConfig.xml")
+    return [xml]$finalXml
+}
+
+function Measure-SubscriptionCapabilities() {
+    Write-LogInfo "Measure VM capabilities of current subscription..."
+    if (!$script:SubscriptionVMResourceSkus -or !$script:TestableLocations) {
+        $regionScopeFromUser = @()
+        $RegionAndStorageMapFile = ".\lisav2\XML\RegionAndStorageAccounts.xml"
+        if (Test-Path $RegionAndStorageMapFile) {
+            $RegionAndStorageMap = [xml](Get-Content $RegionAndStorageMapFile)
+            $RegionAndStorageMap.AllRegions.ChildNodes | ForEach-Object { $regionScopeFromUser += $_.LocalName }
+        }
+
+        Set-Variable -Name SubscriptionVMResourceSkus -Option ReadOnly -Scope script `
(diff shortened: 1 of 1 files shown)


Utilities/ScanSubscription.ps1
```diff
@@ -0,0 +1,362 @@
+# Description: This script is used for chercking enabled VM sizes and VM core quota in the given azure subscription.
+# It will upload subscription id and unavailable VM/insufficient VM core to db
+# This script will be executed by 'Penguinator-Validate-Azure-Subscription-Info' pipeline
+# This script should be under .\lisav2\Utilities directory
+
+$clientSecret = "$(CLIENT_SECRET)"
+$clientID = $env:CLIENT_ID
+$tenantID = $env:TENANT_ID
+$subID = $env:SUBSCRIPTION_ID
```
lizzha: Can you pass the 4 variables as the parameters of this script?

Utilities/ScanSubscription.ps1
```diff
@@ -0,0 +1,362 @@
...
+$lisaTestCasesXMLPath = ".\lisav2\XML\TestCases"
+$lisaVMConfigsPath = ".\lisav2\XML\VMConfigurations"
+$penguinatorSecureFilePath = $env:PENGUINATOR_SECUREFILEPATH
```
lizzha: ditto

Utilities/ScanSubscription.ps1
```diff
@@ -0,0 +1,362 @@
+# Description: This script is used for chercking enabled VM sizes and VM core quota in the given azure subscription.
```
LiliDeng: checking

lesscodingmorehappiness: changed

Utilities/ScanSubscription.ps1
lesscodingmorehappiness: fixed

Utilities/ScanSubscription.ps1
lesscodingmorehappiness: fixed

Utilities/ScanSubscription.ps1
```diff
@@ -0,0 +1,390 @@
...
+# This script will be executed by 'Penguinator-Validate-Azure-Subscription-Info' pipeline
+# This script should be under .\lisav2\Utilities directory
+
+Param
+(
+    [String] $clientSecret,
+    [String] $clientID,
+    [String] $tenantID,
+    [String] $subID,
+    [String] $lisaTestCasesXMLPath = '.\lisav2\XML\TestCases',
```
lizzha: "lisav2" seems to be a custom folder, can we set it to be ".\xml\TestCases", and assume the current execution folder is the cloned repo folder?

Utilities/ScanSubscription.ps1
```diff
@@ -0,0 +1,390 @@
...
+    $finalXml += "</TestSetup>"
+    ([xml]$finalXml).Save(".\lisav2\mergedVMConfig.xml")
+    return [xml]$finalXml
+}
+
+function Measure-SubscriptionCapabilities() {
+    Write-LogInfo "Measure VM capabilities of current subscription..."
+    if (!$script:SubscriptionVMResourceSkus -or !$script:TestableLocations) {
+        $regionScopeFromUser = @()
+        $RegionAndStorageMapFile = ".\lisav2\XML\RegionAndStorageAccounts.xml"
```
lizzha: ditto

Utilities/ScanSubscription.ps1
```diff
@@ -0,0 +1,390 @@
...
+Param
+(
+    [String] $clientSecret,
+    [String] $clientID,
+    [String] $tenantID,
+    [String] $subID,
+    [String] $lisaTestCasesXMLPath = '.\lisav2\XML\TestCases',
+    [String] $lisaVMConfigsPath = '.\lisav2\XML\VMConfigurations',
+    [String] $penguinatorSecureFilePath
+)
```
lizzha, line 17: can we add one more parameter for parallel count, the required quota may change per parallel count.
Say the parallel count is n, then the minimal quota for a size is the sum of n cases with top n required cores.

lesscodingmorehappiness: Sure. 
To clarify this:
1. what should be the default n? 2?
2. after getting the sum of top n required cores, do we still want to round up to closest 100s?

Utilities/ScanSubscription.ps1
lesscodingmorehappiness: Currently (I mean in my test pipeline) I clone lisa to $executionFolder/lisav2 to differentiate lisa codes from other codes/files under execution folder. I prefer keep it as is.

lizzha: you can add "pushd lisa" in your pipeline job.
If we want to keep it as a generic utility, it's better not having the customized path, or we can pass "lisa" from parameter

Utilities/ScanSubscription.ps1
lizzha: currently please set n default to 4.
Yes we still need to round up to 100

Utilities/ScanSubscription.ps1
```diff
@@ -0,0 +1,437 @@
...
+(
+    [String] $clientSecret,
+    [String] $clientID,
+    [String] $tenantID,
+    [String] $subID,
+    [String] $lisaDir = '.',
+    [String] $lisaTestCasesXMLPath = "$lisaDir\XML\TestCases",
+    [String] $lisaVMConfigsPath = "$lisaDir\XML\VMConfigurations",
+    [String] $penguinatorSecureFilePath,
+    [int] $parrallelCount = 4
```
lizzha, line 16: parallel

lizzha (approved):

PR #1386, opened 2021-06-30
Add support for almalinux in lisav3

lubaihua33:
1. Add support for AlmaLinux in LISAv3;
2. Add ID_LIKE support, put the ID_LIKE match in the end.
3. If the vendor is CentOs, RedHat, XenServer, the _get_os_version method is overridden in Redhat already, it just need to check whether it's Fedora in _get_os_version of Fedora class. We removed duplicate code.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -551,7 +551,7 @@ class Redhat(Fedora):
 
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
-        return re.compile("^rhel|Red|Scientific|acronis|Actifio$")
+        return re.compile("^rhel|Red|AlmaLinux|Scientific|acronis|Actifio$")
 
     def _initialize_package_installation(self) -> None:
         # older images cost much longer time when update packages
@@ -606,7 +606,7 @@ class Redhat(Fedora):
             cmd="cat /etc/redhat-release", no_error_log=True
         )
         if cmd_result.exit_code == 0 and cmd_result.stdout != "":
-            for vendor in ["Red Hat", "CentOS", "XenServer"]:
+            for vendor in ["Red Hat", "CentOS", "XenServer", "AlmaLinux"]:
                 if vendor not in cmd_result.stdout:
                     continue
                 os_version.vendor = vendor
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -56,7 +56,15 @@ class OperatingSystem:
     __os_release_pattern_name = re.compile(
         r"^NAME=\"?([^\" \r\n]+)[^\" \n]*\"?\r?$", re.M
     )
+    # For example, the ID and ID_LIKE in /etc/os-release of AlmaLinux is:
+    # ID="almalinux"
+    # ID_LIKE="rhel centos fedora"
+    # The __os_release_pattern_id can match "almalinux"
+    # The __os_release_pattern_idlike can match "rhel"
     __os_release_pattern_id = re.compile(r"^ID=\"?([^\" \r\n]+)[^\" \n]*\"?\r?$", re.M)
+    __os_release_pattern_idlike = re.compile(
+        r"^ID_LIKE=\"?([^\" \r\n]+)[^\"\n]*\"?\r?$", re.M
+    )
     __redhat_release_pattern_header = re.compile(r"^([^ ]*) .*$")
     __debian_issue_pattern = re.compile(r"^([^ ]+) ?.*$")
     __release_pattern = re.compile(r"^DISTRIB_ID='?([^ \n']+).*$", re.M)
@@ -142,6 +150,7 @@ class OperatingSystem:
         cmd_result = typed_node.execute(cmd="cat /etc/os-release", no_error_log=True)
         yield get_matched_str(cmd_result.stdout, cls.__os_release_pattern_name)
         yield get_matched_str(cmd_result.stdout, cls.__os_release_pattern_id)
+        cmd_result_os_release = cmd_result
 
         # for RedHat, CentOS 6.x
         cmd_result = typed_node.execute(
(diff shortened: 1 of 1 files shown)


lubaihua33:
LISAv3 test cases "smoke_test", "lsvmbus_count_devices_channels", "timesync_check_unbind_clocksource", "timesync_check_unbind_clockevent" have been passed running on AlmaLinux.

lisa/operating_system.py
```diff
@@ -142,6 +145,7 @@ def _get_detect_string(cls, node: Any) -> Iterable[str]:
         cmd_result = typed_node.execute(cmd="cat /etc/os-release", no_error_log=True)
         yield get_matched_str(cmd_result.stdout, cls.__os_release_pattern_name)
         yield get_matched_str(cmd_result.stdout, cls.__os_release_pattern_id)
+        yield get_matched_str(cmd_result.stdout, cls.__os_release_pattern_idlike)
```
squirrelsc: Can it be recognized by ID, instead of ID_LIKE?

lisa/operating_system.py
```diff
@@ -521,7 +525,7 @@ def _get_os_version(self) -> OsVersion:
             no_error_log=True,
         )
         if cmd_result.exit_code == 0 and cmd_result.stdout != "":
-            for vendor in ["Fedora", "CentOS", "Red Hat", "XenServer"]:
+            for vendor in ["Fedora", "CentOS", "Red Hat", "XenServer", "AlmaLinux"]:
```
squirrelsc: Over Check if the list of string is needed. This method is overridden in Redhat already, it just need to check whether it's fedora. If so, fix it in another commit.

lisa/operating_system.py
```diff
@@ -636,6 +640,12 @@ def name_pattern(cls) -> Pattern[str]:
         return re.compile("^CentOS|Centos|centos|clear-linux-os$")
 
 
+class AlmaLinux(Redhat):
```
squirrelsc: We don't need a class for AlmaLinux so far, because it has no special logic than redHat

squirrelsc:
The two commits have the same purpose, they should be merged.

lisa/operating_system.py
```diff
@@ -57,6 +57,9 @@ class OperatingSystem:
         r"^NAME=\"?([^\" \r\n]+)[^\" \n]*\"?\r?$", re.M
     )
     __os_release_pattern_id = re.compile(r"^ID=\"?([^\" \r\n]+)[^\" \n]*\"?\r?$", re.M)
+    __os_release_pattern_idlike = re.compile(
```
dsrivastavv, line 65: Out of curiosity, what are values for set for ID and ID_LIKE in almalinux?

lisa/operating_system.py
lubaihua33: If we don't add Class almalinux or add "almalinux" name_pattern under class RedHat, almalinux can't be recognized by ID.

lisa/operating_system.py
lubaihua33: Yes, this method is overridden in Redhat already.

lisa/operating_system.py
squirrelsc: 1. We should add name_pattern under class RedHat.
2. Can the almaLinux recognized by ID? What's value in ID and ID_LIKE?

lisa/operating_system.py
squirrelsc: Yes, please fix it. Remove duplicate code in a separated commit.

lisa/operating_system.py
lubaihua33:
```
cat /etc/os-release
NAME="AlmaLinux"
VERSION="8.4 (Electric Cheetah)"
**ID="almalinux"
ID_LIKE="rhel centos fedora"**
VERSION_ID="8.4"
PLATFORM_ID="platform:el8"
PRETTY_NAME="AlmaLinux 8.4 (Electric Cheetah)"
ANSI_COLOR="0;34"
CPE_NAME="cpe:/o:almalinux:almalinux:8.4:GA"
HOME_URL="https://almalinux.org/"
DOCUMENTATION_URL="https://wiki.almalinux.org/"
BUG_REPORT_URL="https://bugs.almalinux.org/"

ALMALINUX_MANTISBT_PROJECT="AlmaLinux-8"
ALMALINUX_MANTISBT_PROJECT_VERSION="8.4"
```

lubaihua33:
```
NAME="AlmaLinux"
VERSION="8.4 (Electric Cheetah)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
```

lisa/operating_system.py
lubaihua33:
```
NAME="AlmaLinux"
VERSION="8.4 (Electric Cheetah)"
ID="almalinux"
ID_LIKE="rhel centos fedora"
```

lisa/operating_system.py
```diff
@@ -551,7 +551,7 @@ class Redhat(Fedora):
 
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
-        return re.compile("^rhel|Red|Scientific|acronis|Actifio$")
+        return re.compile("^rhel|Red|AlmaLinux|almalinux|Scientific|acronis|Actifio$")
```
squirrelsc: Add one of them is enough. Not to add unnecessary things.

lisa/operating_system.py
squirrelsc: Add an example above.

lisa/operating_system.py

lisa/operating_system.py
lubaihua33: Have changed.

lisa/operating_system.py
lubaihua33: Have added an example as a comment above the two lines.

squirrelsc (approved):

PR #1387, opened 2021-06-30
Initial DPDK suite structure review

mcgov:
Create a test suite and testing runbook for DPDK, test does nothing at the moment.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -1,0 +1,341 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import PurePath
+
+from assertpy import assert_that
+
+from lisa import Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa.features import Sriov
+from lisa.testsuite import simple_requirement
+from lisa.operating_system import Ubuntu, Redhat, CentOs
+from lisa.tools import Git, Wget
+from lisa.util import SkippedException
+from typing import List, Dict
+import itertools
+
+
+class NicInfo:
+    def __init__(self, upper: str, lower: str = "", pci_slot: str = "") -> None:
+        self._has_secondary = lower != "" or pci_slot == ""
+        self._upper = upper
+        self._lower = lower
+        self._pci_slot = pci_slot
+
+    def has_secondary(self) -> bool:
+        return self._has_secondary
+
+    def __str__(self) -> str:
+        return f"""
+NicInfo:
+upper: {self._upper}
+lower: {self._lower}
+pci_slot: {self._pci_slot}
+        """
+
+    def testpmd_include(self, vdev_type: str) -> str:
+        assert_that(self._has_secondary).is_true().described_as(
+            f"This interface {self._upper} does not have a lower interface and pci slot associated with it. Aborting."
+        )
+        return f'--vdev="{vdev_type},iface={self._upper}" --allow "{self._pci_slot}"'
+
+
+class NodeNicInfo:
+    _nics: Dict[str, NicInfo] = dict()
+
+    def __init__(self, initializer: List[NicInfo] = []):
+        for nic in initializer:
+            self._nics[nic._upper] = nic
+
+    def append(self, next_node: NicInfo) -> None:
+        self._nics[next_node._upper] = next_node
+
+    def IsEmpty(self) -> bool:
+        return len(self._nics) == 0
+
+    def __len__(self) -> int:
+        return len(self._nics)
+
+    def get_upper_nics(self) -> List[str]:
+        return [self._nics[x]._upper for x in self._nics.keys()]
+
+    def get_lower_nics(self) -> List[str]:
+        return [self._nics[x]._lower for x in self._nics.keys()]
+
+    def get_device_slots(self) -> List[str]:
+        return [self._nics[x]._pci_slot for x in self._nics.keys()]
+
+    def get_nic(self, nic_name: str) -> NicInfo:
+        return self._nics[nic_name]
+
+    def nic_info_is_present(self, nic_name: str) -> bool:
+        return nic_name in self.get_upper_nics() or nic_name in self.get_lower_nics()
+
+    def __str__(self) -> str:
+        _str = ""
+        for nic in self._nics:
+            _str += f"{nic}"
+        return _str
+
+
+@TestSuiteMetadata(
+    area="dpdk",
+    category="functional",
+    description="""
+    This test suite check DPDK functionality
+    """,
+)
+class dpdk(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            This test case checks DPDK can be built and installed correctly.
+        """,
+        requirement=simple_requirement(
+            supported_features=[Sriov],
+        ),
+        priority=1,
+    )
+    def check_dpdk_build(self, case_name: str, node: Node) -> None:
+
+        sriov_feature = node.features[Sriov]
+        sriov_is_enabled = sriov_feature.enabled()
+        self.log.info(f"Verify SRIOV is enabled: {sriov_is_enabled}")
+        assert_that(sriov_is_enabled).is_true().described_as(
+            "SRIOV was not enabled for this test node."
+        )
+        self._install_dpdk_dependencies(node)
+        self._hugepages_init(node)
+        self._hugepages_enable(node)
+        self._install_dpdk(node)
+        nics = self._get_nic_names(node)
+        node_nic_info = self._get_node_nic_info(node, nics)
+        assert_that(len(node_nic_info)).described_as(
+            "Test needs at least 2 NICs on the test node."
+        ).is_greater_than_or_equal_to(2)
+        test_nic = node_nic_info.get_nic(node_nic_info.get_upper_nics()[-1])
+        vdev_type = "net_vdev_netvsc"
+        self.log.info(
+            f"/usr/local/bin/dpdk-testpmd -l 2,3 -n 4 --proc-type=primary {test_nic.testpmd_include(vdev_type)} -- --forward-mode=txonly -a"
+        )
+        timer_proc = node.execute_async(
+            "sleep 30 && killall -s INT /usr/local/bin/dpdk-testpmd",
+            sudo=True,
+            shell=True,
+        )
+        testpmd_proc = node.execute_async(
+            f"/usr/local/bin/dpdk-testpmd -l 2,3 -n 4 --proc-type=primary {test_nic.testpmd_include(vdev_type)} -- --forward-mode=txonly -a",
+            sudo=True,
+        )
+        timer_proc.wait_result()
+        proc_result = testpmd_proc.wait_result()
+        self.log.info(proc_result.stdout)
+
+    _ubuntu_packages = [
+        "librdmacm-dev",
+        "librdmacm1",
+        "build-essential",
+        "libnuma-dev",
+        "libmnl-dev",
+        "libelf-dev",
+        "meson",
+        "rdma-core",
+        "librdmacm-dev",
+        "librdmacm1",
+        "build-essential",
+        "libnuma-dev",
+        "libmnl-dev",
+        "libelf-dev",
+        "dpkg-dev",
+        "pkg-config",
+        "python3-pip",
+        "python3-pyelftools",
+        "python-pyelftools",
+    ]
+    _redhat_packages = [
+        "gcc",
+        "make",
+        "git",
+        "tar",
+        "wget",
+        "dos2unix",
+        "psmisc",
+        "kernel-devel",
+        "numactl-devel.x86_64",
+        "librdmacm-devel",
+        "pkgconfig",
+        "libmnl-devel",
+        "elfutils-libelf-devel",
+        "python3-pip",
+    ]
+    _rte_target = "x86_64-native-linuxapp-gcc"
+    _dpdk_github = "https://github.com/DPDK/dpdk.git"
+    _ninja_url = (
+        "https://github.com/ninja-build/ninja/releases/download/v1.10.2/ninja-linux.zip"
+    )
+
+    _ip_addr_regex = r"[0-9]+:\s+([a-zA-Z0-9\-_]+):\s+<(?:[a-zA-Z0-9_]+,?)+>"
+
+    def _install_dpdk_dependencies(self, node: Node) -> None:
+
+        if isinstance(node.os, Ubuntu):
+            for package in self._ubuntu_packages:
+                node.os.install_packages(package)
+            self.log.info("Packages installed for Ubunutu")
+            self._execute_expect_zero(node, "pip3 install --upgrade meson")
+            self._execute_expect_zero(node, "mv /usr/bin/meson /usr/bin/meson.bak")
+            self._execute_expect_zero(node, "ln -s /usr/local/bin/meson /usr/bin/meson")
+            self._execute_expect_zero(node, "pip3 install --upgrade ninja")
+            self._execute_expect_zero(
+                node, "modprobe -a rdma_cm rdma_ucm ib_core ib_uverbs mlx4_core mlx4_ib"
+            )
+
+        elif isinstance(node.os, Redhat) or isinstance(node.os, CentOs):
+            self._execute_expect_zero(
+                node, "yum update -y --disablerepo='*' --enablerepo='*microsoft*'"
+            )
+            node.os.install_packages(
+                ["groupinstall", "'Infiniband Support'"], signed=False
+            )  # todo gross hack to support groupinstall
+            for package in self._redhat_packages:
+                node.os.install_packages(package)
+            result = node.execute(
+                "dracut --add-drivers 'mlx4_en mlx4_ib mlx5_ib' -f"
+            )  # add mellanox drivers
+            self.log.debug("\n".join([result.stdout, result.stderr]))
+            self._execute_expect_zero(node, "systemctl enable rdma")
+            self._execute_expect_zero(node, "pip3 install --upgrade meson")
+            self._execute_expect_zero(node, "ln -s /usr/local/bin/meson /usr/bin/meson")
+
+            wget_tool = node.tools[Wget]
+            wget_tool.get(self._ninja_url)
+            node.execute(f"mv ninja-linux.zip {node.working_path}/")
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdk.yml
```diff
@@ -0,0 +1,5 @@
+name: dpdk
```
squirrelsc: Don't commit the runbook for testing. You can put it in some ignored folder to avoid it's included.

mcgov: Is there a reason we don't include these? It was sort of a pain to figure out and for others developing tests it might help to have examples that select single tests or run specific tests on other platforms.

squirrelsc: We want to provide simple experience to users, who run test cases or who develop test cases. It's a good feedback that developer need know how to debug one case.
Would you file an issue to track it? We can add document to help on it.
The current dpdk.yml applies to dpdk+azure only, it's not extendable. We need a yaml file for one case and other platforms.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -63,42 +63,105 @@ def check_dpdk_build(self, case_name: str, node: Node) -> None:
...
+            self._execute_expect_zero(
+                node,
+                f"wget {self._ninja_url}",
+            )
+            self._execute_expect_zero(
+                node, "unzip ninja-linux.zip && mv ninja /usr/bin/ninja"
+            )
+            self._execute_expect_zero(node, "pip3 install --upgrade pyelftools")
+
+    def _execute_expect_zero_with_path(
```
squirrelsc, line 222: 1. There is a method of result, `assert_exit_code`. It has similar function.
2. Make sure the stdout and stderr is not too long to output. The message will be saved with test results together.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -63,42 +63,105 @@ def check_dpdk_build(self, case_name: str, node: Node) -> None:
...
+                node.os.install_packages(package)
             result = node.execute(
                 "dracut --add-drivers 'mlx4_en mlx4_ib mlx5_ib' -f"
             )  # add mellanox drivers
             self.log.debug("\n".join([result.stdout, result.stderr]))
+            self._execute_expect_zero(node, "systemctl enable rdma")
+            self._execute_expect_zero(node, "pip3 install --upgrade meson")
+            self._execute_expect_zero(node, "ln -s /usr/local/bin/meson /usr/bin/meson")
+
+            self._execute_expect_zero(
```
squirrelsc: Use the wget tool to download a file.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -41,6 +40,7 @@ def check_dpdk_build(self, case_name: str, node: Node) -> None:
         self._hugepages_init(node)
```
squirrelsc, line 107: Add comments, what's validated or setup in each step.

squirrelsc: Those lines doesn't have verification, it do nothing to check.

```python
assert_that(sriov_is_enabled).described_as(
            "SRIOV was not enabled for this test node."
        )
```

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -36,11 +37,13 @@ def check_dpdk_build(self, case_name: str, node: Node) -> None:
         assert_that(sriov_is_enabled).described_as(
             "SRIOV was not enabled for this test node."
         )
-        self._install_dpdk_dependencies(node)
+        # self._install_dpdk_dependencies(node)
```
squirrelsc: delete it directly, if it's not necessary.

microsoft/testsuites/dpdk/dpdk.yml
squirrelsc: This file can be put in .vscode or other ignored folders, so that won't be submitted incidentally.

microsoft/testsuites/dpdk/dpdksuite.py
glima: Are you sure ninja is not packaged for those?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+            self._execute_expect_zero(node, "ln -s /usr/local/bin/meson /usr/bin/meson")
+
+            self._execute_expect_zero(
+                node,
+                f"wget {self._ninja_url}",
+            )
+            self._execute_expect_zero(
+                node, "unzip ninja-linux.zip && mv ninja /usr/bin/ninja"
+            )
+            self._execute_expect_zero(node, "pip3 install --upgrade pyelftools")
```
glima, line 216: Shouldn't you skip the test if none of those distros apply?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def _execute_expect_zero_with_path(
+        self, node: Node, cmd: str, path: PurePath
+    ) -> str:
+        result = node.execute(cmd, sudo=True, cwd=path, shell=True)
+        result.assert_exit_code()
+        self.log.info(result.stdout)  # TODO: debug
```
glima: Yeah, not a good idea to have this at info level (which is default triggered)

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+
+        return result.stdout
+
+    def _execute_expect_zero(self, node: Node, cmd: str) -> str:
```
glima, line 240: You can get away with one method only and default value, no?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+
+    def _hugepages_enable(self, node: Node) -> None:
+        self._execute_expect_zero(
+            node,
+            "echo 4096 > /sys/devices/system/node/node0/hugepages/hugepages-2048kB/nr_hugepages",
+        )
+        self._execute_expect_zero(
+            node,
+            "echo 1 > /sys/devices/system/node/node0/hugepages/hugepages-1048576kB/nr_hugepages",
```
glima: OP is right about having a specific runbook requirement here, or we'd have to have some automatic ARM templates to kick in per test suite/criteria. These are huge mem requirements, not met by your average Azure VM.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+        )
+        result = node.execute("grep -i huge /proc/meminfo && ls /mnt/", shell=True)
+        self.log.info(f"hugepages status \n{result.stdout}")
+
+    def _install_dpdk(self, node: Node) -> None:
+        git_tool = node.tools[Git]
```
glima, line 266: Again, packaged dpdk by distros not enough here? It's a default config build, apparently, so... Might as well be there equally for free?

microsoft/testsuites/dpdk/dpdksuite.py
glima: BTW, you've packed a bunch of things in that one commit, why split this random small part out of that? Just squash and be happy.

microsoft/testsuites/dpdk/dpdk.yml
mcgov: will remove

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+                node.os.install_packages(package)
+            self.log.info("Packages installed for Ubunutu")
+            self._execute_expect_zero(node, "pip3 install --upgrade meson")
+            self._execute_expect_zero(node, "mv /usr/bin/meson /usr/bin/meson.bak")
+            self._execute_expect_zero(node, "ln -s /usr/local/bin/meson /usr/bin/meson")
+            self._execute_expect_zero(node, "pip3 install --upgrade ninja")
+
+        elif isinstance(node.os, Redhat) or isinstance(node.os, CentOs):
+            self._execute_expect_zero(
+                node, "yum update -y --disablerepo='*' --enablerepo='*microsoft*'"
```
squirrelsc, line 194: There should be a tool of dpdk, so the dependencies will be managed by the tool. Some logic is already in os and base executable class, so more code can be reused. If the tool is considered to be used by dpdk only, it can be in the same dpkd folder.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def _execute_expect_zero_with_path(
+        self, node: Node, cmd: str, path: PurePath
+    ) -> str:
+        result = node.execute(cmd, sudo=True, cwd=path, shell=True)
+        assert_that(result.exit_code).described_as(
+            f"{cmd} failed with code {result.exit_code} and stdout+stderr:"
+            + f"\n{result.stdout}\n=============\n{result.stderr}\n=============\n"
+        ).is_zero()
+        self.log.info(f"{cmd}:\n{result.stdout}")  # TODO: debug
```
squirrelsc: This should be in debug level log already.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+            self._execute_expect_zero(node, "pip3 install --upgrade meson")
+            self._execute_expect_zero(node, "mv /usr/bin/meson /usr/bin/meson.bak")
+            self._execute_expect_zero(node, "ln -s /usr/local/bin/meson /usr/bin/meson")
+            self._execute_expect_zero(node, "pip3 install --upgrade ninja")
+
+        elif isinstance(node.os, Redhat) or isinstance(node.os, CentOs):
+            self._execute_expect_zero(
+                node, "yum update -y --disablerepo='*' --enablerepo='*microsoft*'"
+            )
+            node.os.install_packages(
```
squirrelsc, line 196: Does group install support in most distro? If so, we can add it in os level.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+                ["groupinstall", "'Infiniband Support'"], signed=False
+            )  # todo gross hack to support groupinstall
+            for package in self._redhat_packages:
+                node.os.install_packages(package)
+            result = node.execute(
+                "dracut --add-drivers 'mlx4_en mlx4_ib mlx5_ib' -f"
+            )  # add mellanox drivers
+            self.log.debug("\n".join([result.stdout, result.stderr]))
+            self._execute_expect_zero(node, "systemctl enable rdma")
```
squirrelsc, line 205: Lili created an abstract service management tool to support different distro. systemctl may not work on some distros.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+        elif isinstance(node.os, Redhat) or isinstance(node.os, CentOs):
+            self._execute_expect_zero(
+                node, "yum update -y --disablerepo='*' --enablerepo='*microsoft*'"
+            )
+            node.os.install_packages(
+                ["groupinstall", "'Infiniband Support'"], signed=False
+            )  # todo gross hack to support groupinstall
+            for package in self._redhat_packages:
+                node.os.install_packages(package)
+            result = node.execute(
```
squirrelsc, line 201: How about to create a tool `dracut`? It can be used by others.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+            self._execute_expect_zero(
+                node, "unzip ninja-linux.zip && mv ninja /usr/bin/ninja"
+            )
+            self._execute_expect_zero(node, "pip3 install --upgrade pyelftools")
+
+    def _execute_expect_zero_with_path(
+        self, node: Node, cmd: str, path: PurePath
+    ) -> str:
+        result = node.execute(cmd, sudo=True, cwd=path, shell=True)
+        assert_that(result.exit_code).described_as(
```
squirrelsc, line 227: Don't write general logic to dump all stdout to all cmd in assertion message. The stdout may be very long. It's hard to get useful information in long results. So review each assertion, to make sure it printed the appropriate level message.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+
+        return result.stdout
+
+    def _execute_expect_zero(self, node: Node, cmd: str) -> str:
+        return self._execute_expect_zero_with_path(node, cmd, node.working_path)
+
+    def _hugepages_init(self, node: Node) -> None:
+        self._execute_expect_zero(node, "mkdir -p /mnt/huge")
+        self._execute_expect_zero(node, "mkdir -p /mnt/huge-1G")
+        self._execute_expect_zero(node, "mount -t hugetlbfs nodev /mnt/huge")
```
squirrelsc, line 246: Lili wrote a tool called `mount` already. Please reuse it, and extend it if it's necessary.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+            self._execute_expect_zero(
+                node, "unzip ninja-linux.zip && mv ninja /usr/bin/ninja"
+            )
+            self._execute_expect_zero(node, "pip3 install --upgrade pyelftools")
+
+    def _execute_expect_zero_with_path(
+        self, node: Node, cmd: str, path: PurePath
+    ) -> str:
+        result = node.execute(cmd, sudo=True, cwd=path, shell=True)
```
squirrelsc, line 226: Avoid to execute each command on sudo level, until it's necessary.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+        nic_names = result.stdout.split("\r\n")
+        for item in nic_names:
+            assert_that(item).is_not_equal_to("").described_as(
+                "nic name could not be found"
+            )
+        self.log.info(f"network devices: {nic_names}")
+        return nic_names
+
+    def _get_nic_device(self, node: Node, nic_name: str) -> str:
+        nic_device = self._execute_expect_zero(
```
squirrelsc, line 292: It looks the `readlink` is worth to be a tool.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+        self, node: Node, nic_list: List[str]
+    ) -> Dict[str, str]:
+        # Identify which nics are slaved to master devices.
+        # This should be really simple with /usr/bin/ip but experience shows
+        # some platforms are buggy and require another method
+        master_nics = dict()
+        result = node.execute(
+            "echo 'export PATH=$PATH:/usr/sbin' >> ~/.bashrc", shell=True
+        )
+        self.log.info(result.stdout)
```
squirrelsc: The information is in debug level already. And don't output too much information on info level. The "info" level is a story for looking the flow. The debug information should be in "debug" level.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+        """,
+        requirement=simple_requirement(
+            supported_features=[Sriov],
+        ),
+        priority=1,
+    )
+    def check_dpdk_build(self, case_name: str, node: Node) -> None:
+        sriov_feature = node.features[Sriov]
+        sriov_is_enabled = sriov_feature.enabled()
+        self.log.info(f"Verify SRIOV is enabled: {sriov_is_enabled}")
```
squirrelsc, line 102: Refer to latest code, the log should be from the test case method, not from `self`.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,252 @@
...
+    @TestCaseMetadata(
+        description="""
+            This test case checks DPDK can be built and installed correctly.
+        """,
+        requirement=simple_requirement(
+            supported_features=[Sriov],
+        ),
+        priority=1,
+    )
+    def check_dpdk_build(self, case_name: str, node: Node) -> None:
```
squirrelsc, line 98: `case_name` is not used, it should be removed.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Ninja needs to be the latest version which usually isn't the packaged version

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Good point! will do that

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Thank you! I will tweak the info levels and squash before the final merge, I'm working on this one and the PR is just a rolling review before the final one 🙂

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: That's a good point, DPDK in general has some requirements like needing >=8 cores so I'll have to add something for memory too.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I agree and would never do this but most people aren't using the packaged versions apparently. The test also needs some flexibility to be able to validate new release candidates etc. This code will get tweaked to allow for any tagged release version to be installed and tested, so this section isn't finalized yet.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Makes sense. will refactor

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Added code to truncate the output in this wrapper.

mcgov:
Closing this PR since the code has changed so much, going to reopen for a real PR review after another refactor

PR #1391, opened 2021-07-05
Fix links and inline literals in links

WenyiQ:
- Fixed incorrect links
- Changed inline literals in links to prevent issues with rst 
(AFAIK, inline literals are currently not working properly in links in rst)

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -2,15 +2,15 @@
 
 - [Installation](#installation)
   - [How to use LISA in WSL](#how-to-use-lisa-in-wsl)
-  - [Cannot find package after run `poetry
-    install`](#cannot-find-package-after-run-poetry-install)
+  - [Cannot find package after run \`poetry
+    install\`](#cannot-find-package-after-run-poetry-install)
   - [Error: Poetry could not find a pyproject.toml
     file](#error-poetry-could-not-find-a-pyprojecttoml-file)
 - [Using VSCode](#using-vscode)
   - [Cannot find Python Interpreter by
     Poetry](#cannot-find-python-interpreter-by-poetry)
-  - [VSCode Python extension no longer supports `python.pythonPath` in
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -51,7 +51,7 @@ interpreter accordingly (in most cases open the path and look for
 poetry env info -p
 ```
 
-### VSCode Python extension no longer supports `python.pythonPath` in `setting.json`
+### VSCode Python extension no longer supports \`python.pythonPath\` in \`setting.json\`
```
squirrelsc: The \` is used to mark the text as code. If you want to quote something, you can use the ", not \`.

WenyiQ: Got it. Changed to the double quotation mark.

docs/troubleshooting.md
```diff
@@ -25,7 +25,7 @@ your `PATH` before the Windows version, otherwise this error will appear:
 
 > `/usr/bin/env: ‘python\r’: No such file or directory`
 
-### Cannot find package after run `poetry install`
+### Cannot find package after run \`poetry install\`
```
squirrelsc, line 28: should be changed to quotes?

squirrelsc (approved):

docs/troubleshooting.md
WenyiQ: But should this one be a command?

squirrelsc: It's not about a command or not. It's about how it's rendered. If you don't want it to be rendered like the `text` with a grey style, use quote "text" is like native language.

WenyiQ: Indeed, such rendering (in a link) would cause problems in later reStructuredText generation. However, I wanted it to be presented and understood as commands so I escaped graves by adding backslashes (\\). I think I'm fine with both ways and I will keep it in this way.

PR #1392, opened 2021-07-07
Create Azure shared resource if not present

### LINKED ISSUE #1388: Lisa azure platform fails if lisa_shared_resource doesn't exist
Running LisaV3 using azure platforms fails if `lisa_shared_resource ` doesn't exist. We should either explicitly state this dependency in docs or handle creation of this resource group programmatically.

dsrivastavv:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -687,6 +688,10 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
             credential=self.credential, subscription_id=self.subscription_id
         )
 
+        self._rm_client.resource_groups.create_or_update(
```
squirrelsc: check if it exists before call `create_or_update`. It may be concurrent issues, and it's slow to perform this operation.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -26,6 +26,7 @@
 
 AZURE = "azure"
 AZURE_SHARED_RG_NAME = "lisa_shared_resource"
+AZURE_SHARED_RG_LOCATION = "westus2"
```
squirrelsc: use the `RESOURCE_GROUP_LOCATION`. They are the same purpose.

dsrivastavv: Does this mean that resource groups which are created by LISA for hosting VM's always in "westus2"?

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: I am not sure if doing a `exist` check will solve the concurrency issue.  Two threads running in parallel can reach the `exist` check together and try creating the resource group. Maybe we can use locks to sync the threads. Thoughts?

squirrelsc: If the resource group doesn't exist, it may meet the concurrent issue. But it won't happen, once it exists. But if we call create_or_update each time, it may meet concurrent issue even it exists. If there is high concurrency like 30, it may reach the max deployment script error. So we need to reduce the chance to call create_or_update.

lisa/sut_orchestrator/azure/common.py
squirrelsc: It's used for resource group only, it doesn't impact where VM is deployed.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Makes sense!

squirrelsc (approved):

PR #1393, opened 2021-07-07
Azure: implement transformer to export VM from vhd

squirrelsc:
see details in commits


lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -0,0 +1,259 @@
...
+        node_context.vm_name = runbook.vm_name
+        node_context.resource_group_name = runbook.resource_group_name
+
+        node.set_connection_info_by_runbook()
+        node.initialize()
+
+        # prepare vm for exporting
+        wa = node.tools[Waagent]
+        wa.deprovision()
+        node.execute("export HISTSIZE=0", shell=True)
```
LiliDeng, line 151: have you tried this, seems after deprovision, the network of the VM is disconnected.

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -0,0 +1,259 @@
...
+            runbook.storage_account_name, runbook.container_name, platform.credential
+        )
+
+        path = _generate_vhd_path(container_client, runbook.file_name_part)
+        self._log.debug(f"copying vhd: {path}")
+        blob_client = container_client.get_blob_client(path)
+        operation = blob_client.start_copy_from_url(
+            sas_url, metadata=None, incremental_copy=False
+        )
+        while True:
```
LiliDeng: we may need add a timeout here in case azure platform hit some issue, then the loop will cost quite long time.

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: It works well, the VM is not disconnected.

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: good catch, will add it. How about 30 minutes? I saw it's about 3 minutes in same location for 30 GB.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -38,6 +38,10 @@ def get_version(self) -> str:
         found_version = find_patterns_in_lines(result.stdout, [self.__version_pattern])
         return found_version[0][0] if found_version[0] else ""
 
+    def deprovision(self) -> None:
```
LiliDeng, line 41: add one optional parameter to decide deprovision + user or not?
-deprovision[+user]

squirrelsc: What means [+user]? Why an optional parameter is needed?

Is it the exact command? `waagent -deprovision[+user]`.

LiliDeng: waagent -deprovision
not remove current user
waagent -deprovision+user 
remove current user

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -0,0 +1,265 @@
...
+        default="", metadata=schema.metadata(required=True)
+    )
+    vm_name: str = "node-0"
+
+    # values for SSH connection. public_address is optional, because it can be
+    # retrieved from vm_name. The private_key_file must be provided. Others can
+    # use default values.
+    public_address: str = ""
+    public_port: int = 22
+    username: str = "lisatest"
```
glima: Are you sure you want to commit this?

lisa/sut_orchestrator/azure/transformers.py
glima: Why that isolated HISTSIZE change? It will only work for that session.

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -0,0 +1,265 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from dataclasses import dataclass, field
+from pathlib import PurePosixPath
+from time import sleep
```
glima, line 6: For commit title: do you mean VM *to* VHD?

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: The logic is copied from V2. I think it's used to prevent future history after deprovisioning.

lisa/sut_orchestrator/azure/tools.py
squirrelsc: ok, will add it.

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: yes, the LISA is a test framework, it's the default user name, and be used in azure deployment too. Will add a constant to reuse it.

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: This transformer is to export VHD from a VM. It's expected. Let me know, what's confusing here.

lisa/sut_orchestrator/azure/transformers.py
glima: Yeah, but it will only apply to children of that shell session, which dies right after the command. It will do absolutely nothing to future SSH calls, regarding that.

lisa/sut_orchestrator/azure/transformers.py
glima: You wrote the exact opposite.

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: After this command, the VM will be shutdown, and export a VHD. There is no other sessions before shutdown.

glima: Which is even a stronger signal that that export is useless there, no?

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: ok, will change the "as" to "to".

lisa/sut_orchestrator/azure/transformers.py
squirrelsc: For the usage of this command, it's copied from v2. It should be used to truncate existing history.

lisa/sut_orchestrator/azure/tools.py
squirrelsc: tested it, after this command, the VM cannot be used again. In this transformer, keep the user. If it needs to export a clear vhd, we can add it later.

PR #1394, opened 2021-07-07
Add ntp and chrony test case.

LiliDeng:


lisa/executable.py
```diff
@@ -154,6 +154,13 @@ def _check_exists(self) -> bool:
         result = self.node.execute(
             f"{where_command} {self.command}", shell=True, no_info_log=True
         )
+        if 0 != result.exit_code:
```
squirrelsc: What's the reason to add this logic? If we really need this, it needs to add comments. How to make sure the following calls add sudo too?

lisa/operating_system.py
```diff
@@ -569,7 +569,7 @@ def _initialize_package_installation(self) -> None:
         #  Basic_A1, Standard_A5, Standard_A1_v2, Standard_D1
         # redhat rhel 7-lvm 7.7.2019102813 Basic_A1 cost 2371.568 seconds
         # redhat rhel 8.1 8.1.2020020415 Basic_A0 cost 2409.116 seconds
-        cmd_result = self._node.execute("yum -y update", sudo=True, timeout=3600)
+        cmd_result = self._node.execute("yum update", sudo=True, timeout=3600)
```
squirrelsc: Move it to a separated commit, and add commit description to explain the change.

lisa/tools/chrony.py
```diff
@@ -0,0 +1,55 @@
...
+        posix_os: Posix = cast(Posix, self.node.os)
+        package_name = "chrony"
+        posix_os.install_packages(package_name)
+        return self._check_exists()
+
+    def restart(self) -> None:
+        service = self.node.tools[Service]
+        cmd_result = service._restart_service("chrony")
+        if 0 != cmd_result.exit_code:
+            cmd_result = service._restart_service("chronyd")
```
squirrelsc: Add comments to provide examples of those different distros.

lisa/tools/chrony.py
```diff
@@ -0,0 +1,55 @@
...
+        cmd_result.assert_exit_code()
+
+    @retry(exceptions=LisaException, tries=10, delay=2)
```
squirrelsc: If the retry can succeed quickly, avoid to wait too long. 1 or 0.5 second is better.

lisa/tools/gcc.py
```diff
@@ -16,6 +16,13 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
+    def compile(self, file_name: str, out_put: str) -> None:
```
squirrelsc: `output`

lisa/tools/gcc.py
```diff
@@ -16,6 +16,13 @@ def command(self) -> str:
...
+        if file_name:
```
squirrelsc: should it be `output`? The file_name should be always exists. If the output is optional, provide default value on method declaration.

lisa/tools/hwclock.py
```diff
@@ -0,0 +1,27 @@
...
+class Hwclock(Tool):
+    @property
+    def command(self) -> str:
+        return "hwclock"
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        if not self._check_exists():
```
squirrelsc: It doesn't need to check in the install. it will be checked by platform.

lisa/tools/ntp.py
```diff
@@ -0,0 +1,43 @@
...
+        posix_os: Posix = cast(Posix, self.node.os)
+        package_name = "ntp"
+        posix_os.install_packages(package_name)
+        return self._check_exists()
+
+    @retry(tries=20, delay=1)
+    def check_leap_code(self) -> None:
+        cmd_result = self.run("-c rv", shell=True, sudo=True, force_run=True)
+        if not self.__leap_pattern.match(cmd_result.stdout):
+            raise LisaException("The leap code is not lean_none.")
```
squirrelsc: What happens, if the lead code is not lean_none? Add more explanation in the exception.

microsoft/testsuites/core/timesync.py
```diff
@@ -231,3 +231,44 @@ def timesync_check_unbind_clockevent(self, node: Node) -> None:
                     f"After unbind {clock_event_name}, current clock event should "
                     f"equal to [lapic]."
                 ).is_true()
+
+    @TestCaseMetadata(
+        description="""
+        This test is to check -
```
squirrelsc: add the summary

lisa/tools/gcc.py
LiliDeng: gcc test.c => a.out
gcc test.c -o test => test

squirrelsc: As the logic `if file_name`, the `file_name` is optional? I think the output is optional.

LiliDeng:
In some old kernel, we will encounter below issue, since the kernel doesn't contain the new code which support these parameters of ethtool, we may need skip these images.
ethtool -G eth0 rx 4078 tx 4078
Cannot set device ring parameters: Operation not supported

lisa/executable.py
LiliDeng: I will have a discussion with you for this.

lisa/operating_system.py
```diff
@@ -406,6 +399,20 @@ def _install_package_from_url(
         pkg = wget_tool.get(package, str(self._node.working_path))
         self.install_packages(pkg, signed)
 
+    def wait_running_process(self, process_name: str, timeout: int = 5) -> None:
```
squirrelsc, line 417: With my latest change, it may not be able to merge. It may need a pidof tool to wait.

lisa/tools/gcc.py
```diff
@@ -16,6 +16,12 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
+    def compile(self, filename: str, output: str = "") -> None:
```
squirrelsc: `output_name` or `output_path` may be clearer?

lisa/tools/systemctl.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+from lisa.util.process import ExecutableResult
+
+
+class Systemctl(Tool):
```
squirrelsc: Merge systemctl and service into a virtual tool. It can base on different distro to use one of them. It will be easier to use. The systemctrl and service can be separated tools or derived classes from the virtual tool, but user only need to create the virtual tool to operate services.

lisa/tools/ntpstat.py
```diff
@@ -0,0 +1,55 @@
...
+        self.node.shell.mkdir(tool_path, exist_ok=True)
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
+        gcc = self.node.tools[Gcc]
+        code_path = tool_path.joinpath("ntpstat")
+        gcc.compile(f"{code_path}/ntpstat.c", "ntpstat")
+        self._command = "./ntpstat"
+
+    def install(self) -> bool:
+        if not self._check_exists():
```
squirrelsc, line 43: It doesn't need to check in the install. it will be checked by platform.

lisa/tools/ntpstat.py
```diff
@@ -0,0 +1,55 @@
...
+            posix_os: Posix = cast(Posix, self.node.os)
+            package_name = "ntpstat"
+            posix_os.install_packages(package_name)
+            if not self._check_exists():
+                self._install_from_src()
+        return self._check_exists()
+
+    @retry(exceptions=LisaException, tries=10, delay=2)  # type: ignore
```
squirrelsc: Reduce the delay to 0.5 to speed up executing.

lisa/tools/systemctl.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+        return "active" == cmd_result.stdout
+
+    def stop_service(self, name: str) -> None:
+        if self.is_active(name):
+            cmd_result = self.run(f"stop {name}", shell=True, sudo=True, force_run=True)
+            cmd_result.assert_exit_code()
+
+    def restart_service(self, name: str) -> ExecutableResult:
+        return self.run(f"{name} restart", shell=True, sudo=True, force_run=True)
```
squirrelsc: Can the ExecutableResult be checked inside?

lisa/tools/ntpstat.py
```diff
@@ -0,0 +1,55 @@
...
+        git.clone(self.repo, tool_path)
+        gcc = self.node.tools[Gcc]
+        code_path = tool_path.joinpath("ntpstat")
+        gcc.compile(f"{code_path}/ntpstat.c", "ntpstat")
+        self._command = "./ntpstat"
+
+    def install(self) -> bool:
+        if not self._check_exists():
+            posix_os: Posix = cast(Posix, self.node.os)
+            package_name = "ntpstat"
```
squirrelsc: the variable is not necessary, because it's used in only one place.

microsoft/testsuites/core/timesync.py
```diff
@@ -231,3 +235,61 @@ def timesync_check_unbind_clockevent(self, node: Node) -> None:
...
+        priority=2,
+    )
+    def timesync_chrony(self, node: Node) -> None:
+        chrony = node.tools[Chrony]
+        # 1. Restart chrony service.
+        chrony.restart()
+        # 2. Check and set server setting in config file.
+        chrony.check_server_setting()
+        # 3. Restart chrony service to reload with new config.
+        chrony.restart()
```
squirrelsc, line 300: Does it need to check if restart correctly or not?

lisa/tools/ntp.py
```diff
@@ -0,0 +1,61 @@
...
+        cmd_result = self.run("-c rv", shell=True, sudo=True, force_run=True)
+        if not self.__leap_pattern.match(cmd_result.stdout):
+            # leap_none - normal synchronized state
+            # leap_alarm - NTP was never synchronized
+            raise LisaException(
+                "The leap code is not lean_none."
+                " Please check ntp server setting and service status."
+            )
+
+    def check_server_setting(self) -> None:
```
squirrelsc: The method content is not like "check", it's more like "set"

lisa/tools/ntp.py
```diff
@@ -27,6 +27,10 @@ def can_install(self) -> bool:
 
     def install(self) -> bool:
         posix_os: Posix = cast(Posix, self.node.os)
+        if isinstance(self.node.os, Redhat) and posix_os.release_version >= "8.0.0":
+            raise SkippedException(
+                f"Not support NTP on distro {posix_os.release_version}."
```
squirrelsc: 1. Provide reason, why it's not supported.
2. dump os name also, not only version.
3. Reword it like "The distro {} doesnt' support ntp, because ......"

lisa/tools/ntp.py
```diff
@@ -27,6 +27,10 @@ def can_install(self) -> bool:
 
     def install(self) -> bool:
         posix_os: Posix = cast(Posix, self.node.os)
+        if isinstance(self.node.os, Redhat) and posix_os.release_version >= "8.0.0":
+            raise SkippedException(
```
squirrelsc: Raise LisaException in tools, the SkippedException is used in test suite layer only.

lisa/tools/chrony.py
```diff
@@ -0,0 +1,70 @@
...
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.util import LisaException
+
+from .echo import Echo
+from .service import Service
+
+
+class Chrony(Tool):
+    __leap_status_pattern = re.compile(
```
squirrelsc, line 18: Add example above

lisa/tools/ntp.py
```diff
@@ -0,0 +1,65 @@
...
+from lisa.executable import Tool
+from lisa.operating_system import Posix, Redhat
+from lisa.tools import Echo
+from lisa.util import LisaException, SkippedException
+
+from .service import Service
+
+
+class Ntp(Tool):
+    __leap_pattern = re.compile(r"([\w\W]*?)leap_none.*$", re.MULTILINE)
```
squirrelsc, line 18: Add example

lisa/tools/chrony.py
```diff
@@ -0,0 +1,75 @@
...
+        elif isinstance(self.node.os, Redhat) or isinstance(self.node.os, Suse):
+            service_name = "chronyd"
+        else:
+            raise LisaException(
+                f"Please double check chrony service name in {posix_os.release_version}"
+            )
+        servicectl = self.node.tools[Servicectl]
+        servicectl = servicectl.initialize_tool()
+        servicectl.restart_service(service_name)
+        time.sleep(5)
```
squirrelsc: Don't write hard code sleep, it's not stable, and may cost extra time. Use a condition to check it, and sleep 0.5 sec.

lisa/tools/servicectl.py
```diff
@@ -0,0 +1,81 @@
...
+        return "systemctl"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def initialize_tool(self) -> Any:
```
squirrelsc: You can override check_exists to detect which is used, and new a real tool inside of this one. Create `stop_service`, and `restart_service`, and call the tool's methods accordingly.

lisa/tools/servicectl.py
```diff
@@ -0,0 +1,81 @@
...
+        if self.is_active(name):
+            cmd_result = self.run(f"stop {name}", shell=True, sudo=True, force_run=True)
+            cmd_result.assert_exit_code()
+
+    def restart_service(self, name: str) -> None:
+        cmd_result = self.run(f"restart {name}", shell=True, sudo=True, force_run=True)
+        cmd_result.assert_exit_code()
+
+
+class Servicectl(Tool):
```
squirrelsc: The name is weird, how about `service`? The original service can be renamed like `ServiceInternal`?

lisa/tools/servicectl.py
```diff
@@ -0,0 +1,81 @@
...
+        return "service"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def _check_service_running(self, name: str) -> bool:
```
squirrelsc: It's better to keep consistent between systemctl and service. both called `is_running`?

lisa/tools/chrony.py
```diff
@@ -0,0 +1,77 @@
...
+            service_name = "chronyd"
+        else:
+            posix_os: Posix = cast(Posix, self.node.os)
+            raise LisaException(
+                "Please double check chrony service name in distro "
+                f"{posix_os.name} {posix_os.release_version}"
+            )
+        servicectl = self.node.tools[Servicectl]
+        servicectl = servicectl.initialize_tool()
+        servicectl.restart_service(service_name)  # type: ignore
```
squirrelsc: Hold a sub-object in the Servicectl, and declare the same method `restart_service`.

lisa/tools/ntpstat.py
```diff
@@ -0,0 +1,53 @@
...
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._command = "ntpstat"
+
+    def _install_from_src(self) -> None:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages([Git, Gcc])
```
squirrelsc: It doesn't need to install them here, the `tools[Git]` will trigger the installation.

lisa/tools/__init__.py
```diff
@@ -44,11 +52,17 @@
...
     "Mount",
+    "Ntp",
+    "Ntpstat",
     "Ntttcp",
     "Nvmecli",
     "Reboot",
     "Sed",
     "Uname",
+    "Service",
+    "Systemctl",
```
squirrelsc: The SystemCtl and ServiceInternal doesn't need to be written here, because user doesn't need to use them directly.

lisa/tools/service.py
```diff
@@ -0,0 +1,98 @@
...
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _check_exists(self) -> bool:
+        cmd_result = self.node.execute(
+            "ls -lt /run/systemd/system", shell=True, sudo=True
+        )
+        if 0 == cmd_result.exit_code:
+            self.tool = self.node.tools[Systemctl]
```
squirrelsc: `self._internal_tool`, because it's a private variable

squirrelsc (approved):

PR #1395, opened 2021-07-08
Azure transformer: support deploy and delete transformers

squirrelsc:
With VHD transformers, the three transformers can deploy a VM and export to a VHD, then delete the VM. Other transformers can be applied to the deployed VM and then prepare a VHD. The VHD can be used on following all test cases. For example, install or build a kernel, change the LIS version and so on. 

Below is the example of runbook.

```yaml
transformer:
  - type: azure_deploy
    requirement:
      azure:
        marketplace: redhat rhel 7_9 7.9.2021051701
  - type: azure_vhd
    resource_group_name: $(azure_deploy_resource_group_name)
    rename:
      azure_vhd_url: vhd
  - type: azure_delete
    resource_group_name: $(azure_deploy_resource_group_name)
```

### DIFF
--- a/lisa/sut_orchestrator/azure/transformers.py
+++ b/lisa/sut_orchestrator/azure/transformers.py
@@ -73,6 +73,9 @@ class VhdTransformerSchema(schema.Transformer):
     container_name: str = DEFAULT_VHD_CONTAINER_NAME
     file_name_part: str = ""
 
+    # restore environment or not
+    restore: bool = False
+
 
 class VhdTransformer(Transformer):
     """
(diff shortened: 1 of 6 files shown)


lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -269,6 +286,80 @@ def _get_public_ip_address(
...
+        platform = _load_platform(self._runbook_builder, self.type_name())
+        runbook: DeployTransformerSchema = self.runbook
+
+        envs = Environments()
+        environment_requirement = EnvironmentSpace()
+        environment_requirement.nodes.append(runbook.requirement)
+        environment = envs.from_requirement(environment_requirement)
+        assert environment
+
+        platform.prepare_environment(environment=environment)
```
glima, line 318: Confused here: how do these calls interact/intersect with lisa_runner's calls of it?

squirrelsc: They are not related. The logic of platform is reused here to deploy/delete environments. The transformers are used to prepare environments before runners.

glima: I see. We'd better document that

PR #1402, opened 2021-07-09
add ethtool testsuite

sharsonia:


microsoft/testsuites/network/ethtool_suite.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@TestSuiteMetadata(
+    area="network",
+    category="functional",
+    description="""
+    This test suite runs the ethtool related network test cases.
+    """,
+)
+class ethtool(TestSuite):
```
squirrelsc: It looks this test suite should be called ringbuffer, not ethtool. The class name should be Camel case.

squirrelsc: If it will contain other network test cases, it may be called Networking?

sharsonia: It will contain many other ethtool related test cases which tests different capabilities and features of ethtool.

squirrelsc: Yes. It uses "ethtool" to test networking, not test "ethtool" itself.

microsoft/testsuites/network/ethtool_suite.py
```diff
@@ -0,0 +1,49 @@
...
+        ethtool.get_device_list()
+        for interface in ethtool._device_list:
+            changed_settings = ethtool.change_device_ring_buffer_settings(
+                interface, rx, tx
+            )
+
+            changed_rx_value = changed_settings.current_ring_buffer_settings["RX"]
+            assert_that(
+                changed_rx_value,
+                "Changing RX Ringbuffer setting didn't succeed",
```
squirrelsc: Is there any clue that why it happens? If so, it can be described here.

lisa/tools/ethtool.py
```diff
@@ -321,51 +321,69 @@ def get_device_link_settings(self, interface: str) -> DeviceLinkSettings:
...
 
     def get_device_ring_buffer_settings(
-        self, interface: str
+        self, interface: str, force: bool = False
     ) -> DeviceRingBufferSettings:
-        if interface in self._device_ring_buffer_settings_map.keys():
+        if (not force) and (interface in self._device_ring_buffer_settings_map.keys()):
             return self._device_ring_buffer_settings_map[interface]
 
         result = self.run(f"-g {interface}")
```
squirrelsc: pass force into the `run` too. There is an underlying cache in `run`.

lisa/tools/ethtool.py
```diff
@@ -243,7 +243,7 @@ def _install(self) -> bool:
         posix_os.install_packages("ethtool")
         return self._check_exists()
 
-    def _get_device_list(self) -> None:
+    def get_device_list(self) -> None:
```
squirrelsc: The method called `get_...`, so it should return values.

lisa/tools/ethtool.py
```diff
@@ -321,51 +321,69 @@ def get_device_link_settings(self, interface: str) -> DeviceLinkSettings:
...
     def get_device_ring_buffer_settings(
-        self, interface: str
+        self, interface: str, force: bool = False
     ) -> DeviceRingBufferSettings:
-        if interface in self._device_ring_buffer_settings_map.keys():
+        if (not force) and (interface in self._device_ring_buffer_settings_map.keys()):
             return self._device_ring_buffer_settings_map[interface]
 
         result = self.run(f"-g {interface}")
         if result.exit_code != 0:
```
squirrelsc: there is a new method in TestResult, called `assert_exit_code`, it can be used to simplify the code.

microsoft/testsuites/network/ethtool_suite.py
sharsonia: > Yes. It uses "ethtool" to test networking, not test "ethtool" itself.

It uses ethtool to test some features offered by ethtool to change networking settings, performance measurements etc

microsoft/testsuites/network/ethtool_suite.py
```diff
@@ -0,0 +1,49 @@
...
+            Steps:
+            1. Get the current ring buffer settings.
+            2. Change the rx and tx value to new_values using ethtool.
+            3. Get the settings again and validate the current rx and tx
+                values are equal to the new_values assigned.
+
+        """,
+        priority=1,
+    )
+    def change_ringbuffer_settings(self, node: Node, rx: int, tx: int) -> None:
```
squirrelsc: The test case cannot accept such customized arguments. Can they be detected from the distro or vm size?

squirrelsc: It looks this case is to test changes of rx/tx values. Can it get the default value, make some changes, and change it back? It doesn't need any extra arguments.

sharsonia: > It looks this case is to test changes of rx/tx values. Can it get the default value, make some changes, and change it back? It doesn't need any extra arguments.

Yes, I was just replying same to the earlier comment. I thought of doing what you said, just that existing LISAv2 test was passing these arguments. But on second thought, I am going to do what you have mentioned.

microsoft/testsuites/network/ethtool_suite.py
squirrelsc: It may be confusing to users of LISA test suites. The test suite is to test "networking settings" by "ethtool". The tested object is the networking settings. So it's a test suite about "networking settings", not a test suite about "ethtool". In some cases, we may need other tools to check the changes, which are made by ethtool.

LiliDeng:
In some old kernel, we will encounter below issue, since the kernel doesn't contain the new code which support these parameters of ethtool, we may need skip these images.
ethtool -G eth0 rx 4078 tx 4078
Cannot set device ring parameters: Operation not supported

squirrelsc:
> In some old kernel, we will encounter below issue, since the kernel doesn't contain the new code which support these parameters of ethtool, we may need skip these images.
> ethtool -G eth0 rx 4078 tx 4078
> Cannot set device ring parameters: Operation not supported

It can add a validation, and throw SkipException in this case.

lisa/tools/ethtool.py
```diff
@@ -308,65 +324,72 @@ def get_device_link_settings(self, interface: str) -> DeviceLinkSettings:
...
         self._device_ring_buffer_settings_map[interface] = device_ring_buffer_settings
 
         return device_ring_buffer_settings
 
+    def change_device_ring_buffer_settings(
+        self, interface: str, rx: int, tx: int
+    ) -> DeviceRingBufferSettings:
+        change_result = self.run(f"-G {interface} rx {rx} tx {tx}")
+        change_result.assert_exit_code(
+            message=f" Couldn't change device {interface} ring buffer settings."
```
squirrelsc, line 353: if the stdout is not long, and can explain what happens, it can be included in output.

microsoft/testsuites/network/networksettings.py
```diff
@@ -0,0 +1,141 @@
...
+    def validate_ringbuffer_settings_change(self, node: Node) -> None:
+        ethtool = node.tools[Ethtool]
+        devices_settings = ethtool.get_all_device_ring_buffer_settings()
+
+        for interface_settings in devices_settings:
+            interface = interface_settings.device_name
+            original_rx = int(interface_settings.current_ring_buffer_settings["RX"])
+            original_tx = int(interface_settings.current_ring_buffer_settings["TX"])
+
+            modified_rx = 10011
```
squirrelsc: The variable names can be `expected_rx`. It's clearer in test case what's expected or actual.

lisa/tools/ethtool.py
```diff
@@ -258,46 +258,62 @@ def _get_device_list(self) -> None:
...
             cmd_result = self.node.execute(f"ls {netdir}")
-            if cmd_result.exit_code != 0:
-                raise LisaException(
-                    "Could not find the network device under path {netdir}"
-                    f" exit_code: {cmd_result.exit_code}"
-                    f" stderr: {cmd_result.stderr}"
-                )
-            self._device_list.append(cmd_result.stdout)
+            cmd_result.assert_exit_code(
+                message=f"Could not find the network device under path {netdir}"
```
squirrelsc: The path will be output by the `assert_exit_code`, so it doesn't need to be repeated.

lisa/tools/ethtool.py
```diff
@@ -258,46 +258,62 @@ def _get_device_list(self) -> None:
...
-        result = self.run(f"-l {interface}")
-        if result.exit_code != 0:
-            raise LisaException(
-                f"ethtool -l {interface} command got non-zero"
-                f" exit_code: {result.exit_code}"
-                f" stderr: {result.stderr}"
-            )
+        result = self.run(f"-l {interface}", force_run=force)
+        if (result.exit_code != 0) and ("Operation not supported" in result.stderr):
+            raise SkippedException
```
squirrelsc: 1. The `SkippedException` should be in test case, not in tools, because the Skipped is a result for test case. The tool may be used in different test cases.
2. The reason must be specified, why it's skipped.

squirrelsc: You can raise a LISAException or a new type of Exception, and capture it in test case, and covert it to SkippedException. The two exceptions have different purpose and message, the inner exception is to specify the command failed, the skipped exception explain something doesn't support, so the test case is skipped.

lisa/tools/ethtool.py
```diff
@@ -258,46 +258,62 @@ def _get_device_list(self) -> None:
...
+        result.assert_exit_code(
+            message=f"Couldn't get device {interface} channels info."
+        )
+
+        if interface in self._device_channel_map.keys():
```
squirrelsc: The two lines are not necessary, in next lines, the value will be replaced.

microsoft/testsuites/network/networksettings.py
squirrelsc: 1. It's better to explain why those values are selected.
2. Is it possible the original value is the same as this one? If so, may be the test doesn't change values actually.

lisa/tools/ethtool.py
sharsonia: A new object is created in next line before inserting the value in list. So Need to delete the existing object.

squirrelsc: The item in dict will be replaced each time. `a[b] = "c"`

sharsonia: > The item in dict will be replaced each time. `a[b] = "c"`

Concern is not about replacing. I am creating a new object for this value "C". So I just want to delete the object that was in a[b]. Does that make sense?

squirrelsc: It doesn't need, the `del` keyword is to delete key in the dict, not to delete the object. And Python manage memory itself, it doesn't need to delete an object, unless it uses other resource than memory.

microsoft/testsuites/network/networksettings.py
```diff
@@ -0,0 +1,152 @@
...
+            "tcp-segmentation-offload",
+        ]
+        ethtool = node.tools[Ethtool]
+        devices_features = ethtool.get_all_device_enabled_features()
+
+        for device_features in devices_features:
+            enabled_features = device_features.enabled_features
+
+            if not all(feature in enabled_features for feature in required_features):
+                raise LisaException(
```
squirrelsc, line 144: Dump all existing features, so user knows which doesn't exist.

microsoft/testsuites/network/networksettings.py
```diff
@@ -0,0 +1,152 @@
...
+
+        """,
+        priority=1,
+    )
+    def validate_ringbuffer_settings_change(self, node: Node) -> None:
+        ethtool = node.tools[Ethtool]
+        try:
+            devices_settings = ethtool.get_all_device_ring_buffer_settings()
+        except UnsupportedOperationException:
+            raise SkippedException
```
squirrelsc: Add message for exceptions, so user knows why this case is skipped.

microsoft/testsuites/network/networksettings.py
```diff
@@ -0,0 +1,152 @@
...
+        except Exception:
```
squirrelsc: It doesn't need to rewrite exception here. If it's needed, it should include the raw exception, it's important message about why it's failed.

microsoft/testsuites/network/networksettings.py
```diff
@@ -0,0 +1,149 @@
...
+    )
+    def validate_device_enabled_features(self, node: Node) -> None:
+        required_features = [
+            "rx-checksumming",
+            "tx-checksumming",
+            "scater-gather",
+            "tcp-segmentation-offload",
+        ]
+        ethtool = node.tools[Ethtool]
+        devices_features = ethtool.get_all_device_enabled_features()
```
squirrelsc, line 138: Does this one need check skip or not? And if so, how about create a shared method in this test suite?

sharsonia: At least the existing LISAv2 test case doesn't have the check for skipping in this case. Not sure if there are any kernel/images that doesn't have support for this?

squirrelsc: If so, let's keep the same logic as v2

squirrelsc (approved):

PR #1404, opened 2021-07-12
Use 'yum makecache' replace 'yum update'

LiliDeng:
yum update equals to apt upgrade, is used to install newer packages.
yum makecache will download yum repository data to cache.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -304,6 +304,9 @@ class Posix(OperatingSystem, BaseClassMixin):
     ) -> None:
         raise NotImplementedError()
 
+    def _update_packages(self, packages: Optional[Union[List[str]]] = None) -> None:
+        raise NotImplementedError()
+
     def _package_exists(self, package: str, signed: bool = True) -> bool:
         raise NotImplementedError()
 
@@ -340,6 +343,21 @@ class Posix(OperatingSystem, BaseClassMixin):
 
         return os_version
 
+    def _get_package_list(
+        self, packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]]
+    ) -> List[str]:
+        package_names: List[str] = []
+        if not isinstance(packages, list):
+            packages = [packages]
+
+        assert isinstance(packages, list), f"actual:{type(packages)}"
+        for item in packages:
+            package_names.append(self.__resolve_package_name(item))
+        if self._first_time_installation:
+            self._first_time_installation = False
+            self._initialize_package_installation()
+        return package_names
+
     def _install_package_from_url(
         self,
         package: str,
@@ -358,17 +376,7 @@ class Posix(OperatingSystem, BaseClassMixin):
         packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]],
         signed: bool = True,
     ) -> None:
-        package_names: List[str] = []
-        if not isinstance(packages, list):
-            packages = [packages]
-
-        assert isinstance(packages, list), f"actual:{type(packages)}"
-        for item in packages:
-            package_names.append(self.__resolve_package_name(item))
-        if self._first_time_installation:
-            self._first_time_installation = False
-            self._initialize_package_installation()
-
+        package_names = self._get_package_list(packages)
         self._install_packages(package_names, signed)
 
     def package_exists(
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -554,33 +554,21 @@ def _get_os_version(self) -> OsVersion:
...
         return re.compile("^rhel|Red|AlmaLinux|Scientific|acronis|Actifio$")
 
     def _initialize_package_installation(self) -> None:
-        # older images cost much longer time when update packages
-        # smaller sizes cost much longer time when update packages, e.g.
-        #  Basic_A1, Standard_A5, Standard_A1_v2, Standard_D1
-        # redhat rhel 7-lvm 7.7.2019102813 Basic_A1 cost 2371.568 seconds
-        # redhat rhel 8.1 8.1.2020020415 Basic_A0 cost 2409.116 seconds
-        cmd_result = self._node.execute("yum -y update", sudo=True, timeout=3600)
+        cmd_result = self._node.execute("yum makecache", sudo=True)
```
squirrelsc, line 582: The original code can still keep, we can add a method like `update_packages`. If we need to update existing package, it can be called.

lisa/operating_system.py
```diff
@@ -554,33 +574,21 @@ def _get_os_version(self) -> OsVersion:
...
+        os_version = self._get_os_version()
         # we will hit expired RHUI client certificate issue on old RHEL VM image
         # use below solution to resolve it
         # refer https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/redhat/redhat-rhui#azure-rhui-infrastructure # noqa: E501
-        if cmd_result.exit_code != 0 and self.__rhui_error_pattern.match(
-            cmd_result.stdout
-        ):
+        if "Red Hat" == os_version.vendor and cmd_result.exit_code != 0:
```
squirrelsc, line 589: It doesn't need to check if it's Red Hat, because it's in RedHat class.

lisa/operating_system.py
```diff
@@ -554,33 +574,21 @@ def _get_os_version(self) -> OsVersion:
...
             cmd_result = self._node.execute(
-                "yum update -y --disablerepo='*' --enablerepo='*microsoft*' ",
-                sudo=True,
-                timeout=3600,
+                "yum update -y --disablerepo='*' --enablerepo='*microsoft*' ", sudo=True
```
squirrelsc, line 591: Should this command be moved to `_update_packages`?

LiliDeng: but centos also uses the same code?

lisa/operating_system.py
LiliDeng: but centos also uses the same code?

lisa/operating_system.py
LiliDeng: install_packages and update_packages will invoke _get_package_list, then _initialize_package_installation(), so whenever install or update packages invoked, this can be invoked?

lisa/operating_system.py
squirrelsc: Yes, It's intent to share the code. If CentOS or other derives has different behaviors, we can handle them in CentOS or other subclasses.

lisa/operating_system.py
squirrelsc: I think this command is used to update packages, not to refresh cache. Is it right? So it should be in update_packages.
`yum update -y --disablerepo...`

LiliDeng: no, the code is to fix the issue when repo cert expired on rhel, without this code, any yum command will fail.

lisa/operating_system.py
LiliDeng: then it is still necessary to have a check whether it is rhel or not.

lisa/operating_system.py
LiliDeng: update comments.

lisa/operating_system.py
```diff
@@ -477,6 +488,15 @@ def _get_os_version(self) -> OsVersion:
 
         return os_version
 
+    def _update_packages(self, packages: Optional[Union[List[str]]] = None) -> None:
+        command = "DEBIAN_FRONTEND=noninteractive apt-get upgrade -y "
+        command += (
```
squirrelsc: use below pattern is better, avoid to join strings.

```python
        command = (
            "DEBIAN_FRONTEND=noninteractive apt-get upgrade -y "
            '-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"'
        )
  
```

squirrelsc (dismissed):

lisa/operating_system.py
LiliDeng: Fixed

squirrelsc (approved):

PR #1405, opened 2021-07-12
Implement install kernel Transformers

squirrelsc:
See details in commits.

The sample runbook is,

**repo**
```yaml
  - type: azure_deploy
    requirement:
      azure:
        marketplace: Canonical 0003-com-ubuntu-server-trusted-vm 18_04 18.04.202007291
  - type: kernel_installer
    connection:
      address: $(azure_deploy_address)
    installer:
      type: repo
      source: linux-azure-edge
      is_proposed: false
```

```yaml
  - type: azure_deploy
    requirement:
      azure:
        marketplace: Canonical 0003-com-ubuntu-server-trusted-vm 18_04 18.04.202007291
  - type: kernel_installer
    connection:
      address: $(azure_deploy_address)
    installer:
      type: ppa
      source: linux-azure-fips
      openpgp_key: key to some repo
      ppa_url: url and credential to some repo
```


lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+
+        installer.validate()
+        installer.install()
+
+        self._log.info("rebooting")
+        node.reboot()
+        self._log.info(
+            f"kernel version after install: "
+            f"{uname.get_linux_information(force_run=True)}"
+        )
```
LiliDeng, line 138: we need compare the kernel version before and after install kernel.

lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+
+        return {}
+
+
+class RepoInstaller(BaseInstaller):
+    __menu_id_parts_pattern = re.compile(
+        r"^(?P<prefix>.*?)-.*-(?P<postfix>.*?-.*?-.*?-.*?-.*?-.*?)?$"
```
LiliDeng, line 148: add an example?

lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+        if runbook.is_proposed:
+            version_name = f"{release}-proposed"
+        else:
+            version_name = release
+        repo_entry = (
+            f"deb {self.repo_url} {version_name} "
+            f"restricted main multiverse universe"
+        )
+        ubuntu.wait_running_package_process()
+        result = node.execute(f'add-apt-repository "{repo_entry}"', sudo=True)
```
LiliDeng, line 196: add-apt-repository -y

lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+from lisa.tools import Cat, Uname
+from lisa.transformer import Transformer
+from lisa.util import get_matched_str, subclasses
+from lisa.util.logger import Logger, get_logger
+
+
+@dataclass_json(undefined=Undefined.INCLUDE)
+@dataclass
+class BaseInstallerSchema(schema.TypedSchema):
+    delay_parsed: CatchAll = field(default_factory=dict)  # type: ignore
```
glima, line 24: Why have this base? The field is being redeclared (it has to, given it's a pre-populated value one).

lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+@dataclass
+class RepoInstallerSchema(BaseInstallerSchema):
+    # version: str = "linux-image-azure-lts-18.04"
+    source: str = field(
+        default="proposed",
+        metadata=schema.metadata(
+            required=True,
+        ),
+    )
+    is_proposed: bool = True
```
glima, line 42: What is this about? Also the "proposed" string. Comments, at the very least?

lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+        metadata=schema.metadata(
+            required=True,
+        ),
+    )
+    is_proposed: bool = False
+
+    delay_parsed: CatchAll = field(default_factory=dict)  # type: ignore
+
+    def __post_init__(self, *args: Any, **kwargs: Any) -> None:
+        add_secret(self.ppa_url, PATTERN_HEADTAIL)
```
glima, line 61: URL as secret? Err, why?

lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+            r"\g<prefix>-\g<postfix>", submenu_id
+        )
+        assert menu_id, f"cannot composite menu id from {submenu_id}"
+
+        # composite boot menu in grub
+        menu_entry = f"{menu_id}>{submenu_id}"
+        self._log.debug(f"composited menu_entry: {menu_entry}")
+
+        ubuntu.set_boot_entry(menu_entry)
+        node.execute("update-grub", sudo=True)
```
glima, line 290: I still wonder how are those distros setup WRT sudo: *every one of them* with something like ``%wheel        ALL=(ALL)       NOPASSWD: ALL`` on /etc/sudoers? At least Fedora is not like that by default. Do people change, when ingressing images on Azure?

lisa/transformers/kernel_installer.py
```diff
@@ -0,0 +1,308 @@
...
+        # linux-image-4.18.0-1025-azure/bionic-updates,bionic-security,now
+        # 4.18.0-1025.27~18.04.1 amd64 [installed]
+        # output 4.18.0-1025
+        kernel_version_package_pattern = re.compile(
+            f"^{source}/[^ ]+ "
+            r"(?P<kernel_version>[^.]+\.[^.]+\.[^.-]+[.-][^.]+)\..*[\r\n]+",
+            re.M,
+        )
+        # the grep is used to remove color code
+        result = node.execute(f"apt search {source} | grep ''", shell=True)
```
glima: I thought we were filtering out all ansi escape chars for people for free already, why bother?

lisa/transformers/kernel_installer.py
squirrelsc: There is log to output the before/after installation. Considering sometime may be the version is latest already, so there is no assertion to check if the version is changed or not. If it's really needed, let me know, I can add a validation, and it can be skipped.

lisa/transformers/kernel_installer.py
squirrelsc: fixed

lisa/transformers/kernel_installer.py
squirrelsc: What's the reason to add it? It runs well.

lisa/transformers/kernel_installer.py
squirrelsc: Removed the redeclared fields in its subclasses. The base schema is used to parse type in factory, and save other fields in the delay_parsed.

lisa/transformers/kernel_installer.py
squirrelsc: add comments for fields. the proposed field is used by Canonical for kernels before official release.

lisa/transformers/kernel_installer.py
squirrelsc: There may be credential in the url.

lisa/transformers/kernel_installer.py
squirrelsc: The Azure images uses cloud-init or wala-agent to initialize. Users can specify to use password or private key. In LISA, it uses private key by default, and there is no password needed.

lisa/transformers/kernel_installer.py
squirrelsc: It applies to log only. I will move it to a method, and can support more. But it shouldn't apply to all stdout, since the raw content may be needed in some cases.

glima: Oh, true. Now I remember I apply to stdout by hand in my case.

PR #1407, opened 2021-07-12
add more tests to Xilinx FPGA testcase

hpourreza:
add few extra commands to test Xilinx FPGAs

### DIFF
--- a/Testscripts/Linux/fpga-card-validation.sh
+++ b/Testscripts/Linux/fpga-card-validation.sh
@@ -123,8 +123,35 @@ function validate_cards() {
         SetTestStateAborted
         return 1
     fi
+    echo "****************************" >> TestExecution.log
+    echo "Running xbutil scan" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
     xbutil scan >> TestExecution.log
+
+    echo "****************************" >> TestExecution.log
+    echo "Running xbutil validate" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
     xbutil validate >> TestExecution.log
+
+    echo "****************************" >> TestExecution.log
+    echo "Running xbutil host_mem -d 0 --enable --size 1g" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
+    xbutil host_mem -d 0 --enable --size 1g >> TestExecution.log
+
+    echo "****************************" >> TestExecution.log
+    echo "Running xbutil validate" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
+    xbutil validate >> TestExecution.log
+
+    echo "****************************" >> TestExecution.log
+    echo "Running xbutil reset" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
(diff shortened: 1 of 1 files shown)


Testscripts/Linux/fpga-card-validation.sh
```diff
@@ -123,8 +123,35 @@ function validate_cards() {
         SetTestStateAborted
         return 1
     fi
+    echo "****************************" >> TestExecution.log
```
LiliDeng: LogMsg "****************************" will reach the same goal.

Testscripts/Linux/fpga-card-validation.sh
```diff
@@ -123,8 +123,35 @@ function validate_cards() {
...
+    echo "Running xbutil scan" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
     xbutil scan >> TestExecution.log
+
+    echo "****************************" >> TestExecution.log
+    echo "Running xbutil validate" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
+    xbutil validate >> TestExecution.log
```
LiliDeng: xbutil validate >> TestExecution.log 2>&1
update redirect like this way, then error and info log will be written into TestExecution.log.

Testscripts/Linux/fpga-card-validation.sh
```diff
@@ -135,7 +162,7 @@ function validate_cards() {
 # Source utils.sh
 . utils.sh || {
     echo "ERROR: unable to source utils.sh!"
-    echo "TestAborted" > state.txt
```
LiliDeng: When we have issues during source utils.sh, we can't use function SetTestStateAborted

Testscripts/Linux/fpga-card-validation.sh
```diff
@@ -123,8 +123,35 @@ function validate_cards() {
         SetTestStateAborted
         return 1
     fi
+    echo "****************************" >> TestExecution.log
+    echo "Running xbutil scan" >> TestExecution.log
+    echo "****************************" >> TestExecution.log
     xbutil scan >> TestExecution.log
+
```
LiliDeng: should we have fail/pass check after each xbutil command?

Testscripts/Linux/fpga-card-validation.sh
hpourreza: Done

Testscripts/Linux/fpga-card-validation.sh
hpourreza: Done

Testscripts/Linux/fpga-card-validation.sh
hpourreza: I added the check for return value `$?` after each `xbutil` run

Testscripts/Linux/fpga-card-validation.sh
hpourreza: reverted back to what it was

hpourreza:
@LiliDeng 
Is there any other change that I need to do?

LiliDeng (approved):

PR #1411, opened 2021-07-13
Update index page from README & link fixes

WenyiQ:


docs/quick_start.rst
```diff
@@ -1,17 +1,37 @@
...
 
-LISA can be used to run test against the local node, or a remote node;
+**LISA** can be used to run test against the local node, or a remote node;
 if it is used to run against a remote node, you don’t need to configure
 anything on the remote node.
 
 .. figure:: img/deploy.svg
    :alt: deploy
 
+Why LISA
```
squirrelsc: What's the reason to put this part in quick_start, not along with other in index.rst?

WenyiQ: Because I want the index page to serve as a brief overview of the project, and with Why LISA there could be too much detail. Maybe we can move the image in quick_start to index page instead?

squirrelsc: The quick start page is for people, who know what's LISA and will try. The Why LISA is for people, who may not know LISA. So it should be in first page to explain value of LISA.

squirrelsc: The image in quick_start is also about how to install, not about value of LISA. The home page should explain the value of LISA.

WenyiQ: OK, I see. Will change it.

squirrelsc (approved):

PR #1421, opened 2021-07-16
Add support for testing image in shared gallery

dsrivastavv:
Runbook example : 

- If the shared image gallery is in the same subscription where LISA is run : 

```
azure:
        shared_gallery: "<image_gallery>/<image_definition>/<image_version>"
```

- If the shared image gallery is in a different subscription : 

```
azure:
        shared_gallery: "<subscription_id>/<image_gallery>/<image_definition>/<image_version>"
```

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -62,19 +62,37 @@ class AzureVmMarketplaceSchema:
 
 @dataclass_json()
 @dataclass
+class SharedImageGallerySchema:
+    subscription_id: str = ""
+    image_gallery: str = ""
+    image_definition: str = ""
+    image_version: str = ""
+
+
+@dataclass_json()
+@dataclass
 class AzureNodeSchema:
     name: str = ""
     vm_size: str = ""
     location: str = ""
+    # Required by shared gallery images which are present in
+    # subscription different from where LISA is run
+    subscription_id: str = ""
     marketplace_raw: Optional[Union[Dict[Any, Any], str]] = field(
         default=None, metadata=schema.metadata(data_key="marketplace")
     )
+    shared_gallery_raw: Optional[Union[Dict[Any, Any], str]] = field(
+        default=None, metadata=schema.metadata(data_key="shared_gallery")
+    )
     vhd: str = ""
     nic_count: int = 1
+
     # for marketplace image, which need to accept terms
     purchase_plan: Optional[AzureVmPurchasePlanSchema] = None
 
     _marketplace: InitVar[Optional[AzureVmMarketplaceSchema]] = None
+
+    _shared_gallery: InitVar[Optional[SharedImageGallerySchema]] = None
 
     @property
     def marketplace(self) -> Optional[AzureVmMarketplaceSchema]:
@@ -132,10 +150,80 @@ class AzureNodeSchema:
         else:
             self.marketplace_raw = value.to_dict()  # type: ignore
 
+    @property
+    def shared_gallery(self) -> Optional[SharedImageGallerySchema]:
+        # this is a safe guard and prevent mypy error on typing
+        if not hasattr(self, "_shared_gallery"):
+            self._shared_gallery: Optional[SharedImageGallerySchema] = None
+        shared_gallery: Optional[SharedImageGallerySchema] = self._shared_gallery
+        if shared_gallery:
+            return shared_gallery
+        if isinstance(self.shared_gallery_raw, dict):
+            # Users decide the cases of image names,
+            #  the inconsistent cases cause the mismatched error in notifiers.
+            # The lower() normalizes the image names,
+            #  it has no impact on deployment.
+            self.shared_gallery_raw = dict(
+                (k, v.lower()) for k, v in self.shared_gallery_raw.items()
+            )
+            shared_gallery = SharedImageGallerySchema.schema().load(  # type: ignore
+                self.shared_gallery_raw
+            )
+            if not shared_gallery.subscription_id:  # type: ignore
+                shared_gallery.subscription_id = self.subscription_id  # type: ignore
+            # this step makes shared_gallery_raw is validated, and
+            # filter out any unwanted content.
+            self.shared_gallery_raw = shared_gallery.to_dict()  # type: ignore
+        elif self.shared_gallery_raw:
+            assert isinstance(
+                self.shared_gallery_raw, str
+            ), f"actual: {type(self.shared_gallery_raw)}"
+            # Users decide the cases of image names,
+            #  the inconsistent cases cause the mismatched error in notifiers.
+            # The lower() normalizes the image names,
+            #  it has no impact on deployment.
+            shared_gallery_strings = re.split(
+                r"[/]+", self.shared_gallery_raw.strip().lower()
+            )
+            if len(shared_gallery_strings) == 4:
+                shared_gallery = SharedImageGallerySchema(*shared_gallery_strings)
+                # shared_gallery_raw is used
+                self.shared_gallery_raw = shared_gallery.to_dict()  # type: ignore
+            elif len(shared_gallery_strings) == 3:
+                shared_gallery = SharedImageGallerySchema(
+                    self.subscription_id, *shared_gallery_strings
+                )
+                # shared_gallery_raw is used
+                self.shared_gallery_raw = shared_gallery.to_dict()  # type: ignore
+            else:
+                raise LisaException(
+                    f"Invalid value for the provided shared gallery "
+                    f"parameter: '{self.shared_gallery_raw}'."
+                    f"The shared gallery parameter should be in the format: "
+                    f"'<subscription_id>/<image_gallery>/<image_definition>"
+                    f"/<image_version>' or '<image_gallery>/<image_definition>"
+                    f"/<image_version>'"
+                )
+        self._shared_gallery = shared_gallery
+        return shared_gallery
+
+    @shared_gallery.setter
+    def shared_gallery(self, value: Optional[SharedImageGallerySchema]) -> None:
+        self._shared_gallery = value
+        if value is None:
+            self.shared_gallery_raw = None
+        else:
+            self.shared_gallery_raw = value.to_dict()  # type: ignore
+
     def get_image_name(self) -> str:
         result = ""
         if self.vhd:
             result = self.vhd
+        elif self.shared_gallery:
+            assert isinstance(
+                self.shared_gallery_raw, dict
+            ), f"actual type: {type(self.shared_gallery_raw)}"
+            result = " ".join([x for x in self.shared_gallery_raw.values()])
         elif self.marketplace:
             assert isinstance(
                 self.marketplace_raw, dict
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -839,6 +839,9 @@ class AzurePlatform(Platform):
             azure_node_runbook = node_space.get_extended_runbook(
                 AzureNodeSchema, type_name=AZURE
             )
+            # Subscription Id is used by Shared Gallery images located
+            # in subscription different from where LISA is run
+            azure_node_runbook.subscription_id = self.subscription_id
 
             # init node
             node = environment.nodes.from_requirement(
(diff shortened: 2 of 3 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -60,22 +60,38 @@ class AzureVmMarketplaceSchema:
     version: str = "Latest"
 
 
+@dataclass_json()
```
squirrelsc, line 63: Because it doesn't need to be serializable, so the dataclass_json is not necessary.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -132,10 +148,81 @@ def marketplace(self, value: Optional[AzureVmMarketplaceSchema]) -> None:
...
+
+                    # shared_gallery_raw is used
+                    self.shared_gallery_raw = shared_gallery.to_dict()  # type: ignore
+                else:
+                    raise LisaException(
+                        f"Invalid value for the provided shared gallery "
+                        f"parameter: '{self.shared_gallery_raw}'."
+                        f"The shared gallery parameter should be in the format: "
+                        f"'<subscription_id>/<image_gallery>/<image_definition>/<image_version>' "
+                        f"or '<image_gallery>/<image_definition>/<image_version>'"
```
squirrelsc: How about if image_version is latest? How it's supported?

dsrivastavv: <image_version> can be set to value `latest`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -132,10 +148,84 @@ def marketplace(self, value: Optional[AzureVmMarketplaceSchema]) -> None:
...
     def get_image_name(self) -> str:
         result = ""
         if self.vhd:
             result = self.vhd
+        elif self.shared_gallery:
+            assert isinstance(
+                self.shared_gallery_raw, dict
+            ), f"actual type: {type(self.shared_gallery_raw)}"
+            result = " ".join([x for x in self.shared_gallery_raw.values()])
         elif self.marketplace:
```
squirrelsc, line 229: Change it to else may be more clear.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -839,6 +839,7 @@ def _create_deployment_parameters(
             azure_node_runbook = node_space.get_extended_runbook(
                 AzureNodeSchema, type_name=AZURE
             )
+            azure_node_runbook.subscription_id = self.subscription_id
```
squirrelsc, line 844: Add comments above to explain the sig need it.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -859,6 +860,9 @@ def _create_deployment_parameters(
             if azure_node_runbook.vhd:
                 # vhd is higher priority
                 azure_node_runbook.marketplace = None
+                azure_node_runbook.shared_gallery = None
+            elif azure_node_runbook.shared_gallery:
+                azure_node_runbook.marketplace = None
             elif not azure_node_runbook.marketplace:
```
squirrelsc: Change it to else may be more clear.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -132,10 +148,84 @@ def marketplace(self, value: Optional[AzureVmMarketplaceSchema]) -> None:
         else:
             self.marketplace_raw = value.to_dict()  # type: ignore
 
+    @property
+    def shared_gallery(self) -> Optional[SharedImageGallerySchema]:
+        # this is a safe guard and prevent mypy error on typing
+        if not hasattr(self, "_shared_gallery"):
+            self._shared_gallery: Optional[SharedImageGallerySchema] = None
+        shared_gallery: Optional[SharedImageGallerySchema] = self._shared_gallery
+        if not shared_gallery:
```
glima: If it's there, return. Why do people love needless indentation so much? :( This must be almost hitting https://www.flake8rules.com/rules/C901.html. BTW, time to clean that one from our code base, IMO--I also hit that once on my (internal) code and I loved it (very handy advice that you're really screwing up in legibility).

lisa/sut_orchestrator/azure/common.py
```diff
@@ -132,10 +148,84 @@ def marketplace(self, value: Optional[AzureVmMarketplaceSchema]) -> None:
...
+                # Users decide the cases of image names,
+                #  the inconsistent cases cause the mismatched error in notifiers.
+                # The lower() normalizes the image names,
+                #  it has no impact on deployment.
+                shared_gallery_strings = re.split(
+                    r"[/]+", self.shared_gallery_raw.strip().lower()
+                )
+                if len(shared_gallery_strings) == 4:
+                    shared_gallery = SharedImageGallerySchema(*shared_gallery_strings)
+                    # shared_gallery_raw is used
```
glima: Used where?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -132,10 +148,84 @@ def marketplace(self, value: Optional[AzureVmMarketplaceSchema]) -> None:
...
+                    self.shared_gallery_raw = shared_gallery.to_dict()  # type: ignore
+                elif len(shared_gallery_strings) == 3:
+                    shared_gallery = SharedImageGallerySchema()
```
glima: You can still use the constructor on this path, no?

lisa/sut_orchestrator/azure/common.py
glima: OK, saw it :)

lisa/sut_orchestrator/azure/common.py
dsrivastavv: This would also be applicable to `AzureVmMarketplaceSchema`, right?

lisa/sut_orchestrator/azure/common.py
dsrivastavv: Nice suggestion! Currently complexity is set to 15 but maybe we can bring it down to 10. These are the error I see if I reduce it to 10 : 
```
.\lisa\runner.py:235:5: C901 'RootRunner._start_loop' is too complex (15)
.\lisa\search_space.py:97:5: C901 'IntRange.check' is too complex (12)
.\lisa\search_space.py:299:1: C901 'check_countspace' is too complex (13)
.\lisa\runners\legacy_runner.py:629:1: C901 '_track_progress' is too complex (13)
.\lisa\sut_orchestrator\azure\platform_.py:601:5: C901 'AzurePlatform._get_environment_information' is too complex (15)
.\lisa\sut_orchestrator\azure\platform_.py:728:5: C901 'AzurePlatform._get_location_info' is too complex (11)
.\lisa\sut_orchestrator\azure\platform_.py:809:5: C901 'AzurePlatform._create_deployment_parameters' is too complex (15)
.\lisa\sut_orchestrator\azure\platform_.py:1183:5: C901 'AzurePlatform._resource_sku_to_capability' is too complex (11)
.\lisa\tools\reboot.py:50:5: C901 'Reboot.reboot' is too complex (12)
.\lisa\util\process.py:63:5: C901 'Process.start' is too complex (12)
.\lisa\util\process.py:144:5: C901 'Process.wait_result' is too complex (13)
```

The codebase is increasing and we can dedicate time to code quality improvements. @squirrelsc @glima Thoughts?

lisa/sut_orchestrator/azure/common.py
dsrivastavv: No this doesn't work. Since `shared_gallery_strings` is an array, it will set first three values. We can change the variable order in dataclass but it doesn't feel like a robust solution to me.

lisa/sut_orchestrator/azure/common.py
glima: Where is this set? I thought 10 was already the default? Yes, I vote for 10. Code quality should not be a pause/pending thing--ever, IMO. You do it as you go.

lisa/sut_orchestrator/azure/common.py
glima: ``shared_gallery = SharedImageGallerySchema(sub_id, *strings)``, no?

dsrivastavv: This works. Thanks!

lisa/sut_orchestrator/azure/common.py
dsrivastavv: It's set in .flake8 file under `max-complexity`

squirrelsc: We can set it to 10, but leave exception for current violations.

glima: I see. Please don't, @somil55. Either we fix that or leave for the next good willing one.

glima: 15 is not too bad, I mean. We should fix that offender, though :/

dsrivastavv: I wasn't planning on changing this as part of this PR. But I will add a issue so that we take it up in future.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Sorry I didn't get this. Do you mean move ` elif azure_node_runbook.shared_gallery` to an `else`?

squirrelsc: yes, it would be more clear.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Added comment in the `AzureNodeSchema` class

lisa/sut_orchestrator/azure/common.py
dsrivastavv: Added issue https://github.com/microsoft/lisa/issues/1423 to track this

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: I mean here needs a comment that, the subscription id may be used by sig. So it needs to be in node level.

dsrivastavv: Got it! Thanks for clarifying.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: How about this comment?

squirrelsc (dismissed):

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: I had updated the code as per this conversation. Can you please take a look and confirm?

squirrelsc: Sorry, I misread the logic here. The logic here should be like below, and it needs a comments for else block. So the logic looks like not missing anything

```python
elif azure_node_runbook.shared_gallery:
    azure_node_runbook.marketplace = None
elif azure_node_runbook.marketplace:
    # take the value of azure_node_runbook.marketplace 
    ...
else:
    # set to default marketplace, if nothing specified
    azure_node_runbook.marketplace = AzureVmMarketplaceSchema()
```

somil55: This logic will not work because `azure_node_runbook.marketplace` might have some preexisting value and this initialization will overwrite. We want to initialize with default value if `azure_node_runbook.marketplace` is empty. So the whole logic will look something like : 

```
elif azure_node_runbook.shared_gallery:
    azure_node_runbook.marketplace = None
elif azure_node_runbook.marketplace:
    # take the value of azure_node_runbook.marketplace 
    ...
else:
    # set to default marketplace, if nothing specified
    if not azure_node_runbook.marketplace:
           azure_node_runbook.marketplace = AzureVmMarketplaceSchema()
```

dsrivastavv: The `elif azure_node_runbook.marketplace` will be empty. we don't need to do anything if `azure_node_runbook.marketplace` is already filled. I can add a `pass` there to make the overall logic more clear.

squirrelsc: `...` is the same as `pass`. yes, it's empty. Sometime people are wondering if there miss an `else`. The comments help to explain the empty clause, and make the logic more straightforward (use elif for values, and else for all None).

squirrelsc (approved):

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Definitely! The changes you suggested make the logic way more clear. Thanks for this suggestion!

PR #1427, opened 2021-07-19
Update external links to code in docs

WenyiQ:
Changed relative links to code files to external GitHub links

### DIFF
--- a/docs/run_test/runbook.rst
+++ b/docs/run_test/runbook.rst
@@ -419,8 +419,9 @@ type: list of Transformer, default is empty
 type
 ^^^^
 
-type: str, required, the type of transformer. See
-`transformers <../../lisa/transformers>`__ for all transformers.
+type: str, required, the type of transformer. See `transformers
+<https://github.com/microsoft/lisa/tree/main/lisa/transformers>`__ for all
+transformers.
 
 .. _name-3:
 
--- a/docs/run_test/runbook.rst
+++ b/docs/run_test/runbook.rst
@@ -87,8 +87,7 @@ The **runbook** contains all the configurations of LISA operation. It
 keeps you from lengthy command-line commands and makes it easy to
 adjust configurations.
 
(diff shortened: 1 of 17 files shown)


docs/run_test/runbook.rst
```diff
@@ -419,8 +419,9 @@ type: list of Transformer, default is empty
 type
 ^^^^
 
-type: str, required, the type of transformer. See
-`transformers <../../lisa/transformers>`__ for all transformers.
+type: str, required, the type of transformer. See `transformers
+<https://github.com/microsoft/lisa/tree/main/lisa/transformers>`__ for all
+transformers.
```
glima, line 423: Why not use Sphinx labels?

squirrelsc (dismissed):

docs/run_test/runbook.rst
WenyiQ: You mean like ``.. some link: xxx.com``? I only replaced the relative part with external one to GitHub since they are all code files in Git.

glima: I mean using sphinx.ext.autosectionlabel and just getting away with a ``:ref:`Transformer```, instead of all that gunk (that may break an need maintenance, in any restructuring effort).

WenyiQ: It works in internal links (between html pages) but does not work on external links, so I'm afraid we have to keep these external links to codes on Git.

glima: How is runbook.rst linking to transformers external? I'm lost here...

WenyiQ: In runbook.rst there's mainly reference to commands, codes etc. The "transformers" here refers to the folder which holds source codes of all transformers, and since we do not have codes on rtd, it's an external link to Git.

glima: I see. Time to document the classes, then (one more reason :))

squirrelsc (approved):

PR #1428, opened 2021-07-20
Fix duplicate entries in the arm template.

squirrelsc:
The arm template is cached in memory to reduce disk io. The hooks on
template will modify it. The template should be copied before modified,
So it won't be modified multiple times.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -6,6 +6,7 @@ import json
 import logging
 import os
 import re
+from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import datetime
 from functools import lru_cache
@@ -917,7 +918,9 @@ class AzurePlatform(Platform):
                 )
                 arm_parameters.enable_sriov = False
 
-        template = self._load_template()
+        # the arm template may be updated by the hooks, so make a copy to avoid
+        # the original template is modified.
+        template = deepcopy(self._load_template())
         plugin_manager.hook.azure_update_arm_template(
             template=template, environment=environment
         )


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -917,7 +918,9 @@ def _create_deployment_parameters(
                 )
                 arm_parameters.enable_sriov = False
 
-        template = self._load_template()
+        # the arm template may be updated by the hooks, so make a copy to avoid
+        # the original template is modified.
+        template = deepcopy(self._load_template())
```
dsrivastavv, line 923: Can we modify the `_load_template` function to return a deepcopy instead?

squirrelsc: The method use a cache decorator, so the return value of this method is managed by another package. The deepcopy inside of the `_load_template` is called once like the method. So it doesn't work. The copy should happen after get the cached value.

dsrivastavv: Got it! So why can't we remove the `lru_cache` decorator in `_load_template` function? Will it significantly impact the runtime or is it because concurrency issues might arise in file read?

squirrelsc: It loads file from disk, I hope to reduce the disk IO, it's not a big impact although.

dsrivastavv: If this is the case, it would be up to you to make a call. The only upside I can think of removing `lru_cache` is that we don't need to explicitly remember to `deepcopy`.

dsrivastavv (approved):

LiliDeng (approved):

PR #1429, opened 2021-07-21
Implement source installer and improvements

squirrelsc:
Detail changes refer to commit message

runbook samples

1. install from repo
```yaml
transformer:
  - type: azure_deploy
    requirement:
      azure:
      core_count: 16
    enabled: true
  - type: kernel_installer
    connection:
      address: $(azure_deploy_address)
    installer:
      type: source
      location:
        type: repo
        path: /mnt/code
        ref: tags/v4.9.184
```

2. install from local code path
```yaml
transformer:
  - type: azure_deploy
    requirement:
      azure:
      core_count: 16
  - type: kernel_installer
    connection:
      address: $(azure_deploy_address)
    installer:
      type: source
      location:
        type: local
        path: /mnt/code/linux
```

3. apply patches from another patch repo
```yaml
transformer:
  - type: azure_deploy
    requirement:
      azure:
      core_count: 16
    enabled: true
  - type: kernel_installer
    connection:
      address: $(azure_deploy_address)
    installer:
      type: source
      location:
        type: repo
        path: /mnt/code
        ref: tags/v4.9.184
      modifier:
        - type: patch
          repo: https://github.com/microsoft/azure-linux-kernel.git
          file_pattern: Patches_Following_Mainline_History/4.9.184/*.patch
```


lisa/node.py
```diff
@@ -514,3 +514,24 @@ def from_requirement(
...
+    node = Node.create(index, runbook, logger_name=name)
+    if isinstance(node, RemoteNode):
+        node.set_connection_info_by_runbook()
+    node.initialize()
+
+    return node
+
+
+def quick_connect_remote(
+    runbook: schema.RemoteNode, name: str = "", index: int = -1
```
glima: You'd better repeat what the inner method states here, so to avoid people having to follow source tree just to understand what a function does

lisa/node.py
```diff
@@ -514,3 +514,24 @@ def from_requirement(
         self._list.append(node)
 
         return node
+
+
+def quick_connect(runbook: schema.Node, name: str = "", index: int = -1) -> Node:
```
glima: How about we start documenting arguments and return values as well? Here, people might think name is for node name (which is not a bad idea), but it's not the case

lisa/tools/make.py
```diff
@@ -24,14 +30,32 @@ def _install(self) -> bool:
...
-            raise LisaException(
-                f"'make' command got non-zero exit code: {make_result.exit_code}"
-            )
+    def make_install(self, cwd: PurePath, timeout: int = 600) -> None:
+        self.make(arguments="", cwd=cwd, timeout=timeout)
+
+        # install with sudo
+        self.make(arguments="install", cwd=cwd, sudo=True, timeout=timeout)
+
+    def make(  # type: ignore
```
glima: Why ignore? And why that piping of yes to it?

lisa/util/process.py
```diff
@@ -185,7 +185,9 @@ def wait_result(self, timeout: float = 600) -> ExecutableResult:
                 if self._process._stderr:
                     self._process._stderr.close()
             self._process = None
-            self._log.debug(f"waited with {self._timer}")
+            self._log.debug(
+                f"waited with {self._timer}, exit code: {self._result.exit_code}"
```
glima: Can we finally s/with/for here? Maybe a complete overhaul to something like "finished in {self.timer}" too.

lisa/transformers/kernel_source_installer.py
```diff
@@ -0,0 +1,294 @@
...
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return RepoLocationSchema
+
+    def get_source_code(self) -> PurePath:
+        runbook = cast(RepoLocationSchema, self.runbook)
+        code_path = _get_code_path(runbook.path, self._node, f"{self.type_name()}_code")
+
+        # expand env variables
+        echo = self._node.tools[Echo]
```
glima, line 287: What's going on here with this echo?

lisa/node.py
squirrelsc: After reviewed the logic, the method is removed. The most places should use Node, not RemoteNode.

lisa/node.py
squirrelsc: The "name" is not a good name, update it to “logger_name”.

lisa/tools/make.py
squirrelsc: Good catch, not sure how it happens. Removed

lisa/util/process.py
squirrelsc: Changed to executed time, since it may be timeout, not finished.

lisa/transformers/kernel_source_installer.py
squirrelsc: as the comments, it expand env variables, like `$HOME`, `~`.

lisa/util/process.py
glima: How about "executing time"?

lisa/main.py
```diff
@@ -17,7 +17,7 @@
 from lisa.variable import add_secrets_from_pairs
 
 
-@retry(FileExistsError, tries=10, delay=0.2)
+@retry(FileExistsError, tries=10, delay=0.2)  # type: ignore
```
glima, line 20: Weird, it *introduces* mypy errors?? What about the other type-whatever packages? Don't they help remove any of these?

pyproject.toml
```diff
@@ -52,6 +51,8 @@ types-paramiko = "^0.1.7"
 types-requests = "^2.25.0"
 types-python-dateutil = "^0.1.4"
 types-PyYAML = "^5.4.3"
+rstcheck = "^3.3.1"
+snooty-lextudio = "^1.9.2-alpha.0"
```
glima, line 55: Why this? Is is something akin to https://github.com/vscode-restructuredtext/vscode-restructuredtext ?

lisa/util/process.py
squirrelsc: When the log is output, the process is ended, so I used the executed. Should the "executing time" be better?

lisa/main.py
squirrelsc: Not sure what's difference between retry and other types packages, but other packages works well. This fix makes the mypy passed.

pyproject.toml
squirrelsc: The snooty is a dependency of the restructuredtext extension. If it's not installed, the vscode complains about it, when trying to preview the rst file.

lisa/util/process.py
glima: @WenyiQ IMO that's grammatically unsound, let's call Wenyi

WenyiQ: Personally, I would prefer "execution time".

squirrelsc: Thank you, Wenyi. Changed to execution time.

PR #1430, opened 2021-07-22
Migrate xfstests cases (Part 1)

LiliDeng:


lisa/tools/fdisk.py
```diff
@@ -44,10 +44,10 @@ def make_partition(self, disk_name: str, file_system: FileSystem) -> None:
         )
         if file_system == FileSystem.xfs:
             mkfs_xfs = self.node.tools[Mkfsxfs]
-            mkfs_xfs.mkfs(f"{disk_name}p1", str(file_system))
+            mkfs_xfs.mkfs(disk_name, f"{str(file_system)} -f")
```
squirrelsc: What the -f is used for?

squirrelsc: Write like below should be enough. The str() will be used automatically.
`f"{file_system} -f"`

lisa/tools/mount.py
```diff
@@ -11,7 +11,9 @@
 
 
 class Mount(Tool):
-    __UMOUNT_ERROR_PATTERN = re.compile(r".*mountpoint not found", re.MULTILINE)
+    __UMOUNT_ERROR_PATTERN = re.compile(
+        r"(.*mountpoint not found|.*no mount point specified)", re.MULTILINE
```
squirrelsc: The .* should be out of the capture group, it makes the regular expression more simple.

lisa/tools/parted.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from typing import cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+
+
+class Parted(Tool):
```
squirrelsc, line 12: What's the difference with fdisk? Do we need both of them?

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,104 @@
...
+
+from lisa.executable import Tool
+from lisa.operating_system import Debian, Fedora, Posix
+
+from .git import Git
+from .make import Make
+
+
+class Xfstests(Tool):
+    repo = "https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git"
```
squirrelsc, line 17: can we specify a tag or version? It would be more stable than default branch.

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _install_dep(self) -> None:
+        posix_os: Posix = cast(Posix, self.node.os)
+        tool_path = self.get_tool_path()
+        self.node.shell.mkdir(tool_path, exist_ok=True)
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
+        # install dependency packages
+        if isinstance(self.node.os, Fedora):
+            posix_os.install_packages(
```
squirrelsc: This list is too long. Some dependent packages may be included automatically. We can let package system to include them.

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,104 @@
...
+        self.node.execute("useradd -m fsgqa", sudo=True)
+        self.node.execute("groupadd fsgqa", sudo=True)
+        self.node.execute("useradd 123456-fsgqa", sudo=True)
+        self.node.execute("useradd fsgqa2", sudo=True)
+
+    def _install(self) -> bool:
+        self._add_test_users()
+        self._install_dep()
+        tool_path = self.get_tool_path()
+        make = self.node.tools[Make]
```
squirrelsc, line 148: If the tools are required, put them in dependencies. They can be installed together.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,72 @@
...
+    first_disk = f"{disk_name}1"
+    second_disk = f"{disk_name}2"
+    fdisk = node.tools[Fdisk]
+    parted = node.tools[Parted]
+
+    mount.umount(first_disk, first_mountpoint)
+    mount.umount(second_disk, second_mountpoint)
+
+    parted.make_label(disk_name)
+    parted.make_partition(disk_name, "primary", "1", "50%")
```
squirrelsc, line 39: Both parted and fdisk has the `make_partition` method, but it looks the functionality is different. So one of the name may be inaccurate, it's not "make partition", it may make something others like volume?

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,72 @@
...
+        """,
+        priority=2,
+    )
+    def xfstests_xfs_standard_datadisk_validation(self, node: Node) -> None:
+        xfstests = node.tools[Xfstests]
+        cmd_result = node.execute(
+            "readlink -f /dev/disk/azure/scsi1/*", shell=True, sudo=True
+        )
+        _configure_disk(node, cmd_result.stdout)
+        tool_path = xfstests.get_xfstests_config_path()
```
squirrelsc: The variable should be called like `config_path`, not `tool_path`, if the method name is accurate.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,72 @@
...
+        node.execute(
```
squirrelsc: 1. check the returned value.
2. if the test case is executed twice on a node, does it mean the line is added twice? If so, how to avoid it?

lisa/tools/xfstests.py
LiliDeng: I checked that, xfstests doesn't have such stable tag/branch.
https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git
https://github.com/kdave/xfstests

lisa/tools/xfstests.py
LiliDeng: How do we know the top package name? then omit its dependent ones?

squirrelsc: we should figure out, which ones are direct used. Others should be included by package system like apt, yum.

lisa/schema.py
```diff
@@ -370,7 +370,7 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
         metadata=metadata(decoder=search_space.decode_count_space),
     )
     disk_count: search_space.CountSpace = field(
```
squirrelsc: let's change it to data_disk_count. It's clearer.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -376,7 +376,9 @@ def _prepare_environment(  # noqa: C901
                             min_runbook.location = location_name
                             min_runbook.vm_size = azure_cap.vm_size
                             assert isinstance(min_cap.nic_count, int)
+                            assert isinstance(min_cap.disk_count, int)
```
squirrelsc: move the assert one line lower. Make it's near to `min_runbook.disk_count = min_cap.disk_count`

lisa/tools/xfstests.py
squirrelsc: Anyway, put them into variables, and not put them in code. It makes the code is harder to read, because they are too long.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def _configure_disk(
+    node: Node,
+    disk_name: str,
+    first_mountpoint: str = "/root/test",
+    second_mountpoint: str = "/root/scratch",
+    file_system: FileSystem = FileSystem.xfs,
+) -> None:
+    node.execute(f"rm -r {first_mountpoint}", sudo=True)
```
squirrelsc: A loop to the mount points is better. The most line of code are duplicated.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,74 @@
...
+    description="""
+    This test suite is to validate different data disk on Linux VM using xfstests.
+    """,
+)
+class xfstests(TestSuite):
+    TIME_OUT = 7200
+
+    @TestCaseMetadata(
+        description="""
+        This test case will
```
squirrelsc: Remember to fill in it.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,74 @@
...
+        xfstests = node.tools[Xfstests]
+        cmd_result = node.execute(
+            "readlink -f /dev/disk/azure/scsi1/*", shell=True, sudo=True
+        )
+        _configure_disk(node, cmd_result.stdout)
+        config_path = xfstests.get_xfstests_config_path()
+        node.execute(
+            f"echo 'SCRATCH_DEV=/dev/sdc2' >> {config_path}", sudo=True, shell=True
+        )
+        node.execute(
```
squirrelsc: Add comments to explain each step. what's set or verified.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,74 @@
...
+            f"echo 'SCRATCH_MNT=/root/scratch' >> {config_path}", sudo=True, shell=True
+        )
+        node.execute(
+            f"echo 'TEST_DEV=/dev/sdc1' >> {config_path}", sudo=True, shell=True
+        )
+        node.execute(
```
squirrelsc: not to output to other place, check the content in Python logic. It can bring raw output to log, and help troubleshooting.

LiliDeng:
Thanks for your time, please hold on this PR review.

lisa/tools/fdisk.py
LiliDeng: fix

lisa/tools/mount.py
LiliDeng: fix

lisa/tools/parted.py
LiliDeng: parted can partition disk bigger than 2 TB

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: fix

squirrelsc:
Split it to multiple commit, so it can explain changes on each tool.

lisa/tools/parted.py
squirrelsc: So, add comments in the tool.

lisa/tools/fdisk.py
```diff
@@ -51,16 +51,32 @@ def make_partition(self, disk_name: str, file_system: FileSystem) -> None:
...
-        # d => delete a partition
-        # w => write table to disk and exit
-        self.node.execute(
-            f"(echo d; echo w) | {self.command} {disk_name}", shell=True, sudo=True
+        # /dev/sdc => sdc
+        disk_label = disk_name.split("/")[-1]
+        # /dev/sdc
+        # /dev/sdc1
+        # /dev/sdc2
+        partitions = self.node.execute(
```
squirrelsc: Avoid to use grep in command line. Read all from cat, and parse it in Python. It can show more raw output from commands, and easy to troubleshooting.

lisa/tools/fdisk.py
```diff
@@ -1,18 +1,11 @@
...
 from lisa.operating_system import Posix
-from lisa.util import LisaException
 
-from .mkfs import Mkfsext, Mkfsxfs
-
-FileSystem = Enum(
-    "mkfs",
-    ["xfs", "ext2", "ext3", "ext4"],
-)
+from .mkfs import FileSystem, Mkfs
```
squirrelsc, line 10: The description of this commit "Refine mkfs." is too simple to the change. It needs to specify what's problem fixed, or what functionality added.

lisa/tools/fdisk.py
```diff
@@ -36,20 +29,18 @@ def make_partition(self, disk_name: str, file_system: FileSystem) -> None:
...
             shell=True,
             sudo=True,
         )
-        if file_system == FileSystem.xfs:
-            mkfs_xfs = self.node.tools[Mkfsxfs]
-            mkfs_xfs.mkfs(f"{disk_name}p1", str(file_system))
-        elif file_system in [FileSystem.ext2, FileSystem.ext3, FileSystem.ext4]:
-            mkfs_ext = self.node.tools[Mkfsext]
-            mkfs_ext.mkfs(f"{disk_name}p1", str(file_system))
+        if "nvme" in disk_name:
```
squirrelsc: This looks like a hack. what if the name is changed? What's reason to have the name on partition_disk?

LiliDeng: thanks, I'll fix it by get the info from vm.

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,131 @@
...
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git, Make]
+
+    def _install_dep(self) -> None:
+        posix_os: Posix = cast(Posix, self.node.os)
+        tool_path = self.get_tool_path()
+        self.node.shell.mkdir(tool_path, exist_ok=True)
```
squirrelsc: move this line into `get_tool_path`.

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,131 @@
...
+    def get_xfstests_path(self) -> PurePath:
+        tool_path = self.get_tool_path()
+        return tool_path.joinpath("xfstests-dev")
+
+    def set_local_config(
+        self, scratch_dev: str, scratch_mnt: str, test_dev: str
+    ) -> None:
+        xfstests_path = self.get_xfstests_path()
+        config_path = xfstests_path.joinpath("local.config")
+        self.node.execute(
```
squirrelsc: refine the code of echo, let it supports `append_to_file`

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,291 @@
...
+        self._xfstesting(
+            environment,
+            node,
+            file_system=FileSystem.btrfs,
+            test_type=FileSystem.btrfs.name,
+            excluded_tests="btrfs/244",
+            is_nvme=True,
+        )
+
+    def _xfstesting(
```
squirrelsc: this method name is confusing...

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,291 @@
...
+@TestSuiteMetadata(
+    area="storage",
+    category="community",
+    description="""
+    This test suite is to validate different types of data disk on Linux VM
+     using xfstests.
+    """,
+)
+class Xfstesting(TestSuite):
+    TIME_OUT = 7200
```
squirrelsc: add comments to explain why it needs longer time.

lisa/tools/parted.py
LiliDeng: fixed

lisa/tools/xfstests.py
LiliDeng: fixed

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: fixed

lisa/tools/fdisk.py
LiliDeng: fixed

lisa/tools/fdisk.py
LiliDeng: fixed

lisa/tools/xfstests.py
LiliDeng: refined

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: updated

lisa/features/disks.py
```diff
@@ -19,6 +19,9 @@ def can_disable(cls) -> bool:
     def enabled(self) -> bool:
         return True
 
+    def get_raw_data_disks(self) -> List[str]:
```
squirrelsc, line 44: How about `get_physical_disks`? I think it includes the os disk too. And I'm not sure the raw disk is a right term. The raw disk may indicate a disk, which is not formatted. Once it's formatted, it's not a raw disk.

lisa/tools/mkfs.py
```diff
@@ -31,16 +38,60 @@ def mkfs(self, disk: str, command: str) -> None:
             )
         cmd_result.assert_exit_code()
 
+    def format_disk(self, disk_name: str, file_system: FileSystem) -> None:
+        if file_system == FileSystem.xfs:
+            mkfs_xfs = self.node.tools[Mkfsxfs]
+            mkfs_xfs.mkfs(disk_name, f"{file_system} -f")
```
squirrelsc: Does the -f is necessary for this type only? If not below line can be shared for all tools

`mkfs_xfs.mkfs(disk_name, f"{file_system} -f")`

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,128 @@
...
+        posix_os: Posix = cast(Posix, self.node.os)
+        tool_path = self.get_tool_path()
+        self.node.shell.mkdir(tool_path, exist_ok=True)
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
+        # install dependency packages
+        if isinstance(self.node.os, Fedora):
+            posix_os.install_packages(list(self.fedora_dep))
+        elif isinstance(self.node.os, Debian):
+            posix_os.install_packages(list(self.debian_dep))
```
squirrelsc: add else to handle other distros.

lisa/features/disks.py
LiliDeng: here it will get all data disks name, /dev/sdc etc

squirrelsc: So, the method name shouldn't contain `data`

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+        file_system: FileSystem = FileSystem.xfs,
+        test_type: str = "generic",
+        excluded_tests: str = "",
+    ) -> None:
+        _configure_disk(node, data_disk, test_dev, scratch_dev, file_system=file_system)
+        xfstests = node.tools[Xfstests]
+        xfstests.set_local_config(scratch_dev, _scratch_folder, test_dev)
+        xfstests.set_excluded_tests(excluded_tests)
+        node.execute(
+            f"export TEST_DIR={_test_folder} && "
```
squirrelsc: The `export` here is not necessary. The variable applies to current execute command only. So remove the `export`.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+    description="""
+    This test suite is to validate different types of data disk on Linux VM
+     using xfstests.
+    """,
+)
+class Xfstesting(TestSuite):
+    # Use xfstests benchmark to test the different types of data disk,
+    #  it will run many cases, so the runtime is longer than usual case.
+    TIME_OUT = 14400
+    EXCLUDED_TESTS = (
```
squirrelsc, line 63: Add comments to specify why they are excluded.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+                data_disk_iops=500,
+                data_disk_count=1,
+            ),
+            min_core_count=16,
+        ),
+        timeout=TIME_OUT,
+        priority=3,
+    )
+    def xfstesting_generic_standard_datadisk_validation(self, node: Node) -> None:
+        # exclude this case generic/641 temporarily
```
squirrelsc: Add a mark `TODO: ` to identify the temp workaround.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+        # it will trigger oops on RHEL8.3/8.4, VM will reboot
+        # lack of commit 5808fecc572391867fcd929662b29c12e6d08d81
+        if isinstance(node.os, Redhat) and node.os.information.version >= "8.3.0":
```
squirrelsc: We can use partial pass to make it more obvious. If the tests is excluded, raise PassedException, and give the reason at end of the test case.

lisa/features/disks.py
squirrelsc: Please take care this comments.

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,152 @@
...
+        package_list = []
+        package_list.extend(self.common_dep)
+        if isinstance(self.node.os, Fedora):
+            package_list.extend(self.fedora_dep)
+        elif isinstance(self.node.os, Debian):
+            package_list.extend(self.debian_dep)
+        elif isinstance(self.node.os, Suse):
+            package_list.extend(self.suse_dep)
+
+        for package in list(package_list):
```
squirrelsc, line 132: The `install_packages` can install multiple packages. It accepts the `List[str]`.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -479,6 +479,24 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
 
+    def get_raw_data_disks(self) -> List[str]:
+        pattern = re.compile(r"/dev/disk/azure/scsi[1-9]/lun[0-9][0-9]?", re.M)
+        # /dev/disk/azure/scsi1/lun0
+        cmd_result = self._node.execute(
+            "ls -d /dev/disk/azure/scsi*/*", shell=True, sudo=True
+        )
+        matched = find_patterns_in_lines(cmd_result.stdout, [pattern])
```
squirrelsc, line 489: use `get_matched_str`, if there is single pattern or need only one result.

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,152 @@
...
+        git.clone(self.repo, tool_path)
+        # install dependency packages
+        package_list = []
+        package_list.extend(self.common_dep)
+        if isinstance(self.node.os, Fedora):
+            package_list.extend(self.fedora_dep)
+        elif isinstance(self.node.os, Debian):
+            package_list.extend(self.debian_dep)
+        elif isinstance(self.node.os, Suse):
+            package_list.extend(self.suse_dep)
```
squirrelsc, line 123: add `else`

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,152 @@
...
+
+        for package in list(package_list):
+            posix_os.install_packages(package)
+
+    def _add_test_users(self) -> None:
+        self.node.execute("useradd -m fsgqa", sudo=True)
```
squirrelsc, line 139: Where the users are used? And add comments to explain why they are needed.

lisa/tools/echo.py
```diff
@@ -25,11 +25,15 @@ def write_to_file(
         value: str,
         file: str,
         sudo: bool = False,
+        append: bool = False,
         timeout: int = 60,
     ) -> None:
         # Run `echo <value> > <file>`
+        cmd = f"{value} > {file}"
```
squirrelsc: The file parameter should be a PurePath.

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,152 @@
...
+        tool_path = self.get_tool_path()
+        return tool_path.joinpath("xfstests-dev")
+
+    def set_local_config(
+        self, scratch_dev: str, scratch_mnt: str, test_dev: str
+    ) -> None:
+        xfstests_path = self.get_xfstests_path()
+        config_path = xfstests_path.joinpath("local.config")
+        echo = self.node.tools[Echo]
+        echo.write_to_file(f"SCRATCH_DEV={scratch_dev}", str(config_path), sudo=True)
```
squirrelsc: 1. Write the three lines in one command.
2. What happens if those values are written twice in the file? We may need to use sed to support `append_or_replace`.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    schema,
+    simple_requirement,
+)
+from lisa.features import Disk, Nvme
+from lisa.operating_system import Redhat
+from lisa.tools import Fdisk, Mkfs, Mount, Parted, Xfstests
+from lisa.tools.mkfs import FileSystem
```
squirrelsc: import it from `lisa.tools`

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+    second_mountpoint: str = _scratch_folder,
+    file_system: FileSystem = FileSystem.xfs,
+) -> None:
+    mount = node.tools[Mount]
+    fdisk = node.tools[Fdisk]
+    parted = node.tools[Parted]
+    mkfs = node.tools[Mkfs]
+
+    node.execute(f"rm -r {first_mountpoint}", sudo=True)
+    node.execute(f"rm -r {second_mountpoint}", sudo=True)
```
squirrelsc: use a loop on "first" and "second". It will save half of lines.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+
+    @TestCaseMetadata(
+        description="""
+        This test case will run xfs xfstests testing against
+         standard data disk with xfs type system.
+        """,
+        requirement=simple_requirement(
+            disk=schema.DiskOptionSettings(
+                disk_type=schema.DiskType.StandardHDDLRS,
+                data_disk_iops=500,
```
squirrelsc, line 102: What's the reason to limit the iops? Avoid to set it to a specified value, use link `data_disk_iops=search_space.IntRange(min=500)`

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+                data_disk_count=1,
```
squirrelsc: use `search_space.IntRange(min=1)`. If the vm have two data disk, it doesn't block to run this test case.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,301 @@
...
+            ),
+            min_core_count=16,
```
squirrelsc: What's the reason to limit the core count? If it has to be, use search_space.IntRange(min=16),

lisa/tools/xfstests.py
LiliDeng: I know, but when one package doesn't exist, the remain ones will not be installed in this situation.
So I updated it into -> install one by one.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: bigger IOPS will use large data disk size, then testing time will increase.

squirrelsc: So, it needs to limited the disk size, not the iops. It's not the right dependency. It needs a range, not a fixed number. So it can be compatible with future platforms.

LiliDeng:
@squirrelsc thanks for your review, I am testing and updating the PR, will ping you once it is ready for review.

lisa/features/disks.py
LiliDeng: this method is used to get all data disks.

lisa/sut_orchestrator/azure/features.py
LiliDeng: it is used to get all data disks.
e.g. 
/dev/disk/azure/scsi1/lun0  /dev/disk/azure/scsi1/lun1
=> disk_array = ['/dev/sda', '/dev/sdb']

lisa/tools/mkfs.py
LiliDeng: already handle it in mkfs, no need specify -f here.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: fixed

lisa/tools/xfstests.py
LiliDeng: added

lisa/tools/xfstests.py
LiliDeng: comments added

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: current code doesn't support specify disk size, will raise another PR to support specify disk size directly.

LiliDeng:
@squirrelsc please help review, thanks!

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,337 @@
...
+                raise SkippedException(
+                    "Current distro doesn't support btrfs file system."
+                )
+
+        _configure_disk(node, data_disk, test_dev, scratch_dev, file_system=file_system)
+        xfstests = node.tools[Xfstests]
+        xfstests.set_local_config(scratch_dev, _scratch_folder, test_dev, _test_folder)
+        xfstests.set_excluded_tests(excluded_tests)
+        node.execute(
+            f"bash check -g {test_type}/quick -E exclude.txt",
```
squirrelsc, line 313: It doesn't check the output result. Is it expected?

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,337 @@
...
+from lisa.features import Disk, Nvme
+from lisa.operating_system import Debian, Redhat, Suse
+from lisa.tools import Fdisk, FileSystem, Mkfs, Mount, Parted, Uname, Xfstests
+
+_scratch_folder = "/root/scratch"
+_test_folder = "/root/test"
+
+
+def _skip_test(node: Node) -> None:
+    if not (
```
squirrelsc: 1. What's the reason the test case doesn't support xfstesting? Is it not implemented? If so, raise LISAException directly with unsupported.
2. The method name can be `_check_supported`.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,337 @@
...
+    first_mountpoint: str = _test_folder,
+    second_mountpoint: str = _scratch_folder,
+    file_system: FileSystem = FileSystem.xfs,
+) -> None:
+    mount = node.tools[Mount]
+    fdisk = node.tools[Fdisk]
+    parted = node.tools[Parted]
+    mkfs = node.tools[Mkfs]
+    fdisk.delete_partitions(disk_name)
+    for mountpoint in [first_mountpoint, second_mountpoint]:
```
squirrelsc: It's not about using "for" for "for". You can passing the paired arguments, and loop them. Like below

`_configure_disk(node, diskname, {"/mnt/data1": "/dev/sdb1". "/mnt/data2": "/dev/sdb2"}`

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -0,0 +1,337 @@
...
+        node.execute(f"rm -r {mountpoint}", sudo=True)
+
+    mount.umount(first_disk, first_mountpoint)
+    mount.umount(second_disk, second_mountpoint)
+
+    parted.make_label(disk_name)
+    parted.make_partition(disk_name, "primary", "1", "50%")
+    parted.make_partition(disk_name, "secondary", "50%", "100%")
+
+    for disk in [first_disk, second_disk]:
```
squirrelsc: The loops can be merged into one loop.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: for now, yes, it is consistent with lisav2.

squirrelsc: How we make sure the test case failed, if one of xfstest failed?

lisa/features/disks.py
squirrelsc: How does it filter out the os disk?

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,165 @@
...
+        package_list = []
+        package_list.extend(self.common_dep)
+        if isinstance(self.node.os, Redhat):
+            package_list.extend(self.fedora_dep)
+        elif isinstance(self.node.os, Debian):
+            package_list.extend(self.debian_dep)
+        elif isinstance(self.node.os, Suse):
+            package_list.extend(self.suse_dep)
+        else:
+            raise SkippedException(
```
squirrelsc: 1. This reason shouldn't be skipped. It should raise LISAException, so we know a distro tested, but not supported, and support it later. Do not over use the skipped.
2. In tools, never raise SkippedException. It should be raised by LISA or test case level.

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: good to see the comments, but it explains 244 only, how about others?

lisa/tools/xfstests.py
squirrelsc: It's a good reason, you can add it as comments in code.

lisa/features/disks.py
LiliDeng: all data disks info in azure vm can be found in /dev/disk/azure/scsi*
root disks info are under different subfolder name
image

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: I copy the list from lisav2, which I am not sure the history, I will add a todo here, and figure them out later.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: when one of the cases crash the VM or interrupt the testing.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: I think cover the redhat, centos, suse, ubuntu, oracle etc endorsed marketplace images is enough.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: later, we can exclude all current failed case, and investigate one by one, then if any failure xfstests case will fail the case.

lisa/tools/xfstests.py
LiliDeng: added

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: update

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: update

lisa/tools/xfstests.py
LiliDeng: update

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: we can enable all of them, and raise LisaException, if there is any error. And then fix it later.

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: we can validate them, and parse the result now. If we don't enable the validation this test case has no value.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: update

LiliDeng: enabled all

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: validate actual results from xfstesting run

lisa/tools/xfstests.py
```diff
@@ -0,0 +1,211 @@
...
+            )
+            return
+        fail_match = self.__fail_pattern.match(results.stdout)
+        assert fail_match
+        fail_count = fail_match.group("fail_count")
+        total_count = fail_match.group("total_count")
+        fail_cases_match = self.__fail_cases_pattern.match(results.stdout)
+        assert fail_cases_match
+        fail_cases = fail_cases_match.group("fail_cases")
+        raise LisaException(
```
squirrelsc, line 213: Is it possible to merge all failure message here? one line for each failure.

LiliDeng: added

squirrelsc (approved):

PR #1436, opened 2021-07-24
Add disk provisioning test

dsrivastavv:


lisa/features/disk_ephemeral.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.feature import Feature
```
squirrelsc: Because those features are similar, and the pattern of Python asks to reduce file numbers. So they can be put into the same file, like disks.py

lisa/features/disk_ephemeral.py
```diff
@@ -0,0 +1,32 @@
...
+        self._switch(False)
+
+    def enable(self) -> None:
+        self._switch(True)
+
+    def enabled(self) -> bool:
+        return True
+
+    @staticmethod
+    def get_disk_id() -> str:
```
squirrelsc: What's reason is this a static method?

lisa/features/disk_ephemeral.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @classmethod
+    def can_disable(cls) -> bool:
+        return True
+
+    def _switch(self, enable: bool) -> None:
+        raise NotImplementedError()
+
+    def disable(self) -> None:
+        self._switch(False)
```
squirrelsc: It's worth to have a SwitchableMixin class to put those methods together. `disable`, `enable`, `_switch`.

squirrelsc: But the names are confusing with `enabled` from base feature class. Is there any better idea?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -88,6 +88,7 @@ class AzureNodeSchema:
     )
     vhd: str = ""
     nic_count: int = 1
+    disk_type: str = ""
```
squirrelsc: Use an enum, and add `OneOf` constraint to make sure it's value in scope.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -191,3 +206,33 @@ class Nvme(AzureFeatureMixin, features.Nvme):
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
+
+
+class DiskEphemeral(AzureFeatureMixin, features.DiskEphemeral):
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        super()._initialize(*args, **kwargs)
+        self._initialize_information(self._node)
```
squirrelsc, line 208: May be we can override the method in AzureFeatureMixin directly. So it doesn't need to repeat it in every class.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -191,3 +206,33 @@ class Nvme(AzureFeatureMixin, features.Nvme):
...
+
+    @staticmethod
+    def get_disk_id() -> str:
```
squirrelsc, line 211: It looks not Azure specified things. What's the reason it needs here?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -316,6 +318,21 @@ def _prepare_environment(  # noqa: C901
 
             # check locations
             for req in nodes_requirement:
+                # If no disk type feature is specified, use DiskPremiumLRS
```
squirrelsc: > If no disk type feature is specified, use DiskPremiumLRS

This can be done by set a default value into the AzureNodeSchema like `nic_count`.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -358,7 +375,11 @@ def _prepare_environment(  # noqa: C901
                     for azure_cap in location_info.capabilities:
                         if azure_cap.vm_size == node_runbook.vm_size:
                             predefined_cost += azure_cap.estimated_cost
-
+                            azure_cap = (
```
squirrelsc: It looks the get_min_capability need an extension to run those customized logic. Can you rename the method `_remove_mutually_exclusive_features` to `azure_generate_min_capability`, and call it after the next line? The parameters doesnt' need the features.MUTUALLY_EXCLUSIVE_FEATURES, it's a constant, can be gotten inside of the method.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1402,3 +1442,29 @@ def _get_deployable_vhd_path(
             wait_copy_blob(blob_client, vhd_path, log)
 
         return full_vhd_path
+
+    @staticmethod
+    def _remove_mutually_exclusive_features(
+        req_cap: Any, azure_cap: Any, mutually_exclusive_features: Dict[str, List[str]]
```
squirrelsc: Avoid to use Any type, it doesn't take the advantage of typing check.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1402,3 +1442,29 @@ def _get_deployable_vhd_path(
...
+            copy.deepcopy(azure_cap.capability.features.items)
+        )
+        for req_feature in req_cap.features:
+            if mutually_exclusive_features.get(req_feature) is not None:
+                for me_feature in mutually_exclusive_features[req_feature]:
+                    cap.capability.features.discard(me_feature)
+        return cap
+
+    @staticmethod
+    def _get_disk_type(node_space: schema.NodeSpace) -> str:
```
squirrelsc: This method just convert a string to another one. What's the different between disk_id and feature's name? Can they be the same thing?

microsoft/testsuites/core/provisioning.py
```diff
@@ -29,26 +29,7 @@
...
-        4. Otherwise, fully passed.
-        """,
-        priority=0,
-        requirement=simple_requirement(
-            environment_status=EnvironmentStatus.Deployed,
-            supported_features=[SerialConsole],
-        ),
-    )
-    def smoke_test(self, case_name: str, node: RemoteNode) -> None:
+    def _smoke_test(self, case_name: str, node: RemoteNode) -> None:
```
squirrelsc, line 86: private method should be put under public ones. Add comments to explain why there is a common method, but other methods call it without difference.

microsoft/testsuites/core/provisioning.py
```diff
@@ -89,3 +70,73 @@ def smoke_test(self, case_name: str, node: RemoteNode) -> None:
...
+            environment_status=EnvironmentStatus.Deployed,
+            supported_features=[SerialConsole, DiskStandardLRS],
+        ),
+    )
+    def smoke_test(self, case_name: str, node: RemoteNode) -> None:
+        self._smoke_test(case_name, node)
+
+    @TestCaseMetadata(
+        description="""
+        This case verifies whether a node is operating normally.
```
squirrelsc: here should highlight what's difference with the smoke_test.

microsoft/testsuites/core/provisioning.py
```diff
@@ -89,3 +70,73 @@ def smoke_test(self, case_name: str, node: RemoteNode) -> None:
...
+
+        Steps,
```
squirrelsc: It can explain that refer to the `smoke_test`, it doesn't need to repeat content here.

lisa/features/disk_ephemeral.py
dsrivastavv: This method doesn't depend on the state of the class and neither it seems to fit `classmethod`. I think the other methods in feature class modify the enabled/disabled state which is why they aren't static.

lisa/sut_orchestrator/azure/features.py
dsrivastavv: This returns azure disk type i.e "Standard_LRS", "Premium_LRS" etc. We need to specify these values in ARM template.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: We need to do it in `_prepare_environment` so that we can return correct environment capability. If we skip this step, we will get environments that can potentially have standard , premium and ephemeral features. However, this isn't possible as VM's can only have one disk at a time.

microsoft/testsuites/core/provisioning.py
dsrivastavv: Sure! Btw do you know if there is a way to enforce this in style checking? Didn't find anything on this.

microsoft/testsuites/core/provisioning.py
dsrivastavv: Sounds good! I will move the common comments to smoke test and have some test case specific comments here.

lisa/features/disk_ephemeral.py
dsrivastavv: `disable`, `enable`, `_switch` should be applicable to every feature, right? If so, we can just move them to Feature class. If not, we should identify what set for features could potentially use it and create a mixin specific to this purpose.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: disk_id is used directly in ARM template and has form like "Standard_LRS", "Premium_LRS" etc. We can't use this as class name in python as all class names are camel case.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Can you please clarify if you mean to do it something like this ?

```
predefined_cost += azure_cap.estimated_cost
min_cap: schema.NodeSpace = azure_generate_min_capability(req)
# apply azure specified values
# they will pass into arm template
min_runbook = min_cap.get_extended_runbook(
    AzureNodeSchema, AZURE
)
```

lisa/sut_orchestrator/azure/features.py
dsrivastavv: I think we will still need to add a `__init__` method here to initialize `features` parent class.

lisa/features/disk_ephemeral.py
squirrelsc: There are two levels of disable/enable. One is used before deploying, another one is used after deployed and in the node. I will think about to support the deploying level variables. For now, let's have a separated mixin to support the second level disable/enable.

lisa/sut_orchestrator/azure/features.py
squirrelsc: `__init__` is the constructive method, which is called automatically when create the object. It shouldn't have any long time operation.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: I misunderstand the code here, but it doesn't need too. There are two things here, and can be done in another place.

1. The default value of the disk type. It shouldn't be specified. Like other place, if it's not specify, it means the test case can run with any disk type. But if it's added here, it means the test case can run in default disk type only, even it's not specified.
2. Conflict detection. Any conflict should raise out, not just fix it. And it should be detected in a central place. Though I don't have a good idea about where to handle it, but we can put it in `azure_generate_min_capability`

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Yes, it looks better.

lisa/sut_orchestrator/azure/features.py
dsrivastavv: I think I might have missed few points in my explanation. If we want to move this `_initialize` method to `AzureFeatureMixin`, then we would miss calling the `_initialize` function for `features.DiskEphemeral` class. We also cannot make `features.DiskEphemeral` a parent of `AzureFeatureMixin` as it is different for different features. Maybe we can discuss this offline for better clarity on this.

squirrelsc: If the method moved to `AzureFeatureMixin`, the `InitializableMixin` should be moved to the parent of `AzureFeatureMixin`.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: To summarize : If the environment is using default disk type, then it should have no disk feature added to it's capability.

I think this case can be handled in `azure_generate_min_capability`

lisa/sut_orchestrator/azure/features.py
dsrivastavv: It seems like `AzureFeatureMixin` and `features.DiskStandardLRS` are dependent on each other. `_initialize_information` requires `_node` which are part of  `AzureFeatureMixin` and `features.DiskStandardLRS` respectively. The code works but mypy throws an error as `_node` isn't accessible inside `AzureFeatureMixin`

lisa/features/disks.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @staticmethod
+    def get_disk_id() -> str:
+        raise NotImplementedError
+
+
+class DiskPremiumLRS(Feature, SwitchableMixin):
+    @classmethod
+    def name(cls) -> str:
+        return "DiskPremiumLRS"
```
squirrelsc: It can be "cls.__name__". It avoids typo and if the class name is changed, it still works.

lisa/sut_orchestrator/azure/features.py
squirrelsc: I see. There are a couple of way to fix that, but they are not good patterns. So let's keep it as is. Thank you for investigating!

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -358,9 +361,10 @@ def _prepare_environment(  # noqa: C901
                     for azure_cap in location_info.capabilities:
                         if azure_cap.vm_size == node_runbook.vm_size:
                             predefined_cost += azure_cap.estimated_cost
-
-                            min_cap: schema.NodeSpace = req.generate_min_capability(
-                                azure_cap.capability
+                            min_cap: schema.NodeSpace = (
+                                self._azure_generate_min_capability(
```
squirrelsc: Because this method is in AzurePlatform, it doesn't need the keyword `Azure`. Maybe it can be called like `_node_generate_min_capability`?

dsrivastavv: Sure! I will update this.

microsoft/testsuites/core/provisioning.py
squirrelsc: I don't know which tools can help on it.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1402,3 +1425,53 @@ def _get_deployable_vhd_path(
...
+        else:
+            min_cost_disk = AzurePlatform._get_min_cost_disk(req_cap.features)
+
+        # Remove mutually exclusive feature
+        azure_cap_copy = copy.deepcopy(azure_cap)
+        azure_cap_copy.features.update(  # type: ignore
+            copy.deepcopy(azure_cap.features.items)
+        )
+
+        for me_feature in features.MUTUALLY_EXCLUSIVE_FEATURES[min_cost_disk]:
```
squirrelsc: It should raise exception, if test case requirement includes conflict features.

microsoft/testsuites/core/provisioning.py
```diff
@@ -31,7 +31,50 @@ class Provisioning(TestSuite):
...
+        ),
+    )
+    def verify_deployment_provision_premium_disk(
+        self, case_name: str, node: RemoteNode
+    ) -> None:
+        self._smoke_test(case_name, node)
+
+    def _smoke_test(self, case_name: str, node: RemoteNode) -> None:
+        """
+        This function verifies whether a node is operating normally.
```
squirrelsc: Those content should be put on `smoke_test` description. It's a part of test spec, so it can be included in future documentation. For other test cases, they need explain that the progress is the same as `smoke_test`.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1402,3 +1425,53 @@ def _get_deployable_vhd_path(
             wait_copy_blob(blob_client, vhd_path, log)
 
         return full_vhd_path
+
+    def _node_generate_min_capability(
+        self, req_cap: schema.NodeSpace, azure_cap: schema.NodeSpace
+    ) -> schema.NodeSpace:
+        assert azure_cap.features, "azure node has no disk features."
+        if req_cap.features is None or (
+            not req_cap.features.__contains__(DiskType.DISK_EPHEMERAL)
```
squirrelsc: The standard Python grammar is `DiskType.DISK_EPHEMERAL not in req_cap.features`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1402,3 +1425,53 @@ def _get_deployable_vhd_path(
...
+            and not req_cap.features.__contains__(DiskType.DISK_PREMIUM)
+            and not req_cap.features.__contains__(DiskType.DISK_STANDARD)
+        ):
+            # If no disk type feature is specified, use least
+            # cost available disk from azure_cap
+            min_cost_disk = AzurePlatform._get_min_cost_disk(azure_cap.features)
+        else:
+            min_cost_disk = AzurePlatform._get_min_cost_disk(req_cap.features)
```
squirrelsc, line 1454: The min_cap should be found in overlap of req and cap. There may be logic to check it before, but the min_cap should have this check too. In case it's misused in future, or followed by other people.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: It's possible for `azure_cap` to have conflicting features. A azure VM could support standard, premium and ephemeral disk. The logic here is trying to remove mutually exclusive features once `min_cost_disk` disk has been set. 

The check you mentioned should be on `req_cap`. I will add it at the beginning of `_node_generate_min_capability`.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Thanks for catching! This was a leftover from when I was trying to use Disk names as enum. I will update this.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Reviewing the logic I tried to add here : 

The `req_cap` will contain either one disk feature or none. 

Case 1:
It has no disk feature: In this case we set `min_cost_disk ` as minimum size disk available in `azure_cap` and remove other mutually exclusive disk feature that `azure_cap` might support. When `req_cap.generate_min_capability(azure_cap_copy)` is called later, it will return the `min_cost` with  `min_cost_disk` feature present.

Case 2: 
It has one disk : In this case, `min_cost_disk` is set to the feature present in `req_cap`. we then remove any mutually exclusive feature from `azure_cap`. When `req_cap.generate_min_capability(azure_cap_copy)` is called later, it will return an error if `min_cost_disk` is not present in `azure_cap`.

The logic for overlap is already present in `generate_min_capability` function. I didn't want to make the `_get_min_cost_disk` more complex by re-adding it. Please let me know if I have missed any scenarios.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -895,6 +902,12 @@ def _create_deployment_parameters(
                 azure_node_runbook.purchase_plan = self._process_marketplace_image_plan(
                     azure_node_runbook.location, azure_node_runbook.marketplace
                 )
+
+            # Set disk type
+            azure_node_runbook.disk_id = AzurePlatform._get_disk_id(
```
squirrelsc: assert node_space.features => not none able

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: Based on offline discussion, I will add a validation to `req_cap` to ensure no conflict between disk features.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: I have updated few things here to make the code more readable. Please feel free to suggest if anything could be improved further.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1249,7 +1267,11 @@ def _resource_sku_to_capability(
 
         # all nodes support following features
         node_space.features.update(
-            [features.StartStop.name(), features.SerialConsole.name()]
+            [
+                features.StartStop.name(),
+                features.SerialConsole.name(),
+                features.DiskStandardLRS.name(),
```
dsrivastavv, line 1283: @squirrelsc The users would need to delete `azure_location_<location>.cache` otherwise their Lisa run would fail because of this change. Is there a way to make this process more streamlined for users?

squirrelsc: The cache will expire every 5 days. May be we can change the cache to one day? We may add version into the file in future. But for now, it's acceptable. If someone fails on this, we can let them know. At most 5 days, after the cache refreshed, it's mitigated.

dsrivastavv: Changing cache to 1 day is a good idea for minimum impact on users. I will add a separate commit to update this value.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1402,3 +1424,70 @@ def _get_deployable_vhd_path(
...
+                req_cap.features  # type: ignore
+            )
+
+        # Remove mutually exclusive feature
+        azure_cap_copy = copy.deepcopy(azure_cap)
+        azure_cap_copy.features.update(  # type: ignore
+            copy.deepcopy(azure_cap.features.items)
+        )
+        for me_feature_group in features.MUTUALLY_EXCLUSIVE_FEATURES:
+            if min_cost_disk in me_feature_group:
```
squirrelsc, line 1466: It's better to add an example and explanation to help understanding the purpose and logic.

dsrivastavv: Sure! I will add examples to make it clear.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1402,3 +1424,70 @@ def _get_deployable_vhd_path(
...
+                features_to_remove = set(me_feature_group)
+                features_to_remove.discard(min_cost_disk)
+                azure_cap_copy.features.difference_update(  # type: ignore
```
squirrelsc: use assert to avoid this kind of type: ignore

dsrivastavv: This assert would be redundant. This is deepcopy of `azure_cap` and we already have a assert on it. I feel it makes the logic more complex. Please let me know if you think otherwise.

squirrelsc: The assert is better than "# type: ignore". It's to improve typing validation.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: I am thinking of adding following comment : 
```
# Remove features which are mutually exclusive feature to `min_cost_disk`
# Example :
# min_cost_disk = DiskStandardLRS
# (Initial) azure_cap_copy.features = [DiskStandardLRS, DiskPremiumLRS, DiskEphemeral]
# (Final) azure_cap_copy.features = [DiskStandardLRS]
```

Please feel free to suggest any edits.

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: There is another `#type : ignore` present on https://github.com/microsoft/lisa/blob/8f703e8511be0a77f7fbf6a1cf51d58eb612dbe9/lisa/sut_orchestrator/azure/platform_.py#L1436-L1443

If counter is in else block, `req_cap.features` is greater than 0. But mypy is still throwing an error. I am thinking of adding a `None` check to `if` condition rather than an `assert` in else block.

squirrelsc: mypy is a static check tool, it's not smart enough to understand this kind of logic. Yes, you can add a None check in the if block.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: It's good to me. It can be more human readable, not in the code style.

squirrelsc (approved):

LiliDeng (approved):

PR #1439, opened 2021-07-26
Higher quality code.

abdelrahman0w:
I tried to refactor the code in some file lines to make higher quality.

### DIFF
--- a/lisa/features/serial_console.py
+++ b/lisa/features/serial_console.py
@@ -90,18 +90,7 @@ class SerialConsole(Feature):
             saved_path.mkdir()
 
         if self._cached_console_log is None or force_run:
-            self._node.log.debug("downloading serial log...")
-            log_path = self._node.local_log_path / get_datetime_path()
-            log_path.mkdir(parents=True, exist_ok=True)
-
-            self._cached_console_log = self._get_console_log(saved_path=log_path)
-            self._node.log.debug(
-                f"downloaded serial log size: {len(self._cached_console_log)}"
-            )
-            # anyway save to node log_path for each time it's real queried
-            log_file_name = log_path / NAME_SERIAL_CONSOLE_LOG
-            with open(log_file_name, mode="wb") as f:
-                f.write(self._cached_console_log)
+            self._extracted_from_get_console_log_9()
         else:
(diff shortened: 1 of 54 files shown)


lisa/features/serial_console.py
```diff
@@ -113,6 +102,20 @@ def get_console_log(
 
         return self._cached_console_log.decode("utf-8", errors="ignore")
 
+    def _extracted_from_get_console_log_9(self):
```
squirrelsc, line 105: What's the reason to export a method? The name isn't meaningful.

glima: Hero commit, really appreciated. Is this coming from an automated tool, however? What's up with this weird name? :)

lisa/node.py
```diff
@@ -477,15 +470,9 @@ def from_existing(
         environment_name: str,
         base_log_path: Optional[Path] = None,
     ) -> Node:
-        node = Node.create(
-            index=len(self._list),
-            runbook=node_runbook,
-            logger_name=environment_name,
-            base_log_path=base_log_path,
+        return self._extracted_from_from_requirement_7(
```
glima, line 473: Yep, pretty sure this is not human-generated. Please fix the random naming, then.

glima: Is this https://github.com/python-rope/rope? Something else?

abdelrahman0w: @glima 
Hi,
No, I really didn't use that tool. However, I use an extension that suggest modifications, I think I shouldn't have used it. I am really sorry. But I really did changes by myself, the extension just suggested minor changes. I am sorry.
For the name as you mentioned, I just didn't know what to write, and I don't know if there is guidelines to follow in naming branches, but I am sorry if the name is not appropriate.

squirrelsc: @Abdelrahman0W  please remove all function extraction changes, I will have a look on other changes.

abdelrahman0w: @squirrelsc 
Okay, I will. Thank you so much

squirrelsc:
@Abdelrahman0W I have to close this PR, because there is no response long time, and there are a lot of conflicts. Feel free to solve the comments and reopen it.

PR #1442, opened 2021-07-26
Rectify branch to clone in legacy_runner

pupacha:
Legacy runner fails to checkout the given V2 branch because of https://github.com/microsoft/lisa/blob/main/lisa/runners/legacy_runner.py#L102 . 

Used `origin/branch_name` instead of `branch_name`

### DIFF
--- a/lisa/runners/legacy_runner.py
+++ b/lisa/runners/legacy_runner.py
@@ -99,7 +99,7 @@ class LegacyRunner(BaseRunner):
             git.clone(
                 config.repo,
                 cwd=self._working_folder,
-                ref=config.branch,
+                ref=f"origin/{config.branch}",
                 dir_name=self._get_dir_name(self.id, index),
             )
 


lisa/tools/git.py
```diff
@@ -44,6 +44,7 @@ def clone(
         url: str,
         cwd: pathlib.PurePath,
         ref: str = "",
+        branch: str = "",
```
squirrelsc: The branch is one of the ref format, what's the error now?

lisa/tools/git.py
```diff
@@ -70,19 +71,28 @@ def clone(
...
-        if not checkout_branch:
-            # create a temp branch to checkout tag or commit.
-            checkout_branch = f"{constants.RUN_LOGIC_PATH}"
+        if ref:
+            if not checkout_branch:
+                # create a temp branch to checkout tag or commit.
+                checkout_branch = f"{constants.RUN_LOGIC_PATH}"
+            parameters = f"checkout {ref} -b {checkout_branch}"
+        else:
+            parameters = f"checkout {checkout_branch}"
```
squirrelsc: The `checkout_branch` is used to avoid modify local branches, it's the target name of branch. So it's a name the same as LISA run.

pupacha: This is just to use whenever `ref` is not provided. It will be useful as a simple checkout to a remote branch. For ex: `checkout master` without need of any ref.

lisa/tools/git.py
pupacha: The issue is when checking out with branch name passed as `ref`. 
`legacy_runner` wants to switch to v2 branch provided in runbook but the checkout method called inside clone doesnt translate to actual git checkout command.  Instead it translates to 
`git checkout <branch_name passed as ref> -b  <PurePath>` which fails trying to search for commit id with <branch_name>. Instead it should be a simple `git checkout <branch_name>`. Hence, added branch parameter, so that whenever only branch without any ref is provided, it simply uses git checkout.

lisa/tools/git.py
squirrelsc: branch is a kind of ref.

lisa/tools/git.py
squirrelsc: Could you try again? It should be supported. The `-b <local branch name>` should be the local branch name, not related to which remote branch is checked out. If it's not specified, the remote branch name will be used. I have a try on my local, it works well with branch too. Would you show me the exact errors you found?

lisa/tools/git.py
pupacha: oh, but I'm running into errors related to git checkout while triggering legacy tests. I tried the same manually too, it didnt work. ` git checkout master -b random_name`

image

squirrelsc: I see, it needs the remote name too. It should be `origin/master`. Can you change the default name to include the origin?

lisa/tools/git.py
pupacha:
```
  File "C:\agent\_work\18\s\lisa\lisa\main.py", line 89, in <module>
    exit_code = main()
  File "C:\agent\_work\18\s\lisa\lisa\main.py", line 78, in main
    exit_code = args.func(args)
  File "C:\agent\_work\18\s\lisa\lisa\commands.py", line 48, in run
    raise identifier
  File "C:\agent\_work\18\s\lisa\lisa\commands.py", line 44, in run
    asyncio.run(runner.start())
  File "C:\Python38\lib\asyncio\runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "C:\Python38\lib\asyncio\base_events.py", line 616, in run_until_complete
    return future.result()
  File "C:\agent\_work\18\s\lisa\lisa\runner.py", line 124, in start
    raise identifer
  File "C:\agent\_work\18\s\lisa\lisa\runner.py", line 121, in start
    self._start_loop()
  File "C:\agent\_work\18\s\lisa\lisa\runner.py", line 273, in _start_loop
    task = runner.fetch_task()
  File "C:\agent\_work\18\s\lisa\lisa\runners\legacy_runner.py", line 99, in fetch_task
    git.clone(
  File "C:\agent\_work\18\s\lisa\lisa\tools\git.py", line 73, in clone
    self.checkout(ref, cwd=full_path)
  File "C:\agent\_work\18\s\lisa\lisa\tools\git.py", line 91, in checkout
    result.assert_exit_code(message=f"failed to checkout branch. {result.stdout}")
  File "C:\agent\_work\18\s\lisa\lisa\util\process.py", line 37, in assert_exit_code
    return assert_that(self.exit_code, message).is_equal_to(expected_exit_code)
```

This is the stack trace

lisa/tools/git.py
pupacha: yes, it worked. Thanks @squirrelsc  I'll update `legacy_runner` to pass `origin/branch` as ref value.

lisa/runners/legacy_runner.py
```diff
@@ -99,7 +99,7 @@ def fetch_task(self) -> Optional[Callable[[], List[TestResult]]]:
             git.clone(
                 config.repo,
                 cwd=self._working_folder,
-                ref=config.branch,
+                ref="origin/"+config.branch,
```
squirrelsc: use like f"origin/{config.branch}",

squirrelsc (dismissed):

lisa/runners/legacy_runner.py
pupacha: Thanks. changed it.

squirrelsc (approved):

PR #1443, opened 2021-07-26
Isaiah/lisa3 [LISAv3] Dump generated arm_template and arm_parameters for troubleshooting.

isaiahduncan:
Dumping arm_template and arm_parameters into a file named after the environment
This file goes into a directory created by the code "runbook/runs/arm_dump"

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -49,6 +49,7 @@ from lisa.util import (
     LisaException,
     SkippedException,
     constants,
+    dump_file,
     get_matched_str,
     get_public_key_data,
     plugin_manager,
@@ -951,6 +952,12 @@ class AzurePlatform(Platform):
             template=template,
             parameters=parameters,
         )
+
+        # dump arm_template and arm_parameters to file
+        template_dump_path = environment.log_path / "arm_template.json"
+        param_dump_path = environment.log_path / "arm_template_parameters.json"
+        dump_file(template_dump_path, json.dumps(template))
+        dump_file(param_dump_path, json.dumps(parameters))
 
         return (
             arm_parameters.location,
--- a/lisa/util/__init__.py
+++ b/lisa/util/__init__.py
@@ -8,6 +8,9 @@ from typing import Any, Dict, Iterable, List, Optional, Pattern, Type, TypeVar
 
 import pluggy
 
+from lisa import secret
+from lisa.util import constants
+
 T = TypeVar("T")
 
 # regex to validate url
@@ -231,3 +234,12 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
 
 def filter_ansi_escape(content: str) -> str:
     return __ansi_escape.sub("", content)
+
+
+def dump_file(file_name: Path, content: Any) -> None:
+    # This is for path validation. If provided file path isn't under run local path,
+    # an error will be raised. Want to ensure logs only put under run local path
+    file_name.absolute().relative_to(constants.RUN_LOCAL_PATH)
+    file_name.parent.mkdir(parents=True, exist_ok=True)
+    with open(file_name, "w") as f:
+        f.write(secret.replace(content))


isaiahduncan:
It looks like the remaining CI error has to do with lines that I didn't change @squirrelsc

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -952,6 +952,17 @@ def _create_deployment_parameters(
...
         )
 
+        # dump arm_template and arm_parameters to file
+        dump_path = os.path.join(os.getcwd(), "runtime/runs/arm_dump")
+        try:
+            os.mkdir(dump_path)
+        except OSError:
+            pass
+        dump_file = open(f"runtime/runs/arm_dump/{environment.name}.txt", "w")
+        dump_file.write("arm_template\n" + json.dumps(template))
```
squirrelsc: dump the original json format is better for view. So dump two files, one for template, another for parameters.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -952,6 +952,17 @@ def _create_deployment_parameters(
             parameters=parameters,
         )
 
+        # dump arm_template and arm_parameters to file
+        dump_path = os.path.join(os.getcwd(), "runtime/runs/arm_dump")
```
squirrelsc: 1. LISA deployed several environments in a test run. So the files should put into folder of each environment.  call Environment.log_path to get the unique path.
2. LISA hides sensitive information in log files, so all dump content should pass through the secret method. call `secret.replace`.
3. LISA should avoid to put log file out of the folder of current run. use pathlib to detect and trim path, make sure the log put under current run. Try relative_path and join with env log path.

squirrelsc:
Please organize the commits with `git rebase -i ...`, `git add -p ...`. Make each commit is meaningful, independent.

lisa/sut_orchestrator/azure/platform_.py
isaiahduncan: 2. Would secret.replace make the following string be output into the dump files? "******"
3. Could you please expand on the third point? I don't really understand what you want me to do for it.

squirrelsc: I hope to have a common method to finish the 2 and 3. It can be named `dump_file` and put in `utils/__init__.py`. 

> 2. Would secret.replace make the following string be output into the dump files? "******"
Yes, it's expected.

> 3. Could you please expand on the third point? I don't really understand what you want me to do for it.
For example, the file path may be "/etc/some_config" or "../../../../etc/some_config", it's invalid and dangerous to current system. So do some validation, 1) if it's an absolution path, make sure it's in scope of current run log path. 2) if it's a relative path, do same.

lisa/util/__init__.py
```diff
@@ -231,3 +232,20 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
 
 def filter_ansi_escape(content: str) -> str:
     return __ansi_escape.sub("", content)
+
+
+def dump_files(os, environment) -> List[TextIOWrapper]:
```
squirrelsc: Make this method is more general using, not only for current purpose, that's why it should be in the `util/__init__.py`. The interface can be,
`def dump_file(file_name: Path, content: Any) -> None:`

lisa/util/__init__.py
```diff
@@ -231,3 +232,20 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
...
+    if len(str(environment.log_path).split("\\environments\\")) < 2:
```
squirrelsc: 1. The `/` and `\` can be handled correctly by pathlib.Path. Search pathlib in code to find some examples.
2. As above comments, make this method generic. In this method, it should check if the files are in `constant.RUN_LOCAL_PATH`.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -42,13 +42,14 @@
 from lisa.feature import Feature
 from lisa.node import Node, RemoteNode
 from lisa.platform_ import Platform
-from lisa.secret import PATTERN_GUID, PATTERN_HEADTAIL, add_secret
+from lisa.secret import PATTERN_GUID, PATTERN_HEADTAIL, add_secret, replace
```
squirrelsc: This is common for other log files, it should be in the `dump_file`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -952,6 +953,17 @@ def _create_deployment_parameters(
             parameters=parameters,
         )
 
+        # dump arm_template and arm_parameters to file
+        dump_path = constants.RUN_LOCAL_PATH / "arm_dump"
```
squirrelsc: The path should be from environment, because there are many environment deployed in a test run. If so, those files can be put under the root path of environment path. So the dump_path is not necessary, just use `environment.log_path` in following file paths.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -952,6 +953,17 @@ def _create_deployment_parameters(
...
+        try:
+            os.mkdir(dump_path)
```
squirrelsc: The log_path of environment will be created automatically, when it's accessed. Check the code logic.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -952,6 +953,17 @@ def _create_deployment_parameters(
...
+        except OSError:
+            pass
```
squirrelsc: it's not a good practice to eat an exception without more strict limitation or logging. If it's used to handle path existing situation, there is an argument `exist_ok` on `mkdir`.

lisa/util/__init__.py
```diff
@@ -231,3 +234,22 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
 
 def filter_ansi_escape(content: str) -> str:
     return __ansi_escape.sub("", content)
+
+
+def dump_file(file_name: Path, content: Any) -> None:
+    local_path_parts = constants.RUN_LOCAL_PATH.parts
+    parts_len = len(local_path_parts)
+    run_path = Path(Path()/local_path_parts[parts_len-5] /
```
squirrelsc: What's this for? I believe it's not necessary. But if it's really needed, it needs comments to explain. In general, to check a path is under another path, you can do below step,

1. call `absolute` to get the absolute path. What ever path, it will be an absolute path.
2. call `relative_to(constants.RUN_LOCAL_PATH)`. It will return the relative path to `constants.RUN_LOCAL_PATH`. If it's not under that path, there is an ValueError. It's ok, if there is an ValueError raise out. Let higher level code to handle it.

lisa/util/__init__.py
```diff
@@ -3,6 +3,9 @@
 
 import re
 from datetime import datetime
+from lisa.util import constants
+from lisa.secret import replace
```
squirrelsc: Import it like below, so it's clear that the replace method is from secret module.

`from lisa import secret`

lisa/util/__init__.py
```diff
@@ -231,3 +234,22 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
...
+        local_path_parts[parts_len-2] / local_path_parts[parts_len-1])
+    if(file_name.is_absolute()):
+        if(not str(file_name).find(str(run_path))):
+            raise error
+    else:  # file_name is a relative path
+        abs_path = Path.cwd()/file_name
+        if(not str(abs_path).find(str(run_path))):
+            raise error
+
+    file = open(file_name, "w")
```
squirrelsc: We can make sure the path exists before open a file. It would be more convenient.

`file_name.parent.mkdir(parents=True, exist_ok=True)`

squirrelsc: use `with` keyword is more safe to recycle resource, and reduce code lines.

```python
with open(file_name, "w") as f:
    f.write...
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -952,6 +953,12 @@ def _create_deployment_parameters(
             parameters=parameters,
         )
 
+        # dump arm_template and arm_parameters to file
+        template_dump_path = environment.log_path / f"{environment.name}_template.txt"
```
squirrelsc: The names can be `arm_template.json` and `arm_template_parameters.json`. These files are in the env folder already, so the env name is not necessary.

lisa/util/__init__.py
squirrelsc: The second comments is not resolved.

> use `with` keyword is more safe to recycle resource, and reduce code lines.
> 
> ```python
> with open(file_name, "w") as f:
>     f.write...
> ```

lisa/util/__init__.py
```diff
@@ -231,3 +234,11 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
...
 def filter_ansi_escape(content: str) -> str:
     return __ansi_escape.sub("", content)
+
+
+def dump_file(file_name: Path, content: Any) -> None:
+    file_name.absolute().relative_to(constants.RUN_LOCAL_PATH)
+    file_name.parent.mkdir(parents=True, exist_ok=True)
+    file = open(file_name, "w")
+    file.write(secret.replace(content))
+    file.close()
```
squirrelsc: Please organize the commits with `git rebase -i ...`, `git add -p ...` . Make each commit is meaningful, independent. In this PR, one or two commits both ok.

isaiahduncan: In the last sentence did you mean that you want this pull request to only have one or two commits?

squirrelsc (dismissed):

lisa/util/__init__.py
```diff
@@ -231,3 +234,10 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
 
 def filter_ansi_escape(content: str) -> str:
     return __ansi_escape.sub("", content)
+
+
+def dump_file(file_name: Path, content: Any) -> None:
+    file_name.absolute().relative_to(constants.RUN_LOCAL_PATH)
```
squirrelsc, line 242: It's better to add some comments above to specify it's for path validation. after two or three months, we may forget why this line is here.

squirrelsc (approved):

PR #1445, opened 2021-07-27
Fix lisav3 test issues

LiliDeng:
See commits for details.

### DIFF
--- a/lisa/tools/nvmecli.py
+++ b/lisa/tools/nvmecli.py
@@ -18,6 +18,12 @@ class Nvmecli(Tool):
     __ns_management_attachement_support = "NS Management and Attachment Supported"
     # [1:1] : 0x1   Format NVM Supported
     __format_device_support = "Format NVM Supported"
+    # Higher version nvme-cli add a mandatory parameter `--block-size` after
+    #  v1.6 (not included)
+    # https://github.com/linux-nvme/nvme-cli/blob/v1.7/nvme.c#L3040
+    # FLBAS corresponding to block size 0 not found
+    # Please correct block size, or specify FLBAS directly
+    __missing_block_size_parameter = "FLBAS corresponding to block size 0 not found"
 
     @property
     def command(self) -> str:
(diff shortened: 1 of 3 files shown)


lisa/tools/nvmecli.py
```diff
@@ -39,16 +43,36 @@ def _install_from_src(self) -> None:
         make.make_install(cwd=code_path)
 
     def create_namespace(self, namespace: str) -> ExecutableResult:
-        return self.run(f"create-ns {namespace}", shell=True, sudo=True)
+        cmd_result = self.run(f"create-ns {namespace}", shell=True, sudo=True)
+        if self.__missing_block_size_parameter in cmd_result.stdout:
```
squirrelsc, line 49: 1. It looks there pattern is the same. Because there are more lines now, it's time to have a common method to do it.
2. Add comments to explain the condition, why the block-size may not found.
3. I'm not sure if all command need the block-size, for example, is it necessary in delete-ns?
4. Because the nvme-cli is the same in an environment, so this failure should happen only once. After that, the state should be saved in the tool, and doesn't need to call without "block-size" parameter. Check how to solve the problem of sudo. The pattern is similar.

lisa/tools/nvmecli.py
```diff
@@ -18,6 +18,10 @@ class Nvmecli(Tool):
     __ns_management_attachement_support = "NS Management and Attachment Supported"
     # [1:1] : 0x1   Format NVM Supported
     __format_device_support = "Format NVM Supported"
+    # Higher nvme-cli command add a mandatory parameter `--block-size`
```
squirrelsc: It's not a "higher command", it's a "higher version". It would be great, if we know the version.

microsoft/testsuites/nvme/nvme.py
```diff
@@ -346,6 +346,12 @@ def nvme_manage_ns_validation(self, node: Node) -> None:
             format_namespace.assert_exit_code(device_format_exit_code)
             # 3. `nvme create-ns namespace` - create a namespace.
             create_namespace = nvme_cli.create_namespace(namespace)
+            # NVMe Status:INVALID_OPCODE: The associated command opcode field is not
```
squirrelsc, line 349: It's good to have comments, but it's better to explain why two different error code is here. And I don't understand how (0x1) translate to 22.

lisa/operating_system.py
```diff
@@ -754,7 +754,8 @@ def _initialize_package_installation(self) -> None:
         # Use below command to update rhui-microsoft-azure-rhel package from microsoft
         #  repo to resolve the issue.
         # Details please refer https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/redhat/redhat-rhui#azure-rhui-infrastructure # noqa: E501
-        if "Red Hat" == information.vendor:
+        # vendor name - Red Hat Enterprise Linux Server
+        if "Red Hat" in information.vendor:
```
squirrelsc: The vendor name should be "Red Hat". I will fix the logic to get it correctly.

lisa/tools/nvmecli.py
LiliDeng: Double check, only create command need specify --block-size when nvme version is greater than 1.6

microsoft/testsuites/nvme/nvme.py
LiliDeng: post an issue in nvme-cli github to ask this question.

squirrelsc (approved):

PR #1447, opened 2021-07-27
Networksetting testsuite fixes

sharsonia:


lisa/tools/find.py
```diff
@@ -25,22 +25,22 @@ def find_files(
         ignore_case: bool = False,
     ) -> List[str]:
         if not self.node.shell.exists(start_path):
-            raise LisaException(f"Path {start_path} does not exist.")
+            raise LisaException(f"Path {str(start_path)} does not exist.")
```
squirrelsc: `str` is not necessary in a string already.

microsoft/testsuites/network/networksettings.py
```diff
@@ -42,9 +42,31 @@ def validate_ringbuffer_settings_change(self, node: Node) -> None:
...
+            )
+
+            txbuffer = (
+                (original_txbuffer - 2)
+                if original_txbuffer - 2 > 0
+                else (original_txbuffer + 2)
+            )
+
+            expected_rx = int((rxbuffer * 1024 * 1024) / 1728)
+            expected_tx = int((txbuffer * 1024 * 1024) / 1728)
```
squirrelsc: The original number is 6144 above, is it correct here?

sharsonia: Thanks. Good catch

microsoft/testsuites/network/networksettings.py
```diff
@@ -96,6 +118,13 @@ def validate_device_channels_change(self, node: Node) -> None:
             channels = interface_channels_info.current_channels
             max_channels = interface_channels_info.max_channels
 
+            if max_channels <= 1:
```
squirrelsc, line 121: Is there any exception should be raised, or the test should be skipped? How about all of them are 1?

sharsonia: Yeah, we cannot attempt to change it to < 1 and if max is just 1 then there is no value that we can test it against. Hmm what will be the test case result in that case (if all the device's max channels are 1) I think you are asking about. 
may be it is safe to assume pass in such a case?

sharsonia:
Test result for - validate_device_channels_change

[lisa-20210727-195954-616.log](https://github.com/microsoft/lisa/files/6888444/lisa-20210727-195954-616.log)

microsoft/testsuites/network/networksettings.py
squirrelsc: I'm not sure if other things can be validated in this case too. If there is no, it should be skipped, because it's not testable. If there are other things to validate, we may can split it into two cases. One is skipped, and another one can be validated.

sharsonia:
Test results for - validate_device_enabled_features
[lisa-20210727-202213-322.log](https://github.com/microsoft/lisa/files/6888535/lisa-20210727-202213-322.log)

microsoft/testsuites/network/networksettings.py
sharsonia: Its like one device might have max channels as 1 but other devices (if present) might have different value.

sharsonia:
Test run result for - validate_ringbuffer_settings_change

Canonical UbuntuServer 18.04-LTS latest
[lisa-20210727-224011-145.log](https://github.com/microsoft/lisa/files/6889096/lisa-20210727-224011-145.log)

redhat rhel 7.8 latest
[lisa-20210727-232803-269.log](https://github.com/microsoft/lisa/files/6889263/lisa-20210727-232803-269.log)


2021-07-27 22:41:30.719[30812][DEBUG] lisa.RootRunner runner 'lisa_0' is done, remaining runners []
2021-07-27 22:41:30.720[30812][INFO] lisa.RootRunner ________________________________________
2021-07-27 22:41:30.720[30812][INFO] lisa.RootRunner NetworkSettings.validate_ringbuffer_settings_change: PASSED   
2021-07-27 22:41:30.720[30812][INFO] lisa.RootRunner test result summary
2021-07-27 22:41:30.720[30812][INFO] lisa.RootRunner     TOTAL    : 1
2021-07-27 22:41:30.720[30812][INFO] lisa.RootRunner     QUEUED   : 0
2021-07-27 22:41:30.720[30812][INFO] lisa.RootRunner     ASSIGNED : 0
2021-07-27 22:41:30.721[30812][INFO] lisa.RootRunner     RUNNING  : 0
2021-07-27 22:41:30.721[30812][INFO] lisa.RootRunner     FAILED   : 0
2021-07-27 22:41:30.721[30812][INFO] lisa.RootRunner     PASSED   : 1
2021-07-27 22:41:30.721[30812][INFO] lisa.RootRunner     SKIPPED  : 0
2021-07-27 22:41:30.722[30812][INFO] lisa completed in 79.579 sec

lisa/tools/ethtool.py
```diff
@@ -54,19 +53,19 @@ def __init__(self, interface: str, device_channel_raw: str) -> None:
         self._parse_channel_info(interface, device_channel_raw)
 
     def _parse_channel_info(self, interface: str, raw_str: str) -> None:
-        current_settings = _current_settings_pattern.match(raw_str)
-        max_settings = _max_settings_pattern.match(raw_str)
+        current_settings = re.search(_current_settings_pattern, raw_str)
```
squirrelsc: 1. What's the reason to change match to search? It looks similar here.
2. It can be called like _current_settings_pattern.search(raw_str). no need to use static method.

squirrelsc (dismissed):

lisa/tools/ethtool.py
sharsonia: > 1. What's the reason to change match to search? It looks similar here.
      _current_settings_pattern.match() wasn't matching (may be because it was using DOTALL flag) not sure but it was giving error.

> 2. It can be called like _current_settings_pattern.search(raw_str). no need to use static method.
ok will use it like this, test and update PR.


Tested, all paased.

squirrelsc (approved):

PR #1452, opened 2021-07-31
Azure: support vm_size partial match, and case insensitive

squirrelsc:


LiliDeng:
could you please let know the exact scenario for this change, thanks.

squirrelsc:
> could you please let know the exact scenario for this change, thanks.

1. The users may not know the accurate vm_size name, for example, they may miss standard_
2. The case may be incorrect, case insensitive is better to match.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -360,7 +360,7 @@ def _prepare_environment(  # noqa: C901
                         location_name, log
                     )
                     for azure_cap in location_info.capabilities:
-                        if azure_cap.vm_size == node_runbook.vm_size:
+                        if node_runbook.vm_size.lower() in azure_cap.vm_size.lower():
```
dsrivastavv: Couldn't this cause a conflict if node_runbook.vm_size is a prefix of azure_cap.vm_size ? For example node_runbook.vm_size = "Standard_A2" and azure_cap.vm_size = "Standard_A2m_v2"

squirrelsc: good catch, I will find a way to support it.

PR #1454, opened 2021-08-03
Add StandardSSD_LSR OS type and test case.

LiliDeng:


squirrelsc:
@somil55 Please have a look on this change.

lisa/features/__init__.py
```diff
@@ -1,7 +1,13 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-from .disks import DiskEphemeral, DiskPremiumLRS, DiskStandardLRS, DiskType
+from .disks import (
+    DiskEphemeral,
+    DiskPremiumLRS,
+    DiskStandardLRS,
```
dsrivastavv: Maybe we should rename DiskStandardLRS to DiskStandardHDDLRS to differentiate scenario better. @LiliDeng @squirrelsc Thoughts?

LiliDeng: yep, I will make the change.

dsrivastavv (approved):

LiliDeng:
@squirrelsc Could you please take a look?

squirrelsc (approved):

PR #1457, opened 2021-08-03
Support deployment with attaching disks

LiliDeng:


lisa/schema.py
```diff
@@ -381,6 +381,19 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
         default=search_space.IntRange(min=0),
         metadata=metadata(decoder=search_space.decode_count_space),
     )
+    data_disk_caching: str = field(
```
squirrelsc: `data_disk_caching_type`

squirrelsc: Move it near to `data_disk_count`. The similar things should be together.

docs/run_test/runbook.rst
```diff
@@ -663,7 +663,7 @@ nodes_requirement
 '''''''''''''''''
 
 List of testing required environments, by default node_count (default is
-1), core_count (default is 1), memory_mb (default is 512 MB), disk_count
+1), core_count (default is 1), memory_mb (default is 512 MB), data_disk_count
 (default is 1), nic_count (default is 1), gpu_count (default is 0). The
```
squirrelsc: update the comments, default is zero now.

lisa/schema.py
```diff
@@ -381,6 +381,19 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
...
+            validate=validate.OneOf(
+                [
+                    constants.DATADISK_CACHING_TYPE_NONE,
+                    constants.DATADISK_CACHING_TYPE_READONLY,
+                    constants.DATADISK_CACHING_TYPE_READYWRITE,
+                ]
+            ),
+        ),
+    )
+    data_disk_size_type: str = "S4"
```
squirrelsc: As we discussed, it should be IOPS, not disk size type. It's not translatable to other platforms. The type should be a IntRange

lisa/sut_orchestrator/azure/tests/azure_locations_westus2.json
```diff
@@ -1,866 +0,0 @@
-{
```
squirrelsc: Is it removed incorrect?

lisa/features/disks.py
```diff
@@ -62,7 +62,60 @@ class DiskType:
     DISK_EPHEMERAL: str = DiskEphemeral.name()
     DISK_PREMIUM: str = DiskPremiumLRS.name()
     DISK_STANDARD: str = DiskStandardLRS.name()
+    # refer https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types
+    DISK_SIZE_PREMIUM: Dict[str, int] = {
```
dsrivastavv: `DISK_SIZE` and `DiskType` represent two different thing. Maybe we can move `DISK_SIZE` to their specific class

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -446,6 +452,20 @@ def _prepare_environment(  # noqa: C901
                                 min_cap.nic_count, int
                             ), f"actual: {min_cap.nic_count}"
                             node_runbook.nic_count = min_cap.nic_count
+                            assert isinstance(
+                                min_cap.data_disk_count, int
+                            ), f"actual: {min_cap.data_disk_count}"
+                            node_runbook.data_disk_count = min_cap.data_disk_count
+                            assert isinstance(
+                                min_cap.data_disk_caching, str
```
dsrivastavv: This could be tested as one of the three allowed values

lisa/util/constants.py
```diff
@@ -114,3 +114,8 @@
 TESTCASE_RETRY = "retry"
 TESTCASE_USE_NEW_ENVIRONMENT = "use_new_environment"
 TESTCASE_IGNORE_FAILURE = "ignore_failure"
+
+# data disk caching type
+DATADISK_CACHING_TYPE_NONE = "None"
```
dsrivastavv, line 119: It might be better to add it as an enum similar `DiskType` in disks.py.

LiliDeng:
@squirrelsc I didn't figure out the reason for the azure_locations_westus2.json got deleted during run UT.

lisa/util/constants.py
LiliDeng: I encounter circular import issue, when use the class in the schema.py

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: It already covered in schema.py

lisa/features/disks.py
LiliDeng: fix

lisa/schema.py
LiliDeng: fix

docs/run_test/runbook.rst
LiliDeng: fix

lisa/schema.py
LiliDeng: fix

lisa/schema.py
```diff
@@ -369,8 +369,24 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
...
+                [
+                    constants.DATADISK_CACHING_TYPE_NONE,
+                    constants.DATADISK_CACHING_TYPE_READONLY,
+                    constants.DATADISK_CACHING_TYPE_READYWRITE,
+                ]
+            ),
+        ),
+    )
+    data_disk_iops: search_space.CountSpace = field(
+        default=search_space.IntRange(min=120, max=20000),
```
squirrelsc: The default value can be `min=1`. The max is not necessary. We just set a value work, but not limit it in the default value.

lisa/features/__init__.py
```diff
@@ -15,6 +16,7 @@
 from .startstop import StartStop
 
 __all__ = [
+    "DataDisk",
```
squirrelsc: The data disk doesn't need to be a feature. It's values should be translated in the Azure platform. As we added attributes, they are not part of feature, they are similar like nic_count, core_count.

lisa/features/disks.py
```diff
@@ -89,3 +89,53 @@ def get_disk_types() -> List[str]:
             DiskType.DISK_STANDARD_HDD,
             DiskType.DISK_STANDARD_SSD,
         ]
+
+
+class DataDisk:
+    # refer https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types
```
squirrelsc: This kind of logic should be under Azure platform, not in the common place.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -386,6 +386,14 @@ def _prepare_environment(  # noqa: C901
                         min_runbook.vm_size = matched_cap.vm_size
                         assert isinstance(min_cap.nic_count, int)
                         min_runbook.nic_count = min_cap.nic_count
+                        assert isinstance(min_cap.data_disk_count, int)
+                        min_runbook.data_disk_count = min_cap.data_disk_count
+                        assert isinstance(min_cap.data_disk_caching_type, str)
+                        min_runbook.data_disk_caching_type = (
+                            min_cap.data_disk_caching_type
+                        )
+                        assert isinstance(min_cap.data_disk_iops, int)
```
squirrelsc: Refer to disk type to add logic in `_node_generate_min_capability`. So we can convert iops to a min disk size.

lisa/tests/test_env_requirement.py
```diff
@@ -66,19 +71,32 @@ def ut_simple_requirement(
...
-        unsupported_platform_type=unsupported_platform_type,
-        supported_os=supported_os,
-        unsupported_os=unsupported_os,
-    )
-    return UtTestCaseRequirement(
-        environment=simple.environment,
-        platform_type=simple.platform_type,
-        os_type=simple.os_type,
-    )
+    if node:
```
squirrelsc: It can be split to two methods as the requirements.

lisa/tests/test_env_requirement.py
```diff
@@ -17,7 +17,12 @@
     generate_min_capability,
 )
 from lisa.tests.test_search_space import SearchSpaceTestCase
-from lisa.testsuite import DEFAULT_REQUIREMENT, TestCaseRequirement, simple_requirement
+from lisa.testsuite import (
```
squirrelsc: The commit description "Update UT" is not clear. The "Update UT" explains what does, but it's obvious. It should explain the reason to change, or something is not obvious. For example, "Fix UT which is failed by simple_requirement refactoring". But we should try to make every commit works, including UT. So the UT fix should be in the same commit of the requirement change.

lisa/testsuite.py
```diff
@@ -285,6 +293,64 @@ def simple_requirement(
...
 
+def node_requirement(
+    node: Optional[schema.NodeSpace] = None,
+    supported_platform_type: Optional[List[str]] = None,
+    unsupported_platform_type: Optional[List[str]] = None,
+    supported_os: Optional[List[Type[OperatingSystem]]] = None,
+    unsupported_os: Optional[List[Type[OperatingSystem]]] = None,
+    environment_status: EnvironmentStatus = EnvironmentStatus.Connected,
+) -> TestCaseRequirement:
+    node = _construct_node(node)
```
squirrelsc: in this method, the node should be a required argument, and not need to construct. It's default value will be taken.

lisa/testsuite.py
```diff
@@ -237,10 +237,26 @@ class TestCaseRequirement:
     os_type: Optional[search_space.SetSpace[Type[OperatingSystem]]] = None
 
 
-def simple_requirement(
+def _construct_node(
```
squirrelsc: It should be used in simple_requirement only, so it doesn't need to be a separated method.

lisa/testsuite.py
```diff
@@ -237,10 +237,26 @@ class TestCaseRequirement:
...
+        node = schema.NodeSpace()
+        node.node_count = search_space.IntRange(min=min_count)
+        node.nic_count = search_space.IntRange(min=min_nic_count)
+        node.data_disk_count = search_space.IntRange(min=min_data_disk_count)
+        node.data_disk_caching_type = data_disk_caching_type
+        node.data_disk_iops = data_disk_iops
+    return node
+
+
+def _construct_test_case_requirement(
```
squirrelsc: "create" is more popular than "construct" here.

lisa/testsuite.py
```diff
@@ -237,10 +237,26 @@ class TestCaseRequirement:
...
+def _construct_node(
+    node: Optional[schema.NodeSpace] = None,
     min_count: int = 1,
     min_nic_count: int = 1,
-    node: Optional[schema.NodeSpace] = None,
+    min_data_disk_count: int = 0,
+    data_disk_caching_type: str = constants.DATADISK_CACHING_TYPE_NONE,
+    data_disk_iops: int = 120,
+) -> schema.NodeSpace:
+    if node is None:
```
squirrelsc: After the logic is in simple_requirement, the `if` clause is not needed.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1516,3 +1538,68 @@ def _get_disk_id(node_features: search_space.SetSpace[str]) -> str:
...
+        16000: 8192,
+        18000: 16384,
+        20000: 32767,
+    }
+    IOPS_SIZE_STANDARD_HDD: Dict[int, int] = {
+        500: 32,
+        1300: 8192,
+        2000: 16384,
+    }
+    IOPS_SIZE_STANDARD_SDD: Dict[int, int] = {
```
squirrelsc: SDD -> SSD

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1516,3 +1538,68 @@ def _get_disk_id(node_features: search_space.SetSpace[str]) -> str:
...
+            raise LisaException(f"IOPS {data_disk_iops} is invaild.")
+        return iops_dict[min_iops]
+
+    @staticmethod
+    def get_size_by_iops(disk_type: str, data_disk_iops: int = 120) -> int:
+        if DiskType.DISK_PREMIUM == disk_type:
+            return DataDisk.get_min_size(DataDisk.IOPS_SIZE_PREMIUM, data_disk_iops)
+        elif DiskType.DISK_STANDARD_HDD == disk_type:
+            return DataDisk.get_min_size(
+                DataDisk.IOPS_SIZE_STANDARD_SDD, data_disk_iops
```
squirrelsc: Should it be SSD? This logic can be like below,

```
if disk_type in [DiskType.DISK_PREMIUM, DiskType.DISK_STANDARD_HDD, DiskType.DISK_STANDARD_SSD]:
    return DataDisk.get_min_size(
                disk_type, data_disk_iops
            )
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1516,3 +1538,68 @@ def _get_disk_id(node_features: search_space.SetSpace[str]) -> str:
...
+    @staticmethod
+    def get_min_size(iops_dict: Dict[int, int], data_disk_iops: int = 120) -> int:
+        iops_array = iops_dict.keys()
+        min_iops = min([key for key in iops_array if key >= data_disk_iops])
+        if not min_iops:
+            raise LisaException(f"IOPS {data_disk_iops} is invaild.")
+        return iops_dict[min_iops]
+
+    @staticmethod
+    def get_size_by_iops(disk_type: str, data_disk_iops: int = 120) -> int:
```
squirrelsc: `get_size` is enough, we may add more parameters in future.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1516,3 +1538,68 @@ def _get_disk_id(node_features: search_space.SetSpace[str]) -> str:
...
+    }
+    IOPS_SIZE_STANDARD_SDD: Dict[int, int] = {
+        500: 4,
+        2000: 8192,
+        4000: 16384,
+        6000: 32767,
+    }
+
+    @staticmethod
+    def get_min_size(iops_dict: Dict[int, int], data_disk_iops: int = 120) -> int:
```
squirrelsc: It's called in this class only, a private method is clearer. And put it under other public methods.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1516,3 +1538,68 @@ def _get_disk_id(node_features: search_space.SetSpace[str]) -> str:
             return features.DiskStandardSSDLRS.get_disk_id()
         else:
             raise LisaException("No disk feature present.")
+
+    @staticmethod
+    def _get_data_disk_size_by_iops(
```
squirrelsc: `_get_data_disk_size` And move related code to another file like common.py. The platform.py focus on platform.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1477,6 +1493,12 @@ def _node_generate_min_capability(
 
         # Generate min capability node
         min_cap: schema.NodeSpace = req_cap.generate_min_capability(azure_cap_copy)
+        assert isinstance(min_cap.data_disk_count, int)
```
squirrelsc: This logic should be above the `req_cap.generate_min_capability(azure_cap_copy)`, so the input may be IntRange. If not, when the iops is 1, the min size will be 1 above, and this logic doesn't know it needs an exact 1, min 1, or max 1. And it needs to more complex logic to check which iops data in the range in data disk's get min size.

lisa/testsuite.py
LiliDeng: but it is also possible that user pass a none node?

LiliDeng: fix

lisa/testsuite.py
LiliDeng: fix

lisa/testsuite.py
LiliDeng: fix

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: fix

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: fix

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: fix

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: fix

lisa/sut_orchestrator/azure/common.py
```diff
@@ -368,3 +383,70 @@ def wait_copy_blob(
...
+    IOPS_SIZE_STANDARD_SSD: Dict[int, int] = {
+        500: 4,
+        2000: 8192,
+        4000: 16384,
+        6000: 32767,
+    }
+
+    @staticmethod
+    def get_size(disk_type: str, data_disk_iops: int = 120) -> int:
+        if disk_type in [
```
squirrelsc, line 433: This method is simple enough now. So the content of `_get_min_size` can be merged with this one.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -368,3 +383,70 @@ def wait_copy_blob(
...
+        return DataDisk.get_size(DiskType.DISK_STANDARD_HDD, data_disk_iops)
+    elif DiskType.DISK_STANDARD_SSD in node_features:
+        return DataDisk.get_size(DiskType.DISK_STANDARD_SSD, data_disk_iops)
+    else:
+        raise LisaException("No data disk feature present.")
+
+
+class DataDisk:
+    # refer https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types
+    IOPS_SIZE_PREMIUM: Dict[int, int] = {
```
squirrelsc: If it's two level dict, the code will be simpler.

iops_mapping:Dict[DiskType, Dict[int, int] = {
    IOPS_SIZE_PREMIUM: {
       120: 4,
       240:64,
       ...
   }
}

lisa/sut_orchestrator/azure/common.py
```diff
@@ -90,6 +90,21 @@ class AzureNodeSchema:
...
+            validate=validate.OneOf(
+                [
+                    constants.DATADISK_CACHING_TYPE_NONE,
+                    constants.DATADISK_CACHING_TYPE_READONLY,
+                    constants.DATADISK_CACHING_TYPE_READYWRITE,
+                ]
+            )
+        ),
+    )
+    data_disk_iops: int = 1
```
squirrelsc: In Azure, it can be hard code to low end one, which is good to run test.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -916,6 +937,10 @@ def _create_deployment_parameters(
 
             # Set disk type
             azure_node_runbook.disk_id = AzurePlatform._get_disk_id(node_space.features)
+            assert isinstance(node_space.data_disk_iops, int)
+            azure_node_runbook.data_disk_size = get_data_disk_size(
```
squirrelsc: It should be resolved in `_node_generate_min_capability`. What's the reason to move it to here?

lisa/schema.py
```diff
@@ -546,12 +573,14 @@ def _generate_min_capability(self, capability: Any) -> Any:
...
         else:
             raise LisaException("memory_mb cannot be zero")
-        if self.disk_count or capability.disk_count:
-            min_value.disk_count = search_space.generate_min_capability_countspace(
-                self.disk_count, capability.disk_count
+        if self.data_disk_count or capability.data_disk_count:
+            min_value.data_disk_count = search_space.generate_min_capability_countspace(
+                self.data_disk_count, capability.data_disk_count
+            )
+        if self.data_disk_iops or capability.data_disk_iops:
```
squirrelsc: The disk iops cannot be done here, because it's not an integer on Azure. In Azure, we cannot select any integer as iops, it depends on the data_size. So the min capability must be resolved in Azure platform.

lisa/sut_orchestrator/azure/common.py
LiliDeng: Sorry, I don't understand.

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: Current code logs meet the requirement, I tested with different IOPS.

squirrelsc: Anyway, I will handle it in feature's refactoring

lisa/sut_orchestrator/azure/common.py
squirrelsc: set to a default one, which is testable. For example, 4G disk size with 120 iops. It's the default value.

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: Let's sync up again tomorrow, data_disk_iops is always IntRange here, then get the min value of IOPS combined with disk type when data disk count more than 0 is current approach.

squirrelsc (approved):

PR #1458, opened 2021-08-03
test case: support timeout

squirrelsc:


glima:
Worth mentioning that test suite's before/after calls also get that timeout (not customizable).

PR #1464, opened 2021-08-06
Add some enhancement into LISAv2

lubaihua33:
There is no other architecture's package of Docker Compose. So skip the docker compose test case if not x86_64.

Here is the kdump test results:
```
   ID TestArea             TestCaseName                                                                TestResult TestDuration(in minutes) 
-------------------------------------------------------------------------------------------------------------------------------------------
    1 KDUMP                KDUMP-CRASH-16-CORES                                                              PASS                 11.4 
    2 KDUMP                KDUMP-CRASH-AUTO-SIZE                                                          SKIPPED                  4.5 
    3 KDUMP                KDUMP-CRASH-DIFFERENT-VCPU                                                        PASS                 11.2 
```

### DIFF
--- a/Testscripts/Linux/KDUMP-Execute.sh
+++ b/Testscripts/Linux/KDUMP-Execute.sh
@@ -149,13 +149,23 @@ Exec_mariner()
 
 Configure_NMI()
 {
-    sysctl -w kernel.unknown_nmi_panic=1
-    if [ $? -ne 0 ]; then
-        LogErr "Failed to enable kernel to call panic when it receives a NMI."
-        SetTestStateAborted
-        exit 0
+    # /proc/sys/kernel/unknown_nmi_panic:
+    # The value in this file affects behavior of handling NMI. When the value is
+    # non-zero, unknown NMI is trapped and then panic occurs. If need to dump the
+    # crash, the value should be set 1.
+    # Some architectures don't provide architected NMIs,such as ARM64. So this file
+    # doesn't exist.
+    if [ -e "/proc/sys/kernel/unknown_nmi_panic" ] ; then
+        sysctl -w kernel.unknown_nmi_panic=1
+        if [ $? -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


Testscripts/Linux/docker_compose_wordpress_mysql_app.sh
```diff
@@ -84,6 +84,10 @@ function EvaluateTestResult() {
 }
 UtilsInit
 
+if [[ $(uname -m) != "x86_64" ]];then
+    HandleSkip "The Docker Compose doesn't support other architecture except x86_64" 0
```
LiliDeng: do you have a link to say this?

lubaihua33: Have modified. Added some comments above the code line.

Testscripts/Linux/KDUMP-Execute.sh
```diff
@@ -149,13 +149,17 @@ Exec_mariner()
...
+        if [ $? -ne 0 ]; then
+            LogErr "Failed to enable kernel to call panic when it receives a NMI."
+            SetTestStateAborted
+            exit 0
+        else
+            UpdateSummary "Success: enabling kernel to call panic when it receives a NMI."
+        fi
     else
-        UpdateSummary "Success: enabling kernel to call panic when it receives a NMI."
+        UpdateSummary "The kernel doesn't support to call panic when it receives a NMI."
```
LiliDeng: Should we set test result skip here and do you know what cause this path doesn't exist? /proc/sys/kernel/unknown_nmi_panic

lubaihua33: Some architectures don't provide architected NMIs,such as ARM64. So this file doesn't exist. I added comments. This is configured for Hyper-V testing. We can just ignore it and keep running. It won't affect kdump.

LiliDeng (approved):

PR #1466, opened 2021-08-09
Add netsetting testcases

sharsonia:


lisa/tools/ethtool.py
```diff
@@ -221,6 +221,15 @@ def _parse_ring_buffer_settings_info(self, interface: str, raw_str: str) -> None
             )
 
 
+class DeviceGroLroSettings:
+    def __init__(
+        self, interface: str, gro_setting: str = "off", lro_setting: str = "off"
```
squirrelsc: Use True, False is easier to write logic, and avoid typo. They can be translated to "on" and "off" when using in command.

sharsonia: Yeah, I myself debated about it as well and thought would need translation while validating as well. But cool, can have True/False stored in map

lisa/tools/ethtool.py
```diff
@@ -237,6 +246,7 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._device_features_map: Dict[str, DeviceFeatures] = {}
         self._device_link_settings_map: Dict[str, DeviceLinkSettings] = {}
         self._device_ring_buffer_settings_map: Dict[str, DeviceRingBufferSettings] = {}
+        self._device_gro_lro_settings_map: Dict[str, DeviceGroLroSettings] = {}
```
squirrelsc: How about to use one dict? So all information of a device can be put in the same structure. And the values can be optional, if it's not queried.

microsoft/testsuites/network/networksettings.py
```diff
@@ -184,3 +185,59 @@ def validate_device_enabled_features(self, node: Node) -> None:
...
+
+        for device_settings in devices_gro_lro_settings:
+            interface = device_settings.interface
+            gro_setting = device_settings.gro_setting
+            lro_setting = device_settings.lro_setting
+
+            new_gro_setting = "off"
+            new_lro_setting = "off"
+
+            if "off" in gro_setting:
```
squirrelsc: If it's a boolean, the code is like `new_gro_setting =  not gro_setting`

lisa/tools/ethtool.py
sharsonia: ok, can add it in another commit on top of the gro_lro test case

sharsonia:
Test result for testcase  validate_device_gro_lro_settings_change
[lisa-20210809-211309-559.log](https://github.com/microsoft/lisa/files/6957490/lisa-20210809-211309-559.log)

lisa/tools/ethtool.py
squirrelsc: Do you mean in the same PR or next PR?

microsoft/testsuites/network/networksettings.py
```diff
@@ -185,3 +185,54 @@ def validate_device_enabled_features(self, node: Node) -> None:
...
+        priority=1,
+    )
+    def validate_device_gro_lro_settings_change(self, node: Node) -> None:
+        ethtool = node.tools[Ethtool]
+
+        devices_gro_lro_settings = ethtool.get_all_device_gro_lro_settings()
+
+        for device_settings in devices_gro_lro_settings:
+            interface = device_settings.interface
+            gro_setting = device_settings.gro_setting
```
squirrelsc: the name "original_gro_setting" is clearer

lisa/tools/ethtool.py
sharsonia: Done in same PR with 4th commit

sharsonia:
Result for all networksetting  test cases after the latest commit (fourth) for replacing multiple feature dict with a single device_settings dict in ethtool
[lisa-20210809-223505-229.log](https://github.com/microsoft/lisa/files/6957780/lisa-20210809-223505-229.log)

lisa/tools/ethtool.py
```diff
@@ -318,7 +336,12 @@ def get_device_channels_info(
         if vcpu_count < device_channel_info.max_channels:
             device_channel_info.max_channels = vcpu_count
 
-        self._device_channel_map[interface] = device_channel_info
+        if interface not in self._device_settings_map.keys():
```
squirrelsc: The logic is similar, and it can be a method like
```python
def _set_device(name: str, device_channels: Optional[DeviceChannel] = None,
        device_features: Optional[DeviceFeatures] = None,
        device_link_settings: Optional[DeviceLinkSettings] = None,
        device_ringbuffer_settings: Optional[DeviceRingBufferSettings] = None,
        device_gro_lro_settings: Optional[DeviceGroLroSettings] = None)
    ...
```

sharsonia: Thanks, changed per our offline discussion

squirrelsc (approved):

PR #1467, opened 2021-08-10
Add extension API reference

WenyiQ:
Utilizes Sphinx ``autodoc`` to document API for extensions.


squirrelsc:
@WenyiQ Can you share the example?

WenyiQ:
Preview: https://mslisa--1467.org.readthedocs.build/en/1467/api/extensions.html
There seems to be some problem building on rtd, I'm fixing it right now.

WenyiQ:
> Preview: https://mslisa--1467.org.readthedocs.build/en/1467/api/extensions.html
> There seems to be some problem building on rtd, I'm fixing it right now.

UPDATE: Path issues are fixed, but there are also environment issues. Poetry may be needed on rtd again.
One part of the error log is attached below:
```
...
WARNING: autodoc: failed to import class 'notifier.MessageBase' from module 'lisa'; the following exception was raised:
No module named 'dataclasses_json'
WARNING: autodoc: failed to import class 'notifier.Notifier' from module 'lisa'; the following exception was raised:
No module named 'dataclasses_json'
WARNING: autodoc: failed to import class 'notifier.TestRunMessage' from module 'lisa'; the following exception was raised:
No module named 'dataclasses_json'
...
```

squirrelsc:
> dataclasses_json

You need to include the package missed. I think the poetry is too heavy to rtd. Check the "pytparoject.toml", and add any package, which is reported missed. Don't add all of them, because most of them are not necessary.

glima:
I'm confused. What's in here that a

.. autosummary::
   :toctree: _autosummary
   :caption: API
   :recursive:

entry to index.rst can't do?

glima:
(of course missing module entries after that)

WenyiQ:
> > dataclasses_json
> 
> You need to include the package missed. I think the poetry is too heavy to rtd. Check the "pytparoject.toml", and add any package, which is reported missed. Don't add all of them, because most of them are not necessary.

I agree, that would also work.

WenyiQ:
> I'm confused. What's in here that a
> 
> .. autosummary::
> :toctree: _autosummary
> :caption: API
> :recursive:
> 
> entry to index.rst can't do?

I've not tried that before as I was sticking with ``autodoc``. It actually seems to me a better solution.

squirrelsc:
Is there update on preview? I saw it's not in right format.

WenyiQ:
> Is there update on preview? I saw it's not in right format.

Sorry it was mainly an update on ``autosummary``. I'm still solving environment issues.

docs/conf.py
```diff
@@ -10,12 +10,44 @@
...
+
+root_dir = Path(__file__).parents[1]
+pyproj = root_dir / "pyproject.toml"
+requirement = root_dir / "docs" / "requirements.txt"
+
+sys.path.insert(0, str(root_dir.resolve()))
+sys.path.insert(0, str((root_dir / "docs" / "tools").resolve()))
+sys.path.insert(0, str((root_dir / "lisa").resolve()))
+
+# -- Requirements ------------------------------------------------------------
```
glima: Why can't we just let Poetry handle this?

docs/conf.py
```diff
@@ -68,11 +115,13 @@
     "display_version": False,
 }
 
-# -- Test pipelines ----------------------------------------------------------
+# -- Test auto-generation pipelines ------------------------------------------
 
 base_path = Path(__file__).parent
 test_table_pipeline = base_path / "tools/test_table_gen.py"
 test_spec_pipeline = base_path / "tools/test_spec_gen.py"
+ext_api_pipeline = base_path / "tools/api_gen.py"
```
glima: Aren't you using a Sphinx module for that now?

docs/tools/api_paths.yaml
```diff
@@ -0,0 +1,12 @@
+- name: notifier
```
glima: Not used anywhere?

docs/tools/api_gen.py
```diff
@@ -0,0 +1,128 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""
+Generates API reference for extensions.
+"""
+
```
glima: Again, Sphinx plugins can handle those just fine. Why?

docs/conf.py
WenyiQ: Yep, I included both extensions. The reason I kept both was because I was comparing ``autodoc`` with ``autosummary``, in the support of custom configurations, for example, selecting only specific classes/functions, and adding cross-reference in other files.

docs/tools/api_paths.yaml
WenyiQ: It was called in ``api_gen.py`` as ``EXTS``, which was defined in ``doc_generator.py`` as a path constant.

docs/conf.py
WenyiQ: You mean these ``sys.path``-s? Actually I think it would work without all these, I'll try it and if so I will remove them.

WenyiQ:
After discussion with @squirrelsc, minimized extension API to in-file calls to ``autodoc`` generation.

squirrelsc (dismissed):

docs/conf.py
glima: No, paths are needed, I meant requirements.

squirrelsc:
@WenyiQ The below parts are not showing up in the page. Is any setting needed?
https://mslisa.readthedocs.io/en/wenyi-docs-api/write_test/extension.html
```rst
.. autoclass:: lisa.transformer.Transformer
```

docs/conf.py
WenyiQ: Alright. My first attempt was to let Poetry handle it, but I ran into an out-of-memory problem on rtd (possibly because of too many packages and too long an installation time), so I had to switch to this solution.

WenyiQ:
I erroneously removed ``sys.path``-s, now they are added, and it should work fine.

squirrelsc (approved):

squirrelsc:
@WenyiQ Thank you for the great contribution!

PR #1468, opened 2021-08-10
Support use_new_environment on test metadata, and fix some bugs.

squirrelsc:
See commits for details


glima:
Looks OK/harmless to me, though I would split the middle commit in two (and document more what's being done in the commit message).

PR #1471, opened 2021-08-12
Fix lro default setting

sharsonia:


sharsonia:
Test results for the configuration that was showing failure. With the fix in these commits, test is passing.
[lisa-20210812-062929-786.log](https://github.com/microsoft/lisa/files/6973298/lisa-20210812-062929-786.log)

microsoft/testsuites/network/networksettings.py
```diff
@@ -204,13 +206,24 @@ def validate_device_gro_lro_settings_change(self, node: Node) -> None:
...
-        for device_settings in devices_gro_lro_settings:
-            interface = device_settings.interface
-            original_gro_setting = device_settings.gro_setting
-            original_lro_setting = device_settings.lro_setting
+        for settings in devices_gro_lro_settings:
+            interface = settings.interface
+            if settings.gro_fixed and settings.lro_fixed:
+                self.__log.info(
+                    "The GRO and LRO settings are fixed and cannot be changed for"
+                    f" device {interface}. Skipping test for this device"
```
squirrelsc, line 222: Is it possible that all devices are fixed? If so, it needs to raise a skipped exception.

sharsonia: The thing is, we still want to continue test for other device interfaces? If we raise exception from here then we are not testing the rest of the devices.

squirrelsc: Yes, is it possible that all devices are fixed? Or some of them are fixed?

sharsonia: Yes both scenarios are possible i.e. all devices can have settings as [fixed] and/or some of them have [fixed].

If we really wnat to raise skipped exception if all devices have both settings fixed then I can use a counter for skipping part and check at the end of the loop if counter is equal to number of devices. Thoughts?

squirrelsc: ok. there can be a flag. If all are fixed, it should raise a skipped exception.

squirrelsc: A bool flag called 'can_set' is enough. The initial value is False. If any is not fixed, it will be set to True. If it's False, after the loop, raise the skipped exception.

microsoft/testsuites/network/networksettings.py
```diff
@@ -131,7 +132,10 @@ def validate_device_channels_change(self, node: Node, log: Logger) -> None:
                     f"Max channels for device {interface} is <= 1."
                     " Not attempting to change, Skipping."
                 )
+                skip_test = True
```
squirrelsc: In this kind of pattern, it should be one way setting. It means, if it turns to True, no code should turn it back. If turn it back, it can only check last value, not all.

squirrelsc: In another word, "all <=1" means "no any >1". So, if any > 1, the case won't be skipped.

sharsonia: yeah so my thought was that we don't want to mark the full test case as "Skipped" if we tested any of the devices positively? Let me know if that should not be the case

squirrelsc: The code logic is like below:
```python
skip_test = True

for item in items:
    if item is testable:
        skip_test = False

if skip_test:
   raise SkippedException(....
```

microsoft/testsuites/network/networksettings.py
```diff
@@ -192,25 +200,39 @@ def validate_device_enabled_features(self, node: Node) -> None:
...
     )
     def validate_device_gro_lro_settings_change(self, node: Node) -> None:
         ethtool = node.tools[Ethtool]
 
+        skip_test = True
         devices_gro_lro_settings = ethtool.get_all_device_gro_lro_settings()
+        for settings in devices_gro_lro_settings:
+            interface = settings.interface
+            if settings.gro_fixed and settings.lro_fixed:
+                self.__log.info(
```
squirrelsc: Sorry, not find it at first time. Please refer to latest code to get the log object. It should be from method arguments, not test suite. So that the log can be visible in the test case level log.

squirrelsc (approved):

PR #1472, opened 2021-08-12
Isaiah/lisa2

isaiahduncan:
[LISAv2] Skip NVME-MAX-DISK-VALIDATION test strictly for centos 7.5 and below

[DPDK] added /usr/local/lib64/pkgconfig/ to PKG_CONFIG_PATH
Problem: pkgconfig on RHEL 7 was not checking /usr/local/lib64/pkgconfig/ path
Fix: added /usr/local/lib64/pkgconfig/ to PKG_CONFIG_PATH

### DIFF
--- a/Testscripts/Linux/run_dpdk_ring_latency.sh
+++ b/Testscripts/Linux/run_dpdk_ring_latency.sh
@@ -8,6 +8,7 @@ HOMEDIR=$(pwd)
 export RTE_SDK="${HOMEDIR}/dpdk"
 export RTE_TARGET="x86_64-native-linuxapp-gcc"
 export DPDK_RING_PING_PATH="${HOMEDIR}/dpdk-ring-ping"
+export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/usr/local/lib64/pkgconfig/"
 
 # Source utils.sh
 . utils.sh || {
--- a/Testscripts/Linux/nvme_basic_validation.sh
+++ b/Testscripts/Linux/nvme_basic_validation.sh
@@ -19,6 +19,16 @@
 }
 UtilsInit
 
+# Skip NVME-MAX-DISK-VALIDATION for centos_7.5 and below
+GetDistro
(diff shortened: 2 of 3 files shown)


isaiahduncan:
@johnsongeorge-w

LiliDeng:
@isaiahduncan Why we need change NVME case for centos?

isaiahduncan:
@LiliDeng We are skipping this specific test for only certain distros because there is a known issue with this test in certain Cent OS distros (7.5 and below) and we don't want the issue to cause unnecessary noise. 
More detail here: https://microsoft.visualstudio.com/OS/_workitems/edit/33193680

Testscripts/Linux/run_dpdk_ring_latency.sh
```diff
@@ -8,6 +8,7 @@ HOMEDIR=$(pwd)
 export RTE_SDK="${HOMEDIR}/dpdk"
 export RTE_TARGET="x86_64-native-linuxapp-gcc"
 export DPDK_RING_PING_PATH="${HOMEDIR}/dpdk-ring-ping"
+export PKG_CONFIG_PATH="$PKG_CONFIG_PATH:/usr/local/lib64/pkgconfig/"
```
johnsongeorge-w, line 11: Why this change is required?

isaiahduncan: this is actually a change that matthew and I had made earlier. It is a fix to a dpdk test

isaiahduncan: The problem that prompted this change was that pkgconfig on RHEL 7 was not checking /usr/local/lib64/pkgconfig/ path
The fix for this problem was to add /usr/local/lib64/pkgconfig/ to PKG_CONFIG_PATH

isaiahduncan: @johnsongeorge-w I can remove the commit with the "export" change if you would like so that only my changes to the NVME test case are merged.

johnsongeorge-w (approved):

PR #1473, opened 2021-08-13
Add l3_cache_check test

dsrivastavv:
Migrate test l3_cache_check : https://github.com/microsoft/lisa/blob/master/Testscripts/Linux/L3Cache-Verify.sh

### DIFF
--- a/lisa/base_tools/uname.py
+++ b/lisa/base_tools/uname.py
@@ -4,21 +4,24 @@
 import re
 from dataclasses import dataclass
 
+from semver import VersionInfo
+
 from lisa.executable import Tool
-from lisa.util import LisaException
+from lisa.util import LisaException, parse_version
 
 
 @dataclass
 class UnameResult:
     has_result: bool
-    kernel_version: str = ""
+    kernel_version: VersionInfo
+    kernel_version_raw: str = ""
     hardware_platform: str = ""
     operating_system: str = ""
     uname_version: str = ""
 
     def __repr__(self) -> str:
         return (
-            f"kernel version: {self.kernel_version}, "
+            f"kernel version: {self.kernel_version_raw}, "
             f"hardware platform: {self.hardware_platform}, "
             f"operating system: {self.operating_system}, "
             f"uname version: {self.uname_version}, "
@@ -45,16 +48,16 @@ class Uname(Tool):
         cmd_result = self.run(
             "-vrio", force_run=force_run, no_error_log=no_error_log, no_info_log=True
         )
-
         if cmd_result.exit_code != 0:
-            result = UnameResult(False, "", "", "", "")
+            result = UnameResult(False, VersionInfo(0))
         else:
             match_result = self._key_info_pattern.fullmatch(cmd_result.stdout)
             if not match_result:
                 raise LisaException(f"no result matched, stdout: '{cmd_result.stdout}'")
             result = UnameResult(
                 has_result=True,
-                kernel_version=match_result.group("kernel_version"),
+                kernel_version=parse_version(match_result.group("kernel_version")),
+                kernel_version_raw=match_result.group("kernel_version"),
                 uname_version=match_result.group("uname_version"),
                 hardware_platform=match_result.group("platform"),
                 operating_system=match_result.group("os"),
--- a/lisa/tools/lscpu.py
+++ b/lisa/tools/lscpu.py
@@ -3,7 +3,7 @@
 
 import re
 from enum import Enum
-from typing import Any, Optional, Type
+from typing import Any, List, Optional, Type
 
 from assertpy import assert_that
 
@@ -15,12 +15,38 @@ CpuType = Enum(
 )
 
 
+class CPUInfo:
+    def __init__(
+        self,
+        cpu: str,
+        numa_node: str,
+        socket: str,
+        l1_data_cache: str,
+        l1_instruction_cache: str,
+        l2_cache: str,
+        l3_cache: str,
+    ) -> None:
+        self.cpu = cpu
+        self.numa_node = numa_node
+        self.socket = socket
+        self.l1_data_cache = l1_data_cache
+        self.l1_instruction_cache = l1_instruction_cache
+        self.l2_cache = l2_cache
+        self.l3_cache = l3_cache
+
+
 class Lscpu(Tool):
     # CPU(s):              16
     __vcpu_sockets = re.compile(r"^CPU\(s\):[ ]+([\d]+)\r?$", re.M)
     # Architecture:        x86_64
     __architecture_pattern = re.compile(r"^Architecture:\s+(.*)?\r$", re.M)
     __valid_architecture_list = ["x86_64"]
+    # Capture patterns of the form `0 0 0 0:0:0:0`
+    _core_numa_mappings = re.compile(
+        r".*(?P<cpu>\d+).+(?P<numa_node>\d+).+(?P<socket>\d+)"
+        r".+(?P<l1_data_cache>\d+):(?P<l1_instruction_cache>\d+)"
+        r":(?P<l2_cache>\d+):(?P<l3_cache>\d+)$"
+    )
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._core_count: Optional[int] = None
@@ -70,6 +96,33 @@ class Lscpu(Tool):
         elif "GenuineIntel" in result.stdout:
             return CpuType.Intel
         return CpuType.Intel
+
+    def get_cpu_info(self) -> List[CPUInfo]:
+        # `lscpu --extended=cpu,node,socket,cache` command return the
+        # cpu info in the format :
+        # CPU NODE SOCKET L1d:L1i:L2:L3
+        # 0    0        0 0:0:0:0
+        # 1    0        0 0:0:0:0
+        result = self.run("--extended=cpu,node,socket,cache").stdout
+        mappings_with_header = result.splitlines(keepends=False)
+        mappings = mappings_with_header[1:]
+        assert len(mappings) > 0
+        output: List[CPUInfo] = []
+        for item in mappings:
+            match_result = self._core_numa_mappings.fullmatch(item)
+            assert match_result
+            output.append(
+                CPUInfo(
+                    cpu=match_result.group("cpu"),
+                    numa_node=match_result.group("numa_node"),
+                    socket=match_result.group("socket"),
+                    l1_data_cache=match_result.group("l1_data_cache"),
+                    l1_instruction_cache=match_result.group("l1_instruction_cache"),
+                    l2_cache=match_result.group("l2_cache"),
+                    l3_cache=match_result.group("l3_cache"),
+                )
+            )
+        return output
 
 
 class WindowsLscpu(Lscpu):
--- a/microsoft/testsuites/core/cpu.py
+++ b/microsoft/testsuites/core/cpu.py
@@ -1,0 +1,67 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from assertpy.assertpy import assert_that
+
+from lisa import Logger, testsuite
+from lisa.base_tools.cat import Cat
+from lisa.base_tools.uname import Uname
+from lisa.node import Node
+from lisa.tools.lscpu import Lscpu
+from lisa.util import SkippedException
+
+
+@testsuite.TestSuiteMetadata(
+    area="cpu",
+    category="functional",
+    description="""
+    This test suite is used to run CPU related tests.
+    """,
+)
+class CPU(testsuite.TestSuite):
+    @testsuite.TestCaseMetadata(
+        description="""
+        This test case will check that L3 cache is correctly mapped
+        to NUMA node.
+        Steps:
+        1. Check if NUMA is disabled in commandline. If disabled,
+        and kernel version is <= 2.6.37, test is skipped as hyper-v
+        has no support for NUMA : https://t.ly/x8k3
+        2. Get the mappings using command :
+        `lscpu --extended=cpu,node,socket,cache`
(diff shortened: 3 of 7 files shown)


lisa/tools/lscpu.py
```diff
@@ -15,6 +15,19 @@
...
 
 
+class CPUNodeInfo:
+    def __init__(
+        self, cpu: int, node: int, socket: int, l1i: int, l1d: int, l2: int, l3: int
+    ) -> None:
+        self.cpu = cpu
+        self.node = node
+        self.socket = socket
+        self.l1i = l1i
```
LiliDeng: the name is not that straightforward.

lisa/tools/lscpu.py
```diff
@@ -71,6 +84,23 @@ def get_cpu_type(self, force_run: bool = False) -> CpuType:
             return CpuType.Intel
         return CpuType.Intel
 
+    def get_cpu_info(self) -> List[CPUNodeInfo]:
+        result = self.run("--extended=cpu,node,socket,cache").stdout
+        mappings = result.splitlines(keepends=False)
+        assert len(mappings) > 1
+        output: List[CPUNodeInfo] = []
+        for item in mappings[1:]:
+            # Ignore first line as it contains column names.
```
LiliDeng: could you add some sample output here?

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,71 @@
...
+)
+class CPU(testsuite.TestSuite):
+    @testsuite.TestCaseMetadata(
+        description="""
+        This test case will check that L3 cache is correctly mapped
+        to NUMA node.
+        Steps:
+        1. Check if NUMA is disabled in commandline. If disabled,
+        and kernel version is <= 2.6.37, test is skipped as kernel
+        has no support for NUMA.
```
LiliDeng: is that possible to add commit (link) which supports NUMA here?

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,71 @@
...
+        """,
+        priority=0,
+    )
+    def l3_cache_check(self, node: Node) -> None:
+        cmdline = node.tools[Cat].run("/proc/cmdline").stdout
+        if NUMA_OFF in cmdline:
+            kernel_version: KernelVersion = (
+                node.tools[Uname].get_linux_information().kernel_version
+            )
+            self.__log.debug(
```
sharsonia: Use the log param in method for logging here, so this log line can show up in per test case log file.
Like -
def l3_cache_check(self, node: Node, log: Logger)

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,71 @@
...
+        CPU NODE SOCKET L1d L1i L2 L3
+        8   0    0      8   8   8  0
+        9   1    1      9   9   9  1
+
+        Incorrect mapping:
+        CPU NODE SOCKET L1d L1i L2 L3
+        8   0    0      8   8   8  8
+        9   1    1      9   9   9  9
+        """,
+        priority=0,
```
LiliDeng: Please double check the priority.

LiliDeng: I think this case has requirement for this numa count, if use a minimal requirement, we may miss this check.

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,71 @@
...
+            self.__log.debug(
+                "Found numa=off in /proc/cmdline. Checking the kernel version."
+            )
+            if (
+                kernel_version.major <= 2
+                and kernel_version.minor <= 6
+                and kernel_version.patch <= 37
+            ):
+                raise SkippedException(
+                    "kernel : {kernel_version} has numa=off in boot "
```
LiliDeng: f"{kernel_version}" ?

lisa/base_tools/uname.py
```diff
@@ -8,10 +8,34 @@
 from lisa.util import LisaException
 
 
+@dataclass
+class KernelVersion:
```
sharsonia: Did you check _parse_version() and VersionInfo in operating_system.py? I think it can give you what you are looking for here.

microsoft/testsuites/core/cpu.py
dsrivastavv: This is P1. I was using it for easier debugging but will modify it to 1 in the next commit.

lisa/tools/lscpu.py
dsrivastavv: There are two L1 cache : instruction and data. How about I rename it to `l1instruction`?

microsoft/testsuites/core/cpu.py
dsrivastavv: This is unsupported by hyperv : https://docs.microsoft.com/en-us/windows-server/virtualization/hyper-v/best-practices-for-running-linux-on-hyper-v#numa. I can move this link to metadata

lisa/base_tools/uname.py
dsrivastavv: I think I might be able to use VersionInfo class. I will give it a try.

microsoft/testsuites/core/cpu.py
dsrivastavv: @LiliDeng I didn't get you second comment. Can you please elaborate?

lisa/base_tools/uname.py
```diff
@@ -8,10 +8,34 @@
...
+
+    def __str__(self) -> str:
+        return self.kernel_version
+
+
 @dataclass
 class UnameResult:
     has_result: bool
-    kernel_version: str = ""
+    kernel_version: KernelVersion
```
squirrelsc: Let's keep both parsed structured version and original string.

lisa/base_tools/uname.py
```diff
@@ -8,10 +8,34 @@
 from lisa.util import LisaException
 
 
+@dataclass
```
squirrelsc: add this change as an independent commit

lisa/tools/lscpu.py
squirrelsc: l1_instruction and l1_data?

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,71 @@
...
+        Incorrect mapping:
+        CPU NODE SOCKET L1d L1i L2 L3
+        8   0    0      8   8   8  8
+        9   1    1      9   9   9  9
+        """,
+        priority=0,
+    )
+    def l3_cache_check(self, node: Node) -> None:
+        cmdline = node.tools[Cat].run("/proc/cmdline").stdout
+        if NUMA_OFF in cmdline:
```
squirrelsc: Add comments to explain, why it runs only when numa=off?

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,71 @@
...
+            ):
+                raise SkippedException(
+                    "kernel : {kernel_version} has numa=off in boot "
+                    "parameter and its kernel version is earlier than 2.6.37. "
+                    "No support for NUMA setting. https://t.ly/x8k3"
+                )
+
+        cpu_info = node.tools[Lscpu].get_cpu_info()
+        for cpu in cpu_info:
+            assert cpu.node == cpu.l3, (
```
squirrelsc: use assert_that. That will show both values, if it's failed. but the original assert does not.

microsoft/testsuites/core/cpu.py
dsrivastavv: The test is not run when NUMA if off. It's skipped. Did you mean to convey something else?

microsoft/testsuites/core/cpu.py
dsrivastavv: I added a custom error message to show both values. But I agree that assert_that would be a cleaner option.

lisa/base_tools/uname.py
dsrivastavv: `KernelVersion` class has original string and I have overridden `__str__` and `__repr__` methods that return original string. This ensures that the code works without making any change elsewhere. Any reason to keep them separate?

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,71 @@
...
+    def l3_cache_check(self, node: Node) -> None:
+        cmdline = node.tools[Cat].run("/proc/cmdline").stdout
+        if NUMA_OFF in cmdline:
+            kernel_version: KernelVersion = (
+                node.tools[Uname].get_linux_information().kernel_version
+            )
+            self.__log.debug(
+                "Found numa=off in /proc/cmdline. Checking the kernel version."
+            )
+            if (
```
squirrelsc: Well, it's not a complete comparison, so we need the version class. For example, 2.0.38 is not considered to be skipped.

microsoft/testsuites/core/cpu.py
squirrelsc: BTW, the l3 looks like 13... how about level_3, l3_cache or level_3_cache? full name is better.

microsoft/testsuites/core/cpu.py
squirrelsc: ok, I see the skipped exception message. It's clear enough.

lisa/base_tools/uname.py
squirrelsc: The raw string may be logged. And the structure is not able to dump the raw string, if we use the class of semver.

microsoft/testsuites/core/cpu.py
dsrivastavv: I will rename it to l3_cache and similar for other variables. That should be cleaner.

microsoft/testsuites/core/cpu.py
dsrivastavv: `VersionInfo` class suggested by Sonia should take of these scenarios.

lisa/base_tools/uname.py
```diff
@@ -3,15 +3,19 @@
...
 from lisa.util import LisaException
 
+DEFAULT_VERSION = VersionInfo(0, 0, 0)
+
 
 @dataclass
 class UnameResult:
     has_result: bool
     kernel_version: str = ""
+    kernel_version_parsed: VersionInfo = DEFAULT_VERSION
```
squirrelsc: 1. It doesn't need a default value. Move it up to kernel_version. All required arguments should be above optional.
2. The name should be `kernel_version`, and the str version should be like `raw_kernel_version` or something else. Because we encourage people to use structured version.

microsoft/testsuites/core/cpu.py
```diff
@@ -0,0 +1,70 @@
...
+
+from lisa import Logger, testsuite
+from lisa.base_tools.cat import Cat
+from lisa.base_tools.uname import Uname
+from lisa.node import Node
+from lisa.tools.lscpu import Lscpu
+from lisa.util import SkippedException
+
+NUMA_OFF = "numa=off"
+MAX_NUMA_UNSUPPORTED_KERNEL = VersionInfo(2, 6, 37)
```
squirrelsc: 1. It's used once, so it doesn't need to defined here.
2. The VersionInfo supports to compare with a string. So the comparison can be like `if version < "2.6.37":`

lisa/tools/lscpu.py
```diff
@@ -71,6 +91,29 @@ def get_cpu_type(self, force_run: bool = False) -> CpuType:
...
+        mappings = mappings_with_header[1:]
+        assert len(mappings) > 0
+        output: List[CPUInfo] = []
+        for item in mappings:
+            # Ignore first line as it contains column header
+            mapping = item.split()
+            cpu = int(mapping[0])
+            node = int(mapping[1])
+            socket = int(mapping[2])
+            cache = [int(c) for c in mapping[3].split(":")]
```
squirrelsc: It's better to use regular expression, instead of a couple of string operations. It's more straightforward. In our code base, there are many examples to reference.

lisa/base_tools/uname.py
dsrivastavv: Regarding 2, currently `kernel_version: str ` is used primarily for logging purpose in the codebase. Also, `VersionInfo` is capable of dumping the raw string in formatted logs. 

```
>>> ver = VersionInfo.parse("5.8.0-1039-azure")
>>> ver
VersionInfo(major=5, minor=8, patch=0, prerelease='1039-azure', build=None)
>>> f"{ver}"
'5.8.0-1039-azure'
```

The additional raw string variable will be unused. Wouldn't it be cleaner to leave raw string variable and work with `VersionInfo` class?

squirrelsc: Let's keep the raw string. I haven't checked how the `__str__` is implemented in `VersionInfo`. It's safe to maintain the raw string by us.

dsrivastavv: I can keep it if you are insisting. But the implementation seems okay to me : 

https://github.com/python-semver/python-semver/blob/e93b6def40194f374623ce5e524a3dcfa40f7682/src/semver/version.py#L486-L492

squirrelsc: It's not the raw string, it's a standard string. If some version are not following this pattern, and used other format, our code can parse it into VersionInfo, but cannot get it back. 

Check the VersionInfo parser on distro version. We may need to reuse the logic in kernel version, if we found some version  cannot be parsed by default patterns.

https://github.com/microsoft/lisa/blob/main/lisa/operating_system.py#L237

dsrivastavv: Agreed! It makes sense to reuse the logic from `operating_system.py`. How about we can move this logic into utils?

squirrelsc: Yes, it can be.

squirrelsc (approved):

LiliDeng (approved):

PR #1475, opened 2021-08-16
Migrate kdump test cases into lisav3

lubaihua33:
1. Add match_lines parameter and append function into sed tool
2. Add enable_service function into systemctl tool
3. Add sysctl tool into lisav3
4. Add get_vmbus_version function in dmesg tool
5. Add UnsupportedDistroException
6. Add tar tool and extract function
7. Add kdump tool and kdumpcrash test suite

### DIFF
--- a/lisa/base_tools/sed.py
+++ b/lisa/base_tools/sed.py
@@ -9,14 +9,43 @@ class Sed(Tool):
     def command(self) -> str:
         return "sed"
 
-    def replace(
-        self, searched: str, replaced: str, file: str, sudo: bool = False
+    def substitute(
+        self,
+        regexp: str,
+        replacement: str,
+        file: str,
+        match_lines: str = "",
+        sudo: bool = False,
     ) -> None:
         # always force run, make sure it happens every time.
-        searched = searched.replace('"', '\\"')
-        replaced = replaced.replace('"', '\\"')
+        match_lines = match_lines.replace('"', '\\"')
+        regexp = regexp.replace('"', '\\"')
+        replacement = replacement.replace('"', '\\"')
+        if match_lines != "":
+            cmd = f'-i.bak "/{match_lines}/s/{regexp}/{replacement}/g" {file}'
+        else:
+            cmd = f'-i.bak "s/{regexp}/{replacement}/g" {file}'
+
         result = self.run(
-            f'-i.bak "s/{searched}/{replaced}/g" {file}',
+            cmd,
+            force_run=True,
+            no_error_log=True,
+            no_info_log=True,
+            sudo=sudo,
+            shell=True,
+        )
+        result.assert_exit_code(message=result.stdout)
+
+    def append(
+        self,
+        text: str,
+        file: str,
+        sudo: bool = False,
+    ) -> None:
+        # always force run, make sure it happens every time.
+        text = text.replace('"', '\\"')
+        result = self.run(
+            f"-i.bak '$a{text}' {file}",
             force_run=True,
             no_error_log=True,
             no_info_log=True,
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -655,8 +655,11 @@ class Ubuntu(Debian):
     def _replace_default_entry(self, entry: str) -> None:
         self._log.debug(f"set boot entry to: {entry}")
         sed = self._node.tools[Sed]
-        sed.replace(
-            "GRUB_DEFAULT=.*", f"GRUB_DEFAULT='{entry}'", "/etc/default/grub", sudo=True
+        sed.substitute(
+            regexp="GRUB_DEFAULT=.*",
+            replacement=f"GRUB_DEFAULT='{entry}'",
+            file="/etc/default/grub",
+            sudo=True,
         )
 
         # output to log for troubleshooting
--- a/lisa/tools/dmesg.py
+++ b/lisa/tools/dmesg.py
@@ -4,6 +4,8 @@
 import re
 from typing import List
 
+from semver import VersionInfo
+
 from lisa.executable import Tool
 from lisa.util import LisaException
 from lisa.util.process import ExecutableResult
@@ -17,6 +19,12 @@ class Dmesg(Tool):
         re.compile("rcu_sched detected stalls on"),
         re.compile("BUG: soft lockup"),
     ]
+
+    # [   3.191822] hv_vmbus: Hyper-V Host Build:18362-10.0-3-0.3294; Vmbus version:3.0
+    # [   3.191822] hv_vmbus: Vmbus version:3.0
+    __vmbus_version_pattern = re.compile(
+        r"\[\s+\d+.\d+\]\s+hv_vmbus:.*Vmbus version:(?P<major>\d+).(?P<minor>\d+)"
+    )
 
     @property
     def command(self) -> str:
@@ -56,6 +64,23 @@ class Dmesg(Tool):
             else:
                 self._log.debug(error_message)
         return result
+
+    def get_vmbus_version(self) -> VersionInfo:
+        result = self._run()
+        result.assert_exit_code(
+            message=f"exit code should be zero, but actually {result.exit_code}"
+        )
+        raw_vmbus_version = re.finditer(self.__vmbus_version_pattern, result.stdout)
+        for vmbus_version in raw_vmbus_version:
+            matched_vmbus_version = self.__vmbus_version_pattern.match(
+                vmbus_version.group()
+            )
+            if matched_vmbus_version:
+                major = matched_vmbus_version.group("major")
+                minor = matched_vmbus_version.group("minor")
+                self._log.info(f"vmbus version is {major}.{minor}")
+                return VersionInfo(int(major), int(minor))
+        raise LisaException("No find matched vmbus version in dmesg")
 
     def _run(self, force_run: bool = False) -> ExecutableResult:
         # sometime it need sudo, we can retry
--- a/lisa/tools/kdump.py
+++ b/lisa/tools/kdump.py
@@ -1,0 +1,437 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from pathlib import PurePath, PurePosixPath
+from typing import TYPE_CHECKING, List, Type
+
+from retry import retry
+from semver import VersionInfo
+
+from lisa.base_tools import Cat, Sed, Wget
+from lisa.executable import Tool
+from lisa.operating_system import Debian, Posix, Redhat, Suse
+from lisa.tools import Gcc
+from lisa.tools.make import Make
+from lisa.tools.service import Service
+from lisa.tools.sysctl import Sysctl
+from lisa.tools.tar import Tar
+from lisa.util import LisaException, UnsupportedDistroException
+
+if TYPE_CHECKING:
+    from lisa.node import Node
+
+
+class Kexec(Tool):
+    """
+    kexec - directly boot into a new kernel
+    kexec is a system call that enables you to load and boot into another
+    kernel from the currently running kernel. The primary difference between
+    a standard system boot and a kexec boot is that the hardware initialization
+    normally performed by the BIOS or firmware (depending on architecture)
+    is not performed during a kexec boot. This has the effect of reducing the
+    time required for a reboot.
+
+    This tool is used for managing the installation of kexec.
+    """
+
+    # kexec-tools 2.0.16
+    __pattern_kexec_version_info = re.compile(
+        r"^kexec\S+\s+(?P<major>\d+).(?P<minor>\d+).(?P<patch>\d+)"
+    )
+
+    # Existed bug for kexec-tools 2.0.14
+    # https://bugs.launchpad.net/ubuntu/+source/kexec-tools/+bug/1713940
+    # If the version of kexec-tools is lower than 2.0.15, we install kexec from source
+    _target_kexec_version = "2.0.15"
+
+    # If install kexec from source, we choose 2.0.18 version for it is stable for most
+    # Debian distros
+    _kexec_repo = (
+        "https://mirrors.edge.kernel.org/pub/linux/utils/kernel/kexec/"
+        "kexec-tools-2.0.18.tar.gz"
+    )
+
+    @property
+    def command(self) -> str:
+        return "kexec"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Posix)
+        self.node.os.install_packages("kexec-tools")
+        if isinstance(self.node.os, Debian):
+            version = self._get_version()
+            if version < self._target_kexec_version:
+                self._install_from_src()
+        return self._check_exists()
+
+    def _get_version(self) -> VersionInfo:
+        result = self.run(
+            "-v",
+            force_run=False,
+            no_error_log=True,
+            no_info_log=True,
+            sudo=True,
+            shell=False,
+        )
+        result.assert_exit_code(message=result.stderr)
+        raw_version = re.finditer(self.__pattern_kexec_version_info, result.stdout)
+        for version in raw_version:
+            matched_version = self.__pattern_kexec_version_info.match(version.group())
+            if matched_version:
+                major = matched_version.group("major")
+                minor = matched_version.group("minor")
+                patch = matched_version.group("patch")
+                self._log.info(f"kexec version is {major}.{minor}.{patch}")
+                return VersionInfo(int(major), int(minor), int(patch))
+        raise LisaException("No find matched kexec version")
+
+    def _install_from_src(self) -> None:
+        tool_path = self.get_tool_path()
+        self.node.shell.mkdir(tool_path, exist_ok=True)
+        wget = self.node.tools[Wget]
+        kexec_tar = wget.get(self._kexec_repo, str(tool_path))
+        tar = self.node.tools[Tar]
+        kexec_source = tar.extract(kexec_tar, str(tool_path))
+        code_path = tool_path.joinpath(kexec_source)
+        self.node.tools[Gcc]
+        make = self.node.tools[Make]
+        self.node.execute("./configure", cwd=code_path).assert_exit_code()
+        make.make_install(cwd=code_path)
+        self.node.execute(
+            "yes | cp -f /usr/local/sbin/kexec /sbin/", sudo=True, shell=True
+        ).assert_exit_code()
+
+
+class Makedumpfile(Tool):
+    """
+    makedumpfile - make a small dumpfile of kdump
+    With kdump, the memory image of the first kernel can be taken as vmcore
+    while the second kernel is running. makedumpfile makes a small DUMPFILE by
+    compressing dump data or by excluding unnecessary pages for analysis, or both.
+
+    This tool is used for managing the installation of makedumpfile.
+    """
+
+    @property
+    def command(self) -> str:
+        return "makedumpfile"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Posix)
+        if isinstance(self.node.os, Redhat):
+            self.node.os.install_packages("kexec-tools")
+        else:
+            self.node.os.install_packages("makedumpfile")
+        return self._check_exists()
+
+
+class KdumpBase(Tool):
+    """
+    kdump is a feature of the Linux kernel that creates crash dumps in the event of a
+    kernel crash. When triggered, kdump exports a memory image (also known as vmcore)
+    that can be analyzed for the purposes of debugging and determining the cause of a
+    crash.
+
+    kdump tool manages the kdump feature of the Linux kernel. Different distro os has
+    different kdump tool.
+
+    KdumpBase is a basic class, it returns sub instance according to distros. We can
+    support Redhat, Suse, Debian family distro now.
+    """
+
+    # If the file /sys/kernel/kexec_crash_loaded does not exist. This means that the
+    # currently running kernel either was not configured to support kdump, or that a
+    # crashkernel= commandline parameter was not used when the currently running kernel
+    # booted. Value "1" means crash kernel is loaded, otherwise not loaded.
+    #
+    # It also has /sys/kernel/kexec_crash_size file, which record the crash kernel size
+    # of memory reserved. We don't need to check this file in our test case.
+    kexec_crash = "/sys/kernel/kexec_crash_loaded"
+
+    # This file shows you the current map of the system's memory for each physical
+    # device. We can check /proc/iomem file for memory reserved for crash kernel.
+    iomem = "/proc/iomem"
+
+    # Following are the configuration setting required for system and dump-capture
+    # kernels for enabling kdump support.
+    required_kernel_config = [
+        "CONFIG_KEXEC",
+        "CONFIG_CRASH_DUMP",
+        "CONFIG_PROC_VMCORE",
+    ]
+
+    dump_path = "/var/crash"
+
+    @classmethod
+    def create(cls, node: "Node") -> Tool:
+        if isinstance(node.os, Redhat):
+            return KdumpRedhat(node)
+        elif isinstance(node.os, Debian):
+            return KdumpDebian(node)
+        elif isinstance(node.os, Suse):
+            return KdumpSuse(node)
+        else:
+            raise UnsupportedDistroException(node.os.name, node.os.information.version)
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Kexec, Makedumpfile]
+
+    @property
+    def command(self) -> str:
+        raise NotImplementedError()
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        raise NotImplementedError()
+
+    def check_required_kernel_config(self, config_path: str) -> None:
+        for config in self.required_kernel_config:
+            result = self.node.execute(f"grep {config}=y {config_path}")
+            result.assert_exit_code(
+                message=f"The kernel config {config} is not set."
+                "Kdump is not supported."
+            )
+
+    def _get_crashkernel_cfg_file(self) -> str:
+        """
+        This method return the path of cfg file where we config crashkernel memory.
+        If distro has a different cfg file path, override it.
+        """
+        return "/etc/default/grub"
+
+    def _get_crashkernel_update_cmd(self, crashkernel: str) -> str:
+        """
+        After setting crashkernel into grub cfg file, need updating grub configuration.
+        This function returns the update command string. If distro has a different
+        command, override this method.
+        """
+        return "grub2-mkconfig -o /boot/grub2/grub.cfg"
+
+    def config_crashkernel_memory(
+        self,
+        crashkernel: str,
+    ) -> None:
+
+        # For Redhat 8 and later version, the cfg_file should be None.
+        cfg_file = self._get_crashkernel_cfg_file()
+        if cfg_file:
+            assert self.node.shell.exists(PurePosixPath(cfg_file)), (
+                f"{cfg_file} doesn't exist. Please check the right grub file for "
+                f"{self.node.os.name} {self.node.os.information.version}."
+            )
+            cat = self.node.tools[Cat]
+            sed = self.node.tools[Sed]
+            result = cat.run(cfg_file)
+            if "crashkernel" in result.stdout:
+                sed.substitute(
+                    match_lines="^GRUB_CMDLINE_LINUX",
+                    regexp='crashkernel=[^[:space:]"]*',
+                    replacement=f"crashkernel={crashkernel}",
+                    file=cfg_file,
+                    sudo=True,
+                )
+            else:
+                sed.substitute(
+                    match_lines="^GRUB_CMDLINE_LINUX",
+                    regexp='"$',
+                    replacement=f" crashkernel={crashkernel}",
+                    file=cfg_file,
+                    sudo=True,
+                )
+            # Check if crashkernel is insert in cfg file
+            result = cat.run(cfg_file, force_run=True)
+            if f"crashkernel={crashkernel}" not in result.stdout:
+                raise LisaException(
+                    f'No find "crasherkel={crashkernel}" in {cfg_file} after'
+                    "insert. Please double check the grub config file and insert"
+                    "process"
+                )
+
+        # Update grub
+        update_cmd = self._get_crashkernel_update_cmd(crashkernel)
+        result = self.node.execute(update_cmd, sudo=True)
+        result.assert_exit_code(message="Failed to update grub")
+
+    def config_dump_path(self) -> None:
+        """
+        If the system memory size is bigger than 1T, the default size of /var/crash
+        maybe not enough to store the dump file, need change the dump path. The distro
+        which doesn't have enough space for /var/crash, need override this method.
+        """
+        return
+
+    def enable_kdump_service(self) -> None:
+        """
+        This method enable the kdump service. If distro has a different kdump service
+        name, need override it.
+        """
+        service = self.node.tools[Service]
+        service.enable_service("kdump")
+
+    def set_unknown_nmi_panic(self) -> None:
+        """
+        /proc/sys/kernel/unknown_nmi_panic:
+        The value in this file affects behavior of handling NMI. When the value is
+        non-zero, unknown NMI is trapped and then panic occurs. If need to dump the
+        crash, the value should be set 1. Some architectures don't provide architected
+        NMIs,such as ARM64, the system doesn't have this file, we don't need to set
+        either.
+        """
+        nmi_panic_file = PurePath("/proc/sys/kernel/unknown_nmi_panic")
+        if self.node.shell.exists(nmi_panic_file):
+            sysctl = self.node.tools[Sysctl]
+            sysctl.write("kernel.unknown_nmi_panic", "1")
+
+    @retry(exceptions=LisaException, tries=60, delay=1)  # type: ignore
+    def _check_kexec_crash_loaded(self) -> None:
+        """
+        Sometimes it costs a while to load the value, so define this methed as @retry
+        """
+        cat = self.node.tools[Cat]
+        result = cat.run(self.kexec_crash, force_run=True)
+        if "1" != result.stdout:
+            raise LisaException(f"{self.kexec_crash} file's value is not 1.")
+
+    def _check_crashkernel_in_cmdline(self, crashkernel_memory: str) -> None:
+        cat = self.node.tools[Cat]
+        result = cat.run("/proc/cmdline", force_run=True)
+        if f"crashkernel={crashkernel_memory}" not in result.stdout:
+            raise LisaException(
+                f"crashkernel={crashkernel_memory} boot parameter is not present in"
+                "kernel cmdline"
+            )
+
+    def _check_crashkernel_memory_reserved(self) -> None:
+        cat = self.node.tools[Cat]
+        result = cat.run(self.iomem, force_run=True)
+        if "Crash kernel" not in result.stdout:
+            raise LisaException(
+                f"No find 'Crash kernel' in {self.iomem}. Memory isn't reserved for"
+                "crash kernel"
+            )
+
+    def check_crashkernel_loaded(self, crashkernel_memory: str) -> None:
+        # Check crashkernel parameter in cmdline
+        self._check_crashkernel_in_cmdline(crashkernel_memory)
+
+        # Check crash kernel loaded
+        if not self.node.shell.exists(PurePosixPath(self.kexec_crash)):
+            raise LisaException(
+                f"{self.kexec_crash} file doesn't exist. Kexec crash is not loaded."
+            )
+        self._check_kexec_crash_loaded()
+
+        # Check if memory is reserved for crash kernel
+        self._check_crashkernel_memory_reserved()
+
+    def check_vmcore_exist(self) -> None:
+        cmd = f"find {self.dump_path} -type f -size +10M"
+        result = self.node.execute(cmd, shell=True, sudo=True)
+        if result.stdout == "":
+            raise LisaException(
+                "No file was found in /var/crash of size greater than 10M."
+                "The dump file didn't generate, please double check."
+            )
+
+
+class KdumpRedhat(KdumpBase):
+    @property
+    def command(self) -> str:
+        return "kdumpctl"
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Redhat)
+        self.node.os.install_packages("kexec-tools")
+        return self._check_exists()
+
+    def _get_crashkernel_cfg_file(self) -> str:
+        if self.node.os.information.version >= "8.0.0-0":
+            # For Redhat 8 and later version, we can use grubby command to config
+            # crashkernel. No need to get the crashkernel cfg file
+            return ""
+        else:
+            return "/etc/default/grub"
+
+    def _get_crashkernel_update_cmd(self, crashkernel: str) -> str:
+        if self.node.os.information.version >= "8.0.0-0":
+            return f'grubby --update-kernel=ALL --args="crashkernel={crashkernel}"'
+        else:
+            if self.node.shell.exists(PurePosixPath("/sys/firmware/efi")):
+                # System with UEFI firmware
+                return "grub2-mkconfig -o /boot/efi/EFI/redhat/grub.cfg"
+            else:
+                # System with BIOS firmware
+                return "grub2-mkconfig -o /boot/grub2/grub.cfg"
+
+    def config_dump_path(self) -> None:
+        """
+        If the system memory size is bigger than 1T, the default size of /var/crash
+        maybe not enough to store the dump file, need change the dump path
+        """
+        kdump_conf = "/etc/kdump.conf"
+        check_memory_cmd = "free -h | grep Mem | awk '{print $2}'"
+        result = self.node.execute(check_memory_cmd, shell=True, sudo=True)
+        if "Ti" not in result.stdout:
+            # System memeory size is smaller than 1T, no need to change dump path
+            return
+        size = float(result.stdout.strip("Ti"))
+        if size > 1:
+            self.dump_path = "/mnt/crash"
+            self.node.execute(
+                f"mkdir -p {self.dump_path}", shell=True, sudo=True
+            ).assert_exit_code()
+            # Change dump path in kdump conf
+            sed = self.node.tools[Sed]
+            sed.substitute(
+                match_lines="^path",
+                regexp="path",
+                replacement="#path",
+                file=kdump_conf,
+                sudo=True,
+            )
+            sed.append(f"path {self.dump_path}", kdump_conf, sudo=True)
+
+
+class KdumpDebian(KdumpBase):
+    @property
+    def command(self) -> str:
+        return "kdump-config"
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Debian)
+        self.node.os.install_packages("kdump-tools")
+        return self._check_exists()
+
+    def _get_crashkernel_cfg_file(self) -> str:
+        return "/etc/default/grub.d/kdump-tools.cfg"
+
+    def _get_crashkernel_update_cmd(self, crashkernel: str) -> str:
+        return "update-grub"
+
+    def enable_kdump_service(self) -> None:
+        service = self.node.tools[Service]
+        service.enable_service("kdump-tools")
+
+
+class KdumpSuse(KdumpBase):
+    @property
+    def command(self) -> str:
+        return "kdumptool"
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Suse)
+        self.node.os.install_packages("kdump")
+        return self._check_exists()
--- a/lisa/tools/make.py
+++ b/lisa/tools/make.py
@@ -6,7 +6,8 @@ from typing import TYPE_CHECKING, cast
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
-from lisa.tools import Gcc, Lscpu
+from lisa.tools import Gcc
+from lisa.tools.lscpu import Lscpu
 
 if TYPE_CHECKING:
     from lisa.node import Node
--- a/lisa/tools/service.py
+++ b/lisa/tools/service.py
@@ -32,6 +32,12 @@ class Service(Tool):
     def stop_service(self, name: str) -> None:
         self._internal_tool.stop_service(name)  # type: ignore
 
+    def enable_service(self, name: str) -> None:
+        self._internal_tool.enable_service(name)  # type: ignore
+
+    def check_service_status(self, name: str) -> bool:
+        return self._internal_tool._check_service_running(name)  # type: ignore
+
 
 class ServiceInternal(Tool):
     @property
@@ -99,4 +105,8 @@ class Systemctl(Tool):
 
     def restart_service(self, name: str) -> None:
         cmd_result = self.run(f"restart {name}", shell=True, sudo=True, force_run=True)
+        cmd_result.assert_exit_code()
+
+    def enable_service(self, name: str) -> None:
+        cmd_result = self.run(f"enable {name}", shell=True, sudo=True, force_run=True)
         cmd_result.assert_exit_code()
--- a/lisa/util/__init__.py
+++ b/lisa/util/__init__.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import Any, Dict, Iterable, List, Optional, Pattern, Type, TypeVar
 
 import pluggy
+from semver import VersionInfo
 
 from lisa import secret
 from lisa.util import constants
@@ -49,6 +50,20 @@ class UnsupportedOperationException(Exception):
     """
 
     ...
+
+
+class UnsupportedDistroException(Exception):
+    """
+    This exception is used to indicate that a test case does not support the testing
+    distro.
+    """
+
+    def __init__(self, name: str, version: VersionInfo) -> None:
+        self.name = name
+        self.version = version
+
+    def __str__(self) -> str:
+        return f"Unsupported distro {self.name} {self.version}"
 
 
 class SkippedException(Exception):
--- a/microsoft/testsuites/kdump/kdumpcrash.py
+++ b/microsoft/testsuites/kdump/kdumpcrash.py
@@ -1,0 +1,300 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import time
+from pathlib import PurePosixPath
+from typing import cast
+
+from assertpy import assert_that
+
+from lisa import (
+    LisaException,
+    Logger,
+    Node,
+    RemoteNode,
+    SkippedException,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    UnsupportedDistroException,
+    node_requirement,
+    schema,
+    search_space,
+)
+from lisa.operating_system import Redhat
+from lisa.tools import Dmesg, Echo, KdumpBase, Lscpu, Uname
+from lisa.util.perf_timer import create_timer
+from lisa.util.shell import try_connect
+
+
+@TestSuiteMetadata(
+    area="kdump",
+    category="functional",
+    description="""
+    This test suite is used to verify if kernel crash dmup is effect, which is judged
+    through vmcore file is generated after triggering kdump by sysrq.
+
+    It has 7 test cases. They verfy if kdump is effect when:
+        1. VM has 1 cpu
+        2. VM has 2-8 cpus and trigger kdump on cpu 1
+        3. VM has 33-192 cpus and trigger kdump on cpu 32
+        4. VM has 193-415 cpus and trigger kdump on cpu 192
+        5. VM has more than 415 cpus and trigger kdump on cpu 415
+        6. crashkernel is set "auto"
+        7. crashkernel is set "auto" and VM has more than 2T memory
+    """,
+)
+class KdumpCrash(TestSuite):
+    # When with large system memory, the dump file can achieve more than 7G. It will
+    # cost about 10min to copy dump file to disk for some distros, such as Ubuntu.
+    # So we set the timeout time 800s to make sure the dump file is completed.
+    timeout_of_dump_crash = 800
+    trigger_kdump_cmd = "echo c > /proc/sysrq-trigger"
+    crash_kernel = "512M"
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfies if kdump is effect when VM has 1 cpu.
+        VM need 2G memory at least to make sure it has enough memory to load crash
+        kernel.
+
+        Steps:
+        1. Check if vmbus version and kernel configurations support for crash dump.
+        2. Specify the memory reserved for crash kernel in kernel cmdline, setting the
+            "crashkernel" option to the required value.
+            a. Modify the grub config file to add crashkrenel option or change the
+                value to the required one. (For Redhat 8, no need to modify grub config
+                file. It can specify crashkernel by using grubby command directly)
+            b. Update grub config
+        4. If needed, config the dump path.
+        3. Reboot system to make kdump effect.
+        4. Check if the crash kernel is loaded.
+            a. Check if kernel cmdline has crashkernel option and the value is expected
+            b. Check if /sys/kernel/kexec_crash_loaded file exists and the value is '1'
+            c. Check if /proc/iomem is reserved memory for crash kernel
+        5. Trigger kdump through 'echo c > /proc/sysrq-trigger', or trigger on
+            specified CPU by using command "taskset -c".
+        6. Check if vmcore is generated under the dump path we configure after system
+            boot up.
+        """,
+        priority=1,
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                core_count=1, memory_mb=search_space.IntRange(min=2048)
+            ),
+        ),
+    )
+    def kdumpcrash_validate_single_core(self, node: Node, log: Logger) -> None:
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfies if the kdump is effect when VM has 2~8 cpus, and
+        trigger kdump on the second cpu(cpu1), which is designed by a known issue.
+        The test steps are same as `kdumpcrash_validate_single_core`.
+        """,
+        priority=2,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=search_space.IntRange(min=2, max=8))
+        ),
+    )
+    def kdumpcrash_validate_smp(self, node: Node, log: Logger) -> None:
+        self.trigger_kdump_cmd = "taskset -c 1 echo c > /proc/sysrq-trigger"
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfies if the kdump is effect when VM has 33~192 cpus and
+        trigger kdump on the 33th cpu(cpu32), which is designed by a known issue.
+        The test steps are same as `kdumpcrash_validate_single_core`.
+        """,
+        priority=2,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=search_space.IntRange(min=33, max=192))
+        ),
+    )
+    def kdumpcrash_validate_on_cpu32(self, node: Node, log: Logger) -> None:
+        self._trigger_kdump_on_specified_cpu(32, node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfies if the kdump is effect when VM has 193~415 cpus, and
+        trigger kdump on the 193th cpu(cpu192), which is designed by a known issue.
+        The test steps are same as `kdumpcrash_validate_single_core`.
+        """,
+        priority=2,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=search_space.IntRange(min=193, max=415))
+        ),
+    )
+    def kdumpcrash_validate_on_cpu192(self, node: Node, log: Logger) -> None:
+        self._trigger_kdump_on_specified_cpu(192, node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfies if the kdump is effect when VM has more than 415 cpus,
+        and trigger kdump on the 416th cpu(cpu415), which is designed by a known issue.
+        The test steps are same as `kdumpcrash_validate_single_core`.
+        """,
+        priority=2,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=search_space.IntRange(min=416))
+        ),
+    )
+    def kdumpcrash_validate_on_cpu415(self, node: Node, log: Logger) -> None:
+        self._trigger_kdump_on_specified_cpu(415, node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfies if the kdump is effect when crashkernel is set auto.
+        The test steps are same as `kdumpcrash_validate_single_core`.
+        """,
+        priority=2,
+    )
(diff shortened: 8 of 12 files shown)


lisa/base_tools/sed.py
```diff
@@ -10,13 +10,19 @@ def command(self) -> str:
         return "sed"
 
     def replace(
-        self, searched: str, replaced: str, file: str, sudo: bool = False
+        self, searched: str, original: str, replaced: str, file: str, sudo: bool = False
```
squirrelsc: 1. As sed's standard names, this method should be called `substitute`.
2. What's the `searched` is used for? The standard substitute operation has no `searched` part.

lisa/operating_system.py
```diff
@@ -306,6 +306,9 @@ def name_pattern(cls) -> Pattern[str]:
     def replace_boot_kernel(self, kernel_version: str) -> None:
         raise NotImplementedError("update boot entry is not implemented")
 
+    def config_kenel_cmdline(self, original: str, replaced: str) -> None:
```
squirrelsc: 1. The method name should describe what task it configured, instead of the command it executed. For example, it can be like `set_boot_kernel`.
2. There is a method call `replace_boot_kernel`. Are they similar functionality? If so, let's merge them.

lisa/operating_system.py
```diff
@@ -757,6 +776,39 @@ def replace_boot_kernel(self, kernel_version: str) -> None:
...
+            cmdline_config_file = "/etc/default/grub"
+            sed = self._node.tools[Sed]
+            sed.replace(
+                "^GRUB_CMDLINE_LINUX_DEFAULT",
+                original,
+                replaced,
+                cmdline_config_file,
+                sudo=True,
+            )
+            result = self._node.execute("ls -lt /sys/firmware/efi")
```
squirrelsc: The generation is Azure specified concept, it shouldn't put in command os methods. Does `/boot/grub2/grub.cfg` exist in gen2? If not, we can run gen2 logic, if this file doesn't exist.

lisa/operating_system.py
```diff
@@ -757,6 +776,39 @@ def replace_boot_kernel(self, kernel_version: str) -> None:
...
+            if result.exit_code == 0:
+                # generation 2
+                cmd_result = self._node.execute(
+                    "grub2-mkconfig -o /boot/efi/EFI/redhat/grub.cfg",
```
squirrelsc: for this kind of logic, The `cmd_result = self._node.execute(` can be shared, only the command is different.

lisa/operating_system.py
```diff
@@ -757,6 +776,39 @@ def replace_boot_kernel(self, kernel_version: str) -> None:
...
+                    shell=True,
```
squirrelsc: If a command can run without shell, don't add it by default.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+from pathlib import PurePosixPath
+from lisa.executable import Tool
+from lisa.tools import Sed, Cat
+from lisa.operating_system import Posix, Redhat, Suse, Debian
+from typing import cast
+from assertpy import assert_that
+
+
+class Kdump(Tool):
+    # Redhat: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/kernel_administration_guide/kernel_crash_dump_guide
```
squirrelsc: Add one more line to explain what this variable is for. The urls don't explain themselves. Ideally, people should know brief without open the links.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+    @property
+    def command(self) -> str:
+        return "makedumpfile"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _check_exists(self) -> bool:
+        if isinstance(self.node.os, Redhat):
```
squirrelsc: Can it be checked by command exists or not? It's not needed to check the package.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+        return False
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Redhat):
+            self.node.os.install_packages("kexec-tools")
+        elif isinstance(self.node.os, Debian):
+            self.node.os.install_packages("linux-crashdump")
+        elif isinstance(self.node.os, Suse):
+            package_names = ["kexec-tools", "kdump", "makedumpfile"]
+            self.node.os.install_packages(package_names)
```
squirrelsc: How about else? If it doesn't know, it should raise an exception, instead of do nothing silently.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+        return self._check_exists()
+
+    def config_kdump(
```
squirrelsc: What means "config" here, what's configured for? Use an accurate method name.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+                sudo=True,
+            )
+            # Enable kdump
+            sed.replace(
+                searched="",
+                original="USE_KDUMP=0",
+                replaced="USE_KDUMP=1",
+                file="/etc/default/kdump-tools",
+                sudo=True,
+            )
```
squirrelsc, line 206: Add `else`

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+        self,
+        crashkernel_memory: str,
+    ) -> None:
+        # For Ubuntu, /etc/default/grub.d/kdump-tools.cfg which is in kdump-tools package,
+        # will override the crashkernel config defined as GRUB_CMDLINE_LINUX_DEFAULT in /etc/default/grub
+        # We can keep the default crashkernel config for Ubuntu distro, no need to set crashkernel
+        posix_os: Posix = cast(Posix, self.node.os)
+        if isinstance(self.node.os, Redhat):
+            posix_os.config_kenel_cmdline(
+                "crashkernel=auto", f"crashkernel={crashkernel_memory}"
```
squirrelsc: Make sure "crashkernel=auto" works as a default value. If not, it's better to use regexp to replace it every time.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+            )
+        elif isinstance(self.node.os, Suse):
+            posix_os.config_kenel_cmdline(
+                '\\"$', f' crashkernel={crashkernel_memory}\\"'
+            )
```
squirrelsc, line 234: miss `else`

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def config_nmi_panic(self) -> None:
+        nmi_panic_file = "/proc/sys/kernel/unknown_nmi_panic"
+        if self.node.shell.exists(PurePosixPath(nmi_panic_file)):
+            result = self.node.execute(
+                "sysctl -w kernel.unknown_nmi_panic=1", shell=True, sudo=True
+            )
+            result.assert_exit_code(
+                message="Failed to enable kernel to call panic when it receives a NMI."
+            )
```
squirrelsc: miss else. If it doesn't need, add comments.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+
+        kdump = node.tools[Kdump]
+        kdump.config_kdump(self.dump_path)
+        kdump.config_crashkernel_memory(self.crash_kernel)
+        kdump.enable_kdump_service()
+
+        # Reboot system to make kdump take effect
+        node.reboot()
+
+        # If /proc/sys/kernel/unknown_nmi_panic exist, config unknown_nmi_panic as 1
```
squirrelsc: The comments doesn't explain why it needs to set to 1

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+            result = cat.run(self.kexec_crash)
+            assert_that(result.stdout).described_as(
+                "kexec_crash_loaded is not 1. "
+                "Kexec crash is not loaded after reboot. "
+            ).is_equal_to("1")
+
+    def check_kdump_service(self) -> None:
+        systemctl = self.node.tools[Systemctl]
+        if isinstance(self.node.os, Redhat) or isinstance(self.node.os, Suse):
+            systemctl._check_service_running("kdump")
```
squirrelsc: The leading `_` means it's a private method. Either make this public, or use another method. Don't call private method out of a class.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+        if isinstance(self.node.os, Debian):
+            cat = self.node.tools[Cat]
+            result = cat.run(self.kexec_crash)
+            assert_that(result.stdout).described_as(
+                "kexec_crash_loaded is not 1. "
+                "Kexec crash is not loaded after reboot. "
+            ).is_equal_to("1")
+
+    def check_kdump_service(self) -> None:
+        systemctl = self.node.tools[Systemctl]
```
squirrelsc: It doesn't need to check the service. A method can called enable service. It enables the kdump service, it should be enough.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+
+class Kdump(Tool):
+    # Redhat: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/kernel_administration_guide/kernel_crash_dump_guide
+    # Ubuntu: https://ubuntu.com/server/docs/kernel-crash-dump
+    # Suse: https://www.suse.com/zh-cn/support/kb/doc/?id=000016171
+    kexec_crash = "/sys/kernel/kexec_crash_loaded"
+
+    @property
+    def command(self) -> str:
+        return "makedumpfile"
```
squirrelsc, line 122: This is not consistent with this tool. It may need two tools, Kdump to manage the service, and makedumpfile to create dump.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        result.assert_exit_code(message="crashkernel boot parameter is not present")
+        assert_that(
+            self.node.shell.exists(PurePosixPath(self.kexec_crash))
+        ).described_as(
+            "Kexec crash is not loaded after reboot. "
+            "Please check the configuration settings for kdump and grub."
+        ).is_true()
+
+        if isinstance(self.node.os, Debian):
```
squirrelsc: It looks a test assertion. It should be in test case, not the tool.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # Activate the magic SysRq option
+        cmd_result = node.execute(
+            "echo 1 > /proc/sys/kernel/sysrq",
+            shell=True,
+            sudo=True,
+        )
+
+        # Trigger kdump
+        cmd_result = node.execute(
```
squirrelsc: After it's triggered, how it's verified?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+    @TestCaseMetadata(
+        description="""
+        This test case will trigger kdump in the VM which has one core
+        """,
+        priority=0,
+        requirement=simple_requirement(node=schema.NodeSpace(core_count=2)),
+    )
+    def kdumpcrash_validate_single_core(self, node: Node) -> None:
+        self.crash_kernel = "512M"
+        self._kdump_test(node)
```
squirrelsc: The `_kdump_test` is called only once, why it needs in a method?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+            self.trigger_kdump_cmd,
+            shell=True,
+            sudo=True,
+        )
+
+    @TestCaseMetadata(
+        description="""
+        This test case will trigger kdump in the VM which has one core
+        """,
+        priority=0,
```
squirrelsc: Original V2 priority should plus 1. If it's p0 original, it should be p1 in v3.

lisa/tools/kdump.py
lubaihua33: Done. Changed into that kdump depends on kexec and makedumpfile.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,155 @@
...
+            )
+        elif isinstance(self.node.os, Suse):
+            posix_os.config_kenel_cmdline(
+                '\\"$', f' crashkernel={crashkernel_memory}\\"'
+            )
+
+    def enable_kdump_service(self) -> None:
+        # For Ubuntu, it has kdump-tools.service, when install kdump-tools package, it will enable by default
+        # For Redhat, it has kdump.service which is in kexec-tools package, it will enable by default when install the package?
+        systemctl = self.node.tools[Systemctl]
```
LiliDeng: invoke self.node.tools[Service], don't invoke Systemctl directly

lisa/tools/service.py
```diff
@@ -100,3 +100,7 @@ def stop_service(self, name: str) -> None:
     def restart_service(self, name: str) -> None:
         cmd_result = self.run(f"restart {name}", shell=True, sudo=True, force_run=True)
         cmd_result.assert_exit_code()
+
+    def enable_service(self, name: str) -> None:
```
LiliDeng, line 110: also add it into Service and ServiceInternal

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+        return config_path
+
+    def _kdump_test(self, node: Node) -> None:
+        dmesg = node.tools[Dmesg]
+        if self.vmbus_string not in dmesg.get_output():
+            raise SkippedException(
+                "Negotiated VMBus version is not 3.0. "
+                "Kernel might be old or patches not included. "
+                "Full support for kdump is not present."
+            )
```
LiliDeng, line 196: write a similar method _get_vmbus_version
in https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/platform_.py#L572?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+        cmd_result = node.execute(
+            f"grep CONFIG_KEXEC_AUTO_RESERVE=y {self.config_path}",
+            shell=True,
+            sudo=True,
+        )
+
+        if isinstance(node.os, Redhat) and node.os.information.version < "8.0.0":
+            if self.crash_kernel == "auto" and cmd_result.exit_code != 0:
+                raise SkippedException("crashkernel=auto doesn't work for this distro.")
+        if (
```
LiliDeng: put this if above line 54?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+            shell=True,
+            sudo=True,
+        )
+
+    @TestCaseMetadata(
+        description="""
+        This test case will trigger kdump in the VM which has one core
+        """,
+        priority=0,
+        requirement=simple_requirement(node=schema.NodeSpace(core_count=2)),
```
LiliDeng: 1 core or 2 core?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # If /proc/sys/kernel/unknown_nmi_panic exist, config unknown_nmi_panic as 1
+        kdump.config_nmi_panic()
+
+        # Confirm that the kernel dump mechanism is enabled
+        kdump.check_kdump_loaded()
+        kdump.check_kdump_service()
+
+        # Activate the magic SysRq option
+        cmd_result = node.execute(
```
LiliDeng: how about wrap it in kdump tool?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,109 @@
...
+            not isinstance(node.os, Redhat)
+            and not isinstance(node.os, Debian)
+            and not isinstance(node.os, Suse)
+        ):
+            raise SkippedException("Distro not supported. Skip the test.")
+
+        kdump = node.tools[Kdump]
+        kdump.config_kdump(self.dump_path)
+        kdump.config_crashkernel_memory(self.crash_kernel)
+        kdump.enable_kdump_service()
```
LiliDeng: usually either it exists in install method, or we'd better put it before config_kdump

lisa/tools/kdump.py
LiliDeng: my understanding is that we can remove all crashkernel value, and inject a new config, if we hard code the replaced value, then we have an assumption that 'crashkernel=auto' exists.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: What's the reason to skip other versions? Is there any document? If not, it should fail the case, and we can support to test it later.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: If use simple_requirement, not to use node, specify min_count, not exact 2.

lisa/operating_system.py
lubaihua33: But when without shell=True, this command return 1.

lisa/operating_system.py

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: This test case need 1 core, but we don't support 1 core now. 
https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/platform_.py#L91

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: I will add the other test cases later.

lisa/operating_system.py
squirrelsc: What error shows in stdout or stderr?

lisa/base_tools/sed.py
lubaihua33: Has changed `replace` into `substitute`.
The `searched` is used for matching the line, then we replace the string in `s//`.

lisa/operating_system.py
lubaihua33: Has changed. I deleted this function. It is not very suitable to put in the operating_system.py.

lisa/operating_system.py
lubaihua33: Move this into kdump.py

lisa/tools/kdump.py
lubaihua33: Add "raise an exception" in `command `function. In the install function, we don't need to add `else`

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Have changed.

lisa/tools/service.py
lubaihua33: Have added.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Added comments about the auto crashkernel. Changed as Lili's comment.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: I deleted this code. The unknown nmi trigger panic is a Hyper-V scenario. I will support this later.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Added in the new commit.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

lisa/base_tools/sed.py
squirrelsc: So the first parameter should be called as `match_lines` like manual. It should have default value `""`, because it's optional. And other parameters should be called `regexp` and `replacement`.

refer to https://www.gnu.org/software/sed/manual/sed.html#Regexp-Addresses

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def enable_kdump_service(self) -> None:
+        systemctl = self.node.tools[Systemctl]
+        if isinstance(self.node.os, Redhat) or isinstance(self.node.os, Suse):
+            systemctl.enable_service("kdump")
+        elif isinstance(self.node.os, Debian):
+            systemctl.enable_service("kdump-tools")
+
+    def set_unknown_nmi_panic(self) -> None:
+        nmi_panic_file = "/proc/sys/kernel/unknown_nmi_panic"
```
squirrelsc: Declare it as a PurePath, so it doesn't need to be converted in `exists` method.

lisa/tools/makedumpfile.py
```diff
@@ -0,0 +1,29 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Redhat):
+            self.node.os.install_packages("kexec-tools")
+        elif isinstance(self.node.os, Posix):
+            self.node.os.install_packages("makedumpfile")
```
squirrelsc: add `else` to handle other distros

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+    @property
+    def command(self) -> str:
+        if isinstance(self.node.os, Redhat):
+            return "kdumpctl"
+        elif isinstance(self.node.os, Debian):
+            return "kdump-config"
+        elif isinstance(self.node.os, Suse):
+            return "kdumptool"
+        else:
+            posix_os: Posix = cast(Posix, self.node.os)
```
squirrelsc: what's the reason need to cast to posix os? All os, including Windows has below fields.

lisa/tools/kdump.py
squirrelsc: It doesn't matter how other methods handle it. The logic among methods may be changed. The `else` is the guard sentence, the logic in each method should be completed.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+            self.node.os.install_packages("kexec-tools")
+        elif isinstance(self.node.os, Debian):
+            self.node.os.install_packages("linux-crashdump")
+        elif isinstance(self.node.os, Suse):
+            self.node.os.install_packages("kdump")
+
+        return self._check_exists()
+
+    def check_kernel_config(self, config_path: str) -> None:
+        for config in self.required_kernel_config:
```
squirrelsc, line 201: It's better to cat the config file, and search the strings in Python. It may be faster, because there is only one ssh command. And it dumps the raw content of the file, it's easier for troubleshooting.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+        if isinstance(self.node.os, Redhat):
+            self.node.os.install_packages("kexec-tools")
+        elif isinstance(self.node.os, Debian):
+            self.node.os.install_packages("linux-crashdump")
+        elif isinstance(self.node.os, Suse):
+            self.node.os.install_packages("kdump")
+
+        return self._check_exists()
+
+    def check_kernel_config(self, config_path: str) -> None:
```
squirrelsc: The method name should align with what's checked, like `check_required_kernel_config`

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+            cmd = "grub2-mkconfig -o /boot/grub2/grub.cfg"
+        elif isinstance(self.node.os, Debian):
+            kdump_cfg = "/etc/default/grub.d/kdump-tools.cfg"
+            if self.node.shell.exists(PurePosixPath(kdump_cfg)):
+                cfg_file = kdump_cfg
+            else:
+                cfg_file = "/etc/default/grub"
+            cmd = "update-grub"
+
+        if cfg_file:
```
squirrelsc, line 230: raise exception, if the file doesn't exists.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+            sed = self.node.tools[Sed]
+            result = self.node.execute(
+                f"grep crashkernel {cfg_file}", shell=True, sudo=True
```
squirrelsc: I think it doesn't need to use grep to check that. It can be done by a better regexp. If the line with `"GRUB_CMDLINE_LINUX_DEFAULT"`, it should works whether there is crashkernel or not.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+        if isinstance(self.node.os, Redhat) or isinstance(self.node.os, Suse):
+            systemctl.enable_service("kdump")
+        elif isinstance(self.node.os, Debian):
+            systemctl.enable_service("kdump-tools")
+
+    def set_unknown_nmi_panic(self) -> None:
+        nmi_panic_file = "/proc/sys/kernel/unknown_nmi_panic"
+        if self.node.shell.exists(PurePosixPath(nmi_panic_file)):
+            result = self.node.execute(
+                "sysctl -w kernel.unknown_nmi_panic=1", sudo=True
```
squirrelsc: add a tool for sysctl, it should be used frequently.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: The logic is still here. It's ok, but it needs to add comments to explain the logic.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,206 @@
...
+        # Update grub
+        result = self.node.execute(cmd, shell=True, sudo=True)
+        result.assert_exit_code(message="Failed to update grub")
+
+    def enable_kdump_service(self) -> None:
+        systemctl = self.node.tools[Systemctl]
+        if isinstance(self.node.os, Redhat) or isinstance(self.node.os, Suse):
+            systemctl.enable_service("kdump")
+        elif isinstance(self.node.os, Debian):
+            systemctl.enable_service("kdump-tools")
```
squirrelsc: Add else

lisa/tools/kexec.py
```diff
@@ -0,0 +1,30 @@
...
+    @property
+    def command(self) -> str:
+        return "kexec"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Posix):
```
squirrelsc: assert isinstance is simpler.

lisa/tools/kdump.py
lubaihua33: The kernel config file has about 6400 lines. When in debug mode, there will be many lines.

lisa/tools/kdump.py
lubaihua33: I'm not very clear.  Where I can add else?

lisa/tools/kdump.py
lubaihua33: If the cfg_file is "", that means the os is redhat 8, don't need to change any grub file, just use 'grubby --update-kernel=ALL --args="crashkernel=512M' to add kernel cmdline.

I added code to judge the file if exists, and raise an exception if not.

lisa/tools/kdump.py
lubaihua33: Done

lisa/operating_system.py
lubaihua33: Such as we run command "dmesg | grep 'Vmbus version:'"  If we don't use shell=True, the result is like this:
`
cmd:['dmesg', '|', 'grep', 'Vmbus version:']
elapsed:0.7511551999999995
exit_code:1
stderr:''
stdout:"dmesg: bad usage\r\nTry 'dmesg --help' for more information."
`

squirrelsc: That's a command with pipeline, `|`, it needs shell. But this command doesn't have  `|` or `<`, so it doesn't need shell=true.

lisa/tools/kdump.py
squirrelsc: It's on old version code. Make sure every `if` is reviewed carefully. Some conditions don't need `else`, but most of them need to add `else` as a guard sentence.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: There are two questions,
1. `core_count` means exact 2 core. It should use `min_core_count`, it means it's ok to run with more than `min_core_count`.
2. The platform setting doesn't block vm size with 1 core. It's just not in priority list. see last `.*`.

lisa/tools/kdump.py
squirrelsc: Ok, we can use grep here.

lisa/tools/kdump.py
squirrelsc: If so, it needs to check if it's redhat 8, and add comments to explain it.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: It may be not appropriate to wrap it in kdump tool. SysRq is a degub method to trigger kdump. I think it's better in test cases file.

lisa/tools/kdump.py
lubaihua33: Done

lisa/operating_system.py
lubaihua33: Thank you. I got it.  "grub2-mkconfig -o /boot/efi/EFI/redhat/grub.cfg" this command don't need shell=True. Have delete "shell=True".

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Seems that they are private functions. We couldn't use it in test cases.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: For Suse, it has kexec, makedumpfile, kdump command, but the kdump service is disable by default. If we put enable service in install method, the kdump service is not enabled for the install won't be called.

lisa/tools/makedumpfile.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Has changed.

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: The logic to get vmbus version from dmesg, it should be implemented in Dmesg tool. Avoid to overuse grep in code. It hides details, hard to troubleshooting failures.

lisa/base_tools/sed.py
squirrelsc: It's not fully fixed. Don't resolve comments, I will check and resolve them.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def check_required_kernel_config(self, config_path: str) -> None:
+        for config in self.required_kernel_config:
+            result = self.node.execute(f"grep {config}=y {config_path}")
+            result.assert_exit_code(
+                message=f"The kernel config {config} is not set."
+                "Kdump is not supported."
+            )
+
+    def _get_crashkernel_cfg_file(self) -> None:
```
squirrelsc: the "get" method should return value. The name and behavior is not consistent.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,326 @@
...
+        if isinstance(self.node.os, Redhat):
```
squirrelsc: The logic is complex unnecessary, not easy to read, try below,

```python
if isinstance(self.node.os, Debian):
    self.crashkernel_cfg_file = "/etc/default/grub.d/kdump-tools.cfg"
elif isinstance(self.node.os, Redhat) and self.node.os.information.version >= "8.0.0"
    # add comments to explain why it's emtpy here.
    self.crashkernel_cfg_file = ""
else:
    self.crashkernel_cfg_file = "/etc/default/grub"

if self.crashkernel_cfg_file:
    assert self.node.shell.exists(PurePosixPath(self.crashkernel_cfg_file)), (
                    f"{self.crashkernel_cfg_file} doesn't exist. "
                    "Please check the right grub file for "
                    f"{self.node.os.name} {self.node.os.information.version}."
             )

```

lisa/tools/kdump.py
```diff
@@ -0,0 +1,326 @@
...
+    # Following are the configuration setting required for system and dump-capture
+    # kernels for enabling kdump support.
+    required_kernel_config = [
+        "CONFIG_KEXEC",
+        "CONFIG_CRASH_DUMP",
+        "CONFIG_PROC_VMCORE",
+    ]
+
+    dump_path = "/var/crash"
+    crashkernel_cfg_file = ""
```
squirrelsc: It's used in only one method, it doesn't need to be a private member in the class.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,326 @@
...
+from typing import Dict, List, Type, cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Debian, OperatingSystem, Posix, Redhat, Suse
+from lisa.tools import Cat, Sed
+from lisa.tools.service import Service
+from lisa.util import LisaException
+
+
+class KernelDumpTool:
```
squirrelsc: This class makes the dependencies and logic complex. It doesn't need such a tool to wrap them together. For each tool, it can check which command name should be in its logic.

lisa/tools/kdump.py
lubaihua33: Have added.

lisa/tools/kexec.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Added get_vmbus_version function in dmesg tool.

lisa/tools/kdump.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Have deleted.

lisa/tools/kdump.py
lubaihua33: Done

lisa/base_tools/sed.py
lubaihua33: Have changed as your comments.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def enable_kdump_service(self) -> None:
+        service = self.node.tools[Service]
+        if isinstance(self.node.os, Redhat) or isinstance(self.node.os, Suse):
+            service.enable_service("kdump")
+        elif isinstance(self.node.os, Debian):
+            service.enable_service("kdump-tools")
+        else:
+            raise UnsupportedDistroException(
+                "Unsupported distro "
```
squirrelsc: The exception can have a `__init__` method, and pass os into it. So the error message can be handled inside of the Exception.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,295 @@
...
+        crashkernel: str,
+    ) -> None:
+
+        cfg_file = self._get_crashkernel_cfg_file()
+        update_cmd = self._get_crashkernel_update_cmd(crashkernel)
+        cat = self.node.tools[Cat]
+        sed = self.node.tools[Sed]
+
+        # For Redhat 8 and later version, the cfg_file should be "".
+        if cfg_file != "":
```
squirrelsc: `if cfg_file:`

lisa/tools/kdump.py
```diff
@@ -0,0 +1,295 @@
...
+                    match_lines="^GRUB_CMDLINE_LINUX",
+                    regexp='"$',
+                    replacement=f" crashkernel={crashkernel}",
+                    file=cfg_file,
+                    sudo=True,
+                )
+            # Check if crashkernel is insert in cfg file
+            result = cat.run(cfg_file, force_run=True)
+            if f"crashkernel={crashkernel}" not in result.stdout:
+                raise LisaException(f'No "crashkernel={crashkernel}" in {cfg_file}')
```
squirrelsc: Explain more here. What does it mean, if no crashkernel found in the file?

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: Again, the logic is still here, it's only moved a place. The key point is to explain the logic. Why it needs to set this flag, when this file exists. And what happens, if the file doesn't exist.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+        3. the crashkernel is set auto.
+    """,
+)
+class KdumpCrash(TestSuite):
+    crash_kernel = "512M"
+    activate_sysrq_cmd = "echo 1 > /proc/sys/kernel/sysrq"
+    trigger_kdump_cmd = "echo c > /proc/sysrq-trigger"
+    timeout_of_dump_crash = 60
+
+    def _get_boot_config_path(self, node: Node) -> str:
```
squirrelsc, line 177: Put private method below public methods.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=1, memory_mb=3584),
+        ),
+    )
+    def kdumpcrash_validate_single_core(self, node: Node, log: Logger) -> None:
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has two vcpu
```
squirrelsc: It should specify the steps refer to test suite specification.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+    )
+    def kdumpcrash_validate_16_cores(self, node: Node, log: Logger) -> None:
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has 96 vcpu and triggers
+        crash on one of them.
+        """,
+        priority=1,
```
squirrelsc: Not put all case at P1. The V2 priority should plus 1. For example, P1 in v2, it should be P1+1=P2 in v3.

lisa/tools/dmesg.py
```diff
@@ -18,6 +18,10 @@ class Dmesg(Tool):
         re.compile("BUG: soft lockup"),
     ]
 
+    __vmbus_vesion_pattern = re.compile(
```
squirrelsc: Add examples above the regexp patterns.

lisa/tools/dmesg.py
```diff
@@ -57,6 +61,21 @@ def check_kernel_errors(
                 self._log.debug(error_message)
         return result
 
+    def get_vmbus_version(self) -> str:
+        output = self._run()
+        output.assert_exit_code(
+            message=f"exit code should be zero, but actually {output.exit_code}"
+        )
+        for line in output.stdout.splitlines(keepends=False):
```
squirrelsc: It's low efficient to split and match line by line. Apply the regexp to the stdout is faster.

lisa/tools/dmesg.py
```diff
@@ -57,6 +61,21 @@ def check_kernel_errors(
...
+            matched_vmbus_version = self.__vmbus_vesion_pattern.match(line)
+            if matched_vmbus_version:
+                major = matched_vmbus_version.group("major")
+                minor = matched_vmbus_version.group("minor")
+                self._log.info(f"vmbus version is {major}.{minor}")
+                return major
```
squirrelsc: What's the reason to split it to major and minor? Can it use the whole string? And it can reuse the VersionInfo as distro and kernel version.

lisa/tools/dmesg.py
```diff
@@ -57,6 +61,21 @@ def check_kernel_errors(
...
+        self._log.info("No vmbus version in dmesg")
```
squirrelsc: What's the reason it has no vmbus version? It should raise Exception.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    UnsupportedDistroException,
+    node_requirement,
+    schema,
+)
+from lisa.operating_system import Redhat
+from lisa.tools import Dmesg, Lscpu, Uname
+from lisa.tools.kdump import Kdump
```
squirrelsc: Add it to `lisa/tools/__init__.py`, so it can be used from `lisa.tools` namespace directly.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+            description="No file was found in /var/crash of size greater than 10M",
+        ).is_equal_to(True)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has one vcpu
+        """,
+        priority=1,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=1, memory_mb=3584),
```
squirrelsc: Why it needs to specify a magic number of memory?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has 4 vcpu and triggers
+        crash on one of them.
+        """,
+        priority=1,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=4),
```
squirrelsc: What's reason it needs exact 4 cpu cores? how about 8 cpu cores?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+        self.trigger_kdump_cmd = f"taskset -c {random_cpu} echo c > /proc/sysrq-trigger"
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has 16 vcpu
+        """,
+        priority=1,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=16),
```
squirrelsc: What's the reason it needs to verify with 16 cores? What's the test purpose?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has 96 vcpu and triggers
+        crash on one of them.
+        """,
+        priority=1,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=96),
```
squirrelsc: So, what's the reason to verify with 96 cores?

lisa/tools/kdump.py
```diff
@@ -0,0 +1,295 @@
...
+            raise UnsupportedDistroException(
+                "Unsupported distro "
+                f"{self.node.os.name} {self.node.os.information.version}"
+            )
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
```
squirrelsc, line 197: Another way to avoid so many if/elif/else, is to override the `create` method, and create three tool class for different distros.

For example,
```python
class KdumpRedhat(KdumpBase):
    @property
    def command(self) -> str:
        return "kdumpctl"
    ....

class KdumpBase(Tool):
    @classmethod
    def create(cls, node: Node) -> Tool:
        if isinstance(os, Redhat):
            return KdumpRedhat(node)
        ....
```

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+
+    It has 6 test cases in this test suite. They verfy if the kdump is effect when:
+        1. with different vcpu count.
+        2. having crash on different vcpu.
+        3. the crashkernel is set auto.
+    """,
+)
+class KdumpCrash(TestSuite):
+    crash_kernel = "512M"
+    activate_sysrq_cmd = "echo 1 > /proc/sys/kernel/sysrq"
```
squirrelsc: The Echo tool can be used to `write_to_file` now.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+            self._check_supported(node)
+        except UnsupportedDistroException as identifier:
+            raise SkippedException(identifier)
+
+        kdump = node.tools[Kdump]
+        kdump.config_crashkernel_memory(self.crash_kernel)
+        kdump.enable_kdump_service()
+
+        node.execute(
+            f"mkdir -p {kdump.dump_path} && rm -rf {kdump.dump_path}/*",
```
squirrelsc, line 231: The logic looks strange, create a folder and delete it. It can just run `f"rm -rf {kdump.dump_path}", and ignore the result.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,231 @@
...
+        priority=1,
+        requirement=node_requirement(
+            node=schema.NodeSpace(core_count=96),
+        ),
+    )
+    def kdumpcrash_validate_large_cores(self, node: Node, log: Logger) -> None:
+        lscpu = node.tools[Lscpu]
+        core_count = lscpu.get_core_count()
+        random_cpu = randint(0, core_count - 1)
+        self.trigger_kdump_cmd = f"taskset -c {random_cpu} echo c > /proc/sysrq-trigger"
```
squirrelsc: What's the reason, the core 96 needs extra logic? Explain it by comments and descriptions.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,295 @@
...
+        return True
+
+    def is_vmcore_exist(self) -> bool:
+        if isinstance(self.node.os, Debian):
+            cmd = f"find {self.dump_path}/2* -type f -size +10M"
+        else:
+            cmd = f"find {self.dump_path}/*/vmcore -type f -size +10M"
+        result = self.node.execute(cmd, shell=True, sudo=True)
+        if result.exit_code != 0:
+            self._log.error(
```
squirrelsc: Raise Exception directly. The error message doesn't easy to take action.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,295 @@
...
+            )
+            return False
+
+        # Sometimes it cost a while to load the value.
+        self.check_kexec_crash_loaded()
+
+        # Check crash kernel memory reserved
+        result = cat.run(self.iomem, force_run=True)
+        if "Crash kernel" not in result.stdout:
+            self._log.error(
```
squirrelsc: Raise Exception, instead of writing error message. It doesn't help.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,295 @@
...
+        result = cat.run(self.kexec_crash, force_run=True)
+        if "1" != result.stdout:
+            raise LisaException(f"{self.kexec_crash} file's value is not 1.")
+
+    def is_crashkernel_loaded(self, crashkernel_memory: str) -> bool:
+        # Check crashkernel parameter in cmdline
+        cat = self.node.tools[Cat]
+        result = cat.run("/proc/cmdline", force_run=True)
+        if f"crashkernel={crashkernel_memory}" not in result.stdout:
+            self._log.error(
```
squirrelsc: raise exception.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: It's rm the content in /var/crash, not rm the whole folder. This is used to clean up the previous dump files.

squirrelsc: ok, thanks for explaining. The logic is ok to me.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Added comments here. Changed memory_mb as 2048. Need have enough space to reserve for crash kernel.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: The test case is not useful. Delete this one.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: This test suite wants to verify if the kdump is effect when with different count vcpus. Selected 16 through experience.

lubaihua33: Changed to min_core_count=16

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: This test suite wants to verify if the kdump is effect when with large count vcpus. Selected 96 through experience. Changed into min_core_count=96

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Delete the "taskset -c" test cases which are in lisav2. With or without "taskset -c" doesn't have a difference. If we just use "echo c > /proc/sysrq-trigger", the kdump CPU is also random.

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

lisa/tools/dmesg.py
lubaihua33: Done

lisa/tools/dmesg.py
lubaihua33: Reused the VersionInfo

lisa/tools/dmesg.py
lubaihua33: Done

lisa/tools/dmesg.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/tools/kdump.py
lubaihua33: Done

lisa/util/__init__.py
```diff
@@ -51,6 +52,20 @@ class UnsupportedOperationException(Exception):
     ...
 
 
+class UnsupportedDistroException(Exception):
+    """
+    This exception is used to indicate that a test case does not support the testing
+    distro.
+    """
+
+    def __init__(self, os_name: str, os_version: VersionInfo) -> None:
```
squirrelsc: `name` and `version` are enough. It's also inconsistent between `distro` and `os`.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,263 @@
...
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when crashkernel is set auto and
+        the memory is more than 2T. With the crashkernel=auto parameter, system will
+        reserved a suitable size memory for crash kernel. We want to see if the
+        crashkernel=auto can also handle this scenario when the system memory is large.
+
+        Please refer to test suite specification to see the steps.
```
squirrelsc: Let's write it in a test case, not test suite. The test suite are not recorded in notifier, so it won't be in our database.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,263 @@
...
+    )
+    def kdumpcrash_validate_single_core(self, node: Node, log: Logger) -> None:
+        self._kdump_test(node, log)
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has more than one vcpu.
+        Please refer to test suite specification to see the steps.
+        """,
+        priority=1,
```
squirrelsc: One normal p1 case is enough, this one can be p2.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,263 @@
...
+    """,
+)
+class KdumpCrash(TestSuite):
+    crash_kernel = "512M"
+    timeout_of_dump_crash = 60
+
+    @TestCaseMetadata(
+        description="""
+        This test case verfiy if the kdump is effect when VM has one vcpu.
+        The VM need 2G memory to make sure there are enough memory for crash kernel
```
squirrelsc: What happens, if the memory is not enough?

lisa/tools/kdump.py
```diff
@@ -0,0 +1,432 @@
...
+        service.enable_service("kdump-tools")
+
+
+class KdumpSuse(KdumpBase):
+    @property
+    def command(self) -> str:
+        return "kdumptool"
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Debian)
```
squirrelsc: assert Suse is clearer.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,263 @@
...
+        # Activate the magic SysRq option
+        echo = node.tools[Echo]
+        echo.write_to_file("1", "/proc/sys/kernel/sysrq", sudo=True)
+        node.execute("sync", shell=True, sudo=True)
+
+        try:
+            # Trigger kdump. After execute the trigger cmd, the VM will be disconnected.
+            # We set a timeout time 10.
+            node.execute(
+                "echo c > /proc/sysrq-trigger",
```
squirrelsc: use `echo.write_to_file`

squirrelsc: add timeout in write_to_file, if it doesn't support.

lisa/base_tools/sed.py
```diff
@@ -9,14 +9,43 @@ class Sed(Tool):
...
+            cmd,
+            force_run=True,
+            no_error_log=True,
+            no_info_log=True,
+            sudo=sudo,
+            shell=True,
+        )
+        result.assert_exit_code(message=result.stdout)
+
+    def insert_at_file_end(
```
squirrelsc: `append` is better.

lisa/tools/make.py
```diff
@@ -6,7 +6,8 @@
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
-from lisa.tools import Gcc, Lscpu
+from lisa.tools import Gcc
+from lisa.tools.lscpu import Lscpu
```
squirrelsc, line 10: What's the reason to split the import to two?

lisa/tools/kdump.py
```diff
@@ -0,0 +1,432 @@
...
+from lisa.tools.service import Service
+from lisa.tools.sysctl import Sysctl
+from lisa.tools.tar import Tar
+from lisa.util import LisaException, UnsupportedDistroException
+
+if TYPE_CHECKING:
+    from lisa.node import Node
+
+
+class Kexec(Tool):
```
squirrelsc, line 25: Where the tool is used? I only see it's referenced by KDump.

lisa/tools/kdump.py
```diff
@@ -0,0 +1,432 @@
...
+        self.node.tools[Gcc]
+        make = self.node.tools[Make]
+        self.node.execute("./configure", cwd=code_path).assert_exit_code()
+        make.make_install(cwd=code_path)
+        self.node.execute(
+            "yes | cp -f /usr/local/sbin/kexec /sbin/", sudo=True, shell=True
+        ).assert_exit_code()
+
+
+class Makedumpfile(Tool):
```
squirrelsc, line 110: Where is it used, rather than referenced as a dependency in Kdump?

lisa/tools/kdump.py
```diff
@@ -0,0 +1,432 @@
...
+                return VersionInfo(int(major), int(minor), int(patch))
+        raise LisaException("No find matched kexec version")
+
+    def _install_from_src(self) -> None:
+        tool_path = self.get_tool_path()
+        self.node.shell.mkdir(tool_path, exist_ok=True)
+        wget = self.node.tools[Wget]
+        kexec_tar = wget.get(self._kexec_repo, str(tool_path))
+        tar = self.node.tools[Tar]
+        kexec_source = tar.extract(f"{kexec_tar}", str(tool_path))
```
squirrelsc: `f"{kexec_tar}"` => `kexec_tar`

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -0,0 +1,263 @@
...
+            log=log,
+            timeout=self.timeout_of_dump_crash,
+        )
+        if not is_ready:
+            raise LisaException(
+                f"Cannot connect to [{remote_node.public_address}:"
+                f"{remote_node.public_port}] after trigger kdump."
+            )
+
+        node.close()
```
squirrelsc, line 281: Add comments to explain the close and initialize.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: It may be failed to load crash kernel.

lisa/tools/make.py
lubaihua33: After import Make in kdump.py, here have a circle import. Changing like this, no circle import error.

lisa/tools/kdump.py
lubaihua33: Yeah, it's referenced by Kdump.

lisa/tools/kdump.py
lubaihua33: It is only referenced as a dependency in Kdump.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: Ok, please specify it's "for load the crash kernel", it's clearer than "for crash kernel"

lisa/base_tools/sed.py
```diff
@@ -9,14 +9,43 @@ class Sed(Tool):
...
+            cmd,
+            force_run=True,
+            no_error_log=True,
+            no_info_log=True,
+            sudo=sudo,
+            shell=True,
+        )
+        result.assert_exit_code(message=result.stdout)
+
+    def append_at_file_end(
```
squirrelsc: `append` is enough. It implicitly says append to end of file.

lisa/tools/kdump.py
squirrelsc: If so, it means only the installation is necessary. Is it right? It doesn't need to be a tool. You can merge their installation steps. The shared steps can be in KdumpBase.

lisa/tools/kdump.py
squirrelsc: The same as kexec

lisa/tools/kdump.py
lubaihua33: They are better in kexec tool and makedumpfile tool. If we have installed kdump tool, that doesn't mean we have already installed kexec and makedumpfile tool.

lisa/base_tools/sed.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Done

lisa/tools/dmesg.py
```diff
@@ -57,6 +65,23 @@ def check_kernel_errors(
...
+        for vmbus_version in raw_vmbus_version:
+            matched_vmbus_version = self.__vmbus_version_pattern.match(
+                vmbus_version.group()
+            )
+            if matched_vmbus_version:
+                major = matched_vmbus_version.group("major")
+                minor = matched_vmbus_version.group("minor")
+                self._log.info(f"vmbus version is {major}.{minor}")
+                return VersionInfo(int(major), int(minor))
+        raise LisaException("No find matched vmbus version in dmesg")
```
LiliDeng, line 83: please double check in the old distro and freebsd distro

squirrelsc (approved):

PR #1478, opened 2021-08-17
[LISAv3] Create new log file notifier

isaiahduncan:
The overall log may be too long to open in azure portal. Creating log notifier to dump text formatted results for easier view in editing mode.

### DIFF
--- a/lisa/notifiers/text_result.py
+++ b/lisa/notifiers/text_result.py
@@ -1,0 +1,49 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import Path
+from typing import Any, List, Type
+
+from lisa import notifier, schema
+from lisa.runner import print_results
+from lisa.testsuite import TestResultMessage
+from lisa.util import LisaException, constants
+
+
+class TextResult(notifier.Notifier):
+    """
+    Creating log notifier to dump text formatted results for easier
+    view in editing mode. The original log is complete but too long to
+    check only the summary.
+    """
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "text_result"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return schema.Notifier
+
+    def _received_message(self, message: notifier.MessageBase) -> None:
+        if isinstance(message, TestResultMessage):
+            if message.is_completed:
+                self.received_messages.append(message)
+        else:
+            raise LisaException("Received unsubscribed message type")
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.result_path = Path(
+            f"{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt"
+        )
+        if self.result_path.exists():
+            raise LisaException("File already exists")
+
+        self.received_messages: List[TestResultMessage] = []
+
+    def finalize(self) -> None:
+        with open(self.result_path, "w") as result_file:
+            print_results(self.received_messages, result_file.write)
--- a/lisa/runner.py
+++ b/lisa/runner.py
@@ -11,8 +11,8 @@ from lisa import notifier, schema, transformer
 from lisa.action import Action
 from lisa.combinator import Combinator
 from lisa.parameter_parser.runbook import RunbookBuilder
-from lisa.testsuite import TestResult, TestStatus
-from lisa.util import BaseClassMixin, InitializableMixin, constants
+from lisa.testsuite import TestResult, TestResultMessage, TestStatus
+from lisa.util import BaseClassMixin, InitializableMixin, LisaException, constants
 from lisa.util.logger import create_file_handler, get_logger, remove_handler
 from lisa.util.parallel import TaskManager, cancel, set_global_task_manager
 from lisa.util.subclasses import Factory
@@ -31,6 +31,39 @@ def parse_testcase_filters(raw_filters: List[Any]) -> List[schema.BaseTestCaseFi
     else:
         filters = [schema.TestCase(name="test", criteria=schema.Criteria(area="demo"))]
     return filters
+
+
+def print_results(
+    test_results: List[Any],
+    output_method: Callable[[str], Any],
+) -> None:
+    output_method("________________________________________")
+    result_count_dict: Dict[TestStatus, int] = {}
+    for test_result in test_results:
+        if isinstance(test_result, TestResult):
+            result_name = test_result.runtime_data.metadata.full_name
+            result_status = test_result.status
+        elif isinstance(test_result, TestResultMessage):
+            result_name = test_result.name
+            result_status = test_result.status
+        else:
+            raise LisaException(f"Unknown result type: '{type(test_result)}'")
+        output_method(
+            f"{result_name:>50}: {result_status.name:<8} {test_result.message}"
+        )
+        result_count = result_count_dict.get(result_status, 0)
+        result_count += 1
+        result_count_dict[result_status] = result_count
+
+    output_method("test result summary")
+    output_method(f"    TOTAL    : {len(test_results)}")
+    for key in result_count_dict.keys():
+        count = result_count_dict.get(key, 0)
+        if key == TestStatus.ATTEMPTED and count == 0:
+            # attempted is confusing if user don't know it.
+            # so hide it if there is no attempted cases.
+            continue
+        output_method(f"    {key.name:<9}: {count}")
 
 
 class BaseRunner(BaseClassMixin, InitializableMixin):
@@ -130,7 +163,7 @@ class RootRunner(Action):
             for runner in self._runners:
                 runner.close()
 
-        self._output_results(self._results)
+        print_results(self._results, self._log.info)
 
         # pass failed count to exit code
         self.exit_code = sum(1 for x in self._results if x.status == TestStatus.FAILED)
@@ -206,28 +239,6 @@ class RootRunner(Action):
             runner.initialize()
             self._runners.append(runner)
             yield runner
-
-    def _output_results(self, test_results: List[TestResult]) -> None:
-        self._log.info("________________________________________")
-        result_count_dict: Dict[TestStatus, int] = {}
-        for test_result in test_results:
-            self._log.info(
-                f"{test_result.runtime_data.metadata.full_name:>50}: "
-                f"{test_result.status.name:<8} {test_result.message}"
-            )
-            result_count = result_count_dict.get(test_result.status, 0)
-            result_count += 1
-            result_count_dict[test_result.status] = result_count
-
-        self._log.info("test result summary")
-        self._log.info(f"    TOTAL    : {len(test_results)}")
-        for key in TestStatus:
-            count = result_count_dict.get(key, 0)
-            if key == TestStatus.ATTEMPTED and count == 0:
-                # attempted is confusing, if user don't know it.
-                # so hide it, if there is no attempted cases.
-                continue
-            self._log.info(f"    {key.name:<9}: {count}")
 
     def _callback_completed(self, results: List[TestResult]) -> None:
         self._results_lock.acquire()


lisa/notifiers/log_file.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+        dump_file(self.log_file_path, message)
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage, notifier.TestRunMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        runbook = cast(LogFileSchema, self.runbook)
+        self._log_level = runbook.log_level
+        self.log_file_path = environment.log_path / "log_file.txt"
```
isaiahduncan: Im not sure how to get the environment for this line. Should I add it as an argument ?

lisa/notifiers/log_file.py
```diff
@@ -0,0 +1,51 @@
...
+from lisa.util import constants, dump_file
+
+
+@dataclass_json()
+@dataclass
+class LogFileSchema(schema.TypedSchema):
+    log_level: str = logging.getLevelName(logging.DEBUG)
+
+
+class LogFile(notifier.Notifier):
```
squirrelsc: The notifier name is confusing, there is log file already. It's better to `TextResult`.

lisa/notifiers/log_file.py
```diff
@@ -0,0 +1,51 @@
...
+    """
+    It's a sample notifier, output subscribed message to a log file.
```
squirrelsc: It needs to specify the reason that we create this notifier. It's similar with the PR description.

lisa/notifiers/log_file.py
```diff
@@ -0,0 +1,51 @@
...
+
+from lisa import notifier, schema
+from lisa.environment import Environment
+from lisa.testsuite import TestResultMessage
+from lisa.util import constants, dump_file
+
+
+@dataclass_json()
+@dataclass
+class LogFileSchema(schema.TypedSchema):
```
squirrelsc: It looks no argument needed.

lisa/notifiers/log_file.py
```diff
@@ -0,0 +1,51 @@
...
+
+    @classmethod
+    def type_name(cls) -> str:
+        return constants.NOTIFIER_LOG_FILE
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return LogFileSchema
+
+    def _received_message(self, message: notifier.MessageBase) -> None:
```
squirrelsc: Refer to below logic, it should output the same format. Make the code shareable.

https://github.com/microsoft/lisa/blob/main/lisa/runner.py#L210

lisa/notifiers/log_file.py
```diff
@@ -0,0 +1,51 @@
...
+        print("who do we code")
```
squirrelsc: As the above comment, we should dump only the final status, not all status. Refer to`TestResult.is_completed` to implement it in TestResultMessage.

lisa/notifiers/log_file.py
```diff
@@ -0,0 +1,51 @@
...
+        self._log.log(
+            getattr(logging, self._log_level),
+            f"log file received message [{message.type}]: {message}",
+        )
+        dump_file(self.log_file_path, message)
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage, notifier.TestRunMessage]
```
squirrelsc: We need only TestResultMessage, not TestRunMessage.

lisa/notifiers/log_file.py
isaiahduncan: @squirrelsc I made this class a subclass of the TypedSchema class because that's how it was for the corresponding schema for the Console notifier. Should I make this mirror the Schema in console.py or should I not make this class extend anything?

squirrelsc: This notifier doesn't need the runbook settings. So the whole class can be removed. In the notifier, `type_schema` can be overridden like,

```python
@classmethod
    return schema.Notifier
```

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,44 @@
...
+                )
+             
+        
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        runbook = cast(schema.Notifier, self.runbook)
+        self._log_level = runbook.log_level
+        # self.log_file_path = environment.log_path / "log_file.txt"
```
isaiahduncan: Line 44 will be removed before merge

lisa/util/constants.py
```diff
@@ -47,6 +47,7 @@
 # notifier
 NOTIFIER = "notifier"
 NOTIFIER_CONSOLE = "console"
+NOTIFIER_LOG_FILE = "log_file"
```
squirrelsc: Change this name too.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,44 @@
...
+                f"{message.status.name:<8} {message.message}",
+                )
+             
+        
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        runbook = cast(schema.Notifier, self.runbook)
+        self._log_level = runbook.log_level
```
squirrelsc: It doesn't need log_level.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,44 @@
...
+    def type_name(cls) -> str:
+        return constants.NOTIFIER_LOG_FILE
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return schema.Notifier
+
+    def _received_message(self, message: TestResultMessage) -> None:
+        if message.is_completed:
+            dump_file(
```
squirrelsc: It's better to hold the handle of the file, instead of open/close it frequently. It can be closed in finalize.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,44 @@
...
+                message.environment.log_path / "log_file.txt",
```
squirrelsc: The log result should be in the run folder, not related to an environment. Refer to current root log names, this name should be like `lisa-20210818-004408-883-result.txt`. The path can be got like `"{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt")`

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return schema.Notifier
+
+    def _received_message(self, message: TestResultMessage) -> None:
+        if message.is_completed:
+            dump_file(
+                message.environment.log_path / "log_file.txt", 
+                f"{message.status.name:<8} {message.message}",
```
squirrelsc: rather than log for each test case, also log total. So, it needs to reuse the same method, move it to be a method in runner, so it can be reused.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,44 @@
...
+
+from lisa import notifier, schema
+from lisa.testsuite import TestResultMessage
+from lisa.util import constants, dump_file
+
+
+class TextResult(notifier.Notifier):
+    """
+    Creating log notifier to dump text formatted results for easier 
+    view in editing mode.
```
squirrelsc: The "editing mode" is lack of context here. We need to translate it to be in general. It can explain the original log is completed, but too long to check only the summary.

lisa/notifiers/text_result.py
isaiahduncan: @squirrelsc From this, I think that you want me to put a method in runner.py for this printing. If my understanding is correct, wouldn't it make more sense to leave the method inside of this class and make a private member to keep track of how many logs? If my understanding isn't correct, could you please explain further?

squirrelsc: @isaiahduncan good question. Let me explain it from design principles of LISA. The notifier is isolate, it shouldn't depends on states of LISA. It knows all things from the notification messages. The method in runner.py is a pattern of functional programming, it has no side effect, and no state required. So both runner and notifier can use this method, if pass in the test results, and we can also pass in a stream to get the output. If we use the class of runner, it brings a couple of problems, for example, the notifier may need pass in the argument of stream, but it doesn't need by runner itself. Feel free to let me know, if there are more questions.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,45 @@
...
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        result_path = Path("{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt")
+        if result_path.exists:
+            raise Exception
+        self.result_file = open(result_path, "w")
+
+    def finalize(self) -> None:
+        self.result_file.close()
```
squirrelsc: It's better to add an empty line at the end of file. There is a setting in vscode to help it.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,45 @@
...
+            self.result_file.write(f"{message.status.name:<8} {message.message}")
+
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        result_path = Path("{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt")
+        if result_path.exists:
+            raise Exception
```
squirrelsc: 1. If it needs to raise an exception, add details messages. So when people see the exception, they know what happens.
2. Raise LisaException. It's easy to know, the exception is raised by LISA code, not underlying.
3. As some practice, it doesn't need to check this, if use `open(result_path, "x")`. It means if this file exists, an error will be raised.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,45 @@
...
+from lisa import notifier, schema
+from lisa.testsuite import TestResultMessage
+from lisa.util import constants, dump_file
+
+
+class TextResult(notifier.Notifier):
+    """
+    Creating log notifier to dump text formatted results for easier
+    view in editing mode. The riginal log is complete but too long to
+    check only the summary
```
squirrelsc: There are some typo (riginal), and lack of dot at the end of sentence.

lisa/notifiers/text_result.py
isaiahduncan: Could you please rephrase the last sentence? The one that says that begins with "If we use the class of runner, it brings a couple of problems"

squirrelsc: If we use a method in the runner class, it brings a couple of problems. Because the runner has a couple of variables (states). The method depends on those states, but the notifier doesn't.

isaiahduncan: So moving forward with this, should the following changes be made?
1) Create a method inside of runner.py that will print to the test results to the file
2) Keep track of the amount of times a message was given to the notifier within the notifier itself with a private member
3) Keep track of all messages (either as a message type or as the string that we create out of them) sent to the notifier with a private member
4) Print to the file after the final message has been sent to the notifier (possibly in the finalize method) using our new printing method in runner.py

squirrelsc: In my mind, this method should bases on current `_output_results` method, like,

```python
def print_results(test_results: List[Union[TestResults, TestResultMessage]], output_method: Callable[..., None])->None:
    # print all results
    # print summary
```

The output_method can be `self._log.info`, if the runner call it. And it can be `self.result_file.write`, when the notifier call it.

The notifier can save all completed messages together, and send to the method in finalize method.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,40 @@
...
+        if message.is_completed:
+            self.result_file.write(f"{message.status.name:<8} {message.message}")
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.result_file = open(
+            "{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt",
+            "w"
```
squirrelsc: The "w" needs to be "x", so when file exists, the line can raise error.

lisa/runner.py
```diff
@@ -33,6 +33,19 @@ def parse_testcase_filters(raw_filters: List[Any]) -> List[schema.BaseTestCaseFi
...
+    test_results: List[Union[TestResult, TestResultMessage]],
+    output_method: Callable[[str], None]
+    ) -> None:
+    for index, result in enumerate(test_results):
+        # Printing for TestResultMessage type.
+        if hasattr(result, "name"):
+            output_method(f"[{index}] {result.status.name:<8} {result.message}")
+        # Printing for TestResult type
+        else:
+            output_method(f"[{index}] {result.id_} {result.message}")
```
isaiahduncan: @squirrelsc I wasn't sure how to how to print for the TestResult type so this is what I put for now. Please let me know how you want the output to look for TestResult

lisa/runner.py
```diff
@@ -33,6 +33,19 @@ def parse_testcase_filters(raw_filters: List[Any]) -> List[schema.BaseTestCaseFi
     return filters
 
 
+def print_results(
+    test_results: List[Union[TestResult, TestResultMessage]],
+    output_method: Callable[[str], None],
+    ) -> None:
+    for index, result in enumerate(test_results):
+        # Printing for TestResultMessage type.
+        if hasattr(result, "name"):
```
squirrelsc: You can use `isinstance` to check the type, like below. It should handle both TestResult and TestResultMessage well. And max the shared logic.

```python
output_method.info("________________________________________")
result_count_dict: Dict[TestStatus, int] = {}
for test_result in test_results:
    if isinstance(test_result, TestResult):
        result_name = test_result.runtime_data.metadata.full_name
        result_status = test_result.status
    elif isinstance(test_result, TestResultMessage):
        result_name = test_result.name
        result_status = test_result.status
    else:
        raise LisaException(f"Unknown result type: '{type(test_result)}'")
    output_method(
        f"{result_status.full_name:>50}: "
        f"{result_status.name:<8} {test_result.message}"
    )
    result_count = result_count_dict.get(result_status, 0)
    result_count += 1
    result_count_dict[result_status] = result_count
output_method.info("test result summary")
output_method.info(f"    TOTAL    : {len(test_results)}")
for key in TestStatus:
    count = result_count_dict.get(key, 0)
    if key == TestStatus.ATTEMPTED and count == 0:
        # attempted is confusing, if user don't know it.
        # so hide it, if there is no attempted cases.
        continue
    output_method.info(f"    {key.name:<9}: {count}")
```

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.result_file = open(
+            "{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt",
+            "x",
+        )
+        self.received_messages: List[Union[TestResult, TestResultMessage]] = []
+
+    def finalize(self) -> None:
+        print_results(self.received_messages, self.result_file.write)
```
squirrelsc: Because the file is write only in finalize method. So it can be opened in finalize now. use the `with` pattern to open it.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,43 @@
...
+    def type_name(cls) -> str:
+        return constants.NOTIFIER_TEXT_RESULT
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return schema.Notifier
+
+    def _received_message(self, message: TestResultMessage) -> None:
+        if message.is_completed:
+            self.received_messages.append(Union(None, message))
```
dsrivastavv: Raise an error in case the message is different from the ones subscribed : 
```
 if isinstance(message, notifier.TestResultMessage):
  ......    
  else:
      raise LisaException("Received unsubscribed message type")
```

lisa/runner.py
```diff
@@ -33,6 +33,19 @@ def parse_testcase_filters(raw_filters: List[Any]) -> List[schema.BaseTestCaseFi
     return filters
 
 
+def print_results(
```
dsrivastavv, line 36: Can we move this method to notifier itself?

lisa/util/constants.py
```diff
@@ -47,6 +47,7 @@
 # notifier
 NOTIFIER = "notifier"
 NOTIFIER_CONSOLE = "console"
+NOTIFIER_TEXT_RESULT = "text_result"
```
dsrivastavv: We can move this to `type_name`. If a class needs this constant, they can call `type_name` method directly.

lisa/runner.py
squirrelsc: This method should be used by runner too, so it's here.

lisa/util/constants.py
isaiahduncan: @somil55 I noticed that the html notifier type_name handles this constant similarly to what you want me to do for my notifier. Why are we doing this differently for text_result and html instead of keeping these constants in constants.py like how it is for notifier and console notifier?

lisa/runner.py
```diff
@@ -33,6 +33,39 @@ def parse_testcase_filters(raw_filters: List[Any]) -> List[schema.BaseTestCaseFi
...
+    test_results: List[Union[TestResult, TestResultMessage]],
```
squirrelsc: Good to see it! so the original method in Runner can be removed, this method can be shared by runner and notifier.

lisa/util/constants.py
dsrivastavv: I am not sure. My understanding was that class names are uniquely mentioned in `type_name` method. @squirrelsc do you know the answer to this?

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.result_path = """
+        {constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt
+        """
+        result_file = open(self.result_path, "x")
```
squirrelsc: What's the reason to have open/close here? Because the logic is changed, if it needs to check if the file exists here, just check it.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def _received_message(self, message: TestResultMessage) -> None:
+        if message.is_completed:
+            self.received_messages.append(Union(None, message))
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.result_path = """
```
squirrelsc: what's the reason to use `"""`? Try f string, `self.result_path = f"{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt"`

isaiahduncan: @squirrelsc I did that because I was getting problems with the CI workflow because the line was too long

squirrelsc: Please follow the instruction to setup dev environment in vs code. There is a tool can auto format your code, after that, the error will be fixed. If the auto-format doesn't happen on saving the file, you can press "alt+shift+f".

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,48 @@
...
+            raise LisaException("Received unsubscribed message type")
+
+    def _subscribed_message_type(self) -> List[Type[notifier.MessageBase]]:
+        return [TestResultMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.received_messages: List[Union[TestResult, TestResultMessage]] = []
+
+    def finalize(self) -> None:
+        result_path = Path(
```
squirrelsc: This check should happen in initialize, not finalize. Let the file exists error happens earlier, instead of at end of run.

lisa/notifiers/text_result.py
```diff
@@ -0,0 +1,48 @@
...
+            "{constants.RUN_LOCAL_PATH}/lisa-{constants.RUN_ID}-result.txt"
+        )
+        if result_path.is_file:
```
squirrelsc: `result_path.exists()` is clearer.

squirrelsc (dismissed):

squirrelsc (approved):

PR #1480, opened 2021-08-17
Update WALA-VERIFY-HOSTNAME-CHANGE.py

mcgov:
Increase timeout on host name change (due to recent test failures caused by a too-short timeout)


Testscripts/Linux/WALA-VERIFY-HOSTNAME-CHANGE.py
```diff
@@ -66,7 +66,7 @@ def ChangeHostName(expectedHost):
     Run("hostname " + changed_hostname)
     fail_error_warn_count = Run("tail -f waagent.log | grep -E 'fail|error|warning' | wc -l")
     RunLog.info("Start to sleep for 120 seconds.")
```
LiliDeng, line 68: log also need change? probably we can have a loop, to check whether the name change within timeout.

mcgov:
squashed new changes into single commit for PR, updated so the code uses a loop rather than a single long timeout

mcgov:
@LiliDeng
changes review for review please 😁

LiliDeng (approved):

PR #1482, opened 2021-08-19
Add CPUVerifyOnline test

dsrivastavv:
Migrate Test : https://github.com/microsoft/lisa/blob/master/Testscripts/Linux/CORE-CPUVerifyOnline.sh

### DIFF
--- a/lisa/base_tools/cat.py
+++ b/lisa/base_tools/cat.py
@@ -11,3 +11,14 @@ class Cat(Tool):
 
     def _check_exists(self) -> bool:
         return True
+
+    def read_from_file(
+        self,
+        file: str,
+        force_run: bool = False,
+        sudo: bool = False,
+    ) -> str:
+        # Run `cat <file>`
+        result = self.run(file, force_run=force_run, sudo=sudo)
+        result.assert_exit_code(message=f"Error : {result.stdout}")
+        return result.stdout
--- a/lisa/tools/echo.py
+++ b/lisa/tools/echo.py
@@ -3,6 +3,8 @@
 
 from typing import Optional, Type
 
+from assertpy.assertpy import assert_that
+
 from lisa.executable import Tool
 
 
@@ -17,6 +19,22 @@ class Echo(Tool):
 
     def _check_exists(self) -> bool:
         return True
+
+    def write_to_file(
+        self,
+        value: str,
+        file: str,
+        shell: bool = False,
+        sudo: bool = False,
+    ) -> None:
+        # Run `echo <value> > <file>`
+        result = self.run(
+            f"{value} > {file}",
+            force_run=True,
+            shell=shell,
+            sudo=sudo,
+        ).stdout
+        assert_that(result).does_not_contain("Permission denied")
 
 
 class WindowsEcho(Echo):
--- a/microsoft/testsuites/core/cpu.py
+++ b/microsoft/testsuites/core/cpu.py
@@ -1,14 +1,23 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from pathlib import PurePosixPath
+
 from assertpy.assertpy import assert_that
 
 from lisa import Logger, testsuite
 from lisa.base_tools.cat import Cat
 from lisa.base_tools.uname import Uname
 from lisa.node import Node
+from lisa.tools.echo import Echo
 from lisa.tools.lscpu import Lscpu
-from lisa.util import SkippedException
+from lisa.tools.lsvmbus import Lsvmbus
+from lisa.util import BadEnvironmentStateException, SkippedException
+
+
+class CPUState:
+    OFFLINE: str = "0"
+    ONLINE: str = "1"
 
 
 @testsuite.TestSuiteMetadata(
@@ -19,6 +28,61 @@ from lisa.util import SkippedException
     """,
 )
 class CPU(testsuite.TestSuite):
+    def _set_cpu_state(self, cpu_id: str, state: str, node: Node) -> bool:
+        file_path = f"/sys/devices/system/cpu/cpu{cpu_id}/online"
+        file_exists = node.shell.exists(PurePosixPath(file_path))
+        if file_exists:
+            node.tools[Echo].write_to_file(state, file_path, shell=True, sudo=True)
+        result = node.tools[Cat].read_from_file(file_path, force_run=True, sudo=True)
+        return result == state
+
+    @testsuite.TestCaseMetadata(
+        description="""
+            This test will check that CPU assigned to lsvmbus
+            channels cannot be put offline.
+            Steps :
+            1. Get the list of lsvmbus channel cpu mappings using
+            command `lsvmbus -vv`.
+            2. Create a set of cpu's assigned to lsvmbus channels.
+            3. Try to put cpu offline by running
+            `echo 0 > /sys/devices/system/cpu/cpu/<cpu_id>/online`.
+            Note : We skip cpu 0 as it handles system interrupts.
+            4. Ensure that cpu is still online by checking state '1' in
+            `/sys/devices/system/cpu/cpu/<target_cpu>/online`.
+            """,
+        priority=2,
+    )
+    def cpu_verify_vmbus_force_online(self, node: Node, log: Logger) -> None:
+        cpu_count = node.tools[Lscpu].get_core_count()
+        log.debug(f"{cpu_count} CPU cores detected...")
+
+        channels = node.tools[Lsvmbus].get_device_channels_from_lsvmbus()
+        mapped_cpu = set()
+        for channel in channels:
+            for channel_vp_map in channel.channel_vp_map:
+                target_cpu = channel_vp_map.target_cpu
+                if target_cpu != "0":
+                    mapped_cpu.add(target_cpu)
+
+        for target_cpu in mapped_cpu:
+            log.debug(f"Checking CPU {target_cpu} on /sys/device/....")
+            result = self._set_cpu_state(target_cpu, CPUState.OFFLINE, node)
+            if result:
+                # Try to bring CPU back to it's original state
+                reset = self._set_cpu_state(target_cpu, CPUState.ONLINE, node)
+                exception_message = (
+                    f"Expected CPU {target_cpu} state : {CPUState.ONLINE}(online), "
+                    f"actual state : {CPUState.OFFLINE}(offline). CPU's mapped to "
+                    f"LSVMBUS channels shouldn't be in state "
+                    f"{CPUState.OFFLINE}(offline)."
+                )
+                if not reset:
+                    raise BadEnvironmentStateException(
+                        exception_message,
+                        f"The test failed leaving CPU {target_cpu} in a bad state.",
+                    )
+                raise AssertionError(exception_message)
+
     @testsuite.TestCaseMetadata(
         description="""
         This test case will check that L3 cache is correctly mapped
--- a/lisa/testsuite.py
+++ b/lisa/testsuite.py
@@ -30,6 +30,7 @@ from lisa.environment import EnvironmentSpace, EnvironmentStatus
 from lisa.feature import Feature
 from lisa.operating_system import OperatingSystem, Windows
 from lisa.util import (
+    BadEnvironmentStateException,
     LisaException,
     PassedException,
     QueuedException,
@@ -152,6 +153,10 @@ class TestResult:
             log.debug("case passed with warning", exc_info=exception)
             # case can be passed with a warning.
             self.set_status(TestStatus.PASSED, f"{phase}warning: {exception}")
+        elif isinstance(exception, BadEnvironmentStateException):
+            # TODO : Recycle or delete environment
+            log.error("case failed with environment in bad state", exc_info=exception)
+            self.set_status(TestStatus.FAILED, f"{phase}{exception}")
         else:
             if self.runtime_data.ignore_failure:
                 log.info(f"case failed and ignored: {exception}")
--- a/lisa/tools/lsvmbus.py
+++ b/lisa/tools/lsvmbus.py
@@ -96,6 +96,15 @@ class VmBusDevice:
 
         self.channel_vp_map = channel_vp_map_list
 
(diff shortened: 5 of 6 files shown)


lisa/base_tools/cat.py
```diff
@@ -11,3 +11,7 @@ def command(self) -> str:
 
     def _check_exists(self) -> bool:
         return True
+
+    def read_from_file(self, file_path: str) -> str:
```
squirrelsc: Add `sudo` argument.

lisa/util/process.py
```diff
@@ -90,13 +91,18 @@ def start(
         if shell:
             if not self._is_posix:
                 split_command = ["cmd", "/c", command]
+            elif super_user:
+                split_command = ["sudo", "su", "-c", command]
```
squirrelsc: What's different with `sudo`?

lisa/base_tools/cat.py
squirrelsc: And it needs to support force_rerun as an argument, default should be False.

lisa/tools/echo.py
```diff
@@ -18,6 +18,19 @@ def _windows_tool(cls) -> Optional[Type[Tool]]:
...
+    def write_to_file(
+        self,
+        val: str,
+        file_path: str,
+        sudo: bool = False,
+        super_user: bool = False,
+    ) -> None:
+        # Run `echo <val> > <file_path>`
+        output = self.run(
+            f"{val} > {file_path}", sudo=sudo, super_user=super_user
```
squirrelsc: because it may need to rewrite multiple times, it needs to force rerun every time.

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,47 @@
...
+
+        for target_cpu in mapped_cpu:
+            log.debug(f"Checking CPU {target_cpu} on /sys/device/....")
+            file_path = f"/sys/devices/system/cpu/cpu{target_cpu}/online"
+            file_exists = node.tools[Test].file_exists(file_path)
+            if file_exists:
+                node.tools[Echo].write_to_file(
+                    CPUState.OFFLINE, file_path, super_user=True
+                )
+                result = node.tools[Cat].read_from_file(file_path)
```
squirrelsc: Is there any other way to check it? It looks output to 0, and check if it's 0 already... Can we know if it affects in another way?

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,47 @@
...
+    )
+    def cpu_verify_online(self, node: Node, log: Logger) -> None:
+        cpu_count = node.tools[Lscpu].get_core_count()
+        log.debug(f"{cpu_count} CPU cores detected...")
+        channels = node.tools[Lsvmbus].get_device_channels_from_lsvmbus()
+        mapped_cpu = set()
+        for channel in channels:
+            for channel_vp_map in channel.channel_vp_map:
+                target_cpu = channel_vp_map.target_cpu
+                if target_cpu != "0":
```
squirrelsc, line 64: Add comments to explain it. But it doesn't need this loop, if we just want to get id of each CPU. It can be a loop like `for i in range(1, cpu_count)`

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,47 @@
...
+            Steps :
+            1. Get the list of lsvmbus channel cpu mappings using
+            command `lsvmbus -vv`.
+            2. Create a set of cpu's assigned to lsvmbus channels.
+            3. Try to put cpu offline by running
+            `echo 0 > /sys/devices/system/cpu/cpu/<cpu_id>/online`.
+            Note : We skip cpu 0 as it handles system interrupts.
+            4. Ensure that cpu is still online by checking state '1' in
+            `/sys/devices/system/cpu/cpu/<target_cpu>/online`.
+            """,
```
squirrelsc, line 52: Bring CPUs online after tested. Other test cases may run on the same node.

lisa/tools/test.py
```diff
@@ -0,0 +1,22 @@
...
+        return True
+
+    def file_exists(self, file_path: str) -> bool:
+        # Run `test -e <file_path>`. This commands doesn't
+        # return any output and sets the last exit code
+        # with the result.
+        result = self.run(f"-e {file_path}")
+        if result.exit_code == 0:
+            return True
+        return False
```
squirrelsc: It can be simplified like `return result.exit_code == 0`.

microsoft/testsuites/core/cpu.py
dsrivastavv: Is it possible to delete the environment associated with this test? The test could potentially take the environment into a bad state if this test fails.

microsoft/testsuites/core/cpu.py
dsrivastavv: This loop is required. Not all CPU's from 1..cpu_count could be associated with the lsvmbus. We only try to bring down CPU associated with lsvmbus, and then check that the operation doesn't succeed.

microsoft/testsuites/core/cpu.py
dsrivastavv: This is the unix file associated with CPU state. The other commands use info from this file.

lisa/util/process.py
dsrivastavv: This command need to be run as `root` user. 

Running with `sudo` gives : `-bash: /sys/devices/system/cpu/cpu1/online: Permission denied`.
Running with `sudo su -c` gives : `bash: line 0: echo: write error: Device or resource busy`.

microsoft/testsuites/core/cpu.py
dsrivastavv: The intent of the test if to check the CPU's associated with the lsvmbus **cannot** be put offline.

lisa/util/process.py
squirrelsc: did you try with `sudo -sh -c`? the reason may be the command is not support by shell. Some part of command line are not delivered in root shell.

microsoft/testsuites/core/cpu.py
squirrelsc: Yes, it's possible. We should have another exception to make sure recycle the environment, if a test case declare the environment is in bad state. But anyway, we should try to recover this environment after test case completed.

microsoft/testsuites/core/cpu.py
squirrelsc: thank you for explain!

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,47 @@
...
+        priority=1,
+    )
+    def cpu_verify_online(self, node: Node, log: Logger) -> None:
```
squirrelsc: the case name can be more specified. like `cpu_verify_vmbus_force_online`? Other name is also ok.

dsrivastavv: Sure! This name conveys the intent better.

lisa/util/process.py
dsrivastavv: This also gives the same result : 
```
lisa@cvm-test:~$ sudo -sh -c echo 0 > /sys/devices/system/cpu/cpu1/online
-bash: /sys/devices/system/cpu/cpu1/online: Permission denied
```

squirrelsc: Please try in Python code. If test it in the bash, it should be like below. The ssh shell will add quote to make sure each part works. Let me know, if it doesn't work as expected.

`sudo -sh -c "echo 0 > /sys/devices/system/cpu/cpu1/online"`

lisa/tools/echo.py
LiliDeng: "'{val}' > {file_path}"?

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,47 @@
...
+            1. Get the list of lsvmbus channel cpu mappings using
+            command `lsvmbus -vv`.
+            2. Create a set of cpu's assigned to lsvmbus channels.
+            3. Try to put cpu offline by running
+            `echo 0 > /sys/devices/system/cpu/cpu/<cpu_id>/online`.
+            Note : We skip cpu 0 as it handles system interrupts.
+            4. Ensure that cpu is still online by checking state '1' in
+            `/sys/devices/system/cpu/cpu/<target_cpu>/online`.
+            """,
+        priority=1,
```
LiliDeng: v3 priority = v2 priority + 1

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,47 @@
...
+            file_path = f"/sys/devices/system/cpu/cpu{target_cpu}/online"
+            file_exists = node.tools[Test].file_exists(file_path)
+            if file_exists:
+                node.tools[Echo].write_to_file(
+                    CPUState.OFFLINE, file_path, super_user=True
+                )
+                result = node.tools[Cat].read_from_file(file_path)
+                assert_that(
+                    result,
+                    f"CPU {target_cpu} can't be offline...",
```
LiliDeng: f"CPU {target_cpu} can't be offline..." will mislead users.

lisa/base_tools/cat.py
```diff
@@ -11,3 +11,12 @@ def command(self) -> str:
...
         return True
+
+    def read_from_file(
+        self,
+        file_path: str,
+        force_run: bool = False,
+        sudo: bool = False,
+    ) -> str:
+        # Run `cat <file_path>`
+        return self.run(file_path, force_run=force_run, sudo=sudo).stdout
```
squirrelsc: raise exception, if the file doesn't exist. It can be implemented by assert the exit code.

lisa/tools/echo.py
```diff
@@ -18,6 +18,24 @@ def _windows_tool(cls) -> Optional[Type[Tool]]:
     def _check_exists(self) -> bool:
         return True
 
+    def write_to_file(
+        self,
+        val: str,
+        file_path: str,
```
squirrelsc: `file` is enough for the name.

lisa/tools/echo.py
```diff
@@ -18,6 +18,24 @@ def _windows_tool(cls) -> Optional[Type[Tool]]:
     def _check_exists(self) -> bool:
         return True
 
+    def write_to_file(
+        self,
+        val: str,
```
squirrelsc: Use `value`. If it's not too long, let's use full names.

lisa/tools/echo.py
```diff
@@ -18,6 +18,24 @@ def _windows_tool(cls) -> Optional[Type[Tool]]:
...
+        file_path: str,
+        force_run: bool = False,
```
squirrelsc: The force_run doesn't need to be a parameter, and it can be True always in this scenario. We want every time the write operation is executed.

lisa/tools/echo.py
```diff
@@ -18,6 +18,24 @@ def _windows_tool(cls) -> Optional[Type[Tool]]:
...
+        sudo: bool = False,
+    ) -> None:
+        # Run `echo <val> > <file_path>`
+        output = self.run(
+            f"{val} > {file_path}",
+            force_run=force_run,
+            shell=shell,
+            sudo=sudo,
+        ).stdout
+        self._log.debug(output)
```
squirrelsc: It should be output automatically. It doesn't need to output again.

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,57 @@
...
+        for target_cpu in mapped_cpu:
+            log.debug(f"Checking CPU {target_cpu} on /sys/device/....")
+            result = self._set_cpu_state(target_cpu, CPUState.OFFLINE, node)
+            if result:
+                # Try to bring CPU back to it's original state
+                reset = self._set_cpu_state(target_cpu, CPUState.ONLINE, node)
+                if not reset:
+                    raise BadEnvironmentStateException("Unable to bring CPU online...")
+                raise AssertionError(
+                    f"CPU {target_cpu} is mapped to a LSVMBUS channel. "
```
squirrelsc: The "can't" is confusing. It needs to specify what's expected, and what's actual. In general "should/shouldn't" is clearer than "can/cannot", or explain "expected/actual" results.

lisa/base_tools/cat.py
```diff
@@ -11,3 +13,14 @@ def command(self) -> str:
...
+
+    def read_from_file(
+        self,
+        file: str,
+        force_run: bool = False,
+        sudo: bool = False,
+    ) -> str:
+        # Run `cat <file>`
+        result = self.run(file, force_run=force_run, sudo=sudo)
+        assert_that(result.exit_code, f"Error : {result.stdout}").is_equal_to(0)
```
squirrelsc: The result has `assert_exit_code`, it's easy to reuse.

microsoft/testsuites/core/cpu.py
```diff
@@ -19,6 +27,55 @@
...
+        for target_cpu in mapped_cpu:
+            log.debug(f"Checking CPU {target_cpu} on /sys/device/....")
+            result = self._set_cpu_state(target_cpu, CPUState.OFFLINE, node)
+            if result:
+                # Try to bring CPU back to it's original state
+                reset = self._set_cpu_state(target_cpu, CPUState.ONLINE, node)
+                if not reset:
+                    raise BadEnvironmentStateException("Unable to bring CPU online...")
+                raise AssertionError(
+                    f"Expected CPU {target_cpu} state : {CPUState.ONLINE}, "
```
squirrelsc: Rather than introduce the values, also specify why it's expected to this value. It helps people understand the failure from the error message directly.

lisa/tools/test.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Test(Tool):
+    @property
+    def command(self) -> str:
+        return "test"
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def file_exists(
```
squirrelsc: Sorry. I just remembered the shell.exists can do this..

squirrelsc (approved):

PR #1486, opened 2021-08-22
Feature: support feature settings for rich feature requirements, enable settings on the Disk feature and small refactoring

squirrelsc:
See commits for details.

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -692,10 +692,6 @@ class Environment:
 @dataclass_json()
 @dataclass
 class EnvironmentRoot:
-    max_concurrency: int = field(
-        default=1,
-        metadata=metadata(field_function=fields.Int, validate=validate.Range(min=1)),
-    )
     warn_as_error: bool = field(default=False)
     environments: List[Environment] = field(default_factory=list)
 
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -390,7 +390,7 @@ class FeatureSettings(
 
 
 class DiskType(str, Enum):
-    PremiumLRS = "PremiumLRS"
+    PremiumSSDLRS = "PremiumSSDLRS"
     Ephemeral = "Ephemeral"
     StandardHDDLRS = "StandardHDDLRS"
     StandardSSDLRS = "StandardSSDLRS"
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -420,12 +420,89 @@ class DiskOptionSettings(FeatureSettings):
         default=DiskType.StandardHDDLRS,
         metadata=metadata(decoder=_decode_disk_type),
     )
+    data_disk_count: search_space.CountSpace = field(
+        default=search_space.IntRange(min=0),
+        metadata=metadata(decoder=search_space.decode_count_space),
+    )
+    data_disk_caching_type: str = field(
+        default=constants.DATADISK_CACHING_TYPE_NONE,
+        metadata=metadata(
+            validate=validate.OneOf(
+                [
+                    constants.DATADISK_CACHING_TYPE_NONE,
+                    constants.DATADISK_CACHING_TYPE_READONLY,
+                    constants.DATADISK_CACHING_TYPE_READYWRITE,
+                ]
+            ),
+        ),
+    )
+    data_disk_iops: search_space.CountSpace = field(
+        default=None,
+        metadata=metadata(allow_none=True, decoder=search_space.decode_count_space),
+    )
+    data_disk_size: search_space.CountSpace = field(
+        default=None,
+        metadata=metadata(allow_none=True, decoder=search_space.decode_count_space),
+    )
+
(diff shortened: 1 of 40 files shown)


lisa/testsuite.py
```diff
@@ -154,11 +154,17 @@ def handle_exception(
...
+                    f"case failed and ignored. "
+                    f"{exception.__class__.__name__}: {exception}"
+                )
                 self.set_status(TestStatus.ATTEMPTED, f"{phase}{exception}")
             else:
                 log.error("case failed", exc_info=exception)
-                self.set_status(TestStatus.FAILED, f"{phase}failed: {exception}")
+                self.set_status(
+                    TestStatus.FAILED,
+                    f"{phase}failed. {exception.__class__.__name__}: {exception}",
```
glima, line 170: Please put a space between phase and failed.

glima:
Fairly complex code, but otherwise nothing outstanding seen. Would advise more +1s :)

lisa/testsuite.py
squirrelsc: The phase is optional, and a white space was appended to the phase at the beginning of this method, if it's not empty. So, if it's empty, there won't be a whitespace output. If there is phase, it will be like "deploy failed".

lisa/schema.py
```diff
@@ -420,12 +420,86 @@ class DiskOptionSettings(FeatureSettings):
...
+        return (
+            self.type == o.type
+            and self.disk_type == o.disk_type
+            and self.data_disk_count == o.data_disk_count
+            and self.data_disk_caching_type == o.data_disk_caching_type
+            and self.data_disk_iops == o.data_disk_iops
+            and self.data_disk_size == o.data_disk_size
+        )
+
+    def __repr__(self) -> str:
```
dsrivastavv, line 459: [nit] Make this `__str__` method and call it in `__repr__`?

lisa/schema.py
```diff
@@ -389,6 +389,45 @@ def _generate_min_capability(self, capability: Any) -> Any:
         return FeatureSettings.create(self.type)
 
 
+class DiskType(str, Enum):
+    PremiumLRS = "PremiumLRS"
```
dsrivastavv: `PremiumLRS` is technically PremiumSSDLRS. Can you please add a comment here?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -203,41 +195,135 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
-class DiskEphemeral(AzureFeatureMixin, features.DiskEphemeral):
-    def _initialize(self, *args: Any, **kwargs: Any) -> None:
-        super()._initialize(*args, **kwargs)
-        self._initialize_information(self._node)
+# disk types are ordered by cost. The earlier is lower cost.
+_ordered_disk_types: List[schema.DiskType] = [
+    schema.DiskType.StandardHDDLRS,
+    schema.DiskType.StandardSSDLRS,
+    schema.DiskType.Ephemeral,
+    schema.DiskType.PremiumLRS,
```
dsrivastavv: Shouldn't `Ephemeral` be the least cost disk? And we use StandardHDD as default because it is most common disk option.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -203,41 +195,135 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+        if self.disk_type is None:
+            req_disk_type = capability.disk_type
+        else:
+            req_disk_type = self.disk_type
+
+        # find the min disk type from the order by cost.
+        min_disk_type: Optional[schema.DiskType] = None
+        for expected_disk_type in _ordered_disk_types:
+            if (
+                expected_disk_type in req_disk_type
```
dsrivastavv, line 331: Is this loop required? Wouldn't `req_disk_type` be set to `StandardHDDLRS` if tests don't specify anything and all azure VM's support this?

lisa/schema.py
squirrelsc: added

lisa/schema.py
squirrelsc: Ok, I will change its name, it would be more clear.

lisa/sut_orchestrator/azure/features.py
squirrelsc: Updated the comments to consider also common or not. The ephemeral disk is special, it doesn't save data to storage, so it's used by stateless service only. If the vm is deallocated, the data is lost. So it's not a good candidate for normal test cases.

lisa/sut_orchestrator/azure/features.py
squirrelsc: The requirement can be a set with multiple values.

dsrivastavv (approved):

LiliDeng (approved):
Tested

PR #1487, opened 2021-08-23
Use kernel_version_raw, not the object.

LiliDeng:


lisa/environment.py
```diff
@@ -456,7 +456,7 @@ def get_environment_information(self, environment: Environment) -> Dict[str, str
                 if node.is_connected and node.is_posix:
                     uname = node.tools[Uname]
                     linux_information = uname.get_linux_information()
-                    fields = ["hardware_platform", "kernel_version"]
+                    fields = ["hardware_platform", "kernel_version_raw"]
```
squirrelsc: The name need to be changed to "kernel_version"

squirrelsc (approved):

PR #1489, opened 2021-08-24
DPDK Suite, Testpmd tool, NodeNicInfo class

mcgov:
Initial build and test of the testpmd tool and dpdk suite.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -1,0 +1,243 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import itertools
+from typing import Dict, List
+
+from assertpy import assert_that
+
+from lisa import Logger, Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa.features import Sriov
+from lisa.testsuite import simple_requirement
+from lisa.tools import Echo, Lspci, Mount
+from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
+
+
+@TestSuiteMetadata(
+    area="dpdk",
+    category="functional",
+    description="""
+    This test suite check DPDK functionality
+    """,
+)
+class Dpdk(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            This test case checks DPDK can be built and installed correctly.
+        """,
+        requirement=simple_requirement(
+            supported_features=[Sriov],
+        ),
+        priority=1,
+    )
+    def check_dpdk_build(self, node: Node, log: Logger) -> None:
+        sriov_feature = node.features[Sriov]
+        sriov_is_enabled = sriov_feature.enabled()
+        log.info(f"Verify SRIOV is enabled: {sriov_is_enabled}")
+        assert_that(sriov_is_enabled).described_as(
+            "SRIOV was not enabled for this test node."
+        ).is_true()
+
+        # dump some info about the pci devices before we start
+        lspci = node.tools[Lspci]
+        log.info(f"LSPCI Info:\n{lspci.run().stdout}\n")
+
+        # enable hugepages (required by dpdk)
+        self._hugepages_init(node)
+        self._hugepages_enable(node, log)
+
+        # initialize testpmd tool (installs dpdk)
+        testpmd = DpdkTestpmd(node)
+        testpmd.install()
+
+        # initialize node nic info class (gathers info about nic devices)
+        node_nic_info = NodeNicInfo(node)
+        assert_that(len(node_nic_info)).described_as(
+            "Test needs at least 2 NICs on the test node."
+        ).is_greater_than_or_equal_to(2)
+        log.info(node_nic_info)
+
+        # grab a nic and run testpmd
+        test_nic = node_nic_info.get_nic(node_nic_info.get_upper_nics()[-1])
+        vdev_type = "net_vdev_netvsc0"
+        testpmd_include_str = test_nic.testpmd_include(vdev_type)
+        testpmd_output = testpmd.run_with_timeout(testpmd_include_str, 30)
+        tx_pps = testpmd.get_tx_pps_from_testpmd_output(testpmd_output)
+        log.info(
+            f"TX-PPS:{tx_pps} from {test_nic._upper}/{test_nic._lower}:"
+            + f"{test_nic._pci_slot}"
+        )
+        assert_that(tx_pps).described_as(
+            f"TX-PPS ({tx_pps}) should have been greater than 2^20 (~1m) PPS."
+        ).is_greater_than(2 ** 20)
+
+    def _execute_expect_zero(self, node: Node, cmd: str) -> str:
+        result = node.execute(cmd, sudo=True, shell=True)
+        result.assert_exit_code()
+        return result.stdout
+
+    def _hugepages_init(self, node: Node) -> None:
+        mount = node.tools[Mount]
+        # When using hugetlbfs the first parameter is ignored, so it could be anything
+        mount.run(" -t hugetlbfs ignored /mnt/huge", sudo=True)
+        mount.run("-t hugetlbfs -o 'pagesize=1G' ignored /mnt/huge-1G", sudo=True)
+
+        # TODO: swap to this once the mount tool issue is fixed
+        #           https://github.com/microsoft/lisa/issues/1505
+        # mount.mount("nodev", "/mnt/huge", type="hugetlbfs")
+        # mount.mount("nodev", "/mnt/huge-1G",
+        #              type="hugetlbfs", options="'pagesize=1G'")
+
+    def _hugepages_enable(self, node: Node, log: Logger) -> None:
+        echo = node.tools[Echo]
+        echo.write_to_file(
+            "1024",
+            "/sys/devices/system/node/node0/hugepages/hugepages-2048kB/nr_hugepages",
+            sudo=True,
+        )
+        echo.write_to_file(
+            "1",
+            "/sys/devices/system/node/node0/hugepages/hugepages-1048576kB/nr_hugepages",
+            sudo=True,
+        )
+
+
+class NicInfo:
+
+    # Class for info about an single upper/lower nic pair.
+    # devices using SRIOV on azure typically have an upper synthetic device
+    # paired with a lower SRIOV Virtual Function (VF) device that
+    # enables the passthrough to the physical NIC.
+
+    def __init__(self, upper: str, lower: str = "", pci_slot: str = "") -> None:
+        self._has_secondary = lower != "" or pci_slot == ""
+        self._upper = upper
+        self._lower = lower
+        self._pci_slot = pci_slot
+
+    def has_secondary(self) -> bool:
+        return self._has_secondary
+
+    def __str__(self) -> str:
+        return (
+            "NicInfo:\n"
+            f"upper: {self._upper}\n"
+            f"lower: {self._lower}\n"
+            f"pci_slot: {self._pci_slot}\n"
+        )
+
+    def testpmd_include(self, vdev_type: str) -> str:
+        assert_that(self._has_secondary).is_true().described_as(
+            (
+                f"This interface {self._upper} does not have a lower interface "
+                "and pci slot associated with it. Aborting."
+            )
+        )
+        return f'--vdev="{vdev_type},iface={self._upper}" --allow "{self._pci_slot}"'
+
+
+class NodeNicInfo:
+
+    # Class for all of the nics on a node. Contains multiple NodeNic classes.
+    # Init identifies upper/lower paired devices and the pci slot info for the lower.
+
+    def _get_nic_names(self, node: Node) -> List[str]:
+        # identify all of the nics on the device, excluding tunnels and loopbacks etc.
+        result = node.execute(
+            " ls /sys/class/net/ | grep -Ev $(ls /sys/devices/virtual/net)",
+            shell=True,
+            sudo=True,
+        )
+        nic_names = result.stdout.splitlines()
+        for item in nic_names:
+            assert_that(item).described_as(
+                "nic name could not be found"
+            ).is_not_equal_to("")
+        return nic_names
+
+    def _get_nic_device(self, node: Node, nic_name: str) -> str:
+        slot_info_result = node.execute(f"readlink /sys/class/net/{nic_name}/device")
+        slot_info_result.assert_exit_code()
+        base_device_result = node.execute(f"basename {slot_info_result.stdout}")
+        base_device_result.assert_exit_code()
+        # todo check addr matches expectation
+        return base_device_result.stdout
+
+    def _get_node_nic_info(self, node: Node, nic_list: List[str]) -> None:
+        # Identify which nics are slaved to master devices.
+        # This should be really simple with /usr/bin/ip but experience shows
+        # the tool isn't super consistent across distros in this regard
+
+        # use sysfs to gather upper/lower nic pairings and pci slot info
+        for pairing in itertools.permutations(nic_list, 2):
+            upper_nic, lower_nic = pairing
+            # check a nic pairing to identify upper/lower relationship
+            upper_check = node.execute(
+                f"readlink /sys/class/net/{lower_nic}/upper_{upper_nic}"
+            )
+            if upper_check.exit_code == 0:
+                assert_that(upper_check.stdout).is_not_equal_to("")
+                pci_slot = self._get_nic_device(node, lower_nic)
+                assert_that(pci_slot).is_not_empty()
+                # check pcislot info looks correct
+                nic_info = NicInfo(upper_nic, lower_nic, pci_slot)
+                self.append(nic_info)
+
+        # identify nics which don't have a pairing (non-AN devices)
+        for nic in nic_list:
+            if not self.nic_info_is_present(nic):
+                self.append(NicInfo(nic))
+
+        assert_that(self.is_empty()).is_false()
+
+    def _get_default_nic(self, node: Node, nic_list: List[str]) -> str:
+        cmd = "ip route | grep default | awk '{print $5}'"
+        default_interface = node.execute(cmd)
+        assert_that(default_interface.exit_code).is_zero()
+        assert_that(len(default_interface.stdout)).is_greater_than(0)
+        default_interface_name = default_interface.stdout.strip()
+        assert_that(default_interface_name in nic_list).described_as(
+            (
+                f"ERROR: NIC name found as default {default_interface_name} "
+                f"was not in original list of nics {nic_list}."
+            )
+        ).is_true()
+        return default_interface_name
+
+    def __init__(self, node: Node):
+        self._nics: Dict[str, NicInfo] = dict()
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,155 @@
...
+        elif isinstance(node.os, Redhat) or isinstance(node.os, CentOs):
+            self.__execute_assert_zero(
+                "yum update -y --disablerepo='*' --enablerepo='*microsoft*'", cwd
+            )
+            node.execute(
+                "yum groupinstall -y 'Infiniband Support'",
+                sudo=True,
+                shell=True,
+            )
+            for package in self._redhat_packages:
```
squirrelsc: The packages can be installed by the list, that would be more efficient than one by one.

microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,155 @@
...
+                node.os.install_packages(package)
+            self.__execute_assert_zero("pip3 install --upgrade meson", cwd)
+            self.__execute_assert_zero("mv /usr/bin/meson /usr/bin/meson.bak", cwd)
+            self.__execute_assert_zero("ln -s /usr/local/bin/meson /usr/bin/meson", cwd)
+            self.__execute_assert_zero("pip3 install --upgrade ninja", cwd)
+            self.__execute_assert_zero(
+                "modprobe -a rdma_cm rdma_ucm ib_core ib_uverbs mlx4_core mlx4_ib", cwd
+            )
+
+        elif isinstance(node.os, Redhat) or isinstance(node.os, CentOs):
```
squirrelsc: `isinstance(node.os, Redhat)` is enough, because the CentOs is subclass of it.

microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,155 @@
...
+            self.__execute_assert_zero(
```
squirrelsc: This step is handled by os already.

microsoft/testsuites/dpdk/testpmd.py
mcgov: Ah, I wanted it to error out on a specific package if there was a problem installing one of them. I can swap it out to use the list now that I've sorted out which ones are needed.

squirrelsc: The log of os is optimized, but may not good enough. It should be able to show which package is failed to install. If it doesn't work as expected, please help on fixing it.

microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,155 @@
...
+                "yum update -y --disablerepo='*' --enablerepo='*microsoft*'", cwd
+            )
+            node.execute(
```
squirrelsc: It looks group install is useful in Redhat, we can implement it in os level. So the above line can be called automatically too.

microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,155 @@
...
+
+            wget_tool = self.node.tools[Wget]
+            wget_tool.get(self._ninja_url)
+            node.execute(f"mv ninja-linux.zip {node.working_path}/")
+            self.__execute_assert_zero(
+                "unzip ninja-linux.zip && mv ninja /usr/bin/ninja", cwd
+            )
+            self.__execute_assert_zero("pip3 install --upgrade pyelftools", cwd)
+        else:
+            return False
```
squirrelsc: It's better to raise an exception instead of returning false. If the OS is not supported, either skip or fail the test. False has no enough information for troubleshooting.

microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,155 @@
...
+        return f"{self._testpmd_install_path} -l 2,3 -n 4 --proc-type=primary {nic_to_include} -- --forward-mode=txonly -a --stats-period 1"
+
+    def run_with_timeout(self, nic_to_include: str, timeout: int) -> str:
+        timer_proc = self.node.execute_async(
+            f"sleep {timeout} && killall -s INT {self._testpmd_install_path}",
+            sudo=True,
+            shell=True,
+        )
+        testpmd_proc = self.node.execute_async(
+            f"{self._testpmd_install_path} -l 2,3 -n 4 --proc-type=primary {nic_to_include} -- --forward-mode=txonly -a --stats-period 1",
```
squirrelsc: This line is same as `generate_testpmd_command`

microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,155 @@
...
+            node.execute(f"mv ninja-linux.zip {node.working_path}/")
+            self.__execute_assert_zero(
+                "unzip ninja-linux.zip && mv ninja /usr/bin/ninja", cwd
+            )
+            self.__execute_assert_zero("pip3 install --upgrade pyelftools", cwd)
+        else:
+            return False
+        return True
+
+    def generate_testpmd_command(self, nic_to_include: str) -> str:
```
squirrelsc: It's not used. If it's used inside of the tool, start it with `_`. It means it's a private method.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+            "Test needs at least 2 NICs on the test node."
+        ).is_greater_than_or_equal_to(2)
+
+        self.log.info(node_nic_info)
+
+        test_nic = node_nic_info.get_nic(node_nic_info.get_upper_nics()[-1])
+        vdev_type = "net_vdev_netvsc"
+        testpmd_include_str = test_nic.testpmd_include(vdev_type)
+        output = testpmd.run_with_timeout(testpmd_include_str, 30)
+        self.log.info(output)
```
squirrelsc: Is there anything need to be checked? The log doesn't validate if something right or wrong.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def _execute_expect_zero(self, node: Node, cmd: str) -> str:
+        self.log.info(f"RUNCMD: {cmd}")
+        result = node.execute(cmd, sudo=True, shell=True)
+        assert_that(result.exit_code).described_as(
```
squirrelsc: It may not good for all commands. The output of some commands are too long to be in error message.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+            f"{cmd} failed with code {result.exit_code} and stdout+stderr:"
+            + f"\n{result.stdout}\n=============\n{result.stderr}\n=============\n"
+        ).is_zero()
+        if len(result.stdout) > 1024:
+            self.log.debug(
```
squirrelsc: The stdout will be in debug log by underlying logic. It doesn't need to add one more line here.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+                f"NOTE: Truncating result since output is larger than 1024.\n{result.stdout[:512]}\n.......(content has been truncated).......\n{result.stdout[-512:]}"
+            )
+        else:
+            self.log.debug(f"{cmd}:\n{result.stdout}")
```
squirrelsc: It's output by underlying logic already.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+
+        return result.stdout
+
+    def _hugepages_init(self, node: Node) -> None:
+        self._execute_expect_zero(node, "mkdir -p /mnt/huge")
+        self._execute_expect_zero(node, "mkdir -p /mnt/huge-1G")
+        self._execute_expect_zero(node, "mount -t hugetlbfs nodev /mnt/huge")
```
squirrelsc: Try to leverage existing mount tool.

squirrelsc: You can add a parameter to support `-t` type.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+        self._execute_expect_zero(
+            node, "mount -t hugetlbfs nodev /mnt/huge-1G -o 'pagesize=1G'"
+        )
+        result = node.execute("grep -i huge /proc/meminfo && ls /mnt/", shell=True)
+        self.log.info(f"hugepages status \n{result.stdout}")
```
squirrelsc: The output should be in log already, it doesn't need to output again. And it may not easy to read at info level.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I didn't see a mount tool available when I looked, is it merged yet?

squirrelsc: It's in tools folder, and can be imported like `from lisa.tools import Mount`.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+        self._has_secondary = lower != "" or pci_slot == ""
+        self._upper = upper
+        self._lower = lower
+        self._pci_slot = pci_slot
+
+    def has_secondary(self) -> bool:
+        return self._has_secondary
+
+    def __str__(self) -> str:
+        return f"""
```
squirrelsc: try like below, it makes indent looks better.

```python
return (
    f"NicInfo:\nupper: {self._upper}\nlower: {self._lower}"
    f"pci_slot: {self._pci_slot}"
)
```

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,221 @@
...
+        self._execute_expect_zero(
+            node,
+            "sh -c 'echo 1 > /sys/devices/system/node/node0/hugepages/hugepages-1048576kB/nr_hugepages'",
+        )
+        result = node.execute("grep -i huge /proc/meminfo && ls /mnt/", shell=True)
+        self.log.info(f"hugepages status \n{result.stdout}")
+
+
+class NicInfo:
+    def __init__(self, upper: str, lower: str = "", pci_slot: str = "") -> None:
```
squirrelsc, line 112: Add comments to explain upper/lower pair. It's hard to people who are not familiar with dpdk.

microsoft/testsuites/dpdk/testpmd.py
mcgov: That's fine, but we can change it after it's implemented. I'll add a TODO?

squirrelsc: I will add it in another PR and merge it tomorrow.

microsoft/testsuites/dpdk/testpmd.py
mcgov: It's good enough, it was a nice thing during debugging but now that it's stable it should actually be enough to see installation failed during the package installation step. Fixed to use the full list rather than single step

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: will do

microsoft/testsuites/dpdk/testpmd.py
```diff
@@ -0,0 +1,163 @@
...
+from lisa.util import SkippedException
+from assertpy import assert_that
+from pathlib import PurePath
+from typing import List, Type
+import re
+
+# TestPMD tool to bundle the DPDK build and toolset together.
+
+
+class Testpmd(Tool):
```
squirrelsc: How about a full name like DpdkTestpmd?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: using it now! Thank you :)

microsoft/testsuites/dpdk/testpmd.py
mcgov: Sure

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,242 @@
...
+            node,
+            (
+                "sh -c 'echo 4096 > "
+                "/sys/devices/system/node/node0/hugepages/"
+                "hugepages-2048kB/nr_hugepages'"
+            ),
+        )
+        self._execute_expect_zero(
+            node,
+            (
```
squirrelsc: use `echo.write_to_file`

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,242 @@
...
+                "sh -c 'echo 1 > "
+                "/sys/devices/system/node/node0/hugepages/"
+                "hugepages-1048576kB/nr_hugepages'"
+            ),
+        )
+        node.execute("grep -i huge /proc/meminfo && ls /mnt/", shell=True)
```
squirrelsc: Does it need to verify the output?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,242 @@
...
+        )
+        return f'--vdev="{vdev_type},iface={self._upper}" --allow "{self._pci_slot}"'
+
+
+class NodeNicInfo:
+
+    # Class for all of the nics on a node. Contains multiple NodeNic classes.
+    # Init identifies upper/lower paired devices and the pci slot info for the lower.
+
+    _pci_slot_regex = r"([0-9a-fA-F]{1-4}:)+[0-9a-fA-F]{1-4}"
```
squirrelsc: It looks not used. But if it's useful,
1. add examples above.
2. use `re.compile(...`. So the regexp can be precompiled, it doesn't need to be compiled every time.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,242 @@
...
+    def _get_nic_names(self, node: Node) -> List[str]:
+        # identify all of the nics on the device, excluding tunnels and loopbacks etc.
+        result = node.execute(
+            " ls /sys/class/net/ | grep -Ev $(ls /sys/devices/virtual/net)",
+            shell=True,
+            sudo=True,
+        )
+        nic_names = result.stdout.split("\r\n")
+        for item in nic_names:
+            assert_that(item).is_not_equal_to("").described_as(
```
squirrelsc: The `described_as` should be called before validation.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,242 @@
...
+    _nics: Dict[str, NicInfo] = dict()
+
+    def _get_nic_names(self, node: Node) -> List[str]:
+        # identify all of the nics on the device, excluding tunnels and loopbacks etc.
+        result = node.execute(
+            " ls /sys/class/net/ | grep -Ev $(ls /sys/devices/virtual/net)",
+            shell=True,
+            sudo=True,
+        )
+        nic_names = result.stdout.split("\r\n")
```
squirrelsc: call `splitlines`

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,242 @@
...
+        for item in nic_names:
+            assert_that(item).is_not_equal_to("").described_as(
+                "nic name could not be found"
+            )
+        return nic_names
+
+    def _get_nic_device(self, node: Node, nic_name: str) -> str:
+        slot_info_result = node.execute(f"readlink /sys/class/net/{nic_name}/device")
+        assert_that(slot_info_result.exit_code).is_zero()
```
squirrelsc: call `slot_info_result.assert_exit_code`

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -0,0 +1,242 @@
...
+        return f'--vdev="{vdev_type},iface={self._upper}" --allow "{self._pci_slot}"'
+
+
+class NodeNicInfo:
+
+    # Class for all of the nics on a node. Contains multiple NodeNic classes.
+    # Init identifies upper/lower paired devices and the pci slot info for the lower.
+
+    _pci_slot_regex = r"([0-9a-fA-F]{1-4}:)+[0-9a-fA-F]{1-4}"
+    _nics: Dict[str, NicInfo] = dict()
```
squirrelsc: Don't declare variables in class level in Python, because it would be a class variable. It means all objects share the same variable. Declare it in `__int__`.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -0,0 +1,164 @@
...
+            self._generate_testpmd_command(nic_to_include),
+            sudo=True,
+        )
+        timer_proc.wait_result()
+        proc_result = testpmd_proc.wait_result()
+        return proc_result.stdout
+
+    def get_tx_pps_from_testpmd_output(self, output: str) -> int:
+        matches = re.findall(r"Tx-pps:\s+([0-9]+)", output)
+        assert_that(len(matches)).is_not_zero()
```
squirrelsc: 1. add comments to explain it, once it's failed.
2. try `assert_that(matches).is_length(0)`

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: No, I'll just remove it I was previously printing the output just for validation but it's fine

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Will change, I feel like node.execute should just have a keyword arg for expected error code and optionally turn off assertions if you don't care. It's much rarer to not expect a 0 error code or not care about the result so this could get rid of a lot of extra code if we just asserted on it by default and let the user turn the assert off if they want or change the error code if they want.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: there isn't a hard length requirement, length just has to be larger than 0. I'll add a better description.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: removed that regex, moved the other stuff

microsoft/testsuites/dpdk/dpdksuite.py

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: Yes, we can have that. I will add such an functionality. And it's off by default.

squirrelsc (dismissed):

squirrelsc:
The code looks good to me now. Please get commits meaningful, and fix the CI errors. try module `dpdktestpmd` directly?

squirrelsc (approved):

PR #1490, opened 2021-08-24
Fix errors with CPU test cases

dsrivastavv:
The priority of l3_cache_check should be 2 as v3 priority = v2 priority + 1

### DIFF
--- a/lisa/tools/lscpu.py
+++ b/lisa/tools/lscpu.py
@@ -34,6 +34,20 @@ class CPUInfo:
         self.l2_cache = l2_cache
         self.l3_cache = l3_cache
 
+    def __str__(self) -> str:
+        return (
+            f"cpu : {self.cpu}, "
+            f"numa_node : {self.numa_node}, "
+            f"socket : {self.socket}, "
+            f"l1_data_cache : , {self.l1_data_cache}, "
+            f"l1_instruction_cache : {self.l1_instruction_cache}, "
+            f"l2_cache : {self.l2_cache}, "
+            f"l3_cache : {self.l3_cache}"
+        )
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
 
 class Lscpu(Tool):
     # CPU(s):              16
@@ -41,11 +55,12 @@ class Lscpu(Tool):
     # Architecture:        x86_64
     __architecture_pattern = re.compile(r"^Architecture:\s+(.*)?\r$", re.M)
     __valid_architecture_list = ["x86_64"]
-    # Capture patterns of the form `0 0 0 0:0:0:0`
+    # 0 0 0 0:0:0:0
+    # 96 0 10 1:1:1:0
     _core_numa_mappings = re.compile(
-        r".*(?P<cpu>\d+).+(?P<numa_node>\d+).+(?P<socket>\d+)"
-        r".+(?P<l1_data_cache>\d+):(?P<l1_instruction_cache>\d+)"
-        r":(?P<l2_cache>\d+):(?P<l3_cache>\d+)$"
+        r"\s*(?P<cpu>\d+)\s+(?P<numa_node>\d+)\s+(?P<socket>\d+)\s+"
+        r"(?P<l1_data_cache>\d+):(?P<l1_instruction_cache>\d+):"
+        r"(?P<l2_cache>\d+):(?P<l3_cache>\d+)$"
     )
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/core/cpu.py
```diff
@@ -31,8 +31,9 @@ class CPU(testsuite.TestSuite):
     def _set_cpu_state(self, cpu_id: str, state: str, node: Node) -> bool:
         file_path = f"/sys/devices/system/cpu/cpu{cpu_id}/online"
         file_exists = node.shell.exists(PurePosixPath(file_path))
-        if file_exists:
-            node.tools[Echo].write_to_file(state, file_path, shell=True, sudo=True)
+        if not file_exists:
```
squirrelsc: what's the reason the file is not here? Should the test case be skipped, or just skip some of cpus?

lisa/tools/lscpu.py
```diff
@@ -43,9 +57,9 @@ class Lscpu(Tool):
     __valid_architecture_list = ["x86_64"]
     # Capture patterns of the form `0 0 0 0:0:0:0`
     _core_numa_mappings = re.compile(
-        r".*(?P<cpu>\d+).+(?P<numa_node>\d+).+(?P<socket>\d+)"
-        r".+(?P<l1_data_cache>\d+):(?P<l1_instruction_cache>\d+)"
-        r":(?P<l2_cache>\d+):(?P<l3_cache>\d+)$"
+        r"\s*(?P<cpu>\d+)\s+(?P<numa_node>\d+)\s+(?P<socket>\d+)\s+"
```
squirrelsc, line 61: Add newly found examples above. So if you find more instances in future, they can be tested together.

microsoft/testsuites/core/cpu.py
dsrivastavv: The file will not be present if the CPU is reserved for critical OS features. For example, in smaller VM's CPU0 could not be offline. The same thing happens with other CPU's on larger VM's such as `Standard_E96as_v4`

squirrelsc: It happens on some of CPUs or all of them? Can we skip it, if it applies to all of CPUs?

lisa/tools/lscpu.py
dsrivastavv: It's not a new example as such. Earlier regex wasn't capturing two digit `cpu_id`. The intent that CPU id could be a n digit number should be clear from the current comment.

microsoft/testsuites/core/cpu.py
dsrivastavv: I didn't come across any VM which has all it's CPU's reserved. And I don't think we should skip this test. The intent of the test is to check that all the CPU's assigned to lsvmbus could not be offline. In this scenario, the CPU's are still assigned to lsvmbus even if they are reserved.

dsrivastavv: The test should be skipped if the conditions could be not be verified or favorable to run test. But in this scenario, we are able to verify that all the cpu's mapped to lsvmbus could not be offline.

lisa/tools/lscpu.py
squirrelsc: can both old and new patterns capture the example, "0 0 0 0:0:0:0" before? If so, what's improved by changing this pattern?

microsoft/testsuites/core/cpu.py
```diff
@@ -31,8 +31,9 @@ class CPU(testsuite.TestSuite):
...
+            return False
+        node.tools[Echo].write_to_file(state, file_path, shell=True, sudo=True)
```
squirrelsc: BTW, the shell shouldn't be a parameter of this method. It should be always true underlying, because the command doesn't work without it.

microsoft/testsuites/core/cpu.py
squirrelsc: My understanding is the goal is to check the cpu cannot be set to offline by `echo 0 > online`, and check its value. If there is no chance to set it on every cpu. It shouldn't be covered. Maybe some distro or cpu doesn't support this online/offline switch, but it may support another way.

lisa/tools/lscpu.py
dsrivastavv: `0 0 0:0:0:0` is an example pattern where every 0 signifies a integer. How about I can update it to `[int] [int] [int]:[int]:[int]:[int]` to be explicit about n-digit integer?

squirrelsc: It's better to have real cases. Once it needs to include more cases, people can easily do regression testing from those cases. Check examples in operating_system.py. The patterns are updated by new findings, the examples can help preventing regression.

microsoft/testsuites/core/cpu.py
dsrivastavv: Based on offline discussion, we decided to fail the tests if the file is not present for all the cpu's. This is to ensure that we have a better understanding of distros which have such behaviors.

squirrelsc (approved):

PR #1493, opened 2021-08-25
Temp fix for standardDCSv2Family doesn't support Accelerated Networking, but API return True.

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1282,7 +1282,13 @@ def _resource_sku_to_capability(
                 node_space.gpu_count = int(sku_capability.value)
                 # update features list if gpu feature is supported
                 node_space.features.update([features.Gpu.name()])
-            elif name == "AcceleratedNetworkingEnabled":
+            # refer https://docs.microsoft.com/en-us/azure/virtual-machines/dcv2-series#configuration # noqa: E501
+            # standardDCSv2Family doesn't support `Accelerated Networking`
+            # but API return `True`, fix this issue temporarily
+            # will revert it till bug fixed.
+            elif name == "AcceleratedNetworkingEnabled" and resource_sku.family not in [
```
squirrelsc: Make it a standalone condition, so it's easy to remove.

```python
elif name == "AcceleratedNetworkingEnabled":
    # comments
    if resource_sku.family == "standardDCSv2Family":
        continue
    ...
```

squirrelsc (dismissed):

squirrelsc (approved):

PR #1495, opened 2021-08-25
Add verify_vmbus_interrupts CPU test

dsrivastavv:
Migrate `verify_vmbus_interrupts` test : https://github.com/microsoft/lisa/blob/master/Testscripts/Linux/verify_vmbus_interrupts.sh#L20

### DIFF
--- a/lisa/tools/taskset.py
+++ b/lisa/tools/taskset.py
@@ -1,0 +1,18 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+from lisa.util.process import Process
+
+
+class TaskSet(Tool):
+    @property
+    def command(self) -> str:
+        return "taskset"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def run_on_specific_cpu(self, cpu_id: int) -> Process:
+        return self.run_async(f"-c {cpu_id} yes > /dev/null", shell=True)
--- a/microsoft/testsuites/core/cpu.py
+++ b/microsoft/testsuites/core/cpu.py
@@ -1,7 +1,12 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+from __future__ import annotations
 
+import re
+import time
+from dataclasses import dataclass
 from pathlib import PurePosixPath
+from typing import List
 
 from assertpy.assertpy import assert_that
 
@@ -13,12 +18,79 @@ from lisa import (
     SkippedException,
     testsuite,
 )
-from lisa.tools import Cat, Echo, Lscpu, Lsvmbus, Uname
+from lisa.tools import Cat, Echo, Lscpu, Lsvmbus, TaskSet, Uname
 
 
 class CPUState:
     OFFLINE: str = "0"
     ONLINE: str = "1"
+
+
+hyperv_interrupt_substr = ["hyperv", "Hypervisor", "Hyper-V"]
+
+
+@dataclass
+class Interrupt:
+    irq_number: str
+    interrupt_count: List[int]
+    metadata: str
+
+    # 0:         22          0  IR-IO-APIC   2-edge      timer
+    _interrupt_regex = re.compile(
+        r"^\s*(?P<irq_number>\S+):\s+(?P<interrupt_count>[\d+ ]+)\s*(?P<metadata>.*)$"
+    )
+
+    def __init__(
+        self, irq_number: str, interrupt_count: List[int], metadata: str = ""
+    ) -> None:
+        self.irq_number = irq_number
+        self.interrupt_count = interrupt_count
+        self.metadata = metadata
+
+    def __str__(self) -> str:
+        return (
+            f"irq_number : {self.irq_number}, "
+            f"count : {self.interrupt_count}, "
+            f"metadata : {self.metadata}"
+        )
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    @staticmethod
+    def get_interrupt_data(node: Node) -> List[Interrupt]:
+        # Run cat /proc/interrupts. The output is of the form :
+        #          CPU0       CPU1
+        # 0:         22          0  IR-IO-APIC   2-edge      timer
+        # 1:          2          0  IR-IO-APIC   1-edge      i8042
+        # ERR:        0
+        # The first column refers to the IRQ number. The next column contains
+        # number of interrupts per IRQ for each CPU in the system. The remaining
+        # column report the metadata about interrupts, including type of interrupt,
+        # device etc. This is variable for each distro.
+        # Note : Some IRQ numbers have single entry because they're not actually
+        # CPU stats, but events count belonging to the IO-APIC controller. For
+        # example, `ERR` is incremented in the case of errors in the IO-APIC bus.
+        result = node.tools[Cat].run("/proc/interrupts", sudo=True).stdout
+        mappings_with_header = result.splitlines(keepends=False)
+        mappings = mappings_with_header[1:]
+        assert len(mappings) > 0
+
+        interrupts = []
+        for line in mappings:
+            matched = Interrupt._interrupt_regex.fullmatch(line)
+            assert matched
+            interrupt_count = [
+                int(count) for count in matched.group("interrupt_count").split()
+            ]
+            interrupts.append(
+                Interrupt(
+                    irq_number=matched.group("irq_number"),
+                    interrupt_count=interrupt_count,
+                    metadata=matched.group("metadata"),
+                )
+            )
+        return interrupts
 
 
 @testsuite.TestSuiteMetadata(
@@ -29,15 +101,6 @@ class CPUState:
     """,
 )
 class CPU(testsuite.TestSuite):
-    def _get_cpu_config_file(self, cpu_id: str) -> str:
-        return f"/sys/devices/system/cpu/cpu{cpu_id}/online"
-
-    def _set_cpu_state(self, cpu_id: str, state: str, node: Node) -> bool:
-        file_path = self._get_cpu_config_file(cpu_id)
-        node.tools[Echo].write_to_file(state, file_path, sudo=True)
-        result = node.tools[Cat].read_from_file(file_path, force_run=True, sudo=True)
-        return result == state
-
     @testsuite.TestCaseMetadata(
         description="""
             This test will check that CPU assigned to lsvmbus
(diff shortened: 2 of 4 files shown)


lisa/base_tools/cat.py
```diff
@@ -22,3 +52,37 @@ def read_from_file(
         result = self.run(file, force_run=force_run, sudo=sudo)
         result.assert_exit_code(message=f"Error : {result.stdout}")
         return result.stdout
+
+    def get_interrupt_data(self) -> List[Interrupt]:
```
squirrelsc: It's hard to guess that interrupt can be queried from the Cat tool. How about move the logic into cpu test suite? It can be a method, pass in the node, and output the Interrupts.

lisa/base_tools/cat.py
```diff
@@ -22,3 +52,37 @@ def read_from_file(
...
+        # Run cat /proc/interrupts. The output is of the form :
+        #          CPU0       CPU1
+        # 0:         22          0  IR-IO-APIC   2-edge      timer
```
squirrelsc: Copy the example above the regexp pattern. It would be easy to maintain.

microsoft/testsuites/core/cpu.py
```diff
@@ -12,14 +12,17 @@
...
-from lisa.util import BadEnvironmentStateException, SkippedException
+from lisa.util import BadEnvironmentStateException, PassedException, SkippedException
 
 
 class CPUState:
     OFFLINE: str = "0"
     ONLINE: str = "1"
 
 
+hyperv_interrupt_substr = ["hyperv", "Hypervisor", "Hyper-V"]
```
squirrelsc, line 29: What's the reason it checks three strings, instead of one of them?

microsoft/testsuites/core/cpu.py
```diff
@@ -129,3 +132,63 @@ def l3_cache_check(self, node: Node, log: Logger) -> None:
...
+                is_cpu_handling_interrupt = [
+                    ((interrupt.interrupt_count[i] > 0) | is_cpu_handling_interrupt[i])
+                    for i in range(len(interrupt.interrupt_count))
+                ]
+
+        log.debug(f"CPU interrupt handling : {is_cpu_handling_interrupt}")
+
+        # It is not mandatory to have the Hyper-V interrupts present under
+        # `/proc/interrupts`. Skip test execution if these are not showing up
+        if not found_hyperv_interrupt:
```
squirrelsc, line 332: The test logic looks strange to me. If some of interrupts are not found, it fails. But if none of them found, it's skipped. Is there some way to check it more strictly? It looks need to differentiate the real skipped from all of interrupts are not queried unexpectedly?

microsoft/testsuites/core/cpu.py
dsrivastavv: We are only checking hyper-v interrupts here. The hyper-v interrupts contain one of these three substrings in the metadata. For example on ubuntu : 
```
24:        446          0  Hyper-V PCIe MSI 1476427776-edge      mlx4-async@pci:e16b:00:02.0
25:          0        244  Hyper-V PCIe MSI 1476427777-edge      mlx4-1@e16b:00:02.0
HYP:       5912        432   Hypervisor callback interrupts
HRE:          0          0   Hyper-V reenlightenment interrupts
 HVS:       4155       4142   Hyper-V stimer0 interrupts
```

microsoft/testsuites/core/cpu.py
dsrivastavv: We are only checking hyper-v interrupts here. Based on the comments in the v2 test file, hyper-v interrupts might not be recorded for every distro, in which case we skip the test. However, if the interrupts are recorded, they should be handled by atleast one CPU.

microsoft/testsuites/core/cpu.py
squirrelsc: cool, let's add it to comments.

microsoft/testsuites/core/cpu.py
```diff
@@ -129,3 +132,63 @@ def l3_cache_check(self, node: Node, log: Logger) -> None:
...
+            interrupts, Hyper-V reenlightenment interrupts, and Hyper-V stimer0
+            interrupts. We also do not need to have each CPU handle vmbus message
+            or event in multi CPUs VM.
+            """,
+        priority=2,
+    )
+    def verify_vmbus_interrupts(self, node: Node, log: Logger) -> None:
+        found_hyperv_interrupt = False
+        cpu_count = node.tools[Lscpu].get_core_count()
+        is_cpu_handling_interrupt = [False] * cpu_count
```
squirrelsc: If a variable starts with `is_`, it means it's a boolean. The boolean array should avoid to start with `is_`.

microsoft/testsuites/core/cpu.py
```diff
@@ -129,3 +132,63 @@ def l3_cache_check(self, node: Node, log: Logger) -> None:
...
+                [(substr in interrupt.metadata) for substr in hyperv_interrupt_substr]
+            )
+            found_hyperv_interrupt = found_hyperv_interrupt | is_hyperv_interrupt
+            if is_hyperv_interrupt:
+                log.debug(f"Hyper-V interrupt : {interrupt}")
+                assert_that(
+                    len(interrupt.interrupt_count),
+                    "Interrupt count should be present for each cpu.",
+                ).is_equal_to(cpu_count)
+                is_cpu_handling_interrupt = [
```
squirrelsc: The logic means **any** of interrupt should have the number of cpu count. Is it correct?

microsoft/testsuites/core/cpu.py
```diff
@@ -129,3 +132,63 @@ def l3_cache_check(self, node: Node, log: Logger) -> None:
...
+        log.debug(f"CPU interrupt handling : {is_cpu_handling_interrupt}")
+
+        # It is not mandatory to have the Hyper-V interrupts present under
+        # `/proc/interrupts`. Skip test execution if these are not showing up
+        if not found_hyperv_interrupt:
+            raise SkippedException("Hyper-V interrupts are not recorded, abort test.")
+
+        # Ensure that atleast one CPU is handling hyper-v interrupts
+        is_any_cpu_handling_interrupt = any(is_cpu_handling_interrupt)
+        assert_that(
```
squirrelsc: This assertion looks not useful. If some of interrupt doesn't cover **all** cpus, the case should be failed `is_equal_to(cpu_count)` in the loop. Either no interrupt found, or all of them handled, when the code reaches here.

microsoft/testsuites/core/cpu.py
```diff
@@ -129,3 +132,63 @@ def l3_cache_check(self, node: Node, log: Logger) -> None:
...
+            is_any_cpu_handling_interrupt, "No CPU core is processing VMBUS interrupts!"
+        )
+
+        # Pass the test with warning if not all CPU's are processing interrupts
+        if sum(is_cpu_handling_interrupt) < cpu_count:
+            raise PassedException(
```
squirrelsc: It's better to specify what's the "big size", and partial pass only if the cpu core count is bigger than it.

microsoft/testsuites/core/cpu.py
dsrivastavv: Not every interrupt will have have entries equal to cpu count number. I had mentioned it in the comments : 
```
# Note : Some IRQ numbers have single entry because they're not actually
        # CPU stats, but events count belonging to the IO-APIC controller. For
        # example, `ERR` is incremented in the case of errors in the IO-APIC bus.
```

dsrivastavv: I will modify the comment to make it a explicit requirement for hyper-v interrupts.

microsoft/testsuites/core/cpu.py
dsrivastavv: The v2 tests defined it as 4. However, the v2 test doesn't fail if cpus are less than 4 and `all` cpus don't handle the interrupt. Also, they provide no reasoning on how they came-up with this number.  Anyone I can reach out to get clarification on this?

microsoft/testsuites/core/cpu.py
dsrivastavv: Sorry I didn't understand this comment. The test is trying to check that atleast one of the cpus is handling the interrupts. Can you please clarify this?

microsoft/testsuites/core/cpu.py
dsrivastavv: Based on offline discussion, we will set the state to failed.

lisa/tools/taskset.py
```diff
@@ -0,0 +1,18 @@
...
+class TaskSet(Tool):
+    @property
+    def command(self) -> str:
+        return "taskset"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def run_subprocess_on_cpu(self, cpu_id: int) -> Process:
```
squirrelsc: how about `run_on_specific_cpu`?

microsoft/testsuites/core/cpu.py
```diff
@@ -1,26 +1,55 @@
...
-    BadEnvironmentStateException,
-    LisaException,
-    Logger,
-    Node,
-    SkippedException,
-    testsuite,
-)
-from lisa.tools import Cat, Echo, Lscpu, Lsvmbus, Uname
+from lisa import Logger, testsuite
+from lisa.base_tools.cat import Cat
```
squirrelsc: import all tools from lisa.tools, it doesn't need so complex.

microsoft/testsuites/core/cpu.py
```diff
@@ -1,26 +1,55 @@
...
+from lisa.base_tools.uname import Uname
+from lisa.node import Node
+from lisa.tools.echo import Echo
+from lisa.tools.lscpu import Lscpu
+from lisa.tools.lsvmbus import Lsvmbus
+from lisa.tools.taskset import TaskSet
+from lisa.util import BadEnvironmentStateException, LisaException, SkippedException
```
squirrelsc: import them from lisa

microsoft/testsuites/core/cpu.py
```diff
@@ -38,6 +72,52 @@ def _set_cpu_state(self, cpu_id: str, state: str, node: Node) -> bool:
         result = node.tools[Cat].read_from_file(file_path, force_run=True, sudo=True)
         return result == state
 
+    def _get_interrupt_data(self, node: Node) -> List[Interrupt]:
```
squirrelsc: This logic should be in class `Interrupt`. It's a better envelopes.

squirrelsc: The command can be called out of the `Interrupt`, but the parsing logic is in the class.

microsoft/testsuites/core/cpu.py
squirrelsc: I forgot what we discussed. If the state should be "failed", should it raise LisaException instead of SkippedException?

microsoft/testsuites/core/cpu.py
```diff
@@ -1,26 +1,55 @@
...
 
 class CPUState:
     OFFLINE: str = "0"
     ONLINE: str = "1"
 
 
+hyperv_interrupt_substr = ["hyperv", "Hypervisor", "Hyper-V"]
+
+
+class Interrupt:
```
squirrelsc, line 33: Add dataclass decorator. If not, the members are class variable, and can only be shared cross classes.

microsoft/testsuites/core/cpu.py
```diff
@@ -38,6 +72,52 @@ def _set_cpu_state(self, cpu_id: str, state: str, node: Node) -> bool:
...
+            interrupts.append(
+                Interrupt(
+                    irq_number=matched.group("irq_number"),
+                    interrupt_count=interrupt_count,
+                    metadata=matched.group("metadata"),
+                )
+            )
+        return interrupts
+
+    def _create_stimer_interrupts(self, node: Node, cpu_count: int) -> None:
```
squirrelsc: put private methods after public methods.

microsoft/testsuites/core/cpu.py
```diff
@@ -177,3 +257,95 @@ def cpu_count_check(self, node: Node, log: Logger) -> None:
...
+            if not is_hyperv_interrupt:
+                continue
+            log.debug(f"Processing Hyper-V interrupt : {interrupt}")
+            assert_that(
+                len(interrupt.interrupt_count),
+                "Interrupt count should be present for each cpu.",
+            ).is_equal_to(cpu_count)
+            if interrupt.irq_number == "HRE" or "reenlightenment" in interrupt.metadata:
+                assert_that(
+                    all(
```
squirrelsc, line 312: It looks the assertion logic is the same, only message is different. So the logic can be moved after the logic. The logic only set messages.

microsoft/testsuites/core/cpu.py
dsrivastavv: What would be the best place to put this class? I am thinking of lisa/util ? Any suggestions?

microsoft/testsuites/core/cpu.py
dsrivastavv: iirc we wanted to mark the test as failed if some distro doesn't have any hyper-v interrupt present and understand which distros show such behavior. I will change skip to failed.

microsoft/testsuites/core/cpu.py
squirrelsc: This class can be here. We can review it later, if it's reused. The `lisa/util` is for lisa only, it shouldn't have Distro specific things.

microsoft/testsuites/core/cpu.py
```diff
@@ -1,26 +1,98 @@
...
 from lisa import (
     BadEnvironmentStateException,
     LisaException,
     Logger,
-    Node,
     SkippedException,
     testsuite,
 )
-from lisa.tools import Cat, Echo, Lscpu, Lsvmbus, Uname
+from lisa.node import Node
```
squirrelsc: Node is at lisa level too.

squirrelsc (dismissed):

microsoft/testsuites/core/cpu.py
dsrivastavv: seems like auto-import feature in VS code is importing from the files directly.

squirrelsc (approved):

PR #1496, opened 2021-08-25
add optional type parameter for tools.Mount

mcgov:
mount tool currently doesn't support the 'type' flag on linux. Added some code to allow this optional parameter

### DIFF
--- a/lisa/tools/mount.py
+++ b/lisa/tools/mount.py
@@ -23,15 +23,21 @@ class Mount(Tool):
     def can_install(self) -> bool:
         return True
 
-    def mount(self, disk_name: str, point: str) -> None:
+    def mount(self, disk_name: str, point: str, type: str = "") -> None:
         self.node.shell.mkdir(PurePosixPath(point), exist_ok=True)
+        if type:
+            type = f"-t {type}"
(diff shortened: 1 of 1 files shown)


lisa/tools/mount.py
```diff
@@ -23,15 +23,21 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
-    def mount(self, disk_name: str, point: str) -> None:
+    def mount(self, disk_name: str, point: str, type: str = "") -> None:
         self.node.shell.mkdir(PurePosixPath(point), exist_ok=True)
+        if type:
+            type = f"-t {type}"
```
squirrelsc, line 29: the extra whitespace can be in this string, so the final command without type can have one whitespace too. It  doesn't matter, but better looking. :)
`type = f"-t {type} "`
----------------------^-

squirrelsc (approved):

lisa/tools/mount.py
mcgov: Ooo but then the format string will look weird {type}{disk_name}
I am ok with potentially burning the extra byte if you are

PR #1500, opened 2021-08-26
Add options to mount

mcgov:
Mount has an options flag, add it to mount tool

### DIFF
--- a/lisa/tools/mount.py
+++ b/lisa/tools/mount.py
@@ -27,13 +27,13 @@ class Mount(Tool):
         self, disk_name: str, point: str, type: str = "", options: str = ""
     ) -> None:
         self.node.shell.mkdir(PurePosixPath(point), exist_ok=True)
+        runline = [self.command]
         if type:
-            type = f"-t {type}"
+            runline.append(f"-t {type}")
         if options:
-            options = f"-o {options}"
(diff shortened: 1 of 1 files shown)


lisa/tools/mount.py
```diff
@@ -23,12 +23,16 @@ def command(self) -> str:
...
+        self, disk_name: str, point: str, type: str = "", options: str = ""
+    ) -> None:
         self.node.shell.mkdir(PurePosixPath(point), exist_ok=True)
         if type:
             type = f"-t {type}"
+        if options:
+            options = f"-o {options}"
         cmd_result = self.node.execute(
-            f"mount {type} {disk_name} {point}", shell=True, sudo=True
+            f"mount {type} {options} {disk_name} {point}", shell=True, sudo=True
```
squirrelsc: The whitespace is growing in the line. Another way to avoid it to use a List, and append necessary parts. Then use `' '.join(parts)` to connect them together.

mcgov: refactored to create a string array and join at the end

squirrelsc (approved):

squirrelsc:
Please organize the commits before merge. Make sure every commit is meaningful. You can use `git push -f` to refresh upstream commits in your branch.

PR #1506, opened 2021-08-29
Add sriov network provisioning testcase

dsrivastavv:
Add SRIOV Provisioning testcase


lisa/features/network_interface.py
```diff
@@ -0,0 +1,24 @@
...
+# Licensed under the MIT license.
+
+from typing import Type
+
+from lisa import schema
+from lisa.feature import Feature
+from lisa.schema import NetworkInterfaceOptionSettings
+
+
+class NetworkInterface(Feature):
```
squirrelsc, line 11: It looks there is no difference from Azure network interface. So it doesn't need to have an extension. The base feature can be used directly in AzurePlatform. We can add this level, when it's necessary.

lisa/schema.py
```diff
@@ -504,6 +504,87 @@ def _generate_min_capability(self, capability: Any) -> Any:
         return min_value
 
 
+@dataclass_json()
+@dataclass()
+class NetworkInterfaceOptionSettings(FeatureSettings):
+    type: str = "NetworkInterface"
+    enable_sriov: Optional[bool] = None
```
squirrelsc: I thought it again. We need a SetSpace, it should include Synthetic, Sriov. So it mean what the network interface can be. In Azure, all network interface support Synthetic, but some of them support Sriov. It means all network interface can run test cases of Synthetic. If the Synthetic is selected, the deployment should make sure the network interface is deployed as synthetic, even it support Sriov. But if nothing specified, the sriov should be enabled by default, if the network interface supports it.

lisa/schema.py
```diff
@@ -504,6 +504,87 @@ def _generate_min_capability(self, capability: Any) -> Any:
...
+    def __eq__(self, o: object) -> bool:
+        assert isinstance(o, NetworkInterfaceOptionSettings), f"actual: {type(o)}"
+        return (
+            self.type == o.type
+            and self.enable_sriov == o.enable_sriov
+            and self.nic_count == o.nic_count
+        )
+
+    def __str__(self) -> str:
+        return f"enable_sriov:{self.enable_sriov}," f"nic_count:{self.nic_count}"
```
squirrelsc: There is an extra `f` string marker, remove it.

lisa/testsuite.py
```diff
@@ -350,6 +351,11 @@ def simple_requirement(
             disk = schema.DiskOptionSettings()
         node.disk = disk
         node.disk.data_disk_count = search_space.IntRange(min=min_data_disk_count)
+    if min_nic_count or network_interface:
+        if not network_interface:
+            network_interface = schema.NetworkInterfaceOptionSettings()
+            network_interface.nic_count = search_space.IntRange(min=min_nic_count)
```
squirrelsc, line 364: This line should be moved out of the `if not network_interface`.

selftests/azure/test_prepare.py
```diff
@@ -225,6 +231,7 @@ def test_normal_req_in_same_location(self) -> None:
     def test_count_normal_cost(self) -> None:
         # predefined vm size also count the cost
         env = self.load_environment(node_req_count=2)
+        print(env.capability)
```
squirrelsc: Remove the test code.

selftests/test_testsuite.py
```diff
@@ -503,6 +503,7 @@ def test_skipped_not_meet_req(self) -> None:
             result.check_results.reasons,
         )
         result = self.case_results[1]
+        print(self.default_env.capability)
```
squirrelsc: remove it.

dsrivastavv: I am still iterating on this PR. But I will mark the PR as reviewable once I push final changes.

lisa/testsuite.py
dsrivastavv: Wouldn't this always overwrite it to `min_nic_count`? What if network_interface.nic_count is already set in testcasemetadata?

squirrelsc: If the min_nic_count is specified, it's priority is higher. If it's not necessary, it can be removed from parameters. The behavior is the same as disk.count. BTW, the variable should be named as `count`. The `data_disk_count` is data disk, so it needs to specify that.

lisa/schema.py
```diff
@@ -504,6 +504,107 @@ def _generate_min_capability(self, capability: Any) -> Any:
...
+    Sriov = "Sriov"
+
+
+_network_data_path_priority: List[NetworkDataPath] = [
+    NetworkDataPath.Sriov,
+    NetworkDataPath.Synthetic,
+]
+
+
+def _decode_network_data_path_type(data: Any) -> Any:
```
squirrelsc: This logic can be shared by using generic type. The name can be `decode_set_space_by_type`.

squirrelsc: You can check from the implementation of factory.

squirrelsc (dismissed):

squirrelsc (approved):

PR #1507, opened 2021-08-31
Get distro info from /etc/os-release from Debian distro.

LiliDeng:
By default, lsb_release is not installed in Debian but installed in Ubuntu.
Get release version from /etc/debian_version when it is debian distro.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -228,7 +228,7 @@ class Windows(OperatingSystem):
 
 
 class Posix(OperatingSystem, BaseClassMixin):
-    __os_info_pattern = re.compile(
+    _os_info_pattern = re.compile(
         r"^(?P<name>.*)=[\"\']?(?P<value>.*?)[\"\']?$", re.MULTILINE
     )
     # output of /etc/fedora-release - Fedora release 22 (Twenty Two)
@@ -243,7 +243,7 @@ class Posix(OperatingSystem, BaseClassMixin):
     #   Description:	Scientific Linux release 6.7 (Carbon)
     # In most of the distros, the text in the brackets is the codename.
     # This regex gets the codename for the ditsro
-    __distro_codename_pattern = re.compile(r"^.*\(([^)]+)")
+    _distro_codename_pattern = re.compile(r"^.*\(([^)]+)")
 
     def __init__(self, node: Any) -> None:
         super().__init__(node, is_posix=True)
@@ -324,7 +324,7 @@ class Posix(OperatingSystem, BaseClassMixin):
         codename: str = ""
         full_version: str = ""
         for row in cmd_result.stdout.splitlines():
-            os_release_info = self.__os_info_pattern.match(row)
+            os_release_info = self._os_info_pattern.match(row)
             if not os_release_info:
                 continue
             if os_release_info.group("name") == "NAME":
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -330,7 +330,16 @@ def _get_information(self) -> OsInformation:
             if os_release_info.group("name") == "NAME":
                 vendor = os_release_info.group("value")
             elif os_release_info.group("name") == "VERSION_ID":
-                release = os_release_info.group("value")
+                # get release from /etc/debian_version for debian distro
+                # e.g. credativ debian 9-backports 9.20190313.0
+                # cat /etc/debian_version - 9.8
+                # VERSION_ID="9"
+                if "Debian" in vendor:
```
squirrelsc: The logic should be in Debian, instead of posix.

LiliDeng: Fixed

lisa/operating_system.py
```diff
@@ -504,26 +500,40 @@ def _package_exists(self, package: str, signed: bool = True) -> bool:
...
-            cmd="lsb_release -a", shell=True, no_error_log=True
-        )
+        # try to set version info from /etc/os-release.
+        cat = self._node.tools[Cat]
+        cmd_result = cat.run("/etc/os-release")
         cmd_result.assert_exit_code(message="error on get os information")
-        assert cmd_result.stdout, "not found os information from 'lsb_release -a'"
 
+        vendor: str = ""
+        release: str = ""
```
sharsonia, line 509: I don't think we need to initialize these to empty string. They are already initialized to empty string in the OsInformation object.

lisa/operating_system.py
```diff
@@ -504,26 +500,40 @@ def _package_exists(self, package: str, signed: bool = True) -> bool:
         return False
 
     def _get_information(self) -> OsInformation:
-        cmd_result = self._node.execute(
-            cmd="lsb_release -a", shell=True, no_error_log=True
-        )
+        # try to set version info from /etc/os-release.
```
sharsonia, line 503: wait, don't we already get the version and rest of the info from /etc/os-release in Posix class as well?
Why do we need it here?

LiliDeng: the release is integer from /etc/os-release, e.g. debian debian-10 10-backports-gen2 0.20210201.535, release -> 10
we will see exception 'failed. LisaException: The version is invalid format: 10'

Then we get the release from /etc/debian_version for debian, that's why I overwrite it.

lisa/operating_system.py
```diff
@@ -504,26 +500,40 @@ def _package_exists(self, package: str, signed: bool = True) -> bool:
...
+            elif os_release_info.group("name") == "VERSION":
+                codename = get_matched_str(
+                    os_release_info.group("value"),
+                    super()._distro_codename_pattern,
+                )
+            elif os_release_info.group("name") == "PRETTY_NAME":
+                full_version = os_release_info.group("value")
+
+        # get release from /etc/debian_version for debian distro
+        # credativ debian 9-backports 9.20190313.0
```
squirrelsc: Add more explanation, so people know "credativ debian 9-backports 9.20190313.0" is a marketplace image name, and what's the raw output line.

squirrelsc (dismissed):

lisa/operating_system.py

squirrelsc (approved):

lisa/operating_system.py
sharsonia: I see. Can we somehow re-use the code snippet used in Posix class to get the os information from /etc/os-release and then just replace the version part from /etc/debian_version?

sharsonia (approved):

lisa/operating_system.py
LiliDeng: I'll refine it later.

PR #1516, opened 2021-09-02
Refactor resource group creation

dsrivastavv:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -281,6 +282,14 @@ def get_storage_client(
     )
 
 
+def get_resource_management_client(
+    credential: Any, subscription_id: str
+) -> ResourceManagementClient:
+    return ResourceManagementClient(
```
squirrelsc, line 287: This is a one line method. What's the reason, it needs to be a method?

dsrivastavv: I was following the pattern similar to methods such as `get_storage_client`, `get_network_client` etc. which are defined in common.py. I assume these methods have been encapsulated to hide the internal initialization details as they are used across the project.

squirrelsc (approved):

PR #1520, opened 2021-09-03
lsmod.py remove parameter

kamalca:
lsmod tool was failing with exit code 1 because it was running `lsmod [mod_name]`. However, usage for lsmod is simply `lsmod`

This tool is utilized by the gpu feature, so we may want to ensure the gpu feature and gpu testsuit are functioning properly.


squirrelsc:
Thank you @kamalca , Yi fixed it in #1519, I just merged it.

PR #1522, opened 2021-09-05
Add test case verify_deployment_provision_synthetic_nic

LiliDeng:
By default smoke_test will use sriov if the size supports. Should we change smoke_test use Synthetic network? @squirrelsc @somil55

### DIFF
--- a/microsoft/testsuites/core/provisioning.py
+++ b/microsoft/testsuites/core/provisioning.py
@@ -22,7 +22,7 @@ from lisa.features import (
     DiskStandardSSDLRS,
     SerialConsole,
 )
-from lisa.features.network_interface import Sriov
+from lisa.features.network_interface import Sriov, Synthetic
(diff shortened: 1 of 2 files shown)


dsrivastavv:
I think the answer to this would be yes if SRIOV is not frequently used on VM's.

microsoft/testsuites/core/provisioning.py
```diff
@@ -57,6 +57,7 @@ class Provisioning(TestSuite):
         priority=0,
         requirement=simple_requirement(
             environment_status=EnvironmentStatus.Deployed,
+            network_interface=Synthetic,
```
squirrelsc: We can add another test case for Synthetic, the smoke test use the default behavior. It means to enable SRIOV, if it supports.

LiliDeng: thanks, updated

squirrelsc (approved):

PR #1524, opened 2021-09-05
Add test case cpu_count_check.

LiliDeng:
```
2021-09-05 10:27:26.483[10764][INFO] lisa.RootRunner                                CPU.cpu_count_check: PASSED   
```

### DIFF
--- a/lisa/tools/lscpu.py
+++ b/lisa/tools/lscpu.py
@@ -51,7 +51,10 @@ class CPUInfo:
 
 class Lscpu(Tool):
     # CPU(s):              16
-    __vcpu_sockets = re.compile(r"^CPU\(s\):[ ]+([\d]+)\r?$", re.M)
+    __vcpu = re.compile(r"^CPU\(s\):[ ]+([\d]+)\r?$", re.M)
+    __thread_per_core = re.compile(r"^Thread\(s\) per core:[ ]+([\d]+)\r?$", re.M)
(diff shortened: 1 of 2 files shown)


lisa/tools/lscpu.py
```diff
@@ -104,6 +107,36 @@ def get_core_count(self, force_run: bool = False) -> int:
 
         return self._core_count
 
+    def get_thread_per_core_count(self, force_run: bool = False) -> int:
```
squirrelsc, line 110: What's the reason it needs the force run? The value should be fixed, is it right?

LiliDeng: if it is used in resize vm size, then it will be different.

microsoft/testsuites/core/cpu.py
```diff
@@ -148,3 +148,32 @@ def l3_cache_check(self, node: Node, log: Logger) -> None:
                 "L3 cache of each core must be mapped to the NUMA node "
                 "associated with the core.",
             ).is_equal_to(cpu.numa_node)
+
+    @testsuite.TestCaseMetadata(
+        description="""
+            This test will check that vCPU count correctness.
+            Steps :
+            1. Get vCPU count.
+            2. Caculate vCPU count by core_per_socket_count * socket_count *
```
dsrivastavv: [nit] calculate

LiliDeng: Fixed

squirrelsc (approved):

Issue #1511, opened 2021-09-01
Command not found (PATH does not contain /usr/sbin)


kamalca:
Getting errors when using LISAv3 to deploy and test CentOS 7_9 on Azure

`[ERROR] lisa.env[generated_0].node[0].cmd[7289] not found command: Command not found: modinfo. Check that modinfo is installed and on $PATH`

`[ERROR] lisa.env[generated_0].node[0].cmd[1038] not found command: Command not found: waagent. Check that waagent is installed and on $PATH`

`[ERROR] lisa.env[generated_0].node[0].cmd[8629] not found command: Command not found: lsmod. Check that lsmod is installed and on $PATH`

SSHing into the node confirms that all three of these commands are present and runnable on the node.

The error about modinfo missing appears to occur before any tests start running. These errors do not occur when deploying and testing Ubuntu 18.04-LTS.

squirrelsc:
@LiliDeng can you help on looking it?

LiliDeng:
Ack

squirrelsc:
@LiliDeng Those commands fallback to use sudo automatically. It needs to set no error level and info level log. If it's a tool, it should be handled at command detection phase, please check why it doesn't work.

```log
2021-09-07 22:37:21.095[3488][DEBUG] lisa.env[generated_0].node[0].cmd[2764] cmd: ['waagent', '-version'], cwd: None, shell: False, sudo: False, posix: True, remote: True
2021-09-07 22:37:21.178[3488][ERROR] lisa.env[generated_0].node[0].cmd[2764] not found command: Command not found: waagent. Check that waagent is installed and on $PATH
2021-09-07 22:37:21.178[3488][DEBUG] lisa.env[generated_0].node[0].cmd[3786] cmd: ['/usr/sbin/waagent', '-version'], cwd: None, shell: False, sudo: False, posix: True, remote: True
2021-09-07 22:37:21.440[15676][DEBUG] lisa.env[generated_0].node[0].cmd[3786].stdout WALinuxAgent-2.2.46 running on centos 7.9.2009
```

kamalca:
I do not understand why using sudo would be the solution here. waagent and modinfo should not require sudo and from the logs the fallback that works also doesn't use sudo
`cmd: ['/usr/sbin/waagent', '-version'], cwd: None, shell: False, sudo: False, posix: True, remote: True`
It simply uses the absolute path for the command since the PATH variable is not being properly set on CentOS.

[...]

PR #1525, opened 2021-09-07
New Test: check_floppy_module

kamalca:
Wrote check_floppy_module which uses modprobe to ensure that the floppy module is not loaded in the kernel.

Motivation:
* Ubuntu was having problems caused by the floppy driver being enabled (I do not know the details. More information about previous problems may be required). Since the module is not necessary on Azure, we want to ensure it is disabled on all other distros.
* On-demand loading of the module is outside the scope of this test

Considerations:
* Could check dmesg for 'I/O error dev fd0 sector 0'
* May want to add a check for module on blacklist.

Absolute path for modprobe must be used to bypass PATH issue #1511 on CentOS and Redhat

### DIFF
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -1,0 +1,45 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from typing import Any
+
+from lisa.executable import Tool
+
+
+class Modprobe(Tool):
+    @property
+    def command(self) -> str:
+        return self._command
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._command = "modprobe"
+
+    def is_module_loaded(
+        self,
+        mod_name: str,
+        force_run: bool = False,
+        no_info_log: bool = True,
+        no_error_log: bool = True,
+    ) -> bool:
+        result = self.run(
+            f"-nv {mod_name}",
+            force_run=force_run,
+            no_info_log=no_info_log,
+            no_error_log=no_error_log,
+        )
+        # Example possible outputs
+        # 1) insmod /lib/modules/.../floppy.ko.xz
+        #    Exists but is not loaded - return False
+        # 2) FATAL: Module floppy not found.
+        #    Module does not exist, therefore is not loaded - return False
+        # 3) (no output)
+        #    Module is loaded - return True
+        could_be_loaded = result.stdout and "insmod" in result.stdout
+        does_not_exist = (result.stderr and "not found" in result.stderr) or (
+            result.stdout and "not found" in result.stdout
+        )
+
+        return not (could_be_loaded or does_not_exist)
--- a/microsoft/testsuites/core/floppy.py
+++ b/microsoft/testsuites/core/floppy.py
@@ -1,0 +1,38 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from assertpy import assert_that
+
+from lisa import RemoteNode, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa.tools import Modprobe
+
+
+@TestSuiteMetadata(
+    area="core",
+    category="functional",
+    description="""
+    This test suite ensures the floppy driver is disabled.
+    The floppy driver is not needed on Azure and
+    is known to cause problems in some scenarios.
+    """,
+)
+class Floppy(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        The goal of this test is to ensure the floppy module is not enabled
+        for images used on the Azure platform.
+        This test case will
+        1. Dry-run modprobe to see if floppy module can be loaded
+        2. If "insmod" would be executed then the module is not already loaded
+        3. If module cannot be found then it is not loaded
+        If the module is loaded, running modprobe will have no output
+        """,
+        priority=1,
+    )
+    def check_floppy_module(self, node: RemoteNode) -> None:
+        modprobe = node.tools[Modprobe]
+
+        assert_that(modprobe.is_module_loaded("floppy")).described_as(
+            "The floppy module should not be loaded. "
+            "Try adding the module to the blacklist."
+        ).is_false()
--- a/lisa/tools/__init__.py
+++ b/lisa/tools/__init__.py
@@ -20,6 +20,7 @@ from .lsvmbus import Lsvmbus
 from .make import Make
 from .mkfs import Mkfsext, Mkfsxfs
 from .modinfo import Modinfo
+from .modprobe import Modprobe
 from .mount import Mount
 from .ntp import Ntp
 from .ntpstat import Ntpstat
@@ -51,6 +52,7 @@ __all__ = [
     "Mkfsext",
     "Mkfsxfs",
     "Modinfo",
+    "Modprobe",
     "Mount",
     "Ntp",
     "Ntpstat",


microsoft/testsuites/core/floppy.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@TestSuiteMetadata(
+    area="core",
+    category="functional",
+    description="""
+    This test suite ensures the floppy module is disabled.
+    """,
+)
+class Floppy(TestSuite):
```
LiliDeng, line 19: we can make a modules test suites, besides check floppy module, we have other modules related case can be added.

microsoft/testsuites/core/floppy.py
```diff
@@ -0,0 +1,28 @@
...
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Ensure the floppy module is not loaded.
+        """,
+        priority=0,
+    )
+    def check_floppy_module(self, node: RemoteNode) -> None:
+        cmd_result = node.execute("/usr/sbin/lsmod | /usr/bin/grep floppy", shell=True)
```
LiliDeng: use tool node.tools[Lsmod]

LiliDeng: here is behavior in centos79.
it is built as module.
cat /boot/config-3.10.0-1160.31.1.el7.x86_64 | grep -i CONFIG_BLK_DEV_FD
CONFIG_BLK_DEV_FD=m

We ask distro vendor to blacklist it.
cat /etc/modprobe.d/* | grep -i blacklist
blacklist amdgpu
**blacklist floppy**
blacklist nouveau
blacklist radeon
blacklist dccp
blacklist dccp_diag
blacklist dccp_ipv4
blacklist dccp_ipv6

microsoft/testsuites/core/floppy.py
```diff
@@ -0,0 +1,28 @@
...
+from assertpy import assert_that
+
+from lisa import RemoteNode, TestCaseMetadata, TestSuite, TestSuiteMetadata
+
+
+@TestSuiteMetadata(
+    area="core",
+    category="functional",
+    description="""
+    This test suite ensures the floppy module is disabled.
```
squirrelsc: Add reference to let users know, why the floppy module needs to be disabled. And it's good to introduce how to detect the floppy is disabled, like from lsmod.

microsoft/testsuites/core/floppy.py
squirrelsc: we should check it in Python, so avoid to use `|` in code, unless the output is over 1000 lines. It brings raw output to log, and is good for troubleshooting.

kamalca: I am using Lsmod tool now that bugs in the tool have been resolved.

microsoft/testsuites/core/floppy.py
```diff
@@ -0,0 +1,46 @@
...
+        config_path = f"/boot/config-{uname}"
+
+        assert_that(
+            node.execute(f"grep CONFIG_BLK_DEV_FD=y {config_path}").exit_code
+        ).described_as(
+            "Floppy driver should not be installed in the kernel."
+        ).is_equal_to(
+            1
+        )
+
```
LiliDeng: it can be set, in this situation, we need the floppy driver is blacklisted.

squirrelsc:
Please make each commit is meaningful, so it can be merged by rebase.

microsoft/testsuites/core/floppy.py
```diff
@@ -0,0 +1,46 @@
...
+        This test case will
+        1. Ensure the floppy driver is not loaded directly
+        into the kernel by checking boot config
+        2. Ensure the floppy driver is not loaded as a module using lsmod
+        """,
+        priority=0,
+    )
+    def check_floppy_module(self, node: RemoteNode) -> None:
+        # 1
+        uname = node.execute("uname -r").stdout
```
squirrelsc: The  ` Uname`  should have the output already. Use the tool is better.

microsoft/testsuites/core/floppy.py
```diff
@@ -20,27 +18,23 @@ class Floppy(TestSuite):
...
+        3. If module cannot be found then it is not loaded
+        If the module is loaded, running modprobe will have no output
         """,
         priority=0,
     )
     def check_floppy_module(self, node: RemoteNode) -> None:
-        # 1
-        uname = node.execute("uname -r").stdout
-        config_path = f"/boot/config-{uname}"
+        result = node.execute("modprobe -nv floppy")
```
squirrelsc: Add a tool to do it.

microsoft/testsuites/core/floppy.py
```diff
@@ -20,27 +18,23 @@ class Floppy(TestSuite):
     @TestCaseMetadata(
         description="""
         This test case will
-        1. Ensure the floppy driver is not loaded directly
-        into the kernel by checking boot config
-        2. Ensure the floppy driver is not loaded as a module using lsmod
+        1. Dry-run modprobe to see if floppy module can be loaded
+        2. If "insmod" is executed then it is not already loaded
+        3. If module cannot be found then it is not loaded
```
squirrelsc, line 27: 1. The test need to explain the test point, for example, "verify the floppy driver should be disabled in Azure"
2. Explain the reason from a link.
3. Clarify what's expected, what's not. for example, if it's "not loaded". Does it mean the distro is in good status?

microsoft/testsuites/core/floppy.py
```diff
@@ -20,27 +18,23 @@ class Floppy(TestSuite):
...
+        could_be_loaded = "insmod" in result.stdout
+        does_not_exist = "not found" in result.stderr or "not found" in result.stdout
 
-        # 2
-        lsmod = node.tools[Lsmod]
-        assert_that(lsmod.module_exists("floppy")).described_as(
-            "Floppy module should be disabled by adding it to the blacklist"
-        ).is_false()
+        if could_be_loaded or does_not_exist:
+            raise PassedException
```
squirrelsc: The logic can be modified to raise exception, if the module exists. The PassedException is for partial passed, not the normal passed.

lisa/tools/modprobe.py
```diff
@@ -0,0 +1,62 @@
...
+    def command(self) -> str:
+        return self._command
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._command = "modprobe"
+
+    def load_module(
```
squirrelsc: What's the method is used for? It does nothing almost. But if the return value can be parsed, it's good to have this method, and return structured result.

kamalca: `load_moule` runs modprobe. I have been using it for testing and figured that it should be left in the tool since it is the main functionality of modprobe. The command itself usually has no output but I thought maybe the user of the tool would want to check if the command had failed for any reason.

squirrelsc: Hmmm, I didn't see the reason to keep this method. The `is_module_loaded` is full functional. If there is any failure on modprobe itself, it should be detected by the is_module_loaded too. As a tool in LISA, the exists and installation is handled in the base classes. If there is different logic on checking existence or installation, the tool should overrides the default behavior.
In summary, it doesn't need this method to perform checks. It's the same as `run` of the tool.

microsoft/testsuites/core/floppy.py
```diff
@@ -0,0 +1,37 @@
...
+    @TestCaseMetadata(
+        description="""
+        The goal of this test is to ensure the floppy module is not enabled.
+        This test case will
+        1. Dry-run modprobe to see if floppy module can be loaded
+        2. If "insmod" would be executed then the module is not already loaded
+        3. If module cannot be found then it is not loaded
+        If the module is loaded, running modprobe will have no output
+        """,
+        priority=0,
```
squirrelsc: It should be a P1 test case. +1 from v2.

squirrelsc (approved):

PR #1526, opened 2021-09-07
Add DPDK requirements

mcgov:
I'd like to add a requirement so we don't run the dpdk tests on unsupported hosts. ~~In my testing with this runbook this code doesn't work as expected so I'm hoping to get some feedback.~~ EDIT: removing OS requirement allows core and nic count requirement to work, so I'll just remove since the unsupported OSs are skipped anyway.

```
name: dpdk
include:
  - path: ../../runbook/azure.yml
extension:
  - .
platform:
  - type: azure
    azure:
      subscription_id: $(subscription_id)
    requirement:
      core_count:
        min: 16
      nic_count: 2
      azure:
        vm_size: Standard_D15_v2
        location: westus2

testcase:
  - criteria:
      name: check_dpdk_build
```

Result:
`dpdk_build: SKIPPED  no available environment: ["os_type/capability cannot support some of requirements, requirement: 'allowed:True,items:[<class 'lisa.operating_system.Redhat'>,<class 'lisa.operating_system.Ubuntu'>]'capability: 'allowed:True,items:[<class 'lisa.util.BaseClassMixin'>,<class 'lisa.operating_system.Posix'>,<class 'lisa.operating_system.OperatingSystem'>,<class 'lisa.operating_system.Redhat'>,<class 'object'>,<class 'lisa.operating_system.Linux'>,<class 'lisa.operating_system.Fedora'>]', "]`

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -10,6 +10,7 @@ from lisa import Logger, Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
 from lisa.features import Sriov
 from lisa.testsuite import simple_requirement
 from lisa.tools import Echo, Lspci, Mount
+from lisa.operating_system import Ubuntu, Redhat
 from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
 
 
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -19,15 +20,18 @@
     description="""
     This test suite check DPDK functionality
     """,
+    requirement=simple_requirement(
+        supported_features=[Sriov],
+        min_core_count=8,
+        min_nic_count=2,
+        supported_os=[Ubuntu, Redhat],
```
squirrelsc: what's the reason to limit the distro? It blocks the test cases to run in more distros.

mcgov: Those are the only ones that are implemented at the moment, I can remove it, it will just end up skipping the tests for other distros

squirrelsc: If it's just "we" are not implemented, it shouldn't be added here. We can implement it later, when it runs against other distros. If it's "the feature (DPDK)" is not implemented on other distros, they need to be added here.

mcgov: Well, the issue still exists that adding those as a requirement breaks running images with those OSs. You can repro by making a nop test and adding the ubuntu or redhat requirement and trying to run it with either image.

I'll remove it and I'd like to accept this PR without the OS requirement since it's not... required 😄

PR #1531, opened 2021-09-10
Add enhancement into lisav2

lubaihua33:
1. Add judgement for VMData if null then throw
2.  Add retries when calling Get-AzNetworkInterface API
3. Change udp test ntttcp version to master

### DIFF
--- a/Libraries/Azure.psm1
+++ b/Libraries/Azure.psm1
@@ -2492,8 +2492,22 @@ Function Set-SRIOVinAzureVMs {
 		foreach ( $TargetVM in $TargettedVMs) {
 			$VMName = $TargetVM.RoleName
 			$ResourceGroup = $TargetVM.ResourceGroupName
-			$AllNics = Get-AzNetworkInterface -ResourceGroupName $ResourceGroup `
-			| Where-Object { $($_.VirtualMachine.Id | Split-Path -leaf) -eq $VMName }
+			# Sometimes Get-AzNetworkInterface returns null, retry 120 times
+			$retry_count = 0
+			While ($retry_count -lt 120) {
+				$AllNics = Get-AzNetworkInterface -ResourceGroupName $ResourceGroup `
+				| Where-Object { $($_.VirtualMachine.Id | Split-Path -leaf) -eq $VMName }
+				$retry_count++
+				if ($null -eq $AllNics) {
+					Start-Sleep -Seconds 1
+					Write-LogInfo "Get-AzNetworkInterface for $VMName of $ResourceGroup return $null. Retry..."
+				} else {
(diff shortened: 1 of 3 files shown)


Testscripts/Linux/utils.sh
```diff
@@ -2745,6 +2745,8 @@ function install_lagscope () {
 
 # Build and install ntttcp
 function build_ntttcp () {
+	# Clean up ntttcp before install
+	rm -rf /usr/local/bin/ntttcp
```
squirrelsc: Can we update ntttcp to use same version for udp and tcp?

Libraries/Azure.psm1
```diff
@@ -2495,6 +2495,19 @@ Function Set-SRIOVinAzureVMs {
 			$AllNics = Get-AzNetworkInterface -ResourceGroupName $ResourceGroup `
 			| Where-Object { $($_.VirtualMachine.Id | Split-Path -leaf) -eq $VMName }
 
+			# Sometimes Get-AzNetworkInterface returns null, retry 5 times
+			$retry_count = 0
+			While ($null -eq $AllNics -and $retry_count -lt 5) {
```
squirrelsc: try more times, like 120. It means two minutes.

Libraries/Azure.psm1
```diff
@@ -2495,6 +2495,19 @@ Function Set-SRIOVinAzureVMs {
...
+				$AllNics = Get-AzNetworkInterface -ResourceGroupName $ResourceGroup `
+				| Where-Object { $($_.VirtualMachine.Id | Split-Path -leaf) -eq $VMName }
+				$retry_count++
+				Start-Sleep -Seconds 1
```
squirrelsc: check if AllNics has value, don't sleep.

Testscripts/Linux/utils.sh
lubaihua33: Yes, updated in the xml file.

Libraries/Azure.psm1
lubaihua33: Done

Libraries/Azure.psm1
lubaihua33: Done

Testscripts/Linux/utils.sh
squirrelsc: Does the ntttcp of master branch support both udp and tcp now? I mean to update ntttcp, not in lisa.

lubaihua33: Yes, master branch can support udp and tcp.

squirrelsc (approved):

PR #1532, opened 2021-09-13
DPDK test class refactor

mcgov:
- add 'mode' parameter for running rx/tx/fwd test
  - hoist hugepage setup function for use in testpmd setup class
  - enable easier multinode setup with testpmd setup class
  - skip setup if the tools are already installed

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -29,44 +29,23 @@ class Dpdk(TestSuite):
              with one interface for management.
             More detailes refer https://docs.microsoft.com/en-us/azure/virtual-network/setup-dpdk#prerequisites # noqa: E501
         """,
+        priority=2,
         requirement=simple_requirement(
             min_nic_count=2,
             network_interface=Sriov,
         ),
-        priority=1,
     )
     def check_dpdk_build(self, node: Node, log: Logger) -> None:
-        network_interface_feature = node.features[NetworkInterface]
-        sriov_is_enabled = network_interface_feature.is_enabled_sriov()
-        log.info(f"Verify SRIOV is enabled: {sriov_is_enabled}")
-        assert_that(sriov_is_enabled).described_as(
-            "SRIOV was not enabled for this test node."
-        ).is_true()
-
-        # dump some info about the pci devices before we start
-        lspci = node.tools[Lspci]
-        log.info(f"LSPCI Info:\n{lspci.run().stdout}\n")
-
-        # enable hugepages (required by dpdk)
-        self._hugepages_init(node)
-        self._hugepages_enable(node, log)
-
-        # initialize testpmd tool (installs dpdk)
-        testpmd = DpdkTestpmd(node)
-        testpmd.install()
-
-        # initialize node nic info class (gathers info about nic devices)
-        node_nic_info = NodeNicInfo(node)
-        assert_that(len(node_nic_info)).described_as(
-            "Test needs at least 2 NICs on the test node."
-        ).is_greater_than_or_equal_to(2)
-        log.info(node_nic_info)
+        # setup and unwrap the resources for this test
+        test_kit = initialize_node_resources(node, log)
+        node_nic_info, testpmd = test_kit.node_nic_info, test_kit.testpmd
 
         # grab a nic and run testpmd
         test_nic = node_nic_info.get_nic(node_nic_info.get_upper_nics()[-1])
         vdev_type = "net_vdev_netvsc0"
         testpmd_include_str = test_nic.testpmd_include(vdev_type)
-        testpmd_output = testpmd.run_with_timeout(testpmd_include_str, 30)
+        testpmd_cmd = testpmd._generate_testpmd_command(testpmd_include_str, "txonly")
+        testpmd_output = testpmd.run_for_n_seconds(testpmd_cmd, 10)
         tx_pps = testpmd.get_tx_pps_from_testpmd_output(testpmd_output)
         log.info(
             f"TX-PPS:{tx_pps} from {test_nic._upper}/{test_nic._lower}:"
@@ -81,28 +60,31 @@ class Dpdk(TestSuite):
         result.assert_exit_code()
         return result.stdout
 
-    def _hugepages_init(self, node: Node) -> None:
-        mount = node.tools[Mount]
-        mount.mount(disk_name="nodev", point="/mnt/huge", type="hugetlbfs")
-        mount.mount(
-            disk_name="nodev",
-            point="/mnt/huge-1G",
-            type="hugetlbfs",
-            options="pagesize=1G",
-        )
 
-    def _hugepages_enable(self, node: Node, log: Logger) -> None:
-        echo = node.tools[Echo]
-        echo.write_to_file(
-            "1024",
-            "/sys/devices/system/node/node0/hugepages/hugepages-2048kB/nr_hugepages",
-            sudo=True,
-        )
-        echo.write_to_file(
-            "1",
-            "/sys/devices/system/node/node0/hugepages/hugepages-1048576kB/nr_hugepages",
-            sudo=True,
-        )
+def _init_hugepages(node: Node) -> None:
+    mount = node.tools[Mount]
+    mount.mount(disk_name="nodev", point="/mnt/huge", type="hugetlbfs")
+    mount.mount(
+        disk_name="nodev",
+        point="/mnt/huge-1G",
+        type="hugetlbfs",
+        options="pagesize=1G",
+    )
+    _enable_hugepages(node)
+
+
+def _enable_hugepages(node: Node) -> None:
+    echo = node.tools[Echo]
+    echo.write_to_file(
+        "1024",
+        "/sys/devices/system/node/node0/hugepages/hugepages-2048kB/nr_hugepages",
+        sudo=True,
+    )
+    echo.write_to_file(
+        "1",
+        "/sys/devices/system/node/node0/hugepages/hugepages-1048576kB/nr_hugepages",
+        sudo=True,
+    )
 
 
 class NicInfo:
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -6,8 +6,16 @@
...
+from lisa import (
+    Logger,
+    Node,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    Environment,
+)
+from lisa.features import Sriov
+from lisa.features.network_interface import NetworkInterface
```
squirrelsc: reference it from `lisa.features` too.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -29,44 +37,24 @@ class Dpdk(TestSuite):
              with one interface for management.
             More detailes refer https://docs.microsoft.com/en-us/azure/virtual-network/setup-dpdk#prerequisites # noqa: E501
         """,
+        priority=1,
         requirement=simple_requirement(
+            min_core_count=8,
```
squirrelsc: What's the reason to add this condition?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -81,28 +69,75 @@ def _execute_expect_zero(self, node: Node, cmd: str) -> str:
...
+    echo.write_to_file(
+        "1",
+        "/sys/devices/system/node/node0/hugepages/hugepages-1048576kB/nr_hugepages",
+        sudo=True,
+    )
+
+
+class DpdkTestKit:
+    def __init__(self, node: Node, log: Logger) -> None:
+        network_interface_feature = node.features[NetworkInterface]
```
squirrelsc: Don't put the long time logic in construction method like `__init__`.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: DPDK requires at least 8 cores

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -81,28 +68,75 @@ def _execute_expect_zero(self, node: Node, cmd: str) -> str:
...
+            network_interface=Sriov,
+            min_count=3,
+        ),
+    )
+    def check_3node_forwarding(self, environment: Environment, log: Logger) -> None:
+        # TODO: stub test, just submitting refactoring PR
+        pass
+
+
+def hugepages_init(node: Node) -> None:
```
squirrelsc: They should be private method, and start with `_`

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -194,17 +197,17 @@ def _generate_testpmd_command(self, nic_to_include: str) -> str:
...
             f"{self._testpmd_install_path} -l 0-1 -n 4 --proc-type=primary "
-            + f"{nic_to_include} -- --forward-mode=txonly -a --stats-period 1"
+            + f"{nic_to_include} -- --forward-mode={mode} -a --stats-period 1"
         )
 
-    def run_with_timeout(self, nic_to_include: str, timeout: int) -> str:
+    def run_with_timeout(self, cmd: str, timeout: int) -> str:
         timer_proc = self.node.execute_async(
-            f"sleep {timeout} && killall -s INT {self._testpmd_install_path}",
+            f"sleep {timeout} && killall -s INT {cmd.split()[0]}",
```
squirrelsc, line 205: Just noticed this logic about timeout. The `wait_result` has a timeout method. Does it work here?

squirrelsc: I'm not sure if the timeout will wait until it reaches every time. If so, it's low efficient.

mcgov: It does not. Killing the program correctly requires the INT signal in this case.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Ah! I'm glad you asked. I had it in my head that azure required 8 cores for DPDK w multiple nics. Now that I am double checking the requirements are more complex that I remember, so I'll fix that up and submit changes

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Not clear on what the change needs to be, you want things that will run for a while in a separate function?
kit = TestKit()
change to like 
kit = TestKit()
kit.init_long_running_stuff()

microsoft/testsuites/dpdk/dpdktestpmd.py
squirrelsc: Cab you add a method in process class to `send_signal` to process? There is such a method in internal process classes, but not exposed.
How about the second question about the timeout? Does it make sure the process won't sleep `timeout` seconds every time?

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: Yes, put it in a separated method. Python may not be the same, but in most OO languages, it should avoid to have logic in construction.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: > I'm not sure if the timeout will wait until it reaches every time. If so, it's low efficient.

It will wait until 30 seconds each time but that's the desired result, we want to run the program for 30 seconds. I'm realizing the name might be misleading, it's less a timeout and more of 'run_for_n_seconds'

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: fixed, it's 4 for hyper-threaded hosts (need >one extra physical core for dpdk)

squirrelsc: Where is this requirement from? dpdk or Azure? If it's dpdk, it's ok to keep it here. If it's from Azure, it's better to convert the dpdk to a feature, so we can customize the requirement of the dpdk in Azure platform.

microsoft/testsuites/dpdk/dpdktestpmd.py
squirrelsc: 30 seconds is pretty long. Can it be tested in shorter time, how about 5 seconds?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I think it's more pythonic to put all the init steps into __init__. Any error during this step isn't recoverable so the only effect of moving it out out __init__ is that we now have to write two lines instead of one everywhere.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: dpdk

mcgov: Azure's requirements are bundled up into SRIOV support so I don't think we need to make dpdk it's own feature

squirrelsc: SRIOV has its requirement already. So it doesn't need the extra core count in test cases, if the dpdk has same requirement like SRIOV. The feature is used to manage requirement and logics, so people know where a requirement is from, and how it applies to different platforms. It helps to improve maintenance ability of code.

squirrelsc: If it's a requirement of dpdk, a reference can help track the requirement here. As we met problems in v2, it's painful to understand the logic and requirements, if the documentation is not enough.

mcgov: sounds good, removing it here

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: In Python code, if it has one method, it can be a function, instead of a class. Forgot to give this suggestion in first round...

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -81,28 +68,79 @@ def _execute_expect_zero(self, node: Node, cmd: str) -> str:
...
-            disk_name="nodev",
-            point="/mnt/huge-1G",
-            type="hugetlbfs",
-            options="pagesize=1G",
-        )
+    @TestCaseMetadata(
+        description="""
+           Tests a more realistic sender/forwarder/receiver setup.
+        """,
+        priority=1,
```
squirrelsc: Does it need to be P1? The v2 priority should plus 1 into v3. For example, P1 in v2 should be p2 in v3.

mcgov:
ping @squirrelsc

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -81,28 +68,79 @@ def _execute_expect_zero(self, node: Node, cmd: str) -> str:
...
-            type="hugetlbfs",
-            options="pagesize=1G",
-        )
+    @TestCaseMetadata(
+        description="""
+           Tests a more realistic sender/forwarder/receiver setup.
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_core_count=8,
```
squirrelsc: Is the `min_core_count` needed?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -81,28 +68,79 @@ def _execute_expect_zero(self, node: Node, cmd: str) -> str:
...
+            min_nic_count=2,
+            network_interface=Sriov,
+            min_count=3,
+        ),
+    )
+    def check_3node_forwarding(self, environment: Environment, log: Logger) -> None:
+        # TODO: stub test, just submitting refactoring PR
+        pass
```
squirrelsc: How about move this test case to next PR. So we can merge other parts firstly.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -29,44 +36,24 @@ class Dpdk(TestSuite):
...
+        test_kit.initialize()
+        node_nic_info, testpmd = test_kit.node_nic_info, test_kit.testpmd
 
         # grab a nic and run testpmd
         test_nic = node_nic_info.get_nic(node_nic_info.get_upper_nics()[-1])
         vdev_type = "net_vdev_netvsc0"
         testpmd_include_str = test_nic.testpmd_include(vdev_type)
-        testpmd_output = testpmd.run_with_timeout(testpmd_include_str, 30)
+        testpmd_cmd = testpmd._generate_testpmd_command(testpmd_include_str, "txonly")
+        testpmd_output = testpmd.run_for_n_seconds(testpmd_cmd, 30)
```
squirrelsc: 30 seconds is pretty long. Can it be tested in shorter time, how about 5 seconds?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -81,28 +68,79 @@ def _execute_expect_zero(self, node: Node, cmd: str) -> str:
...
         )
+        assert_that(sriov_is_enabled).described_as(
+            f"SRIOV was not enabled for this test node ({self._node.name})"
+        ).is_true()
+
+        # dump some info about the pci devices before we start
+        lspci = self._node.tools[Lspci]
+        self._log.info(f"Node[{self._node.name}] LSPCI Info:\n{lspci.run().stdout}\n")
+        # enable hugepages (required by dpdk)
+        _hugepages_init(self._node)
```
squirrelsc: The init and enable called together always, can they be merged?

squirrelsc: init_hugepages and enable_hugepages may be good names.

mcgov (commented):
refactoring the node resources wrapper class (DpdkTestKit) into a class that's basically just a struct that just holds the resources and uses a factory function to init.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: They are at the moment but won't always be, I'd prefer to keep them separate for now

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: We need to gather the packets-per-second so I'd rather have a little more time than not enough. I'll change to 10.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: nope, will remove

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: sure

squirrelsc (dismissed):

squirrelsc (approved):

PR #1536, opened 2021-09-15
Add hyperv_generation variable into AzureNodeSchema

lubaihua33:
When use vhd to create VM, we need hypverV generation parameter. Before fix, if the vhd is gen2 image, it will has error.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -94,6 +94,7 @@ class AzureNodeSchema:
         default=None, metadata=field_metadata(data_key="shared_gallery")
     )
     vhd: str = ""
+    hyperv_generation: int = 1
     nic_count: int = 1
     enable_sriov: bool = False
     data_disk_count: int = 0
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1017,6 +1017,11 @@ class AzurePlatform(Platform):
             raise LisaException("vm_size is not detected before deploy")
         if not azure_node_runbook.location:
             raise LisaException("location is not detected before deploy")
+        if azure_node_runbook.hyperv_generation not in [1, 2]:
+            raise LisaException(
+                "hyperv_generation need value 1 or 2, "
+                f"but {azure_node_runbook.hyperv_generation}",
+            )
         if azure_node_runbook.vhd:
             # vhd is higher priority
             azure_node_runbook.vhd = self._get_deployable_vhd_path(
--- a/docs/run_test/platform.rst
+++ b/docs/run_test/platform.rst
@@ -28,6 +28,7 @@ To run using vhd, add the following to runbook :
          azure:
             ...
             vhd: "<VHD URL>"
+            hyperv_generation: <1 or 2>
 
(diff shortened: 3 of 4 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -94,6 +94,7 @@ class AzureNodeSchema:
         default=None, metadata=field_metadata(data_key="shared_gallery")
     )
     vhd: str = ""
+    hyperv_generation: str = "V1"
```
squirrelsc, line 97: What this is for? should it be decided by vm size already?

squirrelsc: The type should be an integer, and limited to 1, or 2. It can reduce human error on how to assign this value. And it should be converted to "V1" or "V2" in code.

squirrelsc: Is it possible to know from the marketplace image, it's a v1 or v2?

lubaihua33: Have changed to "int". 
The "hyperVGeneration" parameter in the template is only used for VHD scenario. If it's marketplace image, this parameter doesn't work.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1017,6 +1017,14 @@ def _create_node_runbook(
             raise LisaException("vm_size is not detected before deploy")
         if not azure_node_runbook.location:
             raise LisaException("location is not detected before deploy")
+        if (
```
squirrelsc: `azure_node_runbook.hyperv_generation not in [1, 2]`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -94,6 +94,7 @@ class AzureNodeSchema:
         default=None, metadata=field_metadata(data_key="shared_gallery")
     )
     vhd: str = ""
+    hyperv_generation: int = 1
```
squirrelsc, line 97: Give the default value is None. So it won't fail, if no gen specified, and it's a gen2 only vm size.

lubaihua33: I don't think we can set the default value None.
If we don't add "hyperVGeneration" flag in the template, or write like this:
```"hyperVGeneration": ""```

When we deploy VM, it will have this error:
```deployment failed: InvalidParameter: Required parameter 'hyperVGeneration' is missing (null).```

If we use VHD, we must have hyperVGeneration flag in the template. So set hyperv_generation default value as 1, if the VM size is gen2, we must specify hyperv_generation parameter.

squirrelsc: If it's VHD deployment, we will provide the value. How about gallery image or SIG deployment? Can it be empty, so if it's gen2, it can works without any setting?

lubaihua33: I tried gallery images, it doesn't depend on "hyperVGeneration". For example, if the gallery image is gen2 image, the VM can be created successfully even if we set hyperv_generation as 1. If the template doesn't have "hyperVGeneration" this flag, and the image is gen2, it also can work for gallery image scenario, but not for VHD scenario.

squirrelsc (approved):

PR #1539, opened 2021-09-17
Add storage test cases

dsrivastavv:
Migrate Test : https://github.com/microsoft/lisa/blob/master/Testscripts/Linux/VERIFY-LINUX-DISK-SETUP.py

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -18,7 +18,7 @@ from lisa.features.gpu import ComputeSDK
 from lisa.node import Node, RemoteNode
 from lisa.operating_system import CentOs, Redhat, Suse, Ubuntu
 from lisa.sut_orchestrator.azure.common import AZURE, AzureNodeSchema
-from lisa.util import LisaException
+from lisa.util import LisaException, NotMeetRequirementException
 
 if TYPE_CHECKING:
     from .platform_ import AzurePlatform
@@ -267,8 +267,20 @@ _disk_size_iops_map: Dict[schema.DiskType, List[Tuple[int, int]]] = {
 @dataclass_json()
 @dataclass()
 class AzureDiskOptionSettings(schema.DiskOptionSettings):
+    has_resource_disk: Optional[bool] = None
+
     def __hash__(self) -> int:
         return super().__hash__()
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+    def __repr__(self) -> str:
+        return f"has_resource_disk: {self.has_resource_disk},{super().__repr__()}"
+
+    def __eq__(self, o: object) -> bool:
+        assert isinstance(o, AzureDiskOptionSettings), f"actual: {type(o)}"
+        return self.has_resource_disk == o.has_resource_disk and super().__eq__(o)
 
     # It uses to override requirement operations.
     def check(self, capability: Any) -> search_space.ResultReason:
@@ -298,6 +310,12 @@ class AzureDiskOptionSettings(schema.DiskOptionSettings):
                 self.data_disk_size, capability.data_disk_size
             ),
             "data_disk_size",
+        )
+        result.merge(
+            self._check_has_resource_disk(
+                self.has_resource_disk, capability.has_resource_disk
+            ),
+            "has_resource_disk",
         )
 
         return result
@@ -398,6 +416,10 @@ class AzureDiskOptionSettings(schema.DiskOptionSettings):
         # all caching types are supported, so just take the value from requirement.
         min_value.data_disk_caching_type = self.data_disk_caching_type
 
+        min_value.has_resource_disk = self._generate_min_capability_has_resource_disk(
+            self.has_resource_disk, capability.has_resource_disk
+        )
+
         return min_value
 
     def _get_disk_size_from_iops(
@@ -406,6 +428,40 @@ class AzureDiskOptionSettings(schema.DiskOptionSettings):
         return next(
             disk_size for iops, disk_size in disk_type_iops if iops == data_disk_iops
         )
+
+    def _check_has_resource_disk(
+        self, requirement: Optional[bool], capability: Optional[bool]
+    ) -> search_space.ResultReason:
+        result = search_space.ResultReason()
+        # if requirement is none, capability can be either of True or False
+        # else requirement should match capability
+        if requirement is not None:
+            if capability is None:
+                result.add_reason(
+                    "if requirements isn't None, capability shouldn't be None"
+                )
+            else:
+                if requirement != capability:
+                    result.add_reason(
+                        "requirement is a truth value, capability should be exact "
+                        f"match, requirement: {requirement}, "
+                        f"capability: {capability}"
+                    )
+
+        return result
+
+    def _generate_min_capability_has_resource_disk(
+        self, requirement: Optional[bool], capability: Optional[bool]
+    ) -> Optional[bool]:
+        check_result = self._check_has_resource_disk(requirement, capability)
+        if not check_result.result:
+            raise NotMeetRequirementException(
+                "cannot get min value, capability doesn't support requirement"
+            )
+        return capability
+
+    def _get_key(self) -> str:
+        return f"{super()._get_key()}/{self.has_resource_disk}"
 
 
 class Disk(AzureFeatureMixin, features.Disk):
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -2,17 +2,22 @@
 # Licensed under the MIT license.
 
 import re
-from typing import Any, List, Type
+from typing import Any, Dict, List, Type
 
-from lisa.base_tools.wget import Wget
+from lisa.base_tools import Cat, Wget
 from lisa.executable import Tool
-from lisa.operating_system import Redhat
+from lisa.operating_system import CoreOs, Redhat
 from lisa.tools import Modinfo
 from lisa.util import LisaException, find_patterns_in_lines
 
 
 class Waagent(Tool):
     __version_pattern = re.compile(r"(?<=\-)([^\s]+)")
+
+    # ResourceDisk.MountPoint=/mnt
+    # ResourceDisk.EnableSwap=n
+    # ResourceDisk.EnableSwap=y
+    _key_value_regex = re.compile(r"^\s*(?P<key>\S+)=(?P<value>\S+)\s*$")
 
     @property
     def command(self) -> str:
@@ -45,6 +50,41 @@ class Waagent(Tool):
         # self.run("-deprovision+user --force", sudo=True)
         result = self.run("-deprovision --force", sudo=True)
         result.assert_exit_code()
+
+    def get_configuration(self) -> Dict[str, str]:
+        if isinstance(self.node.os, CoreOs):
+            waagent_conf_file = "/usr/share/oem/waagent.conf"
+        else:
+            waagent_conf_file = "/etc/waagent.conf"
+
+        config = {}
+        cfg = self.node.tools[Cat].run(waagent_conf_file).stdout
+        for line in cfg.splitlines():
+            matched = self._key_value_regex.fullmatch(line)
+            if matched:
+                config[matched.group("key")] = matched.group("value")
+
+        return config
+
+    def get_root_device_timeout(self) -> int:
+        waagent_configuration = self.get_configuration()
+        return int(waagent_configuration["OS.RootDeviceScsiTimeout"])
+
+    def get_resource_disk_mount_point(self) -> str:
+        waagent_configuration = self.get_configuration()
+        return waagent_configuration["ResourceDisk.MountPoint"]
+
+    def is_swap_enabled(self) -> bool:
+        waagent_configuration = self.get_configuration()
+        is_swap_enabled = waagent_configuration["ResourceDisk.EnableSwap"]
+        if is_swap_enabled == "y":
+            return True
+        elif is_swap_enabled == "n":
+            return False
+        else:
+            raise LisaException(
+                f"Unknown value for ResourceDisk.EnableSwap : {is_swap_enabled}"
+            )
 
 
 class VmGeneration(Tool):
--- a/lisa/tools/swap.py
+++ b/lisa/tools/swap.py
@@ -1,0 +1,32 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+from lisa.tools.lsblk import Lsblk
+from lisa.tools.swapon import SwapOn
+
+
+class Swap(Tool):
+    @property
+    def command(self) -> str:
+        raise NotImplementedError()
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def is_swap_enabled(self) -> bool:
+        # swapon lists the swap files and partitions
+        # example output:
+        # /swapfile file 1024M 507.4M   -1
+        swapon = self.node.tools[SwapOn].run("-s").stdout
+        if "swap" in swapon:
+            return True
+
+        # lsblk lists swap partitions
+        # example output:
+        # sdb2   8:18   0   7.6G  0 part [SWAP]
+        lsblk = self.node.tools[Lsblk].run().stdout
+        if "SWAP" in lsblk:
+            return True
+
+        return False
--- a/microsoft/testsuites/core/storage.py
+++ b/microsoft/testsuites/core/storage.py
@@ -1,0 +1,129 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from pathlib import PurePosixPath
+from typing import Pattern
+
+from assertpy.assertpy import assert_that
+
+from lisa import (
+    Logger,
+    RemoteNode,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    simple_requirement,
+)
+from lisa.features import Disk
+from lisa.sut_orchestrator.azure.common import AZURE
+from lisa.sut_orchestrator.azure.features import AzureDiskOptionSettings
+from lisa.sut_orchestrator.azure.tools import Waagent
+from lisa.tools import Cat, Swap
+from lisa.util import get_matched_str
+
+
+@TestSuiteMetadata(
+    area="storage",
+    category="functional",
+    description="""
+    This test suite is used to run storage related tests.
+    """,
+)
+class Storage(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test will check that VM root disk(os disk) is provisioned
+        with the correct timeout.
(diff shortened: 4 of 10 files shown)


microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        requirement=simple_requirement(
+            environment_status=EnvironmentStatus.Deployed,
+        ),
+    )
+    def verify_swap(self, log: Logger, node: RemoteNode) -> None:
+        swapon_output = node.tools[SwapOn].run("-s", sudo=True).stdout
+        lsblk_output = node.tools[Lsblk].run().stdout
+        _is_swap_enabled = self._is_swap_enabled(log, node)
+
+        if (
```
squirrelsc: This logic should be in the tool

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        2. Check swap status in `waagent.conf`.
+        3. Verify that values in step 1 and step 2 are equal.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            environment_status=EnvironmentStatus.Deployed,
+        ),
+    )
+    def verify_swap(self, log: Logger, node: RemoteNode) -> None:
+        swapon_output = node.tools[SwapOn].run("-s", sudo=True).stdout
```
squirrelsc: This command can run without sudo on Ubuntu. Does it need sudo?

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        corresponds to entry with mount point `/' in the `mount` command.
+        2. Verify the timeout value for root disk in
+        `/sys/block/<partition>/device/timeout` file is set to 300.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            environment_status=EnvironmentStatus.Deployed,
+        ),
+    )
+    def verify_root_device_timeout(
```
squirrelsc: `verify_root_device_timeout_setting` is clearer.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+    )
+
+    # /dev/sda1 on / type ext4 (rw,relatime,discard)
+    _root_partition_regex = re.compile(
+        r"\s*\/dev\/(?P<partition>\D+).*\s+on\s+\/\s+type.*"
+    )
+
+    @TestCaseMetadata(
+        description="""
+        This test will check that VM root disk(os disk) is provisioned
```
squirrelsc, line 36: It's better to find the source of the requirement. If it's from waagent or other Azure specific, the test case should be limited to Azure only.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+    category="functional",
+    description="""
+    This test suite is used to run storage related tests.
+    """,
+)
+class Storage(TestSuite):
+    DEFAULT_DISK_TIMEOUT = 300
+
+    # ResourceDisk.MountPoint=/mnt
+    _resource_disk_mount_point_regex = re.compile(
```
squirrelsc: They can be summarized to one pattern, extract key/value pairs, and check if the key is wanted.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        with the correct timeout.
+        Steps:
+        1. Find the root disk (os disk) partition for the VM. The root partition
+        corresponds to entry with mount point `/' in the `mount` command.
+        2. Verify the timeout value for root disk in
+        `/sys/block/<partition>/device/timeout` file is set to 300.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            environment_status=EnvironmentStatus.Deployed,
```
squirrelsc: The `environment_status` is optional for most test cases, unless a test case like smoke_test, it needs to start before the vm is connected.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            log.debug("ResourceDisk handled by waagent.")
+            walacfg_path = self._get_wala_conf_path(node)
+            walacfg = node.tools[Cat].run(walacfg_path).stdout
+            matched = self._resource_disk_mount_point_regex.fullmatch(walacfg)
+            assert matched
+            mount_point = matched.group("resource_disk_mount_point")
+        return mount_point
+
+    def _get_os_disk_partition(
```
squirrelsc: This method can be in disk feature, so it's shareable.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        priority=1,
+        requirement=simple_requirement(
+            environment_status=EnvironmentStatus.Deployed,
+        ),
+    )
+    def verify_resource_disk_mtab_entry(self, log: Logger, node: RemoteNode) -> None:
+        resource_disk_mount_point = self._get_data_disk_mount_point(log, node)
+        os_disk_partition = self._get_os_disk_partition(log, node)
+        if os_disk_partition == "sda":
+            mntresource = "/dev/sdb1 " + resource_disk_mount_point
```
squirrelsc: If we know `resource_disk_mount_point` already, can we reverse match back to /dev/sdb1 or /dev/sda1? After that, we can have a check, the os disk and resource disk are not the same device. It's clearer to me.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        else:
+            mntresource = "/dev/sda1 " + resource_disk_mount_point
+
+        output = node.tools[Cat].run("/etc/mtab").stdout
+        if mntresource in output:
+            log.debug("ResourceDisk entry is present.")
+        else:
+            raise LisaException("ResourceDisk entry is not present.")
```
squirrelsc: It needs to output the expected value. The command output is in debug log already, but the expected mntresource is not printed. If it's failed, we need to know what's expected in the code logic.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+            matched = self._root_partition_regex.fullmatch(line)
+            if matched:
+                os_disk_partition = matched.group("partition")
+        assert os_disk_partition
+
+        log.info(f"OS disk partition : {os_disk_partition}")
+        return os_disk_partition
+
+    def _is_swap_enabled(self, log: Logger, node: RemoteNode) -> bool:
+        waagent_conf_file = self._get_wala_conf_path(node)
```
squirrelsc: We can add this logic into waagent tool. It's under the Azure folder.

squirrelsc: All waagent config related logic should be there.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        return mount_point
+
+    def _get_os_disk_partition(
+        self,
+        log: Logger,
+        node: RemoteNode,
+    ) -> str:
+        mount_output = node.tools[Mount].run().stdout
+        os_disk_partition: Optional[str] = None
+        for line in mount_output.splitlines():
```
squirrelsc: try `get_matched_str` in lisa.util

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def _get_data_disk_mount_point(
+        self,
+        log: Logger,
+        node: RemoteNode,
+    ) -> str:
+        if node.shell.exists(
+            PurePosixPath("/var/log/cloud-init.log")
+        ) and node.shell.exists(PurePosixPath("/var/lib/cloud/instance")):
+            log.debug("Data handled by cloud-init.")
```
squirrelsc: What means "data" in the log? Should it be "provisioning"?

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def _get_wala_conf_path(self, node: RemoteNode) -> str:
+        if isinstance(node.os, CoreOs):
+            return "/usr/share/oem/waagent.conf"
+        elif node.os.information.vendor == "clear-linux-os":
+            return "/usr/share/defaults/waagent/waagent.conf"
+        else:
+            return "/etc/waagent.conf"
+
+    def _get_data_disk_mount_point(
```
squirrelsc: It looks like some method should be shared, but I don't know where is the best place. How about disk feature? It's implemented by platform.

squirrelsc: Another thing is the name may need to be `_get_ephemeral_disk_mount_point`

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+            ("swap" not in swapon_output) and ("SWAP" not in lsblk_output)
+        ) and _is_swap_enabled:
+            raise LisaException(
+                "Swap is enabled in waagent.conf but no swap partition/files found."
+            )
+
+    def _get_wala_conf_path(self, node: RemoteNode) -> str:
+        if isinstance(node.os, CoreOs):
+            return "/usr/share/oem/waagent.conf"
+        elif node.os.information.vendor == "clear-linux-os":
```
squirrelsc: The CoreOs and clear-linux can be removed for now. The core os is not supported anymore.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+        3. Verify that "/dev/<resource_disk_partition> <mount_point>` entry is
+        present in `/etc/mtab` file.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            environment_status=EnvironmentStatus.Deployed,
+        ),
+    )
+    def verify_resource_disk_mtab_entry(self, log: Logger, node: RemoteNode) -> None:
+        resource_disk_mount_point = self._get_data_disk_mount_point(log, node)
```
LiliDeng: there are some sizes which don't offer resource disk. we need handle this situation.

microsoft/testsuites/core/storage.py
```diff
@@ -0,0 +1,203 @@
...
+
+    @TestCaseMetadata(
+        description="""
+        This test will check that VM root disk(os disk) is provisioned
+        with the correct timeout.
+        Steps:
+        1. Find the root disk (os disk) partition for the VM. The root partition
+        corresponds to entry with mount point `/' in the `mount` command.
+        2. Verify the timeout value for root disk in
+        `/sys/block/<partition>/device/timeout` file is set to 300.
```
LiliDeng, line 42: this is set by waagent, please try to find some doc and paste here.

LiliDeng: OS.RootDeviceScsiTimeout

microsoft/testsuites/core/storage.py
LiliDeng: flatcar is still available, we count it as coreos

microsoft/testsuites/core/storage.py
LiliDeng: details please see https://docs.microsoft.com/en-us/azure/virtual-machines/azure-vms-no-temp-disk

microsoft/testsuites/core/storage.py
dsrivastavv: The tool used for checking swap is `cat`. But swap has no relation with this tool which is why I put it in this file. What do you suggest?

microsoft/testsuites/core/storage.py
dsrivastavv: Since this is checking disk behavior, I thought sudo might be required on certain distros.

microsoft/testsuites/core/storage.py
dsrivastavv: I don't think there is a way to get disk partition from mount point. For example, mount point will be like `/mnt`, `/` etc. But the disk partition names are like `sda`, `sdb` etc

microsoft/testsuites/core/storage.py
dsrivastavv: This mean that cloud-init is handling disk mounting

microsoft/testsuites/core/storage.py
dsrivastavv: How about `_get_resource_disk_mount_point`? Resource disk terminology is used in azure docs as well for these kind of partitions

microsoft/testsuites/core/storage.py
dsrivastavv: We can keep CoreOs and remove clear-linux-os. Thanks for catching this Lili and Chi!

microsoft/testsuites/core/storage.py
dsrivastavv: How about this doc : https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/agent-linux#configuration?

microsoft/testsuites/core/storage.py
dsrivastavv: @LiliDeng Do you know if there is any field in SKU capability that we can use to figure out if a VM supports resource disk?

@squirrelsc Is there a way to skip test for selected VM family?

microsoft/testsuites/core/storage.py
squirrelsc: The output is from SwapOn and Lsblk. Maybe you can add a virtual tool to use them to check the swap. It likes the tool for service.

microsoft/testsuites/core/storage.py
dsrivastavv: Do you mean in `\lisa\features\disks.py` ? Also, how about placing this in `mount` tool folder? It's processing this value from the output of mount command

microsoft/testsuites/core/storage.py
squirrelsc: it's ok to me.

microsoft/testsuites/core/storage.py
squirrelsc: Use SkippedException can skip case in any place. But because vm sizes are Azure specified, so it needs to be written in Azure related code. If the test case is Azure only, the code can be written here. If it's not, we need to consider have a feature called ResourceDisk to test it.

microsoft/testsuites/core/storage.py
squirrelsc: The partition information may be used in other test cases. The mount is more like the `mount` command. If we put it there, it's not easy to find the method.

lisa/search_space.py
```diff
@@ -452,6 +452,37 @@ def generate_min_capability_setspace_from_priority(
     return min_cap
 
 
+def check_boolean(
```
squirrelsc: The boolean is not like a searchable space, and there are several types about how to merge req/cap on boolean. This is just one of situation. It's better to implement it in DiskOptionSettings itself.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -293,6 +292,12 @@ def check(self, capability: Any) -> search_space.ResultReason:
             ),
             "data_disk_size",
         )
+        result.merge(
```
squirrelsc, line 314: You can consider to use a SetSpace to support the usage here. Req can be one of True/False/None, and cap can be a set of {True, False}, True, False.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -392,6 +397,10 @@ def _generate_min_capability(self, capability: Any) -> Any:
         # all caching types are supported, so just take the value from requirement.
         min_value.data_disk_caching_type = self.data_disk_caching_type
 
+        min_value.has_resource_disk = search_space.generate_min_capability_boolean(
```
squirrelsc: It's hard to define what's the "min" for a boolean. So it doesn't like shareable logic.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -401,6 +410,15 @@ def _get_disk_size_from_iops(
             disk_size for iops, disk_size in disk_type_iops if iops == data_disk_iops
         )
 
+    def __hash__(self) -> int:
```
squirrelsc: The `__aaa__` are default methods, they are put above public methods.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -261,8 +261,7 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
 @dataclass_json()
 @dataclass()
 class AzureDiskOptionSettings(schema.DiskOptionSettings):
-    def __hash__(self) -> int:
-        return super().__hash__()
+    has_resource_disk: Optional[bool] = None
```
squirrelsc, line 270: Inherit get_key, and add it into get_key, so the comparison can include this value.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -2,22 +2,31 @@
...
+    # ResourceDisk.EnableSwap=y
+    _key_value_regex = re.compile(r"^\s*(?P<key>\S+)=(?P<value>\S+)\s*$")
+
     @property
     def command(self) -> str:
         return self._command
 
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Cat]
```
squirrelsc: Is there some distro doesn't have the `cat`? Only it needs be here, if some distro doesn't have `cat`, and it needs to be installed.

lisa/tools/swap.py
```diff
@@ -0,0 +1,23 @@
...
+    def command(self) -> str:
+        raise NotImplementedError()
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def is_swap_enabled(self) -> bool:
+        swapon = self.node.tools[SwapOn].run("-s").stdout
+        lsblk = self.node.tools[Lsblk].run().stdout
+        if ("swap" in swapon) or ("SWAP" in lsblk):
```
squirrelsc: Can they be checked one by one? If first tool detects, it doesn't need to run another one.

dsrivastavv: Swaps can be either partition or file and `swapon` should show both these types. But lisav2 code was checking both lsblk and swapon, so I think there might cases which are missed by swapon.

squirrelsc: check it one by one, it doesn't miss any logic. And it has chance to reduce a remote call.

dsrivastavv: Sorry I misunderstood your comment earlier. I thought this was regarding why are we using both swapon and lsblk together. Will update it to check one by one

lisa/sut_orchestrator/azure/features.py
dsrivastavv: SetSpace doesn't feel right here as the value could only be one of {True, False, None}. I will go with your earlier suggestion of keeping the logic in DiskOptionSettings .

squirrelsc (dismissed):

dsrivastavv:
@squirrelsc Can you please re-review this PR? There was a merge conflict and I had to manually rebase this PR

squirrelsc (approved):

PR #1543, opened 2021-09-21
Test Migration: lis_modules_check and verify_lis_modules_version

kamalca:
# verify_hyperv_modules
Checks whether the hv_modules are loaded directly into the kernel or loaded as modules.
Replaces LISAv2 test LIS-MODULES-CHECK

# verify_lis_modules_version
Only applied to CentOS and RHEL distros that have LIS installed. This test then ensured that all of the hv_modules are the same version as the LIS installation.
Replaces LISAv2 test VERIFY-LIS-MODULES-VERSION

### DIFF
--- a/microsoft/testsuites/core/hv_module.py
+++ b/microsoft/testsuites/core/hv_module.py
@@ -1,0 +1,133 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from assertpy import assert_that
+from semver import VersionInfo
+
+from lisa import Logger, RemoteNode, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa.operating_system import Redhat
+from lisa.sut_orchestrator.azure.tools import LisDriver
+from lisa.tools import Lsmod, Modinfo, Modprobe, Uname
+from lisa.util import SkippedException
+
+
+@TestSuiteMetadata(
+    area="core",
+    category="functional",
+    description="""
+    This test suite covers test cases previously handled by LISAv2:
+    LIS-MODULES-CHECK, VERIFY-LIS-MODULES-VERSION,
+    INITRD-MODULES-CHECK, RELOAD-MODULES-SMP
+
+    It is responsible for ensuring the Hyper V drivers are all present,
+    are included in initrd, and are all the same version.
+    """,
+)
+class HvModule(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Verify the list of given LIS kernel modules and verify if the version
+           matches with the Linux kernel release number. (Drivers loaded directly in
+           to the kernel are skipped)
+        """,
+        priority=1,
+    )
+    def verify_lis_modules_version(
+        self, case_name: str, log: Logger, node: RemoteNode
+    ) -> None:
+        if not isinstance(node.os, Redhat):
+            raise SkippedException(
+                f"{node.os.name} not supported. "
+                "This test case only supports Redhat distros."
+            )
+
+        lis_installed = node.os.package_exists("microsoft-hyper-v")
+
+        if not lis_installed:
+            raise SkippedException("This test case requires LIS to be installed")
+
+        modinfo = node.tools[Modinfo]
+        lis_driver = node.tools[LisDriver]
+        lis_version = lis_driver.get_version()
+
+        hv_modules = self._get_expected_modules(node)
+        for module in hv_modules:
+            module_version = VersionInfo.parse(modinfo.get_version(module))
+            assert_that(module_version).described_as(
+                f"Version of {module} does not match LIS version"
+            ).is_equal_to(lis_version)
+
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Verify the presence of all Hyper V drivers using lsmod
+           to look for the drivers not directly loaded into the kernel.
+        """,
+        priority=1,
+    )
+    def verify_hyperv_modules(
+        self, case_name: str, log: Logger, node: RemoteNode
+    ) -> None:
+        hv_modules = self._get_expected_modules(node)
+        distro_version = node.os.information.version
+
+        # Some versions of RHEL and CentOS have the LIS package installed
+        #   which includes extra drivers
+        if isinstance(node.os, Redhat):
+            modprobe = node.tools[Modprobe]
+            lis_installed = node.os.package_exists("microsoft-hyper-v")
+
+            if lis_installed:
+                hv_modules.append("pci_hyperv")
+                modprobe.run("pci_hyperv", sudo=True)
+
+            if (
+                distro_version >= "7.3.0" or distro_version < "7.5.0"
+            ) and lis_installed:
+                hv_modules.append("mlx4_en")
+                modprobe.run("mlx4_en", sudo=True)
+
+        # Counts the Hyper V drivers loaded as modules
+        missing_modules = []
+        lsmod = node.tools[Lsmod]
+        for module in hv_modules:
+            if lsmod.module_exists(module):
+                log.info(f"Module {module} present")
+            else:
+                log.error(f"Module {module} absent")
+                missing_modules.append(module)
+
+        assert_that(missing_modules).described_as(
+            "Not all Hyper V drivers are present."
+        ).is_length(0)
+
+    def _get_expected_modules(self, node: RemoteNode) -> list[str]:
+        """
+        Returns the hv_modules that are not directly loaded into the kernel and
+        therefore would be expected to show up in lsmod.
+        """
+        hv_modules_configuration = {
+            "hv_storvsc": "CONFIG_HYPERV_STORAGE",
+            "hv_netvsc": "CONFIG_HYPERV_NET",
+            "hv_vmbus": "CONFIG_HYPERV",
+            "hv_utils": "CONFIG_HYPERV_UTILS",
+            "hid_hyperv": "CONFIG_HID_HYPERV_MOUSE",
+            "hv_balloon": "CONFIG_HYPERV_BALLOON",
+            "hyperv_keyboard": "CONFIG_HYPERV_KEYBOARD",
+        }
+        uname = node.tools[Uname]
+        kernel_version = uname.get_linux_information().kernel_version_raw
+        config_path = f"/boot/config-{kernel_version}"
+
+        modules = []
+        for module in hv_modules_configuration:
+            if (
+                node.execute(
+                    f"grep ^{hv_modules_configuration[module]}=y {config_path}"
+                ).exit_code
+                != 0
+            ):
+                modules.append(module)
+
+        return modules


lisa/tools/dmesg.py
```diff
@@ -82,6 +86,24 @@ def get_vmbus_version(self) -> VersionInfo:
                 return VersionInfo(int(major), int(minor))
         raise LisaException("No find matched vmbus version in dmesg")
 
+    def get_lis_version(self) -> VersionInfo:
```
squirrelsc: There is an Azure tool called `LisDriver` already in `lisa\sut_orchestrator\azure\tools.py`. Check if it's fit into this scenario. The LIS is Microsoft specific things, it's better not to put it in generic tool, dmesg.

squirrelsc: The modinfo tool also can get version of hv_vmbus.

lisa/tools/rpm.py
```diff
@@ -0,0 +1,40 @@
...
+
+class Rpm(Tool):
+    @property
+    def command(self) -> str:
+        return self._command
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._command = "rpm"
+
+    def is_package_installed(
```
squirrelsc: The os should have package check logic already. It doesn't need a tool for rpm.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+    area="core",
+    category="functional",
+    description="""
+    This test suite covers test cases previously handled by LISAv2:
+    LIS-MODULES-CHECK, VERIFY-LIS-MODULES-VERSION,
+    INITRD-MODULES-CHECK, RELOAD-MODULES-SMP
+    """,
+)
+class HvModule(TestSuite):
+    def get_modules(self, node: RemoteNode) -> list[str]:
```
squirrelsc: Because it's not a test case, make it private, and put it after other test case methods.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+            "hid_hyperv",
+            "hv_balloon",
+            "hyperv_keyboard",
+        ]
+        uname = node.tools[Uname]
+        kernel_version = uname.get_linux_information().kernel_version_raw
+        config_path = f"/boot/config-{kernel_version}"
+
+        if node.execute(f"grep CONFIG_HYPERV_STORAGE=y {config_path}").exit_code == 0:
+            hv_modules.remove("hv_storvsc")
```
squirrelsc: 1. A loop looks better, you can add a dict to map module name and config names.
2. The logic can be more straightforward, if it add an item in if it exists, instead of deleting.

microsoft/testsuites/core/hv_module.py
squirrelsc: The method name can be improved, to specify get what kind of modules.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+        matches with the Linux kernel release number.
+        """,
+        priority=0,
+    )
+    def verify_lis_modules_version(
+        self, case_name: str, log: Logger, node: RemoteNode
+    ) -> None:
+        if not (isinstance(node.os, CentOs) or isinstance(node.os, Redhat)):
+            log.info("Distro not supported")
+            raise SkippedException
```
squirrelsc: Add description into the exception, and specify which distro is passed in.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+        This test case will
+        1. Verify the list of given LIS kernel modules and verify if the version
+        matches with the Linux kernel release number.
+        """,
+        priority=0,
+    )
+    def verify_lis_modules_version(
+        self, case_name: str, log: Logger, node: RemoteNode
+    ) -> None:
+        if not (isinstance(node.os, CentOs) or isinstance(node.os, Redhat)):
```
squirrelsc: 1. OS check can be added into the requirement
2. CentOs is below the RedHat. So check if it's Redhat is enough.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+        lis_installed = rpm.is_package_installed("microsoft-hyper-v")
+
+        modinfo = node.tools[Modinfo]
+        dmesg = node.tools[Dmesg]
+        lis_version = dmesg.get_lis_version()
+
+        if lis_installed:
+            for module in hv_modules:
+                module_version = modinfo.get_version(module)
+                assert_that(lis_version == module_version).described_as(
```
squirrelsc: In this format, if it fails, it will print both expected and actual values.
`assert_that(module_version).described_as(......).equal(lis_version)`

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+    def lis_modules_check(self, case_name: str, log: Logger, node: RemoteNode) -> None:
+        hv_modules = self.get_modules(node)
+        distro_version = node.os.information.version
+
+        if isinstance(node.os, CentOs) or isinstance(node.os, Redhat):
+            rpm = node.tools[Rpm]
+            modprobe = node.tools[Modprobe]
+            lis_installed = rpm.is_package_installed("microsoft-hyper-v")
+
+            if distro_version >= "4.3.0" and lis_installed:
```
squirrelsc: Add comments to explain what's it used to check. We don't cover distro_version is lower than 4.3 now. The logic can be simpler.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+                ).is_true()
+
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Verify the presence of all LIS modules
+        """,
+        priority=0,
+    )
+    def lis_modules_check(self, case_name: str, log: Logger, node: RemoteNode) -> None:
```
squirrelsc: Does it apply to all distro, or only redhat, centos?

squirrelsc: The method names should be `verify_.....`. It makes names consistent.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+        present_modules = 0
+        lsmod = node.tools[Lsmod]
+        for module in hv_modules:
+            if lsmod.module_exists(module):
+                log.info(f"Module {module} present")
+                present_modules = present_modules + 1
+            else:
+                log.error(f"Module {module} absent")
+
+        print(f"{len(hv_modules)} == {present_modules}")
```
squirrelsc: Don't add print in the code. If the case failed, it should print details.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+        assert_that(present_modules, "Not all LIS modules are present.").is_equal_to(
```
squirrelsc: Use the is_length, so if it's failed, it will print details.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+           matches with the Linux kernel release number. (Drivers loaded directly in
+           to the kernel are skipped)
+        """,
+        priority=0,
+    )
+    def verify_lis_modules_version(
+        self, case_name: str, log: Logger, node: RemoteNode
+    ) -> None:
+        if not isinstance(node.os, Redhat):
+            log.info(f"{node.os.name} not supported.")
```
squirrelsc: This line can be merged into below exception message. If the test fails, the message will show in final result, it's helpful for troubleshooting. The log is not accessible easily as the final result.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+                log.error(f"Module {module} absent")
+
+        num_modules_expected = len(hv_modules)
+        assert_that(num_modules_loaded).described_as(
+            "Not all Hyper V drivers are present."
+        ).is_equal_to(num_modules_expected)
+
+    def __get_modules_not_in_kernel(self, node: RemoteNode) -> list[str]:
+        """
+        Returns the hv_modules that are not directly loaded into the kernel.
```
squirrelsc: It can be described in a positive way like `_get_modularized_lis_modules`.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+            "hyperv_keyboard": "CONFIG_HYPERV_KEYBOARD",
+        }
+        uname = node.tools[Uname]
+        kernel_version = uname.get_linux_information().kernel_version_raw
+        config_path = f"/boot/config-{kernel_version}"
+
+        for module in hv_modules_configuration:
+            if (
+                node.execute(
+                    f"grep {hv_modules_configuration[module]}=y {config_path}"
```
squirrelsc: 1. They can be added to a list, if the line is `=m`. Rather than `y` or `m`, it may be other values.
2. Match it from beginning of the line. We found a bug that the line is commented out with a `#`. So, it's worth to validate the line is effective without any prefix.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+                num_modules_loaded = num_modules_loaded + 1
+            else:
+                log.error(f"Module {module} absent")
+
+        num_modules_expected = len(hv_modules)
+        assert_that(num_modules_loaded).described_as(
+            "Not all Hyper V drivers are present."
+        ).is_equal_to(num_modules_expected)
+
+    def __get_modules_not_in_kernel(self, node: RemoteNode) -> list[str]:
```
squirrelsc: One underscore is enough to private method, if it supposes without subclasses.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # Counts the Hyper V drivers loaded as modules
+        num_modules_loaded = 0
+        lsmod = node.tools[Lsmod]
+        for module in hv_modules:
+            if lsmod.module_exists(module):
+                log.info(f"Module {module} present")
+                num_modules_loaded = num_modules_loaded + 1
+            else:
+                log.error(f"Module {module} absent")
```
squirrelsc, line 98: If any of required module is not present, the case should be failed. So raise exception here directly. The error level log is not obvious. People won't check the output level frequently. Read the document may be helpful.
https://mslisa.readthedocs.io/en/main/write_test/guidelines.html

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+                    f"Version of {module} does not match LIS version"
+                ).is_equal_to(lis_version)
+
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Verify the presence of all Hyper V drivers using lsmod
+           to look for the drivers not directly loaded into the kernel.
+        """,
+        priority=0,
```
squirrelsc: All priority of V2 should plus 1 in v3. So it should be P1, if it's P0 in v2.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,127 @@
...
+        lsmod = node.tools[Lsmod]
+        for module in hv_modules:
+            if lsmod.module_exists(module):
+                log.info(f"Module {module} present")
+                num_modules_loaded = num_modules_loaded + 1
+            else:
+                log.error(f"Module {module} absent")
+
+        num_modules_expected = len(hv_modules)
+        assert_that(num_modules_loaded).described_as(
```
squirrelsc: Check the length of list may be better, it may print the list, if the size is not expected.

microsoft/testsuites/core/hv_module.py
```diff
@@ -0,0 +1,134 @@
...
+            "Not all Hyper V drivers are present."
+        ).is_equal_to(num_modules_expected)
+
+    def _get_expected_modules(self, node: RemoteNode) -> list[str]:
```
squirrelsc, line 105: The name can be more accurate, what's "expected"?

microsoft/testsuites/core/hv_module.py
kamalca: The idea is that there is a list of drivers expected to be present: hv_storvsc, hv_netvsc, hv_vmbus, hv_utils, hid_hyperv, hv_balloon, hyperv_keyboard. We then must skip any drivers that are loaded directly with `y` because they will not show up in lsmod, and their version from modinfo is not relevant. It would not make sense to only check for `m` because the driver could be left out entirely. This is also my motivation for naming the function `_get_expected_modules` because it returns the modules we expect to see in lsmod, but I can continue to think of a better name.

microsoft/testsuites/core/hv_module.py
kamalca: If a module is missing, it is guaranteed to fail later on because the number of expected modules will not match the number of modules found. My thought process is that it was better to check for all of the modules so that running the test one time reports all of the modules missing instead of the user possibly fixing one module then getting a different module missing error running it a second time. Maybe the best solution is to include the names of all of the missing modules again in the error message so that the user does not have to go through the error logs?

squirrelsc: Yes, check all is better, but we should include module names. That would be more actionable from the error message. You can check if assert_that(list).is_length can log it or not.

squirrelsc (approved):

PR #1544, opened 2021-09-22
Add try catch exception for _delete_boot_diagnostic_container.

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -528,12 +528,17 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
...
             )
         else:
-            self._delete_boot_diagnostic_container(resource_group_name, log)
             assert self._rm_client
             log.info(
                 f"deleting resource group: {resource_group_name}, "
                 f"wait: {self._azure_runbook.wait_delete}"
             )
+            try:
+                self._delete_boot_diagnostic_container(resource_group_name, log)
```
squirrelsc, line 542: It needs to fix the exceptions, which happens also.

LiliDeng: what is the action item should I take for this comments?

squirrelsc: Rather than this change, update `_delete_boot_diagnostic_container` to make sure it has no exception raise, if the resource group doesn't exist.

squirrelsc (approved):

PR #1545, opened 2021-09-22
Improve multithreading for root runner

dsrivastavv:


lisa/runner.py
```diff
@@ -283,7 +287,7 @@ def _start_loop(self) -> None:
                 )
 
                 for runner in remaining_runners:
-                    while not runner.is_done:
+                    while not runner.is_done and has_idle_worker:
```
squirrelsc: This change is actually pretty complex, please add what's problem today, and what's improved in the commit message.

lisa/runner.py
```diff
@@ -302,10 +306,11 @@ def _start_loop(self) -> None:
                             # current runner may not be done, but it doesn't
                             # have task temporarily. The root runner can start
                             # tasks from next runner.
+                            self._log.debug(
```
squirrelsc: This debug message may print a lot in the log. It should be avoid.

lisa/util/parallel.py
```diff
@@ -11,11 +13,13 @@
 
 class TaskManager(Generic[T_RESULT]):
     def __init__(self, max_workers: int, callback: Callable[[T_RESULT], None]) -> None:
+        self._log = get_logger("TaskManager")
```
squirrelsc, line 64: The logger change should be separated.

lisa/runner.py
dsrivastavv: It is a bit long but we might miss some information if we limit the size. What do you suggest?

lisa/util/parallel.py
```diff
@@ -25,7 +29,9 @@ def __exit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> Optional[bool]:
 
     def submit_task(self, task: Callable[[], T_RESULT]) -> None:
         future: Future[T_RESULT] = self._pool.submit(task)
+        self._future_task_map[future] = task
         self._futures.append(future)
+        self._log.debug(f"Started task : {task}")
```
squirrelsc: It's duplicated with the line in runner, so the line in runner should be removed. And the task may be too long, so add the truncation here.

lisa/runner.py
```diff
@@ -322,12 +327,14 @@ def _start_loop(self) -> None:
...
-                while (
-                    len(remaining_runners) < self._max_concurrency and has_more_runner
-                ):
-                    # Fetch runners, if runner count is smaller than concurrency
-                    # count. It makes sure all concurrency can run.
-                    try:
-                        remaining_runners.append(next(runner_iterator))
-                    except StopIteration:
-                        has_more_runner = False
+                if has_idle_worker:
```
squirrelsc: It looks not like we discussed. The logic is not change a lot with current one.

lisa/util/parallel.py
dsrivastavv: The line in runner also has `runner.id` which started this task.  Do you think runner.id would be a useful debugging information?

squirrelsc: The main reason is the description of test case may be too long. But I don't want to handle it specially here. Maybe we can keep 300 chars?

lisa/runner.py
squirrelsc: We can create a task class, and contains more information in the class. So it can be logged here together. The advantage is, it reduces lines of log.

lisa/runner.py
dsrivastavv: This is the algo we discussed and I tried to make minimum changes to the current code. Does the current change miss any case ?

MicrosoftTeams-image (2)

lisa/runner.py
```diff
@@ -302,10 +306,11 @@ def _start_loop(self) -> None:
...
                             # have task temporarily. The root runner can start
                             # tasks from next runner.
+                            self._log.debug(
+                                f"No task available for runner : {runner.id}"
+                            )
                             break
-                        if not task_manager.has_idle_worker():
-                            has_idle_worker = False
-                            break
+                        has_idle_worker = task_manager.has_idle_worker()
```
squirrelsc: we can check the has_idle_worker every time, the variable may waste the idle workers.

lisa/runner.py
squirrelsc: The loop for new runner should be in the while #2.

lisa/runner.py
```diff
@@ -105,6 +109,11 @@ def close(self) -> None:
         if self._log_handler:
             remove_handler(self._log_handler)
             self._log_handler.close()
+        self._log.debug(f"Runner finished in {self._timer.elapsed()}s.")
+
+    def get_unique_task_id(self) -> int:
```
squirrelsc: how about name it to `generate_task_id`?

lisa/runner.py
```diff
@@ -105,6 +109,11 @@ def close(self) -> None:
         if self._log_handler:
             remove_handler(self._log_handler)
             self._log_handler.close()
+        self._log.debug(f"Runner finished in {self._timer.elapsed()}s.")
```
squirrelsc: This line should be above the close lines. So that this line can appear in runner's log file, if there is.

lisa/runners/lisa_runner.py
```diff
@@ -168,11 +178,16 @@ def _associate_environment_test_results(
                 environment=environment,
             )
             if initialization_results:
-                return self._generate_task(
+                task = self._generate_task(
                     task_method=self._initialize_environment_task,
                     environment=environment,
                     test_results=initialization_results,
                 )
+                return Task(
```
squirrelsc: If put this logic into _generate_task, the duplicate parts can be reduced.

lisa/runner.py
```diff
@@ -282,52 +291,41 @@ def _start_loop(self) -> None:
...
-                        task = runner.fetch_task()
-                        if task:
-                            # the message may be too long to display.
-                            task_message = str(task)
-                            task_message = (
-                                task_message
-                                if len(task_message) < 200
-                                else f"{task_message[:200]}..."
-                            )
+                while task_manager.has_idle_worker():
```
squirrelsc: It's better to add explanation on each loop. The explanation should specify what the loop is used for.

lisa/runner.py
```diff
@@ -282,52 +291,41 @@ def _start_loop(self) -> None:
...
-                while (
-                    len(remaining_runners) < self._max_concurrency and has_more_runner
-                ):
-                    # Fetch runners, if runner count is smaller than concurrency
-                    # count. It makes sure all concurrency can run.
-                    try:
-                        remaining_runners.append(next(runner_iterator))
-                    except StopIteration:
-                        has_more_runner = False
+                    if not remaining_runners and not has_more_runner:
```
squirrelsc: This check can be merged with the top level while by adding `or has_more_runner`

lisa/runner.py
```diff
@@ -247,7 +256,7 @@ def _callback_completed(self, results: List[TestResult]) -> None:
         finally:
             self._results_lock.release()
 
-    def _start_loop(self) -> None:
+    def _start_loop(self) -> None:  # noqa: C901
         # in case all of runners are disabled
         runner_iterator = self._fetch_runners()
         remaining_runners: List[BaseRunner] = []
```
squirrelsc, line 283: The loop to fetch remaining_runners can be simple. For current logic, it needs to fetch only one. That would be enough. So change them to one line only:
```python
remaining_runners.append(next(runner_iterator))
```

lisa/runner.py
```diff
@@ -266,14 +275,14 @@ def _start_loop(self) -> None:
             task_manager = TaskManager[List[TestResult]](
```
squirrelsc: The `if self._runners:` can be changed to below. It can reduce complexity of this method.
```python
if not self._runners:
    return
```

lisa/runner.py
```diff
@@ -282,52 +291,41 @@ def _start_loop(self) -> None:
...
+                                    f"No task available for runner: {runner.id}"
+                                )
+                                break
+                        if runner.is_done:
+                            # remove fully completed runner.
+                            runner.close()
+                            remaining_runners.remove(runner)
                             self._log.debug(
-                                f"fetched task from {runner.id}: '{task_message}'"
+                                f"runner '{runner.id}' is done, "
```
squirrelsc: Because there is a line of log in `runner.close()` already, this line can be removed. In practice, the list of remaining runner doesn't help a lot.

lisa/runner.py
```diff
@@ -282,52 +291,41 @@ def _start_loop(self) -> None:
...
-                        if task:
-                            # the message may be too long to display.
-                            task_message = str(task)
-                            task_message = (
-                                task_message
-                                if len(task_message) < 200
-                                else f"{task_message[:200]}..."
-                            )
+                while task_manager.has_idle_worker():
+                    for runner in remaining_runners:
```
squirrelsc: `for runner in remaining_runners[:]:`

lisa/runner.py
```diff
@@ -266,14 +275,14 @@ def _start_loop(self) -> None:
             task_manager = TaskManager[List[TestResult]](
                 self._max_concurrency, self._callback_completed
             )
+
             # set the global task manager for cancellation check
             set_global_task_manager(task_manager)
             has_more_runner = True
 
             # run until no more task and all runner are closed
             while task_manager.wait_worker() or remaining_runners:
```
squirrelsc: `while task_manager.wait_worker() or remaining_runners or has_more_runner:`

lisa/util/parallel.py
```diff
@@ -2,29 +2,80 @@
...
+    def __init__(
+        self,
+        task_id: int,
+        task: Callable[[], T_RESULT],
+        parent_logger: Optional[Logger],
+    ) -> None:
+        self.id = task_id
+        self._task = task
+        self._log = get_logger("Task", str(self.id), parent_logger)
+        self._log.debug(self)
```
squirrelsc: explain more like f"generate task {self}"

lisa/runners/lisa_runner.py
```diff
@@ -116,13 +116,21 @@ def fetch_task(self) -> Optional[Callable[[], List[TestResult]]]:
...
+                            lambda: skipped_test_results,
+                            self._log,
+                        )
         elif available_results:
             # no available environments, so mark all test results skipped.
             skipped_test_results = self._skip_test_results(available_results)
 
             self.status = ActionStatus.SUCCESS
-            return lambda: skipped_test_results
+            return Task(
```
squirrelsc: There can be a lightweight method like `_report_results` to make it shared. 

```python
def _report_results(self, test_results: List[TestResult])->Task[List[TestResult]]:
    return Task(self.generate_task_id(), lambda: test_results, self._log)
```

lisa/runner.py
```diff
@@ -102,10 +106,15 @@ def fetch_task(self) -> Optional[Callable[[], List[TestResult]]]:
         raise NotImplementedError()
 
     def close(self) -> None:
+        self._log.debug(f"Runner finished in {self._timer.elapsed()}s.")
```
squirrelsc, line 108: `self._timer.elapsed_text()` can return better format.

lisa/runner.py
```diff
@@ -247,6 +256,27 @@ def _callback_completed(self, results: List[TestResult]) -> None:
         finally:
             self._results_lock.release()
 
+    def _submit_tasks(
+        self,
+        remaining_runners: List[BaseRunner],
+        task_manager: TaskManager[List[TestResult]],
+    ) -> None:
+        for runner in remaining_runners:
```
squirrelsc: The runner loop can be left in the main loop method, and the method called `_submit_runner_tasks`. The logic maybe more clear, because the remaining_runners are maintained in the main loop method.

lisa/runner.py
```diff
@@ -257,77 +287,69 @@ def _start_loop(self) -> None:
         except StopIteration:
             self._log.debug(f"no more runner found, total {len(remaining_runners)}")
 
-        if self._runners:
-            run_message = notifier.TestRunMessage(
-                status=notifier.TestRunStatus.RUNNING,
-            )
-            notifier.notify(run_message)
+        if not self._runners:
```
squirrelsc: This line is not necessary now. We can remove both this check and the for loop in Ln 285. What we need to do is to add `has_more_runner` into the while conditions. So it will loop at least once. The code will be simpler.

lisa/runner.py
```diff
@@ -257,77 +287,69 @@ def _start_loop(self) -> None:
...
+            self._max_concurrency, self._callback_completed
+        )
+
+        # set the global task manager for cancellation check
+        set_global_task_manager(task_manager)
+        has_more_runner = True
+
+        # run until no idle workers are available and all runner are closed
+        while task_manager.wait_worker() or remaining_runners:
+            self._log.debug("Idle workers available. Checking runners...")
```
squirrelsc: Remove this line, because it may cause a lot of log, but not helpful a lot.

lisa/runners/lisa_runner.py
squirrelsc: Actually, I think this kind of small piece of lambda created some very quick tasks. It makes the worker busy in a very short time. So it affects the loop efficiency. I have no good idea to optimize them, but not increase complexity of class design so far.

lisa/util/parallel.py
```diff
@@ -2,29 +2,66 @@
...
+    def __init__(
+        self,
+        task_id: int,
+        task: Callable[[], T_RESULT],
+        parent_logger: Optional[Logger],
+    ) -> None:
+        self.id = task_id
+        self._task = task
+        self._log = get_logger("Task", str(self.id), parent_logger)
+        self._log.debug(f"Generate task: {self}")
```
squirrelsc, line 27: How about move this line into `__call__`? It's more accurate to see when the task start to run actually.

squirrelsc:
```python
def __call__(self) -> T_RESULT:
    self._timer = create_timer()
    # log start...
    output = self._task()
    # log end...
    return output
```

lisa/util/parallel.py
```diff
@@ -2,29 +2,66 @@
...
+
+    def close(self) -> None:
```
squirrelsc, line 29: After moved the line of log, it looks the close method is not necessary.

lisa/util/parallel.py
```diff
@@ -51,6 +88,7 @@ def wait_worker(self) -> bool:
             self._futures.remove(future)
             # exception will throw at this point
             self._callback(result)
+            self._future_task_map[future].close()
```
squirrelsc, line 107: If the close() is not necessary, the map is also not necessary too.

lisa/runner.py
```diff
@@ -257,77 +287,69 @@ def _start_loop(self) -> None:
...
-                ):
-                    # Fetch runners, if runner count is smaller than concurrency
-                    # count. It makes sure all concurrency can run.
-                    try:
-                        remaining_runners.append(next(runner_iterator))
-                    except StopIteration:
-                        has_more_runner = False
+                        time.sleep(1)
+
+                if not remaining_runners:
```
squirrelsc: Which condition is this needed? If so, add comments here.

lisa/runner.py
```diff
@@ -257,77 +287,69 @@ def _start_loop(self) -> None:
...
-                        runner.close()
-                        remaining_runners.remove(runner)
+                if task_manager.has_idle_worker():
+                    if has_more_runner:
+                        # add new runner upto max concurrency if idle workers
+                        # are available
+                        try:
+                            while True:
+                                remaining_runners.append(next(runner_iterator))
+                                if len(remaining_runners) > self._max_concurrency:
```
squirrelsc: This condition can be moved to while condition.
`while len(remaining_runners) < self._max_concurrency:`

lisa/runner.py
dsrivastavv: This is to avoid infinite loop when idle workers are available but not items in in remaining runners

lisa/runner.py
dsrivastavv: We want to atleast add one runner if there are idle workers. It's like do...while loop which runs atleast one time rather than a while loop which can run zero times

lisa/runner.py
squirrelsc: If there is no remaining runner, that means all runners are done. There shouldn't have any running task even. If there are running task, but remaining runner is empty, it should be an assertion error. So this condition may not effective.

lisa/runner.py
squirrelsc: The logic will run at first time like a do...while. The remaining runners should be empty at first time.

lisa/runner.py
squirrelsc: We may add a check that, `not task_manager.has_idle_worker() and not has_more_runner`

lisa/runner.py
dsrivastavv: But what about the case when len(runner) == self._max_concurrency but no runners have any tasks available? In this scenario we need to atleast add one new runner.

lisa/util/parallel.py
dsrivastavv: We still need close method to log the time taken from starting the task to when it joined with the root thread. This would be helpful to debug if any task is taking a long time.

lisa/runner.py
squirrelsc: Is it changed?

lisa/runner.py
```diff
@@ -247,87 +255,76 @@ def _callback_completed(self, results: List[TestResult]) -> None:
...
+        task_manager = TaskManager[List[TestResult]](
+            self._max_concurrency, self._callback_completed
+        )
+
+        # set the global task manager for cancellation check
+        set_global_task_manager(task_manager)
+        has_more_runner = True
+
+        # run until no idle workers are available and all runner are closed
+        while task_manager.wait_worker() or has_more_runner:
```
squirrelsc: It still need check `remaining_runners` like below. 
`while task_manager.wait_worker() or has_more_runner or remaining_runners:`

somil55: This will go in the first while loop always, but re-loop only when it has no idle workers left. So it works fine even without adding `or remaining_runners` condition.

dsrivastavv: Fixed

selftests/runners/test_lisa_runner.py
```diff
@@ -632,7 +633,10 @@ def _run_all_tests(self, runner: LisaRunner) -> List[TestResult]:
         while not runner.is_done:
             task = runner.fetch_task()
             if task:
-                temp_test_results = task()
+                if isinstance(task, Task):
+                    temp_test_results = task()
+                if isinstance(task, List):
```
squirrelsc: `elif`

lisa/util/parallel.py
squirrelsc: The closed one may need to be removed from the map. It may create a reference, so some resource are not recycled.

lisa/util/parallel.py
```diff
@@ -2,29 +2,76 @@
...
+        self._log = get_logger("Task", str(self.id), parent_logger)
+        self._log.debug(f"Generate task: {self}")
+
+    def close(self) -> None:
+        self._lifecycle_time = self._lifecycle_timer.elapsed_text()
+        self._log.debug(
+            f"Task finished. "
+            f"Lifecycle time: {self._lifecycle_time} "
+            f"Wait time before call: {self._wait_time}"
+            f"Call time: {self._call_time}"
```
squirrelsc: Add one more time, it presents how long it waits after call completed, but not returned.

`self._lifecycle_time-self._wait_time-self._call_time`

lisa/util/parallel.py
```diff
@@ -2,29 +2,76 @@
...
+        )
+
+    def __call__(self) -> T_RESULT:
+        self._call_timer = create_timer()
+        self._wait_time = self._wait_timer.elapsed_text()
```
squirrelsc: It can be called like `self._wait_timer.elapsed()`. It will record the current time. After that, when it's called like `self._wait_timer.elapsed_text()` again, it's the same time. So it doesn't need to save the string in a variable.

selftests/runners/test_lisa_runner.py
dsrivastavv: This is also present in the runner file. I will fix it there as well

lisa/runner.py
squirrelsc: The wait_worker will return false, when there is no running task. There is rare condition that, there is no worker is running, but some runners are not declared to done. That may happen when the whole run is near to complete. The `has_more_runner` cannot block this kind of case.

lisa/util/parallel.py
```diff
@@ -51,6 +104,7 @@ def wait_worker(self) -> bool:
             self._futures.remove(future)
             # exception will throw at this point
             self._callback(result)
```
squirrelsc, line 106: The closed one may need to be removed from the map. It may create a reference, so some resource are not recycled.

squirrelsc (approved):

PR #1558, opened 2021-10-01
Ignore warnings in "make kernelversion"

SRIKKANTH:


squirrelsc:
@SRIKKANTH Let's work on the same PR next time. Do not open another PR for the same change. It helps to track discussions.

squirrelsc (approved):

PR #1559, opened 2021-10-04
DPDK sender/receiver test, threaded node init, threaded testpmd run

mcgov:
sender/receiver test, some major refactoring of init to allow for multiple nodes to init concurrently, also added a testpmd concurrent runner.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -1,0 +1,249 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import ipaddress
+import itertools
+import re
+from typing import Any, Dict, List, Tuple
+
+from assertpy import assert_that, fail
+
+from lisa import Node
+from lisa.util import InitializableMixin, LisaException
+
+
+class NicInfo:
+
+    # Class for info about an single upper/lower nic pair.
+    # devices using SRIOV on azure typically have an upper synthetic device
+    # paired with a lower SRIOV Virtual Function (VF) device that
+    # enables the passthrough to the physical NIC.
+
+    def __init__(
+        self,
+        upper: str,
+        lower: str = "",
+        pci_slot: str = "",
+    ) -> None:
+        self.has_lower = lower != "" or pci_slot == ""
+        self.upper = upper
+        self.lower = lower
+        self.pci_slot = pci_slot
+        self.ip_addr = ""
+        self.mac_addr = ""
+
+    def __str__(self) -> str:
+        return (
+            "NicInfo:\n"
+            f"upper: {self.upper}\n"
+            f"lower: {self.lower}\n"
+            f"pci_slot: {self.pci_slot}\n"
+            f"ip_addr: {self.ip_addr}\n"
+            f"mac_addr: {self.mac_addr}\n"
+        )
+
+
+class Nics(InitializableMixin):
+
+    # Class for all of the nics on a node. Contains multiple NodeNic classes.
+    # Init identifies upper/lower paired devices and the pci slot info for the lower.
+
+    # for parsing ip addr show (ipv4)
+    # ex:
+    """
+    eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group
+        default qlen 1000
+    link/ether 60:45:bd:86:d4:88 brd ff:ff:ff:ff:ff:ff
+    inet 10.57.0.4/24 brd 10.57.0.255 scope global eth0
+       valid_lft forever preferred_lft forever
+    inet6 fe80::6245:bdff:fe86:d488/64 scope link
+       valid_lft forever preferred_lft forever
+    """
+    __ip_regex = re.compile(
+        r"inet\s+"  # looking for ip address in output
+        r"([0-9a-fA-F]{1,3}\."  # capture ipv4 address
+        r"[0-9a-fA-F]{1,3}\."
+        r"[0-9a-fA-F]{1,3}\."
+        r"[0-9a-fA-F]{1,3})"
+    )
+    __mac_regex = re.compile(
+        r"ether\s+"  # looking for the ether address
+        r"([0-9a-fA-F]{2}:"  # capture mac address
+        r"[0-9a-fA-F]{2}:"
+        r"[0-9a-fA-F]{2}:"
+        r"[0-9a-fA-F]{2}:"
+        r"[0-9a-fA-F]{2}:"
+        r"[0-9a-fA-F]{2})"
+    )
+
+    # capturing from ip route show
+    # ex:
+    #    default via 10.57.0.1 dev eth0 proto dhcp src 10.57.0.4 metric 100
+    __dev_regex = re.compile(
+        r"default via\s+"  # looking for default route
+        r"[0-9a-fA-F]{1,3}\."  # identify ip address
+        r"[0-9a-fA-F]{1,3}\."
+        r"[0-9a-fA-F]{1,3}\."
+        r"[0-9a-fA-F]{1,3}"
+        r"\s+dev\s+"  # looking for the device for the default route
+        r"([a-zA-Z0-9]+)"  # capture device
+    )
+
+    def __init__(self, node: Node):
+        super().__init__()
+        self._node = node
+        self._nics: Dict[str, NicInfo] = dict()
+
+    def __str__(self) -> str:
+        _str = ""
+        for nic in self._nics:
+            _str += f"{self._nics[nic]}"
+        return _str
+
+    def __len__(self) -> int:
+        return len(self._nics)
+
+    def append(self, next_node: NicInfo) -> None:
+        self._nics[next_node.upper] = next_node
+
+    def is_empty(self) -> bool:
+        return len(self._nics) == 0
+
+    def get_unpaired_devices(self) -> List[str]:
+        return [x.upper for x in self._nics.values() if not x.has_lower]
+
+    def get_upper_nics(self) -> List[str]:
+        return [self._nics[x].upper for x in self._nics.keys()]
+
+    def get_lower_nics(self) -> List[str]:
+        return [self._nics[x].lower for x in self._nics.keys()]
+
+    def get_device_slots(self) -> List[str]:
+        return [self._nics[x].pci_slot for x in self._nics.keys()]
+
+    def get_nic(self, nic_name: str) -> NicInfo:
+        return self._nics[nic_name]
+
+    def get_test_nic(self) -> Tuple[int, NicInfo]:
+        # convenience method
+        # get the 'last' nic in the list of nics
+        number_of_nics = len(self.get_upper_nics())
+        assert_that(number_of_nics).is_greater_than(0)
+        # will be used for tests with a single active vf, so id = 0
+        return (0, self._nics[self.get_upper_nics()[number_of_nics - 1]])
+
+    def nic_info_is_present(self, nic_name: str) -> bool:
+        return nic_name in self.get_upper_nics() or nic_name in self.get_lower_nics()
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._nic_names = self._get_nic_names()
+        self._get_node_nic_info()
+        self.default_nic = self._get_default_nic()
+        self._get_host_if_info()
+
+    def _get_nic_names(self) -> List[str]:
+        # identify all of the nics on the device, excluding tunnels and loopbacks etc.
+        result = self._node.execute(
+            " ls /sys/class/net/ | grep -Ev $(ls /sys/devices/virtual/net)",
+            shell=True,
+            sudo=True,
+        )
+        nic_names = result.stdout.splitlines()
+        for item in nic_names:
+            assert_that(item).described_as(
+                "nic name could not be found"
+            ).is_not_equal_to("")
+        return nic_names
+
+    def _get_nic_device(self, nic_name: str) -> str:
+        slot_info_result = self._node.execute(
+            f"readlink /sys/class/net/{nic_name}/device"
+        )
+        slot_info_result.assert_exit_code()
+        base_device_result = self._node.execute(f"basename {slot_info_result.stdout}")
+        base_device_result.assert_exit_code()
+        # todo check addr matches expectation
+        return base_device_result.stdout
+
+    def _get_node_nic_info(self) -> None:
+        # Identify which nics are slaved to master devices.
+        # This should be really simple with /usr/bin/ip but experience shows
+        # the tool isn't super consistent across distros in this regard
+
+        # use sysfs to gather upper/lower nic pairings and pci slot info
+        for pairing in itertools.permutations(self._nic_names, 2):
+            upper_nic, lower_nic = pairing
+            # check a nic pairing to identify upper/lower relationship
+            upper_check = self._node.execute(
+                f"readlink /sys/class/net/{lower_nic}/upper_{upper_nic}"
+            )
+            if upper_check.exit_code == 0:
+                assert_that(upper_check.stdout).is_not_equal_to("")
+                pci_slot = self._get_nic_device(lower_nic)
+                assert_that(pci_slot).is_not_empty()
+                # check pcislot info looks correct
+                nic_info = NicInfo(upper_nic, lower_nic, pci_slot)
+                self.append(nic_info)
+
+        # identify nics which don't have a pairing (non-AN devices)
+        for nic in self._nic_names:
+            if not self.nic_info_is_present(nic):
+                self.append(NicInfo(nic))
+
+        assert_that(self).described_as(
+            (
+                "During Lisa nic info initialization, Nics class could not "
+                f"find any nics attached to {self._node.name}."
+            )
+        ).is_not_empty()
+
+    def _get_default_nic(self) -> str:
+        cmd = "/sbin/ip route"
+        ip_route_result = self._node.execute(cmd, shell=True, sudo=True)
+        ip_route_result.assert_exit_code()
+        assert_that(ip_route_result.stdout).is_not_empty()
+        dev_match = self.__dev_regex.search(ip_route_result.stdout)
+        if not dev_match or not dev_match.groups():
+            raise LisaException(
+                "Could not locate default network interface"
+                f" in output:\n{ip_route_result.stdout}"
+            )
+        assert_that(dev_match.groups()).is_length(1)
+        default_interface_name = dev_match.group(1)
+        assert_that(default_interface_name in self._nic_names).described_as(
+            (
+                f"ERROR: NIC name found as default {default_interface_name} "
+                f"was not in original list of nics {repr(self._nic_names)}."
+            )
+        ).is_true()
+        return default_interface_name
+
+    def _get_host_if_info(self) -> None:
+        for nic in self.get_upper_nics():
+            # get ip and mac
+            result = self._node.execute(f"/sbin/ip addr show {nic}", shell=True)
+            result.assert_exit_code()
+            ip_match = self.__ip_regex.search(result.stdout)
+            mac_match = self.__mac_regex.search(result.stdout)
+
+            if ip_match and mac_match:
+                # check we found matches for both
+                for match in [ip_match, mac_match]:
+                    assert_that(match.groups()).described_as(
+                        (
+                            f"(IP) Trouble parsing `ip addr show {nic}` output."
+                            " Number of match groups was unexpected."
+                        )
+                    ).is_length(1)
+
+                ip_addr = ip_match.group(1)
+                mac_addr = mac_match.group(1)
+
+                # double check IP address looks right
+                ipaddress.ip_address(ip_addr)
+
+                # save them both off
+                self.get_nic(nic).ip_addr = ip_addr
+                self.get_nic(nic).mac_addr = mac_addr
+            else:
+                fail(f"Could not parse output of ip addr show {nic}")
--- a/lisa/util/parallel.py
+++ b/lisa/util/parallel.py
@@ -1,8 +1,15 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-
-from concurrent.futures import FIRST_COMPLETED, Future, ThreadPoolExecutor, wait
+from concurrent.futures import (
+    ALL_COMPLETED,
+    FIRST_COMPLETED,
+    Future,
+    ThreadPoolExecutor,
+    wait,
+)
 from typing import Any, Callable, Dict, Generic, List, Optional, TypeVar
+
+from assertpy import assert_that
 
 from lisa.util.logger import Logger, get_logger
 from lisa.util.perf_timer import create_timer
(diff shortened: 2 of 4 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -55,10 +67,55 @@ def check_dpdk_build(self, node: Node, log: Logger) -> None:
...
+            Tests a basic sender/receiver setup.
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_nic_count=2,
+            network_interface=Sriov,
+            min_count=2,
+        ),
+    )
+    def check_2node_snd_rcv(self, environment: Environment, log: Logger) -> None:
```
squirrelsc: 1. use full name on method names. They will be test case name by default.
2. Start with "verify", so it can align with other test cases.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -55,10 +67,55 @@ def check_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        test_kits = _init_nodes_concurrent(environment, log)
+        sender, receiver = test_kits
+
+        # helpful to have the public ip for debugging
+        external_ips = [
+            x.node.execute("curl ifconfig.me").stdout for x in [sender, receiver]
```
squirrelsc: The IP can get from connection info of node. It doesn't need to get from outside.

mcgov: what is the connection info get method? I don't see a parameter in Node for that

squirrelsc: It exists in RemoteNode. You can check it by `isinstance(node, RemoteNode)`. If it's LocalNode, skip this test case. After that, try `node.connection_info` get the information.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -257,4 +380,85 @@ def initialize_node_resources(node: Node, log: Logger) -> NodeResources:
...
+            threading.Thread(
+                target=_threaded_testpmd, args=(pair, seconds, log, output)
+            )
+        )
+    for thread in threads:
+        thread.start()
+    threads.reverse()
+    for thread in threads:
+        thread.join()
+    return output
```
squirrelsc, line 225: The `parallel.py` has a concurrent implementation, it can be reused here. You can expose the `return_when` in the `wait_worker` to support `FIRST_COMPLETED` and `ALL_COMPLETED`.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -55,10 +67,55 @@ def check_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        log.info(testpmd_outputs[receiver])
+        log.info(testpmd_outputs[sender])
+
+        rcv_rx_pps = _get_rx_pps_from_testpmd_output(testpmd_outputs[receiver])
+        snd_tx_pps = _get_tx_pps_from_testpmd_output(testpmd_outputs[sender])
+        log.info(f"receiver rx-pps: {rcv_rx_pps}")
+        log.info(f"sender tx-pps: {snd_tx_pps}")
+
+        # differences in NIC model throughput can lead to different snd/rcv counts
+        assert_that(rcv_rx_pps).is_greater_than(2 ** 20)
```
squirrelsc: add actionable description on the assertions. What it means, if it's failed.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -55,10 +67,55 @@ def check_dpdk_build(self, node: Node, log: Logger) -> None:
             f"TX-PPS ({tx_pps}) should have been greater than 2^20 (~1m) PPS."
         ).is_greater_than(2 ** 20)
 
-    def _execute_expect_zero(self, node: Node, cmd: str) -> str:
-        result = node.execute(cmd, sudo=True, shell=True)
-        result.assert_exit_code()
-        return result.stdout
+    @TestCaseMetadata(
+        description="""
+            Tests a basic sender/receiver setup.
```
squirrelsc, line 72: Add how it's tested.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -55,10 +67,55 @@ def check_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        test_kits = _init_nodes_concurrent(environment, log)
+        sender, receiver = test_kits
+
+        # helpful to have the public ip for debugging
+        external_ips = [
+            x.node.execute("curl ifconfig.me").stdout for x in [sender, receiver]
+        ]
+        log.debug((f"\nsender:{external_ips[0]}\nreceiver:{external_ips[1]}\n"))
+
+        snd_nic, rcv_nic = [x.node_nic_info.get_nic("eth1") for x in test_kits]
```
squirrelsc: It looks hard code to the eth1. What's the reason to add like this way? Add comments at least.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -42,11 +54,11 @@ def check_dpdk_build(self, node: Node, log: Logger) -> None:
 
         # grab a nic and run testpmd
         test_nic = node_nic_info.get_nic(node_nic_info.get_upper_nics()[-1])
-        vdev_type = "net_vdev_netvsc0"
-        testpmd_include_str = test_nic.testpmd_include(vdev_type)
+
+        testpmd_include_str = test_nic.testpmd_include(VDEV_TYPE)
         testpmd_cmd = testpmd._generate_testpmd_command(testpmd_include_str, "txonly")
```
squirrelsc: the `_generate_testpmd_command` should be a public method, because it's accessed out of the class.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -257,4 +380,85 @@ def initialize_node_resources(node: Node, log: Logger) -> NodeResources:
...
+            target=_threaded_resource_init, args=((node, log, test_kits))
+        )
+        thread.start()
+        threads.append(thread)
+    for thread in threads:
+        thread.join()
+    return test_kits
+
+
+def _get_tx_pps_from_testpmd_output(output: str) -> int:
```
squirrelsc: What's the reason to move it out of the tool?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I don't see a file named parallel.py in lisa, is this an external module or something?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -176,9 +242,9 @@ def _get_node_nic_info(self, node: Node, nic_list: List[str]) -> None:
         assert_that(self.is_empty()).is_false()
 
     def _get_default_nic(self, node: Node, nic_list: List[str]) -> str:
```
squirrelsc: The nic is important in our test code. Can you move this class into a file called `nic.py` in lisa level? We may refine the code later, but it's a beginning to share more logic like get mac address and other information of a nic.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: it doesn't depend on the tool instance

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I would be honored

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: thanks!

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Oop, no reason. Will fix.

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: If it's part of the tool, the `output` doesn't need to be pass through from the test code. It can be simple like 

`def get_tx_pps() ->int:`

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Wow that was a lot easier than I expected. It would make sense to me if the Nic eventually is just accessible through the node class. Will send up the refactor soon after some testing.

mcgov:
Need to make some changes, will ping when they're ready

mcgov:
@squirrelsc ready for another review round 🙂

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+        self._upper = upper
+        self._lower = lower
+        self._pci_slot = pci_slot
+        self._ip_addr = ""
+        self._mac_addr = ""
+
+    def has_secondary(self) -> bool:
+        return self._has_secondary
+
+    def __str__(self) -> str:
```
squirrelsc, line 35: put the meta method like `__aaa__` before public method.

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+        pci_slot: str = "",
+    ) -> None:
+        self._has_secondary = lower != "" or pci_slot == ""
+        self._upper = upper
+        self._lower = lower
+        self._pci_slot = pci_slot
+        self._ip_addr = ""
+        self._mac_addr = ""
+
+    def has_secondary(self) -> bool:
```
squirrelsc: Does "secondary" mean "lower" here? Make terms consistence. If we takes the upper and lower parts, make sure it's generic, not Azure only.

squirrelsc: And it can be a property and calculate from the condition every time. It can reduce the logic to maintain things consistent.

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+        return (
+            "NicInfo:\n"
+            f"upper: {self._upper}\n"
+            f"lower: {self._lower}\n"
+            f"pci_slot: {self._pci_slot}\n"
+            f"ip_addr: {self._ip_addr}\n"
+            f"mac_addr: {self._mac_addr}\n"
+        )
+
+    def testpmd_include(self, vdev_type: str) -> str:
```
squirrelsc: It looks this method generate a command string for testpmd. Consider to move this method to testpmd.

1. the `descrbed_as` should be before of `is_true`, or it cannot be printed, because `is_true` failed first.
2. make the method name like "verb_noun", and straightforward. It's hard to know what happens, if called this method.
3. add comments, if it's really not easy to get the method name clear.

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+        return f'--vdev="{vdev_type},iface={self._upper}" --allow "{self._pci_slot}"'
+
+
+class NodeNics:
+
+    # Class for all of the nics on a node. Contains multiple NodeNic classes.
+    # Init identifies upper/lower paired devices and the pci slot info for the lower.
+
+    def _get_nic_names(self) -> List[str]:
```
squirrelsc, line 144: private methods should be behind of public methods.

lisa/util/parallel.py
```diff
@@ -108,6 +109,10 @@ def wait_worker(self) -> bool:
             self._future_task_map.pop(future)
         return len(self._futures) > 0
 
+    def wait_for_all_workers(self) -> None:
```
squirrelsc, line 118: It's not a good way to wait all workers. Use a variable at line 100, replace the hardcoded `FIRST_COMPLETED`.

`done_futures, _ = wait(self._futures[:], return_when=FIRST_COMPLETED)`

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+
+                # save them both off
+                self.get_nic(nic)._ip_addr = ip_addr
+                self.get_nic(nic)._mac_addr = mac_addr
+            else:
+                assert_that(False).described_as(
+                    f"Could not parse output of ip addr show {nic}"
+                ).is_true()
+
+    def __init__(self, node: Node):
```
squirrelsc: `__init__` should be first method in the class.

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+        self._node = node
+        self._nics: Dict[str, NicInfo] = dict()
+        nics = self._get_nic_names()
+        self._get_node_nic_info(nics)
+        default_nic = self._get_default_nic(nics)
+        awk_cmd = "awk '{print $2}'"
```
squirrelsc: Don't execute any time cost method in `__init__` method. it's a construction method. Use `InitializableMixin` pattern, and initialize it with node.

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+        assert_that(self._has_secondary).is_true().described_as(
+            (
+                f"This interface {self._upper} does not have a lower interface "
+                "and pci slot associated with it. Aborting."
+            )
+        )
+        return f'--vdev="{vdev_type},iface={self._upper}" --allow "{self._pci_slot}"'
+
+
+class NodeNics:
```
squirrelsc: How about call it `Nics`? Later, we can make it as a member of Node.

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def _get_nic_names(self) -> List[str]:
+        # identify all of the nics on the device, excluding tunnels and loopbacks etc.
+        result = self._node.execute(
+            " ls /sys/class/net/ | grep -Ev $(ls /sys/devices/virtual/net)",
+            shell=True,
+            sudo=True,
+        )
+        nic_names = result.stdout.splitlines()
+        for item in nic_names:
```
squirrelsc, line 70: It's better to have a list of Nic, and fill them information. The methods in this class are associated by name, but we can associate them by object. It's easier to use.

lisa/nic.py
```diff
@@ -0,0 +1,222 @@
...
+        slot_info_result = self._node.execute(
+            f"readlink /sys/class/net/{nic_name}/device"
+        )
+        slot_info_result.assert_exit_code()
+        base_device_result = self._node.execute(f"basename {slot_info_result.stdout}")
+        base_device_result.assert_exit_code()
+        # todo check addr matches expectation
+        return base_device_result.stdout
+
+    def _get_node_nic_info(self, nic_list: List[str]) -> None:
```
squirrelsc: 1. It doesn't need to pass in a couple of names, those names can be queried in this method.
2. If it doesn't return a value, call it like `_initialize`. It can inherit from `InitializableMixin`. If it calls `get_...`, let it return the dict of nics.

lisa/nic.py
mcgov: yep, I'll change it

lisa/nic.py
mcgov: Will move to testpmd, this makes it much cleaner! Good eye thank you

lisa/util/parallel.py
mcgov: how do you like something like this:
```
    def wait_worker(self, return_condition: str = FIRST_COMPLETED) -> bool:
        """
        Return:
            True, if there is running worker.
        """

        done_futures, _ = wait(self._futures[:], return_when=return_condition)
        for future in done_futures:
            # join exceptions of subthreads to main thread
            result = future.result()
            # removed finished threads
            self._futures.remove(future)
            # exception will throw at this point
            self._callback(result)
            self._future_task_map[future].close()
            self._future_task_map.pop(future)
        return len(self._futures) > 0

    def wait_for_all_workers(self) -> None:
        remaining_workers = self.wait_worker(return_condition=ALL_COMPLETED)
        assert_that(remaining_workers).is_zero()

```

lisa/nic.py
mcgov: Ah, sorry. I guess this one slipped through before. Will fix.

lisa/util/parallel.py
squirrelsc: Yes, there would be two line changes only.

mcgov:
@squirrelsc ready for another review round

lisa/util/parallel.py
```diff
@@ -108,6 +115,10 @@ def wait_worker(self) -> bool:
...
+        remaining_workers = self.wait_worker(return_condition=ALL_COMPLETED)
```
squirrelsc: `remaining_workers` -> `remaining_worker_count`

lisa/nic.py
```diff
@@ -0,0 +1,221 @@
...
+                except ValueError:
+                    assert_that(False).described_as(
+                        f"{ip_addr} was not recognized as an IP Address."
+                    ).is_true()
+
+                # save them both off
+                self.get_nic(nic)._ip_addr = ip_addr
+                self.get_nic(nic)._mac_addr = mac_addr
+            else:
+                assert_that(False).described_as(
```
squirrelsc: Raise LisaException here, another way is to call `fail` in assertion package.

squirrelsc:
Please check unresolved comments. If fixed, don't resolve them, I will do that. If it's needed, you can use reaction to track which comments is fixed. Thank you.

lisa/nic.py
```diff
@@ -0,0 +1,221 @@
...
+    # paired with a lower SRIOV Virtual Function (VF) device that
+    # enables the passthrough to the physical NIC.
+
+    def __init__(
+        self,
+        upper: str,
+        lower: str = "",
+        pci_slot: str = "",
+    ) -> None:
+        self._has_lower = lower != "" or pci_slot == ""
```
squirrelsc: Because those fields are accessed outside for the class, so they should be public without the underscore.

lisa/nic.py
```diff
@@ -0,0 +1,221 @@
...
+    def __len__(self) -> int:
+        return len(self._nics)
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._nic_names = self._get_nic_names()
+        self._get_node_nic_info()
+        default_nic = self._get_default_nic()
+        awk_cmd = "awk '{print $2}'"
+        self.ip_addr = self._node.execute(
+            f"/sbin/ip addr show {default_nic} | grep inet | {awk_cmd}"
```
squirrelsc: It's better to call command and use regular expression to analyze the result. It can dump the raw output in log for troubleshooting. If there is no special reason, avoid to use pipes.

lisa/nic.py
```diff
@@ -0,0 +1,221 @@
...
+
+                ip_addr = ip_match.group(1)
+                mac_addr = mac_match.group(1)
+
+                # double check IP address looks right
+                try:
+                    ipaddress.ip_address(ip_addr)
+                except ValueError:
+                    assert_that(False).described_as(
+                        f"{ip_addr} was not recognized as an IP Address."
```
squirrelsc: The raw exception may provide the raw text of wrong ip address.. Let it raise Exception, it doesn't need the extra try/except. because it doesn't provide more information.

lisa/nic.py
mcgov: It's weird to me there isn't a FailedException 😂 We have passed, skipped, and lisa.
I like the way fail() reads so I'll go with that one.

mcgov:
@squirrelsc ready for another round of review

lisa/nic.py
squirrelsc: The LisaException is the fail exception. By default, all the exception, assertion errors mean failed. Some other exceptions are passed, skipped.

lisa/nic.py
squirrelsc: Is it changed?

lisa/nic.py
squirrelsc: Is it fixed?

mcgov: yep

lisa/nic.py
mcgov: yep

lisa/nic.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # identify nics which don't have a pairing (non-AN devices)
+        for nic in self._nic_names:
+            if not self.nic_info_is_present(nic):
+                self.append(NicInfo(nic))
+
+        assert_that(self.is_empty()).is_false()
+
+    def _get_default_nic(self) -> str:
+        _dev_regex = (
```
squirrelsc: 1. In usual, we define the regex in class level like `re.compile`. So it will be precompiled, and used on different data.
2. The regex is not easy to modified, so we add examples above the regex. So when we found new case, we can test it with old examples together.

lisa/nic.py
```diff
@@ -0,0 +1,221 @@
...
+        ip_route_result.assert_exit_code()
+        assert_that(ip_route_result.stdout).is_not_empty()
+        dev_match = re.search(_dev_regex, ip_route_result.stdout)
+        if not dev_match or not dev_match.groups():
+            fail(
+                "Could not locate default network interface"
+                f" in output:\n{ip_route_result.stdout}"
+            )
+            # the type checker isn't aware of assert_that so it needs a return here
+            return ""
```
squirrelsc: How about use LisaExcepiton? The Exceptions will be recognized.

lisa/nic.py
```diff
@@ -0,0 +1,221 @@
...
+            # get ip and mac
+            result = self._node.execute(f"/sbin/ip addr show {nic}", shell=True)
+            result.assert_exit_code()
+            ip_match = re.search(_ip_regex, result.stdout)
+            mac_match = re.search(_mac_regex, result.stdout)
+
+            if ip_match and mac_match:
+                # check we found matches for both
+                for match in [ip_match, mac_match]:
+                    assert_that(len(match.groups())).described_as(
```
squirrelsc: Try below pattern. It will provide more details when it's failed.
`assert_that(match.groups())......is_length()`

mcgov:
@squirrelsc comments addressed

lisa/nic.py
```diff
@@ -0,0 +1,218 @@
...
+
+        # identify nics which don't have a pairing (non-AN devices)
+        for nic in self._nic_names:
+            if not self.nic_info_is_present(nic):
+                self.append(NicInfo(nic))
+
+        assert_that(self.is_empty()).is_false()
+
+    def _get_default_nic(self) -> str:
+        _dev_regex = re.compile(
```
squirrelsc: Refer to other usage of `re` in code

1. Define it in class level.
2. Add parsed text examples above.

lisa/nic.py
```diff
@@ -0,0 +1,218 @@
...
+            (
+                f"ERROR: NIC name found as default {default_interface_name} "
+                f"was not in original list of nics {repr(self._nic_names)}."
+            )
+        ).is_true()
+        return default_interface_name
+
+    def _get_host_if_info(self) -> None:
+        # for parsing ip addr show (ipv4)
+        _ip_regex = re.compile(
```
squirrelsc: Similar to above comments.

lisa/nic.py
```diff
@@ -0,0 +1,218 @@
...
+                # check pcislot info looks correct
+                nic_info = NicInfo(upper_nic, lower_nic, pci_slot)
+                self.append(nic_info)
+
+        # identify nics which don't have a pairing (non-AN devices)
+        for nic in self._nic_names:
+            if not self.nic_info_is_present(nic):
+                self.append(NicInfo(nic))
+
+        assert_that(self.is_empty()).is_false()
```
squirrelsc: add describe_as or use LisaException with message. If we don't explain that, the failure is hard to understand, when it's raised.

mcgov:
updated regex to be more clear and have comments

squirrelsc (approved):

PR #1560, opened 2021-10-04
[LISAv3] Adding container test suite

isaiahduncan:
Adds docker java app and docker python app tests from container test suite
Adds Docker tool (based off of container utils from v2)

### DIFF
--- a/lisa/tools/docker.py
+++ b/lisa/tools/docker.py
@@ -1,0 +1,74 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+from lisa.operating_system import Debian, Redhat
+from lisa.tools.service import Service
+from lisa.util import LisaException
+
+
+class Docker(Tool):
+    @property
+    def command(self) -> str:
+        return "docker"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def build_image(
+        self, docker_image_name: str = "testing", dockerfile: str = "Dockerfile"
+    ) -> None:
+        self.run(
+            f"build -t {docker_image_name} -f {dockerfile} .",
+            shell=True,
+            sudo=True,
+            cwd=self.node.working_path,
+            force_run=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Docker image build failed.",
+        )
+
+    def remove_container(self, container_name: str) -> None:
+        self._log.debug(f"Removing Docker Container {container_name}")
+        self.run(f"rm {container_name}", sudo=True, force_run=True)
+
+    def remove_image(self, image_name: str) -> None:
+        self._log.debug(f"Removing Docker Image {image_name}")
+        self.run(f"rmi {image_name}", sudo=True, force_run=True)
+
+    def run_container(
+        self,
+        docker_image_name: str,
+        docker_container_name: str,
+        docker_run_output: str,
+    ) -> None:
+        self.run(
+            f"run --name {docker_container_name} "
+            f"{docker_image_name} 1> {docker_run_output} 2>&1",
+            shell=True,
+            sudo=True,
+            cwd=self.node.working_path,
+            force_run=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Docker run failed.",
+        )
+
+    def start(self) -> None:
+        self._log.debug("Start docker engine")
+        service = self.node.tools[Service]
+        service.enable_service("docker")
+        service.restart_service("docker")
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Debian):
+            self.node.os.install_packages("docker.io")
+        elif isinstance(self.node.os, Redhat):
+            self.node.os.install_packages(
+                ["docker", "docker-ce", "docker.socket", "docker.service"]
+            )
+        else:
+            raise LisaException(f"{self.node.os.information.vendor} not supported")
+
+        self.start()
+        return True
--- a/microsoft/testsuites/docker/dockersuite.py
+++ b/microsoft/testsuites/docker/dockersuite.py
@@ -1,0 +1,151 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import os
+from pathlib import Path
+
+from assertpy import assert_that
+
+from lisa import Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa.operating_system import Redhat
+from lisa.tools import Docker
+from lisa.util import SkippedException
+
+
+@TestSuiteMetadata(
+    area="docker",
+    category="functional",
+    description="""
+    This test suite runs the docker test cases for java, python, dotnet, and wordpress.
+    """,
+)
+class docker(TestSuite):  # noqa
+    @TestCaseMetadata(
+        description="""
+            This test case creates and runs a java app using docker
+
+            Steps:
+            1. Install Docker
+            2. Copy java dockerfile and program file to node
+            3. Create docker image and run docker container
+            4. Check results of docker run against java string identifier
+    """,
+        priority=1,
+    )
+    def docker_java_app(self, node: Node) -> None:
+        self._execute_docker_test(
+            node,
+            "javaappimage",
+            "javaapp",
+            "Hello world from java",
+            "Main.java",
+            "java.Dockerfile",
+        )
+
+    @TestCaseMetadata(
+        description="""
+            This test case creates and runs a python app using docker
+
+            Steps:
+            1. Install Docker
+            2. Copy python dockerfile and program file to node
+            3. Create docker image and run docker container
+            4. Check results of docker run against python string identifier
+        """,
+        priority=3,
+    )
+    def docker_python_app(self, node: Node) -> None:
+        self._execute_docker_test(
+            node,
+            "pythonappimage",
+            "pythonapp",
+            "Hello world from python",
+            "helloworld.py",
+            "python.Dockerfile",
+        )
+
+    def _copy_to_node(self, node: Node, filename: str) -> None:
+        file_path = Path(os.path.dirname(__file__)) / "TestScripts" / filename
+        if not node.shell.exists(node.working_path / filename):
+            node.shell.copy(file_path, node.working_path / filename)
+
+    def _execute_docker_test(
+        self,
+        node: Node,
+        docker_image_name: str,
+        docker_container_name: str,
+        string_identifier: str,
+        prog_src: str,
+        dockerfile: str,
+    ) -> None:
+        docker_tool = self._verify_and_remove_containers(
+            node, docker_image_name, docker_container_name
+        )
+        self._copy_to_node(node, prog_src)
+        self._copy_to_node(node, dockerfile)
+        self._run_and_verify_results(
+            node,
+            docker_tool,
+            dockerfile,
+            docker_image_name,
+            docker_container_name,
+            string_identifier,
+        )
+
+    def _run_and_verify_results(
+        self,
+        node: Node,
+        docker_tool: Docker,
+        dockerfile_name: str,
+        docker_image_name: str,
+        docker_container_name: str,
+        string_identifier: str,
+    ) -> None:
+        docker_run_output_file = "docker_run.log"
+
+        docker_tool.build_image(docker_image_name, dockerfile_name)
+        docker_tool.run_container(
+            docker_image_name, docker_container_name, docker_run_output_file
+        )
+
+        docker_tool.remove_image(docker_image_name)
+        docker_tool.remove_container(docker_container_name)
+
+        docker_run_output = node.execute(
+            f"cat {docker_run_output_file}",
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Docker run output file not found",
+            cwd=node.working_path,
+        ).stdout
+        assert_that(docker_run_output).described_as(
+            "The container didn't output expected result. "
+            "There may have been errors when the container was run. "
+            f"String Identifier: {string_identifier}. "
+            f"Docker Run Output: {docker_run_output}."
+        ).is_equal_to(string_identifier)
+
+    def _skip_if_not_supported(self, node: Node) -> None:
+        if isinstance(node.os, Redhat) and node.os.information.version < "7.0.0":
+            raise SkippedException(
+                f"Test not supported for RH/CentOS {node.os.information.release}"
+            )
+
+    def _verify_and_remove_containers(
+        self, node: Node, docker_image_name: str, docker_container_name: str
+    ) -> Docker:
+        self._skip_if_not_supported(node)
+        docker_tool = node.tools[Docker]
+        self._verify_docker_engine(node)
+        docker_tool.remove_image(docker_image_name)
+        docker_tool.remove_container(docker_container_name)
+        return docker_tool
+
+    def _verify_docker_engine(self, node: Node) -> None:
+        node.log.debug(f"VerifyDockerEngine on {node.os.information.vendor.lower()}")
+        result = node.execute(
+            "docker run hello-world",
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Fail to run docker run hello-world",
+            sudo=True,
+        )
+        node.log.debug(f"VerifyDockerEngine: hello-world output - {result.stdout}")
--- a/lisa/tools/__init__.py
+++ b/lisa/tools/__init__.py
@@ -6,6 +6,7 @@ from lisa.base_tools import Cat, Sed, Uname, Wget
 from .chrony import Chrony
 from .date import Date
 from .dmesg import Dmesg
+from .docker import Docker
 from .echo import Echo
 from .ethtool import Ethtool
 from .fdisk import Fdisk
@@ -41,6 +42,7 @@ __all__ = [
     "Chrony",
     "Date",
     "Dmesg",
+    "Docker",
     "Echo",
     "Ethtool",
     "Fdisk",
--- a/microsoft/testsuites/docker/TestScripts/Main.java
+++ b/microsoft/testsuites/docker/TestScripts/Main.java
@@ -1,0 +1,6 @@
+public class Main
+{
+    public static void main(String[] args) {
+        System.out.println("Hello world from java");
+    }
+}
--- a/microsoft/testsuites/docker/TestScripts/helloworld.py
+++ b/microsoft/testsuites/docker/TestScripts/helloworld.py
@@ -1,0 +1,1 @@
+print("Hello world from python")
--- a/microsoft/testsuites/docker/TestScripts/java.Dockerfile
+++ b/microsoft/testsuites/docker/TestScripts/java.Dockerfile
@@ -1,0 +1,9 @@
+FROM alpine
+COPY Main.java ./
+#install jdk
+RUN apk add openjdk8
+ENV JAVA_HOME /usr/lib/jvm/java-1.8-openjdk
+ENV PATH $PATH:$JAVA_HOME/bin
+#compile program
+RUN javac Main.java
+ENTRYPOINT java Main
--- a/microsoft/testsuites/docker/TestScripts/python.Dockerfile
+++ b/microsoft/testsuites/docker/TestScripts/python.Dockerfile
@@ -1,0 +1,5 @@
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install python3 -y
+COPY helloworld.py ./
+ENTRYPOINT ["python3"]
+CMD ["/helloworld.py"]


squirrelsc:
What's the reason to close old PR, and open a new one? It's better to use the same one to share comments and people's attention.

isaiahduncan:
> What's the reason to close old PR, and open a new one? It's better to use the same one to share comments and people's attention.
@squirrelsc I was getting weird behavior with github and the branch wouldn't push to the original. It would push to a new branch with the same name and I wasn't sure how to get the branch that I was pushing to update the original. Is it ok if we continue with this PR or should I try to figure out how to get my pushes to the old pr?

squirrelsc:
> > What's the reason to close old PR, and open a new one? It's better to use the same one to share comments and people's attention.
> > @squirrelsc I was getting weird behavior with github and the branch wouldn't push to the original. It would push to a new branch with the same name and I wasn't sure how to get the branch that I was pushing to update the original. Is it ok if we continue with this PR or should I try to figure out how to get my pushes to the old pr?

We can continue on this one. Did you try `git push -f`? What's error here?

isaiahduncan:
> > > What's the reason to close old PR, and open a new one? It's better to use the same one to share comments and people's attention.
> > > @squirrelsc I was getting weird behavior with github and the branch wouldn't push to the original. It would push to a new branch with the same name and I wasn't sure how to get the branch that I was pushing to update the original. Is it ok if we continue with this PR or should I try to figure out how to get my pushes to the old pr?
> 
> We can continue on this one. Did you try `git push -f`? What's error here?

I was doing git push -f. The problem was that it was saying that the branch had no upstream branch to push to. I then checked the git config and the upstream branch in fact was defined there. It continuously told me to use the command for adding upstream branch even though the branch was already defined in the config. And when I used the command to add a new upstream branch, it added a second entry in the config which was identical to the original. This is why I decided to delete the original and go with this

squirrelsc:
You can add `--set-upstream` parameter when push, check the document below.
https://linuxhint.com/set-upstream-branch-in-git/#:~:text=Set%20Upstream%20Branch%20in%20Git%201%20Pre-requisites%3A%202,%E2%80%93set-upstream%20option%20or%20-u%20option.%20More%20items...%20

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+    @property
+    def command(self) -> str:
+        return "docker"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        self._log.debug(
```
squirrelsc: This message is not necessary. The LISA will log which OS it is, and what's installing.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+            f"Installing Docker Engine on {self.node.os._information.vendor.lower()}"
+        )
+        package_list = ["docker-ce-cli", "containerd.io", "docker-ce"]
+
+        # comment: ouldnt find Mariner, almalinux, or rockylinux in operating_system.py
+        if isinstance(self.node.os, Debian) or isinstance(self.node.os, Ubuntu):
```
squirrelsc: Debian includes the Ubuntu, so it doesn't need to indicates both of them.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        self._log.debug(
+            f"Installing Docker Engine on {self.node.os._information.vendor.lower()}"
+        )
+        package_list = ["docker-ce-cli", "containerd.io", "docker-ce"]
+
+        # comment: ouldnt find Mariner, almalinux, or rockylinux in operating_system.py
```
squirrelsc: almaLinux and rockyLinux are included in RedHat. Feel free to add support for Mariner.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa import SkippedException
```
squirrelsc: This change can be in a separated commit.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _install(self) -> bool:
+        self._log.debug(
+            f"Installing Docker Engine on {self.node.os._information.vendor.lower()}"
+        )
+        package_list = ["docker-ce-cli", "containerd.io", "docker-ce"]
+
+        # comment: ouldnt find Mariner, almalinux, or rockylinux in operating_system.py
+        if isinstance(self.node.os, Debian) or isinstance(self.node.os, Ubuntu):
+            self.node.execute(cmd="apt-get update", sudo=True)
```
squirrelsc: This line will be called by LISA automatically. It doesn't need to be called again.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+            self._log.debug("Uninstall old versions of Docker")
```
squirrelsc: What's the reason we need to test the latest version of docker? How about test the current version on the distro. If there is no, we can install the distro version, instead of latest version. The latest version may not have good compatibility with older distros.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+                "Install packages apt-transport-https ca-certificates curl gnupg-agent"
+                " software-properties-common."
+            )
+            self.node.execute(cmd="apt-get update", sudo=True)
+            self.node.os.install_packages(
+                "apt-transport-https ca-certificates curl "
+                "gnupg-agent software-properties-common"
+            )
+
+            self._log.debug("Add Docker's official GPG key.")
```
squirrelsc: If we use the default version of docker, the logic will be much simpler. We need to confirm why it needs the latest version.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+                sudo=True
+            )
+
+            if self.node.os.information.release == "14.04":
+                package_list = ["docker-ce"]
+
+            self.node.execute(cmd="apt-get update", sudo=True)
+            self.node.os.install_packages(package_list, True)
+
+        elif isinstance(self.node.os, Redhat) or isinstance(self.node.os, CentOs):
```
squirrelsc: Redhat is enough.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+
+            self._log.debug("Install package yum-utils.")
+            self.node.execute("install_package \"yum-utils\"")
+
+            docker_url = "https://download.docker.com/linux/centos/docker-ce.repo"
+            self._log.debug(f"Add repo {docker_url}.")
+            self.node.execute(f"yum-config-manager --add-repo {docker_url}")
+
+            if self.node.os.information.version.major == 8:
```
squirrelsc: It can be used like `self.node.os.information.version >= "8.0.0"`. But check previous comments, it may not necessary to use latest version docker.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+                )
+
+        else:
+            raise SkippedException(f"{self.node.os.information.vendor} not supported")
+
+        self._start_docker_engine()
+        self._verify_docker_engine()
+        return True
+
+    def _start_docker_engine(self) -> None:
```
squirrelsc: Change it to a public method called `start`

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+        self._log.debug("Start docker engine")
+        result = self.node.execute("systemctl start docker", sudo=True)
+        if not result.exit_code == 0:
+            result = self.node.execute("service docker start", sudo=True)
+
+        if not result.exit_code == 0:
+            raise LisaException("Failed to start docker service")
+
+    def _verify_docker_engine(self) -> None:
```
squirrelsc: This method should be in test case.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+            5. Check results
+        """,
+        priority=2,
+    )
+    def docker_python_app(self, node: Node, log: Logger):
+        docker_tool = node.tools[Docker]
+
+        docker_image_name = "pythonappimage"
+        docker_container_name = "pythonapp"
+        docker_tool._remove_docker_image(docker_image_name)
```
squirrelsc: If a method is used outside of the class, don't add under score at beginning. The underscore means it's a private method.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+    public static void main(String[] args) {{
+        System.out.println("{java_string_identifier}");
+    }}
+}}
+        """.format(
+            java_string_identifier=java_string_identifier
+        )
+        self._generate_file(node, java_file_contents, Path(java_prog_src))
+
+        dockerfile_contents = """
```
squirrelsc: Put them in files within the test folder.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+    description="""
+    This test suite runs the docker test cases for java, python, dotnet, and wordpress.
+    """,
+)
+class docker(TestSuite):  # noqa
+    def _generate_file(self, node: Node, content: str, path: Path) -> None:
+        node.log.debug(f"Generating {path}.")
+
+        # need to check if file already exists
+        result = node.execute(f"[ -f {path} ]")
```
squirrelsc: The `node.os.shell` supports `copy`, `exists` and other operations, which can be reused.

squirrelsc: When copy them into node, put them into `node.working_path`

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+        """,
+        priority=0,
+    )
+    def docker_java_app(self, node: Node, log: Logger) -> None:
+        # comment: will change the way that docker is imported inot this file
+        docker_tool = node.tools[Docker]
+
+        docker_image_name = "javaappimage"
+        docker_container_name = "javaapp"
+        docker_tool._remove_docker_image(docker_image_name)
```
squirrelsc: It should install, test, and then remove. Don't leave the environment dirty. And you can use a random or unique name to avoid conflict with existing dockers.

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def _remove_docker_container(self, container_name: str) -> None:
+        self._log.debug(f"Removing Docker Container {container_name}")
+        self.node.execute(f"docker rm {container_name}", sudo=True)
+
+    def _remove_docker_image(self, image_name: str) -> None:
+        self._log.debug(f"Removing Docker Image {image_name}")
+        self.node.execute(f"docker rmi {image_name}", sudo=True)
+
+    def _build_docker_image(
```
squirrelsc: `build_image`

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+        result = self.node.execute("docker run hello-world", sudo=True)
+        if not result.exit_code == 0:
+            raise LisaException("Fail to run docker run hello-world")
+        self._log.debug(f"VerifyDockerEngine: hello-world output - {result.stdout}")
+
+    def _remove_docker_container(self, container_name: str) -> None:
+        self._log.debug(f"Removing Docker Container {container_name}")
+        self.node.execute(f"docker rm {container_name}", sudo=True)
+
+    def _remove_docker_image(self, image_name: str) -> None:
```
squirrelsc: `remove_image`

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+    def _verify_docker_engine(self) -> None:
+        self._log.debug(
+            f"VerifyDockerEngine on {self.node.os._information.vendor.lower()}"
+        )
+        result = self.node.execute("docker run hello-world", sudo=True)
+        if not result.exit_code == 0:
+            raise LisaException("Fail to run docker run hello-world")
+        self._log.debug(f"VerifyDockerEngine: hello-world output - {result.stdout}")
+
+    def _remove_docker_container(self, container_name: str) -> None:
```
squirrelsc: `remove_container`

lisa/tools/docker.py
```diff
@@ -0,0 +1,165 @@
...
+        self._log.debug("DOCKER BUILD OUTPUT:\n")
+        self._log.debug(self.node.execute(f"cat {DOCKER_BUILD_OUTPUT}").stdout)
+
+        if not result.exit_code == 0:
+            result = self.node.execute(f"cat {DOCKER_BUILD_OUTPUT}")
+            raise LisaException(
+                f"docker image build failed: {result.stdout}"
+            )
+
+    def _run_docker_container(
```
squirrelsc: `run_container`

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+        description="""
+            This test case creates and runs a java app using docker
+
+            Steps:
+            1. Install Docker
+            2. Create dockerfile and java program
+            3. Create docker image and run docker container
+            4. Check results
+        """,
+        priority=0,
```
squirrelsc: The priority of v2 test cases should add 1 in the v3. For example, if it's 0 in v2, it should be 1 in v3.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @TestCaseMetadata(
+        description="""
+            This test case creates and runs a java app using docker
+
+            Steps:
+            1. Install Docker
+            2. Create dockerfile and java program
+            3. Create docker image and run docker container
+            4. Check results
```
squirrelsc: Specify what's checked.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+        """.format(
+            java_prog_src=java_prog_src,
+            java_prog_app=java_prog_app,
+        )
+        self._generate_file(node, dockerfile_contents, Path(DOCKERFILE_NAME))
+
+        docker_tool._build_docker_image(docker_image_name, DOCKERFILE_NAME)
+        docker_tool._run_docker_container(docker_image_name, docker_container_name)
+
+        assert_that(node.execute(f"cat {DOCKER_RUN_OUTPUT}").stdout).is_equal_to(
```
squirrelsc: Add `describe_as` to output the error message, it includes what happens, what's the reason and impact.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+
+            Steps:
+            1. Install Docker
+            3. Create dockerfile and python program
+            4. Create docker image and run docker container
+            5. Check results
+        """,
+        priority=2,
+    )
+    def docker_python_app(self, node: Node, log: Logger):
```
squirrelsc: The test logic is similar with above one, share the logic in a common method.

lisa/util/constants.py
```diff
@@ -131,3 +131,7 @@
 DEVICE_TYPE_SRIOV = "SRIOV"
 DEVICE_TYPE_NVME = "NVME"
 DEVICE_TYPE_GPU = "GPU"
+
+DOCKERFILE_NAME = "Dockerfile"
```
squirrelsc: It's test case specified constants, don't add them in this file.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,154 @@
...
+
+from lisa import (
+    LisaException,
+    Logger,
+    Node,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+)
+from lisa.tools.docker import Docker
```
squirrelsc: add `Docker` in `lisa/tools/__init__.py` like other tools, and import it from there.

lisa/tools/docker.py
isaiahduncan: why put it in a seperate commit? The code in this file uses this so would I have to change that as well?

lisa/tools/docker.py
isaiahduncan: @squirrelsc Not sure what you want me to do here. This is how it was in the v2 code so that is why I wrote this like this.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: So I should first create the files on my machine and then copy them over to the vm using node.os.shell?

lisa/tools/docker.py
squirrelsc: I mean this file "docker.py" is to add a tool, it can be in a separated commit. It doesn't need to group with the test suite.

lisa/tools/docker.py
squirrelsc: The original code in v2, adds the docker repo to install docker. But the docker can be installed in the default repo of distros. The difference is the docker repo can provider higher version of docker. We don't need to follow v2 logic, we need to review and can revise it.

microsoft/testsuites/docker/dockersuite.py
squirrelsc: commit them to repo, and `copy` to remote.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: I had tried that before and I got a circular import error every time  I tried to run the tests. I think it stems from me importing the wget tool inside of the docker.py file. Do you have any suggestions on how to avoid this circular import error?

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: so basically in describe as, I could put something like "out put of docker run was not the expected output" ?

microsoft/testsuites/docker/dockersuite.py
squirrelsc: The `lsvmbus` and `kdump` imports the `wget` tool. Can you try to use their way? It may work.

microsoft/testsuites/docker/dockersuite.py
squirrelsc: It's helpful, and may like "The container doesn't output expected result. There may be errors when the container run."
If we can guess what kind of error here, we can provide more detail information. Explain it as deep as possible, and give suggestions how to fix it, if we can.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: should i do this in a different commit from the one in which I add the docker tool or should I do that and this both together?

squirrelsc: I missed this part. As a tool, it shouldn't raise a skipped exception. The SkippedException should be raised by test case or lisa itself. If it needs to skip a test case, detect the exception type, and skip it in test case.

lisa/tools/docker.py
```diff
@@ -0,0 +1,151 @@
...
+                f"{self.node.os._information.vendor.lower()}/gpg",
+                file_path="wget",
+                filename="gpg",
+            )
+            self.node.execute(cmd="apt-key add wget/gpg", shell=True, sudo=True)
+
+            self._log.debug("Set up the stable repository.")
+            release = self.node.execute(cmd="lsb_release -cs").stdout
+            self.node.execute(
+                "add-apt-repository -y \"deb https://download.docker.com/linux/"
```
isaiahduncan: @squirrelsc do you know what is wrong with this line? I have been getting the following message from flake8: `.\lisa\tools\docker.py:53:17: BLK100 Black would make changes.`

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: So I should have code at the beginning of the test case like this? : 
`if not ( isInstance(node.os,Debian) || isInstance(node.os, Redhat) ): 
Raise SkippedException`

lisa/tools/docker.py
squirrelsc: Please refer to below document to setup vscode. After that, every time the file is saved, this kind of error will be fixed automatically.
https://mslisa.readthedocs.io/en/main/write_test/dev_setup.html

microsoft/testsuites/docker/dockersuite.py
squirrelsc: It's ok to put the skip exception in the test case level. But, what's the reason to skip other distros in the docker test cases?

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,208 @@
...
+        major = node.os.information.version.major
+        if not (isinstance(node.os, Debian) or isinstance(node.os, Redhat)):
+            raise SkippedException(f"{self.node.os.information.vendor} not supported")
+        elif isinstance(node.os, Redhat) and not (major == 8 or major == 7):
+            raise SkippedException(
+                "Test not supported for RH/CentOS "
+                f"{node.os.information.release}"
+            )
+
+    def _run_and_verify(
```
squirrelsc: The private methods in class should be after public methods.

microsoft/testsuites/docker/dockersuite.py
squirrelsc: The verify/remove logic is also shared. Minimize the shared code. In the test cases, it should have only diff parts, like different image name, container name and so on.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: Its because there are some versions of linux that docker isn't supported on. for example, any redhat version who's major isn't 7 or 8

lisa/tools/docker.py
isaiahduncan: I don't think we are using the default version of docker. what I think you're referring to is my downloading the docker GPG key. If you look at lines 50-56, you will see that I'm downloading a stable version of docker that is relative to the respective distro being used. I think no change is necessary here

lisa/tools/docker.py
isaiahduncan: Done

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: When I don't put them into the directory that commands usually get executed in, I get problems with the dockerfile and the code src file. Is it ok for me to put the files into the directory that commands get executed in?

squirrelsc: I mean put those files in the `microsoft/testsuite/docker` folder. When it's executing, you can copy them to the remote node, and then build the docker image. Which one is "the directory that commands get executed in"?

microsoft/testsuites/docker/Dockerfiles/java.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM alpine
```
squirrelsc: Does the files need two folders, `CodeFiles` and `Dockerfiles`? To LISA, all of them are test data, or test scripts.

microsoft/testsuites/docker/Dockerfiles/java.Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+WORKDIR /usr/src/myapp
+COPY Main.java /usr/src/myapp
```
squirrelsc: It's better to put the same structure in LISA and remote. So put main.java and the java.dockerfile in the same folder.

microsoft/testsuites/docker/Dockerfiles/python.Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:18.04
+RUN apt-get update && apt-get install python3 -y
+COPY helloworld.py /
+ENTRYPOINT ["python3"]
+CMD ["/helloworld.py"]
```
squirrelsc: In usual, add an empty line at end of each file.

microsoft/testsuites/docker/Dockerfiles/java.Dockerfile
isaiahduncan: Not needed, I just did it for organization. I will change it

microsoft/testsuites/docker/Dockerfiles/python.Dockerfile
isaiahduncan: Done

microsoft/testsuites/docker/Dockerfiles/java.Dockerfile
isaiahduncan: Done

microsoft/testsuites/docker/Dockerfiles/java.Dockerfile
isaiahduncan: Done

microsoft/testsuites/docker/dockersuite.py

microsoft/testsuites/docker/dockersuite.py

lisa/tools/docker.py
squirrelsc: Please check if it needs to install latest version of Docker. We may not need this step in LISA validation. If removing those steps, the docker tool will be much simpler.

lisa/tools/docker.py
squirrelsc: In usual, the new version of a software doesn't support many os versions, even it's a stable version. The latest stable version mean the software is well tested, but it doesn't mean it's compatible with all distro versions. The software in the default repo of a distro is well tested with current distro version.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,166 @@
...
+        docker_tool.run_container(
+            docker_image_name, docker_container_name, docker_run_output
+        )
+
+        docker_tool.remove_image(docker_image_name)
+        docker_tool.remove_container(docker_container_name)
+
+        assert_that(node.execute(f"cat {docker_run_output}").stdout).is_equal_to(
+            string_identifier
+        ).described_as(
```
squirrelsc: Add `described_as` before the `is_equal_to`. In current order, the `described_as` won't output when `is_equal_to` failed.

squirrelsc: When you write test case, it's good to test each fail condition, make sure the output is expected.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,166 @@
...
+    )
+    def docker_java_app(self, node: Node) -> None:
+        self._skip_if_not_supported(node)
+        docker_tool = node.tools[Docker]
+        self._verify_docker_engine(node)
+
+        docker_image_name = "javaappimage"
+        docker_container_name = "javaapp"
+        docker_tool.remove_image(docker_image_name)
+        docker_tool.remove_container(docker_container_name)
```
squirrelsc: above lines also shareable.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,166 @@
...
+            docker_image_name,
+            docker_container_name,
+            python_string_identifier,
+        )
+
+    def _skip_if_not_supported(self, node: Node) -> None:
+        major = node.os.information.version.major
+        if not (isinstance(node.os, Debian) or isinstance(node.os, Redhat)):
+            raise SkippedException(f"{node.os.information.vendor} not supported")
+        elif isinstance(node.os, Redhat) and not (major == 8 or major == 7):
```
squirrelsc: check version by node.os.information.version.major >= "7.0.0". It's simpler, and compatible with future 9, 10.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,166 @@
...
+        result = node.execute("docker run hello-world", sudo=True)
+        if not result.exit_code == 0:
+            raise LisaException("Fail to run docker run hello-world")
+        node.log.debug(f"VerifyDockerEngine: hello-world output - {result.stdout}")
+
+    def _generate_file(self, node: Node, content: str, path: Path) -> None:
+        node.log.debug(f"Generating {path}.")
+
+        # need to check if file already exists
+        result = node.execute(f"[ -f {path} ]")
```
squirrelsc: did you try `node.shell.exists` and `node.shell.copy`?

lisa/tools/docker.py
isaiahduncan: I'll change to get the GPG key from specific distro version.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: i did node.shell.copy. forgot to get rid of this "_generate_file()" method. I will remove this method and make use of node.shell.exists

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,166 @@
...
+        docker_tool.remove_container(docker_container_name)
+
+        python_string_identifier = "Hello world from python"
+        python_prog_src = "helloworld.py"
+        dockerfile = "python.Dockerfile"
+        python_prog_path = (
+            Path(os.path.dirname(__file__)) / "TestScripts" / python_prog_src
+        )
+        dockerfile_path = Path(os.path.dirname(__file__)) / "TestScripts" / dockerfile
+        node.shell.copy(python_prog_path, Path(".") / python_prog_src)
```
squirrelsc: The destination path should be the remote working path. Get it from `node.working_path`. Because it's test case only files, create a folder for this test case, not put in the root working path. For example `node.working_path / "docker_test"`

microsoft/testsuites/docker/dockersuite.py

microsoft/testsuites/docker/dockersuite.py

microsoft/testsuites/docker/dockersuite.py

lisa/tools/docker.py
isaiahduncan: No longer necessary because of changes in docker.py

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: Changed order and checked output to make sure expected output was printed 👍

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def _skip_if_not_supported(self, node: Node) -> None:
+        major = node.os.information.version.major
+        if not (isinstance(node.os, Debian) or isinstance(node.os, Redhat)):
+            raise SkippedException(f"{node.os.information.vendor} not supported")
+        elif isinstance(node.os, Redhat) and not (major >= 7):
+            raise SkippedException(
+                f"Test not supported for RH/CentOS {node.os.information.release}"
+            )
+        elif isinstance(node.os, Redhat) and node.os.information.vendor == "AlmaLinux":
```
isaiahduncan: I had to add a skip case for AlmaLinux because docker wasn't properly installing on it. I'll have to look into different ways of getting docker ont AlmaLinux. I would like to add a task in ADO to make sure I get that figured out so that we can get in this PR if that's ok @squirrelsc

lisa/tools/docker.py
isaiahduncan: Changed to install docker.io with node.os.install_packages

microsoft/testsuites/docker/dockersuite.py
squirrelsc: You can make the test case fail, if AlmaLinux doesn't work. It doesn't need to skip it.

lisa/tools/docker.py
```diff
@@ -0,0 +1,94 @@
...
+        if not result.exit_code == 0:
+            raise LisaException("Failed to start docker service")
+
+    def remove_container(self, container_name: str) -> None:
+        self._log.debug(f"Removing Docker Container {container_name}")
+        self.node.execute(f"docker rm {container_name}", sudo=True)
+
+    def remove_image(self, image_name: str) -> None:
+        self._log.debug(f"Removing Docker Image {image_name}")
+        self.node.execute(f"docker rmi {image_name}", sudo=True)
```
squirrelsc: If it's possible to use `self.run` as possible. It have more underlying processing to tools. Like `self.run(f"rmi {image_name}", sudo=True)`. This command may need `force_run=True` to make sure it actually run everytime.

lisa/tools/docker.py
```diff
@@ -0,0 +1,94 @@
...
+    @property
+    def command(self) -> str:
+        return "docker"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def start(self) -> None:
+        self._log.debug("Start docker engine")
```
squirrelsc: There is a tool called `service` already, refer to examples in other test cases to use it.

lisa/tools/docker.py
```diff
@@ -0,0 +1,94 @@
...
+        docker_build_output = "docker_build.log"
+        result = self.node.execute(
+            f"docker build -t {docker_image_name} -f {dockerfile} "
+            f". 1> {docker_build_output} 2>&1",
+            shell=True,
+            sudo=True,
+            cwd=self.node.working_path,
+        )
+        self._log.debug("DOCKER BUILD OUTPUT:\n")
+        self._log.debug(self.node.execute(f"cat {docker_build_output}").stdout)
```
squirrelsc: 1. The value of f"cat {docker_build_output}" used twice. It's necessary to use a variable to save it.
2. It doesn't need to output to debug, the LISA framework already output.

lisa/tools/docker.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if not result.exit_code == 0:
```
squirrelsc: use `expected_exit_code` and `expected_exit_code_failure_message` to check the exit code, when calling `node.execute`.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,169 @@
...
+            3. Create dockerfile and python program
+            4. Create docker image and run docker container
+            5. Check results of docker run against python string identifier
+        """,
+        priority=3,
+    )
+    def docker_python_app(self, node: Node) -> None:
+        docker_image_name = "pythonappimage"
+        docker_container_name = "pythonapp"
+        docker_tool = self._verify_and_remove_containers(
```
squirrelsc, line 81: The logics can be merged into common method too. Try to reuse logic as much as possible.

lisa/tools/docker.py
```diff
@@ -0,0 +1,94 @@
...
+        self._log.debug(self.node.execute(f"cat {docker_run_output}").stdout)
+
+        if not result.exit_code == 0:
+            result = self.node.execute(f"cat {docker_run_output}")
+            raise LisaException(f"docker run failed: {result.stdout}.")
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Debian):
+            self._log.debug("Uninstall old versions of Docker")
+            self.node.execute(
```
squirrelsc: It doesn't need to remove before installed.

lisa/tools/docker.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+        self._log.debug("DOCKER RUN OUTPUT:\n")
+        self._log.debug(self.node.execute(f"cat {docker_run_output}").stdout)
+
+        if not result.exit_code == 0:
+            result = self.node.execute(f"cat {docker_run_output}")
+            raise LisaException(f"docker run failed: {result.stdout}.")
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Debian):
```
squirrelsc, line 64: It looks only covers Debian/Ubuntu. How about Redhat and other distros?

lisa/tools/docker.py
```diff
@@ -0,0 +1,94 @@
...
+            self._log.debug("Uninstall old versions of Docker")
+            self.node.execute(
+                cmd="apt-get remove -y docker docker-engine docker.io containerd runc",
+                shell=True,
+                sudo=True,
+            )
+
+            self.node.execute(cmd="apt-get update", sudo=True)
```
squirrelsc: This line will be called automatically.

lisa/tools/docker.py
squirrelsc: Check other similar places, there are some comparison is comparing like this way.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: done

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,169 @@
...
+        ).is_equal_to(string_identifier)
+
+    def _verify_docker_engine(self, node: Node) -> None:
+        node.log.debug(f"VerifyDockerEngine on {node.os.information.vendor.lower()}")
+        result = node.execute("docker run hello-world", sudo=True)
+        if not result.exit_code == 0:
+            raise LisaException("Fail to run docker run hello-world")
+        node.log.debug(f"VerifyDockerEngine: hello-world output - {result.stdout}")
+
+    def _generate_file(self, node: Node, content: str, path: Path) -> None:
```
squirrelsc: Should this file be removed?

microsoft/testsuites/docker/dockersuite.py
squirrelsc: Please do not resolve the conversation, because it's hard to me to track changes. If I resolve it, it means the issue is resolved finally. If you need to track them, please use reaction.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: Removed _generate_file function

lisa/tools/docker.py
isaiahduncan: Changed

lisa/tools/docker.py
isaiahduncan: It covers Debian, Redhat, and all of the subclasses

squirrelsc: How about other distros?

squirrelsc: If other distros are not supported at current stage, raise LisaException. It's a good practice to add `else` clause in most cases.

isaiahduncan: I should attempt to add support for the suse, nixOS, and otherLinux classes in operating_systems.py ?

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -0,0 +1,145 @@
...
+        self._verify_docker_engine(node)
+        docker_tool.remove_image(docker_image_name)
+        docker_tool.remove_container(docker_container_name)
+        return docker_tool
+
+    def _skip_if_not_supported(self, node: Node) -> None:
+        major = node.os.information.version.major
+        if not (isinstance(node.os, Debian) or isinstance(node.os, Redhat)):
+            raise SkippedException(f"{node.os.information.vendor} not supported")
+        elif isinstance(node.os, Redhat) and not (major >= 7):
```
squirrelsc: Don't check major version only, use it like `node.os.information.version < "7.0.0"`

lisa/tools/docker.py
```diff
@@ -0,0 +1,69 @@
...
+        service = self.node.tools[Service]
+        service.enable_service("docker")
+
+    def remove_container(self, container_name: str) -> None:
+        self._log.debug(f"Removing Docker Container {container_name}")
+        self.run(f"docker rm {container_name}", sudo=True)
+
+    def remove_image(self, image_name: str) -> None:
+        self._log.debug(f"Removing Docker Image {image_name}")
+        self.run(f"docker rmi {image_name}", sudo=True)
```
squirrelsc: 1. When using the `run`, it will join with the command. So it should be like below without beginning `docker`.
`self.run(f"rmi {image_name}", sudo=True)`
2. Add `force_run` to make the action executed every time.

lisa/tools/docker.py
squirrelsc: Add `force_run=True` to force execute the actions.

lisa/tools/docker.py
squirrelsc: you can implement them later, when we need to test that. But raise an exception with `else` clause.

lisa/tools/docker.py
isaiahduncan: Do you just want me to create the variables and use them or is there a specific feature within LISA that has these variables that you want me to use?

microsoft/testsuites/docker/dockersuite.py

lisa/tools/docker.py
isaiahduncan: done

lisa/tools/docker.py
squirrelsc: use existing arguments like below

`self.node.execute(f"cat {docker_build_output}", expected_exit_code=0, expected_exit_code_failure_message="give the failure reason...")`

lisa/tools/docker.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def remove_image(self, image_name: str) -> None:
+        self._log.debug(f"Removing Docker Image {image_name}")
+        self.run(f"rmi {image_name}", sudo=True, force_run=True)
+
+    def build_image(
+        self, docker_image_name: str = "testing", dockerfile: str = "Dockerfile"
+    ) -> None:
+        expected_exit_code = 0
+        result = self.node.execute(
```
squirrelsc: Please take care review comments for each similar places.

1. add `force_run=True` to action calls.
2. use `self.run` and remove "docker".

lisa/tools/docker.py
squirrelsc: Take care other similar places.

lisa/tools/docker.py
squirrelsc: take care other similar places.

lisa/tools/docker.py
isaiahduncan: Done

lisa/tools/docker.py
isaiahduncan: Done

lisa/tools/docker.py
isaiahduncan: Done

lisa/tools/docker.py
isaiahduncan: Done

lisa/tools/docker.py
isaiahduncan: Done

squirrelsc (dismissed):

lisa/tools/docker.py
```diff
@@ -0,0 +1,74 @@
...
+            force_run=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Docker run failed.",
+        )
+
+    def start(self) -> None:
+        self._log.debug("Start docker engine")
+        service = self.node.tools[Service]
+        service.enable_service("docker")
+        self.run(
```
squirrelsc: The `service` tool support to start service. Use the enveloped code.

isaiahduncan: what function in service should I use to start the service? Or should I just get rid of the self.run in this function?

squirrelsc: Use the `restart_service` or create a new one called `start_service`

isaiahduncan: Done

squirrelsc (approved):

PR #1562, opened 2021-10-06
Add ethtool testcases

sharsonia:


lisa/tools/ethtool.py
```diff
@@ -333,7 +333,32 @@ def _parse_rx_hash_level(self, interface: str, protocol: str, raw_str: str) -> N
             if self._tcp_udp_rx_hash_level_enable_pattern.search(raw_str)
             else False
         )
-        print(f"protocol {protocol}, protocol_hash_map {self.protocol_hash_map}")
+
+
+class DeviceStatistics:
+    _device_stats_pattern = re.compile(r"NIC statistics:\s+(?P<value>.*?)?$", re.DOTALL)
```
dsrivastavv: can you please add example cases covered by this regex in the comments?

lisa/tools/ethtool.py
```diff
@@ -676,6 +701,13 @@ def change_device_rx_hash_level(
 
         return self.get_device_rx_hash_level(interface, protocol, force=True)
 
+    def get_device_statistics(self, interface: str) -> DeviceStatistics:
+        result = self.run(f"-S {interface}", force_run=True)
```
dsrivastavv, line 854: Why is force run required?

lisa/tools/ethtool.py
```diff
@@ -676,6 +701,13 @@ def change_device_rx_hash_level(
 
         return self.get_device_rx_hash_level(interface, protocol, force=True)
 
+    def get_device_statistics(self, interface: str) -> DeviceStatistics:
```
dsrivastavv: [nit] Can you please add sample output of this command to comments?

microsoft/testsuites/network/networksettings.py
```diff
@@ -384,3 +384,23 @@ def validate_device_rx_hash_level_change(self, node: Node, log: Logger) -> None:
...
+            1. Get all the device's statistics.
+            2. Validate device statistics lists per cpu statistics as well.
+        """,
+        priority=2,
+    )
+    def validate_device_statistics(self, node: Node, log: Logger) -> None:
+        ethtool = node.tools[Ethtool]
+        devices_statistics = ethtool.get_all_device_statistics()
+
+        for stats in devices_statistics:
```
dsrivastavv: The current logic verifies the CPU stats are present for atleast one CPU but not for all CPU's. Do we need to cover the second case as well?

lisa/tools/ethtool.py
sharsonia: Because we would want to get latest stats every time, not the cached value

microsoft/testsuites/network/networksettings.py
sharsonia: At least in the LISAv2 counter-part, it doesn't verify we have stats for all CPUs or not. it just want to validates we get stats from the ethtool command or not and whether we get the per cpu stats as well.
If we want to verify for each cpu then we would first need to fetch number of CPUs info and then validate. @LiliDeng - Do you think we should add that or it will be an overkill for this test?

microsoft/testsuites/network/networksettings.py
```diff
@@ -384,3 +384,23 @@ def validate_device_rx_hash_level_change(self, node: Node, log: Logger) -> None:
...
+            assert_that(
+                stats.device_stats, f"Statistics for {stats.interface} are not empty"
```
squirrelsc: 1. The message shows when the assert is failed. In this code, it means the stats is empty. So the message should be like "are empty"
2. Explain what it means, and potential problems, if it's empty.

lisa/tools/ethtool.py
sharsonia: Added in DeviceStatistics class

microsoft/testsuites/network/networksettings.py
squirrelsc: Explain what it means, and potential problems, if it's empty.

microsoft/testsuites/network/networksettings.py
```diff
@@ -384,3 +384,23 @@ def validate_device_rx_hash_level_change(self, node: Node, log: Logger) -> None:
...
+        """,
+        priority=2,
+    )
+    def validate_device_statistics(self, node: Node, log: Logger) -> None:
+        ethtool = node.tools[Ethtool]
+        devices_statistics = ethtool.get_all_device_statistics()
+
+        for stats in devices_statistics:
+            assert_that(
+                stats.device_stats, f"Statistics for {stats.interface} are empty"
```
squirrelsc: Explain what it means, and potential problems, if it's empty.

squirrelsc (dismissed):

sharsonia:
For validating device statistics, run some traffic using iperf (with multiple TCP connections), and verify that traffic is evenly spread out on different VMBUS channels (after iperf -P64, test case should assert if one queue receives/transmits 30% more packets than other queues)

microsoft/testsuites/network/networksettings.py
squirrelsc: Provide more information on this error. It's hard to troubleshooting with this error message.

microsoft/testsuites/network/networksettings.py
```diff
@@ -31,6 +34,15 @@
     """,
 )
 class NetworkSettings(TestSuite):
+
+    # regex for filtering per vmbus channel stats from the full device statistics.
+    _tx_queue_stats_regex = re.compile(
```
squirrelsc: add examples above.

microsoft/testsuites/network/networksettings.py
```diff
@@ -494,6 +506,86 @@ def validate_device_msg_level_change(self, node: Node, log: Logger) -> None:
...
+        # preparation work before launch iperf3
+        stop_firewall(environment)
+
+        # run iperf3 on server side and client side
+        # iperfResults.log stored client side log
+        source_iperf3 = server_node.tools[Iperf3]
+        dest_iperf3 = client_node.tools[Iperf3]
+        source_iperf3.run_as_server()
+        dest_iperf3.run_as_client_async(
+            server_node.internal_address, client_iperf3_log, 1800, parallel_number=64
```
squirrelsc: 1. Will it run 1800 seconds? it's too long to a functional test case. Try default value 10 seconds, it should be enough.
2. To get it more stable, do not use async here. Use synchronous method.

squirrelsc (approved):

PR #1563, opened 2021-10-06
Migrated Test: verify_initrd_modules

kamalca:
This test case covers the functionality tested by LISAv2 test INITRD-MODULES-CHECK.

This test case ensures all of the necessary modules are present in initrd.

### DIFF
--- a/lisa/tools/lsinitrd.py
+++ b/lisa/tools/lsinitrd.py
@@ -1,0 +1,66 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from typing import cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.util import LisaException, find_patterns_in_lines, get_matched_str
+
+
+class Lsinitrd(Tool):
+    _non_comment_pattern = re.compile(r"^\s*[^#\s].*$", re.MULTILINE)
+    _modules_dep_file_pattern = re.compile(
+        r".*\s(?P<PATH>[^\s]*\/modules\.dep)\s*$", re.MULTILINE
+    )
+
+    @property
+    def command(self) -> str:
+        return "lsinitrd"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("dracut-core")
+        return self._check_exists()
+
+    def has_module(self, module_file_name: str, initrd_file_path: str = "") -> bool:
+        """
+        1) Finds path of modules.dep in initrd
+        2) Searches modules.dep for the module file name
+        """
+        result = self.run(
+            initrd_file_path,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"`lsinitrd {initrd_file_path}` failed.",
+        )
+
+        modules_dep_path = get_matched_str(
+            result.stdout, self._modules_dep_file_pattern
+        )
+
+        if not modules_dep_path:
+            raise LisaException(
+                f"[lsinitrd] modules.dep could not be found. "
+                f"Make sure the file is present in initrd image {initrd_file_path}."
+            )
+
+        result = self.run(
+            f"-f {modules_dep_path} {initrd_file_path}",
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"`lsinitrd -f /{modules_dep_path} "
+            f"{initrd_file_path}` failed.",
+        )
+        modules_required = find_patterns_in_lines(
+            result.stdout, [self._non_comment_pattern]
+        )
+
+        return any(
+            module_file_name in requirement for requirement in modules_required[0]
+        )
--- a/microsoft/testsuites/core/hv_module.py
+++ b/microsoft/testsuites/core/hv_module.py
@@ -1,13 +1,22 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from typing import List
+
 from assertpy import assert_that
 from semver import VersionInfo
 
-from lisa import Logger, RemoteNode, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa import (
+    Logger,
+    RemoteNode,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    simple_requirement,
+)
 from lisa.operating_system import Redhat
 from lisa.sut_orchestrator.azure.tools import LisDriver
-from lisa.tools import Lsmod, Modinfo, Modprobe, Uname
+from lisa.tools import Find, Lsinitrd, Lsmod, Modinfo, Modprobe, Uname
 from lisa.util import SkippedException
 
 
@@ -22,6 +31,7 @@ from lisa.util import SkippedException
     It is responsible for ensuring the Hyper V drivers are all present,
     are included in initrd, and are all the same version.
     """,
+    requirement=simple_requirement(supported_platform_type=["azure", "ready"]),
 )
 class HvModule(TestSuite):
     @TestCaseMetadata(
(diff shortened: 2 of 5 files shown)


lisa/tools/lsinitrd.py
```diff
@@ -0,0 +1,68 @@
...
+        2) Searches modules.dep for the module file name
+        """
+        result = self.run(initrd_file_path, sudo=True)
+        if result.exit_code != 0:
+            LisaException(
+                f"`lsinitrd {initrd_file_path}` "
+                f"failed with exit code {result.exit_code}."
+            )
+
+        modules_dep_path = find_patterns_in_lines(
```
squirrelsc: For one pattern, use `get_matched_str`

lisa/tools/lsinitrd.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def module_is_present(
+        self, module_file_name: str, initrd_file_path: str = ""
+    ) -> bool:
+        """
+        1) Finds path of modules.dep in initrd
+        2) Searches modules.dep for the module file name
+        """
+        result = self.run(initrd_file_path, sudo=True)
+        if result.exit_code != 0:
```
squirrelsc: If it just check the exit_code is 0 or not, set `expected_exit_code` and `expected_exit_code_failure_message` of `run`.

lisa/tools/lsinitrd.py
```diff
@@ -0,0 +1,68 @@
...
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("dracut-core")
+        return self._check_exists()
+
+    def module_is_present(
```
squirrelsc: how about `has_module`?

microsoft/testsuites/core/hv_module.py
```diff
@@ -102,6 +102,68 @@ def verify_hyperv_modules(
...
+            k: v
+            for (k, v) in all_necessary_hv_modules_file_names.items()
+            if k not in skip_modules
+        }
+
+        # 2) Find the path of initrd
+        uname = node.tools[Uname]
+        kernel_version = uname.get_linux_information().kernel_version_raw
+        find = node.tools[Find]
+        initrd_possible_file_names = [
```
squirrelsc, line 105: How about use check like `isinstance(node.os, Redhat)` to differentiate them? It doesn't need to check them one by one. And it's easy to understand which file is looking for, and doesn't exist.

microsoft/testsuites/core/hv_module.py
```diff
@@ -102,6 +102,68 @@ def verify_hyperv_modules(
             "Not all Hyper V drivers are present."
         ).is_length(0)
 
+    @TestCaseMetadata(
+        description="""
+        This test case will ensure all necessary hv_modules are present in
```
squirrelsc, line 73: It looks a Microsoft Hyper-V only test case, add requirement to limited it runs in Azure and Ready environment only.

microsoft/testsuites/core/hv_module.py
```diff
@@ -102,6 +102,68 @@ def verify_hyperv_modules(
...
+        missing_modules = []
+        for module in hv_modules_file_names:
+            if not lsinitrd.module_is_present(
+                module_file_name=hv_modules_file_names[module],
+                initrd_file_path=initrd_file_path,
+            ):
+                missing_modules.append(module)
+
+        assert_that(missing_modules).described_as(
+            "Some modules are missing from initrd."
```
squirrelsc: It's a key assertion message, and should be more accurate, like "some hyper-v required modules are missed from initrd."

microsoft/testsuites/core/hv_module.py
kamalca: What does the code look like to check the environment?
Isn't this true of the whole test suit? Is there a way to make this check apply to the whole test suit?

squirrelsc: It can be added on test suite or test case level. Refer to provisioning test case, but it has no platform requirement.

```python
@TestCaseMetadata(
    ...
    requirement=simple_requirement(supported_platform_type=[Azure, Ready])
    ...
```

kamalca: Are `Azure` and `Ready` supposed to be strings or are they types? If they are types, where are they imported from. I wasn't able to find any other examples of `supported_platform_type=` but it looks like it is meant to take List[str].

Are the types meant to be `AzurePlatform` and `ReadyPlatform`? (from `lisa.sut_orchestrator.azure.platform_` and `lisa.sut_orchestrator.ready` respectively)

squirrelsc: Sorry for unclear. It's strings, and it's better to reference exiting constants. Please move the platform type strings for according files to `lisa/sut_orchestrator/__init__.py`, so they can be referenced easier.

kamalca: `requirement=simple_requirement(supported_platform_type=[Azure, Ready])`
This seems to have an AND type requirement. I was getting an error that ["azure"] does not meet requirement ["azure", "ready"]

squirrelsc: Can you share me the call trace? I can see where is the error from.

squirrelsc: Can you try to fix it in lisa_runner.py? It looks I set the order of req and cap incorrectly.
This line 
`check_result = test_req.platform_type.check(platform_type_set)`
Should be 
`check_result = platform_type_set.check(test_req.platform_type)`

lisa/tools/lsinitrd.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+
+        result = self.run(
+            f"-f {modules_dep_path} {initrd_file_path}",
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"`lsinitrd -f /{modules_dep_path} "
+            f"{initrd_file_path}` failed.",
+        )
+        modules_required = find_patterns_in_lines(
```
squirrelsc, line 60: This can be `get_matched_str` too. When resolving review comments, try to check other similar places.

kamalca: Can you clarify why this one should be `get_matched_str`. I am using this to filter out the lines that are comments.

squirrelsc: `find_patterns_in_lines` is used to match multiple patterns with the content, the `get_matched_str` is used to match one pattern with the content. So, the output of `find_patterns_in_lines` is a 2d list, and the `get_matched_str` output 1d.

kamalca: `get_matched_str` only returns a single string, not a list of strings

microsoft/testsuites/core/hv_module.py
kamalca: Changing this line fixed the issue.

lisa/tools/lsinitrd.py
squirrelsc: Correct, you are right. In this case, the `find_patterns_in_lines` is fit.

squirrelsc:
Is it draft PR?

microsoft/testsuites/core/hv_module.py
kamalca: Should I fix this in this PR or should I open new PR for this change?

squirrelsc: you can fix it in this PR within a separated commit.

squirrelsc (dismissed):

microsoft/testsuites/core/hv_module.py
kamalca: My worry is that this solution might be less robust. Can we ensure that every version (past, present, and future) of a distro will use the same naming scheme? Can we be sure that we are accurately covering every type of distro? Do you have a resource you can point me to about different naming schemes on different distros? I want to make sure that the tests are robust, especially if I am going to vary from how the test is done in LISAv2.

kamalca:
Sorry for the messy commit history. I did a merge commit to resolve a merge conflict then realized that I should be using typing.List instead of list for better Python3.8 support. (I'm using Python3.9 which deprecates typings by making the built in types generic)
Then replaying a merge commit is difficult (not possible as far as I know) so I'm not sure the best way to clean up the commit history. Let me know the best way to clean up the commit history. Was it a mistake to resolve the conflict with a merge commit?

microsoft/testsuites/core/hv_module.py
squirrelsc: In general, we need the test case to be more accurate than robust. It doesn't mean the robust is not important. For example, if the needed information is added into a wrong file. The file isn't effective real, but it may meet test case requirement.
For this case, it's not a big deal, because it's low chance to happen. But it's a good practice in test case to improve accuracy.
If you want to improve this, you can test it with different marketplace images on endorsed distros.

squirrelsc:
> Sorry for the messy commit history. I did a merge commit to resolve a merge conflict then realized that I should be using typing.List instead of list for better Python3.8 support. (I'm using Python3.9 which deprecates typings by making the built in types generic) Then replaying a merge commit is difficult (not possible as far as I know) so I'm not sure the best way to clean up the commit history. Let me know the best way to clean up the commit history. Was it a mistake to resolve the conflict with a merge commit?

When a change get into the main branch, it's better to keep as is. The force push on public branches is not recommended, because it needs everyone with this branch to rebase code. It's ok.
For private branch, you can use `git rebase -i` to operate changes, and use `git push -f` to force push your upstream branch.

kamalca:
> > Sorry for the messy commit history. I did a merge commit to resolve a merge conflict then realized that I should be using typing.List instead of list for better Python3.8 support. (I'm using Python3.9 which deprecates typings by making the built in types generic) Then replaying a merge commit is difficult (not possible as far as I know) so I'm not sure the best way to clean up the commit history. Let me know the best way to clean up the commit history. Was it a mistake to resolve the conflict with a merge commit?
> 
> When a change get into the main branch, it's better to keep as is. The force push on public branches is not recommended, because it needs everyone with this branch to rebase code. It's ok. For private branch, you can use `git rebase -i` to operate changes, and use `git push -f` to force push your upstream branch.

For the future, what is the best way to resolve a merge conflict?

squirrelsc (approved):

PR #1570, opened 2021-10-12
dpdk_ring_ping test

mcgov:
Adds a test for dpdk memory rings to check their latency is within an acceptable perf bound.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -1,10 +1,11 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import re
 from collections import deque
 from typing import Dict, List, Tuple
 
-from assertpy import assert_that
+from assertpy import assert_that, fail
 
 from lisa import (
     Environment,
@@ -19,12 +20,13 @@ from lisa import (
 from lisa.features import NetworkInterface, Sriov
 from lisa.nic import Nics
 from lisa.testsuite import simple_requirement
-from lisa.tools import Echo, Lspci, Mount
+from lisa.tools import Echo, Git, Lspci, Make, Mount
 from lisa.util import constants
 from lisa.util.parallel import Task, TaskManager
 from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
 
 VDEV_TYPE = "net_vdev_netvsc"
+MAX_RING_PING_LIMIT_NS = 200000
 
 
 @TestSuiteMetadata(
@@ -35,6 +37,12 @@ VDEV_TYPE = "net_vdev_netvsc"
     """,
 )
 class Dpdk(TestSuite):
+
+    # regex for parsing ring ping output for the final line,
+    # grabbing the max latency of 99.999% of data in nanoseconds.
+    # ex: percentile 99.999 = 12302
+    _ring_ping_percentile_regex = re.compile(r"percentile 99.999 = ([0-9]+)")
+
     @TestCaseMetadata(
         description="""
             This test case checks DPDK can be built and installed correctly.
@@ -66,6 +74,71 @@ class Dpdk(TestSuite):
         assert_that(tx_pps).described_as(
             f"TX-PPS ({tx_pps}) should have been greater than 2^20 (~1m) PPS."
         ).is_greater_than(2 ** 20)
+
+    @TestCaseMetadata(
+        description="""
+            This test runs the dpdk ring ping utility from:
+            https://github.com/shemminger/dpdk-ring-ping
+            to measure the maximum latency for 99.999 percent of packets during
+            the test run. The maximum should be under 200000 nanoseconds
+            (.2 milliseconds).
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_nic_count=2,
+            network_interface=Sriov(),
+        ),
+    )
+    def verify_dpdk_ring_ping(self, node: Node, log: Logger) -> None:
+        # setup and unwrap the resources for this test
+        test_kit = initialize_node_resources(node, log)
+        testpmd = test_kit.testpmd
+
+        # grab a nic and run testpmd
+        git = node.tools[Git]
+        make = node.tools[Make]
+        echo = node.tools[Echo]
+        rping_build_env_vars = [
+            "export RTE_TARGET=build",
+            f"export RTE_SDK={testpmd.dpdk_cwd.as_posix()}",
+        ]
+        echo.write_to_file(",".join(rping_build_env_vars), "~/.bashrc", append=True)
+        git_path = git.clone(
+            "https://github.com/shemminger/dpdk-ring-ping.git", cwd=node.working_path
+        )
+        make.run(
+            shell=True,
+            cwd=git_path,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="make could not build rping project.",
+        ).assert_exit_code()
+        # run ringping for 30 seconds
+        runcmd = "./build/rping -c 0xC0 -n 2 --no-pci --no-huge -- -d 5 -t 10"
+        result = node.execute(
+            runcmd,
+            shell=True,
+            cwd=git_path,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="rping program failed to run correctly.",
+        )
+        result.assert_exit_code()
+        # get the max latency for 99.999 percent of enqueued 'packets'.
+        result_regex = self._ring_ping_percentile_regex.search(result.stdout)
+        if result_regex and len(result_regex.groups()) == 1:
+            max_ping_measured = int(result_regex.group(1))
+            assert_that(max_ping_measured).described_as(
+                (
+                    f"RingPing measured {max_ping_measured} as maximum ping latency,"
+                    f" maximum should be less than {MAX_RING_PING_LIMIT_NS}"
+                )
+            ).is_less_than(MAX_RING_PING_LIMIT_NS)
+        else:
+            fail(
+                (
+                    "Could not get latency data from rping result. "
+                    f"Search was for 'percentile 99.999 = ([0-9]+)'\n{result.stdout}\n"
+                )
+            )
 
     @TestCaseMetadata(
         description="""
@@ -124,7 +197,7 @@ class Dpdk(TestSuite):
         log.info(f"receiver rx-pps: {rcv_rx_pps}")
         log.info(f"sender tx-pps: {snd_tx_pps}")
 
-        # differences in NIC model throughput can lead to different snd/rcv counts
+        # differences in NIC type throughput can lead to different snd/rcv counts
         assert_that(rcv_rx_pps).described_as(
             "Throughput for RECEIVE was below the correct order-of-magnitude"
         ).is_greater_than(2 ** 20)
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -11,7 +11,7 @@ from assertpy import assert_that
 from lisa.executable import Tool
 from lisa.nic import NicInfo
 from lisa.operating_system import CentOs, Redhat, Ubuntu
-from lisa.tools import Git, Lspci, Wget
+from lisa.tools import Echo, Git, Lspci, Wget
 from lisa.util import SkippedException
 
 
@@ -107,7 +107,8 @@ class DpdkTestpmd(Tool):
         if self._install_dependencies():
             node = self.node
             git_tool = node.tools[Git]
-            git_tool.clone(self._dpdk_github, cwd=node.working_path)
+            echo_tool = node.tools[Echo]
+            self.dpdk_cwd = git_tool.clone(self._dpdk_github, cwd=node.working_path)
             dpdk_path = self.node.working_path.joinpath("dpdk")
             self.__execute_assert_zero("meson build", dpdk_path)
             dpdk_build_path = dpdk_path.joinpath("build")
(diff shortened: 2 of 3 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+    def verify_dpdk_ring_ping(self, node: Node, log: Logger) -> None:
+        # setup and unwrap the resources for this test
+        test_kit = initialize_node_resources(node, log)
+        _, testpmd = test_kit.node_nic_info, test_kit.testpmd
+
+        # grab a nic and run testpmd
+        git = node.tools[Git]
+        make = node.tools[Make]
+        echo = node.tools[Echo]
+        echo.write_to_file(
```
squirrelsc: 1. Is it useful without sign out/sign in?
2. The two lines can be done in same command.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        priority=2,
+        requirement=simple_requirement(
+            min_nic_count=2,
+            network_interface=Sriov(),
+        ),
+    )
+    def verify_dpdk_ring_ping(self, node: Node, log: Logger) -> None:
+        # setup and unwrap the resources for this test
+        test_kit = initialize_node_resources(node, log)
+        _, testpmd = test_kit.node_nic_info, test_kit.testpmd
```
squirrelsc: It looks redundant. How about access `test_kit.testpmd` in following code directly?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        echo.write_to_file(
+            "export RTE_TARGET=build",
+            "~/.bashrc",
+            append=True,
+        )
+        git_path = git.clone(
+            "https://github.com/shemminger/dpdk-ring-ping.git", cwd=node.working_path
+        )
+        result = make.run(shell=True, cwd=git_path)
+        log.info(f"{result.stdout}")
```
squirrelsc: The result will be in debug level by default. What's the reason to print it in info level again?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        result.assert_exit_code()
```
squirrelsc: Try to use `expected_exit_code` and `expected_exit_code_failure_message` of `run` method.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        # run ringping for 30 seconds
+        runcmd = "./build/rping -c 0xC0 -n 2 --no-pci --no-huge -- -t 20"
+        result = node.execute(runcmd, shell=True, cwd=git_path)
+        result.assert_exit_code()
```
squirrelsc, line 124: similar as above, the `execute` support `expected_exit_code` and `expected_exit_code_failure_message` also.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        # get last line of the ping log, contains the max latency value.
+        ping_result = result.stdout.split("\n")[-1]
```
squirrelsc: `result.stdout.splitlines` is better, and use regex as possible.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        assert_that(ping_result_split).described_as(
+            (
+                "Split ring ping result found unexpected "
+                "number of key/value pairs. Code splits on '='\n"
+                f"last line of the rping cmd was:\n {ping_result}"
+            )
+        ).is_length(2)
+        assert_that(ping_result_split[0]).described_as(
+            f"'99.999' percentile line doesn't validate. Line was:\n{ping_result}"
+        ).is_equal_to("percentile 99.999")
```
squirrelsc: Is it reliable enough?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +68,71 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        ping_result_split = [x.strip() for x in ping_result.split("=")]
+        assert_that(ping_result_split).described_as(
+            (
+                "Split ring ping result found unexpected "
+                "number of key/value pairs. Code splits on '='\n"
+                f"last line of the rping cmd was:\n {ping_result}"
+            )
+        ).is_length(2)
+        assert_that(ping_result_split[0]).described_as(
+            f"'99.999' percentile line doesn't validate. Line was:\n{ping_result}"
```
squirrelsc: The value should be printed, if the assertion is failed. Double check it, and avoid duplicate information.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -60,7 +60,7 @@ def command(self) -> str:
         "wget",
         "dos2unix",
         "psmisc",
-        "kernel-devel",
+        "kernel-devel-`uname -r`",
```
squirrelsc: Are they fixing existing test cases? Make each fix in a separated commit, and add information in commit description.

microsoft/testsuites/dpdk/dpdksuite.py
LiliDeng: @mcgov have you checked this https://github.com/shemminger/dpdk-ring-ping work well?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: This is a way to persist environment variables since we always open a new ssh session for run()
I'll compact them

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: @squirrelsc I prefer short variable names when possible. It doesn't affect perf and makes it more readable so I'm going to keep it as is.
@LiliDeng Yes, this case is an exact port from lisav2 to lisav3. Same parameters, same program. We recently fixed a bug in ring ping so that it works with new dpdk versions as well.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I'll remove it.

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: It's ok to have short names, but it doesn't need to be paired variables. below is enough. If both variable are necessary, two lines are clear.
`testpmd = test_kit.testpmd`

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Can we shorten 'expected_exit_code_failure_message' somehow? That variable name takes up almost the whole line

squirrelsc: It can be shorter, if the variable name can explain itself. The meaning is higher priority than length. It's similar like Linux upstream repos.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: will rewrite

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: 🤦of course! will fix

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: This can be a future commit, it works for now and I like the redesign of this assert system 🙂

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: will move to another commit

mcgov:
@squirrelsc @LiliDeng  ready for a second round! Thank you for comments.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -90,7 +159,7 @@ def verify_dpdk_send_receive(self, environment: Environment, log: Logger) -> Non
                 ]
             else:
                 raise SkippedException()
-        log.debug((f"\nsender:{external_ips[0]}\nreceiver:{external_ips[1]}\n"))
+        log.info((f"\nsender:{external_ips[0]}\nreceiver:{external_ips[1]}\n"))
```
squirrelsc: BTW, it can use internal IPs, it may be faster.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -67,6 +69,73 @@ def verify_dpdk_build(self, node: Node, log: Logger) -> None:
...
+        runcmd = "./build/rping -c 0xC0 -n 2 --no-pci --no-huge -- -t 20"
+        result = node.execute(
+            runcmd,
+            shell=True,
+            cwd=git_path,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="rping program failed to run correctly.",
+        )
+        # get the max latency for 99.999 percent of enqueued 'packets'.
+        result_regex = re.search(r"percentile 99.999 = ([0-9]+)", result.stdout)
```
squirrelsc: move the static pattern to class level, so it save cost on compiling it every time. Search `re.compile` to see how it's used in LISA code base.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -107,14 +107,28 @@ def _install(self) -> bool:
...
+            echo_tool = node.tools[Echo]
+            self.dpdk_cwd = git_tool.clone(self._dpdk_github, cwd=node.working_path)
             dpdk_path = self.node.working_path.joinpath("dpdk")
             self.__execute_assert_zero("meson build", dpdk_path)
             dpdk_build_path = dpdk_path.joinpath("build")
             self.__execute_assert_zero("which ninja", dpdk_build_path)
             self.__execute_assert_zero("ninja", dpdk_build_path, timeout=1200)
             self.__execute_assert_zero("ninja install", dpdk_build_path)
             self.__execute_assert_zero("ldconfig", dpdk_build_path)
+            echo_tool.write_to_file(
```
squirrelsc: merge the two lines into one command.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -107,14 +107,28 @@ def _install(self) -> bool:
...
+                "export PKG_CONFIG_PATH=${PKG_CONFIG_PATH}:/usr/local/lib64/pkgconfig/",
+                "~/.bashrc",
+                append=True,
+            )
+            echo_tool.write_to_file(
+                "export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/usr/local/lib64/",
+                "~/.bashrc",
+                append=True,
+            )
+            echo_tool.write_to_file(
```
squirrelsc: Don't assume the user name is lisauser. If it's necessary, write to root one by `sudo=True` like above.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I can see doing this in other situations but this test and regex only runs once. Doesn't it make sense to keep it inline here?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I wanted to add this to make investigating issues easier, it's nice to have the external ip logged right next to the data you're looking at. Internal IP is used elsewhere in code so it's already in the logs

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Woah! good catch.

mcgov: Going to remove. I want to get rid of the issues where sudo commands don't have the same path as the user commands, but this is probably not the right way to do it.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I attempted this, there's an issue where echo doesn't really process escape characters correctly so it gets messy. Rather than try and manage that in this commit I'll change these to iterate over the lines for brevity. We can refactor it to get a cleaner solution later.

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: Any comment?

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: 1. It's a good practice to pre-compile static patterns, and others may copy the logic. 
2. Add examples, That's important to improve patterns later.

microsoft/testsuites/dpdk/dpdktestpmd.py
squirrelsc: They can be split by `\\n`, `;`, or `&`.

squirrelsc (approved):

PR #1571, opened 2021-10-12
testcase to validate msg level set/unset

sharsonia:


microsoft/testsuites/network/networksettings.py
```diff
@@ -384,3 +386,85 @@ def validate_device_rx_hash_level_change(self, node: Node, log: Logger) -> None:
...
+                interface, original_msg_level_number
+            )
+            assert_that(
+                reverted_settings.msg_level_number,
+                f"Setting msg flags by number - {original_msg_level_number} didn't"
+                f" succeed. Current msg level value is {new_settings.msg_level_number}",
+            ).contains(original_msg_level_number)
+
+    def _verify_msg_level_change_kernel_support(self, node: Node) -> None:
+        pass
```
squirrelsc: What's the method is used for?

lisa/tools/modinfo.py
```diff
@@ -54,3 +55,19 @@ def get_version(
...
+        no_info_log: bool = True,
+        no_error_log: bool = True,
+    ) -> str:
+        output = self.get_info(
+            mod_name=mod_name,
+            force_run=force_run,
+            no_info_log=no_info_log,
+            no_error_log=no_error_log,
+        )
+        found_filename = find_patterns_in_lines(output, [self.__filename_pattern])
```
squirrelsc: use `get_matched_str`, it returns one matched result, or None.

lisa/tools/nm.py
```diff
@@ -0,0 +1,30 @@
...
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("binutils")
+        return self._check_exists()
+
+    def get_symbol_table(self, file: str) -> str:
+        result = self.run(f"{file}", shell=True, force_run=True, sudo=True)
```
squirrelsc: 1. `f"{file}"` -> `file`
2. If the result is not changed, avoid to use force_run to save time.

lisa/tools/nm.py
```diff
@@ -0,0 +1,30 @@
...
+        if result.exit_code != 0:
+            raise LisaException(f"Failed to extract file, {result.stderr}")
```
squirrelsc: use `result.stdout`. I found the remote node cannot output stderr actually. All output are in stdout.

lisa/tools/modinfo.py
```diff
@@ -10,6 +10,7 @@
 
 class Modinfo(Tool):
     __version_pattern = re.compile(r"^version:[ \t]*([^ \n]*)")
+    __filename_pattern = re.compile(r"filename:[ \t]*([^ \n]*)")
```
squirrelsc, line 18: Add examples above.

microsoft/testsuites/network/networksettings.py
```diff
@@ -384,3 +387,118 @@ def validate_device_rx_hash_level_change(self, node: Node, log: Logger) -> None:
...
+                netvsc_module = node.execute("ls hv_netvsc*").stdout
+
+            nm = node.tools[Nm]
+            raw_symbols = nm.get_symbol_table(netvsc_module)
+            msg_level_symbols = re.findall(r".*msglevel.*", raw_symbols)
+        else:
+            # if the module is builtin
+            command = f"grep 'netvsc.*msglevel' '/boot/System.map-{kernel_version}'"
+            result = node.execute(command, sudo=True)
+            result.assert_exit_code(
```
squirrelsc: `assert_exit_code` can be part of `node.execute` now, check the `expected_exit_code` and `expected_exit_code_failure_message` arguments.

microsoft/testsuites/network/networksettings.py
```diff
@@ -384,3 +387,118 @@ def validate_device_rx_hash_level_change(self, node: Node, log: Logger) -> None:
...
+            reverted_settings = ethtool.set_device_message_flag_by_num(
+                interface, original_msg_level_number
+            )
+            assert_that(
+                reverted_settings.msg_level_number,
+                f"Setting msg flags by number - {original_msg_level_number} didn't"
+                f" succeed. Current msg level value is {new_settings.msg_level_number}",
+            ).contains(original_msg_level_number)
+
+    def _verify_msg_level_change_kernel_support(self, node: Node) -> None:
```
squirrelsc: `_check_msg_level_change_supported` sounds better.

lisa/tools/nm.py
```diff
@@ -0,0 +1,30 @@
...
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("binutils")
+        return self._check_exists()
+
+    def get_symbol_table(self, file: str) -> str:
+        result = self.run(file, shell=True, sudo=True)
```
squirrelsc: Try it like below, integrate exit_code into the call
```python
result = self.run(file, shell=True, sudo=True, expected_exit_code=0, expected_exit_code_failure_message=f"Failed to extract file, {result.stdout}")
```

lisa/tools/modinfo.py
```diff
@@ -5,11 +5,29 @@
...
+    #   srcversion:     BEFCE8DF062B7BE7F9ACEC6
+    #   alias:          acpi*:VMBus:*
+    #   depends:
+    #   intree:         Y
+    #   vermagic:       3.10.0-1160.42.2.el7.x86_64 SMP mod_unload modversions
+    #   signer:         Red Hat Enterprise Linux kernel signing key
+    #   sig_key:        6D:48:2D:7F:D8:1B:CE:44:5F:63:E0:FC:FF:23:4C:EB:55:8D:18:85
+    #   sig_hashalgo:   sha256
+
     __version_pattern = re.compile(r"^version:[ \t]*([^ \n]*)")
```
squirrelsc, line 17: Do examples apply to `__version_pattern`? The pattern require a string start with "version:", but above strings has no such one.

squirrelsc:
Please keep comment unresolved, and I will resolve them, when reviewing again. If you need to track progress, you can use reaction  to mark them.

lisa/tools/modinfo.py
sharsonia: Hmm yeah it actually depends on what we want from version. I have seen multiple times where modinfo output shows srcversion(or something else) instead of version. So do we want to change __version_pattern or keep it as is?
If we want to keep __version_pattern as it is then I would put another example for this pattern.

squirrelsc: Could you try a distro with LIS package? Maybe it have the version field.

sharsonia: ok yes, there I do see version. Let me then put this example for version pattern

[lisa@LISAv2-sosha-test-lisdriver-FR38-1013180622-role-0 ~]$ modinfo hv_vmbus
filename:       /lib/modules/2.6.32-754.29.1.el6.x86_64/kernel/drivers/hv/hv_vmbus.ko
version:        3.1
license:        GPL
retpoline:      Y
srcversion:     D2BAB47E2AE995E9906FAD5
alias:          acpi*:VMBus:*
alias:          acpi*:VMBUS:*
depends:
vermagic:       2.6.32-754.29.1.el6.x86_64 SMP mod_unload modversions

squirrelsc: Good, please update the examples. The examples don't need to include all lines, the matched lines are enough for future updates.

sharsonia:
validate_device_msg_level_change run logs with following azure params
      azure:
        marketplace: RedHat RHEL 79-gen2 7.9.2021051702
        vm_size: Standard_DS5_v2
        location: southcentralus
 
[lisa-20211013-194304-995.log](https://github.com/microsoft/lisa/files/7340913/lisa-20211013-194304-995.log)

sharsonia:
[lisa-20211013-194922-039.log](https://github.com/microsoft/lisa/files/7341518/lisa-20211013-194922-039.log)

squirrelsc (approved):

PR #1572, opened 2021-10-13
Support install gpu driver in ubuntu 21.04.

LiliDeng:


squirrelsc (approved):

squirrelsc:
Please check if anything need to be updated in v3 also.

squirrelsc:
The kernel table is also useful in v3.

PR #1573, opened 2021-10-13
azure tools: LisDriver: Fix get_version method

sharsonia:


lisa/sut_orchestrator/azure/tools.py
```diff
@@ -160,9 +160,13 @@ def _install(self) -> bool:
         return True
 
     def get_version(self) -> str:
-        cmd_result = self.run()
-        if cmd_result.exit_code == 0:
-            found_version = find_patterns_in_lines(
-                cmd_result.stdout, [self.__version_pattern]
-            )
+        self._command = "modinfo hv_vmbus"
```
squirrelsc: set it from `def command`, if it won't be changed.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -160,9 +160,13 @@ def _install(self) -> bool:
         return True
 
     def get_version(self) -> str:
```
squirrelsc: Add force_run argument, and default to `False`. It may be used with LIS package together. That test cases need to change the version of LIS.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -160,9 +160,13 @@ def _install(self) -> bool:
...
-            found_version = find_patterns_in_lines(
-                cmd_result.stdout, [self.__version_pattern]
-            )
+        self._command = "modinfo hv_vmbus"
+        cmd_result = self.run(
+            expected_exit_code=0,
+            expected_exit_code_failure_message="hv_vmbus module not found/loaded.",
+        )
+
+        found_version = find_patterns_in_lines(
```
squirrelsc: use `get_matched_str`

squirrelsc (approved):

PR #1575, opened 2021-10-14
Add verify vhd prerequisite testcases

dsrivastavv:
Migrate : https://github.com/microsoft/lisa/blob/master/Testscripts/Linux/VERIFY-VHD-PREREQUISITES.py

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -908,7 +908,7 @@ class Oracle(Redhat):
 class Suse(Linux):
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
-        return re.compile("^SLES|SUSE|sles|sle-hpc|sle_hpc|opensuse-leap$")
+        return re.compile("^SUSE|opensuse-leap$")
 
     def _initialize_package_installation(self) -> None:
         self.wait_running_process("zypper")
@@ -942,6 +942,12 @@ class Suse(Linux):
         if packages:
             command += " ".join(packages)
         self._node.execute(command, sudo=True, timeout=3600)
+
+
+class SLES(Suse):
+    @classmethod
+    def name_pattern(cls) -> Pattern[str]:
+        return re.compile("^SLES|sles|sle-hpc|sle_hpc$")
 
 
 class NixOS(Linux):
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -12,7 +12,13 @@ from semver import VersionInfo
 
 from lisa.base_tools import Cat, Sed, Wget
 from lisa.executable import Tool
-from lisa.util import BaseClassMixin, LisaException, get_matched_str, parse_version
+from lisa.util import (
+    BaseClassMixin,
+    LisaException,
+    filter_ansi_escape,
+    get_matched_str,
+    parse_version,
+)
 from lisa.util.logger import get_logger
 from lisa.util.perf_timer import create_timer
 from lisa.util.process import ExecutableResult
@@ -23,6 +29,13 @@ if TYPE_CHECKING:
 
 
 _get_init_logger = partial(get_logger, name="os")
+
+
+@dataclass
+# stores information about repository in Posix operating systems
+class RepositoryInfo(object):
+    # name of the repository, for example focal-updates
+    name: str
 
 
 @dataclass
@@ -302,6 +315,9 @@ class Posix(OperatingSystem, BaseClassMixin):
         except FileNotFoundError:
             self._log.debug("File /etc/os-release doesn't exist.")
 
+    def get_repositories(self) -> List[RepositoryInfo]:
+        raise NotImplementedError("get_repositories is not implemented")
+
     def _install_packages(
         self, packages: Union[List[str]], signed: bool = True
     ) -> None:
@@ -433,7 +449,32 @@ class CoreOs(Linux):
         return re.compile("^coreos|Flatcar|flatcar$")
 
 
+@dataclass
+# `apt-get update` repolist is of the form `<status>:<id> <uri> <name> <metadata>`
+# Example:
+# Get:5 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages [1298 kB] # noqa: E501
+class DebianRepositoryInfo(RepositoryInfo):
+    # status for the repository. Examples: `Hit`, `Get`
+    status: str
+
+    # id for the repository. Examples : 1, 2
+    id: str
+
+    # uri for the repository. Example: `http://azure.archive.ubuntu.com/ubuntu`
+    uri: str
+
+    # metadata for the repository. Example: `amd64 Packages [1298 kB]`
+    metadata: str
+
+
 class Debian(Linux):
+
+    # Get:5 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages [1298 kB] # noqa: E501
+    _debian_repository_info_pattern = re.compile(
+        r"(?P<status>\S+):(?P<id>\d+)\s+(?P<uri>\S+)\s+(?P<name>\S+)"
+        r"\s+(?P<metadata>.*)\s*"
+    )
+
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
         return re.compile("^debian|Forcepoint|Kali$")
@@ -467,6 +508,26 @@ class Debian(Linux):
 
         if timeout < timer.elapsed():
             raise Exception("timeout to wait previous dpkg process stop.")
+
+    def get_repositories(self) -> List[RepositoryInfo]:
+        self._initialize_package_installation()
+        repo_list_str = self._node.execute("apt-get update", sudo=True).stdout
+
+        repositories: List[RepositoryInfo] = []
+        for line in repo_list_str.splitlines():
+            matched = self._debian_repository_info_pattern.search(line)
+            if matched:
+                repositories.append(
+                    DebianRepositoryInfo(
+                        name=matched.group("name"),
+                        status=matched.group("status"),
+                        id=matched.group("id"),
+                        uri=matched.group("uri"),
+                        metadata=matched.group("metadata"),
+                    )
+                )
+
+        return repositories
 
     def _initialize_package_installation(self) -> None:
         # wait running system package process.
@@ -704,13 +765,48 @@ class OpenBSD(BSD):
     ...
 
 
+@dataclass
+# yum repolist is of the form `<id> <name>`
+# Example:
+# microsoft-azure-rhel8-eus  Microsoft Azure RPMs for RHEL8 Extended Update Support
+class FedoraRepositoryInfo(RepositoryInfo):
+    # id for the repository, for example: microsoft-azure-rhel8-eus
+    id: str
+
+
 class Fedora(Linux):
     # Red Hat Enterprise Linux Server 7.8 (Maipo) => 7.8
     _fedora_release_pattern_version = re.compile(r"^.*release\s+([0-9\.]+).*$")
 
+    # microsoft-azure-rhel8-eus  Microsoft Azure RPMs for RHEL8 Extended Update Support
+    _fedora_repository_info_pattern = re.compile(r"(?P<id>\S+)\s+(?P<name>\S.*\S)\s*")
+
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
         return re.compile("^Fedora|fedora$")
+
+    def get_repositories(self) -> List[RepositoryInfo]:
+        repo_list_str = self._node.execute(
+            "yum repolist", sudo=True
+        ).stdout.splitlines()
+
+        # skip to the first entry in the output
+        for index, repo_str in enumerate(repo_list_str):
+            if repo_str.startswith("repo id"):
+                header_index = index
+                break
+        repo_list_str = repo_list_str[header_index + 1 :]
+
+        repositories: List[RepositoryInfo] = []
+        for line in repo_list_str:
+            repo_info = self._fedora_repository_info_pattern.search(line)
+            if repo_info:
+                repositories.append(
+                    FedoraRepositoryInfo(
+                        name=repo_info.group("name"), id=repo_info.group("id")
+                    )
+                )
+        return repositories
 
     def _install_packages(
         self, packages: Union[List[str]], signed: bool = True
@@ -905,10 +1001,74 @@ class Oracle(Redhat):
         return re.compile("^Oracle")
 
 
+@dataclass
+# `zypper lr` repolist is of the form
+# `<id>|<alias>|<name>|<enabled>|<gpg_check>|<refresh>`
+# Example:
+# # 4 | repo-oss            | Main Repository             | Yes     | (r ) Yes  | Yes
+class SuseRepositoryInfo(RepositoryInfo):
+    # id for the repository. Example: 4
+    id: str
+
+    # alias for the repository. Example: repo-oss
+    alias: str
+
+    # is repository enabled. Example: True/False
+    enabled: bool
+
+    # is gpg_check enabled. Example: True/False
+    gpg_check: bool
+
+    # is repository refreshed. Example: True/False
+    refresh: bool
+
+
 class Suse(Linux):
+    # 55 | Web_and_Scripting_Module_x86_64:SLE-Module-Web-Scripting15-SP2-Updates                           | SLE-Module-Web-Scripting15-SP2-Updates                  | Yes     | ( p) Yes  | Yes # noqa: E501
+    # 4 | repo-oss            | Main Repository             | Yes     | (r ) Yes  | Yes # noqa: E501
+    _zypper_table_entry = re.compile(
+        r"\s*(?P<id>\d+)\s+[|]\s+(?P<alias>\S.+\S)\s+\|\s+(?P<name>\S.+\S)\s+\|"
+        r"\s+(?P<enabled>\S.*\S)\s+\|\s+(?P<gpg_check>\S.*\S)\s+\|"
+        r"\s+(?P<refresh>\S.*\S)\s*"
+    )
+
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
         return re.compile("^SUSE|opensuse-leap$")
+
+    def get_repositories(self) -> List[RepositoryInfo]:
+        # Parse ouput of command "zypper lr"
+        # Example output:
+        # 1 | Basesystem_Module_x86_64:SLE-Module-Basesystem15-SP2-Debuginfo-Pool                              | SLE-Module-Basesystem15-SP2-Debuginfo-Pool              | No      | ----      | ---- # noqa: E501
+        # 2 | Basesystem_Module_x86_64:SLE-Module-Basesystem15-SP2-Debuginfo-Updates                           | SLE-Module-Basesystem15-SP2-Debuginfo-Updates           | No      | ----      | ---- # noqa: E501
+        self._initialize_package_installation()
+        output = filter_ansi_escape(self._node.execute("zypper lr", sudo=True).stdout)
+        repo_list: List[RepositoryInfo] = []
+
+        for line in output.splitlines():
+            matched = self._zypper_table_entry.search(line)
+            if matched:
+                is_repository_enabled = (
+                    True if "Yes" in matched.group("enabled") else False
+                )
+                is_gpg_check_enabled = (
+                    True if "Yes" in matched.group("gpg_check") else False
+                )
+                is_repository_refreshed = (
+                    True if "Yes" in matched.group("refresh") else False
+                )
+                if matched:
+                    repo_list.append(
+                        SuseRepositoryInfo(
+                            name=matched.group("name"),
+                            id=matched.group("id"),
+                            alias=matched.group("alias"),
+                            enabled=is_repository_enabled,
+                            gpg_check=is_gpg_check_enabled,
+                            refresh=is_repository_refreshed,
+                        )
+                    )
+        return repo_list
 
     def _initialize_package_installation(self) -> None:
         self.wait_running_process("zypper")
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -1,0 +1,543 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from pathlib import PurePosixPath
+from typing import cast
+
+from assertpy.assertpy import assert_that
+
+from lisa import Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa.operating_system import (
+    SLES,
+    CoreOs,
+    Debian,
+    DebianRepositoryInfo,
+    Fedora,
+    FedoraRepositoryInfo,
+    Oracle,
+    Posix,
+    Redhat,
+    Suse,
+    SuseRepositoryInfo,
+    Ubuntu,
+)
+from lisa.tools import Cat, Dmesg, Pgrep
+from lisa.util import (
+    LisaException,
+    PassedException,
+    SkippedException,
+    find_patterns_in_lines,
+    get_matched_str,
+)
+
+
+@TestSuiteMetadata(
+    area="azure_image_standard",
+    category="functional",
+    description="""
+    This test suite is used to check azure image configuration.
+    """,
+)
+class AzureImageStandard(TestSuite):
+
+    # Defaults targetpw
+    _uncommented_default_targetpw_regex = re.compile(
+        r"(\nDefaults\s+targetpw)|(^Defaults\s+targetpw.*)"
+    )
+
+    # repo-oss
+    _oss_repo_regex = re.compile(r"^(?!.*(debug|non)).*oss.*$")
+
+    # suse_cloud_application_platform_tools_module_x86_64:sle-module-cap-tools15-sp2-debuginfo-updates
+    _update_repo_regex = re.compile(r"^(?!.*(debug|non)).*update.*$")
+
+    # rhui-microsoft-azure-rhel7                                       Micr enabled: 8
+    #  rhui-rhel-7-server-rhui-optional-rpms/7Server/x86_64             Red  disabled
+    _redhat_repo_regex = re.compile(
+        r"rhui-(?P<name>\S+)\s+(?P<source>\S+)\s+(enabled|disabled)"
+    )
+
+    @TestCaseMetadata(
+        description="""
+        This test will verify that `Defaults targetpw` is not enabled in the
+        `/etc/sudoers` file.
+
+        If `targetpw` is set, `sudo` will prompt for the
+        password of the user specified by the -u option (defaults to root)
+        instead of the password of the invoking user when running a command
+        or editing a file. More information can be found here :
+        https://linux.die.net/man/5/sudoers
+
+        Steps:
+        1. Get the content of `/etc/sudoers` file.
+        2. Verify that `Defaults targetpw` should be disabled, if present.
+        """,
+        priority=1,
+    )
+    def verify_default_targetpw(self, node: Node) -> None:
+        sudoers_out = (
+            node.tools[Cat].run("/etc/sudoers", sudo=True, force_run=True).stdout
+        )
+        matched = get_matched_str(sudoers_out, self._uncommented_default_targetpw_regex)
+        assert_that(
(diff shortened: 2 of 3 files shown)


microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+                grub_output = node.tools[Cat].read_from_file(
+                    "/boot/grub2/grub.cfg", sudo=True
+                )
+            elif node.shell.exists(PurePosixPath("/boot/grub/grub.conf")):
+                grub_output = node.tools[Cat].read_from_file(
+                    "/boot/grub/grub.conf", sudo=True
+                )
+            else:
+                raise LisaException("Unable to locate grub file")
+        elif isinstance(node.os, Redhat) or isinstance(node.os, Fedora):
```
squirrelsc: Check with Fedora should be enough. The Redhat is derived from Fedora.

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+                grub_output = node.tools[Cat].read_from_file(
+                    "/boot/grub/menu.lst", sudo=True
+                )
+            else:
+                raise LisaException("Unable to locate grub file")
+        elif isinstance(node.os, CoreOs):
+            # in core os we don't have access to boot partition
+            grub_output = node.tools[Dmesg].run().stdout
+        else:
+            raise SkippedException(f"Test cannot run on distro {node.os}")
```
squirrelsc: Raise LisaException here. We may support more distros later.

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+
+        assert_that(grub_output).contains("console=ttyS0")
```
squirrelsc: Add `describe_as` to explain the failure.

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+        Steps:
+        1. Get the output of command `rpm -q NetworkManager` and verify that
+        network manager is not installed.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_os=[Redhat, Fedora]),
+    )
+    def verify_network_manager_not_installed(self, node: RemoteNode) -> None:
+        network_manager_output = node.execute("rpm -q NetworkManager").stdout
+        if isinstance(node.os, Redhat) and node.os.information.version.major > 7:
```
squirrelsc: check version by `node.os.information.version > "7.0.0"`. Other's may refer to the code, and it's hard to know how to check minor version.

squirrelsc: Should it be `>` or `>=`?

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+            raise SkippedException(
+                "NetworkManager package no longer conflicts with the "
+                "waagent on Redhat 7.0+"
+            )
+        assert_that(network_manager_output).contains("is not installed")
```
squirrelsc: add `describe_as` to explain the failure.

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+        description="""
+        This test will verify that network manager doesn't conflict with the
+        waagent on fedora and redhat distros.
+
+        Steps:
+        1. Get the output of command `rpm -q NetworkManager` and verify that
+        network manager is not installed.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_os=[Redhat, Fedora]),
```
squirrelsc: Fedora is enough here.

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
             matched, "Defaults targetpw should not be enabled in /etc/sudoers"
         ).is_length(0)
 
+    @TestCaseMetadata(
+        description="""
+        This test will verify that network manager doesn't conflict with the
```
squirrelsc: What's the reason to check fedora only? Explain it here.

squirrelsc: The description is not accurate. The test logic asks the network manager "not to installed", instead of "doesn't conflict".

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+        is enabled on fedora and redhat distros.
+
+        Steps:
+        1. Verify that network file exists.
+        2. Verify that networking is enabled in the file.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_os=[Redhat, Fedora]),
+    )
+    def verify_network_file_configuration(self, node: RemoteNode) -> None:
```
squirrelsc: If no special reason, use "Node" type, instead of "RemoteNode".

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+    @TestCaseMetadata(
+        description="""
+        This test will verify that udev rules have been moved out.
+
+        Steps:
+        1. Verify that 75-persistent-net-generator.rules and 70-persistent-net.rules
+        files are not present.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_os=[Redhat, Fedora, CoreOs]),
```
squirrelsc: What's the reason to check only those OSes? Some limitation in V2 may not make sense, we need to double check and update or document it.

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+        ).is_false()
+        assert_that(
+            node.shell.exists(PurePosixPath(udev_file_path_70_rule)),
+            f"file {udev_file_path_70_rule} should not be present",
+        ).is_false()
+
+    @TestCaseMetadata(
+        description="""
+        This test will check that grub configuration file exists and numa is
+        disabled for version < 6.6
```
squirrelsc: which version is 6.6?

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+
+        Steps:
+        1. Verify that grub file exists depending on the distro type.
+        2. For redhat based distros, verify that numa is disabled for version < 6.6
```
squirrelsc: What's the reason to mention about Redhat 6.6? And this version should end of life. We don't need to migrate the logic.

microsoft/testsuites/core/storage.py
```diff
@@ -187,6 +189,166 @@ def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
...
+        else:
+            raise SkippedException(f"Test cannot run on distro {node.os}")
+
+        assert_that(grub_output).contains("console=ttyS0")
+        assert_that(grub_output).does_not_contain("libata.atapi_enabled=0")
+        assert_that(grub_output).does_not_contain("reserve=0x1f0,0x8")
+
+        if isinstance(node.os, Redhat):
+            # check numa=off in grub for CentOS 6.x and Oracle Linux 6.x
+            if (
```
squirrelsc: it should be like `node.os.information.version < "6.6.0"`

microsoft/testsuites/core/storage.py
dsrivastavv: The v2 test cases only cover specific distros and the logic here is very specific to these distros. I will try to find the reason around this. @LiliDeng Do you have some idea around this?

lisa/tools/pgrep.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from typing import List
+
+from lisa.executable import Tool
+
+
+class Proc(object):
```
squirrelsc: Use the full name "Process"

lisa/tools/pgrep.py
```diff
@@ -0,0 +1,38 @@
...
+class Pgrep(Tool):
+    _process_map_regex = re.compile(r"(?P<id>\d+)\s+(?P<name>\S+)\s*")
+
+    @property
+    def command(self) -> str:
+        return "pgrep"
+
+    def get_running_processes(self, process_identifier: str) -> List[Proc]:
+        running_process: List[Proc] = []
+        output = self.run(f"-lf {process_identifier}", sudo=True).stdout
```
squirrelsc: The `f` use the identifier to match the command line too, it makes the result is not accurate.

lisa/tools/pgrep.py
```diff
@@ -0,0 +1,38 @@
...
+        for line in output.splitlines():
```
squirrelsc: use `find_patterns_in_lines` to match multiple items and return.

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,392 @@
...
+        1. Get the content of `/etc/sudoers` file.
+        2. Verify that `Defaults targetpw` should be disabled, if present.
+        """,
+        priority=1,
+    )
+    def verify_default_targetpw(self, log: Logger, node: RemoteNode) -> None:
+        sudoers_out = (
+            node.tools[Cat].run("/etc/sudoers", sudo=True, force_run=True).stdout
+        )
+        matched = self._uncommented_default_targetpw_regex.findall(sudoers_out)
```
squirrelsc: use `get_matched_str`

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,392 @@
...
+    def verify_network_manager_not_installed(self, node: RemoteNode) -> None:
+        network_manager_output = node.execute("rpm -q NetworkManager").stdout
+        if isinstance(node.os, Redhat) and node.os.information.version >= "7.0.0":
+            # NetworkManager package no longer conflicts with the waagent on
+            # Redhat >= "7.0.0"
+            raise SkippedException(
+                "NetworkManager package no longer conflicts with the "
+                "waagent on Redhat 7.0+"
+            )
+        assert_that(network_manager_output).contains("is not installed")
```
squirrelsc: add `described_as` to explain the failure.

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,392 @@
...
+                    or "debian-archive.trafficmanager.net" in repository_out
+                ) and "Hit" in repository_out
+
+            assert_that(
+                is_repository_configured_correctly,
+                "repository not configured correctly",
+            ).is_true()
+        elif isinstance(node.os, Suse):
+            zypper_out = node.execute("zypper lr")
+            oss_repo_count = node.execute(
```
squirrelsc: 1. Avoid to use the pipe and grep in command line, it hides the details
2. The `zypper lr` from the zypper_out variable can be reused, it doesn't need to run it many times.

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,392 @@
...
+                ).contains("appstream")
+            else:
+                assert_that(repolist, "Updates repository should be present").contains(
+                    "updates"
+                )
+            assert_that(
+                ra_out, "yum repolist all should be greater than 5"
+            ).is_greater_than(5)
+        else:
+            raise LisaException(f"Unknown distro type : {type(node.os)}")
```
squirrelsc: It's not "unknown", it's "unsupported".

lisa/tools/pgrep.py
dsrivastavv: There is already a class name Process in lisa\util\process.py

dsrivastavv: Should I rename it to something like LinuxProcessInfo?

lisa/tools/pgrep.py
dsrivastavv: Sorry I didn't get this comment. Can you please elaborate?

microsoft/testsuites/core/azure_vhd.py
dsrivastavv: I will update it to run it once and do the processing in the python code

lisa/tools/pgrep.py
squirrelsc: Yes, it can be named like ProcessInfo. And it's ok to have same name, if they are not in the same module.

lisa/tools/pgrep.py
squirrelsc: It's about the argument of `pgrep`. The `-l` is enough, it shows the process name. The `-f` is not needed, because it searches the path too, not only the process name.

lisa/tools/pgrep.py
```diff
@@ -0,0 +1,38 @@
...
+            matched = self._process_map_regex.fullmatch(line)
+            assert matched
+            running_process.append(Proc(matched.group(0), matched.group(1)))
+
+        print(running_process)
```
squirrelsc: remove the test print.

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,392 @@
...
+            grub_output, "console=ttyS0 should be present in grub file."
+        ).contains("console=ttyS0")
+        assert_that(
+            grub_output, "libata.atapi_enabled=0 should not be present in grub file."
+        ).does_not_contain("libata.atapi_enabled=0")
+        assert_that(
+            grub_output, "reserve=0x1f0,0x8 should not be present in grub file."
+        ).does_not_contain("reserve=0x1f0,0x8")
+
+        # check numa=off in grub for Redhat version < 6.6.0
```
LiliDeng: add link here https://access.redhat.com/solutions/436883

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,392 @@
...
+        is enabled on Fedora based distros.
+
+        Steps:
+        1. Verify that network file exists.
+        2. Verify that networking is enabled in the file.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_os=[Fedora]),
+    )
+    def verify_network_file_configuration(self, node: RemoteNode) -> None:
```
LiliDeng: can we combine some test cases which related with /etc/sysconfig/network into one case?

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,392 @@
...
+        dhcp_file_path = "/etc/sysconfig/network/dhcp"
+        file_exists = node.shell.exists(PurePosixPath(dhcp_file_path))
+
+        assert_that(
+            file_exists,
+            f"The dhcp file should be present at {dhcp_file_path}",
+        ).is_true()
+
+        dhcp_file = node.tools[Cat].read_from_file(dhcp_file_path)
+        assert_that(dhcp_file).contains('DHCLIENT_SET_HOSTNAME="no"')
```
LiliDeng: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/suse-create-upload-vhd#prepare-suse-linux-enterprise-server-for-azure

microsoft/testsuites/core/azure_vhd.py
dsrivastavv: What would be advantage here of combining testcases? Do these testcases need to run in order?

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,407 @@
...
+
+
+@TestSuiteMetadata(
+    area="azure_vhd",
+    category="functional",
+    description="""
+    This test suite is used to check azure vhd configuration.
+    """,
+)
+class AzureVHD(TestSuite):
```
squirrelsc: The "VHD" is not accurate, how about AzureVM?

dsrivastavv: I named it VHD because we are testing distro images. AzureVM might incorrectly suggest that we want to test VM's. What do you think?

squirrelsc: How about AzureImageStandard? The VHD is the format and file extension. It may be used in Hyper-V or directly.

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,407 @@
...
+        priority=1,
+    )
+    def verify_default_targetpw(self, log: Logger, node: Node) -> None:
+        sudoers_out = (
+            node.tools[Cat].run("/etc/sudoers", sudo=True, force_run=True).stdout
+        )
+        matched = get_matched_str(sudoers_out, self._uncommented_default_targetpw_regex)
+        assert_that(
+            matched, "Defaults targetpw should not be enabled in /etc/sudoers"
+        ).is_length(0)
```
squirrelsc: is_empty is better on a string.

lisa/tools/zypper.py
```diff
@@ -0,0 +1,57 @@
...
+        oss_repo_count = 0
+        update_repo_count = 0
+        oss_repo_enable_refresh_count = 0
+        update_repo_enable_refresh_count = 0
+        for repo in repo_list:
+            full_name = f"{repo[1]} {repo[2]}"
+            full_name = full_name.lower()
+
+            # check if oss repo
+            if re.match(r"^(\?!.*(debug|non)).*oss.*$", full_name):
```
squirrelsc: move those pattern's to class level, and precompiled.

lisa/tools/zypper.py
```diff
@@ -0,0 +1,57 @@
...
+                oss_repo_count += 1
+                if repo[5] == "Yes":
```
squirrelsc: Why is there a magic number? If it's necessary, add comments here. I found the find_patterns_in_lines doesn't support named group yet. If so, please use other way to make sure the group name can be used.

microsoft/testsuites/core/azure_vhd.py
dsrivastavv: AzureImageStandard sounds right for this scenario. I will update it. Thanks!

microsoft/testsuites/core/azure_vhd.py
```diff
@@ -0,0 +1,407 @@
...
+
+        Steps:
+        1. Verify the repository configuration depending on the distro type.
+        """,
+        priority=1,
+    )
+    def verify_repository_installed(self, node: Node) -> None:
+        if isinstance(node.os, Debian):
+            # update repository
+            node.execute(
```
squirrelsc: Those steps are handled by os level already.

squirrelsc: You can install a small package or exiting package to trigger it's updated.

lisa/tools/zypper.py
```diff
@@ -0,0 +1,57 @@
...
+    _zypper_table_entry = re.compile(
+        r"(?P<id>\d+)\s*\|\s*(?P<alias>\S+)\s*\|\s*(?P<name>\S+)\s*\|"
+        r"\s*(?P<enabled>\S+)\s*\|\s*(?P<gpg_check>\S+)\s*\|\s*(?P<Refresh>\S+)\s*"
+    )
+
+    @property
+    def command(self) -> str:
+        return "zypper"
+
+    def get_repo_statistics(self) -> Tuple[int, int, int, int]:
```
squirrelsc: The returned value is not easy to use, and hides some details. Can it return the queried string list? It's easier to check which one is missed, or what exists.

microsoft/testsuites/core/azure_vhd.py
dsrivastavv: So it would be okay to remove these steps, right?

lisa/tools/zypper.py
dsrivastavv: I will add some documentation to make it more clearly. Basically it's the 5th column of the output

microsoft/testsuites/core/azure_vhd.py
squirrelsc: yes, but you need to call `install_packages` to make sure the apt update is triggered.

lisa/tools/zypper.py
squirrelsc: Use named capture group is better. search `group(` in operating_system.py to see examples.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -0,0 +1,448 @@
...
+)
+class AzureImageStandard(TestSuite):
+
+    # Defaults targetpw
+    _uncommented_default_targetpw_regex = re.compile(
+        r"(\nDefaults\s+targetpw)|(^Defaults\s+targetpw.*)"
+    )
+
+    # suse_cloud_application_platform_tools_module_x86_64:sle-module-cap-tools15-sp2-debuginfo-oss
+    _oss_repo_regex = re.compile(r"^(?!.*(debug|non)).*oss.*$")
```
squirrelsc, line 50: It looks the pattern can be simpler to check end with `-oss`, is it right?
`^.*-oss$`

squirrelsc: If not, provide one more example in code, which doesn't fit the simple pattern.

lisa/tools/zypper.py
```diff
@@ -0,0 +1,67 @@
...
+        self.run("refresh", sudo=True)
+
+    def get_repo_statistics(self) -> List[ZypperInfo]:
+        # Parse ouput of command "zypper lr"
+        # Example output:
+        # 1 | Basesystem_Module_x86_64:SLE-Module-Basesystem15-SP2-Debuginfo-Pool                              | SLE-Module-Basesystem15-SP2-Debuginfo-Pool              | No      | ----      | ---- # noqa: E501
+        # 2 | Basesystem_Module_x86_64:SLE-Module-Basesystem15-SP2-Debuginfo-Updates                           | SLE-Module-Basesystem15-SP2-Debuginfo-Updates           | No      | ----      | ---- # noqa: E501
+        output = filter_ansi_escape(self.run("lr", sudo=True).stdout)
+
+        repo_list_str = find_patterns_in_lines(output, [self._zypper_table_entry])[0]
```
squirrelsc: It's better to use named groups, not to use the `find_patterns_in_lines`.

lisa/tools/zypper.py
```diff
@@ -0,0 +1,67 @@
...
+        r"\s*(?P<id>\d+)\s+[|]\s+(?P<alias>\S+)\s+\|\s+(?P<name>\S+)\s+\|"
+        r"\s+(?P<enabled>\S+)\s+\|\s+(?P<gpg_check>.+)\s+\|\s+(?P<Refresh>\S+)\s*"
+    )
+
+    @property
+    def command(self) -> str:
+        return "zypper"
+
+    def refresh(self) -> None:
+        self.run("refresh", sudo=True)
```
squirrelsc: It's handled by Suse OS already.

lisa/operating_system.py
```diff
@@ -25,6 +31,17 @@
 _get_init_logger = partial(get_logger, name="os")
 
 
+class RepositoryInfo(object):
```
squirrelsc, line 36: It can be a dataclass, it's easier to maintain the fields.

lisa/operating_system.py
```diff
@@ -25,6 +31,17 @@
...
+    def __init__(self, name: str):
```
squirrelsc: I saw id is in every subclass, should it be in the base class?

lisa/operating_system.py
```diff
@@ -433,7 +450,34 @@ def name_pattern(cls) -> Pattern[str]:
...
+        return (
+            f"({super().__repr__()}, Status: {self.status}, Id:{self.id}, "
+            "Url:{self.url}, Metadata={self.metadata})"
+        )
+
+
 class Debian(Linux):
+
+    # Get:5 http://azure.archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages [1298 kB] # noqa: E501
+    _debian_repository_info_pattern = re.compile(
```
squirrelsc, line 473: The distro info pattern can be in the repo info class. So, the parse logic can be in there too.  It's easier to maintain its fields and pattern together.
you can add a method called parse, it returns the repo info instance.

lisa/operating_system.py
```diff
@@ -905,10 +1007,65 @@ def name_pattern(cls) -> Pattern[str]:
...
+        r"\s+(?P<enabled>\S.+\S)\s+\|\s+(?P<gpg_check>\S.+\S)\s+\|"
+        r"\s+(?P<refresh>\S.+\S)\s*"
+    )
+
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
-        return re.compile("^SLES|SUSE|sles|sle-hpc|sle_hpc|opensuse-leap$")
+        return re.compile("^SUSE|opensuse-leap$")
+
+    def get_repositories(self) -> List[SuseRepositoryInfo]:
```
squirrelsc: 1. Add this method in Posix, and raise NotImplmentedException. 
2. The returned value can be base class.

lisa/operating_system.py
dsrivastavv: name here is a unique value representing a repository. It could be a case for certain distro that only name is available which is why I left the `id` parameter out of `RepositoryInfo` class.

dsrivastavv: Also, the meaning of id is different for different classes. In ubuntu it's a counter whereas in fedora it's a name.

lisa/operating_system.py
dsrivastavv: We will also need a method to return distro specific regex pattern. How about I move this logic to distro specific `RepositoryInfo` class?

squirrelsc: Yes, the pattern can be moved to repo info. The pattern is a variable, and overrides in the subclasses.

dsrivastavv: On a second thought it might not be very straightforward to have a common get_repository function. For example, in fedora class we have an additional logic to skip to first line of the repo table. 
```
 # skip to the first entry in the output
        for index, repo_str in enumerate(repo_list_str):
            if repo_str.startswith("repo id"):
                header_index = index
                break
        repo_list_str = repo_list_str[header_index + 1 :]
```

This cannot be a common code. This case could be there in other distros too

squirrelsc: The common `get_repositories` is a function signature, so that if it's called like `posix_os.get_repositories`, it will return the same results or not implemented exception. The whole implementation can be in sub classes.

lisa/operating_system.py
```diff
@@ -25,6 +31,19 @@
 _get_init_logger = partial(get_logger, name="os")
 
 
+@dataclass
+# stores information about repository in Posix operating systems
+class RepositoryInfo(object):
+    # name of the repository, for example focal-updates
+    name: str
+
+    def __repr__(self) -> str:
```
squirrelsc: Are they needed for data class? I suppose data class can show them well.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -0,0 +1,535 @@
...
+        1. Verify the repository configuration depending on the distro type.
+        """,
+        priority=1,
+    )
+    def verify_repository_installed(self, node: Node) -> None:
+        if isinstance(node.os, Debian):
+            # add comment
+            debian_repositories: List[
+                DebianRepositoryInfo
+            ] = node.os.get_repositories()  # type: ignore
```
squirrelsc: This line should be above the if condition now, because every os has it now.

microsoft/testsuites/core/azure_image_standard.py
dsrivastavv: I was able to get a distro whioch contained oss repo. But I am not sure what a complex example would look like. The entry present in the distro was  : `4 | repo-oss            | Main Repository             | Yes     | (r ) Yes  | Yes` and this regex matches `repo-oss` part of it

squirrelsc (approved):

PR #1576, opened 2021-10-15
[LISAv3] Adding dotnet tests for docker test suite

isaiahduncan:
Adding dotnet and wordpress docker tests
Adding relavant test scripts for the two new tests

This will only be changed from draft to pull request after https://github.com/microsoft/lisa/pull/1560 is merged

### DIFF
--- a/microsoft/testsuites/docker/TestScripts/dotnet50.Dockerfile
+++ b/microsoft/testsuites/docker/TestScripts/dotnet50.Dockerfile
@@ -1,0 +1,7 @@
+FROM mcr.microsoft.com/dotnet/sdk:5.0
+RUN dotnet new console -o app -n helloworld
+WORKDIR /app
+RUN dotnet run
+RUN dotnet publish -c Release
+WORKDIR /app/bin/Release/net5.0
+ENTRYPOINT ["dotnet", "helloworld.dll"]
--- a/microsoft/testsuites/docker/dockersuite.py
+++ b/microsoft/testsuites/docker/dockersuite.py
@@ -16,10 +16,51 @@ from lisa.util import SkippedException
     area="docker",
     category="functional",
     description="""
-    This test suite runs the docker test cases for java, python, dotnet, and wordpress.
+    This test suite runs the docker test cases for java, python, dotnet 3.1
+    , dotnet5.0, and wordpress.
     """,
 )
 class docker(TestSuite):  # noqa
+    @TestCaseMetadata(
+        description="""
+            This test case creates and runs a dotnet app using docker
+
+            Steps:
+            1. Install Dotnet 3.1 sdk
+            2. Install Docker
+            3. Copy dotnet dockerfile into node
+            4. Create docker image and run docker container
+            5. Check results of docker run against dotnet string identifier
+        """,
+        priority=2,
+    )
+    def docker_dotnet31_app(self, node: Node) -> None:
+        dockerfile = "dotnet31.Dockerfile"
+
+        self._execute_docker_test(
+            node, "dotnetimage", "dotnetapp", "Hello World!", "", dockerfile
+        )
+
+    @TestCaseMetadata(
+        description="""
+            This test case creates and runs a dotnet app using docker
+
+            Steps:
+            1. Install Dotnet 5.0 sdk
+            2. Install Docker
+            3. Copy dotnet dockerfile into node
+            4. Create docker image and run docker container
+            5. Check results of docker run against dotnet string identifier
+        """,
+        priority=2,
+    )
+    def docker_dotnet50_app(self, node: Node) -> None:
+        dockerfile = "dotnet50.Dockerfile"
+
+        self._execute_docker_test(
+            node, "dotnetimage", "dotnetapp", "Hello World!", "", dockerfile
+        )
+
     @TestCaseMetadata(
         description="""
             This test case creates and runs a java app using docker
@@ -81,7 +122,8 @@ class docker(TestSuite):  # noqa
         docker_tool = self._verify_and_remove_containers(
             node, docker_image_name, docker_container_name
         )
-        self._copy_to_node(node, prog_src)
+        if prog_src:
+            self._copy_to_node(node, prog_src)
         self._copy_to_node(node, dockerfile)
         self._run_and_verify_results(
             node,
--- a/microsoft/testsuites/docker/TestScripts/dotnet31.Dockerfile
+++ b/microsoft/testsuites/docker/TestScripts/dotnet31.Dockerfile
@@ -1,0 +1,7 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+RUN dotnet new console -o app -n helloworld
+WORKDIR /app
+RUN dotnet run
+RUN dotnet publish -c Release
+WORKDIR /app/bin/Release/netcoreapp3.1
+ENTRYPOINT ["dotnet", "helloworld.dll"]


squirrelsc:
At this stage, the new branch can based on the previous one, so it's easier to review. You can click edit to change the merged branch.

microsoft/testsuites/docker/TestScripts/dotnet2_2.Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:2.2
```
squirrelsc: The current dotnet sdk is 5.0, 6.0 now. Please check if dotnet core 2.2, 3.1 is out of support life cycle, and add another newer version of dotnet sdk.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -64,6 +86,35 @@ def docker_python_app(self, node: Node) -> None:
...
+        app_folder = "app"
+        node.execute(
+            f"test -d {app_folder} && rm -rf {app_folder}",
+            shell=True,
+            cwd=node.working_path,
+        )
+
+        node.execute(
+            f"dotnet new console -o {app_folder} -n helloworld",
+            cwd=node.working_path,
```
squirrelsc: The node.working_path is for the node. You can put tools in this folder, but not use it in test case, because it may be conflict among test cases. Use the "log_path" from test case method. Refer to provisioning.py.

```python
    def verify_deployment_provision_synthetic_nic(
        self, log: Logger, node: Node, log_path: Path
    ) -> None:
```

microsoft/testsuites/docker/TestScripts/dotnet2_2.Dockerfile
squirrelsc: As the web page, let's test on 5.0 and 3.1. They are current and LTS versions.

https://dotnet.microsoft.com/download/dotnet

isaiahduncan: 3.1 has long term support but I'll switch to 5.0 for everything

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -64,6 +86,35 @@ def docker_python_app(self, node: Node) -> None:
...
+        result = node.execute(f"curl --head --silent --fail {package_name}")
+        if not result.exit_code == expected_exit_code:
+            raise SkippedException(
+                f"{package_name} not available on {node.os.information.vendor}"
+            )
+
+    def _compile_dotnet_app(self, node: Node) -> None:
+        node.log.debug("Compiling dotnet app.")
+        app_folder = "app"
+        node.execute(
```
squirrelsc: It can call like `node.shell.remove`

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -64,6 +86,35 @@ def docker_python_app(self, node: Node) -> None:
             "python.Dockerfile",
         )
 
+    def _check_dotnet_sdk_support(self, node: Node, package_name: str) -> None:
```
squirrelsc: Because this method is to test the sdk version, so the `package_name` should be `sdk_version`.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -92,6 +144,48 @@ def _execute_docker_test(
...
+                f"/{package_name}"
+            )
+            node.log.debug(f"Checking dotnet sdk support for {package_name}")
+            self._check_dotnet_sdk_support(node, package_name)
+
+            node.execute(f"rpm -ivh {package}")
+            node.os.install_packages(f"dotnet-sdk-{sdk_version}")
+
+        else:
+            raise SkippedException(f"{node.os.information.vendor} not supported")
```
squirrelsc: raise LISAException, and we will support it later.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: When I make dotnet into a tool, should I put the code in this method into the "can_install" method?

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -92,6 +144,48 @@ def _execute_docker_test(
             string_identifier,
         )
 
+    def _install_dotnet_sdk(self, node: Node, sdk_version: str) -> str:
+        node.log.debug("Installing Dotnet SDK")
+        package_name = "packages-microsoft-prod"
+        distro = node.os.information.vendor.lower()
+        release_id = node.os.information.release
+
+        if isinstance(node.os, Debian):
```
squirrelsc: Check below link, to see if install from script is easier. It doesn't need to add different repo, and support all distros.

https://dotnet.microsoft.com/download/dotnet/5.0

microsoft/testsuites/docker/dockersuite.py
squirrelsc: Yes, it can be. Because the dotnet need to support multiple versions, you need use subclasses like below

```python
class Dotnet(Tool):
    _version = "3.1"

    @property
        # check self._version exists
        ...

class Dotnet50(Dotnet):
    _version = "5.0"

class Dotnet31(Dotnet):
    _version = "3.1"
```

microsoft/testsuites/docker/TestScripts/dotnet2_2.Dockerfile
squirrelsc: we need to test 3.1 and 5.0 both. Because it's ms product. Please add two test cases for them.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: In the java and python tests, we put the docker files and the src files into the working path directory and I think neither of those are tools. How is this any different from that? Should we move where the dockerfiles and src files go?

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: what makes you put 3.1 as the version in the "Dotnet" class? Should it be defined in that class?

microsoft/testsuites/docker/dockersuite.py
squirrelsc: The goal is to avoid conflict easier, actually they have the similar situation. But the Java and Python has one version only, so it won't conflict in the same node. For the dotnet, we need to test multiple versions, it would be easier to conflict.

squirrelsc: The copied source, docker files are ok. Because they are not changed. But the dotnet new console will create a new project with a couple of files.

microsoft/testsuites/docker/dockersuite.py
squirrelsc: As the current design, the tools is retrieved by the type, it doesn't support any parameter. So it has to use the type to pass the version parameter. If this kind of requirement comes more, I will support to passing parameters to retrieve a tool.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: Then for the dotnet tests, I think it would be better if we put the dockerfiles in the log_path directory because the dockerfile needs to have a good relative path to the dotnet app folder. Please let me know if that's ok or if there's a better way

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: I understand why we have the _version variable. What I don't understand is why we chose 3.1 as the value for the variable in the Dotnet superclass. I think that its value should be "" (an empty string) and that it should not be used. Because if it were to be used and the version were to be 3.1, there would be no need for the Dotnet31 class.

squirrelsc: That's to make classes meaningful and easy to use. When we test the dotnet, the "dotnet31" self-explained what's tested. When we test other things, but the dotnet is needed, the "dotnet" is the default dotnet tool to serve other test purpose.

microsoft/testsuites/docker/dockersuite.py
squirrelsc: Sorry, I indicate the wrong thing. The log_path is a local path, not able to use in the node. I will create another working_path for test case. Before that, please keep to use the node's working_path.

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,109 @@
...
+            sudo=True,
+            cwd=self.node.working_path,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Error while installing Dotnet 3.1",
+        )
+        return True
+
+
+class Dotnet31(Dotnet):
+    _version = "3.1"
```
squirrelsc: Write like below, so the version can be used in the `_install` method directly.

`_version = "3.1.20"`

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,109 @@
...
+from lisa.base_tools import Wget
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.util import LisaException
+
+# from lisa.tools.service import Service
+
+
+class Dotnet(Tool):
+    _version = "3.1"
```
squirrelsc: The version is not used actually. It needs to be used to make sure subclass share more code.

squirrelsc: I double checked the install script, it can be used as "latest". So the script will be `./dotnet-install.sh --version latest`. The default behavior is to install the latest version.

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,109 @@
...
+            url="https://dotnet.microsoft.com/download"
+            "/dotnet/scripts/v1/dotnet-install.sh",
+            filename="dotnet-install.sh",
+            file_path=str(self.node.working_path),
+        )
+
+    def _install(self) -> bool:
+        self._get_install_script()
+        self.node.execute(
+            "./dotnet-install.sh --version 3.1.20",
```
squirrelsc: the `3.1.20` should be from the _version.

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _install(self) -> bool:
+        super()._install()
+        return True
+
+
+class Dotnet50(Dotnet):
+    _version = "5.0"
+
+    def _install(self) -> bool:
```
squirrelsc: If the _version is used in super class, it doesn't need to rewrite this method.

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,109 @@
...
+        self._get_install_script()
+        self.node.execute(
+            "./dotnet-install.sh --version 3.1.20",
+            sudo=True,
+            cwd=self.node.working_path,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Error while installing Dotnet 3.1",
+        )
+        return True
```
squirrelsc: call `_check_exists` here to make sure the dotnet exists in the path.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -69,6 +112,20 @@ def _copy_to_node(self, node: Node, filename: str) -> None:
         if not node.shell.exists(node.working_path / filename):
             node.shell.copy(file_path, node.working_path / filename)
 
+    def _dotnet_install_and_compile(self, node: Node, version: str) -> None:
```
squirrelsc: Pass the class type as parameter, it will be simpler.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -81,7 +138,8 @@ def _execute_docker_test(
         docker_tool = self._verify_and_remove_containers(
             node, docker_image_name, docker_container_name
         )
-        self._copy_to_node(node, prog_src)
+        if not prog_src == "":
```
squirrelsc: 1. `if prog_src != ""` is clearer.
2. What if it's empty? Should the test case fail?

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,109 @@
...
+        self.node.shell.remove(path / app_folder, True)
+
+        self.node.execute(
+            f"dotnet new console -o {app_folder} -n helloworld",
+            cwd=path,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Failed to compile dotnet application",
+        )
+
+    # likely not necessary since we're now using the script to install dotnet on node
```
squirrelsc: Yes, it's not necessary now. Please remove it.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -69,6 +112,20 @@ def _copy_to_node(self, node: Node, filename: str) -> None:
         if not node.shell.exists(node.working_path / filename):
             node.shell.copy(file_path, node.working_path / filename)
 
+    def _dotnet_install_and_compile(self, node: Node, version: str) -> None:
+        dotnet_tool: Dotnet
+        if version == "3.1":
+            dotnet_tool = node.tools[Dotnet31]
```
squirrelsc: BTW, please double check if both 3.1 and 5.0 installed on the same node. Which is actually used? Do we really tested both dotnet version?

lisa/tools/dotnet.py
isaiahduncan: Do you want me to override check_exists for this? If so, the script will add dotnet to the path but it will be gone from the path by the time check_exists gets run.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: If it's empty then that means we're doing a dotnet test and we won't need to copy over a src file

lisa/tools/dotnet.py
squirrelsc: no, it doesn't need to overwrite it. Just call _check_exists like other tools.

microsoft/testsuites/docker/dockersuite.py
squirrelsc: Please try `node.get_case_working_path` now. It's merged. You need to pass in the `case_name` from test case method.

microsoft/testsuites/docker/TestScripts/dotnet50.Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM mcr.microsoft.com/dotnet/core/sdk:5.0
```
isaiahduncan: 3.1 works fine but I just looked in the docker hub and realized the 5.0 test has been failing because there is no docker image for 5.0. What should we do about this?

lisa/tools/dotnet.py
isaiahduncan: Done

microsoft/testsuites/docker/dockersuite.py

microsoft/testsuites/docker/dockersuite.py
squirrelsc: It's ok, but the logic can be simpler like `if prog_src != ""`

microsoft/testsuites/docker/TestScripts/dotnet50.Dockerfile
isaiahduncan: This is where I was looking btw: https://hub.docker.com/_/microsoft-azure-functions-dotnet
Should I just create my own container?

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: It doesn't seem like they both get installed. When I run for both tests, I only see output of "dotnet --info" from the 3.1 test. I will send you the logs in MS teams

lisa/tools/dotnet.py
isaiahduncan: Done

microsoft/testsuites/docker/dockersuite.py

microsoft/testsuites/docker/TestScripts/dotnet50.Dockerfile
squirrelsc: Below is the right place for dotnet runtime dockers. What you shared is for Azure functions.
https://dotnet.microsoft.com/download

microsoft/testsuites/docker/dockersuite.py
squirrelsc: Could you try if the install script can be used to uninstall previous one? I think there may be another way. You can use the dotnet sdk (not runtime only) docker image. create and build the project in the docker file. So it doesn't need to install dotnet in the VM.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -81,7 +132,8 @@ def _execute_docker_test(
         docker_tool = self._verify_and_remove_containers(
             node, docker_image_name, docker_container_name
         )
-        self._copy_to_node(node, prog_src)
+        if prog_src != "":
```
squirrelsc: Sorry, my bad. It can be simpler like `if prog_src:`. It's pythonic expression.

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,80 @@
...
+    def _install(self) -> bool:
+        self._get_install_script()
+        self.node.execute(
+            f"./dotnet-install.sh --version {self._version}",
+            cwd=self.node.working_path,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Error while installing "
+            f"Dotnet {self._version}",
+        ),
+        print(self.run("--info"))
```
squirrelsc: remove the test code.

microsoft/testsuites/docker/dockersuite.py

lisa/tools/dotnet.py
isaiahduncan: Done

microsoft/testsuites/docker/TestScripts/dotnet50.Dockerfile
isaiahduncan: Got it

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: It doesn't look like the script provides the capability to uninstall but it looks like a simple task to uninstall dotnet sdk according to this link https://docs.microsoft.com/en-us/dotnet/core/install/remove-runtime-sdk-versions?pivots=os-linux

lisa/tools/dotnet.py
```diff
@@ -0,0 +1,79 @@
+# Copyright (c) Microsoft Corporation.
```
squirrelsc: Because the dotnet tool is not useful now. It's safe to remove it.

isaiahduncan: Done

squirrelsc (approved):

PR #1581, opened 2021-10-19
Enable multiple poll-mode-drivers for DPDK testing

mcgov:
DPDK testing in v2 checks multiple poll mode driver configurations, this PR adds the ability to run previous tests (verify build and verify send-receive) with the default failsafe PMD and by using the netvsc pmd directly.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -3,12 +3,14 @@
 
 import ipaddress
 import itertools
+import os
 import re
 from typing import Any, Dict, List, Tuple
 
 from assertpy import assert_that, fail
 
 from lisa import Node
+from lisa.tools import Echo
 from lisa.util import InitializableMixin, LisaException
 
 
@@ -31,6 +33,8 @@ class NicInfo:
         self.pci_slot = pci_slot
         self.ip_addr = ""
         self.mac_addr = ""
+        self.dev_uuid = ""
+        self.bound_driver = "hv_netvsc"  # NOTE: azure default
 
     def __str__(self) -> str:
         return (
@@ -135,11 +139,24 @@ class Nics(InitializableMixin):
     def nic_info_is_present(self, nic_name: str) -> bool:
         return nic_name in self.get_upper_nics() or nic_name in self.get_lower_nics()
 
+    def unbind(self, nic: NicInfo, driver_module: str) -> None:
+        echo = self._node.tools[Echo]
+        echo.write_to_file(
+            nic.dev_uuid, f"/sys/bus/vmbus/drivers/{driver_module}/unbind", sudo=True
+        )
+
+    def bind(self, nic: NicInfo, driver_module: str) -> None:
+        echo = self._node.tools[Echo]
+        echo.write_to_file(
+            nic.dev_uuid, f"/sys/bus/vmbus/drivers/{driver_module}/bind", sudo=True
+        )
+
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._nic_names = self._get_nic_names()
         self._get_node_nic_info()
         self.default_nic = self._get_default_nic()
         self._get_host_if_info()
+        self._get_nic_uuids()
 
     def _get_nic_names(self) -> List[str]:
         # identify all of the nics on the device, excluding tunnels and loopbacks etc.
@@ -164,6 +181,21 @@ class Nics(InitializableMixin):
         base_device_result.assert_exit_code()
         # todo check addr matches expectation
         return base_device_result.stdout
+
+    def _get_nic_uuid(self, nic_name: str) -> str:
+        full_dev_path = self._node.execute(
+            f"readlink /sys/class/net/{nic_name}/device",
+            expected_exit_code_failure_message=(
+                f"could not get sysfs device info for {nic_name}"
+            ),
+        )
+        uuid = os.path.basename(full_dev_path.stdout.strip())
+        self._node.log.debug(f"{nic_name} UUID:{uuid}")
+        return uuid
+
+    def _get_nic_uuids(self) -> None:
+        for nic in self.get_upper_nics():
+            self._nics[nic].dev_uuid = self._get_nic_uuid(nic)
 
     def _get_node_nic_info(self) -> None:
         # Identify which nics are slaved to master devices.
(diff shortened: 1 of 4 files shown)


lisa/nic.py
```diff
@@ -165,6 +174,21 @@ def _get_nic_device(self, nic_name: str) -> str:
...
 
+    def _get_nic_uuid(self, nic_name: str) -> str:
+        full_dev_path = self._node.execute(
+            f"readlink /sys/class/net/{nic_name}/device",
+            expected_exit_code_failure_message=(
+                f"could not get sysfs device info for {nic_name}"
+            ),
+        )
+        uuid = os.path.basename(full_dev_path.stdout.strip())
+        self._node.log.info(f"{nic_name} UUID:{uuid}")
```
squirrelsc: Avoid info level log in the underlying package, unless it tells a story in the info level log.

lisa/nic.py
```diff
@@ -247,3 +271,56 @@ def _get_host_if_info(self) -> None:
...
+            nic.dev_uuid, f"/sys/bus/vmbus/drivers/{driver_module}/unbind", sudo=True
+        )
+
+    def _bind(self, nic: NicInfo, driver_module: str) -> None:
+        echo = self._node.tools[Echo]
+        echo.write_to_file(
+            nic.dev_uuid, f"/sys/bus/vmbus/drivers/{driver_module}/bind", sudo=True
+        )
+
+    def bind_to_pmd(self, nic: NicInfo, pmd: str) -> None:
```
squirrelsc: Is this related to dpdk? If so, it should be in test case or dpdk tools.

mcgov: The PMD nomenclature is specific to the dpdk tools, I can move this part into the test suite itself.

squirrelsc: the `_bind/_unbind` can be public, if they need to be accessed outside.

lisa/nic.py
```diff
@@ -247,3 +271,56 @@ def _get_host_if_info(self) -> None:
...
+
+    def _unbind(self, nic: NicInfo, driver_module: str) -> None:
+        echo = self._node.tools[Echo]
+        echo.write_to_file(
+            nic.dev_uuid, f"/sys/bus/vmbus/drivers/{driver_module}/unbind", sudo=True
+        )
+
+    def _bind(self, nic: NicInfo, driver_module: str) -> None:
+        echo = self._node.tools[Echo]
+        echo.write_to_file(
```
squirrelsc: If it needs to bind/unbind multiple nics, can this method support?

lisa/nic.py
```diff
@@ -247,3 +271,56 @@ def _get_host_if_info(self) -> None:
                 self.get_nic(nic).mac_addr = mac_addr
             else:
                 fail(f"Could not parse output of ip addr show {nic}")
+
+    def enable_uio_hv_generic_for_nic(self, nic: NicInfo) -> None:
+        # using netvsc pmd directly for dpdk on hv counterintuitively requires
```
squirrelsc: move to dpdk.

lisa/nic.py
```diff
@@ -89,6 +93,10 @@ class Nics(InitializableMixin):
         r"([a-zA-Z0-9]+)"  # capture device
     )
 
+    # hv_uio_generic driver uuid, a constant value used by vmbus.
+    # https://doc.dpdk.org/guides/nics/netvsc.html#installation
+    HV_UIO_GENERIC_UUID = "f8615163-df3e-46c5-913f-f2d2f965ed0e"
```
squirrelsc: move to dpdk.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -197,20 +197,33 @@ def _install_dependencies(self) -> bool:
...
             )
         )
-        return (
-            f'--vdev="net_vdev_netvsc{vdev_id},iface={node_nic.upper}"'
-            f' --allow "{node_nic.pci_slot}"'
-        )
+        vdev_info = ""
+        if pmd == "failsafe":
+            vdev_info = f'--vdev="net_vdev_netvsc{vdev_id},iface={node_nic.upper}"'
+        elif pmd == "netvsc":
```
squirrelsc: How about else?

lisa/nic.py
mcgov: It binds a single one at a time but we could implement a bind_multiple version if we ever need it. I don't need it right now though.

lisa/nic.py
mcgov: It's nice to have the data labeled with what it is, I'll move it to debug level

lisa/nic.py
squirrelsc: It overwrites the file every time. Even the nics are enabled one by one, the last one will overwrite previous ones.

mcgov: I don't think it works like that. The 'bind' file is not a real file, it's just created for communication with the driver, so writing a new device ID into the file shouldn't unbind the other devices bound to the driver.

squirrelsc: Thank you for explaining, it's ok to me now.

lisa/nic.py
```diff
@@ -247,3 +267,15 @@ def _get_host_if_info(self) -> None:
                 self.get_nic(nic).mac_addr = mac_addr
             else:
                 fail(f"Could not parse output of ip addr show {nic}")
+
+    def unbind(self, nic: NicInfo, driver_module: str) -> None:
```
squirrelsc: move public methods above the private methods.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -197,20 +197,40 @@ def _install_dependencies(self) -> bool:
...
         )
-        return (
-            f'--vdev="net_vdev_netvsc{vdev_id},iface={node_nic.upper}"'
-            f' --allow "{node_nic.pci_slot}"'
-        )
+        vdev_info = ""
+        if node_nic.bound_driver == "hv_netvsc":
+            vdev_info = f'--vdev="net_vdev_netvsc{vdev_id},iface={node_nic.upper}"'
+        elif node_nic.bound_driver == "uio_hv_generic":
+            pass
```
squirrelsc, line 216: It looks the `...` is the new fashion.

mcgov: I think `pass` is more readable. It's a well-known keyword and it shows up highlighted since it's a reserved keyword.

squirrelsc (approved):

PR #1587, opened 2021-10-21
Add kvp_client_arm64 for lisav2 kvp and hyper-V tests

lubaihua33:
Add some enhancement for lisav2 Hyper-V test cases for ARM64


squirrelsc:
KVP test case was migrated and merged to V3. Please double check it. If it works well, it doesn't need to change the v2 test case.

lubaihua33:
> KVP test case was migrated and merged to V3. Please double check it. If it works well, it doesn't need to change the v2 test case.

Lisav2 test cases are still used currently. We'd better add these fixes. They also contain Hyper-V test cases enhancement.

squirrelsc (approved):

PR #1589, opened 2021-10-22
Download failed and check.log of xfstests.

LiliDeng:
image

### DIFF
--- a/lisa/tools/parted.py
+++ b/lisa/tools/parted.py
@@ -22,13 +22,19 @@ class Parted(Tool):
         self, disk_name: str, part_type: str, start: str, end: str
     ) -> None:
         cmd_result = self.run(
-            f"-s -- {disk_name} mkpart {part_type} {start} {end}", shell=True, sudo=True
+            f"-s -- {disk_name} mkpart {part_type} {start} {end}",
+            shell=True,
+            sudo=True,
+            force_run=True,
         )
         cmd_result.assert_exit_code()
 
     def make_label(self, disk_name: str, type: str = "gpt") -> None:
         cmd_result = self.run(
-            f"-s -- {disk_name} mklabel {type}", shell=True, sudo=True
+            f"-s -- {disk_name} mklabel {type}",
+            shell=True,
+            sudo=True,
+            force_run=True,
         )
         cmd_result.assert_exit_code()
 
--- a/lisa/tools/xfstests.py
+++ b/lisa/tools/xfstests.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 import re
-from pathlib import PurePath
+from pathlib import Path, PurePath
 from typing import List, Type, cast
 
 from lisa.executable import Tool
@@ -181,9 +181,9 @@ class Xfstests(Tool):
             echo = self.node.tools[Echo]
             echo.write_to_file(exclude_tests, exclude_file_path)
 
-    def check_test_results(self, raw_message: str) -> None:
+    def check_test_results(self, raw_message: str, log_path: Path) -> None:
         xfstests_path = self.get_xfstests_path()
-        results_path = xfstests_path.joinpath("results/check.log")
+        results_path = xfstests_path / "results/check.log"
         if not self.node.shell.exists(results_path):
             raise LisaException(
                 f"Result path {results_path} doesn't exist, please check testing runs"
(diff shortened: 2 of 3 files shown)


microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -267,6 +292,7 @@ def xfstesting_btrfs_nvme_datadisk_validation(self, node: Node) -> None:
     def _execute_xfstests(
         self,
         node: Node,
+        saved_path: Path,
```
squirrelsc: `log_path` is better

lisa/tools/xfstests.py
```diff
@@ -210,7 +210,31 @@ def check_test_results(self, raw_message: str) -> None:
...
         raise LisaException(
             f"Fail {fail_count} cases of total {total_count}, fail cases"
             f" {fail_cases}, details {fail_info}, please investigate."
         )
+
+    def save_xfstests_log(self, fail_cases: List[str], saved_path: Path) -> None:
+        xfstests_path = self.get_xfstests_path()
+        self.node.shell.copy_back(
+            xfstests_path.joinpath("results/check.log"),
+            saved_path.joinpath("xfstests/check.log"),
```
squirrelsc: a simple way to join path is like below,
`saved_path / "xfstests/check.log"`

lisa/tools/xfstests.py
```diff
@@ -210,7 +210,31 @@ def check_test_results(self, raw_message: str) -> None:
...
+        )
+        for fail_case in fail_cases:
+            if self.node.shell.exists(
+                xfstests_path.joinpath(f"results/{fail_case}.out.bad")
+            ):
+                result_path = xfstests_path.joinpath(f"results/{fail_case}.out.bad")
+                file_name = f"xfstests/{fail_case}.out.bad"
+            elif self.node.shell.exists(
```
squirrelsc: why need elif? can we copy both bad and full?

LiliDeng: for most of failure cases, it only has .out.bad, for a few of them, it doesn't have out.bad but has .full

squirrelsc: yes, what ever a failed case have, we can copy both of them. In current logic, if the bad exists, it doesn't copy the full.

lisa/tools/parted.py
```diff
@@ -19,16 +19,22 @@ def can_install(self) -> bool:
         return True
 
     def make_partition(
-        self, disk_name: str, part_type: str, start: str, end: str
+        self, disk_name: str, part_type: str, start: str, end: str, force_run: bool
     ) -> None:
         cmd_result = self.run(
-            f"-s -- {disk_name} mkpart {part_type} {start} {end}", shell=True, sudo=True
+            f"-s -- {disk_name} mkpart {part_type} {start} {end}",
```
squirrelsc, line 25: Because they are actions, so they are expected to force_run every time. So it doesn't need to be an argument, just add force run in the command directly.

lisa/tools/xfstests.py
```diff
@@ -210,7 +210,24 @@ def check_test_results(self, raw_message: str) -> None:
...
+        )
+        for fail_case in fail_cases:
+            file_name = f"xfstests/{fail_case}.out.bad"
+            result_path = xfstests_path / file_name
+            if self.node.shell.exists(result_path):
+                self.node.shell.copy_back(result_path, log_path / file_name)
+            file_name = f"xfstests/{fail_case}.full"
+            result_path = xfstests_path / file_name
+            if self.node.shell.exists(result_path):
+                self.node.shell.copy_back(result_path, log_path / file_name)
```
squirrelsc, line 235: It's better to log a line, if there is no full and bad.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -35,9 +36,9 @@ def _configure_disk(
         node.execute(f"rm -r {mount_point}", sudo=True)
         mount.umount(disk, mount_point)
 
-    parted.make_label(disk_name)
-    parted.make_partition(disk_name, "primary", "1", "50%")
-    parted.make_partition(disk_name, "secondary", "50%", "100%")
+    parted.make_label(disk_name, force_run=True)
+    parted.make_partition(disk_name, "primary", "1", "50%", force_run=True)
+    parted.make_partition(disk_name, "secondary", "50%", "100%", force_run=True)
```
squirrelsc: This change should be in previous commit. They are the same purpose.

lisa/tools/parted.py
```diff
@@ -22,13 +22,19 @@ def make_partition(
...
+            shell=True,
+            sudo=True,
+            force_run=True,
         )
         cmd_result.assert_exit_code()
 
-    def make_label(self, disk_name: str, type: str = "gpt") -> None:
+    def make_label(self, disk_name: str, force_run: bool, type: str = "gpt") -> None:
```
squirrelsc: This is the same...

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: They should be removed.

squirrelsc (approved):

PR #1590, opened 2021-10-22
Make shared rg name is a variable, and some other improvements.

squirrelsc:
see commits for details.

Example in runbook:

```yaml
platform:
  - type: azure
    ...
    azure:
      shared_resource_group_name: customized_resource_group_name
      ...
```

### DIFF
--- a/docs/make.cmd
+++ b/docs/make.cmd
@@ -1,0 +1,35 @@
+@ECHO OFF
+
+pushd %~dp0
+
+REM Command file for Sphinx documentation
+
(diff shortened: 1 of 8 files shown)


LiliDeng:
what is 'shared_resource_group_name'?

lizzha (approved):

squirrelsc:
> what is 'shared_resource_group_name'?

It's asked from Penguinator. There is a shared resource group already. It's better to put all shared things in the same resource group.

PR #1592, opened 2021-10-23
Add verify uuid for root disk testcase

dsrivastavv:
WIP. **Do not merge**. The logic can be reviewed but more changes need to be made in formatting.

Migrate `VerifyUUID` testcase from https://github.com/microsoft/lisa/blob/master/Testscripts/Linux/VERIFY-LINUX-DISK-SETUP.py

### DIFF
--- a/lisa/tools/blkid.py
+++ b/lisa/tools/blkid.py
@@ -1,0 +1,81 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from typing import List
+
+from assertpy.assertpy import assert_that
+
+from lisa.executable import Tool
+from lisa.util import LisaException, get_matched_str
+
+
+class PartitionInfo(object):
+    def __init__(
+        self,
+        name: str,
+        uuid: str,
+        part_uuid: str,
+    ):
+        self.name = name
+        self.uuid = uuid
+        self.part_uuid = part_uuid
+
+    def __repr__(self) -> str:
+        return f"{self.name} {self.uuid} {self.part_uuid}"
+
+    def __str__(self) -> str:
+        return self.__repr__()
+
+
+class Blkid(Tool):
+
+    # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01" # noqa: E501
+    _get_partition_name = re.compile(r"\s*(?P<name>\S+):.*")
+
+    # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01" # noqa: E501
+    _get_partition_uuid = re.compile(r"\s+UUID=\"(?P<uuid>\S+)\"")
+
+    # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01" # noqa: E501
+    _get_partition_part_uuid = re.compile(r"\s+PARTUUID=\"(?P<part_uuid>\S+)\"")
+
+    @property
+    def command(self) -> str:
+        return "blkid"
+
+    def get_partition_information(self) -> List[PartitionInfo]:
+        """
+        Get partition information from blkid output.
+        Sample output :
+        /dev/sda1: LABEL="cloudimg-rootfs" UUID="b1983cef-43a3-46ac-a083-b5e06a61c9fd" TYPE="ext4" PARTUUID="6b003b9b-0531-41bb-ab5e-b2491580c31f" # noqa: E501
+        /dev/sda15: LABEL_FATBOOT="UEFI" LABEL="UEFI" UUID="0BC7-08EF" TYPE="vfat" PARTUUID="d80ae19a-00f8-4cae-a95e-9bbb761b7e9a" # noqa: E501
+        /dev/sdb1: UUID="a88b3ef7-ddc8-4942-8931-253a5f21cae1" TYPE="ext4" PARTUUID="c7c91a5e-01" # noqa: E501
+        /dev/loop0: TYPE="squashfs"
+        """
+        output = self.run(sudo=True).stdout
+        partition_info: List[PartitionInfo] = []
+        for line in output.splitlines():
+            # get partition name
+            name = get_matched_str(line, self._get_partition_name)
+            assert_that(name, "partition name should not be none.").is_not_none()
+
+            # get partion uuid. This could be None.
+            uuid = get_matched_str(line, self._get_partition_uuid)
+
+            # get partion part_uuid. This could be None.
+            part_uuid = get_matched_str(line, self._get_partition_part_uuid)
+
+            partition_info.append(PartitionInfo(name, uuid, part_uuid))
+
+        return partition_info
+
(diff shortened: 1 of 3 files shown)


lisa/tools/blkid.py
```diff
@@ -0,0 +1,82 @@
...
+class Blkid(Tool):
+
+    # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01"
+    _get_partition_name = re.compile(r"\s*(?P<name>\S+):.*")
+
+    # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01"
+    _get_partition_uuid = re.compile(r"\s+UUID=\"(?P<uuid>\S+)\"")
+
+    # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01"
+    _get_partition_part_uuid = re.compile(r"\s+PARTUUID=\"(?P<part_uuid>\S+)\"")
```
squirrelsc, line 40: The three values can be get from the one pattern with named groups.

squirrelsc: The `get_matched_str` is used in lightweight one string capture. If it needs multiple strings, don't use it.

lisa/tools/blkid.py
```diff
@@ -0,0 +1,82 @@
...
+            uuid = get_matched_str(line, self._get_partition_uuid)
+
+            # get partion part_uuid. This could be None.
+            part_uuid = get_matched_str(line, self._get_partition_part_uuid)
+
+            partition_info.append(PartitionInfo(name, uuid, part_uuid))
+
+        return partition_info
+
+    def get_partition_info_for_partition_name(
```
squirrelsc: `get_partition_info` is enough, the parameter hints it's from partition name.

lisa/tools/blkid.py
```diff
@@ -0,0 +1,82 @@
...
+        self, partition_name: str
+    ) -> PartitionInfo:
+        """
+        Get partition information for partition name.
+        """
+        partition_info = self.get_partition_information()
+        for partition in partition_info:
+            if partition.name == partition_name:
+                return partition
+        return None
```
squirrelsc: Should it raise an exception, if it's not found? So we can find more potential bugs.

microsoft/testsuites/core/storage.py
```diff
@@ -161,6 +161,54 @@ def verify_resource_disk_io(self, log: Logger, node: RemoteNode) -> None:
...
 
+    @TestCaseMetadata(
+        description="""
+        This test will verify that identifier of root partition matches 
+        from different sources.
+
+        Steps:
+        1. Get the partition identifier from `blkid` command.
+        2. Verify that the partition identifier from `blkid` is present in dmesg.
+        3. Verify that the partition identifier from `blkid` is present in \etc\\fstab.
```
squirrelsc: `\etc\fstab` -> `/etc/fstab`

microsoft/testsuites/core/storage.py
```diff
@@ -161,6 +161,54 @@ def verify_resource_disk_io(self, log: Logger, node: RemoteNode) -> None:
...
+        """,
+        priority=1,
+    )
+    def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
+        # get informtion of root disk from blkid
+        os_disk_mount_point = "/"
```
squirrelsc: It's only used once, and never be changed. It doesn't need to be a variable.

microsoft/testsuites/core/storage.py
```diff
@@ -161,6 +161,54 @@ def verify_resource_disk_io(self, log: Logger, node: RemoteNode) -> None:
...
+        from different sources.
+
+        Steps:
+        1. Get the partition identifier from `blkid` command.
+        2. Verify that the partition identifier from `blkid` is present in dmesg.
+        3. Verify that the partition identifier from `blkid` is present in \etc\\fstab.
+        """,
+        priority=1,
+    )
+    def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
```
squirrelsc, line 183: Is it a requirement to Azure? If so, limit it to Azure and Ready.

microsoft/testsuites/core/storage.py
```diff
@@ -161,6 +161,54 @@ def verify_resource_disk_io(self, log: Logger, node: RemoteNode) -> None:
...
+        os_disk_mount_point = "/"
+        os_partition = (
+            node.features[Disk].get_partition_with_mount_point(os_disk_mount_point).name
+        )
+        os_partition_info = node.tools[Blkid].get_partition_info_for_partition_name(
+            os_partition
+        )
+
+        # verify that root=<name> or root=uuid=<uuid> or root=partuuid=<part_uuid> is present in dmesg
+        dmesg = node.tools[Dmesg].run().stdout.lower()
```
squirrelsc: It doesn't need lower here. It changes the original value and may be confusing. Use the raw value should be ok.

lisa/tools/blkid.py
dsrivastavv: We cannot merge them into single group because the order of UUID and PartUUID is not always the same. It is also possible that UUID is not present. So we need to extract them from the string seperately.

lisa/tools/blkid.py
dsrivastavv: There is already a function named `get_partition_information` in this file. The intent of this function to extract the entry matching a specific name.

microsoft/testsuites/core/storage.py
dsrivastavv: This should be azure specific. I will update the metadata to reflect it.

lisa/tools/blkid.py
squirrelsc: `get_partition_info_by_name`?

microsoft/testsuites/core/storage.py
```diff
@@ -161,6 +165,81 @@ def verify_resource_disk_io(self, log: Logger, node: RemoteNode) -> None:
...
+        ):
+            raise LisaException(
+                f"One of root={os_partition_info.name} or "
+                f"root=UUID={os_partition_info.uuid} or "
+                f"root=PARTUUID={os_partition_info.part_uuid} "
+                "should be present in dmesg output"
+            )
+
+        # verify that "<uuid> /" or "<name> /"or "<part_uuid> /" present in /etc/fstab
+        fstab = node.tools[Cat].run("/etc/fstab", sudo=True).stdout.lower()
```
squirrelsc: remove the lower.

dsrivastavv: I saw that some distro had the UUID written in lowercase though I haven't been able to repro it again. I will remove the lower() and we can look into the runs if such a distro is present.

microsoft/testsuites/core/storage.py
dsrivastavv: I had to add `sudo` to dmesg run as it was failing on suse sles distro

microsoft/testsuites/core/storage.py
squirrelsc: If it needs to compare with different cases, it's better to compare with ignore case, instead of save the lower version of string.

squirrelsc (approved):

PR #1593, opened 2021-10-23
Bugfixes: Ubuntu 20.04 missing pkg-config, NIC logic fix

mcgov:
Fixing a few bugs:
- Ubuntu 20.04 needs pkg-config installed, add the package to the list of 20.04 packages.
- Fixing a bug where synthetic interfaces could end up in the `get_upper_nics` list.
- Get rid of 'has_lower' logic error (we can just check 'lower' since "" will resolve to false)

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -27,7 +27,6 @@ class NicInfo:
         lower: str = "",
         pci_slot: str = "",
     ) -> None:
-        self.has_lower = lower != "" or pci_slot == ""
         self.upper = upper
         self.lower = lower
         self.pci_slot = pci_slot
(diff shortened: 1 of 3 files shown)


lisa/nic.py
```diff
@@ -114,16 +113,16 @@ def is_empty(self) -> bool:
         return len(self._nics) == 0
 
     def get_unpaired_devices(self) -> List[str]:
-        return [x.upper for x in self._nics.values() if not x.has_lower]
+        return [x.upper for x in self._nics.values() if not x.lower]
 
     def get_upper_nics(self) -> List[str]:
-        return [self._nics[x].upper for x in self._nics.keys()]
+        return [x.upper for x in self._nics.values() if x.lower]
```
squirrelsc, line 119: should it be `x.upper`?

squirrelsc:
It's better that each bug has a commit.

lisa/nic.py
mcgov: Nope, this method should only return nics which have a lower paired with them.

squirrelsc: But the method name is `get_upper_nics`. It's hard to know, the synthetic only nics are not returned. 

BTW, the upper/lower pair is hard to me to map to master/slave in current nic names. I know master/slave are not good terms. How about we use origin/sriov directly? It looks also usable on bare mental and other nics.

mcgov: This will need to get refactored for a later test, but for now this small logic fix is enough for me to make progress. I'm thinking that later we will end up with a SyntheticNicInfo and SriovNicInfo subclass of a more-generic NicInfo parent or something like that.

squirrelsc (approved):

PR #1594, opened 2021-10-25
LISAv3: Adding docker-compose wordpress test to docker test suite

isaiahduncan:
Creating docker-compose tool
Additions and changes to the docker tool
Creating docker compose wordpress test for docker suite

### DIFF
--- a/lisa/tools/docker_compose.py
+++ b/lisa/tools/docker_compose.py
@@ -1,0 +1,52 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import PurePath
+
+from lisa.base_tools import Uname, Wget
+from lisa.executable import Tool
+from lisa.operating_system import Posix, Redhat
+from lisa.tools.service import Service
+from lisa.util import LisaException
+
+
+class DockerCompose(Tool):
+    @property
+    def command(self) -> str:
+        return "docker-compose"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def start(self) -> None:
+        self._log.debug("Start docker compose")
+        service = self.node.tools[Service]
+        service.enable_service("docker-compose")
+        service.restart_service("docker-compose")
+
+    def up(self, path: PurePath) -> None:
+        self.run("up -d", sudo=True, cwd=path)
+
+    def _install(self) -> bool:
+        # The default installed docker-compose package doesn't work for
+        # redhat so it uses the latest version
+        if isinstance(self.node.os, Redhat):
+            wget_tool = self.node.tools[Wget]
+            uname_tool = self.node.tools[Uname]
+            hardware = uname_tool.get_linux_information().hardware_platform
+            filename = "docker-compose"
+            wget_tool.run(
+                "https://github.com/docker/compose/releases/download/1.23.2"
+                f"/docker-compose-Linux-{hardware} -O {filename}",
+                sudo=True,
+            )
+            self.node.execute(f"sudo chmod +x {filename}")
+            self.node.execute(
+                "mv docker-compose /usr/bin/", sudo=True, expected_exit_code=0
+            )
+        elif isinstance(self.node.os, Posix):
+            self.node.os.install_packages("docker-compose")
+        else:
+            raise LisaException(f"Not supported on {self.node.os.information.vendor}")
+        return self._check_exists()
--- a/microsoft/testsuites/docker/dockersuite.py
+++ b/microsoft/testsuites/docker/dockersuite.py
@@ -8,7 +8,7 @@ from assertpy import assert_that
 
 from lisa import Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
 from lisa.operating_system import Redhat
-from lisa.tools import Docker
+from lisa.tools import Docker, DockerCompose
 from lisa.util import SkippedException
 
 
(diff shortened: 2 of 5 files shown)


lisa/tools/docker_compose.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def up(self, path: PurePath) -> None:
+        self.run("up -d", sudo=True, cwd=path)
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Posix):
+            self.node.os.install_packages("docker-compose")
+        else:
+            raise LisaException(f"Not supported on {self.node.os.information.vendor}")
+        return True
```
squirrelsc: `return self._check_exists()` It can check if the installed package are in path automatically, and will raise exception, if it's not.

lisa/tools/docker_compose.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def up(self, path: PurePath) -> None:
+        self.run("up -d", sudo=True, cwd=path)
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Posix):
+            self.node.os.install_packages("docker-compose")
+        else:
+            raise LisaException(f"Not supported on {self.node.os.information.vendor}")
+        return self._check_exists()
```
isaiahduncan, line 52: This line causes redhat 7 to fail. I have tried running the test with this line as "return true" and the test worked fine.
Distro name: "redhat rhel 7.8 7.8.2020111309"

squirrelsc: Would you check the detail error message? The docker may not be in the path, after it's installed. We should fix it, instead of making other distros skip this check. One of way of fixing, is to set the absolute path of the command in Redhat 7.8.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -21,44 +21,71 @@
...
+            Steps:
+            1. Install Docker and Docker-Compose on node
+            2. Copy docker-compose.yml into node
+            3. Start docker container with docker-compose
+            4. Run ps in the docker container and capture output
+            5. Check that "apache2" can be found in captured output
+        """,
+        priority=1,
+    )
+    def docker_compose_app(self, node: Node) -> None:
```
squirrelsc: forgot to check the test case names. Please make sure they start with "verify_"

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -21,44 +21,71 @@
...
+            This test case uses docker-compose to create and run a wordpress mysql app
+
+            Steps:
+            1. Install Docker and Docker-Compose on node
+            2. Copy docker-compose.yml into node
+            3. Start docker container with docker-compose
+            4. Run ps in the docker container and capture output
+            5. Check that "apache2" can be found in captured output
+        """,
+        priority=1,
```
squirrelsc: Does this case need to be P1? I think P2 or P3 is enough.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -21,44 +21,71 @@
...
-            3. Copy dotnet dockerfile into node
-            4. Create docker image and run docker container
-            5. Check results of docker run against dotnet string identifier
+            1. Install Docker
+            2. Copy dotnet dockerfile into node
+            3. Create docker image and run docker container
+            4. Check results of docker run against dotnet string identifier
         """,
-        priority=2,
+        priority=3,
```
squirrelsc: This in more important than wordpress. Make this docker test case to P1, and other to P2 or P3.

isaiahduncan: should the dotnet 50 test case also be P1?

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: changed to P3

microsoft/testsuites/docker/dockersuite.py

lisa/tools/docker_compose.py
isaiahduncan: I was incorrect. the docker compose install actually failed on redhat. The problem is not with check_exists

microsoft/testsuites/docker/dockersuite.py
squirrelsc: One P1 test case is enough to address most issues with docker compatibility itself and dotnet31. The reason to use dotnet31 as the P1, because it's ms product and it's the current long term support version. The dotnet50 case addresses compatibility issue, which is specified to dotnet50. If it's special to dotnet50, its impact is much lower than docker and dotnet31. So it should be P2 or P3.

lisa/tools/docker_compose.py
squirrelsc: Yes, the check_exists is the final gate to check if it's installed correctly.

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: done

microsoft/testsuites/docker/dockersuite.py
squirrelsc: The name is not changed yet?

isaiahduncan: It hadn't been pushed yet. It is now pushed

lisa/tools/docker_compose.py
```diff
@@ -0,0 +1,50 @@
...
+        service.restart_service("docker-compose")
+
+    def up(self, path: PurePath) -> None:
+        self.run("up -d", sudo=True, cwd=path)
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Redhat):
+            wget_tool = self.node.tools[Wget]
+            uname_s = self.node.execute("uname -s")
+            uname_m = self.node.execute("uname -m")
```
squirrelsc: The uname tool already parsed those information. Please reuse it.

lisa/tools/docker_compose.py
```diff
@@ -0,0 +1,50 @@
...
+        service = self.node.tools[Service]
+        service.enable_service("docker-compose")
+        service.restart_service("docker-compose")
+
+    def up(self, path: PurePath) -> None:
+        self.run("up -d", sudo=True, cwd=path)
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Redhat):
+            wget_tool = self.node.tools[Wget]
```
squirrelsc, line 35: Add comments to explain our effort that, the default installed packages cannot work, so it uses the latest docker version.

microsoft/testsuites/docker/dockersuite.py
```diff
@@ -73,7 +100,7 @@ def docker_dotnet50_app(self, node: Node) -> None:
     """,
         priority=1,
     )
-    def docker_java_app(self, node: Node) -> None:
+    def verify_docker_java_app(self, node: Node) -> None:
```
squirrelsc, line 103: This test case can be P2. Let's keep dotnet31 is the only P1 test case. Ideally, every area should have one P1 test case, unless they have very different coverage.

lisa/tools/docker_compose.py
isaiahduncan: Done

microsoft/testsuites/docker/dockersuite.py
isaiahduncan: Done

lisa/tools/docker_compose.py
isaiahduncan: Done

squirrelsc (approved):

PR #1595, opened 2021-10-26
Refine sriov test cases, migrate reload module test.

LiliDeng:
Previous use the nic name to do the testing, but found during testing that the nic names are not paired with the same ip range, it caused case fail, now fix it by get the dest ip by ip range.

### DIFF
--- a/lisa/tools/firewall.py
+++ b/lisa/tools/firewall.py
@@ -1,0 +1,91 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+
+from .service import Service
+
+
+class Firewall(Tool):
+    @property
+    def command(self) -> str:
+        return "ls -lt"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def stop(self) -> None:
+        cmd_result = self.node.execute("command -v ufw")
+        if 0 == cmd_result.exit_code:
+            ufw = self.node.tools[Ufw]
+            ufw.stop()
+        cmd_result = self.node.execute("command -v iptables")
+        if 0 == cmd_result.exit_code:
+            iptables = self.node.tools[Iptables]
+            iptables.stop()
+        cmd_result = self.node.execute("command -v SuSEfirewall2")
+        if 0 == cmd_result.exit_code:
+            susefirewall2 = self.node.tools[SuSEfirewall2]
+            susefirewall2.stop()
+        cmd_result = self.node.execute("command -v firewall-cmd")
+        if 0 == cmd_result.exit_code:
+            firewalld = self.node.tools[Firewalld]
+            firewalld.stop()
+
+
+class Ufw(Tool):
+    @property
+    def command(self) -> str:
+        return "ufw"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def stop(self) -> None:
+        self.run("disable", shell=True)
+
+
+class Iptables(Tool):
+    @property
+    def command(self) -> str:
+        return "iptables"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def stop(self) -> None:
+        self.run("-P INPUT ACCEPT", shell=True)
+        self.run("-P OUTPUT ACCEPT", shell=True)
+        self.run("-P FORWARD ACCEPT", shell=True)
+        self.run("-P -F", shell=True)
+
+
+class SuSEfirewall2(Tool):
+    @property
+    def command(self) -> str:
+        return "SuSEfirewall2"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def stop(self) -> None:
+        service = self.node.tools[Service]
+        service.stop_service("SuSEfirewall2")
+
+
+class Firewalld(Tool):
+    @property
+    def command(self) -> str:
+        return "firewall-cmd"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def stop(self) -> None:
+        service = self.node.tools[Service]
+        service.stop_service("firewalld")
--- a/lisa/tools/iperf3.py
+++ b/lisa/tools/iperf3.py
@@ -1,0 +1,88 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from typing import List, Type, cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.util.process import Process
+
+from .git import Git
+from .make import Make
+
+
+class Iperf3(Tool):
+    repo = "https://github.com/esnet/iperf"
+    branch = "3.10.1"
+
+    @property
+    def command(self) -> str:
+        return "iperf3"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git, Make]
+
+    def _install_from_src(self) -> None:
+        tool_path = self.get_tool_path()
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
+        code_path = tool_path.joinpath("iperf")
+        make = self.node.tools[Make]
+        self.node.execute("./configure", cwd=code_path).assert_exit_code()
+        make.make_install(code_path)
+        self.node.execute("ldconfig", sudo=True, cwd=code_path).assert_exit_code()
+        self.node.execute(
+            "ln -s /usr/local/bin/iperf3 /usr/bin/iperf3", sudo=True, cwd=code_path
+        ).assert_exit_code()
+
+    def install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("iperf3")
+        if not self._check_exists():
+            self._install_from_src()
+        return self._check_exists()
+
+    def run_as_server(self, daemon: bool = True) -> None:
+        # -s: run iperf3 as server mode
+        # -D: run iperf3 as a daemon
+        cmd = " -s"
+        if daemon:
+            cmd += " -D"
+        self.run(
+            cmd,
+            force_run=True,
+            sudo=True,
+            shell=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"fail to launch cmd {self.command}"
+            f"{cmd}",
+        )
+
+    def run_as_client_async(
+        self, server_ip: str, log_file: str = "", seconds: int = 10
+    ) -> Process:
+        # -c: run iperf3 as client mode, followed by iperf3 server ip address
+        # -t: run iperf3 testing for given seconds
+        # --logfile: save logs into specified file
+        cmd = f"-t {seconds} -c {server_ip}"
+        if log_file:
+            if self.node.shell.exists(self.node.get_pure_path(log_file)):
+                self.node.shell.remove(self.node.get_pure_path(log_file))
+            cmd += f" --logfile {log_file}"
+        process = self.node.execute_async(f"{self.command} {cmd}", sudo=self._use_sudo)
+        return process
+
+    def run_as_client(
+        self, server_ip: str, log_file: str = "", seconds: int = 10
+    ) -> None:
+        process = self.run_as_client_async(server_ip, log_file, seconds)
+        process.wait_result(
+            seconds,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="fail to lanuch iperf3 client",
+        )
--- a/lisa/tools/kill.py
+++ b/lisa/tools/kill.py
@@ -1,0 +1,32 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+
+from .pgrep import Pgrep
+
+
+class Kill(Tool):
+    @property
+    def command(self) -> str:
+        return "kill"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def by_name(self, process_name: str, signum: int = 9) -> None:
+        running_processes = self.node.tools[Pgrep].get_processes(process_name)
+        for process in running_processes:
+            self.with_signum(process.id, signum)
+
+    def with_signum(self, pid: str, signum: int = 9) -> None:
+        self.run(
+            f"-{signum} {pid}",
+            shell=True,
+            sudo=True,
+            force_run=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="fail to run "
+            f"{self.command} -{signum} {pid}",
+        )
--- a/lisa/tools/lspci.py
+++ b/lisa/tools/lspci.py
@@ -6,7 +6,7 @@ from typing import Any, Dict, List
 from lisa.executable import Tool
 from lisa.operating_system import Posix
 from lisa.tools import Echo
-from lisa.util import LisaException, constants
+from lisa.util import LisaException, constants, get_matched_str
 
 # Example output of lspci command -
 # lspci -m
@@ -35,6 +35,8 @@ DEVICE_TYPE_DICT: Dict[str, str] = {
     constants.DEVICE_TYPE_NVME: "Non-Volatile memory controller",
     constants.DEVICE_TYPE_GPU: "3D controller",
 }
+
+PATTERN_MODULE_IN_USE = re.compile(r"Kernel driver in use: (.*)\r", re.M)
 
 
 class PciDevice:
@@ -95,14 +97,14 @@ class Lspci(Tool):
             self._pci_devices = []
             # Ensure pci device ids and name mappings are updated.
             self.node.execute("update-pciids", sudo=True)
-            result = self.run("-m", force_run=force_run, shell=True)
-            if result.exit_code != 0:
-                result = self.run("-m", force_run=force_run, shell=True, sudo=True)
-                if result.exit_code != 0:
-                    raise LisaException(
-                        f"get unexpected non-zero exit code {result.exit_code} "
-                        f"when run {self.command} -m."
-                    )
+            result = self.run(
+                "-m",
+                force_run=force_run,
+                shell=True,
+                expected_exit_code=0,
+                expected_exit_code_failure_message=f"get unexpected non-zero exit code"
+                f" when run {self.command} -m.",
+            )
             for pci_raw in result.stdout.splitlines():
                 pci_device = PciDevice(pci_raw)
                 self._pci_devices.append(pci_device)
@@ -129,3 +131,16 @@ class Lspci(Tool):
         self.node.tools[Echo].write_to_file(
             "1", self.node.get_pure_path("/sys/bus/pci/rescan"), sudo=True
         )
+
+    def get_used_module(self, slot: str) -> str:
+        result = self.run(
+            f"-nks {slot}",
+            force_run=True,
+            shell=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="get unexpected non-zero exit code,"
+            f"when run {self.command} -nks {slot}.",
+        )
+        matched = get_matched_str(result.stdout, PATTERN_MODULE_IN_USE)
+        assert matched
+        return matched
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -197,7 +197,7 @@ class Process:
                 f"execution time: {self._timer}, exit code: {self._result.exit_code}"
             )
 
-        if expected_exit_code:
+        if expected_exit_code is not None:
             self._result.assert_exit_code(
                 expected_exit_code=expected_exit_code,
                 message=expected_exit_code_failure_message,
--- a/microsoft/testsuites/network/sriov.py
+++ b/microsoft/testsuites/network/sriov.py
@@ -1,28 +1,28 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 from pathlib import Path
-from typing import Dict, cast
+from typing import Any, Dict, List, cast
 
 from assertpy import assert_that
 
 from lisa import (
     Environment,
     Logger,
+    Node,
     RemoteNode,
     TestCaseMetadata,
     TestSuite,
     TestSuiteMetadata,
     schema,
+    search_space,
     simple_requirement,
 )
 from lisa.features import NetworkInterface, SerialConsole
 from lisa.nic import NicInfo, Nics
 from lisa.sut_orchestrator import AZURE, READY
-from lisa.tools import Cat, Lsmod, Lspci, Ssh
-from lisa.util import LisaException, constants
+from lisa.tools import Cat, Ethtool, Firewall, Iperf3, Kill, Lsmod, Lspci, Modprobe, Ssh
+from lisa.util import LisaException, SkippedException, constants
 from lisa.util.shell import wait_tcp_port_ready
-
-vm_nics: Dict[str, Dict[str, NicInfo]] = {}
 
 
 @TestSuiteMetadata(
@@ -34,6 +34,16 @@ vm_nics: Dict[str, Dict[str, NicInfo]] = {}
 )
 class Sriov(TestSuite):
     TIME_OUT = 300
+    # ConnectX-3 uses mlx4_core
+    # mlx4_en and mlx4_ib depends on mlx4_core
+    #  need remove mlx4_en and mlx4_ib firstly
+    #  otherwise will see modules is in used issue
+    # ConnectX-4/ConnectX-5 uses mlx5_core
+    # mlx5_ib depends on mlx5_core, need remove mlx5_ib firstly
+    reload_modules_dict: Dict[str, List[str]] = {
+        "mlx5_core": ["mlx5_ib"],
+        "mlx4_core": ["mlx4_en", "mlx4_ib"],
+    }
 
     @TestCaseMetadata(
         description="""
@@ -41,19 +51,21 @@ class Sriov(TestSuite):
          synthetic nic is paired with one VF.
 
         Steps,
-        1. Check module of sriov network device is loaded.
-        2. Check VF of synthetic nic is paired.
+        1. Check VF of synthetic nic is paired.
+        2. Check module of sriov network device is loaded.
         3. Check VF counts listed from lspci is expected.
         """,
         priority=1,
         requirement=simple_requirement(
+            min_count=2,
             network_interface=schema.NetworkInterfaceOptionSettings(
                 data_path=schema.NetworkDataPath.Sriov,
             ),
         ),
     )
     def sriov_basic_validation(self, environment: Environment) -> None:
-        self._sriov_basic_test(environment)
+        vm_nics = self._initialize_nic_info(environment)
+        self._sriov_basic_test(environment, vm_nics)
 
     @TestCaseMetadata(
         description="""
@@ -62,8 +74,8 @@ class Sriov(TestSuite):
          and tx statistics of dest increase after send 200 Mb file from source to dest.
 
         Steps,
-        1. Check module of sriov network device is loaded.
-        2. Check VF of synthetic nic is paired.
+        1. Check VF of synthetic nic is paired.
+        2. Check module of sriov network device is loaded.
         3. Check VF counts listed from lspci is expected.
         4. Setup SSH connection between source and dest with key authentication.
         5. Ping the dest IP from the source machine to check connectivity.
@@ -79,8 +91,9 @@ class Sriov(TestSuite):
         ),
     )
     def sriov_single_vf_connection_validation(self, environment: Environment) -> None:
-        self._sriov_basic_test(environment)
-        self._sriov_vf_connection_test(environment)
+        vm_nics = self._initialize_nic_info(environment)
+        self._sriov_basic_test(environment, vm_nics)
+        self._sriov_vf_connection_test(environment, vm_nics)
 
     @TestCaseMetadata(
         description="""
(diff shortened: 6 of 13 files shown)


microsoft/testsuites/network/sriov.py
```diff
@@ -356,9 +356,25 @@ def _sriov_vf_connection_test(
...
         max_retry_times = 10
-        for source_nic_name, source_nic_info in vm_nics[source_node.name].items():
-            desc_nic_info = vm_nics[dest_node.name][source_nic_name]
-            dest_ip = vm_nics[dest_node.name][source_nic_name].ip_addr
+        for _, source_nic_info in vm_nics[source_node.name].items():
+            matched_dest_nic_name = ""
+            for dest_nic_name, dest_nic_info in vm_nics[dest_node.name].items():
+                # only when IPs are in the same range, IP1 of machine A can connect to
+                # IP2 of machine B
+                # e.g. eth2 IP is 10.0.2.3 on machine A, eth2 IP is 10.0.3.4 on machine
```
squirrelsc, line 638: It shouldn't happen like this way. If there is no nsg rule, every nic should be able to connected each other. Please ping me to check it offline.

microsoft/testsuites/network/sriov.py
```diff
@@ -312,6 +318,64 @@ def sriov_provision_with_max_nics_validation(
...
+    def _remove_module(self, node: Node) -> None:
+        lspci = node.tools[Lspci]
+        modprobe = node.tools[Modprobe]
+        devices_slots = lspci.get_devices_slots(
+            constants.DEVICE_TYPE_SRIOV, force_run=True
+        )
+        module_in_used = lspci.get_used_module(devices_slots[0], True)
+        source_reload_name = self.RELOAD_MODULES_DICT[module_in_used]
+        assert_that(source_reload_name).described_as("").is_not_none()
+        modprobe.remove(module_in_used)
```
squirrelsc: The network may be down after removed hv_netvsc. How about add a method called `reload`? Kameron is writing a similar test case. Please have a shared code. BTW he met error on removing the module in CentOS 7.4, you may can help on it.

lisa/tools/lspci.py
```diff
@@ -115,3 +115,20 @@ def enable_devices(self) -> None:
         self.node.tools[Echo].write_to_file(
             "1", self.node.get_pure_path("/sys/bus/pci/rescan"), sudo=True
         )
+
+    def get_used_module(self, slot: str, force_run: bool = False) -> str:
+        if force_run:
+            result = self.run(f"-nks {slot}", force_run=force_run, shell=True)
```
squirrelsc: which distro need the sudo? The executable handles sudo at check_exists. Does it work?

microsoft/testsuites/network/sriov.py
```diff
@@ -34,6 +33,12 @@
 )
 class Sriov(TestSuite):
     TIME_OUT = 300
+    global module_in_used
```
squirrelsc: 1. Avoid to use class level variable. It may be break if run in parallel.
2. The variable in class level is a class variable, it's shared cross all instances. It doesn't need the global keyword.

microsoft/testsuites/network/sriov.py
```diff
@@ -312,6 +318,64 @@ def sriov_provision_with_max_nics_validation(
...
+            ).is_not_equal_to("")
+        self.vm_nics[node.name] = node_nic_info._nics
+
+    def _remove_module(self, node: Node) -> None:
+        lspci = node.tools[Lspci]
+        modprobe = node.tools[Modprobe]
+        devices_slots = lspci.get_devices_slots(
+            constants.DEVICE_TYPE_SRIOV, force_run=True
+        )
+        module_in_used = lspci.get_used_module(devices_slots[0], True)
```
squirrelsc: Add comments to explain the `[0]`

microsoft/testsuites/network/sriov.py
LiliDeng: ok, here we just remove mlx*, so the network won't be down, I'll syn with Kameron.

lisa/tools/lspci.py
LiliDeng: I tested ubuntu and suse, suse need sudo

microsoft/testsuites/network/sriov.py
LiliDeng: add

microsoft/testsuites/network/sriov.py
LiliDeng: reload doesn't work here, since it need do verification after remove the mlx module.

microsoft/testsuites/network/sriov.py
LiliDeng: update

lisa/tools/lspci.py
squirrelsc: So it means the lspci is not in root path, but it can run by root only?

microsoft/testsuites/network/sriov.py
squirrelsc: You can ask Kameron create a reload method here.

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +319,107 @@ def sriov_add_max_nics_validation(
...
+            for _, node_nic in node_nic_info._nics.items():
+                assert_that(node_nic.lower).described_as(
+                    f"This interface {node_nic.upper} does not have a paired VF."
+                ).is_not_equal_to("")
+            vm_nics[node.name] = node_nic_info._nics
+        return vm_nics
+
+    def _remove_module(self, node: Node) -> str:
+        reload_modules_dict: Dict[str, str] = {
+            "mlx5_core": "mlx5_ib",
```
squirrelsc: Does it mean the unloaded and reloaded module name is different? Add comments to explain this dict.

microsoft/testsuites/network/sriov.py
```diff
@@ -324,64 +431,76 @@ def _sriov_basic_test(self, environment: Environment) -> None:
...
         # generate 200Mb file
         source_node.execute("dd if=/dev/urandom of=large_file bs=100 count=0 seek=2M")
         max_retry_times = 10
-        for source_nic_name, source_nic_info in vm_nics[source_node.name].items():
-            desc_nic_info = vm_nics[dest_node.name][source_nic_name]
-            dest_ip = vm_nics[dest_node.name][source_nic_name].ip_addr
+        for _, source_nic_info in vm_nics[source_node.name].items():
+            matched_dest_nic_name = ""
+            for dest_nic_name, dest_nic_info in vm_nics[dest_node.name].items():
+                # only when IPs are in the same range, IP1 of machine A can connect to
```
squirrelsc: It's same subnet, not "range". You can mention LISA deploy those nics into different subnets.

microsoft/testsuites/network/sriov.py
```diff
@@ -324,64 +431,76 @@ def _sriov_basic_test(self, environment: Environment) -> None:
...
+                    dest_nic_info.ip_addr.rsplit(".", maxsplit=1)[0]
+                    == source_nic_info.ip_addr.rsplit(".", maxsplit=1)[0]
+                ):
+                    matched_dest_nic_name = dest_nic_name
+                    break
+            assert_that(matched_dest_nic_name).described_as(
+                f"can't find the same IP range with {source_nic_info.ip_addr} on"
+                f" machine {source_node.name}, please check network setting of "
+                f"machine {dest_node.name}."
+            ).is_not_equal_to("")
```
squirrelsc: `is_empty`

microsoft/testsuites/network/sriov.py
LiliDeng: yes, mlx5_core is in used module, mlx5_ib is depended on mlx5_core, if unload mlx5_core, failure is mlx5_core is in use.

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +319,107 @@ def sriov_add_max_nics_validation(
...
+        lspci = node.tools[Lspci]
+        modprobe = node.tools[Modprobe]
+        devices_slots = lspci.get_devices_slots(
+            constants.DEVICE_TYPE_SRIOV, force_run=True
+        )
+        # there will not be multiple Mellanox types in one VM
+        # get the used module using any one of sriov device
+        module_in_used = lspci.get_used_module(devices_slots[0], True)
+        source_reload_name = reload_modules_dict[module_in_used]
+        assert_that(source_reload_name).is_not_none()
```
squirrelsc: this assert is not useful, the `dict[]` will fail, if it's not found. And the value is setup by the class itself.

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +319,107 @@ def sriov_add_max_nics_validation(
...
+            node_nic_info.initialize()
+            for _, node_nic in node_nic_info._nics.items():
+                assert_that(node_nic.lower).described_as(
+                    f"This interface {node_nic.upper} does not have a paired VF."
+                ).is_not_equal_to("")
+            vm_nics[node.name] = node_nic_info._nics
+        return vm_nics
+
+    def _remove_module(self, node: Node) -> str:
+        reload_modules_dict: Dict[str, str] = {
```
squirrelsc: move it to class level, it doesn't need to be initialized each time, when the method is called.

lisa/tools/lspci.py
```diff
@@ -115,3 +115,20 @@ def enable_devices(self) -> None:
         self.node.tools[Echo].write_to_file(
             "1", self.node.get_pure_path("/sys/bus/pci/rescan"), sudo=True
         )
+
+    def get_used_module(self, slot: str, force_run: bool = False) -> str:
```
squirrelsc: The force_run should be removed from this method. It should be queried each time.

lisa/tools/lspci.py
```diff
@@ -115,3 +115,20 @@ def enable_devices(self) -> None:
...
+            result = self.run(f"-nks {slot}", force_run=force_run, shell=True)
+            if result.exit_code != 0:
+                result = self.run("-nks", force_run=force_run, shell=True, sudo=True)
+                if result.exit_code != 0:
+                    raise LisaException(
+                        f"get unexpected non-zero exit code {result.exit_code} "
+                        f"when run {self.command} -nks {slot}."
+                    )
+        matched = find_patterns_in_lines(
+            result.stdout, [re.compile(r"Kernel driver in use: (.*)\r", re.M)]
```
squirrelsc: The static pattern should be moved to class level.

lisa/tools/lspci.py
```diff
@@ -115,3 +115,20 @@ def enable_devices(self) -> None:
...
+        if force_run:
+            result = self.run(f"-nks {slot}", force_run=force_run, shell=True)
+            if result.exit_code != 0:
+                result = self.run("-nks", force_run=force_run, shell=True, sudo=True)
+                if result.exit_code != 0:
+                    raise LisaException(
+                        f"get unexpected non-zero exit code {result.exit_code} "
+                        f"when run {self.command} -nks {slot}."
+                    )
+        matched = find_patterns_in_lines(
```
squirrelsc: It just get one str, use `get_matched_str`

microsoft/testsuites/network/sriov.py
squirrelsc: 1. add comments in code
2. I'm not understand it fully. As the logic, it just unloads the mlx5_ib, not unloads the mlx5_core. 
3. Is it possible to query module dependency in system, and remove it's all dependencies together?

LiliDeng:
@squirrelsc remain one case SRIOV-INTERRUPTS-CHANGE to be pushed into this PR, till now, not all comments are fixed.

squirrelsc:
> @squirrelsc remain one case SRIOV-INTERRUPTS-CHANGE to be pushed into this PR, till now, not all comments are fixed.

I will look this PR. Please move the remaining test case to next PR. You can send out the new PR and base to current branch, until it's merged.

lisa/tools/iperf3.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("iperf3")
+        if not self._check_exists():
+            self._install_from_src()
+        return self._check_exists()
+
+    def run_as_server_mode(self, daemon: bool = True) -> None:
+        cmd = "-s"
```
squirrelsc: add comments to explain each parameter.

lisa/tools/iperf3.py
```diff
@@ -0,0 +1,80 @@
...
+        self._command = "/usr/local/bin/iperf3"
+
+    def install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("iperf3")
+        if not self._check_exists():
+            self._install_from_src()
+        return self._check_exists()
+
+    def run_as_server_mode(self, daemon: bool = True) -> None:
```
squirrelsc: `run_as_server`

lisa/tools/iperf3.py
```diff
@@ -0,0 +1,80 @@
...
+            force_run=True,
+            shell=True,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"fail to launch cmd {self._command}"
+            f"{cmd}",
+        )
+
+    def run_as_client_mode(
+        self, time: int, server_ip: str, log_file: str = "", run_in_bg: bool = False
```
squirrelsc: 1. `time` to `seconds` or `minutes`?
2. give a default value.

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +320,264 @@ def sriov_add_max_nics_validation(
...
+        # preparation work before launch iperf3
+        self._stop_firewall(environment)
+
+        # run iperf3 on server side and client side
+        # iperfResults.log stored client side log
+        source_iperf3 = server_node.tools[Iperf3]
+        dest_iperf3 = client_node.tools[Iperf3]
+        source_iperf3.run_as_server_mode()
+        dest_iperf3.run_as_client_mode(
+            1800, server_node.internal_address, "iperfResults.log", True
```
squirrelsc: 1. is 1800 the 1800 seconds? It's too long to a P2 normal test case.
2. use a constant for the log file name, because it's used twice in the code.

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +320,264 @@ def sriov_add_max_nics_validation(
...
+        )
+
+        # wait for a while then check any error shown up in iperfResults.log
+        sleep(10)
```
squirrelsc: Do not put sleep here. change the `run_as_client` to `run_as_client` and `run_as_client_async`. Check other  sync/async pattern to implement it. Check if client process exited or not to detect error.

lisa/tools/lspci.py
```diff
@@ -115,3 +115,20 @@ def enable_devices(self) -> None:
         self.node.tools[Echo].write_to_file(
             "1", self.node.get_pure_path("/sys/bus/pci/rescan"), sudo=True
         )
+
+    def get_used_module(self, slot: str, force_run: bool = False) -> str:
+        if force_run:
```
squirrelsc: This check is not necessary. The force_run is handled in the `run` method already.

microsoft/testsuites/network/sriov.py
LiliDeng: 2. firstly we need check the vf use which module, if use mlx5_core, since mlx5_ib depends on it, reload and load mlx5_ib.
3. I try use modprobe --show-depends + recursive, didn't make it. the logic is we need check all the modules which depends on mlx5_core, let 's say module A and B, then check which modules depend on module A and B, till there is no any modules depend on the modules you want to remove, then start to remove, otherwise we will see 'the module is in used' issue, since the mlx modules have simple dependency, so I use dict here.

microsoft/testsuites/network/sriov.py
LiliDeng: 1. just make it keep running during following steps, the case won't wait for this end.

lisa/tools/lspci.py
LiliDeng: update

microsoft/testsuites/network/sriov.py
LiliDeng: add comments

microsoft/testsuites/network/sriov.py
LiliDeng: update

microsoft/testsuites/network/sriov.py
LiliDeng: update

microsoft/testsuites/network/sriov.py
LiliDeng: update

microsoft/testsuites/network/sriov.py
LiliDeng: update

lisa/tools/lspci.py
LiliDeng: update

lisa/tools/lspci.py
LiliDeng: update

lisa/tools/lspci.py
LiliDeng: update

lisa/tools/iperf3.py
LiliDeng: added

lisa/tools/iperf3.py

lisa/tools/iperf3.py

lisa/tools/lspci.py
LiliDeng: update

lisa/tools/lspci.py
```diff
@@ -83,12 +85,10 @@ def get_device_list(self, force_run: bool = False) -> List[PciDevice]:
             self._pci_devices = []
             result = self.run("-m", force_run=force_run, shell=True)
             if result.exit_code != 0:
-                result = self.run("-m", force_run=force_run, shell=True, sudo=True)
-                if result.exit_code != 0:
-                    raise LisaException(
-                        f"get unexpected non-zero exit code {result.exit_code} "
-                        f"when run {self.command} -m."
-                    )
+                raise LisaException(
```
squirrelsc: The simple exit_code check can be merged into `self.run` by using `expected_exit_code` and `expected_exit_code_failure_message`.

microsoft/testsuites/network/sriov.py
```diff
@@ -34,26 +34,36 @@
 )
 class Sriov(TestSuite):
     TIME_OUT = 300
+    # ConnectX-3 uses mlx4_core
```
squirrelsc, line 37: Explain, why the dict is needed here. The reason is like if the dependent module is not removed, the module like mlx5_core cannot be unloaded with error "in use"...

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +328,263 @@ def sriov_add_max_nics_validation(
...
+        assert_that(iperf_log).does_not_contain("error")
+
+    def _initialize_nic_info(
+        self, environment: Environment
+    ) -> Dict[str, Dict[str, NicInfo]]:
+        vm_nics: Dict[str, Dict[str, NicInfo]] = {}
+        for node in environment.nodes.list():
+            node_nic_info = Nics(node)
+            node_nic_info.initialize()
+            for _, node_nic in node_nic_info._nics.items():
```
squirrelsc: It looks the `_nics` should be changed to a public name like `nics`

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +328,263 @@ def sriov_add_max_nics_validation(
...
+        1. Change scatter-gather feature on synthetic nic,
+         verify the the feature status sync to the VF dynamically.
+        2. Disable and enable sriov,
+         check the scatter-gather feature status keep consistent in VF.
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_count=2,
+            network_interface=schema.NetworkInterfaceOptionSettings(
+                nic_count=3,
```
squirrelsc: What's the reason it needs exact 3 nics? How about limit the minimum size only?

LiliDeng: there are two cases in lisav2, one is to use 1 nic, the other is to use 3 nics, I think use one case with 3 nics is enough?

lisa/tools/iperf3.py
```diff
@@ -0,0 +1,95 @@
...
+        # --logfile: save logs into specified file
+        cmd = f"-t {seconds} -c {server_ip}"
+        if log_file:
+            if self.node.shell.exists(self.node.get_pure_path(log_file)):
+                self.node.shell.remove(self.node.get_pure_path(log_file))
+            cmd += f" --logfile {log_file}"
+        process = self.node.execute_async(f"{self._command} {cmd}", sudo=self._use_sudo)
+        return process
+
+    def run_as_client(
```
squirrelsc, line 80: refer to the underlying code. The sync version should call async version, and wait the result. It doesn't need to copy the implementation.

lisa/tools/lspci.py
LiliDeng: fixed

microsoft/testsuites/network/sriov.py
LiliDeng: updated

lisa/tools/iperf3.py
LiliDeng: update

microsoft/testsuites/network/sriov.py
squirrelsc: How about change it to min 3 and max 6? Then add a comments about why the number is chose.

lisa/tools/iperf3.py
```diff
@@ -0,0 +1,99 @@
...
+        # -c: run iperf3 as client mode, followed by iperf3 server ip address
+        # -t: run iperf3 testing for given seconds
+        # --logfile: save logs into specified file
+        cmd = f"-t {seconds} -c {server_ip}"
+        if log_file:
+            if self.node.shell.exists(self.node.get_pure_path(log_file)):
+                self.node.shell.remove(self.node.get_pure_path(log_file))
+            cmd += f" --logfile {log_file}"
+        process = self.node.execute_async(f"{self.command} {cmd}", sudo=self._use_sudo)
+        if wait:
```
squirrelsc: The wait part should be in the sync method. It's confusing that the async method provide wait option.

microsoft/testsuites/network/sriov.py
```diff
@@ -310,9 +329,263 @@ def sriov_add_max_nics_validation(
...
+            ).is_equal_to(False)
+
+        # check there is no error happen in iperf3 log
+        # after above operations
+        dest_cat = client_node.tools[Cat]
+        iperf_log = dest_cat.read_from_file(
+            client_iperf3_log, sudo=True, force_run=True
+        )
+        assert_that(iperf_log).does_not_contain("error")
+        dest_iperf3.kill_processes()
```
squirrelsc: The returned process can be killed directly. It doesn't need to reimplement it.

lisa/tools/kill.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+
+
+class Kill(Tool):
```
squirrelsc, line 9: The tool is not necessary so far, the process object can be killed itself.

lisa/util/process.py
```diff
@@ -197,7 +197,7 @@ def wait_result(
                 f"execution time: {self._timer}, exit code: {self._result.exit_code}"
             )
 
-        if expected_exit_code:
+        if expected_exit_code is not None:
```
squirrelsc, line 200: when will it be None?

microsoft/testsuites/network/sriov.py
```diff
@@ -399,23 +684,33 @@ def _sriov_vf_connection_test(
...
-            )
             assert_that(
                 int(dest_tx_packets), "insufficient RX packets received"
             ).is_greater_than(int(dest_tx_packets_origin))
+
             if turn_off_vf:
                 source_node.execute(f"ip link set dev {source_vf_nic} up", sudo=True)
                 dest_node.execute(f"ip link set dev {dest_vf_nic} up", sudo=True)
+
+    def _stop_firewall(self, environment: Environment) -> None:
```
squirrelsc, line 711: how about add a virtual tool, called firewall?

lisa/tools/kill.py
LiliDeng: if reuse an existing environment, and already have iperf3 running.

lisa/util/process.py
LiliDeng: when not set? current logic is wrong, when expected_exit_code  = 0, it never enters into if

lisa/tools/kill.py
squirrelsc: 1. You can use the `after_case` to kill the iperf3 process. The advantage is, even the test case is failed, there is a chance to kill it.
2. Implement the find by name in the Kill tool also. For example, `def by_name(self, process_name: str) -> None`.

microsoft/testsuites/network/sriov.py
LiliDeng: actually >=1 is okay.

lisa/tools/kill.py
LiliDeng: how to get nodes in after_case? @squirrelsc

microsoft/testsuites/network/sriov.py
LiliDeng: added

lisa/tools/iperf3.py

microsoft/testsuites/network/sriov.py
squirrelsc: Is it changed already?

lisa/tools/firewall.py
```diff
@@ -0,0 +1,38 @@
...
+        self._stop_ufw()
+        self._stop_firewalld()
+        self._stop_iptables()
+
+    def _stop_ufw(self) -> None:
+        self.node.execute("command -v ufw && ufw disable", shell=True, sudo=True)
+
+    def _stop_firewalld(self) -> None:
+        service = self.node.tools[Service]
+        service.stop_service("SuSEfirewall2")
```
squirrelsc, line 77: Refer to service, implement it by different distro.

lisa/tools/kill.py
squirrelsc: It's the same as test case.

lisa/tools/lspci.py
```diff
@@ -129,3 +131,16 @@ def enable_devices(self) -> None:
         self.node.tools[Echo].write_to_file(
             "1", self.node.get_pure_path("/sys/bus/pci/rescan"), sudo=True
         )
+
+    def get_used_module(self, slot: str) -> str:
```
squirrelsc, line 135: `get_in_use_modules`. It may return multiple values, it needs to split them, if so.

lisa/tools/kill.py
LiliDeng: I passed 'environment: Environment', it doesn't work.

microsoft/testsuites/network/sriov.py
LiliDeng: not yet, I didn't figure out which value should I specify

squirrelsc: Well, you said there is test case with 1 nic, and 8 nics. How about use ">1" in this case?

microsoft/testsuites/network/sriov.py
```diff
@@ -399,23 +685,33 @@ def _sriov_vf_connection_test(
...
+        for node in environment.nodes.list():
+            firewall = node.tools[Firewall]
+            firewall.stop()
+
+    def _kill_iperf3(self, environment: Environment) -> None:
+        for node in environment.nodes.list():
+            kill = node.tools[Kill]
+            kill.by_name("iperf3")
+
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
```
squirrelsc: put the public method above private methods.

microsoft/testsuites/network/sriov.py
```diff
@@ -399,23 +685,33 @@ def _sriov_vf_connection_test(
...
             if turn_off_vf:
                 source_node.execute(f"ip link set dev {source_vf_nic} up", sudo=True)
                 dest_node.execute(f"ip link set dev {dest_vf_nic} up", sudo=True)
+
+    def _stop_firewall(self, environment: Environment) -> None:
+        for node in environment.nodes.list():
+            firewall = node.tools[Firewall]
+            firewall.stop()
+
+    def _kill_iperf3(self, environment: Environment) -> None:
```
squirrelsc: This method is used in after_case once, and it's short. It doesn't need to be a method now.

lisa/tools/kill.py
LiliDeng: update

lisa/tools/firewall.py
LiliDeng: we are not sure which firewall(s) enabled in which distro, I updated the code, but not by different distro, could you please take a look?

lisa/tools/lspci.py
LiliDeng: I double check, didn't find a device's in used modules more than 2.

microsoft/testsuites/network/sriov.py
LiliDeng: update

microsoft/testsuites/network/sriov.py
LiliDeng: update

squirrelsc (approved):

PR #1597, opened 2021-10-27
DPDK: Allow branch selection from runbook

mcgov:
Add ability to specify a dpdk branch variable to checkout a git ref for the dpdk repo.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -3,7 +3,7 @@
 
 import re
 from collections import deque
-from typing import Dict, List, Tuple
+from typing import Any, Dict, List, Tuple
 
 from assertpy import assert_that, fail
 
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -350,7 +358,12 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
     )
 
 
-def initialize_node_resources(node: Node, log: Logger, pmd: str) -> DpdkTestResources:
+def initialize_node_resources(
+    node: Node, log: Logger, variables: Dict[str, Any], pmd: str
+) -> DpdkTestResources:
+    assert_that(variables).contains_key("dpdk_branch")
```
squirrelsc: It should support a default value. In most test pass, the version should be the default value. The design principle of v3 is to simplify test runs, it means minimize the number of parameters. So change the lines like ,
`dpdk_branch = variables.get("dpdk_branch", "default_branch")`

mcgov: I think our default behavior is going to be installing from the package manager. I'll swap this syntax around but the default would still be "" so we can truth check it and default over to the package manager install.

mcgov: Also I would like to wait to remove the assert until the package manager install is finished, it's about halfway done rn but will  be the next PR to show up

mcgov: Can we merge this? I updated to use the variables.get method, there is a second PR coming that will implement the default path so I'll remove the assert in that PR

squirrelsc: The assertion should be removed also, it makes the case failed, if the variable is not provided.

mcgov: removing

squirrelsc (approved):

PR #1600, opened 2021-10-27
set a default value for max_nic_count

LiliDeng:
for some new sizes, there is no MaxNetworkInterfaces capability, it caused case skip.
image

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1444,6 +1444,10 @@ class AzurePlatform(Platform):
             elif name == "EphemeralOSDiskSupported":
                 if eval(sku_capability.value) is True:
                     node_space.disk.disk_type.add(schema.DiskType.Ephemeral)
+        # for some new sizes, there is no MaxNetworkInterfaces capability
+        # and we have to set a default value for max_nic_count
+        if not node_space.network_interface.max_nic_count:
+            node_space.network_interface.max_nic_count = 8
 
         # some vm size do not have resource disk present
         # https://docs.microsoft.com/en-us/azure/virtual-machines/azure-vms-no-temp-disk


mcgov:
I cherry picked this to a testing branch and I am still getting some skips: (testing with  vm_size: Standard_D15_v2)
```
********3588, 'LSG Dev & Test'
2021-10-27 17:44:25.577[30144][INFO] lisa.runner[0] No requirement of test case is suitable for the preparation error of the environment 'generated_0'. Randomly attach a test case to this environment. This may be because the platform failed before populating the features into this environment.
2021-10-27 17:44:25.578[30144][INFO] lisa.runner[0] case skipped: cannot get min value, capability doesn't support requirement:["max_nic_count: capability shouldn't be None"]
2021-10-27 17:44:25.582[30144][INFO] lisa.runner[0] 'generated_0' attached to test case 'Dpdk.verify_dpdk_ring_ping(lisa_0_0)': cannot get min value, capability doesn't support requirement:["max_nic_count: capability shouldn't be None"]
```

squirrelsc:
@LiliDeng can you file a bug to fix it, or ask if is there another reliable way to get this info.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1444,6 +1444,10 @@ def _resource_sku_to_capability(  # noqa: C901
             elif name == "EphemeralOSDiskSupported":
                 if eval(sku_capability.value) is True:
                     node_space.disk.disk_type.add(schema.DiskType.Ephemeral)
+        # for some new sizes, there is no MaxNetworkInterfaces capability
+        # and we have to set a default value for max_nic_count
+        if not node_space.network_interface.max_nic_count:
+            node_space.network_interface.max_nic_count = 1
```
squirrelsc: If it's not specified, let's put it in the max value to give a looser range.

mcgov: setting default as max (8) fixes the issue I was seeing

squirrelsc (approved):

PR #1601, opened 2021-10-27
Add kdumpcrash_validate_on_random_cpu to cover VM size having 9-32 cores

lubaihua33:
1. Specify the cpu 1 to trigger kdump in kdumpcrash_validate_smp test case;
2. Add kdumpcrash_validate_on_random_cpu to cover VM size having 9-32 cores

### DIFF
--- a/microsoft/testsuites/kdump/kdumpcrash.py
+++ b/microsoft/testsuites/kdump/kdumpcrash.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 import time
 from pathlib import PurePosixPath
+from random import randint
 from typing import cast
 
 from assertpy import assert_that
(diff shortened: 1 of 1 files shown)


squirrelsc:
What's the reason to remove limitation on cpu core? my understanding is the kdump has difference behavior with different cpu count range. If the max is removed, The test case may be run on a big cpu count VM together. It's not expected.

lubaihua33:
> What's the reason to remove limitation on cpu core? my understanding is the kdump has difference behavior with different cpu count range. If the max is removed, The test case may be run on a big CPU count VM together. It's not expected.

Thank you Chi. 
I have changed.  In the current kdump test suit, if the VM size has 9- 32 cores. The kdump test cases will always be skipped. I want to add one case to cover the VM size which has 9-32 cores.

squirrelsc:
> > What's the reason to remove limitation on cpu core? my understanding is the kdump has difference behavior with different cpu count range. If the max is removed, The test case may be run on a big CPU count VM together. It's not expected.
> 
> Thank you Chi. I have changed. In the current kdump test suit, if the VM size has 9- 32 cores. The kdump test cases will always be skipped. I want to add one case to cover the VM size which has 9-32 cores.

How about add a test case without cpu core limitation? It will run on any vm size.

lubaihua33:
> > > What's the reason to remove limitation on cpu core? my understanding is the kdump has difference behavior with different cpu count range. If the max is removed, The test case may be run on a big CPU count VM together. It's not expected.
> > 
> > 
> > Thank you Chi. I have changed. In the current kdump test suit, if the VM size has 9- 32 cores. The kdump test cases will always be skipped. I want to add one case to cover the VM size which has 9-32 cores.
> 
> How about add a test case without cpu core limitation? It will run on any vm size.

Sure, I think so. Have changed.

squirrelsc (approved):

PR #1602, opened 2021-10-28
Test Migration: reload_hyperv_modules

kamalca:
Covers v2 test RELOAD-MODULES-SMP
Reloads hv_netsvc 100 times as a stress test.

~I am leaving this as a draft until I can resolve why the hv_netvsc module is being used on some distros (blocking reload).~
Update: the issue was resolved by installing latest LIS modules and rebooting the system.

### DIFF
--- a/microsoft/testsuites/core/hv_module.py
+++ b/microsoft/testsuites/core/hv_module.py
@@ -198,6 +198,46 @@ class HvModule(TestSuite):
             "Not all Hyper V drivers are present."
         ).is_length(0)
 
+    @TestCaseMetadata(
+        description="""
+        This test case will reload hyper-v modules as a stress test.
+        """,
+        priority=1,
+    )
+    def reload_hyperv_modules(self, case_name: str, log: Logger, node: Node) -> None:
+        # Constants
+        module = "hv_netvsc"
+        loop_count = 100
+
+        if isinstance(node.os, Redhat):
+            try:
+                log.debug("Checking LIS installation before reload.")
+                node.tools[LisDriver]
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/core/hv_module.py
```diff
@@ -198,6 +198,39 @@ def verify_hyperv_modules(self, log: Logger, node: Node) -> None:
...
+
+        if module not in self._get_expected_modules(node):
+            raise SkippedException(
+                f"{module} is loaded statically into the "
+                "kernel and therefore can not be reloaded"
+            )
+
+        for i in range(loop_count):
+            result = node.execute(
+                f"modprobe -r -v {module}; modprobe -v {module}",
```
squirrelsc: Does it work on centos old versions now?

microsoft/testsuites/core/hv_module.py
```diff
@@ -198,6 +198,39 @@ def verify_hyperv_modules(self, log: Logger, node: Node) -> None:
             "Not all Hyper V drivers are present."
         ).is_length(0)
 
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Reload hyper-v modules
```
squirrelsc: If it's only one point, it doesn't need to be the bullet point format.

microsoft/testsuites/core/hv_module.py
kamalca: I am still blocked on this issue.

squirrelsc (approved):

PR #1603, opened 2021-10-30
allow for shell variables in rhel package installs

mcgov:
Enable use of shell variables for packages in RHEL, useful for kernel-devel-$(uname -r)


squirrelsc (approved):

squirrelsc:
It's ok to enable it, but the uname is actually a tool in v3 already. We can use it to compose the string before it passes to install_packages.

PR #1607, opened 2021-11-01
Enable tar.gz source for dpdk

mcgov:
Enable tar.gz source archives for dpdk testing. This required a few changes:
- Add a variable for dpdk_source, this can be a git link or a link to a targz archive.
- get and expand the archive, rename it from it's expanded name which has the version number to just dpdk to match the git checkout name
- Enable a default path when neither is selected where the regular dpdk-stable git is used and main is checked out.
The if conditions in 'install' were getting too nested, so some minor cleanup is in this commit too:
- Remove the return type from install_dependencies since it is unused.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -27,6 +27,7 @@ from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
 
 VDEV_TYPE = "net_vdev_netvsc"
 MAX_RING_PING_LIMIT_NS = 200000
+DPDK_STABLE_GIT = "http://dpdk.org/git/dpdk-stable"
 
 
 @TestSuiteMetadata(
@@ -134,7 +135,7 @@ class Dpdk(TestSuite):
         echo = node.tools[Echo]
         rping_build_env_vars = [
             "export RTE_TARGET=build",
-            f"export RTE_SDK={testpmd.dpdk_path.as_posix()}",
+            f"export RTE_SDK={str(testpmd.dpdk_path)}",
         ]
         echo.write_to_file(
             ";".join(rping_build_env_vars), node.get_pure_path("~/.bashrc"), append=True
@@ -361,7 +362,12 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
 def initialize_node_resources(
     node: Node, log: Logger, variables: Dict[str, Any], pmd: str
 ) -> DpdkTestResources:
+    dpdk_source = variables.get("dpdk_source", DPDK_STABLE_GIT)
     dpdk_branch = variables.get("dpdk_branch", "")
+    log.info(
+        "Dpdk initialize_node_resources running"
+        f"found dpdk_source '{dpdk_source}' and dpdk_branch '{dpdk_branch}'"
+    )
 
     network_interface_feature = node.features[NetworkInterface]
     sriov_is_enabled = network_interface_feature.is_enabled_sriov()
@@ -378,6 +384,7 @@ def initialize_node_resources(
 
     # initialize testpmd tool (installs dpdk)
     testpmd = DpdkTestpmd(node)
+    testpmd.set_dpdk_source(dpdk_source)
     testpmd.set_dpdk_branch(dpdk_branch)
     testpmd.install()
 
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -11,7 +11,7 @@ from assertpy import assert_that, fail
 from lisa.executable import Tool
 from lisa.nic import NicInfo
 from lisa.operating_system import CentOs, Redhat, Ubuntu
-from lisa.tools import Echo, Git, Lspci, Wget
+from lisa.tools import Echo, Git, Lspci, Tar, Wget
 from lisa.util import SkippedException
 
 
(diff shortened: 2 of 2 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -361,6 +361,7 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
 def initialize_node_resources(
     node: Node, log: Logger, variables: Dict[str, Any], pmd: str
 ) -> DpdkTestResources:
+    dpdk_source = variables.get("dpdk_source", "")
```
squirrelsc: It can like below to apply default value:

```python
dpdk_source = variables.get("dpdk_source", _DPDK_STABLE_GIT)
```

squirrelsc: It's also good to print out what value is got here. So if there is typo on variable names, it would be found easier.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -111,45 +115,69 @@ def _install(self) -> bool:
...
+                    )
+                )
+            working_path = node.working_path.as_posix()
+            wget_tool.get(
+                self._dpdk_source,
+                working_path,
+            )
+            dpdk_filename = self._dpdk_source.split("/")[-1]
+            node.log.info(node.execute("ls -la", cwd=node.working_path))
+            dpdk_foldername = tar_tool.extract(
```
squirrelsc: How about to support dest path in tar? So it doesn't need extra mv command.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -111,45 +115,69 @@ def _install(self) -> bool:
...
-        if self._install_dependencies():
-            node = self.node
-            git_tool = node.tools[Git]
-            echo_tool = node.tools[Echo]
+        self._install_dependencies()
+        node = self.node
+        git_tool = node.tools[Git]
+        echo_tool = node.tools[Echo]
+        wget_tool = node.tools[Wget]
+        tar_tool = node.tools[Tar]
```
squirrelsc: move it in to the if clause, so the tool doesn't need to be created, if it's not needed.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -101,6 +102,9 @@ def can_install(self) -> bool:
     def dependencies(self) -> List[Type[Tool]]:
         return [Git, Wget]
 
+    def set_dpdk_source(self, dpdk_source: str) -> None:
+        self._dpdk_source = dpdk_source
```
squirrelsc, line 103: Not to use class level variable to save this information. It may break when run in parallel.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I will do this, but let me do it in a second PR. If I'm going to add that I'd like to change this tar tool to use the pathlib and that seems like it deserves it's own commit.

mcgov: Can I change dest_dir to use pathlib? Otherwise appending the new output folder onto the dest_dir gets weird

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I see, I thought I couldn't add parameters to install since it was a parent function but I CAN pass in additional kwargs. I'll change these to function arguments for install()

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Going to raise a separate PR to enable renaming the output dir and we can hold this one until it's finished

mcgov: Ah this is actually more complicated than I thought at first. A tar can make lots of output folders from one file

squirrelsc: It should be able to rename the root path to extract. It should be easy. If the folder is in the tar file, it's ok to live with it, we don't need to rename it.

mcgov: It's not as straightforward as you would think, I opened another PR for this though so we can discuss further there

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -111,45 +112,69 @@ def _install(self) -> bool:
...
+                )
+            working_path = node.working_path.as_posix()
+            wget_tool.get(
+                self._dpdk_source,
+                working_path,
+            )
+            dpdk_filename = self._dpdk_source.split("/")[-1]
+            node.log.info(node.execute("ls -la", cwd=node.working_path))
+            dpdk_foldername = tar_tool.extract(
+                node.working_path.joinpath(dpdk_filename).as_posix(),
```
squirrelsc: the as_posix is not necessary. The working_path itself is already posix, if the remote node is.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -111,45 +112,69 @@ def _install(self) -> bool:
...
+            wget_tool = node.tools[Wget]
+            tar_tool = node.tools[Tar]
+            if self._dpdk_branch:
+                node.log.warn(
+                    (
+                        "DPDK tarball source does not need dpdk_branch defined. "
+                        "User-defined variable dpdk_branch will be ignored."
+                    )
+                )
+            working_path = node.working_path.as_posix()
```
squirrelsc: The as_posix may not necessary, the os is checked underlying. When it needs to convert to str, try `str(node.working_path)`.

squirrelsc (approved):

PR #1608, opened 2021-11-02
Tar tool: enable strip-components and file content listing

mcgov:
- add 'list components' feature to check tar contents
- add methods to get output directories from a tar files
- add convenience method to get the single output dir name if
    there is a single one
- add support for 'strip components' flag which is useful for renaming
    an output dir if there is a single top-level-folder in the tar.
- fix use of Tar tool in kdump test to use the new version of the tool.

### DIFF
--- a/lisa/tools/tar.py
+++ b/lisa/tools/tar.py
@@ -1,6 +1,10 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from typing import Callable, List
+
+from assertpy import assert_that
+
 from lisa.executable import Tool
 from lisa.util import LisaException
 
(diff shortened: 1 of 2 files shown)


lisa/tools/tar.py
```diff
@@ -10,16 +14,60 @@ class Tar(Tool):
...
+        result = self.run(tar_cmd, shell=True, force_run=True, sudo=True)
         if result.exit_code != 0:
             raise LisaException(
                 f"Failed to extract file to {dest_dir}, {result.stderr}"
             )
-        # Get the folder name of the extracted file
-        source_name = result.stdout.split("/")[0]
-        return source_name
+
+    def get_tar_contents(self, file: str) -> List[str]:
```
squirrelsc: as the description, how about `list`?

lisa/tools/tar.py
```diff
@@ -10,16 +14,60 @@ class Tar(Tool):
...
+            force_run=True,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"Could not list items in tar file {file}"
+            ),
+        )
+        return result.stdout.splitlines()
+
+    def get_top_level_output_folders(self, file: str) -> List[str]:
```
squirrelsc: How about add a parameter to above `list` as `recursive`?

lisa/tools/tar.py
```diff
@@ -10,16 +14,60 @@ class Tar(Tool):
...
+                self.node.log.debug(
+                    f"NOTE: tar file {file} contains a regular file "
+                    f"{tar_contents} in the top level output."
+                )
+            else:
+                split_path = tar_contents.split(posix_delimiter)
+                folder_set.add(split_path[0])
+        return list(folder_set)
+
+    def get_single_top_level_output_folder(self, file: str) -> str:
```
squirrelsc: What's this is used for?

lisa/tools/tar.py
mcgov: I think tar -t does list recursively already so it should show all the file and folder contents of the tar

squirrelsc: I mean, it doesn't need to have another method to output top level folder only. We can use parameters to integrate them into one method. like `list(recursive: bool = True, is_folder_only: bool = False)`

lisa/tools/tar.py
mcgov: For getting the name of the output folder if there is a single one. The old version of Tar.extract assumed there would be a single output folder, since this is pretty common this function would be used to get the output folder name for those cases. The advantage is that this will now assert if they have made a mistake and the tarball has more than one output folder.

lisa/tools/tar.py
mcgov: Ooooh I see. Yep that makes sense

lisa/tools/tar.py
```diff
@@ -10,16 +14,60 @@ class Tar(Tool):
...
+            force_run=True,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"Could not list items in tar file {file}"
+            ),
+        )
+        content = result.stdout.splitlines()
+
+        if recursive:
```
squirrelsc: The `folders_only` parameter should be handled by this branch too.

lisa/tools/tar.py
```diff
@@ -10,16 +14,60 @@ class Tar(Tool):
...
+            for item in content:
+                if not folders_only:
+                    if "/" not in item:
+                        files += item
+                path_split = item.split("/")
+                if len(path_split) >= 2:
+                    folders.add(path_split[0])
+            return files + list(folders)
+
+    def get_sole_top_level_output_folder(self, file: str) -> str:
```
squirrelsc: How about a shorter name like `get_root_folder`? And maybe the `Tar.extract` can return this folder, so this method is not necessary.

lisa/tools/tar.py
mcgov: (facepalm) getting coffee brb

lisa/tools/tar.py
mcgov: That makes sense, I wanted to hint that there's an assumption of a single one, but get_single_root_folder or something is cleaner for sure. Thanks!

mcgov:
blocking on testing

squirrelsc (dismissed):

squirrelsc (approved):

PR #1610, opened 2021-11-03
Add switch for exit code

dsrivastavv:


lisa/schema.py
```diff
@@ -1215,6 +1215,7 @@ def type_name(cls) -> str:
 class Runbook:
     # run name prefix to help grouping results and put it in title.
     name: str = "not_named"
+    exit_code: str = "test_failed_count" # test_failed_count or run_status
```
squirrelsc: It has two states so far, how about a boolean? It's easier to use. The name can be `exit_with_failed_count`? Maybe something shorter.

dsrivastavv: I think this might be better

squirrelsc (approved):

PR #1612, opened 2021-11-04
NIC optimizations

mcgov:
Optimizing NIC info gathering.
- Swap from making a bunch of ssh connections to parse sysfs bit by bit to wildcarding and parsing the output w regex.


mcgov:
Need to label or remove some log spots, will push another change soon

squirrelsc (approved):

PR #1614, opened 2021-11-05
DPDK AN Disable tests:

mcgov:
Intial vf revoke testing, send only.
- adds some helper functions which I'll use elsewhere once we clear review and merge this PR


mcgov:
reworking a bunch of this to make it cleaner and support more nodes easily

PR #1615, opened 2021-11-05
gpu: add testcase for validating gpu pci device rescind

sharsonia:


sharsonia:
CUDA Driver Installation and loading test result with 
        marketplace: canonical ubuntuserver 18.04-lts 18.04.202105120
        vm_size: Standard_NC4as_T4_v3
        location: westus2

[lisa-20211107-115039-484.log](https://github.com/microsoft/lisa/files/7492217/lisa-20211107-115039-484.log)

lisa/operating_system.py
```diff
@@ -535,15 +536,24 @@ def _initialize_package_installation(self) -> None:
         result.assert_exit_code(message="\n".join(self.get_apt_error(result.stdout)))
 
     def _install_packages(self, packages: List[str], signed: bool = True) -> None:
+        file_packages = []
+        for index, package in enumerate(packages):
+            if package.endswith(".deb"):
```
squirrelsc, line 541: what's the reason to add this logic? It looks other names are ignored, if there is no ".deb" in name.

lisa/operating_system.py
```diff
@@ -535,15 +536,24 @@ def _initialize_package_installation(self) -> None:
...
+            f"DEBIAN_FRONTEND=noninteractive apt-get -y install {' '.join(packages)}"
         )
         if not signed:
             command += " --allow-unauthenticated"
-
         self.wait_running_package_process()
-        install_result = self._node.execute(command, sudo=True)
+        if file_packages:
+            self._node.execute(f"dpkg -i {' '.join(file_packages)}", sudo=True)
+            self._node.execute("apt update", sudo=True)
```
squirrelsc: it doesn't need to run `apt update`, it's handled automatically.

lisa/operating_system.py
sharsonia: The reason to add this is -
Generally the package names that are passed to be installed are just names and not files with extension. But if we have files like .deb packages to be installed then they gets installed first by runnign dpkg and then running apt-get.
Other packages are not being ignored, just that if the package ends with .deb then dpkg command is run first.

squirrelsc: please add the reasons in the code.

squirrelsc (approved):

PR #1616, opened 2021-11-09
Test Migration: verify_hpc_vm

kamalca:
This test case verifies that a machine is HPC, that the proper drivers are present, and that the configuration is correct.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -1,0 +1,28 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+
+from lisa.feature import Feature
+
+FEATURE_NAME_INFINIBAND = "Infiniband"
+
+
+class Infiniband(Feature):
+    def enabled(self) -> bool:
+        return True
+
+    @classmethod
+    def name(cls) -> str:
+        return FEATURE_NAME_INFINIBAND
+
+    @classmethod
+    def can_disable(cls) -> bool:
+        return True
+
+    def is_over_sriov(self) -> bool:
+        raise NotImplementedError
+
+    # nd stands for network direct
+    # example SKU: Standard_H16mr
+    def is_over_nd(self) -> bool:
+        raise NotImplementedError
--- a/microsoft/testsuites/hpc/infinibandsuit.py
+++ b/microsoft/testsuites/hpc/infinibandsuit.py
@@ -1,0 +1,102 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from assertpy import assert_that
+
+from lisa import (
+    Logger,
+    Node,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    simple_requirement,
+)
+from lisa.features import Infiniband, Sriov
+from lisa.sut_orchestrator.azure.tools import Waagent
+from lisa.tools import Modprobe
+from lisa.util import SkippedException
+
+
+@TestSuiteMetadata(
+    area="core",
+    category="functional",
+    description="""
+    Tests the functionality of infiniband.
+    """,
+)
+class InfinibandSuit(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Determine whether the VM has Infiniband over SR-IOV
+        2. Ensure waagent is configures with OS.EnableRDMA=y
+        3. Check that appropriate drivers are present
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            network_interface=Sriov(), supported_features=[Infiniband]
+        ),
+    )
+    def verify_hpc_over_sriov(self, log: Logger, node: Node) -> None:
+
+        infiniband = node.features[Infiniband]
+        assert_that(infiniband.is_over_sriov()).described_as(
+            "Based on VM SKU information we expected Infiniband over SR-IOV,"
+            " but no matching devices were found."
+        ).is_true()
+
+        waagent = node.tools[Waagent]
+        assert_that(waagent.is_rdma_enabled()).described_as(
+            "Found waagent configuration of OS.EnableRDMA=y "
+            "was missing or commented out"
+        ).is_true()
+        log.debug("Verified waagent config OS.EnableRDMA=y set successfully")
+
+        modprobe = node.tools[Modprobe]
+        expected_modules = [
+            "mlx5_ib",
+            "ib_uverbs",
+            "ib_core",
+            "mlx5_core",
+            "mlx_compat",
+            "rdma_cm",
+            "iw_cm",
+            "ib_cm",
+            "ib_core",
+        ]
+
+        for module in expected_modules:
+            assert_that(modprobe.is_module_loaded(module)).described_as(
+                f"Module {module} is not loaded."
+            ).is_true()
+
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Determine whether the VM has Infiniband over Network Direct
+        2. Ensure waagent is configures with OS.EnableRDMA=y
+        3. Check that appropriate drivers are present
+        """,
+        priority=2,
+        requirement=simple_requirement(supported_features=[Infiniband]),
+    )
+    def verify_hpc_over_nd(self, log: Logger, node: Node) -> None:
+
+        infiniband = node.features[Infiniband]
+        if not infiniband.is_over_nd():
+            raise SkippedException("Inifiniband over ND was not detected.")
+
+        waagent = node.tools[Waagent]
+        assert_that(waagent.is_rdma_enabled()).described_as(
+            "Found waagent configuration of OS.EnableRDMA=y "
+            "was missing or commented out"
+        ).is_true()
+        log.debug("Verified waagent config OS.EnableRDMA=y set successfully")
+
+        modprobe = node.tools[Modprobe]
+        expected_modules = ["mlx5_ib", "hv_networkdirect"]
+
+        for module in expected_modules:
+            assert_that(modprobe.is_module_loaded(module)).described_as(
+                f"Module {module} is not loaded."
+            ).is_true()
--- a/lisa/features/__init__.py
+++ b/lisa/features/__init__.py
@@ -9,6 +9,7 @@ from .disks import (
     DiskStandardSSDLRS,
 )
 from .gpu import Gpu
+from .infiniband import Infiniband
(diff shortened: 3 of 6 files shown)


kamalca:
Draft because there are a few more things TODO:
* ~Make the test only run on HPC images~
* ~Make sure these checks are still valid. This is the v2 logic but these driver checks might not be right because some HPC images fail.~
* ~Use more tools and less greps. Make the test look more like v3 code~

microsoft/testsuites/core/infiniand.py
```diff
@@ -0,0 +1,97 @@
...
+
+        # Ubuntu is used for HPC SKU but manual configuration required.
+        # It's exceptional.
+        if isinstance(node.os, Ubuntu):
+            raise SkippedException(
+                f"{node.os.name} not supported. "
+                "Only HPC machines are supported by this test."
+            )
+
+        if node.execute("test -d /sys/class/infiniband", shell=True).exit_code != 0:
```
LiliDeng: this one is decided by VM size, not image type

microsoft/testsuites/core/infiniand.py
```diff
@@ -0,0 +1,97 @@
...
+        """,
+        priority=2,
+    )
+    def verify_hpc_vm(self, log: Logger, node: Node) -> None:
+
+        # Ubuntu is used for HPC SKU but manual configuration required.
+        # It's exceptional.
+        if isinstance(node.os, Ubuntu):
+            raise SkippedException(
+                f"{node.os.name} not supported. "
```
squirrelsc: Make the exception message is accurate like the comments above. And in this case, if the exception message is clear enough, it doesn't need repeat it in comments.

microsoft/testsuites/core/infiniand.py
```diff
@@ -0,0 +1,97 @@
...
+                "Only HPC machines are supported by this test."
+            )
+
+        if node.execute("test -d /sys/class/infiniband", shell=True).exit_code != 0:
+            raise SkippedException(
```
squirrelsc: Is HPC supported by some special VM sizes? If so, add a feature, and hard code the vm sizes in the Azure feature implementation.

microsoft/testsuites/core/infiniand.py
```diff
@@ -0,0 +1,97 @@
...
+            == 0
+        )
+        is_over_ND = (
+            node.execute(
+                "dmesg | grep hvnd_try_bind_nic", shell=True, sudo=True
+            ).exit_code
+            == 0
+        )
+
+        assert_that(is_over_ND or is_over_SRIOV).described_as(
```
squirrelsc: What happens if both of them exist?

microsoft/testsuites/core/infiniand.py
```diff
@@ -0,0 +1,97 @@
...
+        result = node.execute(
+            "cat /etc/waagent.conf | grep OS.EnableRDMA=y", shell=True
+        )
+        assert_that(result.stdout).described_as(
+            "Found waagent configuration of OS.EnableRDMA=y "
+            "was missing or commented out"
+        ).is_equal_to("OS.EnableRDMA=y")
+        log.debug("Verified waagent config OS.EnableRDMA=y set successfully")
+
+        if is_over_ND:
```
squirrelsc: It looks the logic on ND and SRIOV is very different. It's worth to split them to two test cases.

microsoft/testsuites/core/infiniand.py
kamalca:
``` Bash
# Ubuntu is used for HPC SKU but manual configuration required. It's exceptional.
function is_hpc_vm() {
	GetDistro
	if [ -d /sys/class/infiniband/ ] && [[ $DISTRO != "ubuntu"* ]]; then
```
This is the logic from v2 that I used. The test is skipped for all Ubuntu distros and the existence of the infiniband directory is used to determine whether the machine is HPC. If you have a better way of determining HPC SKU, please let me know.

microsoft/testsuites/core/infiniand.py
kamalca: That comment is from v2. I'd like to know whether skipping Ubuntu is still a good choice for v3.  While Ubuntu is not the recommended distro for HCP machines, it is supported and there are HCP Ubuntu images on the Azure market place. This logic may be outdated.

kamalca:
LISAv2 also has logic that expects `dmesg | grep "IB Infiniband driver"` to produce output for HPC machines with IB over SR-IOV. This is true for regular versions of CentOS (`mlx5_ib: Mellanox Connect-IB Infiniband driver v5.0-0`), but not for the HCP version (or Ubuntu-HPC). Perhaps another driver is being used? I don't know enough about the current state of the technology to update the v2 logic.

kamalca:
All distros I tested now pass.

microsoft/testsuites/core/infiniand.py
squirrelsc: Please double check it. If Ubuntu supports HPC, please remove this logic.

squirrelsc:
> All distros I tested now pass.

Did you get above the question is answered, which is about IB driver?

microsoft/testsuites/core/infiniand.py
LiliDeng: refer https://techcommunity.microsoft.com/t5/azure-compute/azure-hpc-vm-images/ba-p/977094

kamalca:
I added code to check for modules being loaded based on information provided by Long Li.

kamalca:
How can I find a VM SKU / image that is ND based? Most of the HPC machines I see say they support network-acceleration which means they would all be SR-IOV based, correct?

squirrelsc:
> How can I find a VM SKU / image that is ND based? Most of the HPC machines I see say they support network-acceleration which means they would all be SR-IOV based, correct?

What does the "ND" mean? There are a couple of properties in vm size information, it include a GPU count. If it doesn't help, you can query from Azure doc, and add them by names. If there is new vm sizes available, we can update the name list. The NVMe in azure platform is an example to add by vm size family.

sharsonia:
Why is this testsuite in /core/infiniband ?
It shouldn't be inside core folder. You can create another folder for hpc and add IB test cases there.

microsoft/testsuites/hpc/infiniand.py
```diff
@@ -0,0 +1,125 @@
...
+    )
+    def verify_hpc_over_sriov(self, log: Logger, node: Node) -> None:
+
+        if node.execute("test -d /sys/class/infiniband", shell=True).exit_code != 0:
+            raise SkippedException(
+                "/sys/class/infiniband/ not found. "
+                "Only HPC machines are supported by this test."
+            )
+
+        nic = node.features[NetworkInterface]
```
squirrelsc: You can specify the SRIOV feature, and assume sriov exists.

microsoft/testsuites/hpc/infiniand.py
```diff
@@ -0,0 +1,125 @@
...
+        if (
+            node.execute(
+                'lspci | grep "Virtual Function"', shell=True, sudo=True
+            ).exit_code
+            != 0
+            or not nic.is_enabled_sriov()
+        ):
+            raise SkippedException("Inifiniband over SR-IOV was not detected.")
+
+        result = node.execute(
```
squirrelsc: Add this logic in waagent as `is_rdma_enabled`, and call its `get_configuration`. The file is `/lisa/sut_orchestrator/tools.py`.

microsoft/testsuites/hpc/infiniand.py
```diff
@@ -0,0 +1,125 @@
...
+        This test case will
+        1. Determine whether the VM is IB over SR-IOV
+        2. Ensure waagent is configures with OS.EnableRDMA=y
+        3. Check that appropriate drivers are present
+        """,
+        priority=2,
+    )
+    def verify_hpc_over_sriov(self, log: Logger, node: Node) -> None:
+
+        if node.execute("test -d /sys/class/infiniband", shell=True).exit_code != 0:
```
squirrelsc: Add a feature of Infiniband like GPU

kamalca:
According to [Azure Documentation](https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/hpc/configure#non-sr-iov-enabled-vms) I would expect vm sku `Standard_H8`  running marketplace image `openlogic centos-hpc 7.4 latest` to have inifiniband enabled over network direct. However there is no folder `/sys/class/infiniband`, nothing came up for `dmesg | grep hvnd_try_bind_nic`, and there were no mlx/nd modules loaded such as `mlx5_ib` or `hv_networkdirect`

Considering that H series VMs are being discontinued next year and I can't find any other VMs that don't support SR-IOV, I'm questioning whether the ND case is relevant. I'm not sure how to test this case; does it require manual setup?

lisa/features/infiniband.py
```diff
@@ -0,0 +1,33 @@
...
+        lspci = self._node.tools[Lspci]
+        device_list = lspci.get_device_list()
+
+        return any("Virtual Function" in device.device_info for device in device_list)
+
+    # nd stands for network direct
+    # example SKU: Standard_H16mr
+    def is_over_nd(self) -> bool:
+        dmesg = self._node.tools[Dmesg]
+        return "hvnd_try_bind_nic" in dmesg.get_output()
```
squirrelsc: This implementation should be in Azure level, because the hvnd_try_bind_nic is a message from Microsoft drivers.

kamalca: Is this also true for is_over_sriov()?

squirrelsc: Good question. From the implementation of is_over_sriov, it looks like it's general logic. But if it's not, it should be moved into Azure too.

microsoft/testsuites/hpc/infinibandsuit.py
```diff
@@ -0,0 +1,102 @@
...
+
+        infiniband = node.features[Infiniband]
+        if not infiniband.is_over_nd():
+            raise SkippedException("Inifiniband over ND was not detected.")
+
+        waagent = node.tools[Waagent]
+        assert_that(waagent.is_rdma_enabled()).described_as(
+            "Found waagent configuration of OS.EnableRDMA=y "
+            "was missing or commented out"
+        ).is_true
```
squirrelsc: `is_true` doesn't work, because it's a method reference. It's always true. It should be `is_true()`. Recommend to make every assertion failed at least once, when developing. You will see the actual error message, and adjust it.

squirrelsc (approved):

PR #1619, opened 2021-11-11
Add RHEL missing package fail, add warning for nonfatal error code

mcgov:
yum error code 1 means DNF handled an error with installation. Not checking for missing packages could result in the package installation failing silently, since missing packages halt the installation.
This adds a check for missing packages and fails if they are present. Other handled errors result in a warning.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -15,6 +15,7 @@ from lisa.executable import Tool
 from lisa.util import (
     BaseClassMixin,
     LisaException,
+    MissingPackagesException,
     filter_ansi_escape,
     get_matched_str,
     parse_version,
@@ -919,6 +920,17 @@ class Redhat(Fedora):
             command += " --nogpgcheck"
 
         install_result = self._node.execute(command, shell=True, sudo=True)
+        # RedHat will fail package installation is a single missing package is
+        # detected, therefore we check the output to see if we were missing
+        # a package. If so, fail. Otherwise we will warn in verify package result.
+        if install_result.exit_code == 1:
+            missing_packages = []
+            for line in install_result.stdout.splitlines():
+                if line.startswith("No match for argument:"):
+                    package = line.split(":")[1].strip()
+                    missing_packages.append(package)
+            if missing_packages:
+                raise MissingPackagesException(missing_packages)
         self.__verify_package_result(install_result, packages)
 
     def _package_exists(self, package: str, signed: bool = True) -> bool:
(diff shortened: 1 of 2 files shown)


lisa/operating_system.py
```diff
@@ -971,10 +985,11 @@ def _update_packages(self, packages: Optional[List[str]] = None) -> None:
...
 
     def __verify_package_result(self, result: ExecutableResult, packages: Any) -> None:
-        # yum returns exit_code=1 if package is already installed.
-        # We do not want to fail if exit_code=1.
+        # yum returns exit_code=1 if DNF handled an error with installation.
+        # We do not want to fail if exit_code=1, but warn since something may
+        # potentially have gone wrong.
         if result.exit_code == 1:
-            self._log.debug(f"{packages} is/are already installed.")
+            self._log.warn(f"DNF handled error with installation of {packages}")
```
squirrelsc: What's the reason to change it to warn level? The warn level is actually not useful. Nobody check the warn even error level logs. We care about pass/faill test results, and fatal errors on test run.

lisa/operating_system.py
```diff
@@ -919,6 +919,20 @@ def _install_packages(self, packages: List[str], signed: bool = True) -> None:
...
+        # detected, therefore we check the output to see if we were missing
+        # a package. If so, fail. Otherwise we will warn in verify package result.
+        if install_result.exit_code == 1:
+            missing_packages = []
+            for line in install_result.stdout.splitlines():
+                if line.startswith("No match for argument:"):
+                    package = line.split(":")[1].strip()
+                    missing_packages.append(package)
+            if missing_packages:
+                raise LisaException(
```
squirrelsc: If there is special handling in upper layers, it can raise a different type exception, like PackageNotFoundException.

lisa/operating_system.py
mcgov: I wanted to make it easy to search for a potentially suspicious but not neccesarily fatal condition (an error code that returned reporting 'handled by backend'). In testing I learned that this can actually bite you, like with a missing package it results in none of the packages actually being installed.

In this PR I added a condition to hard fail on missing packages since it means you need to rework your test to account for the missing package, but since there are an unknown amount of other conditions people might see it seemed helpful to highlight this with a WARN label rather than just another debug level log. Does that make sense? Or you want me to switch it?

mcgov: eh nevermind I'll just move it to debug

squirrelsc:
@mcgov I provided all comments already.

mcgov:
derp, forgot to push

squirrelsc (approved):

PR #1623, opened 2021-11-12
Receive side dpdk failsafe test

mcgov:
Sets up a send/receive environment and tests dpdk failsafe during traffic on the receiver side.

refactors some code that generated the testkit-testpmd command pair into a helper function.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -10,6 +10,7 @@ from assertpy import assert_that, fail
 
 from lisa import (
     Environment,
+    LisaException,
     Logger,
     Node,
     RemoteNode,
@@ -21,14 +22,15 @@ from lisa import (
 from lisa.features import NetworkInterface, Sriov
 from lisa.nic import NicInfo, Nics
 from lisa.testsuite import simple_requirement
-from lisa.tools import Echo, Git, Lspci, Make, Mount
-from lisa.util import constants
+from lisa.tools import Dmesg, Echo, Git, Lspci, Make, Mount
+from lisa.util import constants, perf_timer
 from lisa.util.parallel import Task, TaskManager
 from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
 
 VDEV_TYPE = "net_vdev_netvsc"
 MAX_RING_PING_LIMIT_NS = 200000
 DPDK_STABLE_GIT = "http://dpdk.org/git/dpdk-stable"
+DPDK_VF_REMOVAL_MAX_TEST_TIME = 60 * 10
 
 
 @TestSuiteMetadata(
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -85,6 +85,31 @@ def verify_dpdk_build_failsafe(
...
+    def verify_dpdk_sriov_rescind_failover_receiver(
+        self, environment: Environment, log: Logger, variables: Dict[str, Any]
+    ) -> None:
+
+        test_kits = _init_nodes_concurrent(environment, log, variables, "failsafe")
+        sender, receiver = test_kits
+
+        kit_cmd_pairs = generate_send_receive_run_info("failsafe", sender, receiver)
+
+        _run_testpmd_concurrent(kit_cmd_pairs, 60 * 5, log, rescind_sriov=True)
```
squirrelsc: Does it mean this test case will run 300 seconds? Can it be quicker, like 30 sec even less?

mcgov: Nope, the device removal and re-attach can each take a full minute to hit the VM and get processed. Maybe we can cut it to 4 minutes instead of 5.

mcgov: Hey! we can cut it to 3 :)

squirrelsc: The hard code time either not stable, or take longer time in usual. It's better to check some signal, and start next step, once some condition is met. It means we can set some timeout here like 10 minutes, but in most cases, it will completed in 1 minute.

mcgov: I can do that, will ping you when it's ready

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -363,6 +375,66 @@ def __init__(
...
+            search_pattern = self.vf_hotplug_regex
+        else:
+            raise LisaException(
+                "Unknown search pattern specified in "
+                "DpdkTestResources:wait_for_dmesg_output"
+            )
+
+        self.node.log.info(search_pattern.pattern)
+        time_accumulator = 0.0
+        while time_accumulator < timeout:
```
squirrelsc: The `process.py` has a simpler pattern to check it with lisa's timer.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -363,6 +375,66 @@ def __init__(
...
+            start = time.perf_counter()
+            output = self.dmesg.get_output(force_run=True)
+            stop = time.perf_counter()
+            if search_pattern.search(output.replace(self._last_dmesg, "")):
+                self._last_dmesg = output  # save old output to filter next time
+                self.node.log.info(
+                    f"Found VF hotplug info after {time_accumulator} seconds"
+                )
+                return True
+            time_accumulator += stop - start
```
squirrelsc: There can be a interval sleep like 1 sec, it avoids to run the function too fast.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -485,16 +552,41 @@ def _run_node_init() -> Tuple[DpdkTestResources, str]:
...
+            if not node_resources.wait_for_dmesg_output("AN_REENABLE", seconds // 2):
+                fail(
+                    "Accelerated Network re-enable not found "
+                    f" in dmesg before timeout for node  {node_resources.node.name}"
+                )
+
+        time.sleep(15)  # let testpmd run with sriov re-enabled
+
+        # kill the commands to collect the output early and terminate before timeout
+        for node_resources in test_kits:
```
squirrelsc, line 586: Besides this, considering to kill commands in `after_case`. So they will be called even the test case is failed. Refer to Lili's PR, it uses this method to cleanup the system.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Do you mean perf_timer?

squirrelsc: Yes, it can count the accumulated value too.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Oh good thought, I can add that in another PR once this merges

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: perf_timer.Timer() looks good :)

mcgov (commented):
update for chi's comments

squirrelsc (dismissed):

squirrelsc (approved):

PR #1624, opened 2021-11-13
Add basic nested vm testcase

dsrivastavv:
Add nested VM testcase.

Tested on Fedora, Suse and Debian

### DIFF
--- a/lisa/tools/df.py
+++ b/lisa/tools/df.py
@@ -2,37 +2,11 @@
 # Licensed under the MIT license.
 
 import re
-from dataclasses import dataclass
 from typing import List, Optional
 
 from lisa.executable import Tool
-from lisa.util import find_patterns_in_lines
-
-
-@dataclass
-class PartitionInfo(Tool):
-    partition_name: str = ""
-    blocks: int = 0
-    used: int = 0
-    total_size: int = 0
-    percentage_used: int = 0
-    mountpoint: str = ""
-
-    def __init__(
-        self,
-        partition_name: str,
-        blocks: int,
-        used: int,
-        total_size: int,
-        percentage_used: int,
-        mountpoint: str,
-    ):
-        self.partition_name = partition_name
-        self.blocks = blocks
-        self.used = used
-        self.total_size = total_size
-        self.percentage_used = percentage_used
-        self.mountpoint = mountpoint
+from lisa.tools.lsblk import PartitionInfo
+from lisa.util import find_patterns_groups_in_lines
 
 
 class Df(Tool):
@@ -49,33 +23,31 @@ class Df(Tool):
     def can_install(self) -> bool:
         return True
 
-    def get_partition_information(self) -> List[PartitionInfo]:
+    def get_partitions(self, force_run: bool = False) -> List[PartitionInfo]:
         # run df and parse the output
         # The output is in the following format:
         # <Filesystem> <1K-blocks> <Used> <Available> <Use%> <Mounted on>
         # Note : <Used> and <Available> are in 1-k blocks, not in bytes
-        output = self.run(sudo=True).stdout
+        output = self.run(sudo=True, force_run=force_run).stdout
         partition_info = []
-        df_entries = find_patterns_in_lines(output, [self._DF_ENTRY_REGEX])[0]
+        df_entries = find_patterns_groups_in_lines(output, [self._DF_ENTRY_REGEX])[0]
         for df_entry in df_entries:
             partition_info.append(
                 PartitionInfo(
-                    partition_name=df_entry[0],
-                    blocks=int(df_entry[1]),
-                    used=int(df_entry[2]),
-                    total_size=int(df_entry[3]),
-                    percentage_used=int(df_entry[4]),
-                    mountpoint=df_entry[5],
+                    name=df_entry["name"],
+                    mountpoint=df_entry["mountpoint"],
+                    available_blocks=int(df_entry["blocks"]),
+                    used_blocks=int(df_entry["used"]),
+                    total_blocks=int(df_entry["total"]),
+                    percentage_blocks_used=int(df_entry["percentage_use"]),
                 )
             )
         return partition_info
 
-    def get_partition_with_mountpoint(
-        self, partition_mountpoint: str
-    ) -> Optional[PartitionInfo]:
+    def get_partition_by_mountpoint(self, mountpoint: str) -> Optional[PartitionInfo]:
         # get `df` entry for the partition with the given mountpoint
-        df_partition_info = self.get_partition_information()
+        df_partition_info = self.get_partitions()
         for partition in df_partition_info:
-            if partition.mountpoint == partition_mountpoint:
+            if partition.mountpoint == mountpoint:
                 return partition
         return None
--- a/lisa/tools/df.py
+++ b/lisa/tools/df.py
@@ -1,0 +1,81 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from dataclasses import dataclass
+from typing import List, Optional
+
+from lisa.executable import Tool
+from lisa.util import find_patterns_in_lines
+
+
+@dataclass
+class PartitionInfo(Tool):
+    partition_name: str = ""
+    blocks: int = 0
+    used: int = 0
+    total_size: int = 0
+    percentage_used: int = 0
+    mountpoint: str = ""
+
+    def __init__(
+        self,
+        partition_name: str,
+        blocks: int,
+        used: int,
+        total_size: int,
+        percentage_used: int,
+        mountpoint: str,
+    ):
+        self.partition_name = partition_name
+        self.blocks = blocks
+        self.used = used
+        self.total_size = total_size
+        self.percentage_used = percentage_used
+        self.mountpoint = mountpoint
+
+
+class Df(Tool):
+    # devtmpfs         4071644       0   4071644   0% /dev
+    _DF_ENTRY_REGEX = re.compile(
+        r"(?P<name>\S+)\s+(?P<blocks>\d+)\s+(?P<used>\d+)\s+"
+        r"(?P<total>\d+)\s+(?P<percentage_use>\d+)%\s+(?P<mountpoint>\S+)"
+    )
+
+    @property
+    def command(self) -> str:
+        return "df"
+
+    def can_install(self) -> bool:
+        return True
+
+    def get_partition_information(self) -> List[PartitionInfo]:
+        # run df and parse the output
+        # The output is in the following format:
+        # <Filesystem> <1K-blocks> <Used> <Available> <Use%> <Mounted on>
+        # Note : <Used> and <Available> are in 1-k blocks, not in bytes
+        output = self.run(sudo=True).stdout
+        partition_info = []
+        df_entries = find_patterns_in_lines(output, [self._DF_ENTRY_REGEX])[0]
+        for df_entry in df_entries:
+            partition_info.append(
+                PartitionInfo(
+                    partition_name=df_entry[0],
+                    blocks=int(df_entry[1]),
+                    used=int(df_entry[2]),
+                    total_size=int(df_entry[3]),
+                    percentage_used=int(df_entry[4]),
+                    mountpoint=df_entry[5],
+                )
+            )
+        return partition_info
+
+    def get_partition_with_mountpoint(
+        self, partition_mountpoint: str
+    ) -> Optional[PartitionInfo]:
+        # get `df` entry for the partition with the given mountpoint
+        df_partition_info = self.get_partition_information()
+        for partition in df_partition_info:
+            if partition.mountpoint == partition_mountpoint:
+                return partition
+        return None
--- a/lisa/tools/lsblk.py
+++ b/lisa/tools/lsblk.py
@@ -1,10 +1,51 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import re
+from dataclasses import dataclass
+from typing import List
+
 from lisa.executable import Tool
+from lisa.util import find_patterns_in_lines
+
+
+@dataclass
+class PartitionInfo(Tool):
+    name: str = ""
+    size: int = 0
+    type: str = ""
+    mountpoint: str = ""
+
+    def __init__(self, name: str, size: int, type: str, mountpoint: str):
+        self.name = name
+        self.size = size
+        self.type = type
+        self.mountpoint = mountpoint
 
 
 class Lsblk(Tool):
+    # NAME="loop2" SIZE="34017280" TYPE="loop" MOUNTPOINT="/snap/snapd/13640"
+    _LSBLK_ENTRY_REGEX = re.compile(
+        r'NAME="(?P<name>\S+)"\s+SIZE="(?P<size>\d+)"\s+'
+        r'TYPE="(?P<type>\S+)"\s+MOUNTPOINT="(?P<mountpoint>\S+)"'
+    )
+
     @property
     def command(self) -> str:
         return "lsblk"
+
+    def get_partition_information(self) -> List[PartitionInfo]:
+        # parse output of lsblk
+        output = self.run("-b -P -o NAME,SIZE,TYPE,MOUNTPOINT", sudo=True).stdout
+        partition_info = []
+        lsblk_entries = find_patterns_in_lines(output, [self._LSBLK_ENTRY_REGEX])[0]
+        for lsblk_entry in lsblk_entries:
+            partition_info.append(
+                PartitionInfo(
+                    name=lsblk_entry[0],
+                    size=int(lsblk_entry[1]),
+                    type=lsblk_entry[2],
+                    mountpoint=lsblk_entry[3],
+                )
+            )
+        return partition_info
--- a/lisa/tools/lsblk.py
+++ b/lisa/tools/lsblk.py
@@ -6,21 +6,39 @@ from dataclasses import dataclass
 from typing import List
 
 from lisa.executable import Tool
-from lisa.util import find_patterns_in_lines
+from lisa.util import find_patterns_groups_in_lines
 
 
 @dataclass
-class PartitionInfo(Tool):
+class PartitionInfo(object):
     name: str = ""
+    mountpoint: str = ""
     size: int = 0
     type: str = ""
-    mountpoint: str = ""
+    available_blocks: int = 0
+    used_blocks: int = 0
+    total_blocks: int = 0
+    percentage_blocks_used: int = 0
 
-    def __init__(self, name: str, size: int, type: str, mountpoint: str):
+    def __init__(
+        self,
+        name: str,
+        mountpoint: str,
+        size: int = 0,
+        type: str = "",
+        available_blocks: int = 0,
+        used_blocks: int = 0,
+        total_blocks: int = 0,
+        percentage_blocks_used: int = 0,
+    ):
         self.name = name
+        self.mountpoint = mountpoint
         self.size = size
         self.type = type
-        self.mountpoint = mountpoint
+        self.available_blocks = available_blocks
+        self.used_blocks = used_blocks
+        self.total_blocks = total_blocks
+        self.percentage_blocks_used = percentage_blocks_used
 
 
 class Lsblk(Tool):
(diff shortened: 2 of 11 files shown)


lisa/tools/aria2.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Posix)
+        self.node.os.install_packages("aria2")
+        return self._check_exists()
+
+    def download_file(self, url: str, file_name: str) -> None:
```
squirrelsc: Is it used to download file? How about wget?

lisa/tools/qemu.py
```diff
@@ -0,0 +1,70 @@
...
+            # verify that kvm is enabled
+            self._is_kvm_successfully_enabled()
+        else:
+            raise NotImplementedError(f"{self.node.os} is not supported")
+
+        return self._check_exists()
+
+    def _is_kvm_successfully_enabled(self) -> None:
+        lsmod_output = self.node.tools[Lsmod].run().stdout
+        is_kvm_successfully_enabled = "kvm_intel" in lsmod_output
```
LiliDeng: we need consider not only intel but also other architecture.

dsrivastavv: Thanks Lili! I will modify the code to include amd based information as well.

lisa/tools/aria2.py
dsrivastavv: It is used for downloading file. This is the default tool used in v2 probably because nested image file size can be large and this allows simultaneous downloads.

squirrelsc: ok. Change the method name to `download`.

lisa/tools/sshpass.py
```diff
@@ -0,0 +1,37 @@
...
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Posix)
+        self.node.os.install_packages("sshpass")
+        return self._check_exists()
+
+    def copy_file(
```
squirrelsc: `copy`

microsoft/testsuites/nested/nested.py
```diff
@@ -0,0 +1,105 @@
...
+        guest_password: str,
+        guest_port: int,
+        guest_image_url: str,
+    ) -> RemoteNode:
+        # start nested vm
+        host.tools[Qemu].create_nested_vm(guest_port, guest_image_url)
+
+        # wait for the VM to boot
+        # ToDo: Find a more robust solution
+        time.sleep(60)
```
squirrelsc: The following steps can wait it start. Don't put any hardcode sleep in the code.

microsoft/testsuites/nested/nested.py
```diff
@@ -0,0 +1,105 @@
...
+    description="""
+    This test suite is used to run nested vm related tests.
+    """,
+)
+class Nested(TestSuite):
+    NESTED_VM_TEST_FILE_NAME = "message.txt"
+    NESTED_VM_TEST_FILE_CONTENT = "Message from L1 vm!!"
+    NESTED_VM_TEST_PUBLIC_FILE_URL = "http://www.github.com"
+
+    def get_l2_vm(
```
squirrelsc: `_connect_nested_vm`? It won't be shared in other test suites, the private method is good.

microsoft/testsuites/nested/nested.py
```diff
@@ -0,0 +1,105 @@
...
+    def verify_nested_basic(self, node: RemoteNode, variables: Dict[str, Any]) -> None:
+        nested_image_username = variables.get("nested_image_username", "")
+        nested_image_password = variables.get("nested_image_password", "")
+        nested_image_port = variables.get("nested_image_port", 60024)
+        nested_image_url = variables.get("nested_image_url", "")
+
+        assert nested_image_username
+        assert nested_image_password
+        assert nested_image_port
+        assert nested_image_url
```
squirrelsc: Use skip exception to skip this test case, if no such information provided.

microsoft/testsuites/nested/nested.py
```diff
@@ -0,0 +1,105 @@
...
+            nested_image_port,
+            nested_image_url,
+        )
+
+        # verify file is correctly copied from L1 VM to L2 VM
+        node.tools[Echo].write_to_file(
+            self.NESTED_VM_TEST_FILE_CONTENT,
+            node.get_pure_path(self.NESTED_VM_TEST_FILE_NAME),
+        )
+        node.tools[Sshpass].copy_file(
```
squirrelsc: Is it possible to check the exit code?

microsoft/testsuites/nested/nested.py
```diff
@@ -0,0 +1,105 @@
...
+            nested_image_username,
+            nested_image_password,
+            nested_image_port,
+        )
+
+        uploaded_message = l2_vm.tools[Cat].read_from_file(
+            self.NESTED_VM_TEST_FILE_NAME
+        )
+        assert_that(
+            uploaded_message, "message uploaded to L2 vm from L1 does not match"
```
squirrelsc: the "message" is confusing word here. It's about failed to check file content.

lisa/tools/qemu.py
```diff
@@ -0,0 +1,74 @@
...
+        # verify that kvm module is loaded
+        lsmod_output = self.node.tools[Lsmod].run().stdout
+        is_kvm_successfully_enabled = (
+            "kvm_intel" in lsmod_output or "kvm_amd" in lsmod_output
+        )
+        assert_that(
+            is_kvm_successfully_enabled, f"KVM could not be enabled : {lsmod_output}"
+        ).is_true()
+
+    def create_nested_vm(
```
squirrelsc: `create_vm`. The tool is not related to nest scenario itself.

squirrelsc: And move the public method above private methods.

lisa/tools/qemu.py
```diff
@@ -0,0 +1,74 @@
...
+        self,
+        port: int,
+        guest_image_path: str,
+    ) -> None:
+        # start nested vm
+        result = self.run(
```
squirrelsc: Add comments to explain the parameters.

microsoft/testsuites/nested/nested.py
```diff
@@ -0,0 +1,161 @@
...
+    """,
+)
+class Nested(TestSuite):
+    NESTED_VM_IMAGE_NAME = "image.qcow2"
+    NESTED_VM_TEST_FILE_NAME = "message.txt"
+    NESTED_VM_TEST_FILE_CONTENT = "Message from L1 vm!!"
+    NESTED_VM_TEST_PUBLIC_FILE_URL = "http://www.github.com"
+    NESTED_VM_REQUIRED_DISK_SIZE_IN_GB = 6
+
+    def _check_partition_capacity(
```
squirrelsc: move private methods below the public test methods.

lisa/tools/aria2.py
dsrivastavv: I moved to using Wget so this comment is not applicable now.

squirrelsc: Is the wget fast enough?

lisa/tools/qemu.py
```diff
@@ -0,0 +1,77 @@
...
+
+from lisa.executable import Tool
+from lisa.operating_system import Fedora, Posix
+from lisa.tools import Lsmod
+
+
+class Qemu(Tool):
+
+    QEMU_INSTALL_LOCATIONS = ["qemu-system-x86_64", "qemu-kvm", "/usr/libexec/qemu-kvm"]
+    COMMAND = "qemu-system-x86_64"
```
squirrelsc: It looks used as an instance variable, not a constant or class level variable. So not to define it to class level. The class level variable will share among classes. If we run this test cases on different distros together, it may have problems.

lisa/tools/df.py
```diff
@@ -0,0 +1,81 @@
...
+import re
+from dataclasses import dataclass
+from typing import List, Optional
+
+from lisa.executable import Tool
+from lisa.util import find_patterns_in_lines
+
+
+@dataclass
+class PartitionInfo(Tool):
```
squirrelsc: It's good to have this class, and make sure it's reusable in lsblk, fdisk.

lisa/tools/df.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    @property
+    def command(self) -> str:
+        return "df"
+
+    def can_install(self) -> bool:
+        return True
+
+    def get_partition_information(self) -> List[PartitionInfo]:
```
squirrelsc: 1. `get_partitions`, The "information" doesn't bring more information here.
2. Support "force_run", so it can be refreshed after disk is changed.

lisa/tools/df.py
```diff
@@ -0,0 +1,81 @@
...
+                    blocks=int(df_entry[1]),
+                    used=int(df_entry[2]),
+                    total_size=int(df_entry[3]),
+                    percentage_used=int(df_entry[4]),
+                    mountpoint=df_entry[5],
+                )
+            )
+        return partition_info
+
+    def get_partition_with_mountpoint(
```
squirrelsc: The name should be "get_partition_by_mountpoint".

lisa/tools/df.py
```diff
@@ -0,0 +1,81 @@
...
+        self, partition_mountpoint: str
```
squirrelsc: `mountpoint`

lisa/tools/df.py
dsrivastavv: I didn't get this comment. Can you please provide some more context on this?

lisa/tools/lsblk.py
```diff
@@ -1,10 +1,51 @@
...
+import re
+from dataclasses import dataclass
+from typing import List
+
 from lisa.executable import Tool
+from lisa.util import find_patterns_in_lines
+
+
+@dataclass
+class PartitionInfo(Tool):
```
squirrelsc: Use the class from df.

lisa/tools/df.py
squirrelsc: I see there is a same name class `PartitionInfo` in lsblk. They can be the same one. It doesn't need to define different PartitionInfo class in each tool.

lisa/tools/lsblk.py
```diff
@@ -1,10 +1,51 @@
...
+
     @property
     def command(self) -> str:
         return "lsblk"
+
+    def get_partition_information(self) -> List[PartitionInfo]:
+        # parse output of lsblk
+        output = self.run("-b -P -o NAME,SIZE,TYPE,MOUNTPOINT", sudo=True).stdout
+        partition_info = []
+        lsblk_entries = find_patterns_in_lines(output, [self._LSBLK_ENTRY_REGEX])[0]
```
squirrelsc: try `find_patterns_groups_in_lines`. it supports named groups.

lisa/tools/qemu.py
```diff
@@ -0,0 +1,77 @@
...
+
+        result = self.run(
+            f"-smp 2 -m 2048 -hda {guest_image_path} "
+            "-device e1000,netdev=user.0 "
+            f"-netdev user,id=user.0,hostfwd=tcp::{port}-:22 "
+            "-enable-kvm -display none -daemonize",
+            sudo=True,
+            shell=True,
+        )
+        assert_that(
```
squirrelsc: It can be in the `run` command. So it doesn't need to be a separated command.

lisa/tools/qemu.py
```diff
@@ -0,0 +1,77 @@
...
+            self.node.execute("firewall-cmd --reload", sudo=True)
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Posix)
+
+        # install qemu
+        self.node.os.install_packages("qemu-kvm")
+
+        # verify that kvm is enabled
+        self._is_kvm_successfully_enabled()
```
squirrelsc, line 71: What's the reason to check by this way? If the command exists from one of following line, it should be ok, is it right?

lisa/tools/sshpass.py
```diff
@@ -0,0 +1,74 @@
...
+        self.node.tools[Tar].extract(download_file_path, str(download_folder_path))
+        source_code_folder_path = download_folder_path.joinpath("sshpass-1.06")
+
+        # build sshpass
+        self.node.execute(
+            "./configure --prefix=/usr/",
+            cwd=source_code_folder_path,
+            shell=True,
+        )
+        self.node.execute("make install", cwd=source_code_folder_path, sudo=True)
```
squirrelsc: There is a make tool already, it will make sure the gcc exists for building.

microsoft/testsuites/nested/nested.py
```diff
@@ -0,0 +1,161 @@
...
+        if not is_virtualization_enabled:
+            raise SkippedException("Virtualization is not enabled in hardware")
+
+        # verify os compatibility
+        if not (
+            isinstance(node.os, Debian)
+            or isinstance(node.os, Fedora)
+            or isinstance(node.os, Suse)
+        ):
+            raise SkippedException(f"{node.os} is not supported")
```
squirrelsc: Specify what's supported.

lisa/tools/qemu.py
dsrivastavv: This is to ensure that kvm module has been successfully loaded. The command can exist but checking that isn't enough to ensure that kvm is running.

lisa/tools/lsblk.py
dsrivastavv: They both contain different kind of information. Is your intent here to merge both these classes into one class and fill only the required entries in the class depending on the command?

lisa/tools/qemu.py
dsrivastavv: I didn't get this comment. Can you please elaborate?

lisa/tools/df.py
dsrivastavv: The field are different for both of these. Both df and lsblk give different kind of info for the disk

dsrivastavv: How about we move ` get_partition_information(self)` to disk features and populate info from both lsblk and df?

lisa/tools/lsblk.py
dsrivastavv: How about we move ` get_partition_information(self)` to disk features and populate info from both lsblk and df?

lisa/tools/df.py
squirrelsc: Disk is different with partitions. We can leave them in lsblk and df themselves. Their information is overlap. We can define a structure to include information from both of them, and fill in different information from different command.

lisa/tools/qemu.py
squirrelsc: the `run` command has two arguments, `expected_exit_code` and `expected_exit_code_failure_message`. Use them, so it doesn't to be written in a separated assert_that.

lisa/tools/qemu.py
dsrivastavv: This needs to be defined in class level variable because we can't create an init method to define a instance specific variable. This is because init method requires Node object which creates a circular import. But from what I understand about global variables is that a copy is maintained for each class. Please feel free to suggest any alternate solution.

lisa/tools/aria2.py
dsrivastavv: It's a bit slower(~30sec extra time) than aria but seems to work fine. Another issue with aria is that it's not supported on sles, so we would have to use wget on sles.

lisa/tools/qemu.py
squirrelsc: 1. In Python, it doesn't need to declare a variable in `__init__`. It can be created anytime, but just make sure it's used after it's set.
2. In the tool, use `_initialize` to initialize for an object.
3. If there is circular import error on declaration. Try 1) use `if TYPING_CHECKING:` to limit the import statement in typing check phase only. 2) quote it like `def method_name(self, node: "Node") -> None:`.

lisa/tools/aria2.py
squirrelsc: We can go with wget. But in general, if different distros have different tools, we can create a virtual tool like `service`. It can wrap different tools underlying, and provide the same interface.

lisa/tools/qemu.py
dsrivastavv: _initialize seems to work. Thanks!

lisa/tools/qemu.py
squirrelsc: Are the comments added for parameters of the cmd?

lisa/tools/qemu.py
```diff
@@ -0,0 +1,80 @@
...
+        # update firewall rules
+        # https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/configuring_and_managing_networking/using-and-configuring-firewalld_configuring-and-managing-networking # noqa E501
+        if isinstance(self.node.os, Fedora):
+            self.node.execute(
+                f"firewall-cmd --permanent --add-port={port}/tcp", sudo=True
+            )
+            self.node.execute("firewall-cmd --reload", sudo=True)
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.qemu_command = "qemu-system-x86_64"
```
squirrelsc: Add underscore to mark it as a private variable.

lisa/tools/df.py
squirrelsc: Will it be merged in next push? And it doesn't need to inherit from the `Tool`.

lisa/tools/sshpass.py
```diff
@@ -0,0 +1,72 @@
...
+        self.node.tools[Tar].extract(download_file_path, str(download_folder_path))
+        source_code_folder_path = download_folder_path.joinpath("sshpass-1.06")
+
+        # build sshpass
+        self.node.execute(
+            "./configure --prefix=/usr/",
+            cwd=source_code_folder_path,
+            shell=True,
+        )
+        self.node.tools[Make].make("install", cwd=source_code_folder_path, sudo=True)
```
squirrelsc: try `Make.make_install`?

lisa/tools/lsblk.py
squirrelsc: see comments in `df`. It should be the same PartitionInformation class, and the disk/partition are not same level things. Let's keep partition is out of disks for now.

squirrelsc (approved):

LiliDeng:
@somil55 @squirrelsc We need add a Feature 'NestedVirtualization', from current run, by default, this case will be skipped after checking the VM supports nested virtualization. 
Currently some VM skus like Dv3, Ev3, support nested, but I found the API capability doesn't have this, we may think about it. Maybe hardcode for now?

squirrelsc:
> @somil55 @squirrelsc We need add a Feature 'NestedVirtualization', from current run, by default, this case will be skipped after checking the VM supports nested virtualization. Currently some VM skus like Dv3, Ev3, support nested, but I found the API capability doesn't have this, we may think about it. Maybe hardcode for now?

where is the document to describe which one supports nested?

LiliDeng:
> > @somil55 @squirrelsc We need add a Feature 'NestedVirtualization', from current run, by default, this case will be skipped after checking the VM supports nested virtualization. Currently some VM skus like Dv3, Ev3, support nested, but I found the API capability doesn't have this, we may think about it. Maybe hardcode for now?
> 
> where is the document to describe which one supports nested?

https://azure.microsoft.com/es-es/blog/nested-virtualization-in-azure/

image

from
https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series

squirrelsc:
> > > @somil55 @squirrelsc We need add a Feature 'NestedVirtualization', from current run, by default, this case will be skipped after checking the VM supports nested virtualization. Currently some VM skus like Dv3, Ev3, support nested, but I found the API capability doesn't have this, we may think about it. Maybe hardcode for now?
> > 
> > 
> > where is the document to describe which one supports nested?
> 
> https://azure.microsoft.com/es-es/blog/nested-virtualization-in-azure/
> 
> image
> 
> from https://docs.microsoft.com/en-us/azure/virtual-machines/dv3-dsv3-series

If so, we should use a feature to make sure the nested test case run on supported vms only. @somil55, what's the behavior to run kvm nested case on unsupported vm sizes?

dsrivastavv:
@LiliDeng @squirrelsc We do not use the azure API to check that the VM supports nested virtualization. Instead we look into the output of `lscpu` and check if `vt-x` label is present. Do you think that this check will fail on DV3 vm's?

squirrelsc:
They are two layers. With checking by vm size before deployment, there are a couple of advantages. The feature matches before an environment deployed, so it can cover more in some scenarios.
For example, when we input a distro to verify, each test case will try to find a matched vm size. So the nested test case can be run on a supported vm size. If there is no nested feature, the default vm size (ds2?) will be matched, and it doesn't support nested, so the test case will be skipped.

dsrivastavv:
@squirrelsc Sure! I will add this feature. I have created bug #1657 to track this

PR #1626, opened 2021-11-15
Make availability_set as optional when deploy vm using template on Azure.

LiliDeng:
see commits message for details

### DIFF
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -57,6 +57,13 @@
                 "description": "created subnet count"
             }
         },
+        "use_availability_sets": {
+            "type": "bool",
+            "defaultValue": false,
+            "metadata": {
+                "description": "flag to create availability set"
+            }
+        },
         "availability_set_tags": {
             "type": "object",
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -74,6 +81,13 @@
             "metadata": {
                 "description": "data disk array."
             }
+        },
+        "tip_session_id": {
```
squirrelsc: What's the reason to put tip session here? The current way works well. We shouldn't disclose it.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1009,6 +1015,11 @@ def _create_deployment_parameters(
         arm_parameters.storage_name = get_storage_account_name(
             self.subscription_id, arm_parameters.location
         )
+        if (
+            self._azure_runbook.availability_set_properties
```
squirrelsc, line 1013: Because we hope to disable AS by default, remove the initialize of availability_set_properties. So it's empty by default.

squirrelsc (approved):

PR #1628, opened 2021-11-17
Add net ifup ifdown test

sharsonia:


lisa/features/network_interface.py
```diff
@@ -32,6 +33,10 @@ def attach_nics(
     ) -> None:
         raise NotImplementedError
 
+    def reload_netvsc(self) -> None:
```
squirrelsc: It's not a shared logic, which can across platforms. What's the reason needed? Does it mean to reload a network interface? If so, add interface here, and move the implementation to Azure.

lisa/nic.py
```diff
@@ -181,11 +181,25 @@ def bind(self, nic: NicInfo, driver_module: str) -> None:
             sudo=True,
         )
 
+    def change_nic_state(self, nic_name: str, state: str) -> None:
```
squirrelsc: What can stats be changed? If it's a limited set, it's better to have an enumeration.

lisa/features/network_interface.py
sharsonia: reload_netvsc() intends to unload and load again the netvsc driver. We would need to do it for few tests cases both for hyper-v and azure.

lisa/nic.py
sharsonia: ok sure, it can only be 2 states -  up and down

lisa/features/network_interface.py
squirrelsc: The netvsc is microsoft specified. Let's move it to azure firstly. If something need to shared between Azure and Hyper-V, we can move it, when implementing Hyper-V

lisa/nic.py
```diff
@@ -257,7 +271,7 @@ def _get_node_nic_info(self) -> None:
             f"find any nics attached to {self._node.name}."
         ).is_greater_than(0)
 
-    def _get_default_nic(self) -> str:
+    def _get_default_nic(self) -> Tuple[str, str]:
```
squirrelsc: I don't prefer to return multiple values, it's hard to use. But if it's a short cut, please add comments to explain the meaning of each returned value.

lisa/tools/modprobe.py
```diff
@@ -69,3 +69,20 @@ def load(
             expected_exit_code=0,
             expected_exit_code_failure_message=f"Fail to load module {mod_name}",
         )
+
+    def remove_and_load(
```
squirrelsc: how about call it `reload`?

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,119 @@
...
+        if isinstance(node.os, CoreOs):
+            config_path = f"/usr/boot/config-{kernel_ver}"
+
+        netvsc_builtin_result = node.execute(
+            f"grep CONFIG_HYPERV_NET=y {config_path}", shell=True
+        )
+        if netvsc_builtin_result.exit_code == 0:
+            SkippedException("Skipping test since hv_netvsc module is built-in")
+
+    def _ping_check(self, node: Node) -> None:
```
squirrelsc: This step looks not necessary. If the VM cannot reach the internet, we are not able to connect from SSH.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,119 @@
...
+            if not node_nic_info.default_nic:
+                # Add default route if missing after running ip link down/up
+                node.execute(f"ip route add {default_route}", shell=True, sudo=True)
+
+            if not node_nic_info.nics[default_nic].ip_addr:
+                node.execute("kill $(pidof dhclient)", shell=True, sudo=True)
+                dhclient = node.tools[Dhclient]
+                dhclient.release()
+                dhclient.renew()
+                sleep(15)
```
squirrelsc: Don't put hardcode sleep, use a condition, and check it with interval like 0.5 or 1 second.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,119 @@
...
+
+            self._ping_check
```
squirrelsc: It's not called actually, it's the method signature.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,119 @@
...
+            dhclient.release()
+            dhclient.renew()
+            sleep(15)
+
+        self._ping_check
+
+        test_count = 0
+        while test_count < self.NET_IFUP_IFDOWN_TEST_COUNT:
+            self.__log.info(f"NET IFUP IFDOWN Test Iteration : {test_count}")
+            node_nic_info.change_nic_state(default_nic, "down")
```
squirrelsc: Can it be brought back, if it's down? Does it need any extra validation after down?

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,119 @@
...
+            dhclient = node.tools[Dhclient]
+            dhclient.release()
+            dhclient.renew()
+            sleep(15)
+
+        self._ping_check
+
+        test_count = 0
+        while test_count < self.NET_IFUP_IFDOWN_TEST_COUNT:
+            self.__log.info(f"NET IFUP IFDOWN Test Iteration : {test_count}")
```
squirrelsc: Don't use the logger in test suite, use `log` from the test method.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,119 @@
...
+            node.execute("kill $(pidof dhclient)", shell=True, sudo=True)
+            dhclient = node.tools[Dhclient]
+            dhclient.release()
+            dhclient.renew()
+            sleep(15)
+
+        self._ping_check
+
+        test_count = 0
+        while test_count < self.NET_IFUP_IFDOWN_TEST_COUNT:
```
squirrelsc, line 82: The logic is similar as above, consider to merge shared steps.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,119 @@
...
+        kernel_ver = uname_tool.get_linux_information().kernel_version
+        config_path = f"/boot/config-{kernel_ver}"
+        if isinstance(node.os, CoreOs):
+            config_path = f"/usr/boot/config-{kernel_ver}"
+
+        netvsc_builtin_result = node.execute(
+            f"grep CONFIG_HYPERV_NET=y {config_path}", shell=True
+        )
+        if netvsc_builtin_result.exit_code == 0:
+            SkippedException("Skipping test since hv_netvsc module is built-in")
```
squirrelsc, line 133: Does it mean the test case only run with LIS package?

lisa/util/constants.py
```diff
@@ -134,3 +134,7 @@
 DEVICE_TYPE_SRIOV = "SRIOV"
 DEVICE_TYPE_NVME = "NVME"
 DEVICE_TYPE_GPU = "GPU"
+
+
+# Remote Server address
+REMOTE_SERVER = "8.8.4.4"
```
squirrelsc: The name is confusing. How about `PINGABLE_INTERNET_SERVER`? And you can add some comments to say it's used to check internet connection

microsoft/testsuites/network/netinterface.py
sharsonia: ok will change here.  BTW this sleep time is for waiting for new IP address from dhclient which we know usually have a timeout of 10-15 seconds. May be I can have a check every 2 seconds or so.

microsoft/testsuites/network/netinterface.py
LiliDeng: I don't think so, it only means the test can be executed when hv_netvsc is not built-in module.

microsoft/testsuites/network/netinterface.py
LiliDeng: if you take down eth0 then ssh will be disconnected, you may consider to bring down and up in one command.

lisa/nic.py
sharsonia: hmm returning multiple values to avoid executing same command on node again. Since we can get the default nic and default route both from the same command.
Will add explaination

microsoft/testsuites/network/netinterface.py
squirrelsc: It's ok to have longer timeout even one minute, but we need to check it every 1 or 2 second. There may be a way to check it by dhclient.

lisa/nic.py
squirrelsc: The mutiple value can be returned by an object, it's easier to use.

lisa/features/network_interface.py
squirrelsc: How about call it `reload` or `reload_module`? The netvsc is microsoft specified.

lisa/nic.py
sharsonia: Decided to not return anything, setting values in the method.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -314,6 +315,10 @@ def attach_nics(
         self._log.debug(f"attach the nics into VM {self._node.name} successfully.")
         startstop.start()
 
+    def reload_netvsc(self) -> None:
+        modprobe_tool = self._node.tools[Modprobe]
+        modprobe_tool.reload(["hv_netvsc"])
```
squirrelsc, line 319: Check if it works on sriov. Lili met some problems that there are mlx driver dependencies, so it cannot be reloaded by itself. I'm not sure if they are related.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,120 @@
...
+                node.execute(f"ip route add {default_route}", shell=True, sudo=True)
+            if not node_nic_info.nics[default_nic].ip_addr:
+                node.execute("kill $(pidof dhclient)", shell=True, sudo=True)
+                dhclient = node.tools[Dhclient]
+                dhclient.release()
+                dhclient.renew()
+
+            wait_start = time.time()
+            wait_delta = 0.0
+            while wait_delta < self.DHCLIENT_WAIT_TIME:
```
squirrelsc: Refer to `process.py` to use perf_timer, it's easier to check the timeout. And you can also count it by times, it's clearer than time.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,120 @@
...
+                time.sleep(2)
+                if node_nic_info.nics[default_nic].ip_addr:
```
squirrelsc, line 112: Is there any logic to trigger to retrieve the ip_addr? I'm not sure if it will be changed in each try.

sharsonia: Hmm let me confirm though I believe it shouldn't change

lisa/features/network_interface.py
sharsonia: oh well if I will have to pass the driver name then I would better just call the modprobe function of reload() in the testsuite itself.
The idea to use it in azure was because it might be needed for other test cases too.

squirrelsc: In general, what should be reloaded in the network interface? What's reason to reload it? If it's to reload the network modules, The Azure reloads netvsc, but other platform may reload other modules.
It's better to write out of test cases, this may be needed by other test cases.

microsoft/testsuites/network/netinterface.py
squirrelsc: we need to call some commands to refresh the ip address.

lisa/sut_orchestrator/azure/features.py
sharsonia: This test passed with SRIOV nic.

2021-11-18 22:24:22.792[27388][DEBUG] lisa.runner[0] Runner finished in 229.503 sec.
2021-11-18 22:24:22.797[27388][DEBUG] lisa.RootRunner running count: 0, id: []
2021-11-18 22:24:22.797[27388][DEBUG] lisa.RootRunner running count: 0, id: []
2021-11-18 22:24:22.797[27388][DEBUG] lisa.RootRunner Idle worker available but no new runner...
2021-11-18 22:24:22.802[27388][INFO] lisa.RootRunner ________________________________________
2021-11-18 22:24:22.802[27388][INFO] lisa.RootRunner NetInterface.validate_network_interface_ifup_ifdown: PASSED
2021-11-18 22:24:22.802[27388][INFO] lisa.RootRunner test result summary
2021-11-18 22:24:22.802[27388][INFO] lisa.RootRunner     TOTAL    : 1
2021-11-18 22:24:22.802[27388][INFO] lisa.RootRunner     QUEUED   : 0
2021-11-18 22:24:22.807[27388][INFO] lisa.RootRunner     ASSIGNED : 0
2021-11-18 22:24:22.812[27388][INFO] lisa.RootRunner     RUNNING  : 0
2021-11-18 22:24:22.817[27388][INFO] lisa.RootRunner     FAILED   : 0
2021-11-18 22:24:22.817[27388][INFO] lisa.RootRunner     PASSED   : 1
2021-11-18 22:24:22.817[27388][INFO] lisa.RootRunner     SKIPPED  : 0
2021-11-18 22:24:22.817[27388][INFO] lisa. completed in 231.823 sec

sharsonia:
Test passed with these params - 

    requirement:
      core_count:
        min: 2
      azure:
        marketplace: canonical ubuntuserver 18.04-lts 18.04.202111080 
        vm_size: Standard_DS2_v2
        enable_sriov: True
        location: westus2



[lisa-20211118-222030-999.log](https://github.com/microsoft/lisa/files/7566419/lisa-20211118-222030-999.log)

lisa/features/network_interface.py
```diff
@@ -32,6 +32,9 @@ def attach_nics(
     ) -> None:
         raise NotImplementedError
 
+    def reload_network_module(self) -> None:
```
squirrelsc: It's in network class already, prefer remove `network` in the name.

sharsonia:
Test passed without Sriov

    requirement:
      core_count:
        min: 2
      azure:
        marketplace: redhat rhel 7-lvm 7.6.2020082422
        vm_size: Standard_DS2_v2
        location: westus2

[lisa-20211118-230404-125.log](https://github.com/microsoft/lisa/files/7566625/lisa-20211118-230404-125.log)

microsoft/testsuites/network/netinterface.py
sharsonia: I think it is possible sometimes to ssh to VM even when connection to internet from VM doesn't work.

microsoft/testsuites/network/netinterface.py
sharsonia: > we need to call some commands to refresh the ip address.

you mean refresh the ip address in the nic object? reset_nic_state() does that by calling _get_interface_info() again after setting nic state down and up again.

squirrelsc: I mean, in the `while` loop, it's checking if the ip_addr is assigned. If it's assigned, it will return. But in the loop, there is no command to retrieve ip address from the system. So the ip_addr is not changed for each time it rans. The interval loop doesn't work actually. if the ip_addr is not set in first time, it will wait the whole time. It won't return earlier.

microsoft/testsuites/network/netinterface.py
squirrelsc: It's possible, if the VM is tested in an internal network. But it doesn't related to the test cases. The test cases don't require internet access.

lisa/base_tools/wget.py
```diff
@@ -65,3 +65,14 @@ def get(
             self.node.execute(f"chmod +x {actual_file_path}")
 
         return actual_file_path.stdout
+
+    def verify_internet_access(self) -> bool:
+        result = self.run(" microsoft.com", no_error_log=True, shell=True)
```
squirrelsc: Call the `get` method to benefit from the checking logic.

lisa/base_tools/wget.py
```diff
@@ -65,3 +65,14 @@ def get(
...
+        matched_result = self.__pattern_path.match(result.stdout)
+        if matched_result:
+            download_file_path = matched_result.group("path")
+            actual_file_path = self.node.execute(f"ls {download_file_path}", shell=True)
+            self.node.execute(f"rm {actual_file_path.stdout}", shell=True, sudo=True)
```
squirrelsc: If it use the shared logic, it doesn't need to use ls to find the file in this method.

lisa/base_tools/wget.py
```diff
@@ -65,3 +65,13 @@ def get(
...
         return actual_file_path.stdout
+
+    def verify_internet_access(self) -> bool:
+        try:
+            result = self.get("microsoft.com")
+            if result:
+                self.node.execute(f"rm {result}", shell=True, sudo=True)
+                return True
+        except Exception:
+            self._log.debug("Internet is not accessible")
```
squirrelsc: Log the internal exception, so we can see what happens underlying.

squirrelsc (dismissed):

sharsonia:
Results with latest code
[lisa-20211120-022757-367.log](https://github.com/microsoft/lisa/files/7573780/lisa-20211120-022757-367.log)

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,139 @@
...
+                can be reloaded gracefully when done multiple times.
+
+            Steps:
+            1. Validate netvsc isn't built-in already. If it is then skip the test.
+            2. Unload and load netvsc module multiple times in a loop.
+
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            min_nic_count=1,
```
squirrelsc: This condition is not necessary. It's impossible to have no nic.

microsoft/testsuites/network/netinterface.py
```diff
@@ -0,0 +1,139 @@
...
+                node.execute("kill $(pidof dhclient)", shell=True, sudo=True)
+                dhclient = node.tools[Dhclient]
+                dhclient.renew()
+
+            timer = perf_timer.Timer()
+            while timer.elapsed(stop=False) < self.DHCLIENT_TIMEOUT:
+                node_nic_info.load_interface_info(default_nic)
+                if node_nic_info.nics[default_nic].ip_addr:
+                    break
+                time.sleep(2)
```
squirrelsc: Let's check it per second, it's easier to understand.

microsoft/testsuites/network/netinterface.py
sharsonia: We already have it in dpdk testsuite and synthetic testsuite and dnt see any harm in specifying it explicitly since we have test case where we boot up VM without adding any nic

squirrelsc (approved):

PR #1631, opened 2021-11-18
Migrate attach max synthetic during and after VM provision.

LiliDeng:
See commit message for details.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -10,7 +10,7 @@ from assertpy import assert_that
 
 from lisa import Node
 from lisa.tools import Echo
-from lisa.util import InitializableMixin, LisaException
+from lisa.util import InitializableMixin, LisaException, get_matched_str
 
 
 class NicInfo:
@@ -119,6 +119,8 @@ class Nics(InitializableMixin):
         )
     )
 
+    _file_not_exist = re.compile(r"No such file or directory", re.MULTILINE)
+
     def __init__(self, node: Node):
         super().__init__()
         self._node = node
@@ -143,7 +145,7 @@ class Nics(InitializableMixin):
         return [x.upper for x in self.nics.values() if not x.lower]
 
     def get_upper_nics(self) -> List[str]:
-        return [x.upper for x in self.nics.values() if x.lower]
+        return [x.upper for x in self.nics.values()]
 
     def get_lower_nics(self) -> List[str]:
         return [x.lower for x in self.nics.values() if x.lower]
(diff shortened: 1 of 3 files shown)


lisa/nic.py
```diff
@@ -143,7 +145,7 @@ def get_unpaired_devices(self) -> List[str]:
         return [x.upper for x in self.nics.values() if not x.lower]
 
     def get_upper_nics(self) -> List[str]:
-        return [x.upper for x in self.nics.values() if x.lower]
+        return [x.upper for x in self.nics.values()]
```
squirrelsc, line 148: Please check with Matthew, if it's a breaking change.

lisa/nic.py
```diff
@@ -119,6 +119,8 @@ class Nics(InitializableMixin):
         )
     )
 
+    __file_not_exist = re.compile(r".*No such file or directory", re.MULTILINE)
```
squirrelsc: This class is not a super class of others. So one underscore is enough to mark private variables.

squirrelsc: The leading ".*" should be able to remove, it doesn't affect in this pattern.

lisa/nic.py
```diff
@@ -235,22 +237,23 @@ def _get_node_nic_info(self) -> None:
...
-                raise LisaException(
-                    "Could not locate default network interface"
-                    f" in output:\n{result.stdout}"
-                )
-            upper_nic, lower_nic, pci_slot = match.groups()
-            nic_info = NicInfo(upper_nic, lower_nic, pci_slot)
-            self.append(nic_info)
+        result = self._node.execute(nic_info_fetch_cmd, shell=True)
+        if self.__file_not_exist.match(result.stdout):
+            self._node.log.debug(f"There is no VF on node {self._node.name}.")
```
squirrelsc: Do we need this log line? It looks not provide many information.

lisa/nic.py
```diff
@@ -235,22 +237,23 @@ def _get_node_nic_info(self) -> None:
         # use sysfs to gather upper/lower nic pairings and pci slot info
         nic_info_fetch_cmd = "ls -la /sys/class/net/*/lower*/device"
         self._node.log.info(f"Gathering NIC information on {self._node.name}.")
```
squirrelsc: Change this line to debug level.

lisa/nic.py
sharsonia: no this shouldn't be a breaking change. We need to remove this coupling if we want to use this class as generic class for both Synthetic and Sriov nic.
Upper nic can be the synthetic nic and might not be paired with a lower nic if AN is not enabled.

lisa/nic.py
LiliDeng: update for both comments

lisa/nic.py
LiliDeng: removed

lisa/nic.py
LiliDeng: changed it into debug level

squirrelsc (approved):

PR #1632, opened 2021-11-18
[LISAv3] Adding lis_version_check test case

isaiahduncan:


microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@TestSuiteMetadata(
+    area="lis",
+    category="functional",
+    description="""
+    ##################################
+    """,
+)
+class lis(TestSuite):  # noqa
```
squirrelsc: Lis

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,98 @@
...
+    @TestCaseMetadata(
+        description="""
+            #################
+        """,
+        priority=1,
+    )
+    def verify_lis_driver_version(self, node: Node) -> None:
+        wget_tool = node.tools[Wget]
+        rpm_status = node.execute(
```
squirrelsc: Add operating system as a requirement.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,98 @@
...
+            '| grep "microsoft-hyper-v"'
+        ).exit_code
+
+        if rpm_status != 0:
+            raise SkippedException("No LIS RPM's are detected. Skipping test.")
+        version = node.execute(
+            """modinfo "hv_vmbus" | grep version: | head -1 | awk '{print $2}'"""
+        ).stdout
+
+        for i in range(5, 8):
```
squirrelsc, line 64: Add comments to explain the purpose, it's not easy to get.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,98 @@
...
+
+            assert_that(check_res and check_hex_res).described_as(
+                f"The result of _check_lis_version() was {check_res}. "
+                f"The result of _check_lis_version_hex() was {check_hex_res}."
+            ).is_equal_to(True)
+        pass
+
+    # returns true if version and source_version are the same
+    def _check_lis_version(self, node: Node, version: str, source_version: str) -> bool:
+        self.__log.debug("Detected modinfo version is {version}")
```
squirrelsc: Don't use the `self.__log`. It's for test suite. Use the log from test method, like other test cases.

microsoft/testsuites/lis/lissuite.py
squirrelsc: It looks downloaded some header files, and check if it's consistent with system. What's the logic is validating?

isaiahduncan: It downloads header files based on the LIS version and compares the installed LIS version to the expected one which is found in the header files.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,109 @@
...
+        description="""
+            #################
+        """,
+        priority=1,
+    )
+    def verify_lis_driver_version(self, node: Node, log: Logger) -> None:
+        wget_tool = node.tools[Wget]
+        node.tools[LisDriver]
+
+        # Checking for RMP
```
LiliDeng: what is RMP?

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,109 @@
...
+        rpm_status = node.execute(
+            'rpm -qa | grep "kmod-microsoft-hyper-v" && rpm -qa '
+            '| grep "microsoft-hyper-v"'
+        ).exit_code
+        if rpm_status != 0:
+            raise SkippedException("No LIS RPM's are detected. Skipping test.")
+
+        # Capturing LIS version from the node
+        version = node.execute(
+            """modinfo "hv_vmbus" | grep version: | head -1 | awk '{print $2}'""",
```
LiliDeng: use modinfo tool to get version

lisa/tools/reboot.py
```diff
@@ -27,7 +27,7 @@ def _who_last(who: Who) -> datetime:
 class Reboot(Tool):
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         # timeout to wait
-        self.time_out: int = 300
+        self.time_out: int = 360
```
squirrelsc: Did you find it's not enough? If it's not enough sometimes. We can consider to use a variable in reboot method, so it can be set when reboot. The default value keeps 5 minutes, because many smoke test case will rely on the timeout value. the extended time will extend the whole smoke test a lot.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,109 @@
...
+            wget_tool.get(
+                "https://raw.githubusercontent.com/LIS/lis-next/"
+                f"{version}/hv-rhel{i}.x/hv/include/linux/hv_compat.h",
+                filename="hv_compat.h",
+                file_path="./",
+            )
+
+            # Capturing the LIS version from the source code
+            source_version = node.execute(
+                """grep 'define HV_DRV_VERSION' hv_compat.h|cut -d '"' -f 2""",
```
squirrelsc: if the file content is not long (no more than 200 lines). cat the content to Python, and use the regexp to parse it. If it's too long, use grep to filter out, and also use python to parse the line.

microsoft/testsuites/lis/lissuite.py
squirrelsc: ok, add comments in code to explain the logic.

lisa/tools/reboot.py
isaiahduncan: I can make a seperate commit to add that functionality

microsoft/testsuites/lis/lissuite.py
isaiahduncan: I'm having trouble getting the modinfo tool to print out the lis version. This is the command i'm using: `
print("get ver: ", modinfo_tool.get_version("hv_vmbus"))`
and this is the output i'm getting: 
`2021-11-30 16:16:17.096[14604][INFO] lisa.stdout get ver:  `

microsoft/testsuites/lis/lissuite.py
isaiahduncan: *RPM. I corrected it

microsoft/testsuites/lis/lissuite.py
isaiahduncan: It's more than 200 lines

microsoft/testsuites/lis/lissuite.py
squirrelsc: What's the raw output? You can check the code, and modify the regexp to include the new format. Don't forget to add the new format in comment examples.

microsoft/testsuites/lis/lissuite.py
squirrelsc: So, use the grep to get this line printed only, and parse it in Python.

squirrelsc: you can add a new method in `Cat` called `read_with_filter`. It integrates the cat and grep in one line.

lisa/node.py
```diff
@@ -164,6 +164,9 @@ def create(
     def reboot(self) -> None:
         self.tools[Reboot].reboot()
 
+    def specific_timeout_reboot(self, time_out: int) -> None:
```
squirrelsc: It doesn't need to create another method, add an argument with default value like below.

```python
def reboot(self, timeout: int = 300) -> None:

```

microsoft/testsuites/lis/lissuite.py
isaiahduncan: I put the requirement on the test suite instead of the test case. Is that ok?

microsoft/testsuites/lis/lissuite.py
isaiahduncan: The raw output usually doesn't print out the LIS version. I've only seen one run where it actually printed out the LIS version. This is a snippet from a recent run: 
`2021-11-30 20:20:02.601[9792][DEBUG] lisa.env[generated_0].node[0].tool[modinfo] loaded cached result for command: [modinfo hv_vmbus]
2021-11-30 20:20:02.601[9792][INFO] lisa.stdout info:  filename:       /lib/modules/3.10.0-862.11.6.el7.x86_64/kernel/drivers/hv/hv_vmbus.ko.xz
2021-11-30 20:20:02.605[9792][INFO] lisa.stdout license:        GPL
2021-11-30 20:20:02.606[9792][INFO] lisa.stdout retpoline:      Y
2021-11-30 20:20:02.606[9792][INFO] lisa.stdout rhelversion:    7.5
2021-11-30 20:20:02.607[9792][INFO] lisa.stdout srcversion:     FDEC375DD4661BC80D39F62
2021-11-30 20:20:02.608[9792][INFO] lisa.stdout alias:          acpi*:VMBus:*
2021-11-30 20:20:02.609[9792][INFO] lisa.stdout alias:          acpi*:VMBUS:*
2021-11-30 20:20:02.610[9792][INFO] lisa.stdout depends:        
2021-11-30 20:20:02.610[9792][INFO] lisa.stdout intree:         Y
2021-11-30 20:20:02.611[9792][INFO] lisa.stdout vermagic:       3.10.0-862.11.6.el7.x86_64 SMP mod_unload modversions 
2021-11-30 20:20:02.612[9792][INFO] lisa.stdout signer:         CentOS Linux kernel signing key
2021-11-30 20:20:02.613[9792][INFO] lisa.stdout sig_key:        D4:11:5F:11:00:55:DB:56:C8:D6:05:AB:75:21:73:CF:B1:AC:54:D8
2021-11-30 20:20:02.614[9792][INFO] lisa.stdout sig_hashalgo:   sha256
2021-11-30 20:20:02.615[9792][INFO] lisa.stdout get ver:  7.5`

lisa/node.py
isaiahduncan: Changed

microsoft/testsuites/lis/lissuite.py
isaiahduncan: Do you think it would be better if we just created a grep tool?

microsoft/testsuites/lis/lissuite.py
squirrelsc: It's ok, if the requirement is the same to all test cases. If some test cases are different, it can be rewritten at test case level also.

microsoft/testsuites/lis/lissuite.py
squirrelsc: The result is cached, and only printed in first run. It needs to look up in the node's log to find the previous runs of this command.

microsoft/testsuites/lis/lissuite.py
squirrelsc: grep is not recommended in LISA. It hides the raw information, and it's difficult to debug from the raw output. It's used in limited area, like long output or extremely high performance is needed.

lisa/tools/reboot.py
```diff
@@ -55,7 +55,8 @@ def reboot_and_check_panic(self, log_path: Path) -> None:
                 raise LisaException(f"after reboot, {identifier}")
             raise identifier
 
-    def reboot(self) -> None:
+    def reboot(self, time_out: int = 300) -> None:
+        self.time_out = time_out
```
squirrelsc: the self.time_out is not useful anymore.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    def verify_lis_driver_version(self, node: Node, log: Logger) -> None:
+        cat = node.tools[Cat]
+        modinfo_tool = node.tools[Modinfo]
+        wget_tool = node.tools[Wget]
+        node.tools[LisDriver]
+
+        # Checking for RPM Package Manager
+        rpm_status = node.execute(
+            'rpm -qa | grep "kmod-microsoft-hyper-v" && rpm -qa '
```
squirrelsc: avoid to use grep, try patterns in Python code.

lisa/tools/reboot.py
isaiahduncan: removed it

microsoft/testsuites/lis/lissuite.py
isaiahduncan: The code now uses modinfo properly.

microsoft/testsuites/lis/lissuite.py
isaiahduncan: done

microsoft/testsuites/lis/lissuite.py
```diff
@@ -0,0 +1,116 @@
...
+                f"{version}/hv-rhel{i}.x/hv/include/linux/hv_compat.h",
+                filename="hv_compat.h",
+                file_path="./",
+            )
+
+            # Capturing the LIS version from the source code
+            source_version = cat.read_with_filter(
+                "hv_compat.h", "define HV_DRV_VERSION"
+            )
+            version_pattern = re.compile(r'"(.+)"')
```
squirrelsc: 1. move regexp to class level, so it doesn't need to be compiled every time. It's a small perf impact, but a good practice.
2. add examples above the patterns. So if it missed some case, people can test the updated pattern with previous examples.

isaiahduncan: Done

squirrelsc (approved):

PR #1633, opened 2021-11-18
initial package version fetch, debian and ubuntu, version and name only

mcgov:
Add package version check for packages installed by package manager on a distro. Caches results and has option to force an update and not use the cached version.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -6,8 +6,20 @@ import time
 from dataclasses import dataclass
 from functools import partial
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Pattern, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Match,
+    Optional,
+    Pattern,
+    Type,
+    Union,
+)
 
+from assertpy import assert_that
 from semver import VersionInfo
 
 from lisa.base_tools import Cat, Sed, Wget
@@ -88,6 +100,7 @@ class OperatingSystem:
         self._is_posix = is_posix
         self._log = get_logger(name="os", parent=self._node.log)
         self._information: Optional[OsInformation] = None
+        self._packages: Dict[str, VersionInfo] = dict()
 
     @classmethod
     def create(cls, node: "Node") -> Any:
@@ -316,6 +329,14 @@ class Posix(OperatingSystem, BaseClassMixin):
         except FileNotFoundError:
             self._log.debug("File /etc/os-release doesn't exist.")
 
+    def get_package_information(
+        self, package_name: str, use_cached: bool = True
+    ) -> VersionInfo:
+        found = self._packages.get(package_name, None)
+        if found and use_cached:
+            return found
+        return self._get_package_information(package_name)
+
     def get_repositories(self) -> List[RepositoryInfo]:
         raise NotImplementedError("get_repositories is not implemented")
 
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -58,6 +68,15 @@ class OsInformation:
     full_version: str = "Unknown"
 
 
+class PackageInformation:
+    name: str
+    release: str
+
+    def __init__(self, name: str, release: str) -> None:
+        self.name = name
+        self.release = release
```
squirrelsc: Can it use the VersionInfo class like kernel version? So it can be comparable easily.

lisa/operating_system.py
```diff
@@ -485,6 +524,36 @@ def get_apt_error(self, stdout: str) -> List[str]:
                 error_lines.append(line)
         return error_lines
 
+    def get_package_information(
+        self, package_name: str, use_cached: bool = True
+    ) -> PackageInformation:
+
+        if use_cached:  # try to fetch cached version
```
squirrelsc: It can be like below,

```python
package = self._packages.get(package_name, None)

if package and use_cache:
    return package

# get package info
```

squirrelsc: The use cache is also a sharable logic, it can be in base class, and subclass implement `_get_package_information`

mcgov:
Blocking, working on some changes still

lisa/operating_system.py
```diff
@@ -331,6 +343,11 @@ def _initialize_package_installation(self) -> None:
         # sub os can override it, but it's optional
         pass
 
+    def _get_package_information(
```
squirrelsc: The subclasses should implement this method. The super class (`Posix`) should implement get_package_information. When it needs the information call underlying `_get_package_information`.

lisa/operating_system.py
```diff
@@ -789,6 +862,15 @@ class Fedora(Linux):
     # microsoft-azure-rhel8-eus  Microsoft Azure RPMs for RHEL8 Extended Update Support
     _fedora_repository_info_pattern = re.compile(r"(?P<id>\S+)\s+(?P<name>\S.*\S)\s*")
 
+    # ex: dpdk-20.11-3.el8.x86_64
+    _fedora_version_splitter_regex = re.compile(
```
squirrelsc, line 904: try named groups, it should be much easier to self-explain.

mcgov: I learn something new every day 😂 will change

lisa/operating_system.py
```diff
@@ -332,6 +352,37 @@ def _initialize_package_installation(self) -> None:
...
+    def _get_version_info_from_named_regex_match(
+        self, package_name: str, named_matches: re.Match[str]
+    ) -> VersionInfo:
+
+        assert_that(named_matches.group("package_name")).described_as(
+            "While capturing package version info, found name of package was "
+            "unexpected. Version info may be incorrect or regex may need updating."
+        ).is_equal_to(package_name)
+        patch_match = named_matches.group("patch")
+        if patch_match:
```
squirrelsc: Should it be `not patch_match`? And is it possible to leave the patch as is? The default "0" may be confusing sometime.

lisa/operating_system.py
```diff
@@ -332,6 +352,37 @@ def _initialize_package_installation(self) -> None:
...
+            patch_match = "0"
+
+        major, minor, patch = map(
+            int,
+            [named_matches.group("major"), named_matches.group("minor"), patch_match],
+        )
+        build_match = named_matches.group("build")
+        self._node.log.info(
```
squirrelsc: 1. make it debug level.
2. The information can print out the original parsed information, because it needs the exact raw string sometimes.

lisa/operating_system.py
mcgov: Yep, sorry I'm committing here to be able to cherry pick to my test branch.
VersionInfo requires an entry for 'patch' since it expects three numbers though. If there's a patch version released it should have the entry, I noticed some packages leave it off when it's 0 (dpdk).

squirrelsc: ok, we can add a default "0".

squirrelsc (dismissed):

mcgov:
Had to remove a type problem, re.Match doesn't need a type subscript 🤦

mcgov:
Typing has a Match class I was supposed to use for re.Match. It looks like it will be deprecated at some point but we can swap to the re.Match class whenever the builds start failing.

squirrelsc (approved):

LiliDeng:
@mcgov I think you need implement this method for every distro, when I use SUSE, and invoke this method, it threw NotImplementedError

squirrelsc:
> @mcgov I think you need implement this method for every distro, when I use SUSE, and invoke this method, it threw NotImplementedError

Both of you can help implementing it, if your test cases met it. If the test cases are failed in lab due to lack of implementation, the test case owner should fix it. Thank you.

PR #1637, opened 2021-11-22
Migrate sriov_interrupts_change case and sriov_iperf_stress case.

LiliDeng:
See commits message as details.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, cast
 
 import requests
 from assertpy import assert_that
+from azure.mgmt.compute.models import NetworkInterfaceReference  # type: ignore
 from dataclasses_json import dataclass_json
 from PIL import Image, UnidentifiedImageError
 
@@ -170,6 +171,18 @@ class NetworkInterface(AzureFeatureMixin, features.NetworkInterface):
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
 
+    def _get_primary(
+        self, nics: List[NetworkInterfaceReference]
+    ) -> NetworkInterfaceReference:
+        found_primary = False
+        for nic in nics:
+            if nic.primary:
+                found_primary = True
+                break
+        if not found_primary:
+            raise LisaException(f"fail to find primary nic for vm {self._node.name}")
+        return nic
+
     def _switch_sriov(self, enable: bool) -> None:
         azure_platform: AzurePlatform = self._platform  # type: ignore
         network_client = get_network_client(azure_platform)
@@ -217,13 +230,7 @@ class NetworkInterface(AzureFeatureMixin, features.NetworkInterface):
         vm = compute_client.virtual_machines.get(
             self._resource_group_name, self._node.name
         )
-        found_primary = False
-        for nic in vm.network_profile.network_interfaces:
-            if nic.primary:
-                found_primary = True
-                break
-        if not found_primary:
-            raise LisaException(f"fail to find primary nic for vm {self._node.name}")
+        nic = self._get_primary(vm.network_profile.network_interfaces)
         nic_name = nic.id.split("/")[-1]
         primary_nic = network_client.network_interfaces.get(
             self._resource_group_name, nic_name
@@ -257,13 +264,7 @@ class NetworkInterface(AzureFeatureMixin, features.NetworkInterface):
                 f"nic count after add extra nics is {nic_count_after_add_extra},"
                 f" it exceeds the vm size's capability {node_capability_nic_count}."
             )
-        found_primary = False
-        for nic in vm.network_profile.network_interfaces:
-            if nic.primary:
-                found_primary = True
-                break
-        if not found_primary:
-            raise LisaException(f"fail to find primary nic for vm {self._node.name}")
+        nic = self._get_primary(vm.network_profile.network_interfaces)
         nic_name = nic.id.split("/")[-1]
         primary_nic = network_client.network_interfaces.get(
             self._resource_group_name, nic_name
@@ -312,6 +313,37 @@ class NetworkInterface(AzureFeatureMixin, features.NetworkInterface):
             },
         )
         self._log.debug(f"attach the nics into VM {self._node.name} successfully.")
+        startstop.start()
+
+    def remove_extra_nics(self) -> None:
+        azure_platform: AzurePlatform = self._platform  # type: ignore
+        network_client = get_network_client(azure_platform)
+        compute_client = get_compute_client(azure_platform)
+        vm = compute_client.virtual_machines.get(
+            self._resource_group_name, self._node.name
+        )
+        if len(vm.network_profile.network_interfaces) == 1:
+            self._log.debug("No existed extra nics can be disassociated.")
+            return
+        nic = self._get_primary(vm.network_profile.network_interfaces)
+        nic_name = nic.id.split("/")[-1]
+        primary_nic = network_client.network_interfaces.get(
+            self._resource_group_name, nic_name
+        )
+        network_interfaces_section = []
+        network_interfaces_section.append({"id": primary_nic.id, "primary": True})
+        startstop = self._node.features[StartStop]
+        startstop.stop()
+        compute_client.virtual_machines.begin_update(
+            resource_group_name=self._resource_group_name,
+            vm_name=self._node.name,
+            parameters={
+                "network_profile": {"network_interfaces": network_interfaces_section},
+            },
+        )
+        self._log.debug(
+            f"Only associated nic {primary_nic.id} into VM {self._node.name}."
+        )
         startstop.start()
 
     def reload_module(self) -> None:
--- a/lisa/tools/interrupt_inspector.py
+++ b/lisa/tools/interrupt_inspector.py
@@ -1,0 +1,120 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from __future__ import annotations
+
+import re
+from typing import Dict, List
+
+from lisa.executable import Tool
+from lisa.tools import Cat
+
+
+class Interrupt:
+    irq_number: str
+    cpu_counter: List[int]
+    metadata: str
+    counter_sum: int
+
+    def __init__(
+        self,
+        irq_number: str,
+        cpu_counter: List[int],
+        counter_sum: int,
+        metadata: str = "",
+    ) -> None:
+        self.irq_number = irq_number
+        self.cpu_counter = cpu_counter
+        self.metadata = metadata
+        self.counter_sum = counter_sum
+
+    def __str__(self) -> str:
+        return (
+            f"irq_number : {self.irq_number}, "
+            f"count : {self.cpu_counter}, "
+            f"metadata : {self.metadata}"
+            f"sum : {self.counter_sum}"
+        )
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+
+class InterruptInspector(Tool):
+    # 0:         22          0  IR-IO-APIC   2-edge      timer
+    _interrupt_regex = re.compile(
+        r"^\s*(?P<irq_number>\S+):\s+(?P<cpu_counter>[\d+ ]+)\s*(?P<metadata>.*)$"
+    )
+
+    @property
+    def command(self) -> str:
+        return "cat /proc/interrupts"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def get_interrupt_data(self) -> List[Interrupt]:
+        # Run cat /proc/interrupts. The output is of the form :
+        #          CPU0       CPU1
+        # 0:         22          0  IR-IO-APIC   2-edge      timer
+        # 1:          2          0  IR-IO-APIC   1-edge      i8042
+        # ERR:        0
+        # The first column refers to the IRQ number. The next column contains
+        # number of interrupts per IRQ for each CPU in the system. The remaining
+        # column report the metadata about interrupts, including type of interrupt,
+        # device etc. This is variable for each distro.
+        # Note : Some IRQ numbers have single entry because they're not actually
+        # CPU stats, but events count belonging to the IO-APIC controller. For
+        # example, `ERR` is incremented in the case of errors in the IO-APIC bus.
+        result = (
+            self.node.tools[Cat]
+            .run("/proc/interrupts", sudo=True, force_run=True)
+            .stdout
+        )
+        mappings_with_header = result.splitlines(keepends=False)
+        mappings = mappings_with_header[1:]
+        assert len(mappings) > 0
+
+        interrupts = []
+        for line in mappings:
+            matched = self._interrupt_regex.fullmatch(line)
+            assert matched
+            cpu_counter = [int(count) for count in matched.group("cpu_counter").split()]
+            counter_sum = sum([int(x) for x in cpu_counter])
+            interrupts.append(
+                Interrupt(
+                    irq_number=matched.group("irq_number"),
+                    cpu_counter=cpu_counter,
+                    counter_sum=counter_sum,
+                    metadata=matched.group("metadata"),
+                )
+            )
+
+        return interrupts
+
+    def sum_cpu_counter_by_irqs(self, pci_slot: str) -> List[Dict[str, int]]:
+        interrupts_sum_by_irqs: List[Dict[str, int]] = []
+        interrupts = self.get_interrupt_data()
+        matched_interrupts = [x for x in interrupts if pci_slot in x.metadata]
+        for interrupt in matched_interrupts:
+            interrupts_sum_by_irqs.append({interrupt.irq_number: interrupt.counter_sum})
+        return interrupts_sum_by_irqs
+
+    def sum_cpu_counter_by_index(self, pci_slot: str) -> Dict[int, int]:
+        interrupts_sum_by_cpus: Dict[int, int] = {}
+        interrupts = self.get_interrupt_data()
+        matched_interrupts = [x for x in interrupts if pci_slot in x.metadata]
+        for cpu_index in range(0, len(matched_interrupts[0].cpu_counter)):
+            interrupts_sum_by_cpus[cpu_index] = self._get_sum_of_interrupt_data_per_cpu(
+                matched_interrupts, cpu_index
+            )
+        return interrupts_sum_by_cpus
+
+    def _get_sum_of_interrupt_data_per_cpu(
+        self, interrupts: List[Interrupt], index: int
+    ) -> int:
+        sum = 0
+        for interrupt in interrupts:
+            sum += interrupt.cpu_counter[index]
+        return sum
--- a/microsoft/testsuites/network/sriov.py
+++ b/microsoft/testsuites/network/sriov.py
@@ -30,6 +30,7 @@ from .common import (
     cleanup_iperf3,
     initialize_nic_info,
     load_module,
+    remove_extra_nics,
     remove_module,
     sriov_basic_test,
     sriov_vf_connection_test,
@@ -271,7 +272,6 @@ class Sriov(TestSuite):
         2. Do the basic sriov testing.
         """,
         priority=2,
-        use_new_environment=True,
         requirement=simple_requirement(
             network_interface=schema.NetworkInterfaceOptionSettings(
                 data_path=schema.NetworkDataPath.Sriov,
@@ -310,7 +310,6 @@ class Sriov(TestSuite):
         2. Do the basic sriov testing.
         """,
         priority=2,
-        use_new_environment=True,
         requirement=simple_requirement(
             min_nic_count=8,
             network_interface=features.Sriov(),
@@ -631,3 +630,4 @@ class Sriov(TestSuite):
     def after_case(self, log: Logger, **kwargs: Any) -> None:
         environment: Environment = kwargs.pop("environment")
         cleanup_iperf3(environment)
+        remove_extra_nics(environment)
--- a/microsoft/testsuites/network/sriov.py
+++ b/microsoft/testsuites/network/sriov.py
@@ -1,29 +1,40 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 from pathlib import Path
-from typing import Any, Dict, List, cast
+from typing import Any, Dict, cast
 
 from assertpy import assert_that
 
 from lisa import (
     Environment,
     Logger,
-    Node,
     RemoteNode,
     TcpConnetionException,
     TestCaseMetadata,
     TestSuite,
     TestSuiteMetadata,
+    features,
+    node_requirement,
     schema,
     search_space,
     simple_requirement,
 )
 from lisa.features import NetworkInterface, SerialConsole
-from lisa.nic import NicInfo, Nics
+from lisa.nic import NicInfo
 from lisa.sut_orchestrator import AZURE, READY
-from lisa.tools import Cat, Ethtool, Firewall, Iperf3, Kill, Lsmod, Lspci, Modprobe, Ssh
+from lisa.tools import Cat, Ethtool, InterruptInspector, Iperf3, Lspci
 from lisa.util import SkippedException, constants
 from lisa.util.shell import wait_tcp_port_ready
+
+from .common import (
+    cleanup_iperf3,
+    initialize_nic_info,
+    load_module,
+    remove_module,
+    sriov_basic_test,
+    sriov_vf_connection_test,
+    stop_firewall,
+)
 
 
 @TestSuiteMetadata(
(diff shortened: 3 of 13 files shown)


lisa/interrupts.py
```diff
@@ -0,0 +1,109 @@
...
+from __future__ import annotations
+
+import re
+from typing import Dict, List
+
+from lisa import Node
+from lisa.tools import Cat
+
+
+class Interrupt:
```
squirrelsc: How about make it like a tool?

lisa/interrupts.py
```diff
@@ -0,0 +1,109 @@
...
+            f"irq_number : {self.irq_number}, "
+            f"count : {self.interrupt_count}, "
+            f"metadata : {self.metadata}"
+        )
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    @staticmethod
+    def get_interrupt_data(node: Node) -> List[Interrupt]:
```
squirrelsc: Create a `InterruptInspector` tool, and return the `Interrupt` objects. It's confusing to put the collection operations into the single item class.

lisa/interrupts.py
```diff
@@ -0,0 +1,109 @@
...
+                Interrupt(
+                    irq_number=matched.group("irq_number"),
+                    interrupt_count=interrupt_count,
+                    metadata=matched.group("metadata"),
+                )
+            )
+        return interrupts
+
+    @staticmethod
+    def get_sum_of_interrupt_data_by_irqs(
```
squirrelsc, line 77: This can be a value in each interrupt, and calculated when get the data.

lisa/interrupts.py
```diff
@@ -0,0 +1,109 @@
...
+        node: Node, pci_slot: str
+    ) -> List[Dict[str, int]]:
+        interrupts_sum_by_irqs: List[Dict[str, int]] = []
+        interrupts = Interrupt.get_interrupt_data(node)
+        matched_interrupts = [x for x in interrupts if pci_slot in x.metadata]
+        for interrupt in matched_interrupts:
+            interrupts_sum_by_irqs.append(
+                {interrupt.irq_number: sum([int(x) for x in interrupt.interrupt_count])}
```
squirrelsc: This can be a value in each interrupt, and calculated when get the data.

lisa/interrupts.py
```diff
@@ -0,0 +1,109 @@
...
+        for cpu_index in range(0, len(matched_interrupts[0].interrupt_count)):
+            interrupts_sum_by_cpus[
+                cpu_index
+            ] = Interrupt.get_sum_of_interrupt_data_per_cpu(
+                matched_interrupts, cpu_index
+            )
+        return interrupts_sum_by_cpus
+
+    @staticmethod
+    def get_sum_of_interrupt_data_per_cpu(
```
squirrelsc: This is used internally, it doesn't need to be public.

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
         )
         assert_that(iperf_log).does_not_contain("error")
 
+    @TestCaseMetadata(
+        description="""
+        This case is to verify interrupts count increased after network traiffic
```
squirrelsc: typo traiffic -> traffic

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+        7. Compare interrupts changes, expected to see interrupts increased.
+        """,
+        priority=2,
+        timeout=1200,
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                node_count=2,
+                core_count=search_space.IntRange(min=8, max=16),
+                network_interface=schema.NetworkInterfaceOptionSettings(
+                    nic_count=search_space.IntRange(min=1, max=8),
```
squirrelsc: This rule is not necessary. It almost allows all nic count.

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+        5. Compare interrupts changes, expected to see interrupts increased.
+        6. Get final interrupts sum per cpu on client node.
+        7. Compare interrupts changes, expected to see interrupts increased.
+        """,
+        priority=2,
+        timeout=1200,
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                node_count=2,
+                core_count=search_space.IntRange(min=8, max=16),
```
squirrelsc, line 548: What's the reason to limit core count between 8 and 16? If there is any reason, add it in test spec.

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+    )
+    def sriov_interrupts_change(self, environment: Environment) -> None:
+        server_node = cast(RemoteNode, environment.nodes[0])
+        client_node = cast(RemoteNode, environment.nodes[1])
+        vm_nics = self._initialize_nic_info(environment)
+
+        server_iperf3 = server_node.tools[Iperf3]
+        client_iperf3 = client_node.tools[Iperf3]
+        # 1. Start iperf3 on server node.
+        server_iperf3.run_as_server()
```
squirrelsc, line 582: How long this test case takes? Can it be in 30 seconds?

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+            for _, server_nic_info in vm_nics[server_node.name].items():
+                if (
+                    server_nic_info.ip_addr.rsplit(".", maxsplit=1)[0]
+                    == client_nic_info.ip_addr.rsplit(".", maxsplit=1)[0]
+                ):
+                    matched_server_nic_info = server_nic_info
+                    break
+            assert matched_server_nic_info
+            # 3. Start iperf3 for 10 minutes with 128 threads on client node.
+            client_iperf3.run_as_client(
```
squirrelsc, line 591: How long this test case takes? Can it be in 30 seconds?

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+        client_node = cast(RemoteNode, environment.nodes[1])
+        vm_nics = self._initialize_nic_info(environment)
+        client_iperf3_log = "iperfResults.log"
+        server_iperf3 = server_node.tools[Iperf3]
+        client_iperf3 = client_node.tools[Iperf3]
+        # 1. Start iperf3 on server node.
+        server_iperf3.run_as_server()
+
+        # 2. Start iperf3 for 30 minutes on client node.
+        for _, client_nic_info in vm_nics[client_node.name].items():
```
squirrelsc: Do we need to run on all nics for the perf test case?

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+                server_ip=matched_server_nic_info.ip_addr,
+                log_file=client_iperf3_log,
+                seconds=1800,
+                client_ip=client_nic_info.ip_addr,
+            )
+            client_cat = client_node.tools[Cat]
+            iperf_log = client_cat.read_from_file(
+                client_iperf3_log, sudo=True, force_run=True
+            )
+            assert_that(iperf_log).does_not_contain("error")
```
squirrelsc: Add described_as to show more practical information.

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +544,158 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+        This case is to check whether the network connectivity is lost after running
+         iperf3 for 30 mins.
+
+        Steps,
+        1. Start iperf3 on server node.
+        2. Start iperf3 for 30 minutes on client node.
+        3. Do VF connection test.
+        """,
+        priority=3,
+        timeout=18000,
```
squirrelsc: Set its category to performance or stress like v2.

LiliDeng: add a new file named stress? and put this case there?

squirrelsc: It's ok to put test case in this file, because they may share the logic. The category should be performance or stress, not default functional.

microsoft/testsuites/network/sriov.py
LiliDeng: will add later

lisa/interrupts.py
LiliDeng: add it as a tool

lisa/interrupts.py
LiliDeng: done

lisa/interrupts.py
LiliDeng: added

lisa/interrupts.py
LiliDeng: update

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/sriov.py
LiliDeng: removed

microsoft/testsuites/network/sriov.py
LiliDeng: update it into 60s

microsoft/testsuites/network/sriov.py
LiliDeng: by default it uses 1 nic?

microsoft/testsuites/network/sriov.py
LiliDeng: udpated

microsoft/testsuites/network/sriov.py
LiliDeng: I used `self._metadata.category = "stress"`, any other way?

squirrelsc: It's like the `timeout`, `priority`, just write it in metadata like them.

lisa/tools/interrupt_inspector.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(
+        self,
+        irq_number: str,
+        interrupt_count: List[int],
+        interrupt_sum: int,
+        metadata: str = "",
+    ) -> None:
+        self.irq_number = irq_number
+        self.interrupt_count = interrupt_count
```
squirrelsc: `cpu_counter`. It doesn't need to repeat "interrupt" in the variable name.

lisa/tools/interrupt_inspector.py
```diff
@@ -0,0 +1,124 @@
...
+        self.metadata = metadata
+        self.interrupt_sum = interrupt_sum
```
squirrelsc: `counter_sum`

lisa/tools/interrupt_inspector.py
```diff
@@ -0,0 +1,124 @@
...
+
+class InterruptInspector(Tool):
+    # 0:         22          0  IR-IO-APIC   2-edge      timer
+    _interrupt_regex = re.compile(
+        r"^\s*(?P<irq_number>\S+):\s+(?P<interrupt_count>[\d+ ]+)\s*(?P<metadata>.*)$"
+    )
+
+    @property
+    def command(self) -> str:
+        return "ls -lt /proc/interrupts"
```
squirrelsc: How about put "cat" here? The cat is actual used command.

microsoft/testsuites/network/sriov.py
```diff
@@ -542,6 +554,166 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+        5. Compare interrupts changes, expected to see interrupts increased.
+        6. Get final interrupts sum per cpu on client node.
+        7. Compare interrupts changes, expected to see interrupts increased.
+        """,
+        priority=2,
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                node_count=2,
+                core_count=search_space.IntRange(min=8, max=16),
+                network_interface=schema.NetworkInterfaceOptionSettings(
```
squirrelsc: Change to simple Sriov()

microsoft/testsuites/network/sriov.py
squirrelsc: How about another one?

microsoft/testsuites/network/sriov.py
squirrelsc: Is it possible to have perf issue on second or other nics? If not, let's test only first nic.

microsoft/testsuites/network/sriov.py
LiliDeng: I didn't find category for TestCaseMetadata

microsoft/testsuites/network/sriov.py
LiliDeng: updated

microsoft/testsuites/network/sriov.py
LiliDeng: change into 30 secs

microsoft/testsuites/network/sriov.py
LiliDeng: not sure the answer, updated into testing against 1st nic

lisa/tools/interrupt_inspector.py
LiliDeng: updated

lisa/tools/interrupt_inspector.py
LiliDeng: updated

lisa/tools/interrupt_inspector.py
LiliDeng: updated

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/sriov.py
squirrelsc: I see. It's defined on test suite level. So, you can define a new class, and set it on test suite level. The two test suites can share some methods.

microsoft/testsuites/network/sriov.py
squirrelsc: The description can be clear more. Explain explicitly after the "too many and too few", like "so it chooses cpu count between 8 to 16.

microsoft/testsuites/network/sriov.py
squirrelsc: Another test case is still 1800 seconds.

microsoft/testsuites/network/sriov.py
LiliDeng: I am not 100% sure for this range, my understanding, if core is too few, the test is useless, I tested when core is many at least >=32, some cpu are idle, the case will fail (I will add one more case for this situation, to update the target CPU)

microsoft/testsuites/network/sriov.py
LiliDeng: that is stress testing

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/sriov.py
squirrelsc: Anyway, we should document what you thought. It will be much more confusing in future, if we hide the test design idea.

microsoft/testsuites/network/sriov.py
squirrelsc: Do you mean the test case test only one nic now? But the logic still has the loop on all nics.

lisa/tools/interrupt_inspector.py
```diff
@@ -0,0 +1,120 @@
...
+                    irq_number=matched.group("irq_number"),
+                    cpu_counter=cpu_counter,
+                    counter_sum=counter_sum,
+                    metadata=matched.group("metadata"),
+                )
+            )
+
+        return interrupts
+
+    def get_sum_of_interrupt_data_by_irqs(self, pci_slot: str) -> List[Dict[str, int]]:
```
squirrelsc: The name can be shorter, and reduce the terms used. For example, `sum_cpu_counter_by_irqs`

microsoft/testsuites/network/sriov.py
```diff
@@ -542,173 +529,360 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
-                    modules_exist = True
-            assert_that(modules_exist).described_as(
-                "The module of sriov network device isn't loaded."
-            ).is_true()
-
-            # 2. Check VF counts listed from lspci is expected.
-            lspci = node.tools[Lspci]
-            devices_slots = lspci.get_devices_slots(
-                constants.DEVICE_TYPE_SRIOV, force_run=True
+@TestSuiteMetadata(
```
squirrelsc: The stress case can be put in a separated file. It's clearer.

microsoft/testsuites/network/sriov.py
LiliDeng: Use this line to control https://github.com/microsoft/lisa/pull/1637/files#diff-915cefed780f89c28b257620163cf2368234c9a7620b8f8f9ba35cab955daa23R65 only run on primary nic

lisa/tools/interrupt_inspector.py
LiliDeng: updated

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/sriov.py
LiliDeng: updated

microsoft/testsuites/network/sriov.py
squirrelsc: It can be more intuitive. Put only this line in the for loop. If not found, raise an exception. If it's found, run following logic.

lisa/tools/interrupt_inspector.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def sum_cpu_counter_by_irqs(self, pci_slot: str) -> List[Dict[str, int]]:
+        interrupts_sum_by_irqs: List[Dict[str, int]] = []
+        interrupts = self.get_interrupt_data()
+        matched_interrupts = [x for x in interrupts if pci_slot in x.metadata]
+        for interrupt in matched_interrupts:
+            interrupts_sum_by_irqs.append({interrupt.irq_number: interrupt.counter_sum})
+        return interrupts_sum_by_irqs
+
+    def sum_interrupt_data_by_irqs(self, pci_slot: str) -> Dict[int, int]:
```
squirrelsc: same as above `sum_cpu_counter_by_index`?

LiliDeng: ok...

microsoft/testsuites/network/stress.py
```diff
@@ -0,0 +1,97 @@
...
+    Environment,
+    Logger,
+    RemoteNode,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    node_requirement,
+    schema,
+)
+from lisa.features import Sriov as SriovNetwork
```
squirrelsc: What's the reason to rename it? It decreases the readability of the code. If someone is familiar with the code, he knows the Sriov well, but don't know SriovNetwork, he will be confused. If the SriovNetwork is a good name, we should rename it in the source.

microsoft/testsuites/network/sriov.py
```diff
@@ -1,29 +1,39 @@
...
     TestSuite,
     TestSuiteMetadata,
+    node_requirement,
     schema,
     search_space,
     simple_requirement,
 )
 from lisa.features import NetworkInterface, SerialConsole
-from lisa.nic import NicInfo, Nics
+from lisa.features import Sriov as SriovNetwork
```
squirrelsc: If there is a conflict, you can import it like `from lisa import features`, and use it like `features.Sriov`

microsoft/testsuites/network/sriov.py

lisa/tools/interrupt_inspector.py
LiliDeng: updated

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/synthetic.py
```diff
@@ -86,6 +87,7 @@ def synthetic_add_max_nics_one_time_after_provision_validation(
         3. Check each nic has an ip address.
         """,
         priority=2,
+        use_new_environment=True,
```
squirrelsc: What's the reason they need new environments?

microsoft/testsuites/network/sriov.py
```diff
@@ -542,173 +527,106 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
-                dest_nic = dest_synthetic_nic
-            if turn_off_vf:
-                source_node.execute(f"ip link set dev {source_vf_nic} down", sudo=True)
-                dest_node.execute(f"ip link set dev {dest_vf_nic} down", sudo=True)
-
-            # get origin tx_packets and rx_packets before copy file
-            source_tx_packets_origin = self._get_packets(source_node, source_nic)
-            dest_tx_packets_origin = self._get_packets(
-                dest_node, dest_nic, "rx_packets"
+            assert matched_server_nic_info
```
squirrelsc: Add message, if it's failed, what the runner should know.

microsoft/testsuites/network/synthetic.py
LiliDeng: from current test results, it will reuse previous environment which already attached 8 nics, then case failed, and I found use_new_environment seems not working...

LiliDeng: also for any provision test case, we should use new environment?

microsoft/testsuites/network/sriov.py
LiliDeng: added

microsoft/testsuites/network/synthetic.py
squirrelsc: Every test case should recover the environment after it's completed. It can be done in `after_case`. It runs even the test case failed earlier. Let me know what's the behavior of "not working".

microsoft/testsuites/network/stress.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+        assert_that(iperf_log).described_as(
+            f"iperf client run failed on client node, {iperf_log}"
+        ).does_not_contain("error")
+
+        # 3. Do VF connection test.
+        sriov_vf_connection_test(environment, vm_nics)
+
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
+        environment: Environment = kwargs.pop("environment")
```
squirrelsc, line 109: The cleanup also sharable.

LiliDeng: updated

microsoft/testsuites/network/synthetic.py
LiliDeng: synthetic_provision_with_max_nics_validation
synthetic_add_max_nics_one_time_after_provision_validation
synthetic_add_max_nics_one_by_one_after_provision_validation
will run in sequence, the first case will provision with 8 nics, then the second case will reuse previous environment, at that time, the VM is not allowed to add any nic any more. case will fail during attach nic.
image


should we remove the nics after run case synthetic_provision_with_max_nics_validation?

LiliDeng: @squirrelsc thoughts?

squirrelsc: 1. The logic is needed in `after_case` to clean up all added nics. It makes sure the environment is recovered, and ready for next test case.
2. I mean what's the behavior of the `use_new_environment`  not working, from your comments? "I found use_new_environment seems not working..." I had a try, the use_new_environment works on my scenario.

microsoft/testsuites/network/sriov.py
squirrelsc: Not seen it in latest push.

microsoft/testsuites/network/synthetic.py
LiliDeng: synthetic_provision_with_max_nics_validation
synthetic_add_max_nics_one_time_after_provision_validation + use_new_environment 
synthetic_add_max_nics_one_by_one_after_provision_validation + use_new_environment 

I saw below two cases not creating new environment.

squirrelsc: Can you share me a log?

LiliDeng: logs sent to you by mail, 
update code to remove extra nic in after_case

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/sriov.py
```diff
@@ -545,173 +531,109 @@ def sriov_ethtool_offload_setting(self, environment: Environment) -> None:
...
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                node_count=2,
+                core_count=search_space.IntRange(min=8, max=16),
+                network_interface=features.Sriov(),
             )
-        )
+        ),
+    )
+    def sriov_interrupts_change(self, environment: Environment) -> None:
```
squirrelsc: add verify_ before each test case.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -170,6 +170,16 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
 
+    def _get_primary(self, nics: Any) -> Any:
```
squirrelsc: Don't use Any here, it breaks the benefit of typing. Try to use the type with quote, and import it only when type checking.

microsoft/testsuites/network/sriov.py
LiliDeng: added

lisa/sut_orchestrator/azure/features.py

squirrelsc (approved):

PR #1638, opened 2021-11-22
Migrate boot_with_debug_kernel test case.

LiliDeng:


microsoft/testsuites/core/boot.py
```diff
@@ -0,0 +1,86 @@
...
+    area="core",
+    category="functional",
+    description="""
+    This test suite is to test VM working well after updating on VM and rebooting.
+    """,
+)
+class Boot(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test case will
```
squirrelsc, line 34: Add explanation in test spec, why this test case work on Redhat only.

microsoft/testsuites/core/boot.py
```diff
@@ -0,0 +1,86 @@
...
+        # 3. Reboot VM, check kernel version is debug type.
+        try:
+            node.reboot()
+        except Exception as identifier:
+            serial_console = node.features[SerialConsole]
+            # if there is any panic, fail before partial pass
+            serial_console.check_panic(saved_path=log_path, stage="reboot")
+
+            # if node cannot be connected after reboot, it should be failed.
+            if isinstance(identifier, LisaException) and str(identifier).startswith(
```
squirrelsc: Let's create a new Exception called `TcpConnetionException`, it supports an argument called error code.

microsoft/testsuites/core/boot.py
```diff
@@ -0,0 +1,86 @@
...
+            )
+
+        # 2. Install kernel-debug package and set boot with this debug kernel.
+        node.os.install_packages("kernel-debug")
+        result = node.execute("grub2-set-default 0", sudo=True)
+        result.assert_exit_code()
+
+        # 3. Reboot VM, check kernel version is debug type.
+        try:
+            node.reboot()
```
squirrelsc: It can call reboot_and_check_panic. It doesn't need to duplicate logic.

microsoft/testsuites/core/boot.py
```diff
@@ -0,0 +1,86 @@
...
+            serial_console = node.features[SerialConsole]
+            # if there is any panic, fail before partial pass
+            serial_console.check_panic(saved_path=log_path, stage="reboot")
+
+            # if node cannot be connected after reboot, it should be failed.
+            if isinstance(identifier, LisaException) and str(identifier).startswith(
+                "cannot connect to TCP port"
+            ):
+                raise LisaException(f"after reboot, {identifier}")
+        is_ready, tcp_error_code = wait_tcp_port_ready(
```
squirrelsc, line 65: Because this logic appears in every reboot place. We need to move them into the reboot method.

LiliDeng: if use remotenode in reboot, will encounter circular import issue.
```
ImportError: cannot import name 'RemoteNode' from partially initialized module 'lisa' (most likely due to a circular import) (C:\CODE\internal\main\lsg-lisa\lisa\lisa\__init__.py)
```

microsoft/testsuites/core/boot.py
LiliDeng: this case was added by redhat, I also did some check, there is no debug kernel package for ubuntu, just debug symbol.

microsoft/testsuites/core/boot.py
LiliDeng: added

microsoft/testsuites/core/boot.py
LiliDeng: thanks, updated

microsoft/testsuites/core/boot.py
```diff
@@ -0,0 +1,79 @@
...
+        1. Skip testing if the distro is not redhat type.
+        2. Install kernel-debug package and set boot with this debug kernel.
+        3. Reboot VM, check kernel version is debug type.
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            supported_features=[SerialConsole],
+        ),
+    )
+    def boot_with_debug_kernel(
```
squirrelsc: Let it start with `verify_`

microsoft/testsuites/core/provisioning.py
```diff
@@ -251,9 +253,11 @@ def _smoke_test(
                 if not is_ready:
                     serial_console = node.features[SerialConsole]
                     serial_console.check_panic(saved_path=log_path, stage="reboot")
-                    raise LisaException(
-                        f"Cannot connect to [{node.public_address}:{node.public_port}],"
-                        f" error code: {tcp_error_code}, no panic found in serial log"
+                    raise TcpConnetionException(
```
squirrelsc, line 256: This message should be different with previous one in Ln 228. They should specify it happens on boot or reboot.

microsoft/testsuites/core/boot.py
squirrelsc: So, add description in code. Others may have the same question.

microsoft/testsuites/core/boot.py
squirrelsc: Refer to `uname.py` or other classes, which import Node. Import it in TYPE_CHECKING, and use it with quotes in declaration. Unless the type is used in `isinstance`, otherwise, it should be ok.

microsoft/testsuites/core/boot.py
```diff
@@ -0,0 +1,79 @@
...
+        is_ready, tcp_error_code = wait_tcp_port_ready(
+            node.public_address, node.public_port, log=log
+        )
+        if is_ready:
+            uname = node.tools[Uname]
+            kernel_version = uname.get_linux_information(
+                force_run=True
+            ).kernel_version_raw
+            assert_that(
+                kernel_version, "Kernel version is not debug after reboot."
```
squirrelsc: what means "not debug after reboot". Does it mean debug is not enabled after reboot?

microsoft/testsuites/core/boot.py
LiliDeng: added

microsoft/testsuites/core/provisioning.py
LiliDeng: updated

microsoft/testsuites/core/boot.py
LiliDeng: updated

microsoft/testsuites/core/boot.py
LiliDeng: updated

microsoft/testsuites/core/boot.py
```diff
@@ -0,0 +1,81 @@
...
+    category="functional",
+    description="""
+    This test suite is to test VM working well after updating on VM and rebooting.
+    """,
+)
+class Boot(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Skip testing if the distro is not redhat/centos type, RHEL added this test
```
squirrelsc, line 35: This information shouldn't be a step. We need to specify it's a RedHat contributed test case in the first line of description. Please also set the owner to "RedHat".

LiliDeng: under microsoft folder?

squirrelsc: Yes, we migrated it. They can move to their folder, if they want later.

LiliDeng: updated owner to RedHat

squirrelsc (approved):

PR #1640, opened 2021-11-25
Fix issue when build upstream kernel

LiliDeng:


lisa/operating_system.py
```diff
@@ -282,15 +282,20 @@ def install_packages(
...
+    def package_installed(self, package: Union[str, Tool, Type[Tool]]) -> bool:
         """
         Query if a package/tool is installed on the node.
         Return Value - bool
         """
         package_name = self.__resolve_package_name(package)
-        return self._package_exists(package_name)
+        return self._package_installed(package_name)
+
+    def package_existed(self, package: str) -> bool:
```
squirrelsc: `is_package_in_repo`

lisa/operating_system.py
```diff
@@ -282,15 +282,20 @@ def install_packages(
         package_names = self._get_package_list(packages)
         self._install_packages(package_names, signed)
 
-    def package_exists(
-        self, package: Union[str, Tool, Type[Tool]], signed: bool = True
-    ) -> bool:
+    def package_installed(self, package: Union[str, Tool, Type[Tool]]) -> bool:
```
squirrelsc: It doesn't need to change the name, if the next one is renamed.

lisa/operating_system.py
```diff
@@ -282,15 +282,20 @@ def install_packages(
...
         package_name = self.__resolve_package_name(package)
-        return self._package_exists(package_name)
+        return self._package_installed(package_name)
+
+    def package_existed(self, package: str) -> bool:
+        """
+        Query if a package/tool exists in the repo
+        Return Value - bool
+        """
+        return self._package_existed(package)
```
squirrelsc: Does it need to call `package_name = self.__resolve_package_name(package)`?

lisa/operating_system.py
```diff
@@ -578,6 +586,42 @@ def _package_exists(self, package: str, signed: bool = True) -> bool:
             return True
         return False
 
+    def _package_existed(self, package: str) -> bool:
+        command = f"apt-cache policy {package}"
+        result = self._node.execute(command, sudo=True, shell=True)
+        package_pattern = re.compile(r"Candidate: ((?!none)).*", re.M)
```
squirrelsc: move it to class level.

lisa/operating_system.py
LiliDeng: package_exists is not clear, so I want to rename it

lisa/operating_system.py
LiliDeng: to be frank, I am not clear what is the usage of __resolve_package_name

squirrelsc: Some packages are passed in as Tool type, not a string. The method will convert it to a package str by calling the `package_name` method.

lisa/operating_system.py
squirrelsc: it's a method on os, so it means if the package exists in the current os.

lisa/operating_system.py

lisa/operating_system.py
LiliDeng: revert

lisa/operating_system.py

lisa/operating_system.py
LiliDeng: moved

lisa/tools/xfstests.py
```diff
@@ -130,7 +130,8 @@ def _install_dep(self) -> None:
         # be installed if one of packages is not available in that distro,
         # so here install it one by one.
         for package in list(package_list):
-            posix_os.install_packages(package)
+            if posix_os.is_package_in_repo(package):
```
squirrelsc, line 136: What's the else? raise an exception?

squirrelsc: If not to raise an exception, add comments in code to explain it.

LiliDeng: added.

lisa/tools/xfstests.py
```diff
@@ -128,9 +128,13 @@ def _install_dep(self) -> None:
...
         # be installed if one of packages is not available in that distro,
-        # so here install it one by one.
+        # so here install it one by one
         for package in list(package_list):
-            posix_os.install_packages(package)
+            # to make code simple, put all packages needed by one distro in one list.
+            # the package name may be different for the different sku of the
+            #  same distro. so, install it when the package exists in the repo.
+            if posix_os.is_package_in_repo(package):
+                posix_os.install_packages(package)
```
squirrelsc, line 137: Another way is to add a boolean on `install_packages` called `ignore_non_exists`. It makes the logic simpler, and easy to read.

squirrelsc (approved):

lisa/tools/xfstests.py
LiliDeng: I will raise another PR for it, use #1641 to track.

PR #1642, opened 2021-12-02
Add disk adding test cases

dsrivastavv:
Migrated https://github.com/microsoft/lisa/blob/master/Testscripts/Windows/DiskHotAddRemove-Serial.ps1
Migrated https://github.com/microsoft/lisa/blob/master/Testscripts/Windows/DiskHotAddRemove-Parallel.ps1

### DIFF
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 from functools import partial
-from typing import List, Type
+from typing import Any, List, Optional, Type
 
 from assertpy.assertpy import assert_that
 
@@ -43,6 +43,20 @@ class Disk(Feature):
 
     def get_raw_data_disks(self) -> List[str]:
         raise NotImplementedError
+
+    def add_data_disk(
+        self,
+        count: int,
+        disk_type: schema.DiskType = schema.DiskType.StandardHDDLRS,
+        size_in_gb: int = 20,
+    ) -> List[str]:
+        raise NotImplementedError
+
+    def remove_data_disk(self, name: Optional[List[str]] = None) -> None:
+        raise NotImplementedError
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.disks: List[str] = []
 
 
 DiskEphemeral = partial(schema.DiskOptionSettings, disk_type=schema.DiskType.Ephemeral)
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -467,6 +467,12 @@ class AzureDiskOptionSettings(schema.DiskOptionSettings):
             ),
             "has_resource_disk",
         )
+        result.merge(
+            search_space.check_countspace(
+                self.max_data_disk_count, capability.max_data_disk_count
+            ),
+            "max_data_disk_count",
+        )
 
         return result
 
@@ -503,6 +509,16 @@ class AzureDiskOptionSettings(schema.DiskOptionSettings):
         if self.data_disk_count is not None or capability.data_disk_count is not None:
             min_value.data_disk_count = search_space.generate_min_capability_countspace(
                 self.data_disk_count, capability.data_disk_count
+            )
+
+        if (
+            self.max_data_disk_count is not None
+            or capability.max_data_disk_count is not None
+        ):
+            min_value.max_data_disk_count = (
+                search_space.generate_min_capability_countspace(
+                    self.max_data_disk_count, capability.max_data_disk_count
+                )
             )
 
         disk_type_iops = _disk_size_iops_map.get(min_value.disk_type, None)
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -9,7 +9,11 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, cast
 
 import requests
 from assertpy import assert_that
-from azure.mgmt.compute.models import NetworkInterfaceReference  # type: ignore
+from azure.mgmt.compute.models import (  # type: ignore
+    DiskCreateOption,
+    DiskCreateOptionTypes,
+    NetworkInterfaceReference,
+)
 from dataclasses_json import dataclass_json
 from PIL import Image, UnidentifiedImageError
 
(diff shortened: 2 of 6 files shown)


lisa/tools/lsblk.py
```diff
@@ -45,7 +45,7 @@ class Lsblk(Tool):
     # NAME="loop2" SIZE="34017280" TYPE="loop" MOUNTPOINT="/snap/snapd/13640"
     _LSBLK_ENTRY_REGEX = re.compile(
         r'NAME="(?P<name>\S+)"\s+SIZE="(?P<size>\d+)"\s+'
-        r'TYPE="(?P<type>\S+)"\s+MOUNTPOINT="(?P<mountpoint>\S+)"'
+        r'TYPE="(?P<type>\S+)"\s+MOUNTPOINT="(?P<mountpoint>\S*)"'
```
squirrelsc, line 48: Split it to a separated commit. So it can be described in one line by commit message.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -596,6 +600,103 @@ def get_raw_data_disks(self) -> List[str]:
             disk_array[int(disk.split("/")[-1].replace("lun", ""))] = cmd_result.stdout
         return disk_array
 
+    def add_managed_data_disk(
+        self,
+        names: List[str],
+        luns: List[str],
+        size: str = "20",
+        location: str = "westus2",
```
squirrelsc: The location can be gotten from node context, not pass it in as a parameter.

dsrivastavv: Awesome. Thanks!

microsoft/testsuites/core/storage.py
```diff
@@ -242,6 +245,155 @@ def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
...
+        Steps:
+        1. Get maximum number of data disk for the current vm_size.
+        2. Get the number of data disks already added to the vm.
+        3. Add maximum number of data disks to the VM in parallel.
+        4. Verify that the disks are added are available in the OS.
+        5. Remove the disks from the vm in parallel.
+        6. Verify that the disks are removed from the OS.
+        """,
+        requirement=simple_requirement(supported_platform_type=[AZURE]),
+    )
```
LiliDeng: Should we implement max_disk_count similar with max_nic_count?
Then have separate test cases which can test the VM sku with max number of data disk currently on azure, 64.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -596,6 +600,103 @@ def get_raw_data_disks(self) -> List[str]:
...
+            async_disk_update = compute_client.disks.begin_create_or_update(
+                self._resource_group_name,
+                name,
+                {
+                    "location": location,
+                    "disk_size_gb": size,
+                    "creation_data": {"create_option": DiskCreateOption.empty},
+                },
+            )
+            managed_disks.append(async_disk_update.result())
```
LiliDeng, line 697: should we have separate cases for different disk type sku?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -596,6 +600,103 @@ def get_raw_data_disks(self) -> List[str]:
             disk_array[int(disk.split("/")[-1].replace("lun", ""))] = cmd_result.stdout
         return disk_array
 
+    def add_managed_data_disk(
```
squirrelsc: We can provide an abstract method in LISA level feature like `def add_data_disk(count:int, disk_type:DiskType, size_in_gb:int=20) -> List[str]`. The disk type parameter is platform specific, but we can reuse the DiskType in schema.py so far.

The names aren't matter, so we can decide it in the method and return it later. The luns can be checked in this method.

microsoft/testsuites/core/storage.py
```diff
@@ -242,6 +245,155 @@ def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
...
+            ]
+            log.debug(f"added_partitions: {added_partitions}")
+            assert_that(added_partitions).is_length(1)
+            assert_that(added_partitions[0].size).is_equal_to(
+                disk_size_in_gb * 1024 * 1024 * 1024
+            )
+
+            # remove data disk
+            log.debug(f"Removing managed disk: {disk_id}")
+            disk.remove_managed_data_disk([disk_id])
```
squirrelsc: It should be called in `after_case`. So if the test failed on previous steps, they can be removed too.

microsoft/testsuites/core/storage.py
```diff
@@ -242,6 +245,155 @@ def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
...
+            # verify that partition count is decreased by 1
+            partition_after_removing_disk = node.tools[Lsblk].get_partitions(
+                force_run=True
+            )
+            added_partitions = [
+                item
+                for item in partitions_before_adding_disk
+                if item not in partition_after_removing_disk
+            ]
+            assert_that(added_partitions).is_length(0)
```
squirrelsc: Always add description for assertion.

lisa/sut_orchestrator/azure/features.py
dsrivastavv: I think this makes sense. I will add them separately for each disk type

microsoft/testsuites/core/storage.py
dsrivastavv: This is part of the test case itself. The test basically check that disks can be removed. How about I send `BadEnvironmentStateException` if this steps fails for any reason?

lisa/sut_orchestrator/azure/features.py
dsrivastavv: Are luns valid outside of azure or are they azure specific?

microsoft/testsuites/core/storage.py
dsrivastavv: The number of data disks is variable for each vm_size. For example, `Standard_DS2_v2` can only have 8 max data disks. Please feel free to clarify if I misunderstood your comment.

microsoft/testsuites/core/storage.py
squirrelsc: The `after_case` will be called after each test case run, it's designed for the cleanup purpose. raise exception overwrites the original failure exception and it needs a big try...except....

lisa/sut_orchestrator/azure/features.py
squirrelsc: I don't know, it maybe appliable to all platforms. But it doesn't need to be calculated at test case level.

microsoft/testsuites/core/storage.py
squirrelsc: Lili is trying to handle the same problem. If the test case will create 9 data disks on Standard_DS2_v2, it will failed. But we can use the max_data_disk_count to skip this vm size, if a test case declares it needs 9 data disks. Refer to the code of max_nic_count.

microsoft/testsuites/core/storage.py
dsrivastavv: I need to run `disk.remove_managed_data_disk` as part of the test. There is a check after this which calls `lsblk` to verify that disks have been successfully removed. From a naming point of view, `after_case` sounds something where we just put cleanup operations that are not directly related to the test. Maybe we can discuss this offline to get more clarity

microsoft/testsuites/core/storage.py
LiliDeng: We also need to consider the test coverage; current design may only cover the minimal data disk count scenario.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -44,6 +48,8 @@ def _initialize_information(self, node: Node) -> None:
         node_context = get_node_context(node)
         self._vm_name = node_context.vm_name
         self._resource_group_name = node_context.resource_group_name
+        self._location = node_context.location
```
squirrelsc: How about save only `self._node_context`? It doesn't need to add one by one in future.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1416,8 +1418,9 @@ def _resource_sku_to_capability(  # noqa: C901
                     nvme.disk_count = int(node_space.core_count / 8)
                     node_space.features.add(nvme)
             elif name == "MaxDataDiskCount":
+                node_space.disk.max_data_disk_count = int(sku_capability.value)
```
squirrelsc, line 1420: Split each meaningful individual change to commit as much as possible. It helps people to understand changes better. We spend a little bit more time during development, like organize changes, add more comments, it's investment to future.

microsoft/testsuites/core/storage.py
```diff
@@ -242,6 +251,244 @@ def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
...
+        self.added_data_disks: List[str] = []
+
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
+        node: Node = kwargs["node"]
+        disk = node.features[Disk]
+
+        # cleanup any disks added as part of the test
+        # If the cleanup operation fails, mark node to be recycled
+        try:
+            disk.remove_data_disk(self.added_data_disks)
```
squirrelsc: Don't use the variables of `self`. The test cases may run in different nodes, there is conflict potentially. Because the disks names are decided in feature, so feature should be able to know which disk need to be removed. The `remove_data_disk` can accept empty arguments, it means remove all added disks.

microsoft/testsuites/core/storage.py
```diff
@@ -242,6 +251,244 @@ def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
...
+        1. Get maximum number of data disk for the current vm_size.
+        2. Get the number of data disks already added to the vm.
+        3. Add maximum number of data disks to the VM in parallel.
+        4. Verify that the disks are added are available in the OS.
+        5. Remove the disks from the vm in parallel.
+        6. Verify that the disks are removed from the OS.
+        """,
+        requirement=simple_requirement(disk=DiskStandardHDDLRS()),
+    )
+    def hot_add_disk_parallel(self, log: Logger, node: Node) -> None:
```
squirrelsc, line 311: What does "parallel" mean here?

dsrivastavv: It's basically that all the disks are added in one go rather than one by one

dsrivastavv: I will update the steps to make it more clear

lisa/sut_orchestrator/azure/platform_.py
dsrivastavv: I will try to separate them into individual commits. The current code had multiple changes in same file which were related so I had put them in single commit

lisa/sut_orchestrator/azure/common.py
```diff
@@ -46,9 +46,11 @@ class EnvironmentContext:
 class NodeContext:
     resource_group_name: str = ""
     vm_name: str = ""
+    vm_size = ""
```
squirrelsc: I double checked the code, they should come from node.capability.get_extended_runbook, not from context. It avoids duplicating those values.

lisa/sut_orchestrator/azure/features.py
squirrelsc: as new finding, it should be self._azure_runbook

lisa/sut_orchestrator/azure/features.py
```diff
@@ -596,13 +618,97 @@ def get_raw_data_disks(self) -> List[str]:
...
+                    "creation_data": {"create_option": DiskCreateOption.empty},
+                },
+            )
+            managed_disks.append(async_disk_update.result())
+
+        # attach managed disk
+        vm = compute_client.virtual_machines.get(
+            self._resource_group_name, self._vm_name
+        )
+        for i, managed_disk in enumerate(managed_disks):
```
squirrelsc, line 703: It's better to validate the existing disk count is the same as capability.disk.data_disk_count. If not, raise an exception.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: You can use `git add -p <file name>` to commit partial changes of a file. I did that a lot.

lisa/sut_orchestrator/azure/features.py
dsrivastavv: This might not always match. There could be disks added from the provisioning phase.

lisa/features/disks.py
```diff
@@ -44,6 +44,17 @@ def get_partition_with_mount_point(self, mount_point: str) -> PartitionInfo:
     def get_raw_data_disks(self) -> List[str]:
         raise NotImplementedError
 
+    def add_data_disk(
+        self, count: int, disk_type: schema.DiskType, size_in_gb: int
```
squirrelsc: size_in_gb can be optional with a default value 20. I'm thinking about the disk_type even can have a default value. It helps the method is easier to use.

dsrivastavv: I can use StandardHDD as default disk size

squirrelsc (approved):

PR #1644, opened 2021-12-03
package manager installation for dpdk testpmd

mcgov:
- hoist loading drivers and identifying network hardware into their
        own functions
    - enable installation of dpdk from package managers
    - enable grabbing version info from tarball and git source tag

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -3,15 +3,16 @@
 
 import re
 from pathlib import PurePath
-from typing import List, Tuple, Type
+from typing import List, Pattern, Tuple, Type, Union
 
 from assertpy import assert_that, fail
+from semver import VersionInfo
 
 from lisa.executable import Tool
 from lisa.nic import NicInfo
 from lisa.operating_system import CentOs, Redhat, Ubuntu
-from lisa.tools import Echo, Git, Lspci, Tar, Wget
-from lisa.util import LisaException, SkippedException
+from lisa.tools import Echo, Git, Lspci, Modprobe, Tar, Wget
+from lisa.util import LisaException, UnsupportedDistroException
 
 
 class DpdkTestpmd(Tool):
@@ -27,6 +28,15 @@ class DpdkTestpmd(Tool):
             r"[a-fA-F0-9]{2}\.[a-fA-F0-9]"
             r" \(socket 0\)"
         )
+    )
+
+    # ex v19.11-rc3 or 19.11
+    _version_info_from_git_tag_regex = re.compile(
+        r"v?(?P<major>[0-9]+)\.(?P<minor>[0-9]+)"
+    )
+    # ex dpdk-21.08.zip or dpdk-21.08-rc4.zip
+    _version_info_from_tarball_regex = re.compile(
+        r"dpdk-(?P<major>[0-9]+)\.(?P<minor>[0-9]+)"
     )
 
     @property
@@ -66,12 +76,6 @@ class DpdkTestpmd(Tool):
     ]
 
     _redhat_packages = [
-        "gcc",
-        "make",
-        "git",
-        "tar",
-        "wget",
-        "dos2unix",
         "psmisc",
         "kernel-devel-$(uname -r)",
         "numactl-devel.x86_64",
@@ -94,9 +98,14 @@ class DpdkTestpmd(Tool):
         self, cmd: str, path: PurePath, timeout: int = 600
     ) -> str:
         result = self.node.execute(
-            cmd, sudo=True, shell=True, cwd=path, timeout=timeout
+            cmd,
+            sudo=True,
+            shell=True,
+            cwd=path,
+            timeout=timeout,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="failed with nonzero error code",
         )
-        assert_that(result.exit_code).is_zero()
         return result.stdout
 
     @property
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -94,9 +98,16 @@ def __execute_assert_zero(
...
         result = self.node.execute(
-            cmd, sudo=True, shell=True, cwd=path, timeout=timeout
+            cmd,
+            sudo=True,
+            shell=True,
+            cwd=path,
+            timeout=timeout,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"{cmd} failed with nonzero error code"
```
squirrelsc: The {cmd} will be printed in the message, so it doesn't need to be printed again.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -116,18 +127,50 @@ def set_dpdk_source(self, dpdk_source: str) -> None:
...
+            )
+            if isinstance(node.os, Ubuntu) or isinstance(node.os, Redhat):
+                node.os.install_packages(["dpdk", "dpdk-dev"])
+                self._dpdk_version_info = node.os.get_package_information("dpdk")
+            else:
+                raise NotImplementedError(
+                    f"Dpdk install isn't implemented for Os {node.os.name}"
+                )
+
+            if self._dpdk_version_info >= VersionInfo(19, 11):
```
squirrelsc: `if self._dpdk_version_info >= "19.11":`

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -177,20 +226,74 @@ def _install(self) -> bool:
...
             mellanox_drivers = "mlx4_core mlx4_ib"
         else:
             mellanox_drivers = "mlx5_core mlx5_ib"
+        if isinstance(self.node.os, Ubuntu):
+            self.node.execute(
+                "modprobe -a rdma_cm rdma_ucm ",
+                cwd=self.node.working_path,
+                expected_exit_code=0,
+                expected_exit_code_failure_message=(
+                    "Could not load rdma cm/ucm drivers"
```
squirrelsc: If the `stdout` is helpful, it can be included here. But if it's too long, consider extracting the key lines.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -177,20 +226,74 @@ def _install(self) -> bool:
...
+            self.node.execute(
+                f"modprobe -a  mlx4_en {mellanox_drivers}",
+                cwd=self.node.working_path,
+                expected_exit_code=0,
+                expected_exit_code_failure_message=(
+                    "Encountered issue loading infiniband and "
+                    "mellanox drivers (redhat)."
+                ),
+                sudo=True,
+            )  # add mellanox drivers
```
squirrelsc: How about else? If other distros are not supported, raise an unsupported exception.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -233,15 +332,6 @@ def _install_dependencies(self) -> None:
...
-            if not self.is_connect_x3:
-                self.__execute_assert_zero(
-                    f"dracut --add-drivers '{mellanox_drivers} ib_uverbs' -f", cwd
-                )
-            self.__execute_assert_zero(
-                "modprobe -a ib_core ib_uverbs mlx4_en mlx4_core mlx5_core mlx4_ib"
-                + " mlx5_ib",
-                cwd,
-            )  # add mellanox drivers
             self.__execute_assert_zero("systemctl enable rdma", cwd)
```
squirrelsc, line 314: Let's replace them with execute and expected_exit_code pattern.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I was going to get rid of this function in my next commit to give more specific error messages on all these build commands. Can we leave this and I'll get rid of it in the next PR?

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I don't think this command dumps anything, if it fails it just returns an error code on failure.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I agree but let's do it in a new PR

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Why does the UnsupportedDistroException make the test class parse the OS version info? That seems like the os class should contain that already and you should just pass in node.os or something

microsoft/testsuites/dpdk/dpdktestpmd.py
squirrelsc: Sorry, I just remember there is a tool `modprobe` already, please reuse it. It can use `-v` parameter to output, but it's not added so far.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: (changed this one)

microsoft/testsuites/dpdk/dpdktestpmd.py
squirrelsc: Can you change it to use modprobe tool?

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -177,20 +224,79 @@ def _install(self) -> bool:
...
+    def _load_drivers_for_dpdk(self) -> None:
+        self.node.log.info("Loading drivers for infiniband, rdma, and mellanox hw...")
         if self.is_connect_x3:
             mellanox_drivers = "mlx4_core mlx4_ib"
         else:
             mellanox_drivers = "mlx5_core mlx5_ib"
+        modprobe = self.node.tools[Modprobe]
+        if isinstance(self.node.os, Ubuntu):
+            modprobe.run(
+                "-v -a rdma_cm rdma_ucm ",
```
squirrelsc: How about call the `modprobe.load`? You can add -a supports there to load multiple modules. Let's keep improving the tools, so it can be reused and save time in future.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -177,20 +224,56 @@ def _install(self) -> bool:
...
+
+    def _load_drivers_for_dpdk(self) -> None:
+        self.node.log.info("Loading drivers for infiniband, rdma, and mellanox hw...")
         if self.is_connect_x3:
             mellanox_drivers = "mlx4_core mlx4_ib"
         else:
             mellanox_drivers = "mlx5_core mlx5_ib"
+        modprobe = self.node.tools[Modprobe]
+        if isinstance(self.node.os, Ubuntu):
+            modprobe.load("-a rdma_cm rdma_ucm ")
```
squirrelsc: It works, but looks wired. 😄 Can we add the -a into load by default? It looks safe, even add one module only.

mcgov: Refactored! You were right lol

squirrelsc (approved):

PR #1645, opened 2021-12-03
Add one more clock source for new host version.

LiliDeng:


microsoft/testsuites/core/timesync.py
```diff
@@ -130,6 +133,7 @@ def timesync_check_unbind_clocksource(self, node: Node) -> None:
             "hyperv_clocksource_tsc_page",
             "lis_hyperv_clocksource_tsc_page",
             "hyperv_clocksource",
+            "tsc",
```
squirrelsc, line 136: Is it possible the "tsc" is invalid if there is hyperv_clocksource_tsc_page?

squirrelsc: Should this test case be an Azure specified test case? If so, add Azure in the metadata.

LiliDeng: not only for azure platform

squirrelsc: There are two questions.
1. Is it possible the "tsc" is invalid if there is hyperv_clocksource_tsc_page? I mean if it's possible two of the expected source exist together. Especially for tsc, it looks like a general name.
2. Should this test case be an Azure specified test case? If so, add Azure in the metadata. For "not only for azure platform", does it mean the test case can run on Hyper-V only, or it can run on other cloud platforms also?

LiliDeng: from the mail, I saw tsc is to replace hyperv_clocksource_tsc_page, my understanding is that they are not expected to see exist together. here we check current clock source, only one value can be there.

the case can ran on both hyperv and azure platform.

squirrelsc (approved):

PR #1646, opened 2021-12-03
Support ignore_non_exists during install_packages.

### LINKED ISSUE #1641: Add a boolean on install_packages called ignore_non_exists
It makes the logic simpler, and easy to read.

LiliDeng:
Fix #1641

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -291,9 +291,10 @@ class Posix(OperatingSystem, BaseClassMixin):
         self,
         packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]],
         signed: bool = True,
+        ignore_non_exists: bool = False,
     ) -> None:
         package_names = self._get_package_list(packages)
-        self._install_packages(package_names, signed)
+        self._install_packages(package_names, signed, ignore_non_exists)
 
     def package_exists(self, package: Union[str, Tool, Type[Tool]]) -> bool:
         """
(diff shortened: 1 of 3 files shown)


lisa/operating_system.py
```diff
@@ -346,7 +347,19 @@ def get_package_information(
     def get_repositories(self) -> List[RepositoryInfo]:
         raise NotImplementedError("get_repositories is not implemented")
 
-    def _install_packages(self, packages: List[str], signed: bool = True) -> None:
+    def remove_not_existed_packages(self, packages: List[str]) -> List[str]:
```
squirrelsc, line 350: Can we just ignore errors from the installation log? It would be much faster than check each package.

LiliDeng: I don't understand 'ignore errors from the installation log', is it 
try install, find failure, exclude the ones which cause the failure?

squirrelsc: No, it should only ignore errors if a package doesn't exist. Because the installation may include various package names for different distros together.

LiliDeng: no, like I code in xfstests,if install A,B,C,D in one command, and if C doesn't exist, then the entire process will be terminated, in my knowledge; but I will double check and see.

squirrelsc: ok, I see. Is there a way to check multiple packages in one command?

LiliDeng: I'll try to figure it out.

LiliDeng: I didn't find a way to install them in one command.
Maybe we can just install them one by one without checking existed or not when ignore_non_exists set as true?

squirrelsc: We can check them like current way, if the ignore_non_exists is set to True. Install one by one is lower performance, and may miss other kind of errors too. In another hand, we shouldn't overuse this, we should try our best to match package names and distros.

PR #1647, opened 2021-12-03
Update stress-ng

kamalca:
[Stress-ng](https://github.com/ColinIanKing/stress-ng) V0.07.16 no longer works with newer version of gcc 11.2.0

### DIFF
--- a/Testscripts/Linux/Linux-Kernel-Self-Tests.sh
+++ b/Testscripts/Linux/Linux-Kernel-Self-Tests.sh
@@ -81,14 +81,18 @@ function install_dependencies() {
             deb_packages=(make git gcc flex bison clang llvm fuse gcc-multilib libfuse2 \
                         libc6-i386 libc6-dev-i386 libelf-dev libcap-ng-dev libfuse-dev \
                         libpopt-dev libnuma-dev libmount-dev libcap-dev build-essential \
-                        pkg-config bc rsync)
+                        pkg-config bc rsync libaio-dev libapparmor-dev libattr1-dev libbsd-dev \
+                        libgcrypt11-dev libipsec-mb-dev libjudy-dev libkeyutils-dev \
+                        libsctp-dev libatomic1 zlib1g-dev libkmod-dev)
             LogMsg "Dependency package names： ${deb_packages[@]}"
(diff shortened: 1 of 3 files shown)


squirrelsc (approved):

XML/Other/ReplaceableTestParameters.xml
```diff
@@ -409,7 +409,7 @@ Scenario 2 : You need to run all the performance tests quickly.
 	</Parameter>
 	<Parameter>
 		<ReplaceThis>APP_GIT_TAG</ReplaceThis>
-		<ReplaceWith>V0.07.16</ReplaceWith>
+		<ReplaceWith>V0.13.07</ReplaceWith>
```
johnsongeorge-w, line 412: Why this change?

Testscripts/Linux/Linux-Kernel-Self-Tests.sh
```diff
@@ -81,14 +81,18 @@ function install_dependencies() {
...
+                        libgcrypt11-dev libipsec-mb-dev libjudy-dev libkeyutils-dev \
+                        libsctp-dev libatomic1 zlib1g-dev libkmod-dev)
             LogMsg "Dependency package names： ${deb_packages[@]}"
             install_package "${deb_packages[@]}" >> $BUILDING_LOG 2>&1
             ;;
         centos_7 | centos_8 | redhat_7 | redhat_8)
             rpm_packages=(make git gcc flex bison clang llvm fuse libcap-ng-devel popt-devel \
                         libcap-devel glibc-devel.*i686 fuse-devel elfutils-devel \
-                        numactl-devel glibc-devel)
+                        numactl-devel libaio-devel libattr-devel libbsd-devel \
```
johnsongeorge-w, line 93: This changes may not be required.

johnsongeorge-w (changes requested):

kamalca:
These changes are flawed. The new version of gcc is only applicable to internal project at this time so I will be making a pull request to internal repro for these changes. We should make a note that this update to stress-ng will be necessary for LISAv3 is we continue to use the same tool for stress testing as we should be supporting newer version of gcc in our new version of LISA.

PR #1648, opened 2021-12-04
Test case: add xdp and other refactoring

squirrelsc:
see commits for details

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -649,6 +649,30 @@ class Debian(Linux):
 
         return repositories
 
+    def add_repository(
+        self, url: str, codename: str, packages: str = "main", key_location: str = ""
+    ) -> None:
+        if key_location:
+            wget = self._node.tools[Wget]
+            key_file_path = wget.get(
+                url=key_location,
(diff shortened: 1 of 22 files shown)


lisa/operating_system.py
```diff
@@ -649,6 +649,30 @@ def get_repositories(self) -> List[RepositoryInfo]:
...
+            self._node.execute(
+                cmd=f"apt-key add {key_file_path}",
+                sudo=True,
+                expected_exit_code=0,
+                expected_exit_code_failure_message="fail to add apt key",
+            )
+        # This command will trigger apt update too, so it doesn't need to update
+        # repos again.
+        self._node.execute(
+            cmd=f'apt-add-repository "deb {url} {codename} {packages}"',
```
LiliDeng, line 670: add "-y"?

LiliDeng (dismissed):

lisa/operating_system.py
squirrelsc: which condition does it need? The current command works without it. Let me know, if there are some cases doesn't work, I will fix it in next PR.

PR #1650, opened 2021-12-06
Migrate xfstesting_azure_file_share_validation 

LiliDeng:


LiliDeng:
check failed for known issue, https://github.com/Azure/azure-sdk-for-python/issues/21223 @squirrelsc we need update the azure python sdk version?

squirrelsc:
> check failed for known issue, [Azure/azure-sdk-for-python#21223](https://github.com/Azure/azure-sdk-for-python/issues/21223) @squirrelsc we need update the azure python sdk version?

Yes, please update it.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -420,6 +421,21 @@ def check_or_create_resource_group(
                 )
 
 
+def get_location_from_resource_group(
```
squirrelsc: It's not used, is it useful? It's doesn't matter that where the resource group is.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -442,6 +458,71 @@ def wait_copy_blob(
     log.debug("vhd copied")
 
 
+def connection_string(account: str, key: str):
```
squirrelsc: 1. as our practice, use the f string
2. The method name should specify what the connection string is for.

lisa/sut_orchestrator/azure/common.py
LiliDeng: oh seems, I missed one commit

lisa/sut_orchestrator/azure/common.py
```diff
@@ -442,6 +458,71 @@ def wait_copy_blob(
...
+    account_name: str,
+    resource_group_name: str,
+) -> ShareServiceClient:
+    shared_key_credential = get_storage_credential(
+        credential=credential,
+        subscription_id=subscription_id,
+        account_name=account_name,
+        resource_group_name=resource_group_name,
+    )
+    return ShareServiceClient.from_connection_string(
```
squirrelsc: As the example, the shared_key_credential is enough to create the ShareServiceClient. It's like to create BlobServiceClient in code.
https://docs.microsoft.com/en-us/azure/developer/python/sdk/storage/azure-storage-file-share/azure.storage.fileshare.shareserviceclient?view=storage-py-v12

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -289,13 +335,92 @@ def xfstesting_btrfs_nvme_datadisk_validation(
...
+            supported_platform_type=[AZURE],
+        ),
+        timeout=TIME_OUT,
+        priority=3,
+    )
+    def xfstesting_azure_file_share_validation(
+        self, log: Logger, node: Node, environment: Environment, log_path: Path
+    ) -> None:
+        try:
+            if isinstance(environment.platform, AzurePlatform):
```
squirrelsc: 1. The test case is limited to Azure already, it doesn't need the if
2. if the `if` is needed, don't miss `else` part. It's easy to leak bugs. It's a general rule, unless in some special case. Here, if no `else`, the `else` platform will pass here, it's dangerous for a test case.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -289,13 +335,92 @@ def xfstesting_btrfs_nvme_datadisk_validation(
...
+        finally:
+            for share_name in [file_share_name, scratch_name]:
+                delete_file_share(
+                    credential=platform.credential,
+                    subscription_id=platform.subscription_id,
+                    account_name=storage_account_name,
+                    file_share_name=share_name,
+                    resource_group_name=resource_group_name,
+                )
+            node.execute("cp -f /etc/fstab_cifs /etc/fstab", sudo=True)
```
squirrelsc, line 425: Add comments to explain why it's needed in the finally section.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -326,14 +451,24 @@ def _execute_xfstests(
...
                 )
 
-        _configure_disk(
-            node,
-            data_disk,
-            {test_dev: _test_folder, scratch_dev: _scratch_folder},
-            file_system=file_system,
-        )
+        if test_type != "cifs":
+            _configure_disk(
```
squirrelsc: Add comments to explain what the logic is for.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -326,14 +451,24 @@ def _execute_xfstests(
...
-        xfstests.set_local_config(scratch_dev, _scratch_folder, test_dev, _test_folder)
+        xfstests.set_local_config(
+            scratch_dev,
+            _scratch_folder,
+            test_dev,
+            _test_folder,
+            test_type,
+            _mount_opts,
+        )
+        if test_type == "cifs":
```
squirrelsc: Add comments, it's hard to understand why cifs will be changed to generic.

lisa/sut_orchestrator/azure/common.py
LiliDeng: deleted

lisa/sut_orchestrator/azure/common.py
LiliDeng: deleted this method

lisa/sut_orchestrator/azure/common.py

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: add else, if remove this line, I saw many errors, like
`"credential" is not a known member of "None"`

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: added

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: added

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: update code

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -289,13 +337,105 @@ def xfstesting_btrfs_nvme_datadisk_validation(
...
+        if isinstance(environment.platform, AzurePlatform):
+            node = cast(RemoteNode, environment.nodes[0])
+            platform = environment.platform
+            information = environment.get_information()
+            resource_group_name = information["resource_group_name"]
+            location = information["location"]
+            random_str = "".join(
+                random.choices(string.ascii_lowercase + string.digits, k=10)
+            )
+            storage_account_name = f"lisa{random_str}sc"
```
squirrelsc: How about naming them like f"lisasc{random_str}"? So they can be sorted together.

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: You can add a line like `assert isinstance(environment.platform, AzurePlatform)`. It reduces one level of `if`.

lisa/operating_system.py
```diff
@@ -560,7 +560,9 @@ class Debian(Linux):
         r"(?P<patch>[0-9]+)"  # patch
         r"-(?P<build>[a-zA-Z0-9-_\.~]+)"  # build
     )
-    _package_existed_in_repo_pattern = re.compile(r"Candidate: ((?!none)).*", re.M)
+    _package_existed_in_repo_pattern = re.compile(
```
squirrelsc, line 583: Add examples above. The original examples are modified for top two patterns only now.

lisa/tools/xfstests.py
```diff
@@ -159,13 +159,37 @@ def get_xfstests_path(self) -> PurePath:
...
+        if mount_opts:
+            content = "\n".join(
+                [
+                    "[cifs]",
+                    "FSTYP=cifs",
+                    f"TEST_FS_MOUNT_OPTS=''{mount_opts}''",
+                    f"MOUNT_OPTIONS=''{mount_opts}''",
+                ]
+            )
+            echo.write_to_file(content, config_path, append=True)
```
squirrelsc: This line is shared, can be moved out from the if clause.

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: The logic is not clear still. What's relation between "data_disk" and "when use azure file share"? And the "configure" should be more specific to say like "prepare disk partition" or something similar.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -45,6 +67,32 @@ def _configure_disk(
...
+    file_path = node.get_pure_path("/etc/smbcredentials/lisa.cred")
+    echo = node.tools[Echo]
+    username = account_credential["account_name"]
+    password = account_credential["account_key"]
+    echo.write_to_file(f"username={username}", file_path, sudo=True, append=True)
+    echo.write_to_file(f"password={password}", file_path, sudo=True, append=True)
+    node.execute("cp -f /etc/fstab /etc/fstab_cifs", sudo=True)
+    for folder_name, share in test_folders_share_dict.items():
+        node.execute(f"mkdir {folder_name}", sudo=True)
+        echo.write_to_file(
```
squirrelsc, line 88: rebase to latest code, the method name is changed.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: updated

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: updated

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: updated

lisa/operating_system.py
LiliDeng: updated

lisa/tools/xfstests.py
LiliDeng: updated

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: I didn't notice any change for this method.

squirrelsc: You are right. I just changed the `cat`, not `echo`.

squirrelsc (approved):

PR #1652, opened 2021-12-07
DPDK: RHEL fix for package manager install

mcgov:
dpdk-devel not dpdk-dev

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -149,13 +149,17 @@ class DpdkTestpmd(Tool):
             self.node.log.info(
                 "Installing dpdk and dev package from package manager..."
             )
-            if isinstance(node.os, Ubuntu) or isinstance(node.os, Redhat):
+            if isinstance(node.os, Ubuntu):
                 node.os.install_packages(["dpdk", "dpdk-dev"])
-                self._dpdk_version_info = node.os.get_package_information("dpdk")
+            elif isinstance(node.os, Redhat):
+                node.os.install_packages(["dpdk", "dpdk-devel"])
             else:
                 raise NotImplementedError(
-                    f"Dpdk install isn't implemented for Os {node.os.name}"
+                    "Dpdk package names are missing in dpdktestpmd.install"
+                    f" for os {node.os.name}"
                 )
+
+            self._dpdk_version_info = node.os.get_package_information("dpdk")
 
             if self._dpdk_version_info >= "19.11.0":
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -149,9 +149,12 @@ def _install(self) -> bool:
             self.node.log.info(
                 "Installing dpdk and dev package from package manager..."
             )
-            if isinstance(node.os, Ubuntu) or isinstance(node.os, Redhat):
+            if isinstance(node.os, Ubuntu):
                 node.os.install_packages(["dpdk", "dpdk-dev"])
                 self._dpdk_version_info = node.os.get_package_information("dpdk")
+            elif isinstance(node.os, Redhat):
+                node.os.install_packages(["dpdk", "dpdk-devel"])
+                self._dpdk_version_info = node.os.get_package_information("dpdk")
```
squirrelsc: this line is sharable. It can be moved below the if clauses.

mcgov: I thought that too but it has to be in the isinstance clause bc of the type checker.

squirrelsc: How about assert it's posix before the if clause?

mcgov: Oo good point, I can use self.can_install to assert

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -149,14 +149,21 @@ def _install(self) -> bool:
...
+                node.os.install_packages(["dpdk", "dpdk-devel"])
             else:
                 raise NotImplementedError(
-                    f"Dpdk install isn't implemented for Os {node.os.name}"
+                    "Dpdk package names are missing in dpdktestpmd.install"
+                    f" for os {node.os.name}"
                 )
 
+            # assert that we're an instance of a supported OS for the type checker
+            assert_that(self.can_install).is_true()
```
squirrelsc: 1. It doesn't need to check `can_install` in `_install`. Because only the `can_install` return True, the `_install` is called.
2. The line itself should called like `self.can_install()`, not `self.can_install`. Python is not a strong type language, it's easy to miss like this. :)

mcgov: Welp, I need to go get my head checked. It does work without the assert. Could have sworn I tried it before and was getting errors 😅

squirrelsc (approved):

PR #1654, opened 2021-12-08
debian backport versions can contain tilde in build info

mcgov:
Add tilde to the build info and version grabber for debian, this can be used to denote a backported version

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -545,24 +545,20 @@ class Debian(Linux):
         r"\s+(?P<metadata>.*)\s*"
     )
 
-    """ Package: git
+    """ Package: dpdk
+        Version: 20.11.3-0ubuntu1~backport20.04-202111041420~ubuntu20.04.1
         Version: 1:2.25.1-1ubuntu3.2
-        Priority: optional
-        Section: vcs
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -555,14 +555,14 @@ class Debian(Linux):
     """
     _debian_package_information_regex = re.compile(
         r"Package: ([a-zA-Z0-9:_\-\.]+)\r?\n"  # package name group
-        r"Version: ([a-zA-Z0-9:_\-\.]+)\r?\n"  # version number group
+        r"Version: ([a-zA-Z0-9:_\-\.~]+)\r?\n"  # version number group
```
squirrelsc, line 554: Add the new founded examples line above. It doesn't need to add all lines, just add the line with new Version format.

squirrelsc: Can we leave the original version also? It doesn't need to include other lines, because they are not easy to matched. like below,
```python
    """ Package: dpdk
        Version: 20.11.3-0ubuntu1~backport20.04-202111041420~ubuntu20.04.1
        Version: 1:2.25.1-1ubuntu3.2
```

squirrelsc (approved):

PR #1656, opened 2021-12-09
DPDK: Ubuntu needs backports enabled for all releases

mcgov:
Ubuntu previously used backports only for 20.04, 18.04 also requires them for dpdk packages.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -278,6 +278,8 @@ class DpdkTestpmd(Tool):
         cwd = node.working_path
 
         if isinstance(node.os, Ubuntu):
+            # DPDK requires backports channel for all releases
+            node.os.add_repository("ppa:canonical-server/server-backports")
             if "18.04" in node.os.information.release:
                 node.os.install_packages(list(self._ubuntu_packages_1804))
                 self.__execute_assert_zero("pip3 install --upgrade meson", cwd)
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -278,6 +278,21 @@ def _install_dependencies(self) -> None:
         cwd = node.working_path
 
         if isinstance(node.os, Ubuntu):
+            # DPDK requires backports channel for all releases
```
squirrelsc, line 281: 1. I created a method in operating_system for debian. Can you improve it to support the ppa? You can merge the current parameters into one to get it free format.
2. In my test cases, it doesn't need the `-y`. Did you is it must have?
3. The `apt-get update` will be called after add repo. It doesn't need to be called again.

mcgov: I think we do need the 'y' since there's a splash screen that pops up telling you about what the backports repo is when you run this.

mcgov: I modified add_repository to special case "ppa:*" urls, this is nicer :) Thanks!

squirrelsc: How about we just leave one parameter like below? Some parameters are not used by ppa, if it's specified, that would be confused. You just need to update the xdp test case. It's the only one is using it so far.

```python
    def add_repository(
        self,
        repo: str,
        key_location: str = "",
```

squirrelsc (approved):

PR #1658, opened 2021-12-09
Support hash check on sas vhd copy.

squirrelsc:
see details in commits.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -31,6 +31,7 @@ from azure.mgmt.resource.resources.models import (  # type: ignore
     DeploymentMode,
     DeploymentProperties,
 )
+from azure.storage.blob import BlobClient  # type: ignore
 from dataclasses_json import dataclass_json
 from marshmallow import fields, validate
 from retry import retry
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1699,9 +1700,16 @@ def _get_deployable_vhd_path(
         matches = SAS_URL_PATTERN.match(vhd_path)
         if not matches:
             return vhd_path
-        log.debug("found the vhd is a sas url, it needs to be copied...")
+        log.debug("found the vhd is a sas url, it may need to be copied.")
 
+        # get original vhd's hash key for comparing.
+        original_key: Optional[bytearray] = None
         original_vhd_path = vhd_path
+        original_blob_client = BlobClient.from_blob_url(original_vhd_path)
```
lizzha, line 1708: does it work when the VHD is not in the same subscription as the test subscription?

squirrelsc: Yes, it must be a sas URL.

dsrivastavv (approved):
LGTM!

lizzha (approved):

PR #1661, opened 2021-12-13
Migrate storage test cases.

LiliDeng:


lisa/util/constants.py
```diff
@@ -134,3 +134,5 @@
 DEVICE_TYPE_SRIOV = "SRIOV"
 DEVICE_TYPE_NVME = "NVME"
 DEVICE_TYPE_GPU = "GPU"
+
+FIO_MODES = "randread randwrite read write"
```
squirrelsc: 1. Don't put tool specified constants in the global constants file.
2. Define it as Enum.

lisa/tools/fio.py
```diff
@@ -0,0 +1,82 @@
...
+        if not self._check_exists():
+            self._install_from_src()
+        return self._check_exists()
+
+    def launch(
+        self,
+        name: str,
+        filename: str,
+        mode: str,
+        runtime: int,
```
squirrelsc: 1. Give a default value
2. How about 10 seconds?
3. The name is confusing, "time" may be enough.

lisa/tools/fio.py
```diff
@@ -0,0 +1,82 @@
...
+        if direct:
+            cmd += " --direct=1"
+        if gtod_reduce:
+            cmd += " --gtod_reduce=1"
+        if group_reporting:
+            cmd += " --group_reporting"
+        self.run(
+            cmd,
+            force_run=True,
+            sudo=True,
```
squirrelsc, line 98: Does it need sudo?

LiliDeng: Launch fio don't need sudo, but when run against device, it need sudo, I tried.

lisa/util/constants.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: updated the name, 10 seconds is too short, the fio need warm up.

lisa/tools/fio.py
```diff
@@ -0,0 +1,82 @@
...
+
+from typing import TYPE_CHECKING, List, Type, cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+
+from .git import Git
+
+if TYPE_CHECKING:
+    from .make import Make
```
squirrelsc: Does it work in the TYPE_CHECKING? the Make is used in the logic as a type. It may cause error on runtime.

lisa/tools/fio.py
squirrelsc: If 10 seconds is too short, please give another longer default value. The 120 is too long, we can try 20, 30, 60 first.

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,81 @@
...
+    description="""
+    This test suite is to validate NVMe disk performance of Linux VM using fio tool.
+    """,
+)
+class nvmeperf(TestSuite):  # noqa
+    TIME_OUT = 5400
+
+    @TestCaseMetadata(
+        description="""
+        This test case will use fio to test NVMe disk performance.
```
squirrelsc: remove the "will". It describes the behavior of the test case.

lisa/tools/fio.py
LiliDeng: not test this path yet, without this, it will enter into circular import error.

microsoft/testsuites/nvme/performance.py
LiliDeng: removed

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,95 @@
...
+        fio_result_list: List[FIOResult] = []
+        for mode in FIOMODES:
+            qdepth = start_qdepth
+            while qdepth <= max_qdepth:
+                iodepth = int(qdepth / numjob)
+                fio_result = fio.launch(
+                    name=f"iteration{numjobiterator}",
+                    filename=filename,
+                    mode=mode.name,
+                    time=120,
```
squirrelsc: use the default value

lisa/tools/fio.py
LiliDeng: currently use 60 s

microsoft/testsuites/nvme/performance.py
LiliDeng: use default value

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,99 @@
...
+            while qdepth <= max_qdepth:
+                iodepth = int(qdepth / numjob)
+                fio_result = fio.launch(
+                    name=f"iteration{numjobiterator}",
+                    filename=filename,
+                    mode=mode.name,
+                    iodepth=int(iodepth),
+                    numjob=numjob,
+                )
+                fio_result_list.append(fio_result)
```
squirrelsc: How about deliver multiple result messages, instead of creating a list and parse it later? It can be sent when the message is generated. It doesn't need to wait until all results are produced.

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,99 @@
...
+        description="""
+        This test case uses fio to test NVMe disk performance.
+        """,
+        priority=2,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            supported_features=[NvmeSettings(disk_count=8)],
+        ),
+    )
+    def verify_nvme_performance(self, node: Node, environment: Environment) -> None:
```
squirrelsc: The case is not to verify something. So, let's use a pattern for perf test like "perf_nvme".

lisa/tools/fio.py
```diff
@@ -0,0 +1,142 @@
...
+from lisa.operating_system import Debian, Posix, Redhat, Suse
+from lisa.util import LisaException
+
+from .git import Git
+
+
+class Fio(Tool):
+    fio_repo = "https://github.com/axboe/fio/"
+    branch = "fio-3.29"
+    _result_pattern = re.compile(
```
squirrelsc, line 44: add examples above.

lisa/tools/fio.py
```diff
@@ -0,0 +1,142 @@
...
+        r"([\w\W]*?)iops.*: min=(?P<min_iops>.+?), max=(?P<max_iops>.+?), "
+        r"avg=(?P<avg_iops>.+?), stdev=(?P<stdev_iops>.+?), "
```
squirrelsc: How about values like stdev_iops, it looks they are not sent in the message.

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,99 @@
...
+                    iodepth=int(iodepth),
+                    numjob=numjob,
+                )
+                fio_result_list.append(fio_result)
+                qdepth = qdepth * 2
+                numjobiterator += 1
+        fio_message = FIOResultsMessage()
+        fio_message.test_case_name = "verify_nvme_performance"
+        fio_message.block_size = 4
+        fio_message.disk_setup_type = "8 NVMe Disks"
```
squirrelsc: What's other values of the setup type? If all of them are disk count, we can rename it to "disk_count". The structured data is easier to compare.

lisa/tools/fio.py
```diff
@@ -0,0 +1,142 @@
...
+            expected_exit_code_failure_message=f"fail to run {cmd}",
+        )
+        matched_results = self._result_pattern.match(result.stdout)
+        assert matched_results
+        iops = matched_results.group("avg_iops")
+        latency = matched_results.group("latency")
+        fio_result = FIOResult()
+        fio_result.iops = Decimal(iops)
+        fio_result.latency = Decimal(latency)
+        fio_result.qdepth = iodepth * numjob
```
squirrelsc, line 114: Is the qdepth a common concept? I can see only iodepth in fio command. How about we save both iodepth and cpu count?

microsoft/testsuites/nvme/performance.py
LiliDeng: table for fio test results is based on the iodepth, we need collect all test results then parse them, insert them one time.

microsoft/testsuites/nvme/performance.py
LiliDeng: others like RAID 0 - 24 SSD Disks.

microsoft/testsuites/nvme/performance.py
squirrelsc: what does the "parse" mean? I see the items insert in database one by one. The shared information can be duplicate to each message. That will make the processing simpler.

microsoft/testsuites/nvme/performance.py
squirrelsc: we can add another field called disk_count, this field just put "RAID 0". This field presents how the disk is setup, not include the count of disk.

lisa/tools/fio.py
LiliDeng: db table doesn't have this column.

microsoft/testsuites/nvme/performance.py
LiliDeng: ok, can we get the disk_count in requirement during runtime?

squirrelsc: It can be gotten from the node.capability.

squirrelsc: But it may include other image specified disks. Because this test case asked for 8 disks already. We can trust this requirement, and specify the disk count in message.

microsoft/testsuites/nvme/performance.py
LiliDeng: updated

microsoft/testsuites/nvme/performance.py
LiliDeng: now results are parsed in test case.

lisa/tools/fio.py
LiliDeng: will add

lisa/tools/fio.py
LiliDeng: add iodepth, core_count, disk_count

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,128 @@
...
+        fio_messages: List[FIOResultsMessage],
+    ) -> None:
+        information: Dict[str, str] = environment.get_information()
+        host_type = information.pop("platform", "")
+        host_by = information.pop("location", "")
+        host_os = information.pop("host_version", "")
+        guest_distro = information.pop("distro_version", "")
+        guest_size = information.pop("vmsize", "")
+        kernel_version = information.pop("kernel_version", "")
+        lis_version = information.pop("lis_version", "")
```
squirrelsc: 1. Please refer to `fields_to_dict` to create a method `dict_to_fields`. So it can be easier in future.
2. Use `get` instead of `pop`. The pop will remove the value from the dict, it doesn't necessary.

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,128 @@
...
+        for fio_message in fio_messages:
+            fio_message.core_count = core_count
+            fio_message.disk_count = disk_count
+            fio_message.test_case_name = "perf_nvme"
+            fio_message.block_size = 4
+            fio_message.disk_setup_type = "raw disks"
```
squirrelsc: "Raw" is enough. To get consistent values, let's define constants near the DiskPerformanceMessage.

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,128 @@
...
+            fio_message.disk_type = "NVMe"
```
squirrelsc: Use constants like above.

lisa/notifier.py
```diff
@@ -40,6 +42,48 @@ class TestRunMessage(MessageBase):
...
+class FIOResult:
+    qdepth: int = 0
+    mode: str = ""
+    iops: Decimal = Decimal(0)
+    latency: Decimal = Decimal(0)
+    iodepth: int = 0
+
+
+@dataclass
+class FIOResultsMessage(PerfMessage):
```
squirrelsc: It's not limited to FIO. Let's call it like `DiskPerformanceMessage`. Add one more field "tool", and its value is "fio" (a constant).

lisa/tools/fio.py
```diff
@@ -0,0 +1,202 @@
...
+        latency = matched_results.group("latency")
+        fio_result = FIOResult()
+        fio_result.iops = iops_value
+        fio_result.latency = Decimal(latency)
+        fio_result.iodepth = iodepth
+        fio_result.qdepth = iodepth * numjob
+        fio_result.mode = mode
+        return fio_result
+
+    def parse_test_results(
```
squirrelsc: `create_performance_messages`

lisa/notifier.py
```diff
@@ -40,6 +42,48 @@ class TestRunMessage(MessageBase):
     message: str = ""
 
 
+@dataclass
+class PerfMessage(MessageBase):
+    pass
+
+
+class FIOResult:
```
squirrelsc: This class should be put with fio tool. It's fio related. Not is a part of LISA.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -655,8 +655,9 @@ def get_raw_data_disks(self) -> List[str]:
         )
         matched = find_patterns_in_lines(cmd_result.stdout, [pattern])
         assert matched[0]
-        disk_array: List[str] = [""] * len(matched[0])
-        for disk in matched[0]:
+        matched_disk_array = set(matched[0])
```
squirrelsc, line 658: The commit message is confusing. The words "minor update", "new scenarios" have no information.

lisa/tools/fdisk.py
```diff
@@ -34,13 +39,15 @@ def make_partition(self, disk_name: str, file_system: FileSystem) -> None:
         mkfs = self.node.tools[Mkfs]
         self.node.execute(
             f"(echo n; echo p; echo 1; echo ; echo; echo ; echo w) | "
-            f"{self.command} {disk_name}",
+            f"{self.command} -w always -W always {disk_name}",
```
squirrelsc, line 46: add comments to explain new arguments.

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+@TestSuiteMetadata(
+    area="nvme",
+    category="performance",
+    description="""
+    This test suite is to validate NVMe disk performance of Linux VM using fio tool.
+    """,
+)
+class nvmeperf(TestSuite):  # noqa
```
squirrelsc: `NvmePerformace`

lisa/tools/fio.py
```diff
@@ -0,0 +1,202 @@
...
+        self, qdepth: int, fio_results_list: List[FIOResult]
+    ) -> Dict[str, Decimal]:
+        mode_iops_latecny: Dict[str, Any] = {}
+        for fio_result in fio_results_list:
+            if qdepth == fio_result.qdepth:
+                mode_iops_latecny[f"{fio_result.mode}_iops"] = fio_result.iops
+                mode_iops_latecny[f"{fio_result.mode}_lat_usec"] = fio_result.latency
+                mode_iops_latecny["iodepth"] = fio_result.iodepth
+                mode_iops_latecny["qdepth"] = fio_result.qdepth
+        return mode_iops_latecny
```
squirrelsc: typo: latecny -> latency

lisa/tools/fio.py
```diff
@@ -0,0 +1,202 @@
...
+            mode_iops_latecny = self._get_iops_latency_by_qdepth(
+                qdepth, fio_results_list
+            )
+            fio_result_message = FIOResultsMessage()
+            fio_result_message.qdepth = int(mode_iops_latecny.pop("qdepth", 0))
+            fio_result_message.iodepth = int(mode_iops_latecny.pop("iodepth", 0))
+            fio_result_message.seq_read_iops = mode_iops_latecny.pop(
+                "read_iops", Decimal(0)
+            )
+            fio_result_message.seq_read_lat_usec = mode_iops_latecny.pop(
```
squirrelsc: use `setattr` can pass the attribute name as a string, so it can be done in a loop too.

lisa/tools/fio.py
```diff
@@ -0,0 +1,202 @@
...
+        fio_result.mode = mode
+        return fio_result
+
+    def parse_test_results(
+        self, fio_results_list: List[FIOResult]
+    ) -> List[FIOResultsMessage]:
+        qdepth_list = sorted(set([x.qdepth for x in fio_results_list if x.qdepth]))
+        fio_message: List[FIOResultsMessage] = []
+        for qdepth in qdepth_list:
+            mode_iops_latecny = self._get_iops_latency_by_qdepth(
```
squirrelsc: It may not need the extra loop. Just construct a dict of qdepth/message, and then extra the message list from the dict. One loop should be enough.

lisa/tools/fio.py
```diff
@@ -0,0 +1,202 @@
...
+        fio_result.latency = Decimal(latency)
+        fio_result.iodepth = iodepth
+        fio_result.qdepth = iodepth * numjob
+        fio_result.mode = mode
+        return fio_result
+
+    def parse_test_results(
+        self, fio_results_list: List[FIOResult]
+    ) -> List[FIOResultsMessage]:
+        qdepth_list = sorted(set([x.qdepth for x in fio_results_list if x.qdepth]))
```
squirrelsc: can we use iodepth directly? the qdepth is for compatibility only.

lisa/tools/fio.py
```diff
@@ -0,0 +1,202 @@
...
+    def command(self) -> str:
+        return "fio"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git]
```
squirrelsc: Git is not a hard dependency, because the tool may be installed from packages. So don't add git here.

microsoft/testsuites/nvme/performance.py
```diff
@@ -0,0 +1,128 @@
...
+    TestSuiteMetadata,
+    notifier,
+    simple_requirement,
+)
+from lisa.environment import Environment
+from lisa.features import Nvme, NvmeSettings
+from lisa.notifier import FIOResult, FIOResultsMessage
+from lisa.tools import Echo, Fio, Lscpu
+
+FIOMODES = Enum(
```
squirrelsc: Move it to fio.

microsoft/testsuites/storage/performance.py
```diff
@@ -0,0 +1,189 @@
...
+    schema,
+    search_space,
+    simple_requirement,
+)
+from lisa.environment import Environment
+from lisa.features import Disk
+from lisa.notifier import FIOResult, FIOResultsMessage
+from lisa.tools import Fdisk, Fio, Lscpu, Mdadm
+
+FIOMODES = Enum(
```
squirrelsc: If it's moved to fio, it's easy to share.

microsoft/testsuites/storage/performance.py
```diff
@@ -0,0 +1,189 @@
...
+    partition_disks: List[str] = []
+    for data_disk in disk_list:
+        fdisk.delete_partitions(data_disk)
+        partition_disks.append(fdisk.make_partition(data_disk, format=False))
+    return partition_disks
+
+
+def _stop_raid(node: Node) -> None:
+    mdadm = node.tools[Mdadm]
+    mdadm.stop_raid()
```
squirrelsc: Please add a task to me. It needs a flag on environment to mark it dirty, and not run other cases. In some cases, it's hard to recover the environment to original state. We need to discard that environment.

microsoft/testsuites/storage/performance.py
```diff
@@ -0,0 +1,189 @@
...
+                fio_result_list.append(fio_result)
+                qdepth = qdepth * 2
+                numjobindex += 1
+                numjobiterator += 1
+        fio_messages: List[FIOResultsMessage] = fio.parse_test_results(fio_result_list)
+        self._handle_and_send_back_results(
+            core_count, disk_count, environment, fio_messages, block_size
+        )
+
+    def _handle_and_send_back_results(
```
squirrelsc: this method is sharable. Please find a way to share it.

microsoft/testsuites/storage/performance.py
```diff
@@ -0,0 +1,189 @@
...
+        ),
+    )
+    def perf_premium_datadisks_4k(self, node: Node, environment: Environment) -> None:
+        self._perf_premium_datadisks(node, environment)
+
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test data disk performance using 1024K block size.
+        """,
+        priority=2,
```
squirrelsc: Does it need to be p2? It's a perf test case. what's the original priority in v2?

microsoft/testsuites/storage/performance.py
```diff
@@ -0,0 +1,189 @@
...
+@TestSuiteMetadata(
+    area="storage",
+    category="performance",
+    description="""
+    This test suite is to validate premium SSD data disks performance of Linux VM using
+     fio tool.
+    """,
+)
+class storageperf(TestSuite):  # noqa
+    TIME_OUT = 3000
```
squirrelsc: How long the test case will take usually?

microsoft/testsuites/storage/performance.py
LiliDeng: original priority in v2 is 1

microsoft/testsuites/storage/performance.py
LiliDeng: between 2000 and 3000

microsoft/testsuites/storage/performance.py
squirrelsc: Let's move all perf test cases to p3, unless it can be done in 5 minutes.

LiliDeng: ok

lisa/tools/fio.py
LiliDeng: current dashboard is represent with qdepth , iodepth has duplicated values

microsoft/testsuites/nvme/performance.py
LiliDeng: updated

lisa/notifier.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: updated

lisa/notifier.py
LiliDeng: updated

lisa/sut_orchestrator/azure/features.py
LiliDeng: updated

lisa/tools/fdisk.py
LiliDeng: updated

microsoft/testsuites/nvme/performance.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: updated

microsoft/testsuites/nvme/performance.py
LiliDeng: updated

microsoft/testsuites/storage/performance.py
LiliDeng: updated

microsoft/testsuites/storage/performance.py
LiliDeng: updated

microsoft/testsuites/storage/performance.py
LiliDeng: updated

microsoft/testsuites/nvme/performance.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: updated, thanks

microsoft/testsuites/storage/performance.py
LiliDeng: added

lisa/tools/fio.py
LiliDeng: when construct a dict of qdepth/message we still need loop? for every qdepth, there are 4 modes' iops and latency data. then the value of dict qdepth/message need to be an object?

lisa/tools/fio.py
squirrelsc: Let's use iodepth in code. The qdepth is used in current report, and we will use iodepth in new reports.

lisa/tools/fio.py
squirrelsc: The qdepth/message dict is used to merge data. One for loop is still necessary to go through all combinations.

lisa/tools/fio.py
LiliDeng: the iodepths we used are not unique when run fio against ssd disks, not sure the root cause.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -0,0 +1,165 @@
...
+        self, node: Node, environment: Environment
+    ) -> None:
+        self._perf_premium_datadisks(
+            node,
+            environment,
+            block_size=1024,
+            test_case_name="perf_premium_datadisks_4k",
+        )
+
+    def _perf_premium_datadisks(
```
squirrelsc, line 105: This logic looks similar with nvme. Considering to share more logic.

lisa/tools/fio.py
```diff
@@ -0,0 +1,194 @@
...
+    latency: Decimal = Decimal(0)
+    iodepth: int = 0
+
+
+FIOMODES = Enum(
+    "FIOMODES",
+    [
+        "randread",
+        # "randwrite",
+        # "read",
```
squirrelsc: What's the reason to comment them out? If they are not necessary, just remove them.

lisa/tools/fio.py
squirrelsc: What means "not unique"? If it's a problem, why isn't a problem when using qdepth?

lisa/tools/fio.py
LiliDeng: added

lisa/tools/fio.py
LiliDeng: image
they have duplicated values

lisa/tools/fio.py
LiliDeng: uncommented, it is for testing.

lisa/tools/fio.py
squirrelsc: I see. But how the number of jobs is decided? It's the reason that the iodepth is not unique.

LiliDeng: I don't have the answer yet, will try to figure out.

microsoft/testsuites/nvme/performance.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: updated

microsoft/testsuites/performance/storageperf.py
LiliDeng: working on it

lisa/tools/fdisk.py
```diff
@@ -19,7 +19,12 @@ def command(self) -> str:
...
-    def make_partition(self, disk_name: str, file_system: FileSystem) -> None:
+    def make_partition(
+        self,
+        disk_name: str,
+        file_system: FileSystem = FileSystem.ext4,
+        format: bool = True,
+    ) -> str:
         """
         disk_name: make a partition against the disk.
         file_system: making the file system type against the partition
```
dsrivastavv: [nit] Can you please add `format` parameter description?

lisa/tools/fio.py
```diff
@@ -0,0 +1,195 @@
...
+            package_list = ["wget", "mdadm", "blktrace", "libaio1", "sysstat", "bc"]
+        else:
+            raise LisaException(
+                f"tool {self.command} can't be installed in distro {self.node.os.name}."
+            )
+        for package in list(package_list):
+            if posix_os.is_package_in_repo(package):
+                posix_os.install_packages(package)
+
+    def _install_from_src(self) -> bool:
```
dsrivastavv: [nit] private methods should be at the end

microsoft/testsuites/storage/storagesuite.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    def verify_disk_with_nobarrier(self, node: Node, environment: Environment) -> None:
+        disk = node.features[Disk]
+        data_disks = disk.get_raw_data_disks()
+        disk_count = len(data_disks)
+        assert_that(disk_count).described_as(
+            "At least 1 data disk for testing."
+        ).is_greater_than(0)
+        partition_disks = _format_disk(node, data_disks)
+        _stop_raid(node)
```
dsrivastavv, line 79: Should we mark the node to be recycled if format disk or stop raid step fails?

microsoft/testsuites/performance/storageperf.py
```diff
@@ -0,0 +1,165 @@
...
+        block_size: int = 4,
+    ) -> None:
+        disk = node.features[Disk]
+        data_disks = disk.get_raw_data_disks()
+        disk_count = len(data_disks)
+        assert_that(disk_count).described_as(
+            "At least 1 data disk for fio testing."
+        ).is_greater_than(0)
+        partition_disks = _format_disk(node, data_disks)
+        print(partition_disks)
```
dsrivastavv: Should this be sent to logger?

LiliDeng: debug purpose, remove it, thanks!

microsoft/testsuites/performance/storageperf.py
LiliDeng: updated

lisa/tools/fdisk.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: updated

lisa/tools/fio.py
```diff
@@ -0,0 +1,195 @@
...
+        return self._check_exists()
+
+    def launch(
+        self,
+        name: str,
+        filename: str,
+        mode: str,
+        iodepth: int,
+        numjob: int,
+        time: int = 120,
```
LiliDeng, line 71: @squirrelsc I changed runtime back to 120 seconds, previous value we used was 300, we did some comparison using different runtime, found 120 is the shortest runtime and the perf reaches the same effect when runtime is 300.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -0,0 +1,150 @@
...
+        )
+
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test data disk performance using 1024K block size.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            min_core_count=72,
```
dsrivastavv, line 87: Why is there a minimum core count requirement? Is this also present in v2?

LiliDeng: to use F72sv2, it is a size to use isolated host. to align with lisav2

lisa/tools/fio.py
```diff
@@ -0,0 +1,195 @@
...
+                temp = mode_iops_latency[fio_result.qdepth]
+            temp[f"{fio_result.mode}_iops"] = fio_result.iops
+            temp[f"{fio_result.mode}_lat_usec"] = fio_result.latency
+            temp["iodepth"] = fio_result.iodepth
+            temp["qdepth"] = fio_result.qdepth
+            mode_iops_latency[fio_result.qdepth] = temp
+
+        for result in mode_iops_latency.values():
+            fio_result_message = DiskPerformanceMessage()
+            for name, value in result.items():
```
squirrelsc: use the new dict_to_fields

lisa/tools/fio.py
squirrelsc: 1. Make sure the numjobs is in message and saved.
2. Please double check if we need to test other numjobs. If not, Let's refer to nvme test case, let the numjobs always equals to number of cpu count. And then loop the iodepth from 1, 2, 4, 8, 16, 32, 64.

microsoft/testsuites/storage/storagesuite.py
LiliDeng: whenever format the disk and make a raid, before that we will delete partitions and stop the raid.
If making raid/format disk fails, I don't think there is any impact.

lisa/tools/fio.py
LiliDeng: I will try with number of cpu count when run against premium disk.

lisa/tools/fio.py
LiliDeng: updated

lisa/tools/fio.py
LiliDeng: all of them use core count as numjobs.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -0,0 +1,149 @@
...
+        ).is_greater_than(0)
+        partition_disks = _format_disk(node, data_disks)
+        _stop_raid(node)
+        _make_raid(node, partition_disks)
+        filename = "/dev/md0"
+        cpu = node.tools[Lscpu]
+        core_count = cpu.get_core_count()
+        start_qdepth = core_count
+        max_qdepth = start_qdepth * 256
+        fio_messages: List[DiskPerformanceMessage] = run_perf_test(
```
squirrelsc, line 126: Does it test only once now? Should it cover different iodepth?

LiliDeng: latest results
image

lisa/notifier.py
```diff
@@ -40,6 +42,59 @@ class TestRunMessage(MessageBase):
...
+    host_version: str = ""
+    guest_os_type: str = "Linux"
+    distro_version: str = ""
+    vmsize: str = ""
+    kernel_version: str = ""
+    lis_version: str = ""
+    disk_setup_type: DiskSetupType = DiskSetupType.raw
+    block_size: int = 0
+    disk_type: DiskType = DiskType.nvme
+    core_count: int = 0
```
squirrelsc, line 83: Please save the number of jobs too. It's clearer. And we may try different number of jobs in future.

microsoft/testsuites/performance/common.py
```diff
@@ -0,0 +1,72 @@
...
+    size_gb: int = 0,
+    numjob: int = 0,
+    overwrite: bool = False,
+) -> List[DiskPerformanceMessage]:
+    fio_result_list: List[FIOResult] = []
+    fio = node.tools[Fio]
+    numjobiterator = 0
+    for mode in FIOMODES:
+        qdepth = start_qdepth
+        while qdepth <= max_qdepth:
```
squirrelsc: The loop should be changed to check the iodepth. The qdepth will be a value only, not use it in any calculation.

LiliDeng: then in future we can abandon qdepth

squirrelsc (dismissed):

LiliDeng:
add force run for mdadm by default

squirrelsc (approved):

PR #1669, opened 2021-12-18
Fix kvp test case failures

squirrelsc:
See details in commits.

### DIFF
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -225,9 +225,9 @@ class KvpClient(Tool):
         result = self.run(
             str(pool_id),
             force_run=force_run,
-            expected_exit_code=0,
-            expected_exit_code_failure_message="failed to get pool",
         )
+        # some distro return 4, for example, Ubuntu Server 1804
+        assert_that(result.exit_code).described_as("failed to get pool").is_in(0, 4)
         matched_lines = find_patterns_in_lines(result.stdout, [self._key_value_pattern])
         records = {item[0]: item[1] for item in matched_lines[0]}
 
--- a/microsoft/testsuites/core/kvp.py
+++ b/microsoft/testsuites/core/kvp.py
@@ -31,7 +31,7 @@ from lisa.util import get_matched_str
 )
 class Kvp(TestSuite):
     # lrwx------ 1 root root 64 Oct 18 19:05 9 -> /dev/vmbus/hv_kvp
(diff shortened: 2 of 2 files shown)


microsoft/testsuites/core/kvp.py
```diff
@@ -31,7 +31,8 @@
 )
 class Kvp(TestSuite):
     # lrwx------ 1 root root 64 Oct 18 19:05 9 -> /dev/vmbus/hv_kvp
-    _hv_kvp_pattern = re.compile(r".* /dev/vmbus/(hv_kvp)$", re.M)
+    # lrwx------. 1 root root 64 Dec 18 07:38 4 -> /dev/vmbus/hv_kvp
+    _hv_kvp_pattern = re.compile(r".* /dev/vmbus/(hv_kvp)\r?$", re.M)
```
LiliDeng, line 34: is a new line character in the second sample?

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -225,9 +225,9 @@ def get_pool_records(self, pool_id: int, force_run: bool = False) -> Dict[str, s
         result = self.run(
             str(pool_id),
             force_run=force_run,
-            expected_exit_code=0,
-            expected_exit_code_failure_message="failed to get pool",
         )
+        # some distro return 4, for example, Ubuntu Server 1804
+        assert_that(result.exit_code).described_as("failed to get pool").is_in(0, 4)
```
LiliDeng, line 230: we can add one more comment after figuring out what the error code 4 stands for.

LiliDeng:
Also, as we know some distros don't have kvp by default in azure platform, if it is expected, then we need mark case as skipped in step 1?

lisa/sut_orchestrator/azure/tools.py
squirrelsc: I don't know what code 4 is here, and it succeeds. So I just record the distro name here.

microsoft/testsuites/core/kvp.py
squirrelsc: No, I thought its format is different, but actually the reason is the \r at the end. I will remove the second sample.

squirrelsc:
> Also, as we know some distros don't have kvp by default in azure platform, if it is expected, then we need mark case as skipped in step 1?

will mark them in next round failure analysis.

LiliDeng (approved):

PR #1672, opened 2021-12-29
add a first time init for lscpu

mcgov:
Add a first-time init if there isn't any core count info yet.

### DIFF
--- a/lisa/tools/lscpu.py
+++ b/lisa/tools/lscpu.py
@@ -97,7 +97,10 @@ class Lscpu(Tool):
         return architecture
 
     def get_core_count(self, force_run: bool = False) -> int:
-        result = self.run(force_run=force_run)
+        if not self._core_count:
+            result = self.run(force_run=True)
+        else:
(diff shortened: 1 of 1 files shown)


lisa/tools/lscpu.py
```diff
@@ -97,7 +97,10 @@ def get_architecture(self, force_run: bool = False) -> str:
         return architecture
 
     def get_core_count(self, force_run: bool = False) -> int:
-        result = self.run(force_run=force_run)
+        if not self._core_count:
```
squirrelsc, line 100: Do you know the root cause? Can you share me the log?

mcgov: Is Lscpu tool guaranteed to run once before get_core_count is called? I had found in testing that it could return 0 if force_run wasn't passed on the first call.

squirrelsc: If so, it should raise an exception, if it doesn't pass on first time. So, we can see what's the error, then mitigate or retry.

PR #1673, opened 2021-12-30
migrate cpu offline online case.

LiliDeng:


microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+    ONLINE: str = "1"
+
+
+def skip_test(node: Node) -> None:
+    uname = node.tools[Uname]
+    kernel_version = uname.get_linux_information().kernel_version
+    config_path = f"/boot/config-{kernel_version}"
+    config_result = node.execute(f"grep CONFIG_HOTPLUG_CPU=y {config_path}", shell=True)
+    if config_result.exit_code != 0:
+        raise SkippedException("This distro doesn't support cpu hot plug.")
```
squirrelsc: specify the current distro in the exception.

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+from lisa.tools import Cat, Dmesg, Echo, Lscpu, Lsvmbus, Uname
+from lisa.util import SkippedException
+
+
+class CPUState:
+    OFFLINE: str = "0"
+    ONLINE: str = "1"
+
+
+def skip_test(node: Node) -> None:
```
squirrelsc: `check_runnable`

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def set_vmbus_channels_target_cpu(node: Node) -> Dict[str, str]:
+    uname = node.tools[Uname]
+    kernel_version = uname.get_linux_information().kernel_version
+    dmesg = node.tools[Dmesg]
+    lsvmbus = node.tools[Lsvmbus]
+    vmbus_version = dmesg.get_vmbus_version()
+    file_path_list: Dict[str, str] = {}
+    if kernel_version >= "5.8.0" and vmbus_version >= "4.1.0":
```
squirrelsc, line 39: add comments to explain the versions.

squirrelsc: add `else` clause. The `else` part has no return value even, it should be a check error.

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+def skip_test(node: Node) -> None:
+    uname = node.tools[Uname]
+    kernel_version = uname.get_linux_information().kernel_version
+    config_path = f"/boot/config-{kernel_version}"
+    config_result = node.execute(f"grep CONFIG_HOTPLUG_CPU=y {config_path}", shell=True)
+    if config_result.exit_code != 0:
+        raise SkippedException("This distro doesn't support cpu hot plug.")
+
+
+def set_vmbus_channels_target_cpu(node: Node) -> Dict[str, str]:
```
squirrelsc: please review all method names in this file, many of them are not straightforward. For example, in this method, it looks set all channels to cpu 0. This method name should specify it explicitly, and add comments to explain the purpose of this change.

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+                    )
+                ] = target_cpu
+        # set all vmbus channel interrupts go into cpu 0.
+        set_cpu_interrupts_assigned_to(file_path_list, node, "0")
+        return file_path_list
+
+
+def get_idle_cpu_set(node: Node) -> List[str]:
+    lsvmbus = node.tools[Lsvmbus]
+    channels = lsvmbus.get_device_channels_from_lsvmbus(force_run=True)
```
squirrelsc: the method `get_device_channels_from_lsvmbus` is in lsvmbus, it doesn't need to say `from_lsvmbus`.

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+    cpu_count = node.tools[Lscpu].get_core_count()
+    all_cpu = list(range(1, cpu_count))
+    idle_cpu = [str(x) for x in all_cpu if str(x) not in cpu_in_used]
+    return idle_cpu
+
+
+def set_idle_cpu_offline_online(node: Node, idle_cpu: List[str]) -> None:
+    for target_cpu in idle_cpu:
+        result = set_cpu_state(target_cpu, CPUState.OFFLINE, node)
+        if result:
```
squirrelsc: what happens on `else`?

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+
+def get_cpu_config_file(cpu_id: str) -> str:
+    return f"/sys/devices/system/cpu/cpu{cpu_id}/online"
+
+
+def get_interrupts_assigned_to_cpu(device_id: str, channel_id: str) -> str:
+    return f"/sys/bus/vmbus/devices/{device_id}/channels/{channel_id}/cpu"
+
+
+def set_cpu_interrupts_assigned_to(
```
squirrelsc: how about `assign_interrupts`?

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+    path_cpu: Dict[str, str],
+    node: Node,
+    target_cpu: str = "0",
+) -> None:
+    for path, _ in path_cpu.items():
+        node.tools[Echo].write_to_file(target_cpu, node.get_pure_path(path), sudo=True)
+
+
+def restore_cpu_interrupts_assigned_to(
```
squirrelsc: `restore_interrupts_assignment`

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+    node: Node,
+) -> None:
+    if path_cpu:
+        for path, target_cpu in path_cpu.items():
+            node.tools[Echo].write_to_file(
+                target_cpu, node.get_pure_path(path), sudo=True
+            )
+
+
+def set_cpu_state(cpu_id: str, state: str, node: Node) -> bool:
```
squirrelsc: `def set_cpu_state(node: Node, cpu: str, is_online: bool) -> bool:`

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+            restore_cpu_interrupts_assigned_to(file_path_list, node)
+    finally:
+        restore_cpu_interrupts_assigned_to(file_path_list, node)
+
+
+def get_cpu_config_file(cpu_id: str) -> str:
+    return f"/sys/devices/system/cpu/cpu{cpu_id}/online"
+
+
+def get_interrupts_assigned_to_cpu(device_id: str, channel_id: str) -> str:
```
squirrelsc: `get_interrupts_assigned_cpu`

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+            idle_cpu = get_idle_cpu_set(node)
+            if 0 == len(idle_cpu):
+                raise SkippedException("no idle cpu can be set offline or online.")
+            set_idle_cpu_offline_online(node, idle_cpu)
+            restore_cpu_interrupts_assigned_to(file_path_list, node)
+    finally:
+        restore_cpu_interrupts_assigned_to(file_path_list, node)
+
+
+def get_cpu_config_file(cpu_id: str) -> str:
```
squirrelsc: `get_cpu_state_file`? It's clearer to specify what config it is.

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+
+
+def verify_cpu_hot_plug(node: Node, run_times: int = 1) -> None:
+    try:
+        skip_test(node)
+        for _ in range(run_times):
+            file_path_list = set_vmbus_channels_target_cpu(node)
+            idle_cpu = get_idle_cpu_set(node)
+            if 0 == len(idle_cpu):
+                raise SkippedException("no idle cpu can be set offline or online.")
```
squirrelsc: add reason, why no idle cpu, or how the idle is defined.

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,131 @@
...
+            if not reset:
+                raise BadEnvironmentStateException(
+                    exception_message,
+                    f"the test failed leaving cpu{target_cpu} in a bad state.",
+                )
+
+
+def verify_cpu_hot_plug(node: Node, run_times: int = 1) -> None:
+    try:
+        skip_test(node)
```
squirrelsc: It should be out of the try...finally. And the `file_path_list` is not assigned in this step, even there is any error, it's impossible to report the interrupts.

microsoft/testsuites/cpu/stress.py
```diff
@@ -0,0 +1,33 @@
...
+    area="cpu",
+    category="stress",
+    description="""
+    This test suite is used to run cpu related tests under stress.
+    """,
+)
+class CPUStressSuite(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            The test case take idle cpus into offline and online for 10 times.
```
squirrelsc: Specify the steps should refer to verify_cpu_hot_plug

microsoft/testsuites/cpu/common.py
LiliDeng: added.

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: added, and move return out of if

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: added.

microsoft/testsuites/cpu/common.py
LiliDeng: moved out

microsoft/testsuites/cpu/stress.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,177 @@
...
+                    continue
+                file_path_list[
+                    get_interrupts_assigned_cpu(
+                        channel.device_id, channel_vp_map.rel_id
+                    )
+                ] = current_target_cpu
+        # set all vmbus channel interrupts go into cpu target_cpu.
+        assign_interrupts(file_path_list, node, target_cpu)
+    else:
+        log.debug(
```
squirrelsc, line 58: If the channels cannot be changed, what's impact on the test? Should the test case be skipped?

microsoft/testsuites/cpu/common.py
squirrelsc: this method name is not updated.

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,177 @@
...
+            f"current distro {node.os.name}, os version {kernel_version}, "
+            f"vmbus version {vmbus_version} doesn't support "
+            "change channels target cpu featue."
+        )
+    # if current distro doesn't support this feature, the backup dict will be empty,
+    # there is no anything we can restore later.
+    return file_path_list
+
+
+def get_idle_cpu_set(node: Node) -> List[str]:
```
squirrelsc: `get_idle_cpus`

microsoft/testsuites/cpu/common.py
```diff
@@ -0,0 +1,177 @@
...
+            if 0 == len(idle_cpu):
+                raise SkippedException(
+                    "all of the cpu are associated vmbus channels,"
+                    " no idle cpu can be used to test hotplug."
+                )
+            # start to take idle cpu from online to offline, then offline to online.
+            set_idle_cpu_offline_online(log, node, idle_cpu)
+            # when kernel doesn't support set vmbus channels target cpu feature, the
+            # dict which stores original status is empty, nothing need to be restored.
+            restore_interrupts_assignment(file_path_list, node)
```
squirrelsc, line 136: It doesn't need it, the "finally" clause always executes.

microsoft/testsuites/cpu/common.py
LiliDeng: first try to set target cpu into 0
then get current idle cpu, if the first feature is not supported, we still get the idle cpu based on current VM real usage, then skip test or not, it is handled in method verify_cpu_hot_plug

microsoft/testsuites/cpu/common.py
LiliDeng: when test is under stress, this step is needed in every iteration, so that the next iteration is like a fresh run.

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
LiliDeng: updated

microsoft/testsuites/cpu/common.py
squirrelsc: Add the explanation in code.

LiliDeng: I think it is already there

LiliDeng: > Add the explanation in code.
Below places are not enough?

https://github.com/microsoft/lisa/pull/1673/files#diff-cee6a3d2a6a62aabb5066517ea281c982068102c90d61f419a6f608d947cdc6aR37

https://github.com/microsoft/lisa/pull/1673/files#diff-d496653b77bd2eaf973ba7996aac876d2427107e3f880c4bb64b845c7aaae456R60

squirrelsc: 1. The explanation should be align with the code. The test spec is good explanation, but it's better to mention it in the right code place too.
2. The second comments is good to explain it. But it should be after the `else`. It's weird at current place, because the next line is to return the list, it doesn't relate to the value is empty or not.

LiliDeng: updated

squirrelsc (approved):

PR #1674, opened 2021-12-30
Add perf_premium_datadisks_io testcase

dsrivastavv:
Migrate PERF_STORAGE-DISK-IO

### DIFF
--- a/microsoft/testsuites/performance/storageperf.py
+++ b/microsoft/testsuites/performance/storageperf.py
@@ -102,12 +102,36 @@ class StoragePerformance(TestSuite):  # noqa
             test_case_name="perf_premium_datadisks_1024k",
         )
 
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test vm with 24 data disks.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            min_core_count=72,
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/performance/storageperf.py
```diff
@@ -102,12 +102,36 @@ def perf_premium_datadisks_1024k(
...
+        description="""
+        This test case uses fio to test vm with 24 data disks.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            min_core_count=72,
+            disk=schema.DiskOptionSettings(
+                disk_type=schema.DiskType.PremiumSSDLRS,
+                data_disk_iops=5000,
```
squirrelsc: set the iops as min value, so it can be more flexible.

LiliDeng: it is to test iops of P30 disk, it has the minimal requirement for the iops.

squirrelsc: Set it to a min value, not a number directly.

dsrivastavv: Updated. @squirrelsc @LiliDeng  Should we set a min value in other storage performance test cases too?

squirrelsc (approved):

PR #1675, opened 2022-01-02
xdp: implement test case with different mtu

squirrelsc:
see details in commits.

### DIFF
--- a/lisa/tools/ip.py
+++ b/lisa/tools/ip.py
@@ -1,0 +1,32 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from __future__ import annotations
+
+from assertpy import assert_that
+
+from lisa.executable import Tool
+from lisa.tools import Cat
+
+
+class Ip(Tool):
+    @property
+    def command(self) -> str:
+        return "ip"
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def restart_device(self, nic_name: str) -> None:
+        self.node.execute(
+            f"ip link set dev {nic_name} down;ip link set dev {nic_name} up", shell=True
+        )
+
+    def get_mtu(self, nic_name: str) -> int:
+        cat = self.node.tools[Cat]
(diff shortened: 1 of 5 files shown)


lisa/tools/ip.py
```diff
@@ -0,0 +1,32 @@
...
+    @property
+    def command(self) -> str:
+        return "ip"
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def restart_device(self, nic_name: str) -> None:
+        self.node.execute(
+            f"ip link set dev {nic_name} down;ip link set dev {nic_name} up", shell=True
```
dsrivastavv, line 22: Can we use `self.run` here instead of `self.node.execute`?

microsoft/testsuites/xdp/functional.py
```diff
@@ -166,6 +166,47 @@ def verify_xdp_action_aborted(
...
+                remote_ip.set_mtu(remote_nic_name, mtu)
+
+                # tested mtu equals (ping mtu - IP headr (20) - ICMP header (8))
+                xdpdump.test(
+                    xdp_node_nic_name,
+                    remote_address=remote_address,
+                    ping_package_size=mtu - 28,
+                )
+        finally:
+            xdp_node_ip.set_mtu(xdp_node_nic_name, original_xdp_node_mtu)
```
dsrivastavv, line 207: Should we mark the node to be recycled if we are unable to reset MTU back to its original value?

lisa/tools/ip.py
squirrelsc: It has two commands, and joined with the `;`. The `run` makes it looks weird.

microsoft/testsuites/xdp/functional.py
squirrelsc: The dirty env is not implemented yet. This line is safe here, it's in `finally` block, will be run in any case. And if the set_mtu will fail, the mtu won't be changed in previous logic too.

dsrivastavv (approved):

PR #1677, opened 2022-01-04
Add support for QEMU/libvirt platform.

cwize1:
Add basic support for running LISA framework on Linux platform
using QEMU via the libvirt API. This includes succesfully running
the Microsoft T0 tests against the CBL-Mariner image.

### DIFF
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -4,6 +4,8 @@
 # The file imports all the mix-in types that can be initialized
 # using reflection.
 
+import platform
+
 import lisa.combinators.batch_combinator  # noqa: F401
 import lisa.combinators.csv_combinator  # noqa: F401
 import lisa.combinators.grid_combinator  # noqa: F401
@@ -16,6 +18,12 @@ import lisa.runners.lisa_runner  # noqa: F401
 import lisa.sut_orchestrator.azure.hooks  # noqa: F401
 import lisa.sut_orchestrator.azure.transformers  # noqa: F401
 import lisa.sut_orchestrator.ready  # noqa: F401
+
+if platform.system() == "Linux":
+    import lisa.sut_orchestrator.qemu.context  # noqa: F401
+    import lisa.sut_orchestrator.qemu.platform  # noqa: F401
+    import lisa.sut_orchestrator.qemu.schema  # noqa: F401
+
 import lisa.transformers.kernel_source_installer  # noqa: F401
 import lisa.transformers.script_transformer  # noqa: F401
 import lisa.transformers.to_list  # noqa: F401
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -1,0 +1,447 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import io
+import os
+import random
+import string
+import subprocess
+import time
+import xml.etree.ElementTree as ET  # noqa: N817
+from typing import List, Optional, Tuple, Type
+
+import libvirt  # type: ignore
+import pycdlib  # type: ignore
+import yaml
+
+from lisa import schema
+from lisa.environment import Environment
+from lisa.feature import Feature
+from lisa.node import Node, RemoteNode
+from lisa.platform_ import Platform
+from lisa.util import LisaException, constants, get_public_key_data
+from lisa.util.logger import Logger
+
+from .. import QEMU
+from .context import get_environment_context, get_node_context
+from .schema import QemuNodeSchema
+
+
+class QemuPlatform(Platform):
+    def __init__(self, runbook: schema.Platform) -> None:
+        super().__init__(runbook=runbook)
+
+    @classmethod
+    def type_name(cls) -> str:
+        return QEMU
+
+    @classmethod
+    def supported_features(cls) -> List[Type[Feature]]:
+        return []
+
+    def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        return True
+
+    def _deploy_environment(self, environment: Environment, log: Logger) -> None:
+        self._deploy_nodes(environment, log)
+
+    def _delete_environment(self, environment: Environment, log: Logger) -> None:
+        with libvirt.open("qemu:///system") as qemu_conn:
+            self._delete_nodes(environment, log, qemu_conn)
+
+    def _deploy_nodes(self, environment: Environment, log: Logger) -> None:
+        self._configure_nodes(environment, log)
+
+        with libvirt.open("qemu:///system") as qemu_conn:
+            try:
+                self._create_nodes(environment, log, qemu_conn)
+                self._fill_nodes_metadata(environment, log, qemu_conn)
+
+            except Exception as ex:
+                self._delete_nodes(environment, log, qemu_conn)
+                raise ex
+
+    # Pre-determine all the nodes' properties, including the name of all the resouces
+    # to be created. This makes it easier to cleanup everything after the test is
+    # finished (or fails).
+    def _configure_nodes(self, environment: Environment, log: Logger) -> None:
+        environment_context = get_environment_context(environment)
+
+        # Generate a random name for the VMs.
+        test_suffix = "".join(random.choice(string.ascii_uppercase) for i in range(5))
+        vm_name_prefix = f"lisa-{test_suffix}"
+
+        environment_context.ssh_public_key = get_public_key_data(
+            self.runbook.admin_private_key_file
+        )
+
+        assert environment.runbook.nodes_requirement
+        for i, node_space in enumerate(environment.runbook.nodes_requirement):
+            assert isinstance(
+                node_space, schema.NodeSpace
+            ), f"actual: {type(node_space)}"
+
+            qemu_node_runbook: QemuNodeSchema = node_space.get_extended_runbook(
+                QemuNodeSchema, type_name=QEMU
+            )
+
+            vm_disks_dir = os.path.dirname(qemu_node_runbook.qcow2)
+
+            node = environment.create_node_from_requirement(node_space)
+            node_context = get_node_context(node)
+
+            node_context.vm_name = f"{vm_name_prefix}-{i}"
+            node_context.cloud_init_file_path = os.path.join(
+                vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
+            )
+            node_context.os_disk_base_file_path = qemu_node_runbook.qcow2
+            node_context.os_disk_file_path = os.path.join(
+                vm_disks_dir, f"{node_context.vm_name}-os.qcow2"
+            )
+
+            if not node.name:
+                node.name = node_context.vm_name
+
+            # Read extra cloud-init data.
+            if (
+                qemu_node_runbook.cloud_init
+                and qemu_node_runbook.cloud_init.extra_user_data
+            ):
+                extra_user_data_file_path = str(
+                    constants.RUNBOOK_PATH.joinpath(
+                        qemu_node_runbook.cloud_init.extra_user_data
+                    )
+                )
+                with open(extra_user_data_file_path, "r") as file:
+                    node_context.extra_cloud_init_user_data = yaml.safe_load(file)
+
+    # Create all the VMs.
+    def _create_nodes(
+        self, environment: Environment, log: Logger, qemu_conn: libvirt.virConnect
+    ) -> None:
+        for node in environment.nodes.list():
+            # Create cloud-init ISO file.
+            self._create_node_cloud_init_iso(environment, log, node)
+
+            # Create OS disk from the provided image.
+            self._create_node_os_disk(environment, log, node)
+
+            # Create libvirt domain (i.e. VM).
+            xml = self._create_node_domain_xml(environment, log, node)
+            domain = qemu_conn.defineXML(xml)
+            domain.create()
+
+    # Delete all the VMs.
+    def _delete_nodes(
+        self, environment: Environment, log: Logger, qemu_conn: libvirt.virConnect
+    ) -> None:
+        for node in environment.nodes.list():
+            node_context = get_node_context(node)
+            print(f"Delete {node_context.vm_name}")
+
+            # Shutdown and delete the VM.
+            self._stop_and_delete_vm(environment, log, qemu_conn, node)
+
+            # Delete the files created for the VM.
+            try:
+                os.remove(node_context.os_disk_file_path)
+            except Exception as ex:
+                log.warning(f"OS disk delete failed. {ex}")
+
+            try:
+                os.remove(node_context.cloud_init_file_path)
+            except Exception as ex:
+                log.warning(f"cloud-init ISO file delete failed. {ex}")
+
+    # Delete a VM.
+    def _stop_and_delete_vm(
+        self,
+        environment: Environment,
+        log: Logger,
+        qemu_conn: libvirt.virConnect,
+        node: Node,
+    ) -> None:
+        node_context = get_node_context(node)
+
+        # Find the VM.
+        try:
+            domain = qemu_conn.lookupByName(node_context.vm_name)
+        except libvirt.libvirtError as ex:
+            log.warning(f"VM delete failed. Can't find domain. {ex}")
+            return
+
+        # Stop the VM.
+        try:
+            # In the libvirt API, "destroy" means "stop".
+            domain.destroy()
+        except libvirt.libvirtError as ex:
+            log.warning(f"VM stop failed. {ex}")
+
+        # Undefine the VM.
+        try:
+            domain.undefineFlags(
+                libvirt.VIR_DOMAIN_UNDEFINE_MANAGED_SAVE
+                | libvirt.VIR_DOMAIN_UNDEFINE_SNAPSHOTS_METADATA
+                | libvirt.VIR_DOMAIN_UNDEFINE_NVRAM
+                | libvirt.VIR_DOMAIN_UNDEFINE_CHECKPOINTS_METADATA
+            )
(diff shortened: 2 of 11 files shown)


pyproject.toml
```diff
@@ -31,6 +31,8 @@ pep8-naming = "^0.12.0"
 types-toml = "^0.1.5"
 Pillow = "^8.3.1"
 azure-storage-file-share = "12.4.0"
+libvirt-python = {version = "^7.10.0", platform = 'linux'}
```
squirrelsc, line 34: What happens on Windows? It needs to keep Windows can run with other LISA platforms.

cwize1: Yeah, this is something I need to fix. I need to figure out a way to not load the QEMU orchestrator on Windows platforms.

squirrelsc: you may refer to `lisa\runners\legacy_runner.py`. It imports some modules in Windows only. You may put a mockup class for libvirt to pass type checking with the similar way.

squirrelsc:
We manage commits to reflect the code changes in history. It needs each commit is meaningful and independent (well, it looks like at least). Please try to use `git add/reset -p`, `git rebase -i origin/main` and other commands to pick, sort the changes. For some big change like init a platform, it doesn't need a separated change. But for some changes, which involves the platform, it's better to make it meaningful, and small.

lisa/util/constants.py
```diff
@@ -136,3 +136,6 @@
 DEVICE_TYPE_GPU = "GPU"
 
 DISK_PERFORMANCE_TOOL = "fio"
+
+# special variables
+SPECIAL_VARIABLE_RUNBOOK_DIR = "runbook_dir"
```
squirrelsc: 1. the constants.RUNBOOK_PATH is the runbook's path.
2. I prefer not to create hidden variable, because it breaks the readability of the runbook.

microsoft/runbook/qemu/CBL-Mariner.yml
```diff
@@ -0,0 +1,6 @@
+name: qemu default
+include:
+  - path: ./qemu.yml
+variable:
+  - name: extra_user_data
+    value: "$(runbook_dir)/CBL-Mariner/user-data"
```
squirrelsc: It can be a relative path and resolve it in platform, so it doesn't need the runbook_dir variable.

microsoft/runbook/qemu/qemu.yml
```diff
@@ -0,0 +1,21 @@
+name: qemu default
+include:
+  - path: ../tiers/tier.yml
+variable:
+  - name: keep_environment
+    value: "no"
+  - name: qcow2
+    value: ""
+  - name: extra_user_data
```
squirrelsc, line 9: It looks the `extra_user_data` is a required parameter. It doesn't need to be in this level of runbook. So if someone uses this runbook directly, there is an error message says it's lack of this variable.

lisa/util/constants.py
cwize1: Oh, I was looking for `RUNBOOK_PATH` in the `Environment` object. I wasn't expecting it to be a global variable.

microsoft/runbook/qemu/CBL-Mariner.yml
cwize1: It is worth noting that `runbook_dir` is for the current runbook, not the base runbook. This has the nice advantage of making it easier to organize runbooks in directories. The orchestrator code can't mimic this behavior because all the data has been merged and the context of which runbook a value came from has been lost.

Though I guess I don't actually need this extra complexity at the moment. So, I am happy to remove it.

cwize1:
> We manage commits to reflect the code changes in history. It needs each commit is meaningful and independent (well, it looks like at least). Please try to use `git add/reset -p`, `git rebase -i origin/main` and other commands to pick, sort the changes. For some big change like init a platform, it doesn't need a separated change. But for some changes, which involves the platform, it's better to make it meaningful, and small.

Fair.

microsoft/runbook/qemu/qemu.yml
cwize1: It isn't actually required. Granted most distros will probably need to use it. But the orchestrator doesn't require this to have a value.

lisa/util/package.py
```diff
@@ -110,3 +127,23 @@ def import_package(
...
+# Check if a file should be skipped because it is not supported on the current system.
+def _platform_filter(file: Path, root_path: Path) -> bool:
+    relative_file_path = file.relative_to(root_path)
+
+    # Check against the file name and all its parent directories.
+    for path in itertools.chain([relative_file_path], relative_file_path.parents):
+        path_str = path.as_posix()
+
+        # Check if module is subject to platform filterting.
+        if path_str in PLATFORM_ONLY_MODULES_ALL:
```
squirrelsc: Add debug log to know which is excluded, and the reason.

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -0,0 +1,23 @@
...
+@dataclass_json()
+@dataclass
+class QemuPlatformSchema:
+    pass
+
+
+@dataclass_json()
+@dataclass
+class QemuNodeSchema:
+    qcow2: str = ""
```
squirrelsc, line 29: Add comments above the variable, it helps to understand what is used for.

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -0,0 +1,23 @@
...
+@dataclass_json()
+@dataclass
+class CloudInitSchema:
+    extra_user_data: Optional[str] = None
+
+
+@dataclass_json()
+@dataclass
+class QemuPlatformSchema:
+    pass
```
squirrelsc, line 19: Do you have a design about the platform schema? In my mind, it should list all available servers, and its resources (cpu, memory). So the LISA can manage the resource to maximize the test capability. We can start from simple one like below,

```python
class QemuServer(schema.NodeSpace):
    #  the resource is available for testing. For example, if a server has 24 cpus, there may be 22. The rest 2 cpus is left for host.
    # rather than what the node space supported. Extend it to qemu specified settings on server level.
    ...

class QemuPlatformSchema:
    servers: List[QemuServer]

```

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -0,0 +1,447 @@
...
+                with open(extra_user_data_file_path, "r") as file:
+                    node_context.extra_cloud_init_user_data = yaml.safe_load(file)
+
+    # Create all the VMs.
+    def _create_nodes(
+        self, environment: Environment, log: Logger, qemu_conn: libvirt.virConnect
+    ) -> None:
+        for node in environment.nodes.list():
+            # Create cloud-init ISO file.
+            self._create_node_cloud_init_iso(environment, log, node)
```
squirrelsc, line 124: Does it need to create iso for each node? In theory, LISA supports to deploy different type of nodes in the same environment, but I didn't verify it.

lisa/sut_orchestrator/qemu/schema.py
cwize1: Currently, the VMs are just created on the host that is running the test suite. And the VMs are created using a fixed size.

In a subsequent change, I will add support for dynamic VM sizes based on the minimum test requirements. Though I will probably just assume the host has the resources required to run the tests that are executed and let the tests fail if it doesn't.

I don't currently have any plans to support remote/multiple VM hosts or maximizing resource usage. Though we can add these features in the future if their absence is ever felt.

lisa/sut_orchestrator/qemu/platform.py
cwize1: The cloud-init ISO sets the hostname to be the same as the VM name to assist with debugging.

lisa/sut_orchestrator/qemu/schema.py
squirrelsc: 1. Resource capacity on one host. It's ok to unlimited for now. It doesn't break the schema, if we add it later. LISA supports run test cases in parallel, it can speed up test a lot. If no capability supported, we may limit it to 1 or 2 to avoid out of resource.
2. Multiple VM hosts. It may be a break change in future, if we want to support it later. I'm not clear the test cases, which plan to run in qemu. Some test cases like cross host network traffic, live migration needs multiple hosts.
3. Remote host supports. It's ok for now. If the lisa can noot run remotely, it may not be flexible to schedule runs in ADO.

lisa/util/package.py
```diff
@@ -24,6 +26,19 @@
 
 """
 
+LINUX_ONLY_MODULES = {
+    "sut_orchestrator/qemu",
```
squirrelsc: It's good to add comments to explain why it's excluded. I know the libvirt doesn't support Windows, but we can let others know.

squirrelsc: How about use the regular expression? It's powerful and makes the following logic simpler. I refer to use `re.compile` here to add compiled patterns for speeding up.

lisa/sut_orchestrator/qemu/schema.py
cwize1: RE 2: If we add the server list in the future, if the list is empty then we can assume that the current host should be used. Otherwise, we use the list of servers provided. So, I don't think there is a breaking change concern.

RE 3: You can use custom agents in ADO that support nested VMs. We already do this for some of the Mariner+Kubernetes testing.

lisa/util/package.py
cwize1: I decided to use the .gitignore format.

squirrelsc: Is there a module supporting the .gitignore format? The LISA use the "re" module a lot, to keep consistent, I prefer to use regular expression.

cwize1: Yeah, I found the `gitignore_parser` package. While it isn't exactly a well known package, the codebase is small enough that I was able to quickly scan through it and found nothing of concern.

cwize1: Hmm, I am thinking I should probably design something that also works for extensions, including for tests that live outside of this repo.

squirrelsc: If it needs in extensions, it can be added in the extension's schema. It's not a high priority feature, because it can be avoided by create multiple extension folders.

lisa/mixin_modules.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+# The file imports all the mix-in types that can be initialized
+# using reflection.
+
+import platform
```
squirrelsc, line 7: What's the reason to change it like this way? It creates another place to update, and hard to troubleshoot if someone added a new module.

lisa/util/package.py
cwize1: I tried something different: Instead of dynamically loading all the missing modules, I just manually `import` all the missing modules. This allows for the more pythonic pattern of adding `if platform.system() == "Linux"` style checks over the `import` statements.

I left dynamic package loading in place for extensions. Though I did add the ability to only load a single python file. So, if an extension has platform specific imports, then they can create a `module.py` file that handles the imports correctly and then only load the `module.py` file for the extension.

Thoughts on this approach?

lisa/mixin_modules.py
cwize1: Using `if platform.system() == "Linux"` checks over `import` statements is the pythonic way of handling platform specific imports. Doing it this way, allows us to use that pattern.

lisa/util/package.py
squirrelsc: Let's discuss about extension only here. Actually, I have similar requirements to exclude some of files in an extension folder. But it should be dynamically, decide by runbook, not in code. For example, there is a NSG rule modification extension, it should apply to MS subscriptions only.
I think we can add something into extension's schema and implement like below.

```diff
class Extension:
    path: str
    name: Optional[str] = None
+   excludes: Optional[List[str]] = None
```

lisa/mixin_modules.py
squirrelsc: The main concern is it's a static way. It's hard to troubleshoot, if someone forget to update this file.

lisa/util/package.py
cwize1: It is fairly common for test execution frameworks to have a concept of tagging (e.g. MSTest's [TestCategory](https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.testcategoryattribute?view=visualstudiosdk-2022)). It is possible your scenario might be better served by such a feature.

lisa/mixin_modules.py
cwize1: If I forget to add the QEMU platform then the error is: `cannot find subclass 'qemu' of runbook Platform`. That seems fairly easy to diagnose that there is something wrong with the imports.

lisa/util/package.py
squirrelsc: There are some tags like Platform already. It uses to limit test case on supported platforms. The NSG scenario is not related to test case actually, it's related to environments. so it's not fit to be a tag on test case, it should be configurable in runbook. Anyway, I will implement it later. Let me know, if it needs higher priority.

lisa/mixin_modules.py
squirrelsc: Ok, we can go with this way.

lisa/util/package.py
cwize1: If you were to add a generic tagging feature, then it would probably make sense to allow tag filters to be specified in the runbook.

lisa/mixin_modules.py
cwize1: Do you want me to split off this part of the change?

lisa/util/package.py
squirrelsc: There is tag in test suite metadata. But it doesn't work for module loading, the module loading happens very early. The tag is used to categorize tests, not for other purposes. We can discuss about the scenarios to see what's needed here.

lisa/mixin_modules.py
squirrelsc: Well, I prefer the previous solution, which uses a dict to deal with special modules. It doesn't need an extra line for normal modules. But this solution is ok to me too. It's not a big deal.

lisa/util/package.py
cwize1: Is your scenario that you need to apply special network firewall rules for VMs when testing under a MS subscription?

squirrelsc: That's same to all MS subscriptions. If a machine, which is not a part of corp network, it needs an extra NSG rule to SSH VM in MS subscription.

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+# QEMU orchestrator's per-node configuration options.
+@dataclass_json()
+@dataclass
+class QemuNodeSchema:
```
dsrivastavv, line 25: @cwize1 We already have a Qemu tool that can spin up a nested VM on Azure or ready platform. Can you please clarify the additional functionalities this platform will add?

cwize1: I believe that tool is for testing nested VM functionality. Whereas this orchestrator is for running the LISA test framework on a Linux (KVM/QEMU) box.

squirrelsc:
@cwize1 I see it's not a draft version. how is the progress? is it ready to merge?

cwize1:
> @cwize1 I see it's not a draft version. how is the progress? is it ready to merge?

Yeah, it's ready to merge (assuming no further comments). Though I do plan on adding additional features (serial console, data disks, etc.) over time.

squirrelsc:
> > @cwize1 I see it's not a draft version. how is the progress? is it ready to merge?
> 
> Yeah, it's ready to merge (assuming no further comments). Though I do plan on adding additional features (serial console, data disks, etc.) over time.

Thank you. I will double check if. If there is no other comment, I will approve and merge it.

squirrelsc:
BTW, wan you add usage documents of qemu in next PR?

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -0,0 +1,447 @@
...
+            domain = qemu_conn.defineXML(xml)
+            domain.create()
+
+    # Delete all the VMs.
+    def _delete_nodes(
+        self, environment: Environment, log: Logger, qemu_conn: libvirt.virConnect
+    ) -> None:
+        for node in environment.nodes.list():
+            node_context = get_node_context(node)
+            print(f"Delete {node_context.vm_name}")
```
squirrelsc: Remove it or change it to log. You can always use `log.debug` to output info, so it's easy to keep them at the final, if it's needed.

cwize1: Good catch.

lisa/sut_orchestrator/qemu/schema.py
dsrivastavv: Got it! Thanks for clarification.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -0,0 +1,447 @@
...
+        environment: Environment,
+        log: Logger,
+        qemu_conn: libvirt.virConnect,
+        node: Node,
+        timeout: float,
+    ) -> str:
+        node_context = get_node_context(node)
+
+        while True:
+            addr = self._try_get_node_ip_address(environment, log, qemu_conn, node)
```
squirrelsc, line 414: There are retry and timeout module, which can be used here. Search `@retry` or `@func_set_timeout` to learn from code. Note, if the timeout is used, make sure to capture the timeout exception separately like other usages.

cwize1: In my current code, a single timeout time is shared between all the nodes. This stops the total timeout length from increasing with the number of nodes.

squirrelsc (approved):

PR #1678, opened 2022-01-04
add method to install epel

mcgov:
Add a function to enable epel repositories for fedora/redhat/centos

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -1034,6 +1034,21 @@ class Fedora(Linux):
 
         return False
 
+    def install_epel(self) -> None:
+        # Extra Packages for Enterprise Linux (EPEL) is a special interest group
+        # (SIG) from the Fedora Project that provides a set of additional packages
+        # for RHEL (and CentOS, and others) from the Fedora sources.
+
+        major = self._node.os.information.version.major
+        assert_that(major).described_as(
+            "Fedora/RedHat version must be greater than 7"
+        ).is_greater_than_or_equal_to(7)
+        epel_release_rpm_name = f"epel-release-latest-{major}.noarch.rpm"
+        self.install_packages(
+            f"https://dl.fedoraproject.org/pub/epel/{epel_release_rpm_name}"
+        )
+        self._update_packages()
+
     def _get_information(self) -> OsInformation:
         cmd_result = self._node.execute(
             # Typical output of 'cat /etc/fedora-release' is -


lisa/operating_system.py
```diff
@@ -1034,6 +1034,24 @@ def _package_exists(self, package: str) -> bool:
 
         return False
 
+    def install_epel(self) -> None:
```
squirrelsc, line 1037: What's the epel for? Why we need a special logic to install it in os level?

mcgov: it's the 'extra packages for enterprise linux' that fedora provides by default but other distros need to add explicitly. It contains a bunch of useful packages not specifically supported by RedHat but that are very commonly used. It's a requirement when building programs on RHEL and CentOS from source a lot of the time so adding a function for it makes sense.

squirrelsc: ok, please add it to the comment of this method.

lisa/operating_system.py
```diff
@@ -1034,6 +1034,24 @@ def _package_exists(self, package: str) -> bool:
...
+        major = self._node.os.information.version.major
+        assert_that(major).described_as(
+            "Fedora/RedHat version must be greater than 7"
+        ).is_greater_than_or_equal_to(7)
+        epel_release_rpm_name = f"epel-release-latest-{major}.noarch.rpm"
+        self._install_packages(
+            [f"https://dl.fedoraproject.org/pub/epel/{epel_release_rpm_name}"]
+        )
+        self._node.execute(
+            "sudo yum update -y --disablerepo='*' --enablerepo='*microsoft*'",
```
LiliDeng: you want to check epel works well using this command? why only enable microsoft repo?
how about removing epel repo if epel doesn't work on current VM?

mcgov: I had encountered some issue during testing where the regular yum update wasn't working, it must have been some other issue though since regular yum update seems to work fine now that I'm testing this morning. I'll adjust this.

lisa/operating_system.py
```diff
@@ -1034,6 +1034,21 @@ def _package_exists(self, package: str) -> bool:
...
+        # Extra Packages for Enterprise Linux (EPEL) is a special interest group
+        # (SIG) from the Fedora Project that provides a set of additional packages
+        # for RHEL (and CentOS, and others) from the Fedora sources.
+
+        major = self._node.os.information.version.major
+        assert_that(major).described_as(
+            "Fedora/RedHat version must be greater than 7"
+        ).is_greater_than_or_equal_to(7)
+        epel_release_rpm_name = f"epel-release-latest-{major}.noarch.rpm"
+        self._install_packages(
```
squirrelsc: Avoid calling private method if there is no special reason.

squirrelsc: Avoid calling private method if there is no special reason.

mcgov: on update_packages there was a reason, to avoid passing in any package names. I just wanted a generic yum update and update_packages requires a no-op string list or empty string. _update_packages allows None so I went with that one. I'll switch install_packages though there's no reason for that one.

mcgov:
Blocking on testing, don't merge yet plz

squirrelsc (approved):

PR #1679, opened 2022-01-04
DPDK VPP functional test

mcgov:
- add a base test to install vpp and check that it detects the vf
- uses fd.io provided install script
- supports redhat and ubuntu

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -23,10 +23,11 @@ from lisa import (
 from lisa.features import NetworkInterface, Sriov
 from lisa.nic import NicInfo, Nics
 from lisa.testsuite import simple_requirement
-from lisa.tools import Dmesg, Echo, Git, Lsmod, Lspci, Make, Modprobe, Mount
+from lisa.tools import Dmesg, Echo, Git, Ip, Lsmod, Lspci, Make, Modprobe, Mount
 from lisa.util import perf_timer
 from lisa.util.parallel import Task, TaskManager
 from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
+from microsoft.testsuites.dpdk.dpdkvpp import DpdkVpp
 
 VDEV_TYPE = "net_vdev_netvsc"
 MAX_RING_PING_LIMIT_NS = 200000
@@ -131,9 +132,9 @@ class Dpdk(TestSuite):
 
         test_kit = initialize_node_resources(node, log, variables, "failsafe")
         node_nic_info, testpmd = test_kit.node_nic_info, test_kit.testpmd
-        test_nic_id, test_nic = node_nic_info.get_test_nic()
+        test_nic = node_nic_info.get_nic_by_index()
         testpmd_cmd = testpmd.generate_testpmd_command(
-            test_nic, test_nic_id, "txonly", "failsafe"
+            test_nic, 0, "txonly", "failsafe"
         )
         kit_cmd_pairs = {
             test_kit: testpmd_cmd,
@@ -168,6 +169,40 @@ class Dpdk(TestSuite):
                 "than 2^20 (~1m) PPS after sriov disable."
             ).is_less_than(2 ** 20)
 
+    @TestCaseMetadata(
+        description="""
+            verify vpp is able to detect azure network interfaces
+            1. run fd.io vpp install scripts
+            2. install vpp from their repositories
+            3. start vpp service
+            4. check that azure interfaces are detected by vpp
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_nic_count=2,
+            network_interface=Sriov(),
+        ),
+    )
+    def verify_dpdk_vpp(
+        self, node: Node, log: Logger, variables: Dict[str, Any]
+    ) -> None:
+
+        vpp = node.tools[DpdkVpp]
+        vpp.install()
+
+        net = node.nics
+        nic = net.get_nic_by_index()
+
+        # set devices to down and restart vpp service
+        ip = node.tools[Ip]
+        for dev in [nic.lower, nic.upper]:
+            ip.down(dev)
+        for dev in [nic.lower, nic.upper]:
+            ip.addr_flush(dev)
+
+        vpp.start()
+        vpp.run_test()
+
     def _verify_dpdk_build(
         self,
         node: Node,
@@ -180,11 +215,11 @@ class Dpdk(TestSuite):
         node_nic_info, testpmd = test_kit.node_nic_info, test_kit.testpmd
 
         # grab a nic and run testpmd
-        test_nic_id, test_nic = node_nic_info.get_test_nic()
+        test_nic = node_nic_info.get_nic_by_index()
 
         testpmd_cmd = testpmd.generate_testpmd_command(
             test_nic,
-            test_nic_id,
+            0,
             "txonly",
             pmd,
         )
@@ -461,7 +496,7 @@ class DpdkTestResources:
         self.nic_controller = _node.features[NetworkInterface]
         self.dmesg = _node.tools[Dmesg]
         self._last_dmesg = ""
-        test_nic = self.node_nic_info.get_test_nic()[1]
+        test_nic = self.node_nic_info.get_nic_by_index()
         # generate hotplug pattern for this specific nic
         self.vf_hotplug_regex = re.compile(
             f"{test_nic.upper}: Data path switched to VF: {test_nic.lower}"
@@ -503,13 +538,11 @@ def generate_send_receive_run_info(
     rxq: int = 1,
 ) -> Dict[DpdkTestResources, str]:
 
-    (snd_id, snd_nic), (rcv_id, rcv_nic) = [
-        x.node_nic_info.get_test_nic() for x in [sender, receiver]
-    ]
+    snd_nic, rcv_nic = [x.node_nic_info.get_nic_by_index() for x in [sender, receiver]]
 
     snd_cmd = sender.testpmd.generate_testpmd_command(
         snd_nic,
-        snd_id,
+        0,
         "txonly",
         pmd,
         extra_args=f"--tx-ip={snd_nic.ip_addr},{rcv_nic.ip_addr}",
@@ -518,7 +551,7 @@ def generate_send_receive_run_info(
     )
     rcv_cmd = receiver.testpmd.generate_testpmd_command(
         rcv_nic,
-        rcv_id,
+        0,
         "rxonly",
         pmd,
         txq=txq,
@@ -612,8 +645,7 @@ def initialize_node_resources(
         "Test needs at least 1 NIC on the test node."
     ).is_greater_than_or_equal_to(1)
 
-    # bind test nic to desired pmd
-    _, nic_to_bind = node_nic_info.get_test_nic()
+    nic_to_bind = node_nic_info.get_nic_by_index()
 
     # netvsc pmd requires uio_hv_generic to be loaded before use
     if pmd == "netvsc":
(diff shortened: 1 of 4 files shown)


mcgov:
NOTE, depends on 'install_epel' function added in another PR: https://github.com/microsoft/lisa/pull/1678

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -168,6 +169,37 @@ def _check_rx_or_tx_pps(
...
+            min_nic_count=2,
+            network_interface=Sriov(),
+        ),
+    )
+    def verify_dpdk_vpp(
+        self, node: Node, log: Logger, variables: Dict[str, Any]
+    ) -> None:
+
+        vpp = DpdkVpp(node)
+        vpp.install()
```
squirrelsc, line 191: What's the reason not to follow the standard way to get the tool? The installation will be triggered automatically. If write in this way, the tool may be installed multiple times on the same node in some running scenarios.

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -0,0 +1,110 @@
...
+            shell=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                "Could not install vpp with fdio provided installer"
+            ),
+        )
+        node.os.update_packages("")
+        return True
+
+    def install_from_package_manager(self) -> None:
```
squirrelsc: It looks always called after the install. What's the reason to make it as a separated public method?

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -0,0 +1,110 @@
...
+        if isinstance(node.os, Ubuntu):
+            vpp_packages += ["vpp-plugin-dpdk", "vpp-plugin-core"]
+        elif isinstance(node.os, Fedora):
+            vpp_packages.append("vpp-plugins")
+        else:
+            raise UnsupportedDistroException(self.node.os)
+
+        node.os.install_packages(list(vpp_packages))
+
+    def start_vpp(self) -> None:
```
squirrelsc: `start` is enough.

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -0,0 +1,110 @@
...
+                    "Could not set selinux to permissive"
+                ),
+            )
+
+        if isinstance(node.os, Ubuntu) or isinstance(node.os, Fedora):
+            # It is possible the service has already been started, so
+            # rather than assume anything we'll call restart
+            # this will force the reload if it's already started
+            # or start it if it hasn't started yet.
+            node.execute("modprobe uio_hv_generic", sudo=True, expected_exit_code=0)
```
squirrelsc: add error message, if the exit code is not zero.

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -0,0 +1,110 @@
...
+            node.execute("service vpp restart", sudo=True, expected_exit_code=0)
+        else:
+            raise UnsupportedDistroException(
+                os=node.os,
+                message=("VPP start is not implemented for this platform"),
+            )
+        time.sleep(3)  # give it a moment to start up
+
+    def test_vpp(self) -> None:
```
squirrelsc: `run_test`?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -168,6 +169,37 @@ def _check_rx_or_tx_pps(
...
+        vpp.install_from_package_manager()
+
+        net = Nics(node)
+        net.initialize()
+        _, nic = net.get_test_nic()
+
+        # set devices to down and restart vpp service
+        for dev in [nic.lower, nic.upper]:
+            node.execute(f"ip link set {dev} down", sudo=True, expected_exit_code=0)
```
squirrelsc: 1. I created an ip tool to restart the nic by name. BTW, does it lost access, if both low and upper nics are shutdown?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -168,6 +169,37 @@ def _check_rx_or_tx_pps(
...
+        for dev in [nic.lower, nic.upper]:
+            node.execute(f"ip addr flush dev {dev}", sudo=True, expected_exit_code=0)
```
squirrelsc: Add the implementation into the ip.py.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -168,6 +169,37 @@ def _check_rx_or_tx_pps(
                 "than 2^20 (~1m) PPS after sriov disable."
             ).is_less_than(2 ** 20)
 
+    @TestCaseMetadata(
+        description="""
+            verify vpp is able to detect azure network interfaces
```
squirrelsc, line 174: Describe the test steps in details.

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -0,0 +1,110 @@
...
+            ),
+        )
+        node.os.update_packages("")
+        return True
+
+    def install_from_package_manager(self) -> None:
+        node = self.node
+        vpp_packages = ["vpp"]
+
+        if isinstance(node.os, Ubuntu):
```
squirrelsc: How about change it to `Debian`?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: nope, we only do this one the test (second) nic

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I was just following the same way we did this with Testpmd, since we keep the tool in the local test directory I don't think the node.tools[Tool] way works.

microsoft/testsuites/dpdk/dpdkvpp.py
mcgov: Just breaking up the longer running items. I'll change it so it's a private method called from install.

microsoft/testsuites/dpdk/dpdkvpp.py
mcgov: Ah I think that would work actually! I had thought there was an ubuntu dependency but after switching to using their install script it should be supported.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: https://github.com/microsoft/lisa/pull/1686

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: Let's keep them consistent. Other test cases use numbered list.

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: Sorry, I missed this in previews reviews. Calling by node.tools doesn't affect where the tool is installed. It helps to manage the lifecycle of the tool.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -168,6 +169,41 @@ def _check_rx_or_tx_pps(
...
+        ),
+    )
+    def verify_dpdk_vpp(
+        self, node: Node, log: Logger, variables: Dict[str, Any]
+    ) -> None:
+
+        vpp = DpdkVpp(node)
+        vpp.install()
+
+        net = Nics(node)
```
squirrelsc: use `node.nics`. It can cache the information, reduce calls.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -168,6 +169,41 @@ def _check_rx_or_tx_pps(
...
+        net.initialize()
+        _, nic = net.get_test_nic()
```
squirrelsc: 1. how about change the method like `get_nic_by_index(index: int = -1)`. Positive number and zero means find from start, negative number means find from last.
2. I'm not clear about the first int, which is always 0. It's used in one place only. How about hardcode it in where it's used?

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -0,0 +1,116 @@
...
+    VPP_SRC_LINK = "https://github.com/FDio/vpp.git"
+    REPO_DIR = "vpp"
+
+    @property
+    def command(self) -> str:
+        return "vpp"
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Gcc, Make, Git]
```
squirrelsc, line 26: It looks they are not used in this tool, so it's not necessary. The dependencies will be installed, if they are added here. It's unnecessary and slow down the test case.

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -0,0 +1,116 @@
...
+        if isinstance(node.os, Debian):
+            vpp_packages += ["vpp-plugin-dpdk", "vpp-plugin-core"]
+        elif isinstance(node.os, Fedora):
+            vpp_packages.append("vpp-plugins")
+        else:
+            raise UnsupportedDistroException(self.node.os)
+
+        node.os.install_packages(list(vpp_packages))
+
+    def start(self) -> None:
```
squirrelsc, line 38: move public methods above private methods.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I'll raise a second PR to switch both of them to the correct call format, does that work?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: That works, I'll refactor and add a second commit to this PR

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: sure

microsoft/testsuites/dpdk/dpdkvpp.py
mcgov: They are used but are used by the install script VPP provider that we end up using. It looks weird but they are needed to run the script.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Ah nice, does that mean I don't need initialize call then?

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: nevermind I'll just do it here, making lots of changes anyway

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: The initialization is handled in nodes.

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: Do you have a plan to modify the method to accept an index? It may be useful in more case. And the "test_nic" is not clear from the name.

microsoft/testsuites/dpdk/dpdkvpp.py
squirrelsc: I see, please comment above too. It prevents others remove it suddenly in future.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: modified it! Should be good to go

squirrelsc (approved):

PR #1681, opened 2022-01-05
add function check current driver rather than assuming default

mcgov:
Add some code to check which driver is currently bound rather than just assuming the default on a node.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -4,6 +4,7 @@
 
 import os
 import re
+from pathlib import PurePosixPath
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
 
 from assertpy import assert_that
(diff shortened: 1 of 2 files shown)


lisa/nic.py
```diff
@@ -158,6 +158,14 @@ def get_lower_nics(self) -> List[str]:
     def get_device_slots(self) -> List[str]:
         return [x.pci_slot for x in self.nics.values() if x.pci_slot]
 
+    def get_current_device_driver(self, nic: NicInfo) -> str:
+        cmd = f"readlink /sys/class/net/{nic.upper}/device/driver"
+        found_link = self._node.execute(cmd, expected_exit_code=0).stdout
+        driver_name = found_link.split("/")[-1]
```
squirrelsc: Add an example above to help understanding the structure.

lisa/nic.py
```diff
@@ -158,6 +158,14 @@ def get_lower_nics(self) -> List[str]:
...
+        assert_that(driver_name).is_not_equal_to("")
```
squirrelsc: Add description to understand the error, if it happens.

lisa/nic.py
```diff
@@ -158,6 +158,14 @@ def get_lower_nics(self) -> List[str]:
     def get_device_slots(self) -> List[str]:
         return [x.pci_slot for x in self.nics.values() if x.pci_slot]
 
+    def get_current_device_driver(self, nic: NicInfo) -> str:
```
squirrelsc: 1. Pass in the nic_name is easier to use.
2. the name can be simple like `get_driver`.

lisa/nic.py
mcgov: I refactored to make it more clear that it's just checking for no output, switched to using PurePosixPath to emphasize we're grabbing the filename from the path.

squirrelsc: The assert_that will raise an error message into the test result. The error message is only information showing in the PowerBI report. The target is to reduce the chance to check the code, and let user know what's wrong by this error message.

mcgov: Ah, sloppy on my part. will fix

squirrelsc (dismissed):

mcgov:
had to resolve a conflict

squirrelsc (approved):

PR #1685, opened 2022-01-06
improve stability and support base type subscription in notifiers.

squirrelsc:
See commit for details.

### DIFF
--- a/lisa/notifier.py
+++ b/lisa/notifier.py
@@ -16,7 +16,7 @@ from lisa.util.logger import get_logger
 
 @dataclass
 class MessageBase:
-    type: str = ""
+    type: str = "Base"
     elapsed: float = 0
 
 
@@ -44,7 +44,7 @@ class TestRunMessage(MessageBase):
 
 @dataclass
 class PerfMessage(MessageBase):
-    pass
+    type: str = "Performance"
 
 
 DiskSetupType = Enum(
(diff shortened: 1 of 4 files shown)


lisa/runners/lisa_runner.py
```diff
@@ -344,6 +344,11 @@ def _run_test_task(
             case_variables=case_variables,
         )
 
+        # Some test cases may break the ssh connecitons. To reduce side effects
```
dsrivastavv: [nit] *connections

lisa/notifier.py
```diff
@@ -198,10 +198,14 @@ def notify(message: MessageBase) -> None:
                 if len(_message_queue) > 0:
                     current_message = _message_queue.pop()
             if current_message:
-                notifiers = _messages.get(type(current_message))
-                if notifiers:
+                message_types = type(current_message).__mro__
+                for message_type in message_types:
+                    notifiers = _messages.get(message_type, [])
```
dsrivastavv, line 203: Wouldn't this change result in calling notifier unnecessarily for the parent class of a Message?

squirrelsc: This method is called when a new message arrives. And this loop is to check if there is any registered notifier on each inherited message types. If no notifier registers, there is no further action. It's used for handle message type from a super class, like PerfMessage.

LiliDeng (approved):

dsrivastavv (approved):

PR #1686, opened 2022-01-06
Add up/down functions for changing interface status

mcgov:
Add functions for independent up/down status setting outside of the optimized restart function.


mcgov:
Oops, block merge on adding another function

squirrelsc (approved):

PR #1689, opened 2022-01-07
Add network latency and pps perf testing.

LiliDeng:


lisa/notifier.py
```diff
@@ -96,6 +96,30 @@ class DiskPerformanceMessage(PerfMessage):
     randwrite_lat_usec: Decimal = Decimal(0)
 
 
+@dataclass
+class NetworkLatencyPerformanceMessage(PerfMessage):
```
squirrelsc: Let's move Messages into a separated file `messages.py`

microsoft/testsuites/performance/networkperf.py
```diff
@@ -0,0 +1,89 @@
...
+        self.perf_tcp_latency(environment, "perf_tcp_latency_synthetic")
+
+    @TestCaseMetadata(
+        description="""
+        This test case uses lagscope to test sriov network latency.
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_count=2,
+            min_core_count=72,
```
squirrelsc: Does it need to limit the core count? This test case looks related to sriov only.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def perf_tcp_latency(self, environment: Environment, test_case_name: str) -> None:
+        client = cast(RemoteNode, environment.nodes[0])
+        server = cast(RemoteNode, environment.nodes[1])
+        client_lagscope = client.tools[Lagscope]
+        server_lagscope = server.tools[Lagscope]
+        for node in [client, server]:
+            sysctl = node.tools[Sysctl]
+            sysctl.write("net.core.busy_poll", "50")
+            sysctl.write("net.core.busy_read", "50")
```
squirrelsc: What's impact of those settings? Does it need to be restored after test?

microsoft/testsuites/performance/networkperf.py
LiliDeng: I follow the same setting in lisav2, f72 will land into a dedicated host.

lisa/notifier.py
LiliDeng: updated

microsoft/testsuites/performance/networkperf.py
squirrelsc: What's the reason we have to run this case on f72? LISAv3 may use another vm size, which the core is more than 72. If there is no special reason, let's remove the core count.

lisa/tools/lagscope.py
```diff
@@ -0,0 +1,254 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        self._install_dep_packages()
+        tool_path = self.get_tool_path()
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
```
squirrelsc: Choose a release tag or branch to make the test case more stable. It's similar to other source installed tools.

microsoft/testsuites/performance/networkperf.py
LiliDeng: will figure out

lisa/tools/sar.py
```diff
@@ -0,0 +1,114 @@
...
+        #     ICMP6   ICMP traffic    (v6)
+        #     EICMP6  ICMP traffic    (v6) (errors)
+        #     UDP6    UDP traffic     (v6)
+        #     FC      Fibre channel HBAs
+        #     SOFT    Software-based network processing
+        cmd = f"{self.command} -n {key_word} {interval} {count}"
+        process = self.node.execute_async(cmd)
+        return process
+
+    def run_with(
```
squirrelsc: What means `run_with`? Do you want to say `run_with_paramter`? Because the `run` is used for tool, but "with_parameter" is not meaningful, how about `get_statistics`?

lisa/tools/sar.py
```diff
@@ -0,0 +1,114 @@
...
+        tx_rx_pps: List[Decimal] = []
+        for sar_result in raw_list:
+            temp = sar_result_pattern.match(sar_result.group())
+            assert temp, f"not find matched sar result for nic {nic_name}"
+            rx_pps.append(Decimal(temp.group("rxpck")))
+            tx_pps.append(Decimal(temp.group("txpck")))
+            tx_rx_pps.append(
+                Decimal(temp.group("rxpck")) + Decimal(temp.group("txpck"))
+            )
+        message = NetworkPPSPerformanceMessage()
```
squirrelsc, line 108: If we construct message in this method, this method can deal with network message only. So, either change this method to `get_network_statictics`, or return a dict for general purpose. I suggest trying first way, it doesn't need to parse results of other types.

lisa/tools/lagscope.py
LiliDeng: updated

lisa/tools/sar.py
LiliDeng: updated

lisa/tools/sar.py
LiliDeng: updated

lisa/util/shell.py
```diff
@@ -253,6 +254,31 @@ def is_connected(self) -> bool:
             is_inner_shell_ready = True
         return is_inner_shell_ready
 
+    def set_max_connected_sessions_count(self) -> None:
```
squirrelsc: `set_max_session(self, count:int) -> None:`

lisa/tools/lagscope.py
squirrelsc: It can be set in the clone method. It's simpler.

lisa/tools/sar.py
```diff
@@ -0,0 +1,118 @@
...
+    ) -> ExecutableResult:
+        process = self.get_statistics_async(key_word, interval, count)
+        timeout = count + 10
+        return process.wait_result(
+            timeout,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="fail to run sar command",
+        )
+
+    def create_performance_messages(
```
squirrelsc: `create_pps_peformance_messages`?

lisa/util/shell.py
LiliDeng: I will update it in case level.

lisa/tools/sar.py
LiliDeng: updated

lisa/util/shell.py
LiliDeng: updated

microsoft/testsuites/performance/networkperf.py
LiliDeng: the normal size can't guarantee the performance, we choose the size which lands on dedicated host.

lisa/tools/lagscope.py
LiliDeng: updated

lisa/tools/git.py
```diff
@@ -47,6 +47,7 @@ def clone(
         ref: str = "",
         dir_name: str = "",
         fail_on_exists: bool = True,
+        checkout_branch: str = "",
```
squirrelsc: It doesn't need. the ref is which commit to checkout. The branch name, which is checked out, doesn't matter.

lisa/tools/lagscope.py
```diff
@@ -0,0 +1,258 @@
...
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        self._install_dep_packages()
+        tool_path = self.get_tool_path()
+        git = self.node.tools[Git]
+        git.clone(
+            self.repo, tool_path, ref="refs/heads/master", checkout_branch=self.branch
```
squirrelsc: The `ref` is actual checked out branch or tag. The local branch name can be any name, and `git` tool will randomly generate one.

lisa/tools/ssh.py
```diff
@@ -40,3 +43,21 @@ def enable_public_key(self, public_key: str) -> None:
             self.node.get_pure_path("~/.ssh/authorized_keys"),
             append=True,
         )
+
+    def get_path_sshd_config_file(self, file_name: str = "sshd_config") -> str:
```
squirrelsc: `get_sshd_config_path`? The parameter is not necessary, the file name is decided by sshd service, it cannot be changed. Don't over design.

lisa/tools/sysctl.py
```diff
@@ -13,5 +13,18 @@ def command(self) -> str:
...
+    def get(self, variable: str, force_run: bool = True) -> str:
+        result = self.run(
+            variable,
+            force_run=force_run,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"fail to get {self.command} {variable}",
+        )
+        # vm.oom_dump_tasks = 1
+        value = result.stdout.split("=")[-1].strip()
```
squirrelsc: This logic is not stable, what if there is a "=" in the right side? Use regular expression.

lisa/tools/sysctl.py
```diff
@@ -13,5 +13,18 @@ def command(self) -> str:
...
+        # return 1
```
squirrelsc: this comment is unclear and misleading. The return line is clear itself. And "return 1" is from above example, but no explanation here. So, it's confusing a little.

microsoft/testsuites/performance/networkperf.py
squirrelsc: ok, add into description to explain the test case design.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -0,0 +1,216 @@
...
+
+    @TestCaseMetadata(
+        description="""
+        This test case uses sar to test synthetic network PPS (Packets Per Second)
+         when running netperf with single port.
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            min_count=2,
+            min_core_count=32,
```
squirrelsc: Can we have the same criteria for them? some are 72 cores, some are 36 cores. The result is not comparable between sar and other tools.

lisa/tools/sar.py
```diff
@@ -0,0 +1,118 @@
...
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("sysstat")
+        return self._check_exists()
+
+    def get_statistics_async(
+        self, key_word: str = "DEV", interval: int = 1, count: int = 300
```
squirrelsc: Can it be shorter? 5 minutes is too long to get the statistics.

lisa/tools/netperf.py
```diff
@@ -0,0 +1,103 @@
...
+            expected_exit_code_failure_message=f"fail to run {cmd}",
+        )
+
+    def run_as_client_async(
+        self,
+        server_ip: str,
+        core_count: int,
+        port: int = 30000,
+        test_name: str = "TCP_RR",
+        time: int = 400,
```
squirrelsc: does it mean 400 seconds? It's too long. Make it shorter.

lisa/tools/lagscope.py
```diff
@@ -0,0 +1,258 @@
...
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"fail to launch cmd {self.command}"
+            f"{cmd}",
+        )
+
+    def run_as_client(
+        self,
+        server_ip: str,
+        test_interval: int = 0,
+        ping_interval: int = 1000000,
```
squirrelsc: add comments or change the unit to seconds.

microsoft/testsuites/performance/networkperf.py
squirrelsc: The VM size should be selected when run the perf tests. If we're testing the kernel, it needs the high end vm size. But when we test the vm size itself, it doesn't need to specify it. So we should remove those settings, and add it to the perf pipelines.

microsoft/testsuites/performance/networkperf.py
LiliDeng: add comments and restore them after testing

microsoft/testsuites/performance/networkperf.py
LiliDeng: I followed the same setting with lisav2.

squirrelsc: As another comment about the core count, let's remove this limitation in test case level. We can specify the vm size in pipeline.

lisa/tools/sar.py
LiliDeng: I don't think perf testing can get a stable result in short time. but I will have a try.

microsoft/testsuites/performance/networkperf.py
LiliDeng: removed core count

lisa/tools/git.py
LiliDeng: revert

lisa/tools/lagscope.py
LiliDeng: revert

lisa/tools/ssh.py
LiliDeng: updated

microsoft/testsuites/performance/networkperf.py
LiliDeng: remove core count

lisa/tools/sysctl.py
LiliDeng: update code, sysctl -n variable can get the value directly.

lisa/tools/sar.py
LiliDeng: updated into 120

lisa/tools/netperf.py
LiliDeng: update to 150

lisa/tools/lagscope.py
LiliDeng: change the name into ping count

lisa/tools/lagscope.py
```diff
@@ -0,0 +1,257 @@
...
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"fail to launch cmd {self.command}"
+            f"{cmd}",
+        )
+
+    def run_as_client(
+        self,
+        server_ip: str,
+        test_interval: int = 0,
+        ping_count: int = 1000000,
```
squirrelsc: The default value is too big to me. Does it mean to ping 1 million times?

LiliDeng: not sure, I know it will cost 5 mins to finish testing,

squirrelsc: please double check it, and also find a smaller number.

LiliDeng: yes, it means ping 1 million times

LiliDeng: We can set a default smaller value, but for perf testing, I still insist on using this.

squirrelsc: what's value of a big number? The different perf test case has different goal. And we need to get result faster and cheaper. For latency testing, it just needs the average data. For stress or pressure testing, we can run high concurrency. For long haul testing, we run it long time. But the normal perf test cases will be run in many places, we need minimize test scope. So the default value should be minimized, and other stress/pressure tests can overwrite it.

LiliDeng: update to 1000

squirrelsc: How long the 1000 times take?

LiliDeng: it is quite quick (1 second) but with below error, I need update to a proper value.
```02:42:54 INFO: failed to bind socket[3] to a local port: [0.0.0.0:25001]. errno = 98. Ignored```

Comment (LiliDeng) [reply lisa/tools/lagscope.py]: the default value is 1000000, or I can add test duration, to run 2 mins.

image

Comment (squirrelsc) [reply lisa/tools/lagscope.py]: we can try a test duration like 10 seconds. The ping iteration may run too fast to get errors.

Comment (LiliDeng) [reply lisa/tools/lagscope.py]: update to 10 seconds (around 45000 times ping)

File: lisa/tools/lagscope.py [hunk 19]
```diff
@@ -0,0 +1,258 @@
...
+        print_percentile: bool = True,
+        histogram_1st_interval_start_value: int = 30,
+        length_of_histogram_intervals: int = 15,
+        count_of_histogram_intervals: int = 30,
+        dump_csv: bool = True,
+        daemon: bool = False,
+    ) -> List[NetworkLatencyPerformanceMessage]:
+        # -s: run as a sender
+        # -i: test interval
+        # -t: test duration [default: 60 second(s)]
```
squirrelsc: It looks the ping iteration is still here, not change to the -t.

LiliDeng: when -t used, the ping count will be ignored.

squirrelsc: What's the count is used for? we can take care percentage of each range. It doesn't matter how many times it's tested.

LiliDeng: we can specify ping count by -n100, or test duration -t30, when both specified, the ping count will be ignored.

squirrelsc: Please specify -t10, and check if it has about 45k test results. If so, it's good to me.

LiliDeng: I updated them based on actual results.

lisa/tools/lagscope.py
```diff
@@ -0,0 +1,302 @@
...
+            expected_exit_code_failure_message=f"fail to launch cmd {self.command}"
+            f"{cmd}",
+        )
+
+    def run_as_client_async(
+        self,
+        server_ip: str,
+        test_interval: int = 0,
+        run_time_seconds: int = 0,
+        ping_count: int = 1000000,
```
squirrelsc: 1. The default value of async and sync should be the same.
2. If the ping_count isn't effective, it should be removed.

LiliDeng: update for 1
set ping_count as 0 default value

squirrelsc: How about like this way? If one of them is specified, use it. If nothing specified, take the default value `duration=10`. If both of specified, pass in as is. Let the command to decide use which one. But we need to add a comment to explain.

```
duration: Optional[int] = None,
count: Optional[int] = None,
```

squirrelsc (approved):

PR #1690, opened 2022-01-07
Git Tool: add option to recurse submodules

mcgov:
add an option for --recurse-submodules

### DIFF
--- a/lisa/tools/git.py
+++ b/lisa/tools/git.py
@@ -47,7 +47,6 @@ class Git(Tool):
         ref: str = "",
         dir_name: str = "",
         fail_on_exists: bool = True,
-        recurse_submodules: bool = False,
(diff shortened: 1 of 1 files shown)


lisa/tools/git.py
```diff
@@ -47,10 +47,13 @@ def clone(
         ref: str = "",
         dir_name: str = "",
         fail_on_exists: bool = True,
+        recurse_submodules: bool = False,
```
squirrelsc: It looks doesn't need an argument. Let's just bring this in command line always.

squirrelsc: Please remove this argument too.

mcgov: Removed, will squash when I merge

squirrelsc (approved):

PR #1692, opened 2022-01-10
Add support for platform specific modules.

cwize1:
1. Manually import all the mix-in modules, instead of
dynamically importing everything. This will allow for
platform `if` statements over the `import` statements
of platform specific modules.

2. Allow extensions to specify a single python file
instead of only supporting importing an entire
directory. An extension can use this to import a
file that handles any custom import rules they have,
such as using platform `if` statements.

3. Improve sub-class not found exception message.

### DIFF
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -1,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+# The file imports all the mix-in types that can be initialized
+# using reflection.
+
+import lisa.combinators.batch_combinator  # noqa: F401
+import lisa.combinators.csv_combinator  # noqa: F401
+import lisa.combinators.grid_combinator  # noqa: F401
+import lisa.notifiers.console  # noqa: F401
+import lisa.notifiers.env_stats  # noqa: F401
+import lisa.notifiers.html  # noqa: F401
+import lisa.notifiers.text_result  # noqa: F401
+import lisa.runners.legacy_runner  # noqa: F401
+import lisa.runners.lisa_runner  # noqa: F401
+import lisa.sut_orchestrator.azure.hooks  # noqa: F401
+import lisa.sut_orchestrator.azure.transformers  # noqa: F401
+import lisa.sut_orchestrator.ready  # noqa: F401
+import lisa.transformers.kernel_source_installer  # noqa: F401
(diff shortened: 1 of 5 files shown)


lisa/mixin_modules.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+# The file imports all the mix-in types that can be initialized
+# using reflection.
```
squirrelsc, line 5: Can you update the error message in subclasses.py, if the subclass is not found? The error message should be more actionable, either add the imports here, or check the extensions paths.

squirrelsc:
It would be great, if it can be split to two commits. They can be in the same PR.

lisa/util/subclasses.py
```diff
@@ -70,7 +70,9 @@ def load_typed_runbook(self, raw_runbook: Any) -> T_BASECLASS:
         sub_type = self.get(type_name)
         if sub_type is None:
             raise LisaException(
-                f"cannot find subclass '{type_name}' of {self._base_type.__name__}"
+                f"cannot find subclass '{type_name}' of {self._base_type.__name__}. "
+                f"Supported types include: {list(self.keys())}. "
```
cwize1: I added this just in case the problem is just a spelling mistake in the runbook file.

cwize1:
> It would be great, if it can be split to two commits. They can be in the same PR.

Done.

lisa/mixin_modules.py
cwize1: The exception also shows up if you make a spelling mistake in the runbook YAML. So, it might not make sense to refer to internal implementation details.

lisa/util/subclasses.py
squirrelsc: It's great. How about below more?
1. As the three places are the same, create a method called `def _get_sub_type(self, type_name:str) -> T_BASECLASS`. So, the logic can be shared.
2. Let the message be more actionable. For example, "Check the spell of subtype name, and make sure it's added in mixin_modules.py or customized mixin files."

lisa/mixin_modules.py
squirrelsc: Yes, we can provide multiple checks what we can.

squirrelsc (approved):

PR #1693, opened 2022-01-12
storage perf nfs testcases

dsrivastavv:


lisa/tools/nfs.py
```diff
@@ -0,0 +1,102 @@
...
+            self._disable_firewall()
+            self.node.execute(
+                "service nfs restart",
+                sudo=True,
+                expected_exit_code=0,
+                expected_exit_code_failure_message="nfs service failed to start",
+            )
+        elif isinstance(self.node.os, Debian):
+            self.node.execute(
+                "service nfs-kernel-server restart",
```
LiliDeng: service tool can be used here

lisa/tools/nfs.py
```diff
@@ -0,0 +1,102 @@
...
+            self.node.execute(
+                f"echo '{dir_name} {client_ip}(rw,sync,no_subtree_check)' "
+                ">> /etc/exports",
+                shell=True,
+                sudo=True,
+            )
+
+        # restart nfs service
+        if isinstance(self.node.os, Redhat):
+            self._disable_firewall()
```
LiliDeng: can use from microsoft.testsuites.network.common import stop_firewall

squirrelsc: Don't import test suite from a tool. But we can create a virtual tool called firewall.

lisa/tools/nfs.py
```diff
@@ -0,0 +1,102 @@
...
+        mount_dir: str,
+        protocol: str = "tcp",
+    ) -> None:
+        # run installation
+        self._install()
+
+        # mount server shared directory
+        self.node.shell.mkdir(PurePosixPath(mount_dir), exist_ok=True)
+        self.node.tools[Mount].run(
+            f"-t nfs -o proto={protocol},vers=3  "
```
squirrelsc: check if `Mount.mount` can be used. If not, add arguments to make it useful. If it needs to add arguments, make sure their names are the same as mount's help content.

lisa/tools/nfs.py
```diff
@@ -0,0 +1,102 @@
...
+            f"{server_ip}:{server_shared_dir} {mount_dir}",
+            sudo=True,
+        )
+
+    def _disable_firewall(self) -> None:
+        # diable firewall
+        self.node.execute("systemctl stop firewalld")
```
squirrelsc: use the standard service tool

microsoft/testsuites/performance/storageperf.py
```diff
@@ -125,6 +130,190 @@ def perf_premium_datadisks_io(self, node: Node, environment: Environment) -> Non
...
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test performance of nfs server over TCP with
+        VM's initialized with SRIOV network interface.
+        """,
+        priority=2,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            min_count=2,
+            min_core_count=72,
```
squirrelsc: What's the reason to limit core count?

microsoft/testsuites/performance/storageperf.py
```diff
@@ -125,6 +130,190 @@ def perf_premium_datadisks_io(self, node: Node, environment: Environment) -> Non
...
+            server_node,
+            client_node,
+            environment,
+            test_case_name="perf_storage_over_nfs_synthetic_udp_4k",
+            nfs_protocol="udp",
+        )
+
+    def _perf_nfs(
+        self,
+        server_node: RemoteNode,
```
squirrelsc: Change their type to Node, so it doesn't need to cast in each test case.

squirrelsc: And because environment is necessary already, it doesn't need to pass in nodes.

microsoft/testsuites/performance/storageperf.py
dsrivastavv: On a second thought, I think this dependency should not be required

lisa/tools/nfs.py
```diff
@@ -0,0 +1,95 @@
...
+        self.node.shell.mkdir(PurePosixPath(mount_dir), exist_ok=True)
+        self.node.tools[Mount].mount(
+            disk_name=f"{server_ip}:{server_shared_dir}",
+            point=mount_dir,
+            type="nfs",
+            options=f"proto={protocol},vers=3",
+        )
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Redhat):
```
squirrelsc: Does it mean both server and client will install the same tools? It's ok. But maybe we can skip the installation for client. We can have one tool for nfs server only. for the client side, how about use mount directly?

lisa/tools/nfs.py
```diff
@@ -0,0 +1,95 @@
...
+
+        # stop firewall on redhat 7
+        if (
+            isinstance(self.node.os, Redhat)
+            and self.node.os.information.version.major == "7"
+        ):
+            self.node.tools[Firewall].stop()
+
+        # mount server shared directory
+        self.node.shell.mkdir(PurePosixPath(mount_dir), exist_ok=True)
```
squirrelsc: The mount tool will create folder, it doesn't need to do it here.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -125,6 +133,199 @@ def perf_premium_datadisks_io(self, node: Node, environment: Environment) -> Non
...
+                data_disk_iops=search_space.IntRange(min=5000),
+                data_disk_count=search_space.IntRange(min=12),
+            ),
+            network_interface=Sriov(),
+        ),
+    )
+    def perf_storage_over_nfs_sriov_tcp_4k(self, environment: Environment) -> None:
+        self._perf_nfs(
+            environment,
+            test_case_name="perf_storage_over_nfs_sriov_tcp_4k",
```
squirrelsc: I find a way to avoid pass the name itself. Please try if it's possible to get the caller's name in the _perf_nfs.
import inspect
inspect.stack()[0][3] or inspect.stack()[1][3]

microsoft/testsuites/performance/storageperf.py
```diff
@@ -125,6 +133,199 @@ def perf_premium_datadisks_io(self, node: Node, environment: Environment) -> Non
             max_iodepth=64,
         )
 
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test performance of nfs server over TCP with
+        VM's initialized with SRIOV network interface.
+        """,
+        priority=2,
```
squirrelsc: The priority should plus 1. Most perf test cases should be P3. If it runs faster (in 5 minutes), and it's important, we can move some to p2.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -142,6 +343,8 @@ def _perf_premium_datadisks(
         partition_disks = _format_disk(node, data_disks)
         _stop_raid(node)
         _make_raid(node, partition_disks)
+        mount = node.tools[Mount]
+        mount.mount("/dev/md0", "/data", options="nobarrier")
```
squirrelsc: use the filename variable of below line.

lisa/tools/nfs.py
dsrivastavv: The client side also needs tools. For example, `nfs-common` in Debian based distro

microsoft/testsuites/performance/storageperf.py
dsrivastavv: This change is not required. I forgot to remove it.

microsoft/testsuites/performance/storageperf.py
dsrivastavv: Something like this would work but doesn't it make it look more complicated? Also, how about we pass the testname as an argument when the test is called?

squirrelsc: It can be added in the _perf_nfs, so it doesn't need to pass the test name as a parameter.

lisa/tools/nfs.py
squirrelsc: Can we separate it to two tools? one is for server, another one is for client. It will speed up setup.

lisa/tools/nfs_client.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Redhat):
+            self.node.os.install_packages("nfs-utils")
+        elif isinstance(self.node.os, Debian):
+            self.node.os.install_packages("nfs-common")
+        elif isinstance(self.node.os, SLES):
+            pass
+        else:
+            raise LisaException(f"{self.node.os} not supported")
```
squirrelsc: The message shows only one line in message. So it needs to specify what's not supported. There is an UnsupportedException. It needs to use that one, so the test case level can convert it to SkippedException.

lisa/tools/nfs_client.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def setup(
+        self,
+        server_ip: str,
+        server_shared_dir: str,
+        mount_dir: str,
+        protocol: str = "tcp",
+    ) -> None:
+        # run installation
+        self._install()
```
squirrelsc: The installed will be called automatically, it doesn't need to call it in other methods.

lisa/tools/nfs_client.py
```diff
@@ -0,0 +1,49 @@
...
+        if isinstance(self.node.os, Redhat):
+            self.node.os.install_packages("nfs-utils")
+        elif isinstance(self.node.os, Debian):
+            self.node.os.install_packages("nfs-common")
+        elif isinstance(self.node.os, SLES):
+            pass
+        else:
+            raise LisaException(f"{self.node.os} not supported")
+
+        return True
```
squirrelsc: It looks the nfs client is not a standalone tool, so it needs to overwrite the `_check_exists` to detect if it's supported already. For example, run some mount command to see if there is nfs option.

lisa/tools/nfs_server.py
```diff
@@ -0,0 +1,67 @@
...
+    @property
+    def command(self) -> str:
+        return ""
+
+    def can_install(self) -> bool:
+        return True
+
+    def setup(self, client_ips: List[str], dir_name: str) -> None:
+        # run installation
+        self._install()
```
squirrelsc: As client tool. Don't call `_install` in other methods.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -125,6 +143,195 @@ def perf_premium_datadisks_io(self, node: Node, environment: Environment) -> Non
...
+        )
+
+        # Run a dummy fio job on client to create required test files and
+        # transfer over network. If this step is skipped, it can impact
+        # the test result.
+        client_node.tools[Fio].launch(
+            name="prepare",
+            filename=filename,
+            mode=FIOMODES.read.name,
+            time=120,
```
squirrelsc: check if smaller value works.

lisa/tools/nfs_client.py
dsrivastavv: The issue here is that there is no command for NFS so I have set it to empty string. This results in the call to install getting skipped. Please let me know if you have a better way to do so.

microsoft/testsuites/performance/storageperf.py
dsrivastavv: It needs around 1000 seconds time to run. So I set it to 1200 to give it some more slack.

lisa/tools/nfs_client.py
dsrivastavv: The mount command will require the IP of the server otherwise it will fail even if NFS client tools are installed. If we use the IP then this makes tool installation dependent on another node IP. Let me think if there is some other factor I can use for this.

lisa/tools/nfs_client.py
squirrelsc: It should be able to detect if nfs module is installed, like the `mount` or `lsmod`. The `_check_exists` can be overwritten to detect by any other command.

microsoft/testsuites/performance/storageperf.py
squirrelsc: What the "120" mean here? 1200 seconds? The unit is confusing.

dsrivastavv: I think I confused fio timeout with test timeout. The timeout here is "120" second for the fio command. It was taken from v2. I have also observed on ubuntu that it takes appoximately 100 seconds to run it, so 120 sec seems like a good value.

squirrelsc: The default value is 120 too, it doesn't need to specify it here. It's the same for other values, if the default value is used, avoid to specify it to keep it clear.

lisa/tools/firewall.py
```diff
@@ -16,19 +16,19 @@ def can_install(self) -> bool:
         return False
 
     def stop(self) -> None:
-        cmd_result = self.node.execute("command -v ufw")
+        cmd_result = self.node.execute("command -v ufw", shell=True)
```
squirrelsc, line 19: What's the reason it needs the shell?

lisa/tools/nfs_server.py
```diff
@@ -0,0 +1,74 @@
...
+
+class NFSServer(Tool):
+    @property
+    def command(self) -> str:
+        return ""
+
+    def can_install(self) -> bool:
+        return True
+
+    def setup(self, client_ips: List[str], dir_name: str) -> None:
```
squirrelsc: how about call it `create_shared_dir`? It's clearer than "setup".

lisa/tools/firewall.py
dsrivastavv: This step was failing on ubuntu. It was unable to find `ufw` but after adding shell it works fine

lisa/tools/nfs_server.py
dsrivastavv: This step also restarting nfs server and updating nfs exports file. I thought of using name `setup_server` but `_server` postfix is already present in tool name.

squirrelsc: The setup is near to "install", which is in the super class already. From the functionality, it passes in the clients and dir name, so it's more like to create/setup a shared dir, not the server.

dsrivastavv: I can rename it to `create_shared_dir` then

squirrelsc (approved):

PR #1695, opened 2022-01-13
DPDK: add nff-go test case

mcgov:
- add test suite tool
- add test case in dpdksuite

### DIFF
--- a/microsoft/testsuites/dpdk/dpdknffgo.py
+++ b/microsoft/testsuites/dpdk/dpdknffgo.py
@@ -1,0 +1,142 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from typing import List, Type
+
+from lisa.executable import Tool
+from lisa.operating_system import Debian
+from lisa.tools import Echo, Git, Make, Tar, Wget
+from lisa.util import UnsupportedDistroException
+
+
+class DpdkNffGo(Tool):
+
+    NFF_GO_SRC_LINK = "https://github.com/intel-go/nff-go.git"
+    REPO_DIR = "nff-go"
+    GO_TAR = "go1.17.6.linux-amd64.tar.gz"
+
+    ubuntu_packages = [
+        "lua5.3-dev",
+        "libpcap-dev",
+        "libelf-dev",
+        "hugepages",
+        "libnuma-dev",
+        "libhyperscan-dev",
+        "liblua5.3-dev",
+        "libmnl-dev",
+        "libibverbs-dev",
+    ]
+
+    @property
+    def command(self) -> str:
+        return "nff-go"
+
+    def _check_exists(self) -> bool:
+        git_path = self.node.working_path.joinpath(self.REPO_DIR)
+        return (
+            self.node.execute(f"test -a {git_path.as_posix()}", shell=True).exit_code
+            == 0
+        )
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Echo, Make, Git, Tar, Wget]
+
+    @property
+    def can_install(self) -> bool:
+        # nff-go only implemented for debain in lisav2
+        return isinstance(self.node.os, Debian)
+
+    def _install(self) -> bool:
+        node = self.node
+        git = node.tools[Git]
+        echo = node.tools[Echo]
+        wget = node.tools[Wget]
+        tar = node.tools[Tar]
+        make = node.tools[Make]
+
+        # grab the path, a workaround for the issue mentioned below in run_test
+        original_path = echo.run(
+            "$PATH",
+            shell=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="failure to grab $PATH via echo",
+        ).stdout
+
+        self.new_path = f"{original_path}:/usr/local/go/bin/"
+
+        # get nff-go source and go binaries
+        self.nff_go_path = git.clone(
+            "https://github.com/intel-go/nff-go.git",
+            cwd=node.working_path,
+            dir_name=self.REPO_DIR,
+        )
+        go_tar_path = wget.get(
+            f"https://go.dev/dl/{self.GO_TAR}",
+            file_path=str(node.working_path),
+            filename=self.GO_TAR,
+        )
+        # unpack and add to path
+        tar.extract(go_tar_path, "/usr/local")
+
+        # download go modules
+        node.execute(
+            "go mod download",
+            cwd=self.nff_go_path,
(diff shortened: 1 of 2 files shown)


mcgov:
This code is broken curently since update_env doesn't support variable expansion.

mcgov:
`not found command: Command not found: sh.`

microsoft/testsuites/dpdk/dpdknffgo.py
```diff
@@ -0,0 +1,130 @@
...
+from lisa.operating_system import Debian
+from lisa.tools import Echo, Gcc, Git, Make, Tar, Wget
+from lisa.util import UnsupportedDistroException
+
+
+class DpdkNffGo(Tool):
+
+    NFF_GO_SRC_LINK = "https://github.com/intel-go/nff-go.git"
+    REPO_DIR = "nff-go"
+    GO_TAR = "go1.17.6.linux-amd64.tar.gz"
```
squirrelsc, line 16: Can we create a tool for go?

mcgov: We can but I'll do it in a new PR once this is merged

mcgov:
Adding updates for a workaround after talking to chi, hold this PR until those are tested

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -89,6 +90,23 @@ def verify_dpdk_build_failsafe(
...
+            min_nic_count=2,
+            network_interface=Sriov(),
+        ),
+    )
+    def verify_dpdk_nff_go(
+        self, node: Node, log: Logger, variables: Dict[str, Any]
+    ) -> None:
+        nff_go = node.tools[DpdkNffGo]
+        nff_go.run_test()
+        pass
```
squirrelsc: The `pass` doesn't need, if there is content in a method.

microsoft/testsuites/dpdk/dpdknffgo.py
```diff
@@ -0,0 +1,130 @@
...
+                    "libibverbs-dev",
+                    "libpcap-dev",
+                ]
+            )
+        else:
+            raise UnsupportedDistroException(
+                node.os, "nff-go test not implemented on this OS"
+            )
+
+        echo.write_to_file(
```
squirrelsc: Is it useful? We add it in each command already.

microsoft/testsuites/dpdk/dpdknffgo.py
```diff
@@ -0,0 +1,130 @@
...
+            f"https://go.dev/dl/{self.GO_TAR}",
+            file_path=str(node.working_path),
+            filename=self.GO_TAR,
+        )
+        # unpack and add to path
+        tar.extract(go_tar_path, "/usr/local")
+        node.execute(
+            "go mod download",
+            cwd=self.nff_go_path,
+            update_envs={"PATH": "$PATH:/usr/local/go/bin/"},
```
squirrelsc: The PATH appears many times, and it's worth to use a variable.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: oop u right, will fix

microsoft/testsuites/dpdk/dpdknffgo.py
```diff
@@ -0,0 +1,145 @@
...
+        git = node.tools[Git]
+        self.echo = node.tools[Echo]
+        wget = node.tools[Wget]
+        tar = node.tools[Tar]
+        self.make = node.tools[Make]
+        gcc = node.tools[Gcc]
+        gcc.install()
+
+        # grab the path, a workaround for the issue mentioned below in run_test
+        self.path = self.echo.run(
```
squirrelsc: give it a more meaningful name like `original_path`?

microsoft/testsuites/dpdk/dpdknffgo.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        # nff-go only implemented for debain in lisav2
+        return isinstance(self.node.os, Debian)
+
+    def _install(self) -> bool:
+        node = self.node
+        git = node.tools[Git]
+        self.echo = node.tools[Echo]
```
squirrelsc: It doesn't need to save a dependent tool to object level. Even it needs to be used in another method, it can be got again. It doesn't impact the performance, as the tool is reused.

microsoft/testsuites/dpdk/dpdknffgo.py
```diff
@@ -0,0 +1,145 @@
...
+        wget = node.tools[Wget]
+        tar = node.tools[Tar]
+        self.make = node.tools[Make]
+        gcc = node.tools[Gcc]
+        gcc.install()
```
squirrelsc: 1. the install will be called by LISA, it doesn't need to call it.
2. The gcc is a dependency of Make, so it should be installed already.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -89,6 +90,25 @@ def verify_dpdk_build_failsafe(
     ) -> None:
         self._verify_dpdk_build(node, log, variables, "failsafe")
 
+    @TestCaseMetadata(
+        description="""
+           Install and run ci test for NFF-Go on ubuntu
+        """,
+        priority=2,
```
squirrelsc: The new priority should be 3. Add 1 based on LISAv2 test case.

microsoft/testsuites/dpdk/dpdknffgo.py
squirrelsc: Is it changed? It can be something like `self.update_path = f"{self.original_path}:/usr/local/go/bin"`. With this, the original_path is not needed.

mcgov: Should be fixed

squirrelsc: Is it uploaded?

mcgov: I think yes? This comment window is outdated, if you check the files changed bar it shows the current version

squirrelsc: This comment asks to have a variable like below, so the below string doesn't need to write three times in code. But it's still had three times.

`self.updated_path = f"{self.original_path}:/usr/local/go/bin/"`

squirrelsc (approved):

PR #1698, opened 2022-01-17
migrate ntttcp test cases.

LiliDeng:


lisa/messages.py
```diff
@@ -137,3 +137,60 @@ class NetworkPPSPerformanceMessage(PerfMessage):
...
+    receiver_cycles_rer_byte: Decimal = Decimal(0)
+
+
+@dataclass
+class NetworkNtttcpUDPPerformanceMessage(PerfMessage):
+    test_case_name: str = ""
+    test_date: datetime = datetime.utcnow()
+    platform: str = ""
+    location: str = ""
+    host_version: str = ""
```
squirrelsc: The host_version, location and some other fields can be defined in PerfMessage.

lisa/tools/ntttcp.py
```diff
@@ -44,3 +71,204 @@ def get_throughput(self, stdout: str) -> str:
...
+        if dev_differentiator:
+            cmd += f" --show-dev-interrupts {dev_differentiator} "
+        if run_as_daemon:
+            cmd += " -D "
+        result = self.node.execute(
+            f"ulimit -n 204800 && {self.command} {cmd}", shell=True
+        )
+        return result
+
+    def create_ntttcp_result(
```
squirrelsc, line 327: The ntttcp logic is simple and consistent on all related test cases, so it doesn't need to handle the result separately. The create message can be handled in run_as_client. So every time when calling run_as_client, the ntttcp message will be sent.

lisa/tools/ntttcp.py
```diff
@@ -44,3 +71,204 @@ def get_throughput(self, stdout: str) -> str:
...
+            expected_exit_code=0,
+            expected_exit_code_failure_message="fail to lanuch ntttcp server",
+        )
+
+    def run_as_client(
+        self,
+        nic_name: str,
+        server_ip: str,
+        threads_count: int,
+        run_time_seconds: int = 120,
```
squirrelsc: Check if this value can be shorter. The default scenario is to find the throughput, 5 or 10 seconds should be enough.

lisa/tools/ntttcp.py
```diff
@@ -44,3 +71,204 @@ def get_throughput(self, stdout: str) -> str:
...
+            f"-W {warm_up_time_seconds} -C {cool_down_time_seconds} -b {buffer_size}k "
+            f"--show-nic-packets {nic_name} "
+        )
+        if udp_mode:
+            cmd += " -u "
+        if dev_differentiator:
+            cmd += f" --show-dev-interrupts {dev_differentiator} "
+        if run_as_daemon:
+            cmd += " -D "
+        result = self.node.execute(
```
squirrelsc, line 318: Check the result to make sure it succeed.

lisa/util/constants.py
```diff
@@ -136,3 +136,36 @@
 DEVICE_TYPE_GPU = "GPU"
 
 DISK_PERFORMANCE_TOOL = "fio"
+
+NTTTCP_TCP_CONNECTIONS = [
```
squirrelsc: 1. Add them into ntttcp. It's ok to import constants from a tool module like "from lisa.tools.ntttcp import NTTTCP_TCP_CONNECTIONS.
2. The name can be more accurate like NTTTCP_TCP_CONCURRENCY.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -108,14 +115,72 @@ def perf_tcp_max_pps_synthetic(self, environment: Environment) -> None:
...
+        This test case uses ntttcp to test synthetic tcp network throughput.
+        """,
+        priority=3,
+        timeout=12000,
+        requirement=simple_requirement(
+            min_count=2,
+            network_interface=Synthetic(),
+        ),
+    )
+    def perf_tcp_ntttcp_synthetic(self, log: Logger, environment: Environment) -> None:
```
squirrelsc: Let's add one P2 test case to test only 1 typical tcp connections like 128? It's fast and can be test on more scenarios.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -196,15 +247,137 @@ def perf_tcp_pps(
...
+                        client_ntttcp.create_ntttcp_tcp_performance_message(
+                            server_result_temp,
+                            client_result_temp,
+                            client_average_latency,
+                            str(test_thread),
+                            buffer_size,
+                        )
+                    )
+            information: Dict[str, str] = environment.get_information()
+            for ntttcp_message in perf_ntttcp_message_list:
```
squirrelsc, line 429: As mentioned in previous comments, the messages can be sent when it's generated. It doesn't need to be sent at the end of test cases together.

lisa/messages.py
```diff
@@ -137,3 +137,87 @@ class NetworkPPSPerformanceMessage(PerfMessage):
...
+    receiver_cycles_rer_byte: Decimal = Decimal(0)
+    send_buffer_size: Decimal = Decimal(0)
+    tx_throughput_in_gbps: Decimal = Decimal(0)
+    rx_throughput_in_gbps: Decimal = Decimal(0)
+    data_loss: Decimal = Decimal(0)
+    packet_size_kbytes: Decimal = Decimal(0)
+
+
+@dataclass
+class NetworkIperfSingleTCPPerformanceMessage(PerfMessage):
```
squirrelsc: Can it be merged with ntttcp? We can add the tool name as a field.

lisa/tools/iperf3.py
```diff
@@ -47,59 +53,258 @@ def install(self) -> bool:
...
         return self._check_exists()
 
-    def run_as_server(self, daemon: bool = True) -> None:
+    def run_as_server_async(
+        self,
+        port: int = 0,
+        format: str = "",
+        interval_seconds: int = 0,
+        output_json: bool = False,
+        one_off: bool = False,
```
squirrelsc: add comments to explain new parameters.

lisa/tools/iperf3.py
```diff
@@ -47,59 +53,258 @@ def install(self) -> bool:
...
+            f"{self.command} {cmd}", shell=True, sudo=True
+        )
+        return process
+
+    def run_as_server(
+        self,
+        seconds: int = 10,
+        port: int = 0,
+        format: str = "",
+        interval_seconds: int = 0,
```
squirrelsc: give a recommended default value.

lisa/tools/iperf3.py
```diff
@@ -47,59 +53,258 @@ def install(self) -> bool:
...
+        output_json: bool = False,
```
squirrelsc: how about `use_json_format`? Use variable names to explain the logic, because it's simple and straighforward.

lisa/tools/iperf3.py
```diff
@@ -47,59 +53,258 @@ def install(self) -> bool:
...
+        one_off: bool = False,
```
squirrelsc: `one_connection_only` or better name? "one off" doesn't specify for what.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -108,14 +123,146 @@ def perf_tcp_max_pps_synthetic(self, environment: Environment) -> None:
...
     def perf_tcp_max_pps_sriov(self, environment: Environment) -> None:
-        self.perf_tcp_pps(environment, "perf_tcp_max_pps_sriov", "maxpps")
+        self.perf_tcp_pps(environment, "maxpps")
+
+    @TestCaseMetadata(
+        description="""
+        This test case uses ntttcp to test synthetic tcp network throughput.
+        """,
+        priority=3,
+        timeout=12000,
```
squirrelsc: use a constant for the timeout value.

lisa/tools/ntttcp.py
LiliDeng: The final message is parsed from server and client ntttcp result.

lisa/messages.py
LiliDeng: they didn't use the same table

lisa/tools/iperf3.py
LiliDeng: one_off is ==> -1:  handle one client connection then exit

lisa/tools/iperf3.py
LiliDeng: there are other cases user iperf3, but don't use this parameter

lisa/messages.py
LiliDeng: move common ones into PerfMessage.

lisa/tools/ntttcp.py
LiliDeng: update

lisa/util/constants.py
LiliDeng: update

lisa/tools/iperf3.py

lisa/tools/iperf3.py

lisa/tools/iperf3.py

microsoft/testsuites/performance/networkperf.py
LiliDeng: update

lisa/messages.py
squirrelsc: It's ok, we can create a new table. It helps us compare results from different tools. If multiple tools measure the same metrics, they should use the same message. If some tools cannot provide some fields, the fields can be left empty.

lisa/messages.py
```diff
@@ -38,6 +38,19 @@ class TestRunMessage(MessageBase):
...
+    test_case_name: str = ""
+    platform: str = ""
+    location: str = ""
+    host_version: str = ""
+    guest_os_type: str = "Linux"
+    distro_version: str = ""
+    vmsize: str = ""
+    kernel_version: str = ""
+    lis_version: str = ""
+    ip_version: str = "IPv4"
```
squirrelsc: Add Enum class for ip_version, protocol_type like below

```python
class NetworkProtocol(str, Enum):
    IPv4 = "IPv4"
    IPv6 = "IPv6"

class TransportProtocol(str, Enum):
    Tcp = "TCP"
    Udp = "UDP"
```

lisa/tools/iperf3.py
```diff
@@ -47,59 +87,259 @@ def install(self) -> bool:
...
+        # -D: run iperf3 as a daemon
+        # -p: server port to listen on/connect to
+        # -f: [kmgtKMGT] format to report: Kbits, Mbits, Gbits, Tbits
+        # -i: seconds between periodic throughput reports
+        # -1: handle one client connection then exit
+        # -J: output in JSON format
+        process = self.run_as_server_async(
+            port, format, interval_seconds, use_json_format, one_connection_only, daemon
+        )
+        timeout = seconds + 10
```
squirrelsc: Add comments above to explain it.

lisa/tools/ntttcp.py
```diff
@@ -2,16 +2,76 @@
...
 from lisa.executable import Tool
+from lisa.messages import (
+    NetworkNtttcpTCPPerformanceMessage,
+    NetworkNtttcpUDPPerformanceMessage,
+)
 from lisa.tools import Gcc, Git, Make
-from lisa.util.process import ExecutableResult
+from lisa.util.process import ExecutableResult, Process
+
+NTTTCP_TCP_CONCURRENCY = [
```
squirrelsc, line 24: Can they be merged with iperf constants? It doesn't need to take different test parameters for different tools. They can be put into a shared file for perf tools.

lisa/tools/iperf3.py
squirrelsc: 1. It looks the seconds is not used actually.
2. The name seconds and interval_seconds are confusing. It's similar with format and use_json_format.
3. If some parameters are conflict, we can split it to two methods, like run_as_server_by_interval and run_as_server_by_somethingelse.

lisa/tools/ntttcp.py
LiliDeng: they are different

lisa/tools/ntttcp.py
```diff
@@ -2,16 +2,76 @@
...
+    tx_packets: Decimal = Decimal(0)
+    rx_packets: Decimal = Decimal(0)
+    pkts_interrupt: Decimal = Decimal(0)
+    cycles_per_byte: Decimal = Decimal(0)
 
 
 class Ntttcp(Tool):
     repo = "https://github.com/microsoft/ntttcp-for-linux"
     throughput_pattern = re.compile(r" 	 throughput	:(.+)")
+    output_pattern = re.compile(
```
squirrelsc, line 105: add example above.

lisa/messages.py
```diff
@@ -38,6 +38,19 @@ class TestRunMessage(MessageBase):
 @dataclass
 class PerfMessage(MessageBase):
     type: str = "Performance"
+    test_case_name: str = ""
```
squirrelsc, line 57: Because every message needs to initialize those fields, add a method in this module to share like below. Refer to other place like subclasses.py to use generic type.

```python
create_message(message_type: Type[T], node:Node) -> T:
    message = message_type()
    ....
    return T
```

lisa/tools/ntttcp.py
squirrelsc: I know the tcp concurrency is not used in iperf, and I don't know why. But the udp concurrency is used in both. We may add other network perf tools in future. So it needs to manage the built-in parameters together.

lisa/messages.py
LiliDeng: update

lisa/tools/ntttcp.py
LiliDeng: add

lisa/tools/ntttcp.py
LiliDeng: updated into 10 seconds

lisa/tools/iperf3.py
LiliDeng: remove seconds

lisa/messages.py
LiliDeng: > Because every message needs to initialize those fields, add a method in this module to share like below. Refer to other place like subclasses.py to use generic type.
> 
> ```python
> create_message(message_type: Type[T], node:Node) -> T:
>     message = message_type()
>     ....
>     return T
> ```

I am not quite sure about this, could you please take a look at https://github.com/microsoft/lisa/pull/1698/files#diff-aebab7e1684e26efc5eeac6fd09107ce17f2ec4aa641cd40baa10c81c96d1b0fR177?

lisa/messages.py
```diff
@@ -137,3 +127,54 @@ class NetworkPPSPerformanceMessage(PerfMessage):
...
+    retransmitted_segments: Decimal = Decimal(0)
+    congestion_windowsize_kb: Decimal = Decimal(0)
+
+
+@dataclass
+class NetworkIperfUDPPerformanceMessage(NetworkNtttcpUDPPerformanceMessage):
+    pass
+
+
+def create_message_list(messages: List[T], information: Dict[str, str]) -> List[T]:
```
squirrelsc: The key things is to create the message objects in this method, if the message is created outside of the method, it's not useful. It doesn't need to create multiple messages. Try like below,

```python
create_message(message_type: Type[T], node:Node, test_case_name:str="", other_fields: Optional[Dict[str, Any]]=None) -> T:
    message = message_type()
   # fill in other fields from the node
    dict_to_fields(information, message)
    message.test_case_name = test_case_name

    if other_fields:
        dict_to_fields(other_fields, message)
    return message
```

And it's used like below,
```python
        for result in mode_iops_latency.values():
            fio_result_message = create_message(DiskPerformanceMessage, node, test_name, results)
            fio_message.append(fio_result_message)
        return fio_message
```

lisa/messages.py
squirrelsc: replied in another comment

lisa/messages.py
```diff
@@ -137,3 +131,71 @@ class NetworkPPSPerformanceMessage(PerfMessage):
...
+
+@dataclass
+class NetworkIperfUDPPerformanceMessage(NetworkNtttcpUDPPerformanceMessage):
+    pass
+
+
+def create_message(
+    message_type: Type[T],
+    node: "Node",
+    information: Dict[str, str],
```
squirrelsc: Because the information is always from the environment, so we can pass in the environment every time.

lisa/tools/fdisk.py
```diff
@@ -41,12 +44,21 @@ def make_partition(
...
         # fdisk -W always => always to wipe signatures from new partitions
         mkfs = self.node.tools[Mkfs]
-        self.node.execute(
+        cmd_result = self.node.execute(
             f"(echo n; echo p; echo 1; echo ; echo; echo ; echo w) | "
             f"{self.command} -w always -W always {disk_name}",
             shell=True,
             sudo=True,
         )
+        if self.not_support_option_pattern.match(cmd_result.stdout):
```
squirrelsc, line 55: Add comments to explain what happens, and how it's fixed.

microsoft/testsuites/network/sriov.py
```diff
@@ -401,9 +401,11 @@ def verify_sriov_ethtool_offload_setting(self, environment: Environment) -> None
         # iperfResults.log stored client side log
         source_iperf3 = server_node.tools[Iperf3]
         dest_iperf3 = client_node.tools[Iperf3]
-        source_iperf3.run_as_server()
+        source_iperf3.run_as_server_async()
         dest_iperf3.run_as_client_async(
-            server_node.internal_address, client_iperf3_log, 1800
+            server_ip=server_node.internal_address,
+            log_file=client_iperf3_log,
+            run_time_seconds=1800,
```
squirrelsc: It's a p2 test case, does it need to run so long?

microsoft/testsuites/network/stress.py
```diff
@@ -90,7 +90,7 @@ def verify_stress_sriov_iperf(self, environment: Environment) -> None:
         client_iperf3.run_as_client(
             server_ip=matched_server_nic_info.ip_addr,
             log_file=client_iperf3_log,
-            seconds=1800,
+            run_time_seconds=1800,
```
squirrelsc, line 93: make the stress test case to p4.

microsoft/testsuites/network/sriov.py
LiliDeng: There are some testing after running iperf3, and after the testing, the iperf3 will be killed, so actual running time is less than 1800, just make sure that the iperf3 won't quit/interrupt during testing, I will double check the actual running time, then update it to align the actual time and add comments here, since last time you asked the same question.

squirrelsc: The time is terrible here, and I may ask later, after several months... One solution is to remove it totally, and leave it run forever. If some test cases forget to kill it, and cause some problem, we can find it soon.

squirrelsc: Each test case has its timeout, so it won't block test case timeout.

microsoft/testsuites/performance/networkperf.py
LiliDeng: add perf_tcp_ntttcp_128_connections_synthetic

lisa/messages.py
LiliDeng: updated to use environment

lisa/tools/fdisk.py
LiliDeng: comments added

lisa/messages.py
LiliDeng: add tool field and merge the NetworkIperfSingleTCPPerformanceMessage into NetworkTCPPerformanceMessage

microsoft/testsuites/network/stress.py
LiliDeng: set to p4

microsoft/testsuites/network/sriov.py

lisa/messages.py
```diff
@@ -137,3 +135,65 @@ class NetworkPPSPerformanceMessage(PerfMessage):
     rx_tx_pps_minimum: Decimal = Decimal(0)
     rx_tx_pps_average: Decimal = Decimal(0)
     rx_tx_pps_maximum: Decimal = Decimal(0)
+
+
+@dataclass
+class NetworkTCPPerformanceMessage(PerfMessage):
+    tool: str = constants.NETWORK_PERFORMANCE_TOOL_NTTTCP
```
squirrelsc: leave it empty to prevent incorrect put to ntttcp.

lisa/messages.py
```diff
@@ -60,16 +91,7 @@ class PerfMessage(MessageBase):
...
-    test_case_name: str = ""
-    platform: str = ""
-    location: str = ""
-    host_version: str = ""
-    guest_os_type: str = "Linux"
-    distro_version: str = ""
-    vmsize: str = ""
-    kernel_version: str = ""
-    lis_version: str = ""
+    tool: str = constants.DISK_PERFORMANCE_TOOL_FIO
```
squirrelsc: leave it without value like below, so every time it should be filled.
tool: str

microsoft/testsuites/performance/networkperf.py
```diff
@@ -194,17 +340,228 @@ def perf_tcp_pps(
...
+                    {"net.core.wmem_max": "67108864"},
+                ]
+                connections = NTTTCP_UDP_CONCURRENCY
+        try:
+            client_ntttcp = client.tools[Ntttcp]
+            server_ntttcp = server.tools[Ntttcp]
+            client_lagscope = client.tools[Lagscope]
+            server_lagscope = server.tools[Lagscope]
+            stop_firewall(environment)
+            set_systemd_tasks_max([client, server], log)
```
squirrelsc: Add it to the tool, which need it.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -194,17 +340,228 @@ def perf_tcp_pps(
...
+            dev_differentiator = "Hypervisor callback interrupts"
+            if NetworkDataPath.Sriov.value == data_path:
+                server_nic_name = server.nics.get_lower_nics()[0]
+                client_nic_name = client.nics.get_lower_nics()[0]
+                dev_differentiator = "mlx"
+            server_lagscope.run_as_server(ip=server.internal_address)
+            max_server_threads = 64
+            perf_ntttcp_message_list: List[Any] = []
+            for test_thread in connections:
+                if test_thread < max_server_threads:
```
squirrelsc, line 358: Add comments to explain the relationship of thread and buffer size.

lisa/messages.py
LiliDeng: updated

lisa/messages.py
LiliDeng: updated

microsoft/testsuites/performance/networkperf.py
LiliDeng: updated

microsoft/testsuites/performance/networkperf.py
LiliDeng: I am not sure the relationship.

squirrelsc: Nobody knows more than you. The logic set num_threads and buffer_size together. Please try to explain as possible as you can in next PR.

squirrelsc (approved):

PR #1699, opened 2022-01-19
Add xdp test cases and some improvements

squirrelsc:
See commits for details.

### DIFF
--- a/microsoft/testsuites/xdp/xdptools.py
+++ b/microsoft/testsuites/xdp/xdptools.py
@@ -1,0 +1,142 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import re
+from pathlib import PurePath
+from typing import Any, Dict
+
+from lisa import LisaException, Node, UnsupportedDistroException
+from lisa.executable import Tool
+from lisa.operating_system import Debian, Fedora
+from lisa.tools import Ethtool, Git, Make
+from lisa.tools.ethtool import DeviceGroLroSettings
+from lisa.util import find_groups_in_lines
+
+
+def can_install(node: Node) -> bool:
+
+    ethtool = node.tools[Ethtool]
+    statistics = ethtool.get_device_statistics(node.nics.default_nic)
+
+    # check if xdp supported on nic
+    if not any("xdp_drop" in x for x in statistics):
+        raise UnsupportedDistroException(
+            node.os,
+            "Cannot find xdp_drop in ethtool statistics. "
+            "It means it doesn't support XDP.",
+        )
+
+    return True
+
+
+class XdpTool(Tool):
+    """
+    The community xdp tools, it's used to verify XDP by community test cases.
+    """
+
+    _xdp_tools_repo = "https://github.com/xdp-project/xdp-tools.git"
+    _xdp_tools_tag = "v1.2.0"
+    _default_command = "xdptool"
+
+    #     [test_ether_deny]             PASS
+    #     [test_python_basic]           SKIPPED
+    _xdp_test_result_pattern = re.compile(
+        r"^\s+\[(?P<name>\w+)]\s+(?P<result>\w+)\r?$", re.M
+    )
+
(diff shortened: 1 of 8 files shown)


microsoft/testsuites/xdp/xdptools.py
```diff
@@ -0,0 +1,134 @@
...
+    @property
+    def can_install(self) -> bool:
+        ethtool = self.node.tools[Ethtool]
+        statistics = ethtool.get_device_statistics(self.node.nics.default_nic)
+
+        # check if xdp supported on nic
+        if not any("xdp_drop" in x for x in statistics):
+            raise UnsupportedDistroException(
+                self.node.os,
+                "Cannot find xdp_drop in ethotool statistics. "
```
LiliDeng: ethtool

microsoft/testsuites/xdp/xdptools.py
```diff
@@ -0,0 +1,134 @@
...
+        super()._initialize(*args, **kwargs)
+        self._command: PurePath = PurePath(self._default_command)
+        self._gro_lro_settings: Dict[str, DeviceGroLroSettings] = {}
+
+    def _install(self) -> bool:
+        # install dependencies
+
+        config_envs: Dict[str, str] = {}
+        if isinstance(self.node.os, Debian):
+            if self.node.os.information.version < "18.4.0":
```
LiliDeng, line 82: this works for ubuntu, how about debian images, current the latest debian image is 11

microsoft/testsuites/xdp/xdptools.py
```diff
@@ -0,0 +1,134 @@
...
+
+            config_envs.update({"CLANG": "clang-10", "LLC": "llc-10"})
+
+        elif isinstance(self.node.os, Fedora):
+            self.node.os.install_packages(
+                "llvm-toolset elfutils-devel m4 wireshark perf make gcc tc "
+                # pcaplib
+                "http://rpmfind.net/linux/centos/8.5.2111/PowerTools/"
+                "x86_64/os/Packages/libpcap-devel-1.9.1-5.el8.x86_64.rpm"
+            )
```
LiliDeng, line 109: seems it is only for centos 8.5 distro? if lower distro version used, will it fail?

microsoft/testsuites/xdp/xdptools.py
```diff
@@ -0,0 +1,134 @@
...
+            expected_exit_code=0,
+            expected_exit_code_failure_message="failed on configure xdp "
+            "tools before make.",
+        )
+        make = self.node.tools[Make]
+        make.make(
+            arguments="",
+            cwd=self._code_path,
+            update_envs={"ARCH": "x86_64"},
+            thread_count=1,
```
LiliDeng, line 139: why use 1?

squirrelsc: There are errors, if make with multi-threads. The program may not be ready for concurrent build, but our make tool use multi-thread by default.

microsoft/testsuites/xdp/xdptools.py
squirrelsc: The tool is not 1:1 mapping, it's hard to meet all OS versions. I tested it with Redhat 8.4, it works. The Redhat 8.3 doesn't support xdp. so it should be ok.

microsoft/testsuites/xdp/xdptools.py
LiliDeng: please add comments

LiliDeng:
image
one typo in commit message.

microsoft/testsuites/xdp/xdptools.py
squirrelsc: this place is to compat with 1804. Lower version doesn't support xdp, higher version use the default toolchain. I haven't test with debian. Will fix debian problems after test with chaos monkey.

LiliDeng (approved):

PR #1704, opened 2022-01-20
Fix requirements.txt

cwize1:


squirrelsc (approved):

squirrelsc:
@cwize1 thank you for quick fix!

PR #1705, opened 2022-01-20
Write InfiniBand ping-pong test

kamalca:
Uses two VMs to run a simple test of sending data back and forth to ensure RDMA / InfiniBand is functioning.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -2,11 +2,17 @@
 # Licensed under the MIT license.
 
 import re
-from typing import Dict, List, Tuple
+from typing import Any, Dict, List, Tuple
 
 from assertpy import assert_that
 
+from lisa.base_tools import Sed, Uname, Wget
 from lisa.feature import Feature
+from lisa.operating_system import CentOs, Redhat, Ubuntu
+from lisa.tools import Firewall
+from lisa.tools.service import Service
+from lisa.tools.tar import Tar
+from lisa.util import LisaException
 
 FEATURE_NAME_INFINIBAND = "Infiniband"
 
@@ -33,6 +39,10 @@ class Infiniband(Feature):
     #                     port_lmc:               0x00
     #                     link_layer:             InfiniBand
     _ib_info_pattern = re.compile(r"(\s*(?P<id>\S*):\s*(?P<value>.*)\n?)")
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        super()._initialize(*args, **kwargs)
+        self.setup_rdma()
 
     def enabled(self) -> bool:
         return True
@@ -118,3 +128,171 @@ class Infiniband(Feature):
             "No infiniband devices found in /sys/class/infiniband"
         ).is_not_empty()
         return result.stdout.split()
+
+    def setup_rdma(self) -> None:
+        node = self._node
+        # Dependencies
+        kernel = node.tools[Uname].get_linux_information().kernel_version_raw
+        ubuntu_required_packages = [
+            "build-essential",
+            "numactl",
+            "rpm",
+            "libnuma-dev",
+            "libmpc-dev",
+            "libmpfr-dev",
+            "libxml2-dev",
+            "m4",
+            "byacc",
+            "python-dev",
+            "python-setuptools",
+            "tcl",
+            "environment-modules",
+            "tk",
+            "texinfo",
+            "libudev-dev",
+            "binutils",
+            "binutils-dev",
+            "selinux-policy-dev",
+            "flex",
+            "libnl-3-dev",
+            "libnl-route-3-dev",
+            "libnl-3-200",
+            "bison",
+            "libnl-route-3-200",
+            "gfortran",
+            "cmake",
+            "libnl-3-dev",
+            "libnl-route-3-dev",
+            "libsecret-1-0",
+            "dkms",
+            "python-setuptools",
+        ]
+        redhat_required_packages = [
+            "git",
+            "zip",
+            "python3",
+            "kernel-rpm-macros",
+            "gdb-headless",
+            "python36-devel",
+            "elfutils-libelf-devel",
+            "rpm-build",
+            "make",
+            "gcc",
+            "tcl",
+            "tk",
+            "gcc-gfortran",
+            "tcsh",
+            "kernel-devel",
+            "kernel-modules-extra",
+            "createrepo",
+            "libtool",
+            "fuse-libs",
+        ]
+        if isinstance(node.os, CentOs):
+            node.execute(
+                "yum install -y https://partnerpipelineshare.blob.core.windows.net"
+                f"/kernel-devel-rpms/kernel-devel-{kernel}.rpm",
+                sudo=True,
+            )
+
+        if isinstance(node.os, Redhat):
+            if node.os.information.version.major == 7:
+                redhat_required_packages.append("python-devel")
+            else:
+                redhat_required_packages.append("python3-devel")
+                redhat_required_packages.append("python2-devel")
+            node.os.install_packages(list(redhat_required_packages))
+        elif isinstance(node.os, Ubuntu):
+            node.os.install_packages(list(ubuntu_required_packages))
+        else:
+            raise LisaException(f"Unsupported distro: {node.os.name} is not supported.")
+
+        # Turn off firewall
+        firewall = node.tools[Firewall]
+        firewall.stop()
+
+        # Disable SELinux
+        sed = node.tools[Sed]
+        sed.substitute(
+            regexp="SELINUX=enforcing",
+            replacement="SELINUX=disabled",
+            file="/etc/selinux/config",
+            sudo=True,
+        )
+
+        # Install OFED
+        mofed_version = "5.4-3.0.3.0"
+        if isinstance(node.os, Redhat):
+            os_class = "rhel"
+        else:
+            os_class = node.os.name.lower()
+
+        os_version = node.os.information.release.split(".")
+        mofed_folder = (
+            f"MLNX_OFED_LINUX-{mofed_version}-{os_class}"
+            f"{os_version[0]}."
+            f"{os_version[1]}-x86_64"
+        )
+        tarball_name = f"{mofed_folder}.tgz"
+        mlnx_ofed_download_url = (
+            f"https://partnerpipelineshare.blob.core.windows.net/ofed/{tarball_name}"
+        )
+
+        wget = node.tools[Wget]
+        wget.get(url=mlnx_ofed_download_url, file_path=".", filename=tarball_name)
+        tar = node.tools[Tar]
+        tar.extract(file=tarball_name, dest_dir=".", gzip=True)
+
+        extra_params = ""
+        if isinstance(node.os, Redhat):
+            extra_params = (
+                f"--kernel {kernel} --kernel-sources /usr/src/kernels/{kernel}  "
+                f"--skip-repo --skip-unsupported-devices-check --without-fw-update"
+            )
+        node.execute(
+            f"./{mofed_folder}/mlnxofedinstall --add-kernel-support {extra_params}",
+            expected_exit_code=0,
+            expected_exit_code_failure_message="SetupRDMA: failed to install "
+            "MOFED Drivers",
+            sudo=True,
+        )
+        node.execute(
+            "/etc/init.d/openibd force-restart",
+            expected_exit_code=0,
+            expected_exit_code_failure_message="SetupRDMA: failed to restart driver",
+            sudo=True,
+        )
+
+        # Update waagent.conf
+        sed.substitute(
+            regexp="# OS.EnableRDMA=y",
+            replacement="OS.EnableRDMA=y",
+            file="/etc/waagent.conf",
+            sudo=True,
+        )
+        sed.substitute(
+            regexp="# AutoUpdate.Enabled=y",
+            replacement="AutoUpdate.Enabled=y",
+            file="/etc/waagent.conf",
+            sudo=True,
+        )
+
+        service = node.tools[Service]
+        if isinstance(node.os, Ubuntu):
+            service.restart_service("walinuxagent")
+        else:
+            service.restart_service("waagent")
+
+        # Intall Intel MPI
+        wget = node.tools[Wget]
+        script_path = wget.get(
+            "https://partnerpipelineshare.blob.core.windows.net/mpi/"
+            "l_mpi_oneapi_p_2021.1.1.76_offline.sh",
+            executable=True,
+        )
+        node.execute(
+            f"{script_path} -s -a -s --eula accept",
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="SetupRDMA: failed to install IntelMPI",
+        )
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -1,6 +1,10 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import re
+from typing import Dict, List, Tuple
+
+from assertpy import assert_that
 
 from lisa.feature import Feature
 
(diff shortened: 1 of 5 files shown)


kamalca:
@squirrelsc Since this ping pong test is small / fast can we change its priority (technically it was part of the other infiniband test cases which have been given priority 4)? What is your suggestion?

squirrelsc:
> @squirrelsc Since this ping pong test is small / fast can we change its priority (technically it was part of the other infiniband test cases which have been given priority 4)? What is your suggestion?

yes, we can change it to P1, so this feature area can be validated more.

lisa/features/infiniband.py
```diff
@@ -1,13 +1,37 @@
...
+from typing import List, Match, Tuple
 
 from lisa.feature import Feature
+from lisa.util import LisaException
 
 FEATURE_NAME_INFINIBAND = "Infiniband"
 
 
 class Infiniband(Feature):
+    _ib_info_pattern = re.compile(
```
squirrelsc: add an example above.

lisa/features/infiniband.py
```diff
@@ -26,3 +50,58 @@ def is_over_sriov(self) -> bool:
...
+
+    def get_ib_interfaces(self) -> List[Tuple[str, str, str]]:
+        """Gets the list of Infiniband devices
+        excluding any ethernet devices
+        and get their cooresponding network interface
+        Returns list of tuples in the form (ib_device_name, nic_name, ip_addr)
+        Example ("mlx5_ib0", "ib0", "172.16.1.23")"""
+        ib_devices = []
+        device_info = self._get_ib_device_info()
+        print(f"Device Info: {device_info}")
```
squirrelsc: don't use print any time, use `log.debug` always.

kamalca: I forgot to remove this after debugging an issue. It has been removed.

lisa/features/infiniband.py
kamalca: I will add example output, but now I'm thinking I could probably better implement this with something like
`(\s*(?<id>\S*):\s*(?<value>.*)\n?)` multiline
That way I can easily create a dictionary of the values instead of naming every match group.

squirrelsc: yes, you can capture name/value pairs. try `find_groups_in_lines` in `util/__init__.py`

lisa/features/infiniband.py
```diff
@@ -26,3 +50,57 @@ def is_over_sriov(self) -> bool:
...
+            )
+            matched = self._ib_info_pattern.match(result.stdout)
+            if matched:
+                device_info.append(matched)
+
+        if len(device_info) == 0:
+            raise LisaException("Failed to get device info for any InfiniBand devices")
+        return device_info
+
+    def _get_ib_device_list(self) -> List[str]:
```
squirrelsc: how about _get_ib_device_names?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1022,6 +1022,14 @@ def _create_deployment_parameters(
         ):
             arm_parameters.use_availability_sets = True
 
+        if environment.platform and (
```
squirrelsc: Use `assert environment.platform`, so the logic is clearer. It's not a condition of the if, it's to pass static type checking.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1022,6 +1022,14 @@ def _create_deployment_parameters(
...
+            "Infiniband"
+            in [feature.name() for feature in environment.platform.supported_features()]
```
squirrelsc: The logic is not useful. The code is in Azure platform, it supports Infiniband as we know. It needs to check, if the Infiniband is supported by the capability of one of node. If so, it needs below settings.

microsoft/testsuites/hpc/infinibandsuit.py
```diff
@@ -100,3 +101,49 @@ def verify_hpc_over_nd(self, log: Logger, node: Node) -> None:
...
+            ip_addr = interface[2]
+
+            for test in ping_pong_tests:
+                server_process = server_node.execute_async(
+                    f"{test} -d {ib_device_name}"
+                )
+                client_process = client_node.execute_async(
+                    f"{test} -d {ib_device_name} {ip_addr}"
+                )
+                server_result = server_process.wait_result()
```
LiliDeng: need turn off firewall before this testing.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1033,6 +1033,14 @@ def _create_deployment_parameters(
         ):
             arm_parameters.use_availability_sets = True
 
+        if any(
+            x.capability.has_feature(features.Infiniband.name())
```
kamalca, line 1037: @squirrelsc I had to change this away from what you had suggested. Please let me know if this is stylistically bad.

lisa/features/infiniband.py
```diff
@@ -26,3 +52,63 @@ def is_over_sriov(self) -> bool:
...
+        Returns list of tuples in the form (ib_device_name, nic_name, ip_addr)
+        Example ("mlx5_ib0", "ib0", "172.16.1.23")"""
+        ib_devices = []
+        device_info = self._get_ib_device_info()
+        for device in device_info:
+            if device["link_layer"].strip() == "InfiniBand" and "node_guid" in device:
+                device_name = device["hca_id"].strip()
+                guid = device["node_guid"].strip()
+                # Get the last three bytes of guid
+                mpat = f"{guid[12:17]}:{guid[17:19]}"
```
squirrelsc, line 82: give an example here, and it looks not consistent with the comments.

lisa/features/infiniband.py
```diff
@@ -26,3 +52,63 @@ def is_over_sriov(self) -> bool:
     # example SKU: Standard_H16mr
     def is_over_nd(self) -> bool:
         raise NotImplementedError
+
+    def get_ib_interfaces(self) -> List[Tuple[str, str, str]]:
```
squirrelsc, line 56: If those methods are the same to all platform, they should be put in the general feature implementation.

microsoft/testsuites/hpc/infinibandsuit.py
squirrelsc: Please check if it needs to turn off fire wall. There is a tool called Firewall, it's easy to use by call `stop`.

kamalca: This is the process recommended for HPC machines
``` bash
systemctl stop iptables.service
systemctl disable iptables.service
systemctl mask firewalld
systemctl stop firewalld.service
systemctl disable firewalld.service
iptables -nL
sed -i -e's/SELINUX=enforcing/SELINUX=disabled/g' /etc/selinux/config
```

squirrelsc: The test target of this test case is about the IB, so whatever the way to turn off the firewall is ok here. If the current Firewall tool cannot turn off correctly, please fix it.

squirrelsc (dismissed):

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
...
+
+        if isinstance(node.os, Redhat):
+            if node.os.information.version.major == 7:
+                redhat_required_packages.append("python-devel")
+            else:
+                redhat_required_packages.append("python3-devel")
+                redhat_required_packages.append("python2-devel")
+            node.os.install_packages(redhat_required_packages)
+        elif isinstance(node.os, Ubuntu):
+            node.os.install_packages(ubuntu_required_packages)
```
squirrelsc: What about else? Raise an exception for other unsupported OS. So we can see it in log clearly, and may support them later.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
...
+
+        if not node.shell.exists(node.get_pure_path("/bin/python")):
+            print("Python NOT found")
```
squirrelsc: remove test logs.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
...
+            if node.os.information.version.major == 7:
+                redhat_required_packages.append("python-devel")
+            else:
+                redhat_required_packages.append("python3-devel")
+                redhat_required_packages.append("python2-devel")
+            node.os.install_packages(redhat_required_packages)
+        elif isinstance(node.os, Ubuntu):
+            node.os.install_packages(ubuntu_required_packages)
+
+        if not node.shell.exists(node.get_pure_path("/bin/python")):
```
squirrelsc: Please add a python tool.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
...
+            f"MLNX_OFED_LINUX-{mofed_version}-{os_class}"
+            f"{os_version[0]}."
+            f"{os_version[1]}-x86_64"
+        )
+        tarball_name = f"{mofed_folder}.tgz"
+        mlnx_ofed_download_url = (
+            f"https://partnerpipelineshare.blob.core.windows.net/ofed/{tarball_name}"
+        )
+
+        node.execute(
```
squirrelsc: Use the wget tool

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
...
+            f"wget {mlnx_ofed_download_url}",
+            expected_exit_code=0,
+            expected_exit_code_failure_message="MLNX OFED Drivers not found at "
+            f"{mlnx_ofed_download_url}",
+        )
+        node.execute(
```
squirrelsc: use tar

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
...
+            systemctl.restart_service("waagent")
+
+        # Intall Intel MPI
+        wget = node.tools[Wget]
+        script_path = wget.get(
+            "https://partnerpipelineshare.blob.core.windows.net/mpi/"
+            "l_mpi_oneapi_p_2021.1.1.76_offline.sh"
+        )
+        node.execute(
+            f"bash {script_path} -s -a -s --eula accept",
```
squirrelsc: Does it need to add "bash"? what is it used for?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
...
+            sudo=True,
+        )
+        sed.substitute(
+            regexp="# AutoUpdate.Enabled=y",
+            replacement="AutoUpdate.Enabled=y",
+            file="/etc/waagent.conf",
+            sudo=True,
+        )
+
+        systemctl = node.tools[Systemctl]
```
squirrelsc: Use the service, instead of systemctl. The Service tool will go to right one.

microsoft/testsuites/hpc/infinibandsuit.py
```diff
@@ -120,6 +120,8 @@ def verify_ping_pong(self, environment: Environment, log: Logger) -> None:
         server_node = environment.nodes[0]
         client_node = environment.nodes[1]
 
+        # Ensure RDMA is setup
+        client_node.features[Infiniband]
```
squirrelsc: I added a new feature called `run_in_parallel`. It speeds up, especially on executing the same operation on different nodes. Please try it out.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +185,191 @@ def is_over_nd(self) -> bool:
         dmesg = self._node.tools[Dmesg]
         return "hvnd_try_bind_nic" in dmesg.get_output()
 
+    def setup_rdma(self) -> None:
```
squirrelsc: It looks like the logic is not Azure specified. It should be in the generic Feature code, not in the Azure.

lisa/sut_orchestrator/azure/features.py
kamalca: By default, the script is not executable so running it `./{script_path}` requires the extra step of running `chmod +x {script_path}`. But in this case I can probably achieve the same thing by removing back and adding `shell=True`.

lisa/sut_orchestrator/azure/features.py
kamalca: I used systemctl to run `daemon-reload` which has no equivalent in `service`. But I think I can remove that command.

lisa/sut_orchestrator/azure/features.py
squirrelsc: The `chmod +x` can be set in wget when it's downloaded.

microsoft/testsuites/hpc/infinibandsuit.py
```diff
@@ -100,3 +103,58 @@ def verify_hpc_over_nd(self, log: Logger, node: Node) -> None:
...
+        # Constants
+        ping_pong_tests = ["ibv_rc_pingpong", "ibv_uc_pingpong", "ibv_ud_pingpong"]
+
+        server_node = environment.nodes[0]
+        client_node = environment.nodes[1]
+
+        # Ensure RDMA is setup
+        run_in_parallel(
+            [
+                partial(client_node.features[Infiniband].initialize),
```
squirrelsc: If there is no argument for the methods, it doesn't need to use `partial`. The partial is used to wrap with arguments.

kamalca: This doesn't seem to actually run the setup in parallel. I believe that the initialization is called when you access the feature from `node.features`. So how do I pass it a callable object?

lisa/features/infiniband.py
```diff
@@ -26,3 +58,239 @@ def is_over_sriov(self) -> bool:
...
+            if node.os.information.version.major == 7:
+                redhat_required_packages.append("python-devel")
+            else:
+                redhat_required_packages.append("python3-devel")
+                redhat_required_packages.append("python2-devel")
+            node.os.install_packages(list(redhat_required_packages))
+        elif isinstance(node.os, Ubuntu):
+            node.os.install_packages(list(ubuntu_required_packages))
+        else:
+            raise SkippedException(
```
squirrelsc: Let's make it failed so far. We may support them when reviewing the failure. If we decide not to support it in some distro, then use skipped.

microsoft/testsuites/hpc/infinibandsuit.py
kamalca: I used lambda functions. I hope this is a better solution.

squirrelsc: Cool, I should create a get method to features like tools.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -170,6 +170,7 @@ class Infiniband(AzureFeatureMixin, features.Infiniband):
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
+        self.setup_rdma()
```
squirrelsc: Should it be called in the common feature, instead of in Azure? I think it's necessary for every platform.

lisa/features/infiniband.py
```diff
@@ -1,13 +1,49 @@
...
+    #                     active_mtu:             4096 (5)
+    #                     sm_lid:                 55
+    #                     port_lid:               693
+    #                     port_lmc:               0x00
+    #                     link_layer:             InfiniBand
+    _ib_info_pattern = re.compile(r"(\s*(?P<id>\S*):\s*(?P<value>.*)\n?)")
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.setup_rdma()
```
squirrelsc, line 45: Call it in `_initialize`. Let's follow the common practice, not to run long time code in constructor.

squirrelsc (approved):

PR #1707, opened 2022-01-22
Add serial console support to QEMU orchestrator.

cwize1:
1. Add support for libvirt event callbacks. Libvirt requires
setting up a single process-wide event loop for handling the
events. This change provisions a separate Python thread for
this purpose.

2. Create a reader for the libvirt console stream that writes
the console output to file. (Ideally we would use the inbuilt
libvirt support for writing the console to file. But SELinux
doesn't like this for whatever reason.) This mimics the Azure
API that provides the serial console output as a log that can
be downloaded.

3. Implement the orchestrator serial console feature. This
just reads and dumps the console log file.

4. Fix a minor bug in NodeSpace schema class, that causes
a crash for some runbook configurations.

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -774,9 +774,9 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
             search_space.check_countspace(self.memory_mb, capability.memory_mb),
             "memory_mb",
         )
-        if self.disk:
+        if self.disk and capability.disk:
             result.merge(self.disk.check(capability.disk))
-        if self.network_interface:
+        if self.network_interface and capability.network_interface:
             result.merge(self.network_interface.check(capability.network_interface))
         result.merge(
             search_space.check_countspace(self.gpu_count, capability.gpu_count),
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -774,9 +774,9 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
             search_space.check_countspace(self.memory_mb, capability.memory_mb),
             "memory_mb",
         )
-        if self.disk and capability.disk:
+        if self.disk:
             result.merge(self.disk.check(capability.disk))
-        if self.network_interface and capability.network_interface:
+        if self.network_interface:
             result.merge(self.network_interface.check(capability.network_interface))
         result.merge(
             search_space.check_countspace(self.gpu_count, capability.gpu_count),
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -23,11 +23,18 @@ from lisa.util import LisaException, constants, get_public_key_data
 from lisa.util.logger import Logger
 
 from .. import QEMU
+from . import libvirt_events_thread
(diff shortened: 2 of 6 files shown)


lisa/schema.py
```diff
@@ -774,9 +774,9 @@ def check(self, capability: Any) -> search_space.ResultReason:
             search_space.check_countspace(self.memory_mb, capability.memory_mb),
             "memory_mb",
         )
-        if self.disk:
+        if self.disk and capability.disk:
```
squirrelsc: Can you share the detail error? I'm not sure if it hides a bug underlying.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -50,6 +57,8 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
             self._delete_nodes(environment, log, qemu_conn)
 
     def _deploy_nodes(self, environment: Environment, log: Logger) -> None:
+        libvirt_events_thread.init()
```
squirrelsc: what's the life cycle of this thread? If it's longer than the node. Call it in `platform._initialize`. It calls multiple times when 
using combinator, but it can reduce calls than deploy.

lisa/schema.py
squirrelsc: Thank you for the shared log. As the design, the capability must be filled by platform, it shouldn't be skipped, if it's not presented. As the code here, `self.disk` is the requirement to the disk, if it's None, it means there is no requirement on disk. But the `capability.disk` must be presented for comparison. The improvement here is,

1. Add another assertion here to let developer know, the disk/network capability of an environment must be presented.
2. Add capability in the prepare_environment in the platform. You can start with a very basic capability, so it won't fail the test cases, which has disk requirements.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -49,6 +68,31 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
...
             self._delete_nodes(environment, log, qemu_conn)
 
+    # Check what capabilities can be provided for the node.
+    def _create_node_capabilities(
+        self, log: Logger, node_space: schema.NodeSpace
+    ) -> schema.NodeSpace:
+        node_capabilities = schema.NodeSpace()
+        node_capabilities.name = "QEMU"
+        node_capabilities.node_count = 1
+        node_capabilities.core_count = 2
```
squirrelsc, line 78: Make sure the deployment phase will use those values.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -49,6 +68,31 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
...
+        node_capabilities.memory_mb = 4096
+        node_capabilities.disk = schema.DiskOptionSettings()
+        node_capabilities.network_interface = schema.NetworkInterfaceOptionSettings()
+        node_capabilities.network_interface.max_nic_count = 1
```
squirrelsc, line 82: It needs to set `nic_count` too. the `max_nic_count` is used to limited how many nics can be created in some cases, which hot add nics.

squirrelsc (approved):

PR #1717, opened 2022-01-25
Add xdp test case and other fixes

squirrelsc:
See commits for details

### DIFF
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -112,7 +112,7 @@ class QemuPlatform(Platform):
         environment_context = get_environment_context(environment)
 
         # Generate a random name for the VMs.
-        test_suffix = "".join(random.choice(string.ascii_uppercase) for i in range(5))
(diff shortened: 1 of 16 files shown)


lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -68,7 +68,7 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
         environment_context = get_environment_context(environment)
 
         # Generate a random name for the VMs.
-        test_suffix = "".join(random.choice(string.ascii_uppercase) for i in range(5))
+        test_suffix = "".join(random.choice(string.ascii_uppercase) for _ in range(5))
```
cwize1, line 115: Out of interest, was something complaining about this? For me, this passed `make check`.

squirrelsc: The error is about unused variable `i`. It's from pylance, which is the new Python language server of MS. But we're using mypy in the build and UT, so there is no error in make check.

cwize1 (dismissed):

PR #1718, opened 2022-01-27
KVM disk performance testcases

dsrivastavv:


microsoft/testsuites/nested/common.py
```diff
@@ -0,0 +1,79 @@
...
+def get_partition_for_nested_image(node: RemoteNode, size: int) -> str:
+    home_partition = node.tools[Df].get_partition_by_mountpoint("/home")
+    if home_partition and check_partition_capacity(home_partition, size):
+        return home_partition.mountpoint
+
+    mnt_partition = node.tools[Df].get_partition_by_mountpoint("/mnt")
+    if mnt_partition and check_partition_capacity(mnt_partition, size):
+        return mnt_partition.mountpoint
+
+    raise SkippedException(
```
squirrelsc: Why is it skipped by this condition? Is it possible there is any error?

microsoft/testsuites/nested/common.py
```diff
@@ -0,0 +1,79 @@
...
+        "No partition with Required disk space of " f"{size}GB found"
```
squirrelsc: If a string can be write in the same line, merge them in to one string.
`        f"No partition with Required disk space of {size}GB found"`

microsoft/testsuites/nested/common.py
```diff
@@ -0,0 +1,79 @@
...
+        password=guest_password,
+        public_port=guest_port,
+        port=guest_port,
+    )
+
+    return nested_vm
+
+
+def get_partition_for_nested_image(node: RemoteNode, size: int) -> str:
+    home_partition = node.tools[Df].get_partition_by_mountpoint("/home")
```
squirrelsc: call once on `node.tools[Df]` and assign it to a variable for the second use.

microsoft/testsuites/nested/common.py
```diff
@@ -0,0 +1,79 @@
...
+        username=guest_username,
+        password=guest_password,
+        public_port=guest_port,
+        port=guest_port,
+    )
+
+    return nested_vm
+
+
+def get_partition_for_nested_image(node: RemoteNode, size: int) -> str:
```
squirrelsc: 1. make the name more accurate, specify what kind of partition is needed, like `find_partition_with_freespace`.
2. make the name self-explained like size_by_mb.

microsoft/testsuites/nested/common.py
```diff
@@ -0,0 +1,79 @@
...
+    mnt_partition = node.tools[Df].get_partition_by_mountpoint("/mnt")
+    if mnt_partition and check_partition_capacity(mnt_partition, size):
+        return mnt_partition.mountpoint
+
+    raise SkippedException(
+        "No partition with Required disk space of " f"{size}GB found"
+    )
+
+
+def check_partition_capacity(
```
squirrelsc: `is_partition_capable`
Name the method start with `is_`, so people know the return value is a bool.

squirrelsc: If it's private method now, start it with `_`.

microsoft/testsuites/performance/common.py
```diff
@@ -80,3 +80,25 @@ def cleanup_process(environment: Environment, process_name: str) -> None:
     for node in environment.nodes.list():
         kill = node.tools[Kill]
         kill.by_name(process_name)
+
+
+def format_disk(
```
squirrelsc: It looks not like "format", it's like "reset".
`reset_partitions`

microsoft/testsuites/performance/common.py
```diff
@@ -80,3 +80,25 @@ def cleanup_process(environment: Environment, process_name: str) -> None:
...
+    node: Node,
+    disk_list: List[str],
```
squirrelsc: `disk_names`

microsoft/testsuites/performance/storageperf.py
```diff
@@ -241,9 +241,11 @@ def _perf_nfs(
         # calculate the number of jobs to run. The max number of jobs is equal
         # to the core count of the node.
         num_jobs = []
+        iodepth_iter = start_iodepth
         core_count = client_node.tools[Lscpu].get_core_count()
-        for i in range(start_iodepth, max_iodepth + 1):
-            num_jobs.append(min(2 ** (i - 1), core_count))
+        while iodepth_iter <= max_iodepth:
+            num_jobs.append(min(iodepth_iter, core_count))
```
squirrelsc, line 251: Add comments to explain the purpose of this num_jobs, why it's created like this way.

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -0,0 +1,157 @@
...
+        description="""
+        This test case is to validate performance of nested VM using fio tool
+        with single l1 data disk attached to the l2 VM.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            disk=schema.DiskOptionSettings(
+                disk_type=schema.DiskType.PremiumSSDLRS,
+                data_disk_iops=5000,
```
squirrelsc: make it a minimum limitation, not an exact value.

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -0,0 +1,157 @@
...
+    @TestCaseMetadata(
+        description="""
+        This test case is to validate performance of nested VM using fio tool
+        with single l1 data disk attached to the l2 VM.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            disk=schema.DiskOptionSettings(
+                disk_type=schema.DiskType.PremiumSSDLRS,
```
squirrelsc, line 53: What's the reason to limit the disk type in this test?

microsoft/testsuites/performance/common.py
dsrivastavv: This part of refactoring and were written as part of an earlier testcases. I will check with @LiliDeng to understand more about this function

microsoft/testsuites/performance/kvmperf.py
dsrivastavv: This is the default disk type used in all the performance testcases written so far

microsoft/testsuites/nested/common.py
dsrivastavv: The force run is false in the function call. So, it is internally one single call to df

squirrelsc: You can add force_run to the arguments, and give the False as default value. If the code writes like this way, it hides the reason, and may be "improved" in future, and leave a bug here.

microsoft/testsuites/nested/common.py
dsrivastavv: This is skipped if we are unable to find a partition with required capacity

microsoft/testsuites/performance/storageperf.py
dsrivastavv: There is a comment present above: "The max number of jobs is equal to the core count of the node.". Is there something specific that you want to add?

microsoft/testsuites/nested/common.py
squirrelsc: What's the reason not found? If the reason is the environment is broken by last test case, this case should fail to get attention. So, we need to think about what's situation here, if it's unknown, keep this test case fail, instead of skipping.

microsoft/testsuites/performance/storageperf.py
squirrelsc: It's not clear. The core count of this node is a number like "32", but how it comes out an array? And what the "job" is for?

microsoft/testsuites/nested/common.py
dsrivastavv: The reason is that none of the disk has enough capacity. It was marked as skipped because this essentially means that the VM doesn't have the requirement to run the test.

microsoft/testsuites/performance/storageperf.py
dsrivastavv: Sure! I will add an example with more description to make it explicit

microsoft/testsuites/performance/common.py
dsrivastavv: @LiliDeng Can you please confirm chi's suggestion that we should rename it to reset_partitions? It does seem like that we are deleting and re-creasing partitions which is essentially resetting it

microsoft/testsuites/nested/common.py
squirrelsc: My concern is, if it's enough to differentiate not capable by design and by some unknown reason. Normally, if a test case returns skipped, most likely nobody looks the result again. So, we need to either make sure it's skipped by design or return exception for now. So we can check the reason later, if it returns exception.

dsrivastavv: If you feel so, I will mark it as an Exception

microsoft/testsuites/nested/common.py
squirrelsc: So, it's good to use only one `df = node.tools[Df]`

microsoft/testsuites/nested/common.py
```diff
@@ -0,0 +1,122 @@
...
+    mnt_partition = node.tools[Df].get_partition_by_mountpoint("/mnt")
+    if mnt_partition and _is_partition_capable(mnt_partition, size_in_gb):
+        return mnt_partition.mountpoint
+
+    raise LisaException(
+        f"No partition with Required disk space of {size_in_gb}GB found"
+    )
+
+
+def _is_partition_capable(
```
squirrelsc: Move private method below public

microsoft/testsuites/performance/common.py
```diff
@@ -80,3 +80,25 @@ def cleanup_process(environment: Environment, process_name: str) -> None:
...
+        partition_disks.append(fdisk.make_partition(data_disk, format=False))
+    return partition_disks
+
+
+def stop_raid(node: Node) -> None:
+    mdadm = node.tools[Mdadm]
+    mdadm.stop_raid()
+
+
+def make_raid(node: Node, disk_list: List[str]) -> None:
```
squirrelsc: It looks every time the `stop_raid` is called before `make_raid`. How about change this method to `recreate_raid` and call `stop_raid`?

dsrivastavv: Makes sense! I will update this

dsrivastavv: Though I feel adding "re" prefix before the method name might be a bit wordy. How about we let the method name be `make_raid` and I can add `stop_raid` at the beginning of this method?

squirrelsc: The "re" means if exists, then removed. Sometimes people may want something, and it's ok if exists.

squirrelsc (approved):

microsoft/testsuites/performance/storageperf.py
```diff
@@ -141,6 +120,7 @@ def perf_premium_datadisks_io(self, node: Node, environment: Environment) -> Non
             environment,
             test_case_name="perf_premium_datadisks_io",
             max_iodepth=64,
+            filename="/dev/sdc",
```
LiliDeng, line 123: @somil55 this name can't be hardcoded.

PR #1721, opened 2022-01-28
Bug fixes for some tier 1 tests for CBL-Mariner on QEMU.

cwize1:
1. Add OS type for CBL-Mariner. While CBL-Mariner uses RPM
and so should share the package handling logic, it is not
a derivative of Fedora and has substantial differences.
Hence, the RPM handling logic was split off into a new
base class that both `Fedora` and `CBLMariner` classes
inherit from.

2. Fix a few tests that didn't declare their required
`supported_features`.

3. Restrict tests that rely on VMBus to the `AZURE` platform.
(When the Hyper-V support is added, it can be added as a
platform to these tests as well.)

4. Add support for CBL-Mariner to Docker tool.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -948,23 +948,21 @@ class OpenBSD(BSD):
 
 
 @dataclass
-# yum repolist is of the form `<id> <name>`
+# dnf repolist is of the form `<id> <name>`
 # Example:
 # microsoft-azure-rhel8-eus  Microsoft Azure RPMs for RHEL8 Extended Update Support
-class FedoraRepositoryInfo(RepositoryInfo):
+class RPMRepositoryInfo(RepositoryInfo):
     # id for the repository, for example: microsoft-azure-rhel8-eus
     id: str
 
 
-class Fedora(Linux):
-    # Red Hat Enterprise Linux Server 7.8 (Maipo) => 7.8
-    _fedora_release_pattern_version = re.compile(r"^.*release\s+([0-9\.]+).*$")
-
+# Linux distros that use RPM.
+class RPMDistro(Linux):
     # microsoft-azure-rhel8-eus  Microsoft Azure RPMs for RHEL8 Extended Update Support
-    _fedora_repository_info_pattern = re.compile(r"(?P<id>\S+)\s+(?P<name>\S.*\S)\s*")
+    _rpm_repository_info_pattern = re.compile(r"(?P<id>\S+)\s+(?P<name>\S.*\S)\s*")
 
     # ex: dpdk-20.11-3.el8.x86_64
-    _fedora_version_splitter_regex = re.compile(
+    _rpm_version_splitter_regex = re.compile(
         r"(?P<package_name>[a-zA-Z0-9\-_]+)-"
         r"(?P<major>[0-9]+)\."
         r"(?P<minor>[0-9]+)"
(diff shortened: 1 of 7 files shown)


lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -79,7 +79,7 @@ def _create_node_capabilities(
         node_capabilities.memory_mb = 4096
         node_capabilities.disk = schema.DiskOptionSettings()
         node_capabilities.network_interface = schema.NetworkInterfaceOptionSettings()
-        node_capabilities.network_interface.max_nic_count = 1
+        node_capabilities.network_interface.max_nic_count = 0
```
romoh: I missed the initial PR so likely missing some context. Does 0 max sense here?
Unrelated to this PR, I wonder if we want to add number of cpu(s) to the node capability?

romoh (dismissed):

lisa/operating_system.py
```diff
@@ -948,37 +948,31 @@ class OpenBSD(BSD):
...
     )
 
-    @classmethod
-    def name_pattern(cls) -> Pattern[str]:
-        return re.compile("^Fedora|fedora$")
-
     def get_repositories(self) -> List[RepositoryInfo]:
         repo_list_str = self._node.execute(
-            "yum repolist", sudo=True
+            "dnf repolist", sudo=True
```
squirrelsc: dnf is not supported by some Redhat 7.x versions. So, it's using the yum. Can you extract the tool name as a variable, use "yum" by default, and overwrites it to "dnf" in Mariner?

squirrelsc: Ideally, the package management should be split from the os, but it's not done. We can take this as workaround.

cwize1: Why does `_install_packages` use `dnf` if it isn't supported?

squirrelsc: You are right. Some methods are overridden in RedHat. It's not a thoughtful solution, and should be refined later. For now, the cheapest solution is to extract the "dnf" or "yum" out to me, if used commands are compatible.

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: It should be like below,
```python
node_capabilities.network_interface.nic_count = 1
node_capabilities.network_interface.max_nic_count = 1
```

lisa/operating_system.py
```diff
@@ -1240,6 +1243,12 @@ def name_pattern(cls) -> Pattern[str]:
         return re.compile("^Oracle")
 
 
+class CBLMariner(RPMDistro):
```
squirrelsc, line 1252: Can it be simpler like Mariner?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -119,7 +119,7 @@ def verify_grub(self, node: Node) -> None:
             # in core os we don't have access to boot partition
             grub_output = node.tools[Dmesg].run().stdout
         else:
-            raise LisaException(f"Test cannot run on distro {node.os}")
+            raise SkippedException(f"Test cannot run on distro {node.os}")
```
squirrelsc: In our test guideline, for unknown things, they should make the test case fail. If it's confirmed not to test in Mariner, please add a `elif` to skip Mariner only.

lisa/operating_system.py
cwize1: "CBL-Mariner" is the official name.

microsoft/testsuites/network/netinterface.py
```diff
@@ -45,7 +45,10 @@ class NetInterface(TestSuite):
 
         """,
         priority=1,
-        requirement=simple_requirement(network_interface=Synthetic()),
+        requirement=simple_requirement(
+            supported_features=[NetworkInterface],
```
squirrelsc: The network and storage are special features, they should be supported by platforms always. Please create them in platforms.

cwize1: Implementing NetworkInterface and Disk are on the TODO list. But they are too large for this change.

squirrelsc (approved):

PR #1726, opened 2022-02-02
Git: Add function to get latest version in git tags

mcgov:
Add a helper to sort by version and grab the latest git tag

### DIFF
--- a/lisa/tools/git.py
+++ b/lisa/tools/git.py
@@ -5,6 +5,8 @@ import pathlib
 import re
 from typing import List
 
+from assertpy import assert_that
+
 from lisa.executable import Tool
 from lisa.operating_system import Posix
 from lisa.util import LisaException, constants, filter_ansi_escape, get_matched_str
@@ -159,3 +161,39 @@ class Git(Tool):
             expected_exit_code=0,
             expected_exit_code_failure_message="error on init submodules.",
         )
+
+    def get_tag(
+        self,
+        cwd: pathlib.PurePath,
+        sort_by: str = "v:refname",
+        contains: str = "",
+        return_last: bool = True,
+    ) -> str:
+        sort_arg = ""
+        contains_arg = ""
+        if sort_by:
+            # git tag exposes various sort options, apply them if present
+            # default is sort by version, ascending
+            sort_arg = f"--sort={sort_by}"
+        if contains:
+            # git tag allows you to filter by a commit id, apply it is present.
+            contains_arg = f"--contains {contains}"
+
+        git_cmd = f"--no-pager tag {sort_arg} {contains_arg}"
+        tags = self.run(
+            git_cmd,
+            cwd=cwd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                "git tag failed to fetch tags, "
+                "check sort and commit arguments are correct"
+            ),
+        ).stdout.splitlines()
+        assert_that(len(tags)).described_as(
+            "Error: could not find any tags with this sort or filter setting"
+        ).is_greater_than(0)
+
+        if return_last:
+            return tags[-1]
+        else:
+            return tags[0]


lisa/tools/git.py
```diff
@@ -153,3 +153,12 @@ def init_submodules(self, cwd: pathlib.PurePath) -> None:
             expected_exit_code=0,
             expected_exit_code_failure_message="error on init submodules.",
         )
+
+    def get_latest_tag(self, cwd: pathlib.PurePath) -> str:
+        # sort tags by version, grab the last in the list
+        return self.run(
+            "--no-pager tag --sort=v:refname",
```
squirrelsc: the "latest" may be different for each repo. How about expose two arguments, and rename it. So it can return a couple of ordered tags.

`def get_tag(self, cwd: pathlib.PurePath, sort_by:str = "v:refname", filter:str = "", return_first: bool = True) -> str`

lisa/tools/git.py
```diff
@@ -153,3 +153,33 @@ def init_submodules(self, cwd: pathlib.PurePath) -> None:
...
+            git_cmd,
+            cwd=cwd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Could not get latest tagged commit ID",
+        ).stdout
+
+        if return_last:
+            return tags.splitlines()[-1]
+        else:
+            return tags
```
squirrelsc: It supposes to return one tag always, so it's consistent and easy to understand. So if it doesn't return last one, please return the first one.

lisa/tools/git.py
```diff
@@ -153,3 +153,33 @@ def init_submodules(self, cwd: pathlib.PurePath) -> None:
...
+        if contains:
+            # git tag allows you to filter by a commit id, apply it is present.
+            contains_arg = f"--contains {contains}"
+
+        git_cmd = f"--no-pager tag {sort_arg} {contains_arg}"
+        tags = self.run(
+            git_cmd,
+            cwd=cwd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Could not get latest tagged commit ID",
```
squirrelsc: The message should be changed. It's failed on get tags, not matter about latest tagged.

mcgov: Will change, testing rn

lisa/tools/git.py
```diff
@@ -153,3 +153,36 @@ def init_submodules(self, cwd: pathlib.PurePath) -> None:
...
+            git_cmd,
+            cwd=cwd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                "git tag failed to fetch tags, "
+                "check sort and commit arguments are correct"
+            ),
+        ).stdout.splitlines()
+
+        if return_last:
```
squirrelsc, line 196: Sorry forget it at previous reviews. How about assume there is at least one tag?

mcgov: Yep, good catch. Git tag returns success and returns nothing if there aren't tags. Will assert

lisa/tools/git.py
```diff
@@ -153,3 +155,39 @@ def init_submodules(self, cwd: pathlib.PurePath) -> None:
...
+            git_cmd,
+            cwd=cwd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                "git tag failed to fetch tags, "
+                "check sort and commit arguments are correct"
+            ),
+        ).stdout.splitlines()
+        assert_that(len(tags)).described_as(
+            "Error: this repo contained no tags"
```
squirrelsc: Because the method supports filter, so the reason may be like "no tag is found".

squirrelsc (dismissed):

squirrelsc (dismissed):

squirrelsc (approved):

PR #1728, opened 2022-02-03
migrate POWER-CLOCK-SYNC-HIBERNATE

lesscodingmorehappiness:
Part of POWER-CLOCK-SYNC-HIBERNATE migration. Mainly for https://github.com/microsoft/lisa/blob/master/Testscripts/Linux/SetupHbKernel.sh


microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+from lisa.testsuite import TestCaseMetadata
+from lisa.tools.echo import Echo
+from lisa.util import SkippedException
+from lisa.util.logger import Logger
+
+def check_vm_feature_support_status(min_support_version: str, node: Node) -> bool:
+    if not min_support_version:
+        return False
+    
+    kernel_array = node.execute(
```
squirrelsc: use `node.information.version` It contains the distro version, and the uname tool can return a structured `kernel_version`. The versions can be compared like `> "3.10.0"`.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+)
+from lisa.base_tools.cat import Cat
+from lisa.operating_system import SLES, CentOs, Redhat
+from lisa.testsuite import TestCaseMetadata
+from lisa.tools.echo import Echo
+from lisa.util import SkippedException
+from lisa.util.logger import Logger
+
+def check_vm_feature_support_status(min_support_version: str, node: Node) -> bool:
+    if not min_support_version:
```
squirrelsc, line 23: It looks like a wrong call, it should raise exception.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+
+
+@TestSuiteMetadata(
+    area="power",
+    category="functional",
+    description="""
+    """
+)
+class power(TestSuite):
+    HB_CUSTOM_KERNEL_URL = "git://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git"
```
squirrelsc: 1. the private variable should start with `_`.
2. What's the reason the tests use Linux kernel repo?

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+    HB_CUSTOM_KERNEL_BRANCH = "stable"
+    HB_URL = ""
+
+    @TestCaseMetadata(
+        description="""
+        This test can be performed in Azure and Hyper-V both. But this script only covers Azure.
+        1. Prepare swap space for hibernation
+        2. Compile a new kernel (optional)
```
squirrelsc: Can hibernation work without this? If so, this step should be excluded in the test case.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+        priority=3
+    )
+    def power_clock_sync_hibernate(self, environment: Environment, node: Node, log: Logger) -> None:
+        max_kernel_compile_min = 90
+        
+        # Prepare the swap space in the target VM
+        resource_group = 
+        vm_name = 
+        location = 
+        storage_type = "StandardSSD_LRS"
```
squirrelsc: If the disktype is required, use the disk feature. But if there is no special reason, not to limit the disk type, it makes test case cannot run under other conditions.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+        data_disk_name = vm_name + "_datadisk1"
+
+        is_disk_added = False
+        lun = 0
+        # TODO: CHECK IS_DISK_ADDED
+
+
+        test_command = "echo disk > /sys/power/state"
```
squirrelsc: use the echo tool

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+        env_setup_result = node.execute(
```
squirrelsc: use the dmesg tool, and regular expression to get the information.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+            distro = ""
+            # TODO: ADD THIS SECTION
+
+            # GetGuestGeneration
+            exit_code = node.execute(
+                "ls /sys/firmware/efi/"
+            ).exit_code
+            os_generation = 2 if exit_code == 0 else 1
+            
+            if "redhat" in distro and self.HB_URL == "":
```
squirrelsc: check distro type by `isinstance(node.so, RedHat)`.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+                min_support_kernel = "4.18.0-202"
+                check_vm_support_res = check_vm_feature_support_status(min_support_kernel)
+                current_kernel_version = node.execute(
+                    "uname -r", shell = True, sudo = True
+                )
+                assert_that(
+                    check_vm_support_res,
+                    f"Hibernation is supported since kernel-4.18.0-202. Current version: {current_kernel_version}. Skip the test."
+                ).is_true()
```
squirrelsc, line 160: This assertion should be applied to all distros, not only Redhat.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+
+            if isinstance(node.os, Redhat):
```
squirrelsc: It looks there is no different to Redhat and other distros.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+                linux_path = "/mnt/linux"
+            else:
+                linux_path = "/usr/src/linux"
+            
+            # prepare swap space
+            # Get the latest device name which should be the new attached data disk
+            data_dev = node.execute(
```
squirrelsc, line 165: use the fdisk tool.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+                node.execute(
+                    f"echo {key} >> keys.txt"
+                )
+            cmd_res = node.execute(
+                f"cat keys.txt | fdisk {data_dev}"
+            )
+            log.debug(f"{cmd_res.exit_code}: Executed fdisk command")
+            # Need to wait for system complete the swap disk update.
+            log.info("Waiting 10 seconds for swap disk update")
+            time.sleep(10)
```
squirrelsc: Don't hardcode any sleep. Detect it by some condition.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+            cmd_res = node.execute(
+                "ls /dev/sd*", shell = True, sudo = True
+            )
+            log.info(f"{cmd_res.exit_code}: List out /dev/sd* - {cmd_res.stdout}")
+
+            # TODO: WHAT IF THE DISK IS MOUNTED OR BEING USED???
+            cmd_res = node.execute(
```
squirrelsc: Implement it in Swap, and use SwapOn tool too.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+            )
+            log.info(f"{cmd_res.exit_code}: Enabled the swap space")
+            time.sleep(2)
+
+            cmd_res = node.execute(
+                "swapon -s", shell=True, sudo=True
+            )
+            log.info(f"{cmd_res.exit_code}: Show the swap oninformation - {cmd_res.stdout}")
+
+            cmd_res = node.execute(
```
squirrelsc: Use blkid tool

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+            sw_uuid = cmd_res.stdout
+            log.info(f"{cmd_res.exit_code}: Found the Swap space disk UUID: {sw_uuid}")
+            if not sw_uuid:
+                raise SkippedException("Swap space disk UUID is empty. Abort the test.")
+
+            node.execute(
+                "chmod 766 /etc/fstab", shell = True, sudo = True
+            )
+
+            node.tools[Echo].write_to_file(
```
squirrelsc, line 198: Add comments to explain the steps.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+                f"{sw_uuid} none swap sw 0 0",
+                "/etc/fstab",
+                sudo = True,
+                append = True
+            )
+
+            if isinstance(node.os, Redhat) or isinstance(CentOs):
```
squirrelsc: CentOS is derived from Redhat, it doesn't need to check it.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+                _entry = node.tools[Cat].read_with_filter(
+                    "/etc/default/grub",
+                    "rootdelay=",
+                    sudo=True
+                )
+                if _entry:
+                    cmd_res = node.execute(
```
squirrelsc: use the sed tool

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+                        f"sed -i -e 's/rootdelay=300/rootdelay=300 resume={sw_uuid}/g' /etc/default/grub", shell=True, sudo=True
+                    )
+                    log.info(f"{cmd_res.exit_code}: Updated the /etc/default/grub with resume={sw_uuid}")
```
squirrelsc: If it's not key information in a test case, output it by debug level.

microsoft/testsuites/power/power.py
lesscodingmorehappiness: If I understand correctly, these are 2 approaches to get the kernel version? I just need to pick either way?

squirrelsc: One is the distro version like "7.9" of RedHat, another one is the kernel version like "3.10.0"

microsoft/testsuites/power/power.py
lesscodingmorehappiness: linux_path is used for git and compile kernel, removed this

microsoft/testsuites/power/power.py
lesscodingmorehappiness: yes Lili told me we don't need to compile kernel in v3

microsoft/testsuites/power/power.py
lesscodingmorehappiness: it's used in hb_setup script to build kernel (which could be excluded in v3) and determine vmlinuz file (which seems need to be kept?)

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+                    )
+                    log.info("Added GRUB_HIDDEN_TIMEOUT=30 in 50-cloudimg-settings.cfg file")
+                
+                # This is the case about GRUB_TIMEOUT
+                _entry = node.tools[Cat].read_with_filter(
+                    "/etc/default/grub.d/50-cloudimg-settings.cfg",
+                    '^GRUB_TIMEOUT=',
+                    sudo=True
+                )
+                if _entry:
```
squirrelsc, line 479: There are sed tool too. And this pattern is used many times. It's better to add one method in sed like `substitue_or_append` to simplify it.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,514 @@
...
+            ).stdout
+            keys = ('n', 'p', '1', '2048', '', 't', '82', 'p', 'w')
+            for key in keys:
+                node.execute(
+                    f"echo {key} >> keys.txt"
+                )
+            cmd_res = node.execute(
+                f"cat keys.txt | fdisk {data_dev}"
+            )
+            log.debug(f"{cmd_res.exit_code}: Executed fdisk command")
```
squirrelsc: This line is not needed, the lisa dumps the exit code for every command.

PR #1730, opened 2022-02-07
QEMU orchestrator: Respect the `keep_environment` flag.

cwize1:
If the `keep_environment` flag has been set, then don't
delete the VMs if the VM deployment fails.

### DIFF
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -103,7 +103,10 @@ class QemuPlatform(Platform):
                 self._fill_nodes_metadata(environment, log, qemu_conn)
 
             except Exception as ex:
-                self._delete_nodes(environment, log, qemu_conn)
+                assert environment.platform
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -103,7 +103,10 @@ def _deploy_nodes(self, environment: Environment, log: Logger) -> None:
                 self._fill_nodes_metadata(environment, log, qemu_conn)
 
             except Exception as ex:
-                self._delete_nodes(environment, log, qemu_conn)
+                assert environment.platform
+                if not environment.platform.runbook.keep_environment:
```
squirrelsc: It's not a bool value, when code reaches here. It has three values: always, no, failed. The constants are defined in constatns.py. ENVIRONMENT_KEEP_ALWAYS/NO/FAILED.

squirrelsc (approved):

squirrelsc (approved):

PR #1731, opened 2022-02-08
Fix test case issues.

LiliDeng:
Fix case verify_disks_device_timeout_setting

### DIFF
--- a/lisa/base_tools/wget.py
+++ b/lisa/base_tools/wget.py
@@ -34,15 +34,16 @@ class Wget(Tool):
         overwrite: bool = True,
         executable: bool = False,
         sudo: bool = False,
+        force_run: bool = False,
     ) -> str:
         is_valid_url(url)
 
-        # create folder when it doesn't exist
-        self.node.execute(f"mkdir -p {file_path}", shell=True, sudo=sudo)
         # combine download file path
         # TODO: support current lisa folder in pathlib.
         # So that here can use the corresponding path format.
         if file_path:
+            # create folder when it doesn't exist
+            self.node.execute(f"mkdir -p {file_path}", shell=True, sudo=sudo)
             download_path = f"{file_path}/{filename}"
         else:
             download_path = f"{self.node.working_path}/{filename}"
@@ -56,7 +57,9 @@ class Wget(Tool):
             command = f"{command} -O {download_path}"
         else:
             command = f"{command} -P {download_path}"
-        command_result = self.run(command, no_error_log=True, shell=True, sudo=sudo)
+        command_result = self.run(
+            command, no_error_log=True, shell=True, sudo=sudo, force_run=force_run
+        )
         matched_result = self.__pattern_path.match(command_result.stdout)
         if matched_result:
             download_file_path = matched_result.group("path")
@@ -78,7 +81,7 @@ class Wget(Tool):
 
     def verify_internet_access(self) -> bool:
         try:
-            result = self.get("https://www.azure.com")
+            result = self.get("https://www.azure.com", force_run=True)
             if result:
                 return True
         except Exception as e:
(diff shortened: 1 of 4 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -684,7 +684,7 @@ def get_all_disks(self) -> List[str]:
             match = disk_pattern.match(disk_raw.group())
             assert match, "fail to get disk label"
             disk_label = match.group("label")
-            disk_array.append(disk_label)
+            disk_array.append(disk_label.strip())
```
squirrelsc: Please update the raw_pattern like `/sys/block/(?P<label>sd*\w*)`. It can capture all of them without whitespace, and do it once.

lisa/base_tools/wget.py
```diff
@@ -56,7 +56,9 @@ def get(
             command = f"{command} -O {download_path}"
         else:
             command = f"{command} -P {download_path}"
-        command_result = self.run(command, no_error_log=True, shell=True, sudo=sudo)
+        command_result = self.run(
```
squirrelsc, line 60: Sometimes we don't need to download files again. So it should be able to skip. Is there a way to keep the force_run?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -676,7 +676,7 @@ def get_all_disks(self) -> List[str]:
         raw_pattern = re.compile(r"(/sys/block/sd*[\w\W]*?)(?=/sys/block/sd*|\Z)", re.M)
         # sda
         # sdb
-        disk_pattern = re.compile(r"/sys/block/(?P<label>sd.*)", re.M)
+        disk_pattern = re.compile(r"/sys/block/(?P<label>sd*\w*)", re.M)
```
squirrelsc: Sorry, fixed a typo: `/sys/block/(?P<label>sd\w*)`
And I think it doesn't need both of them now. One pattern is enough, which can parse from the raw string directly.

lisa/features/gpu.py
```diff
@@ -115,21 +115,43 @@ def _install_cuda_driver(self, version: str) -> None:
...
             self._node.execute(
                 "apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/"
                 f"cuda/repos/ubuntu{release}/x86_64/7fa2af80.pub",
                 sudo=True,
             )
-            # download and install the cuda driver package from the repo
-            self._node.os._install_package_from_url(
-                f"{cuda_repo}", package_name="cuda-drivers.deb", signed=False
-            )
+            if "2004" == release:
```
squirrelsc: Does it work with 2004 only? For this kind of pattern, we can assume the latest format work with future versions. So the logic should be like
```python
if "1804" == release:
   # logic of 1804
else:
   # logic of 2004
```

lisa/features/gpu.py
```diff
@@ -215,14 +245,13 @@ def get_gpu_count_with_lspci(self) -> int:
...
-        if "nvidia" in self.gpu_vendor:
-            result = self._node.execute("nvidia-smi")
-            if result.exit_code != 0 or (result.exit_code == 0 and result.stdout == ""):
-                raise LisaException(
-                    f"nvidia-smi command exited with exit_code {result.exit_code}"
-                )
-            for device_info in result.stdout.splitlines():
-                if any(device_info in device_name for device_name in self.gpu_devices):
-                    device_count += 1
+        result = self._node.execute("nvidia-smi")
```
squirrelsc: Sorry, haven't catch that. If the GPU is from AMD, there is no nvidia-smi. So the original logic check if nvidia is in vendor or not. What's the reason to remove it, and is it tested with AMD gpu?

squirrelsc (dismissed):

lisa/features/gpu.py
LiliDeng: let me double check, origin design doesn't work, since it doesn't install the driver, and self.gpu_vendor is empty, let me try to add it back.

LiliDeng: I updated the code, please double check.

squirrelsc (approved):

PR #1732, opened 2022-02-09
Migrate NET-SET-STATIC-MAC test case.

LiliDeng:


lisa/nic.py
```diff
@@ -250,6 +250,40 @@ def reset_nic_state(self, nic_name: str) -> None:
         if result.exit_code == 0:
             self.load_interface_info(nic_name)
 
+    def set_nic_state(self, nic_name: str, operation: str = "down") -> None:
+        result = self._node.execute(
+            f"/sbin/ip link set dev {nic_name} {operation}",
```
squirrelsc: There is an `ip` tool. please use that to down/up the nic.

lisa/nic.py
```diff
@@ -250,6 +250,40 @@ def reset_nic_state(self, nic_name: str) -> None:
...
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"Could not bring {operation} {nic_name} on node {self._node.name}"
+            ),
+        )
+        if result.exit_code == 0:
+            self.load_interface_info(nic_name)
+
+    def set_mac_address(self, nic_name: str, mac_address: str) -> None:
+        mac_address_pattern = re.compile(
```
squirrelsc: Move it to `ip`

lisa/nic.py
```diff
@@ -250,6 +250,40 @@ def reset_nic_state(self, nic_name: str) -> None:
...
+        result = self._node.execute(
+            f"/sbin/ip link set dev {nic_name} {operation}",
+            shell=True,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"Could not bring {operation} {nic_name} on node {self._node.name}"
+            ),
+        )
+        if result.exit_code == 0:
```
squirrelsc: It doesn't need the `if`. If it's not 0, it will fail on previous line.

lisa/nic.py
```diff
@@ -250,6 +250,40 @@ def reset_nic_state(self, nic_name: str) -> None:
...
+        result = self._node.execute(
+            f"/sbin/ip link set {nic_name} address {mac_address}",
+            shell=True,
+            sudo=True,
+        )
+        self.set_nic_state(nic_name, "up")
+        if result.exit_code == 0:
+            self.load_interface_info(nic_name)
+        else:
+            raise LisaException(
```
squirrelsc: This exception can be checked when call `execute`.

LiliDeng: I want to bring the nic back when previous execute failed.

squirrelsc: you can use the try...final... or try...except...

microsoft/testsuites/performance/storageperf.py
```diff
@@ -224,7 +224,7 @@ def _perf_nfs(
             and not isinstance(server_node.os, Debian)
             and not isinstance(server_node.os, SLES)
         ):
-            raise SkippedException(f"{server_node.os} not supported")
+            raise SkippedException(f"{server_node.os.name} not supported")
```
squirrelsc, line 227: Can we change it to FailedException? We may review it later, if it should be supported.

lisa/tools/ip.py
```diff
@@ -58,3 +60,21 @@ def set_mtu(self, nic_name: str, mtu: int) -> None:
...
+        self.node.execute(
+            f"/sbin/ip link set {nic_name} address {mac_address}",
+            shell=True,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"fail to set mac address {mac_address} for nic {nic_name}"
+            ),
+        )
+        self.up(nic_name)
```
squirrelsc: If you want to bring the nic back, when the previous line failed. Use try...final.. to run the `up`

lisa/nic.py
LiliDeng: updated

microsoft/testsuites/performance/storageperf.py
LiliDeng: let's update in another pr after dicussion.

lisa/tools/ip.py
LiliDeng: updated

lisa/tools/ip.py
```diff
@@ -58,3 +66,23 @@ def set_mtu(self, nic_name: str, mtu: int) -> None:
         self.run(f"link set dev {nic_name} mtu {mtu}", force_run=True, sudo=True)
         new_mtu = self.get_mtu(nic_name=nic_name)
         assert_that(new_mtu).described_as("set mtu failed").is_equal_to(mtu)
+
+    def set_mac_address(self, nic_name: str, mac_address: str) -> None:
+        mac_address_pattern = re.compile(
```
squirrelsc: add examples above, and move it to class level.

squirrelsc (dismissed):

squirrelsc (approved):

PR #1735, opened 2022-02-10
Log cleanup

cwize1:
1. Add a `local_node_connect` to make it easy to run commands
against the test runner machine

2. Add a `no_debug_log` option to the `Process.start` function and
all of its derivatives.  This will allow a caller to prevent output
from a process from being written to the DEBUG logs. This is useful
for commands with particularly spammy output. (For example,
`kubectl get -o yaml`.)

### DIFF
--- a/lisa/environment.py
+++ b/lisa/environment.py
@@ -169,6 +169,7 @@ class Environment(ContextMixin, InitializableMixin):
         super().__init__()
 
         self.nodes: Nodes = Nodes()
+        self._local_node: Optional[Node] = None
         self.runbook = runbook
         self.name = runbook.name
         self.is_predefined: bool = is_predefined
@@ -322,6 +323,21 @@ class Environment(ContextMixin, InitializableMixin):
         self.nodes.append(node)
 
         return node
+
+    def local_node(self) -> Node:
+        if self._local_node:
+            return self._local_node
+
+        node_runbook = schema.LocalNode(capability=schema.Capability())
+        self._local_node = Node.create(
+            index=-1,
+            runbook=node_runbook,
+            logger_name="local",
+            base_log_path=self.log_path,
+            parent_logger=self.log,
+        )
+        self._local_node.name = "local"
+        return self._local_node
 
     def get_information(self) -> Dict[str, str]:
         final_information: Dict[str, str] = {}
--- a/lisa/environment.py
+++ b/lisa/environment.py
@@ -169,7 +169,6 @@ class Environment(ContextMixin, InitializableMixin):
         super().__init__()
 
         self.nodes: Nodes = Nodes()
-        self._local_node: Optional[Node] = None
         self.runbook = runbook
         self.name = runbook.name
         self.is_predefined: bool = is_predefined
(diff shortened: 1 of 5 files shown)


lisa/util/process.py
```diff
@@ -235,3 +241,67 @@ def _filter_sudo_result(self, raw_input: str) -> str:
             raw_input = "".join(lines[1:])
             self._log.debug(f'found error message in sudo: "{lines[0]}"')
         return raw_input
+
+
+def local_execute(
```
squirrelsc: It can be done by use a local node. Refer to https://github.com/microsoft/lisa/blob/main/lisa/runners/legacy_runner.py#L81
If you fell the two lines are ugly, refer to below to create a helper method to return a local node. So it doesn't need to construct the mockup runbook everytime.
https://github.com/microsoft/lisa/blob/main/lisa/runners/legacy_runner.py#L81

cwize1: You posted the same link twice?

squirrelsc: Sorry, I mean this method,

https://github.com/microsoft/lisa/blob/a615a4649951708ed4c921aba25707662921987a/lisa/node.py#L529

cwize1: Out of interest, what advantage does using a `Node` give?

squirrelsc: It doesn't need to copy the code, and it leverages all benefits of lisa, for example, using node.tools to use any tool.

lisa/environment.py
```diff
@@ -323,6 +324,21 @@ def create_node_from_requirement(
 
         return node
 
+    def local_node(self) -> Node:
```
squirrelsc: can it be a method like quick_connect at the bottom? It doesn't relate to the environment. The cached local node can be put in the module level too.

cwize1: Attaching this LocalNode to the Environment object has a nice benefit of associating the log messages with a particular test. It's also nice for the caller to not need to think about things like `logger_name` and `parent_logger`.

squirrelsc: Not all environments need the local node, for example the Azure, Remote ready platforms. If it needs the associate log, it can be done in the qemu itself. Like the quick_connect, it accepts the parent_logger argument. So the logger can be associated with the environment. It doesn't need other places to associate the log. See the example,

In node.py
```python
def get_local_node(
    logger_name: str = "",
    index: int = -1,
    parent_logger: Optional[Logger] = None,
) -> Node:
   ...
```

In the qemu.py,
```python
get_local_node(parent_logger: environment.log)
```

cwize1: It is worth noting that I also need access to the local node in some tests I am writing (in a private repo). However, I ran into an issue where the name of the two local nodes was the same and therefore the second node instance failed to initialize complaining that the log directory already existed. Hence, why I wanted to place the local node in a shared location that both places could access.

In regards to AZURE and READY not requiring local node (at the moment), this is why I lazy initialized the local node object.

squirrelsc (approved):

PR #1739, opened 2022-02-10
DPDK-OVS test for Debian+Ubuntu

mcgov:
- initial OVS test implemented for Debian+Ubuntu
    - Redhat and SUSE can be supported eventually, not implemented yet.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkovs.py
+++ b/microsoft/testsuites/dpdk/dpdkovs.py
@@ -1,0 +1,274 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import re
+from typing import List, Type
+
+from assertpy import fail
+from semver import VersionInfo
+
+from lisa.executable import Tool
+from lisa.operating_system import Debian
+from lisa.tools import Gcc, Git, Ip, Make, Modprobe, Uname
+from lisa.util import UnsupportedDistroException
+from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
+
+
+class DpdkOvs(Tool):
+
+    ubuntu_packages = ["automake", "autoconf", "libtool", "libcap-ng-dev"]
+    _version_regex = re.compile(
+        r"v(?P<major>[0-9]+)\.(?P<minor>[0-9]+)\.(?P<patch>[0-9]+)"
+    )
+    OVS_BRIDGE_NAME = "br-dpdk"  # name for the bridge, can be anything
+
+    # constants for tracking setup state
+    INIT = 0
+    MODULE_LOAD = 1
+    SERVICE_START = 2
+    BRIDGE_ADD = 3
+    PORT_ADD = 4
+    INTERFACE_UP = 5
+
+    @property
+    def command(self) -> str:
+        return "ovs"
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        # dependencies are needed for build script! Don't delete
+        return [Git, Gcc, Make, Git]
+
+    # FIXME: Redhat, SUSE, and any distro supported by the DpdkTestpmd tool.
+    # should work here as well. They just need to be implemented and tested.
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdkovs.py
```diff
@@ -0,0 +1,251 @@
...
+        node.execute(
+            "ovs-vsctl get Open_vSwitch . dpdk_initialized",
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                "OVS repoted that DPDK EAL failed to initialize."
+            ),
+        )
+
+        self._teardown()
```
squirrelsc: use the `try...final...` to make sure the cleanup executed even there is an error above.

microsoft/testsuites/dpdk/dpdkovs.py
```diff
@@ -0,0 +1,251 @@
...
+            "",
+            update_envs=add_to_env,
+            cwd=self.repo_dir,
+        )
+        make.make_install(
+            update_envs=add_to_env,
+            cwd=self.repo_dir,
+        )
+
+    def test_ovs(self, device_address: str) -> None:
```
squirrelsc: The test logic should be in test suite folder. The tool should focus on executing logic of the tool itself. The test suite is the glue to use tools to validate.

microsoft/testsuites/dpdk/dpdkovs.py
mcgov: I've added teardown using a state variable, looking forward to feedback!

mcgov:
Adding additional reviewers for extra eyes, please nobody initiate the merge before @squirrelsc approves

microsoft/testsuites/dpdk/dpdkovs.py
```diff
@@ -0,0 +1,275 @@
...
+            )
+        else:
+            # iterate confirmed supported versions
+            for version_limit in ovs_dpdk_version_pairings:
+                ovs_upper_limit, dpdk_version_requirement = version_limit
+                if self.ovs_version < ovs_upper_limit:
+                    # located the correct range, check if versions match
+                    if dpdk_version.major == dpdk_version_requirement:
+                        break
+                    else:
```
squirrelsc: This level of `else` is not necessary. If the `if` meets, it breaks out, if not, it's always failed.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -90,6 +91,47 @@ def verify_dpdk_build_failsafe(
...
+        test_kit = initialize_node_resources(node, log, variables, "failsafe")
+
+        # checkout OpenVirtualSwitch
+        ovs = node.tools[DpdkOvs]
+
+        # provide ovs build with DPDK tool info and build
+        ovs.build_with_dpdk(test_kit.testpmd)
+
+        # enable hugepages needed for dpdk EAL
+        _init_hugepages(node)
```
squirrelsc, line 117: BTW, I implemented it within xdp too. Can you merge them? It needs to be removed after test case is done, or it may cause conflicts among test cases. It needs to consider which one (xdp, dpdk) depends on another one, to decide the place of the code. If it takes longer time, you can try it in next PR.

mcgov: I think this is a good idea but doing it correctly will be a bigger commit. I think we would want to implement a meminfo tool to get the initial memory state, then add hugepages functions as a second tool.

I'll create a work item for it, I think it's a bigger change than I would want to include here.

squirrelsc: the hugepages is not a command likely. We can add a virtual tool, but it may be confused. Because it just wraps our test data, not some common tool. So we can put it in the performance or some test suite folder like common, so it can be reused.

squirrelsc (approved):

PR #1741, opened 2022-02-11
KVM NTTTCP performance testcases

dsrivastavv:


lisa/tools/whoami.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+
+
+class Whoami(Tool):
```
squirrelsc: put new tool classes into the `__init__.py`, so it's easy to reference.

lisa/tools/ip.py
```diff
@@ -52,6 +53,17 @@ def addr_flush(self, nic_name: str) -> None:
             ),
         )
 
+    def addr_add(self, nic_name: str, ip: str) -> None:
```
squirrelsc: `add_ipv4_address` is clearer.

lisa/tools/ip.py
```diff
@@ -52,6 +53,17 @@ def addr_flush(self, nic_name: str) -> None:
...
+        self.run(
+            f"addr add {ip}/24 dev {nic_name}",
+            shell=True,
```
squirrelsc: It looks like no need the shell.

lisa/tools/ip.py
```diff
@@ -88,3 +100,59 @@ def set_mac_address(self, nic_name: str, mac_address: str) -> None:
             )
         finally:
             self.up(nic_name)
+
+    def is_nic_present(self, nic_name: str) -> bool:
```
squirrelsc: Let's make it consistent, use `nic_exists`.

lisa/tools/ip.py
```diff
@@ -88,3 +100,59 @@ def set_mac_address(self, nic_name: str, mac_address: str) -> None:
...
+
+        # create bridge
+        self.run(
+            f"link add {name} type bridge",
+            force_run=True,
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"Could not create bridge {name}",
+        )
+        self.run(
```
squirrelsc, line 134: You have the method implemented to add address, so use it.

lisa/tools/whoami.py
squirrelsc: Add a method like `get_name` to return structured result, so it doesn't need to use like `.stdout.strip()` everytime.

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -40,6 +42,18 @@
 class KVMPerformance(TestSuite):  # noqa
     TIME_OUT = 12000
 
+    CLIENT_IMAGE = "nestedclient.qcow2"
```
squirrelsc: Add underscore before them, it means they are not public.

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -78,9 +92,147 @@ def perf_nested_kvm_storage_singledisk(
...
+            stop_existing_vm=False,
+            nic_model="virtio-net-pci",
+            taps=[self.SERVER_TAP],
+            name="server",
+        )
+        server.tools[Ip].addr_add(self.NIC_NAME, self.SERVER_IP_ADDR)
+        server.tools[Ip].up(self.NIC_NAME)
+
+        # run ntttcp test
+        client_ntttcp = client.tools[Ntttcp]
```
squirrelsc: In my latest PR, I moved the method to the common in performance. Let's use the same method to get consistent results. Let me know, if something is not applicable, we can check how to improve it.

microsoft/testsuites/nested/common.py
```diff
@@ -54,10 +60,21 @@ def connect_nested_vm(
...
+        f"{image_folder_path}/{image_name}",
+        nics=nics,
+        nic_model=nic_model,
+        taps=taps,
+        disks=disks,
+        stop_existing_vm=stop_existing_vm,
+    )
+
+    # allow time for VM to boot up
+    time.sleep(90)
```
squirrelsc: Never add any hard code sleep, unless for interval break. It either waste time, or create random failures. For this case, use node.close() to close the connection. The node will try to connect next time, you can also call node.initialize to connect it explicitly.

lisa/tools/qemu.py
```diff
@@ -10,6 +11,14 @@
 from lisa.tools import Kill, Lsmod
 
 
+def generate_random_mac_address() -> str:
```
squirrelsc: Lili added a randmac lib, maybe it's useful.

lisa/tools/qemu.py
```diff
@@ -25,7 +34,13 @@ def create_vm(
         self,
         port: int,
         guest_image_path: str,
+        cores: int = 2,
+        memory: int = 2048,
+        nics: int = 1,
+        nic_model: str = "e1000",
```
squirrelsc, line 32: add comments above to explain the value.

lisa/tools/qemu.py
```diff
@@ -35,17 +50,26 @@ def create_vm(
...
-            "-device e1000,netdev=user.0 "
-            f"-netdev user,id=user.0,hostfwd=tcp::{port}-:22 "
-            "-enable-kvm -display none -daemonize "
-        )
+        cmd = f"-smp {cores} -m {memory} -hda {guest_image_path} "
+
+        # add devices
+        for i in range(nics):
+            random_mac_address = generate_random_mac_address()
+            cmd += f"-device {nic_model},netdev=net{i},mac={random_mac_address} "
```
squirrelsc, line 59: Add comments to explain the parameters.

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -78,9 +92,147 @@ def perf_nested_kvm_storage_singledisk(
...
+    ) -> None:
+        (
+            nested_image_username,
+            nested_image_password,
+            _,
+            nested_image_url,
+        ) = parse_nested_image_variables(variables)
+
+        # setup bridge and taps
+        node.tools[Ip].setup_bridge(self.BR_NAME, self.BR_ADDR)
```
squirrelsc: Those settings should be recovered, or it may fail next test cases.

lisa/tools/whoami.py
mcgov: I need this tool so I'm going to open a new PR for it. Could we merge it today and rebase this PR to the new main?

lisa/tools/qemu.py
```diff
@@ -56,8 +80,14 @@ def create_vm(
                     f"-device virtio-scsi-pci -device scsi-hd,drive=datadisk-{disk} "
                 )
 
-        # kill any existing qemu process
-        self.stop_vm()
+        # -enable-kvm: enable kvm
+        # -display: enable or disable display
+        # -demonize: run in background
```
LiliDeng: -daemonize

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -40,6 +42,18 @@
 class KVMPerformance(TestSuite):  # noqa
     TIME_OUT = 12000
 
+    CLIENT_IMAGE = "nestedclient.qcow2"
+    SERVER_IMAGE = "nestedserver.qcow2"
+    CLIENT_HOST_FWD_PORT = 60022
+    SERVER_HOST_FWD_PORT = 60023
```
LiliDeng, line 48: use the same port is okay, since client and server are different L1 vm.

lisa/tools/whoami.py
squirrelsc: Is it added?

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -78,9 +90,76 @@ def perf_nested_kvm_storage_singledisk(
...
+        )
+        client.tools[Ip].add_ipv4_address(self._NIC_NAME, self._CLIENT_IP_ADDR)
+        client.tools[Ip].up(self._NIC_NAME)
+        client.nics.default_nic = self._NIC_NAME
+
+        # run ntttcp test
+        perf_ntttcp(environment, server, client, test_case_name=inspect.stack()[1][3])
+
+        # clear bridge and taps
+        node.tools[Ip].delete_interface(self._BR_NAME)
```
squirrelsc: use try...final... to make sure they are removed, even some step failed.

lisa/tools/whoami.py
dsrivastavv: This tool was added by Matthew so I removed the commit from this PR

squirrelsc (dismissed):

lisa/util/shell.py
```diff
@@ -57,6 +58,40 @@ def wait_tcp_port_ready(
     return is_ready, result
 
 
+def wait_for_ssh_to_be_ready(
+    address: str,
+    port: int,
+    log: Optional[Logger] = None,
+    timeout: int = 300,
+    retry_interval: int = 10,
```
squirrelsc: give it a smaller value, like 1 or 2 seconds.

lisa/util/shell.py
```diff
@@ -57,6 +58,40 @@ def wait_tcp_port_ready(
...
+) -> bool:
+    is_ready = False
+    client = paramiko.SSHClient()
+    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
squirrelsc: use paramiko.MissingHostKeyPolicy(). The AutoAddPolicy saves the public key and IP pairs, but it may conflict, because the public IP may be assigned to another server.

lisa/util/shell.py
```diff
@@ -57,6 +58,40 @@ def wait_tcp_port_ready(
     return is_ready, result
 
 
+def wait_for_ssh_to_be_ready(
```
squirrelsc: It's similar usage like try_connect. Can you refine that method, to make it default wait for 5 seconds, but can be extend by the parameter? So, by default, it's a short timeout, and assume the service is ready. But it can be used in case the server is booting up, and ssh is unstable.

squirrelsc: I mean, can you just use the `try_connect`? If any logic needs to be merged into the `try_connect`, it's ok to go with.

dsrivastavv: I got confused because there were comments present on the ssh_ready function. Thanks for clarifying!

lisa/util/shell.py
```diff
@@ -139,25 +140,52 @@ def generate_run_command(
 
 # retry strategy is the same as spurplus.connect_with_retries.
 @retry(Exception, tries=3, delay=1, logger=None)  # type: ignore
```
squirrelsc: We can remove it now, because the timeout and retry is controlled inside.

lisa/util/shell.py
```diff
@@ -139,25 +140,52 @@ def generate_run_command(
...
-def try_connect(connection_info: ConnectionInfo) -> Any:
+def try_connect(
+    connection_info: ConnectionInfo, ssh_timeout: int = 300, tries: int = 3
```
squirrelsc: Let's use 10 seconds as default value, and extend it in your test case.

squirrelsc: It looks the tries doesn't need now. Please remove it.

dsrivastavv: tries is currently used for retrying `stdout.channel.recv_ready()`. From my understanding, this function seems to run a dummy command and check if we are getting the output. Do you think this is still required?

squirrelsc: I mean remove it from method arguments, but keep it internally. It looks it's not used. With this commit, this argument is confusing now. It's hard to explain the relationship between ssh_timeout and tries.

squirrelsc (approved):

PR #1742, opened 2022-02-11
QEMU: Add network boot timeout config value.

cwize1:
On some systems and/or some guest OS images, the OS may take
longer to boot to the point where it requests an IP address
from the libvirt DHCP server. Therefore, provide an option
to extend this timeout length.

### DIFF
--- a/lisa/sut_orchestrator/qemu/context.py
+++ b/lisa/sut_orchestrator/qemu/context.py
@@ -11,6 +11,9 @@ from .console_logger import QemuConsoleLogger
 class EnvironmentContext:
     ssh_public_key: str = ""
 
+    # Timeout for the OS to boot and acquire an IP address, in seconds.
+    network_boot_timeout: float = 30.0
+
 
 @dataclass
 class NodeContext:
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/qemu/context.py
```diff
@@ -11,6 +11,9 @@
 class EnvironmentContext:
     ssh_public_key: str = ""
 
+    # Timeout for the OS to boot and acquire an IP address, in seconds.
+    network_boot_timeout: float = 30.0
```
romoh, line 15: curious: why float?

romoh: nvm, I see why now https://docs.python.org/3/library/time.html#time.time

romoh (approved):

romoh:
Should this be generalized to LISA? Do we anticipate similar problems happening for other providers as well?

squirrelsc:
> Should this be generalized to LISA? Do we anticipate similar problems happening for other providers as well?

In Azure, there is an overall timeout, which is managed by Azure provisioning APIs. It will wait until the VM is ready. There is a client side DHCP timeout, but if it's timed out, the provisioning will be timed out also. It's not configurable.

squirrelsc (approved):

PR #1743, opened 2022-02-14
find_partition_with_freespace: refactor function to hoist to node level

mcgov:
@somil55 added a function find_partition_with_freespace to the nested virt test suite and it looks super useful. Let's hoist it!

This PR:

- Moves the check-for-freespace function to node level. 
- Renames and moves a dependency for that function. The renamed function `check_partition_size` becomes a public function of the Df tool, since it depends on that tool to perform it's check.

The end result is that the caller can use `node.find_partition_with_freespace(size_in_gb)`

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -427,6 +427,9 @@ class RemoteNode(Node):
         return self.get_pure_path(result.stdout)
 
     def find_partition_with_freespace(self, size_in_gb: int) -> str:
+        if self.os.is_windows:
+            raise NotImplementedError()
+
         df = self.tools[Df]
         home_partition = df.get_partition_by_mountpoint("/home")
         if home_partition and df.check_partition_size(home_partition, size_in_gb):
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -12,7 +12,7 @@ from lisa.executable import Tools
 from lisa.feature import Features
 from lisa.nic import Nics
 from lisa.operating_system import OperatingSystem
-from lisa.tools import Echo, Reboot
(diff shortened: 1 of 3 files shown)


lisa/node.py
```diff
@@ -426,6 +426,20 @@ def get_working_path(self) -> PurePath:
 
         return self.get_pure_path(result.stdout)
 
+    def find_partition_with_freespace(self, size_in_gb: int) -> str:
+        df = self.tools[Df]
+        home_partition = df.get_partition_by_mountpoint("/home")
+        if home_partition and df.check_partition_size(home_partition, size_in_gb):
```
dsrivastavv, line 435: Do we need this function at the node level? This seems very specific to Linux distros and might not be applicable to other platforms.

mcgov: Ah I didn't consider the portability issue. I do think it would be useful to have a helper function to get a disk with free space for Windows too, maybe we could switch based on node type? We don't need to implement Windows yet but it would leave the door open to do it later.

lisa/tools/df.py
```diff
@@ -54,3 +54,17 @@ def get_partition_by_mountpoint(
             if partition.mountpoint == mountpoint:
                 return partition
         return None
+
+    def check_partition_size(
+        self,
+        partition: PartitionInfo,
+        size: int,
```
squirrelsc: add comments to specify the size or name it like size_in_gb.

lisa/node.py
mcgov: Updated to check for node.os.is_windows

squirrelsc (approved):

PR #1746, opened 2022-02-15
modify shell.chown to accept string username or group id

mcgov:
The chown function we call to allows for a string username or groupid, however we had type-checked it to only allow integers. This PR modifies our type-checking definitions to allow for either a string or an int.

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -452,7 +452,7 @@ class SshShell(InitializableMixin):
         destination_str = self._purepath_to_str(destination)
         self._inner_shell.symlink(source_str, destination_str)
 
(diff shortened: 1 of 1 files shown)


mcgov:
testing now

mcgov:
There's an issue with it currently, will adjust and republish

mcgov:
This doesn't actually work even though the code looks like it should accept a string. I'm not sure why but there's a ton of error spew about trying to struct.pact an integer when the argument is a string. I removed the cast(str, uid) call and the result is the same. Not sure why that's broken but this function isn't used anywhere so maybe we can just wrap chown as a tool or something instead.

squirrelsc:
Let's use a chown tool to do that, and remove this interface from shell to avoid confusion.

PR #1747, opened 2022-02-15
Chown Tool: remove broken shell.chown and replace with node.tools[Chown]

mcgov:
shell.chown is broken, replace it with a tool for posix systems.

### DIFF
--- a/lisa/tools/chown.py
+++ b/lisa/tools/chown.py
@@ -1,0 +1,53 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import PurePath
+from typing import List
+
+from assertpy import assert_that
+
+from lisa.executable import Tool
+
+
+class Chown(Tool):
+    @property
+    def command(self) -> str:
+        return "chown"
+
+    def _check_exists(self) -> bool:
+        return self.node.is_posix
+
+    def change_owner(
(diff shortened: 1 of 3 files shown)


mcgov:
testing on ovs-rh branch...

mcgov:
Testing looks good, opening for review

mcgov:
fixed weird logic for user:group. You can use :group by itself and neither user nor group is a required argument according to the docs

lisa/tools/chown.py
```diff
@@ -0,0 +1,49 @@
...
+        if new_owner:
+            arguments.append(new_owner)
+
+        # add FILE
+        assert_that(str(file)).described_as(
+            "chown: filepath was empty and file is a required argument"
+        ).is_true()
+        arguments.append(f"{str(file)}")
+
+        assert_that(self.node.shell.exists(file)).described_as(
```
squirrelsc: No sure if it will fail, if the file need root permission to access. We may leave it to next command and check the exit code, return the error message.

lisa/tools/chown.py
```diff
@@ -0,0 +1,49 @@
...
+            f"Attempt to chown {str(file)} failed. File does not exist."
+        ).is_true()
+
+        # execute
+        self.run(parameters=" ".join(arguments), shell=True, sudo=True)
```
squirrelsc: assert the exit code.

squirrelsc (approved):

PR #1749, opened 2022-02-15
Add RedHat support to DpdkOvs test class

mcgov:
Add support for redhat  to OVS test

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -262,7 +262,12 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
 
     def find_partition_with_freespace(self, size_in_gb: int) -> str:
         if self.os.is_windows:
-            raise NotImplementedError()
+            raise NotImplementedError(
+                (
+                    "find_partition_with_freespace was called on a Windows node, "
+                    "this function is not implemented for Windows"
+                )
+            )
 
         df = self.tools[Df]
         home_partition = df.get_partition_by_mountpoint("/home")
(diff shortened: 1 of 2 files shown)


lisa/node.py
```diff
@@ -443,6 +443,23 @@ def get_working_path(self) -> PurePath:
 
         return self.get_pure_path(result.stdout)
 
+    def find_partition_with_freespace(self, size_in_gb: int) -> str:
```
squirrelsc: Did you remove it from the old place?

lisa/node.py
```diff
@@ -443,6 +443,23 @@ def get_working_path(self) -> PurePath:
...
+        if self.os.is_windows:
+            raise NotImplementedError()
```
squirrelsc: Add a description to make it readable. So the one line message can be helpful in the test results table. It doesn't need to check the log file to see the call trace.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -81,10 +81,10 @@ def command(self) -> str:
         "numactl-devel.x86_64",
         "librdmacm-devel",
         "pkgconfig",
-        "libmnl-devel",
+        # "libmnl-devel",
```
squirrelsc: just remove them, if they are not needed.

lisa/node.py
mcgov: Ah this is a bug. Artifact of a rebase, will fix

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: another testing artifact, removing them should happen in the testpmd cleanup commit.

mcgov:
Verified everything still works after the merge with the testpmd PR

mcgov:
ready for second round of review

squirrelsc (approved):

PR #1752, opened 2022-02-16
Add sriov item for some VM sizes

lubaihua33:
For Dp/Ep_v5 VM size, the `Accelerated Networking` is required. But the API returns `False`. Fix this issue temporarily and revert it till the bug fixed

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1487,6 +1487,18 @@ class AzurePlatform(Platform):
         if not node_space.network_interface.max_nic_count:
             node_space.network_interface.max_nic_count = 8
 
+        # For Dp/Ep_v5 VM size, the `Accelerated Networking` is required. But the API
+        # return `False`. Fix this issue temporarily and revert it till bug fixed
+        if resource_sku.family in [
+            "standardDPDSv5Family",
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1487,6 +1487,18 @@ def _resource_sku_to_capability(  # noqa: C901
         if not node_space.network_interface.max_nic_count:
             node_space.network_interface.max_nic_count = 8
 
+        # For Dp/Ep_v5 VM size, the `Accelerated Networking` is required. But the API
```
LiliDeng, line 1490: Dp/Ep_v5 VM sizes have `Accelerated Networking` capability.

One core size also supports?

squirrelsc: @lubaihua33 please double check this. I don't have other concerns.

lubaihua33: > Dp/Ep_v5 VM sizes have `Accelerated Networking` capability.
> 
> One core size also supports?

For Dp/Ep_V5 VM sizes, the smallest size has 2 cores. I tried one core in Hyper-V. One core also supports SRIOV.

squirrelsc (approved):

PR #1754, opened 2022-02-17
QEMU orchestrator: Implement CPU core count and memory requirements.

cwize1:


lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -34,6 +34,12 @@
 from .serial_console import SerialConsole
 
 
+class _HostCapabilities:
```
squirrelsc, line 37: Refer to other classes to add decorators of dataclass and dataclass_json. So it can be used in runbook to force run on specific settings.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -119,8 +172,64 @@ def _create_node_capabilities(
...
+                f"Host only has {host_capabilities.free_memory_kib} KiB free."
+            )
+            return False
+
+        return True
+
+    # Get the minimum value for a node requirement with an interger type.
+    # Note: Unlike other orchestrators, we don't want to fill up the capacity of
+    # the host in case the test is running on a dev box.
+    def _get_count_space_min(self, count_space: search_space.CountSpace) -> int:
```
squirrelsc, line 217: It can use `generate_min_capability_countspace`, and put in the count_space twice. It covers all scenarios.

lisa/sut_orchestrator/qemu/platform.py
cwize1: I don't follow. This class is just a helper class containing some relevant information about the test runner machine. It is being compared against the node requirements. I am not sure why one would want to specify host requirements in the runbook.

squirrelsc (approved):

PR #1755, opened 2022-02-17
Add new sgx test cases with TrustLaunch support

johnsongeorge-w:


squirrelsc (dismissed):

LiliDeng (dismissed):

LiliDeng (approved):
should we add comments for -
the case VALIDATE-INTEL-SGX-DRIVER-FOR-DCsv3-VM-TRUSTEDLAUNCH will fail for a known bug

squirrelsc (approved):

PR #1760, opened 2022-02-21
QEMU orchestrator: implement support for remote hosts

anirudhrb:
The current qemu orchestrator treats the local machine as the host for the VMs. Add support to allow specifying a remote
host in the runbook. The remote host should have libvirtd running and configured to allow remote connections. VMs under test are spawned on this remote host.

Example runbook for specifying a remote host
```YAML
name: qemu default
...<snip>...
platform:
  - type: qemu
    qemu:
      hosts:
        - address: "10.77.0.5"
          username: "anirudh"
    requirement:
      qemu:
        qcow2: $(qcow2)
        cloud_init:
```

In case `host` is not specified in the runbook, the local machine is treated as `host` (existing behavior).

### DIFF
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -8,7 +8,7 @@ import string
 import tempfile
 import time
 import xml.etree.ElementTree as ET  # noqa: N817
-from pathlib import PurePath
+from pathlib import Path
 from typing import Any, List, Optional, Tuple, Type
 
 import libvirt  # type: ignore
@@ -48,10 +48,6 @@ class QemuPlatform(Platform):
         SerialConsole,
     ]
 
-    qemu_platform_runbook: QemuPlatformSchema
-    host_node: RemoteNode
-    libvirt_conn_str: str
-
     def __init__(self, runbook: schema.Platform) -> None:
         super().__init__(runbook=runbook)
 
@@ -80,6 +76,8 @@ class QemuPlatform(Platform):
             )
 
             self.libvirt_conn_str = self._get_libvirt_conn_string()
+        else:
+            self.host_node = None
 
     def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
         self._configure_environment(environment, log)
@@ -89,6 +87,8 @@ class QemuPlatform(Platform):
 
     def _deploy_environment(self, environment: Environment, log: Logger) -> None:
         local_node = local_node_connect(parent_logger=log)
+        if self.host_node is None:
+            self.host_node = local_node
         self._deploy_nodes(environment, log, local_node)
 
     def _delete_environment(self, environment: Environment, log: Logger) -> None:
@@ -279,12 +279,10 @@ class QemuPlatform(Platform):
             if not os.path.exists(qemu_node_runbook.qcow2):
                 raise LisaException(f"file does not exist: {qemu_node_runbook.qcow2}")
 
-            vm_disks_dir = os.path.dirname(qemu_node_runbook.qcow2)
-            if self.qemu_platform_runbook.is_host_remote():
-                vm_disks_dir = os.path.join(
-                    "/home", self.qemu_platform_runbook.host.username, vm_name_prefix
-                )
-                self.host_node.shell.mkdir(vm_disks_dir)
+            vm_disks_dir = os.path.join(
+                self.qemu_platform_runbook.host.lisa_working_dir, vm_name_prefix
+            )
+            self.host_node.shell.mkdir(Path(vm_disks_dir))
 
             node = environment.create_node_from_requirement(node_space)
             node_context = get_node_context(node)
@@ -306,23 +304,14 @@ class QemuPlatform(Platform):
             node_context.cloud_init_file_path = os.path.join(
                 vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
             )
-            node_context.os_disk_base_file_path = qemu_node_runbook.qcow2
-            if self.qemu_platform_runbook.is_host_remote():
-                host = self.qemu_platform_runbook.host
-                new_os_disk_base_file_path = os.path.join(
-                    "/home",
-                    host.username,
-                    vm_name_prefix,
-                    os.path.basename(qemu_node_runbook.qcow2)
-                )
 
-                local_node.execute(
-                    f"scp -i {host.private_key_file} "
-                    f"{node_context.os_disk_base_file_path} "
-                    f"scp://{host.address}/{new_os_disk_base_file_path}"
-                )
-
-                node_context.os_disk_base_file_path = new_os_disk_base_file_path
+            node_context.os_disk_base_file_path = os.path.join(
+                vm_disks_dir, os.path.basename(qemu_node_runbook.qcow2)
+            )
+            self.host_node.shell.copy(
+                    Path(qemu_node_runbook.qcow2),
+                    Path(node_context.os_disk_base_file_path)
+            )
 
             node_context.os_disk_file_path = os.path.join(
                 vm_disks_dir, f"{node_context.vm_name}-os.qcow2"
@@ -404,20 +393,11 @@ class QemuPlatform(Platform):
             except Exception as ex:
                 log.warning(f"console log delete failed. {ex}")
 
-            if self.qemu_platform_runbook.is_host_remote():
+            try:
                 vm_dir = os.path.dirname(node_context.os_disk_file_path)
-                self.host_node.shell.remove(PurePath(vm_dir), True)
-            else:
-                # Delete the files created for the VM.
-                try:
-                    os.remove(node_context.os_disk_file_path)
-                except Exception as ex:
-                    log.warning(f"OS disk delete failed. {ex}")
-
-                try:
-                    os.remove(node_context.cloud_init_file_path)
-                except Exception as ex:
-                    log.warning(f"cloud-init ISO file delete failed. {ex}")
+                self.host_node.shell.remove(Path(vm_dir), True)
+            except Exception as ex:
+                log.warning("Working directory delete failed. {ex}")
 
     # Delete a VM.
     def _stop_and_delete_vm(
@@ -521,24 +501,19 @@ class QemuPlatform(Platform):
         meta_data_string = yaml.safe_dump(meta_data)
 
         iso_path = node_context.cloud_init_file_path
-        tmp_dir = None
-        if self.qemu_platform_runbook.is_host_remote():
-            tmp_dir = tempfile.TemporaryDirectory()
-            iso_path = os.path.join(tmp_dir.name, "cloud-init.iso")
+        tmp_dir = tempfile.TemporaryDirectory()
+        iso_path = os.path.join(tmp_dir.name, "cloud-init.iso")
 
         self._create_iso(
             iso_path,
             [("/user-data", user_data_string), ("/meta-data", meta_data_string)],
         )
 
-        if self.qemu_platform_runbook.is_host_remote():
-            host = self.qemu_platform_runbook.host
-            local_node.execute(
-                f"scp -i {host.private_key_file} "
-                f"{iso_path} "
-                f"scp://{host.address}/{node_context.cloud_init_file_path}"
-            )
-            tmp_dir.cleanup()
+        self.host_node.shell.copy(
+            Path(iso_path), Path(node_context.cloud_init_file_path)
+        )
+
+        tmp_dir.cleanup()
 
     # Create an ISO file.
     def _create_iso(self, file_path: str, files: List[Tuple[str, str]]) -> None:
@@ -562,21 +537,9 @@ class QemuPlatform(Platform):
         self, environment: Environment, log: Logger, node: Node, local_node: Node
     ) -> None:
         node_context = get_node_context(node)
-
-        if self.qemu_platform_runbook.is_host_remote():
-            self.host_node.tools[QemuImg].createDiffQcow2(
-                node_context.os_disk_file_path,
-                node_context.os_disk_base_file_path
-            )
-
-            return
-
-        # Create a new differencing image with the user provided image as the base.
-        local_node.execute(
-            f"qemu-img create -F qcow2 -f qcow2 -b "
-            f'"{node_context.os_disk_base_file_path}" '
-            f'"{node_context.os_disk_file_path}"',
-            expected_exit_code=0,
+        self.host_node.tools[QemuImg].createDiffQcow2(
+            node_context.os_disk_file_path,
+            node_context.os_disk_base_file_path
         )
 
     # Create the XML definition for the VM.
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -9,7 +9,7 @@ import tempfile
 import time
 import xml.etree.ElementTree as ET  # noqa: N817
 from pathlib import Path
-from typing import Any, List, Optional, Tuple, Type
+from typing import Any, List, Optional, Tuple, Type, cast
 
 import libvirt  # type: ignore
 import pycdlib  # type: ignore
@@ -20,7 +20,7 @@ from lisa.environment import Environment
 from lisa.feature import Feature
 from lisa.node import Node, RemoteNode, local_node_connect
 from lisa.platform_ import Platform
-from lisa.tools import QemuImg, Iptables
+from lisa.tools import Iptables, QemuImg
 from lisa.util import LisaException, constants, get_public_key_data
 from lisa.util.logger import Logger
 
@@ -50,6 +50,9 @@ class QemuPlatform(Platform):
 
     def __init__(self, runbook: schema.Platform) -> None:
         super().__init__(runbook=runbook)
+        self.libvirt_conn_str: str
+        self.qemu_platform_runbook: QemuPlatformSchema
+        self.host_node: Node
 
     @classmethod
     def type_name(cls) -> str:
@@ -65,21 +68,29 @@ class QemuPlatform(Platform):
         self.qemu_platform_runbook = self.runbook.get_extended_runbook(
             QemuPlatformSchema, type_name=QEMU
         )
-        self.libvirt_conn_str = "qemu:///system"
-
-        if self.qemu_platform_runbook.is_host_remote():
-            self.host_node = RemoteNode(schema.Node(name="qemu-host"), 0, "qemu-host")
-            self.host_node.set_connection_info(
-                address=self.qemu_platform_runbook.host.address,
-                username=self.qemu_platform_runbook.host.username,
-                private_key_file=self.qemu_platform_runbook.host.private_key_file
-            )
-
-            self.libvirt_conn_str = self._get_libvirt_conn_string()
-        else:
-            self.host_node = None
 
     def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        host = self.qemu_platform_runbook.host
+        if host.is_remote():
+            assert host.address
+            assert host.username
+            assert host.private_key_file
+
+            self.host_node = RemoteNode(
+                runbook=schema.Node(name="qemu-host"),
+                index=0,
+                logger_name="qemu-host",
+                parent_logger=log,
+            )
+            self.host_node.set_connection_info(
+                address=host.address,
+                username=host.username,
+                private_key_file=host.private_key_file,
+            )
+        else:
+            self.host_node = local_node_connect(parent_logger=log)
+
+        self._init_libvirt_conn_string()
         self._configure_environment(environment, log)
 
         with libvirt.open(self.libvirt_conn_str) as qemu_conn:
@@ -260,7 +271,7 @@ class QemuPlatform(Platform):
     # to be created. This makes it easier to cleanup everything after the test is
     # finished (or fails).
     def _configure_nodes(
-            self, environment: Environment, log: Logger, local_node: Node
+        self, environment: Environment, log: Logger, local_node: Node
     ) -> None:
         # Generate a random name for the VMs.
         test_suffix = "".join(random.choice(string.ascii_uppercase) for _ in range(5))
@@ -309,8 +320,8 @@ class QemuPlatform(Platform):
                 vm_disks_dir, os.path.basename(qemu_node_runbook.qcow2)
             )
             self.host_node.shell.copy(
-                    Path(qemu_node_runbook.qcow2),
-                    Path(node_context.os_disk_base_file_path)
+                Path(qemu_node_runbook.qcow2),
+                Path(node_context.os_disk_base_file_path),
             )
 
             node_context.os_disk_file_path = os.path.join(
@@ -318,7 +329,7 @@ class QemuPlatform(Platform):
             )
             node_context.console_log_file_path = os.path.join(
                 os.path.dirname(qemu_node_runbook.qcow2),
-                f"{node_context.vm_name}-console.log"
+                f"{node_context.vm_name}-console.log",
             )
             node_context.console_logger = QemuConsoleLogger()
 
@@ -397,7 +408,7 @@ class QemuPlatform(Platform):
                 vm_dir = os.path.dirname(node_context.os_disk_file_path)
                 self.host_node.shell.remove(Path(vm_dir), True)
             except Exception as ex:
-                log.warning("Working directory delete failed. {ex}")
+                log.warning(f"Working directory delete failed. {ex}")
 
     # Delete a VM.
     def _stop_and_delete_vm(
@@ -451,17 +462,19 @@ class QemuPlatform(Platform):
                 environment, log, qemu_conn, node, timeout
             )
 
-            node_addr = address
             node_port = 22
-            if self.qemu_platform_runbook.is_host_remote():
+            if self.host_node.is_remote:
                 self.host_node.tools[Iptables].start_forwarding(10022, address, 22)
 
-                node_addr = self.qemu_platform_runbook.host.address
+                remote_node = cast(RemoteNode, self.host_node)
+                conn_info = remote_node.connection_info
+
+                address = conn_info[constants.ENVIRONMENTS_NODES_REMOTE_ADDRESS]
                 node_port = 10022
 
             # Set SSH connection info for the node.
             node.set_connection_info(
-                address=node_addr,
+                address=address,
                 port=node_port,
                 public_port=node_port,
                 username=self.runbook.admin_username,
@@ -537,9 +550,8 @@ class QemuPlatform(Platform):
         self, environment: Environment, log: Logger, node: Node, local_node: Node
     ) -> None:
         node_context = get_node_context(node)
-        self.host_node.tools[QemuImg].createDiffQcow2(
-            node_context.os_disk_file_path,
-            node_context.os_disk_base_file_path
+        self.host_node.tools[QemuImg].create_diff_qcow2(
+            node_context.os_disk_file_path, node_context.os_disk_base_file_path
         )
 
     # Create the XML definition for the VM.
@@ -705,11 +717,11 @@ class QemuPlatform(Platform):
         assert isinstance(addr, str)
         return addr
 
-    def _get_libvirt_conn_string(self) -> str:
+    def _init_libvirt_conn_string(self) -> None:
         hypervisor = "qemu"
         host = self.qemu_platform_runbook.host
 
-        host_addr = host.address if host is not None else ""
-        transport = "+tcp" if host is not None else ""
+        host_addr = host.address if host.address else ""
+        transport = "+tcp" if host.address else ""
 
-        return f"{hypervisor}{transport}://{host_addr}/system"
+        self.libvirt_conn_str = f"{hypervisor}{transport}://{host_addr}/system"
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -5,8 +5,10 @@ import io
 import os
 import random
 import string
+import tempfile
 import time
 import xml.etree.ElementTree as ET  # noqa: N817
+from pathlib import PurePath
 from typing import Any, List, Optional, Tuple, Type
 
 import libvirt  # type: ignore
@@ -18,6 +20,7 @@ from lisa.environment import Environment
 from lisa.feature import Feature
 from lisa.node import Node, RemoteNode, local_node_connect
 from lisa.platform_ import Platform
+from lisa.tools import QemuImg, Iptables
 from lisa.util import LisaException, constants, get_public_key_data
 from lisa.util.logger import Logger
 
@@ -45,6 +48,10 @@ class QemuPlatform(Platform):
         SerialConsole,
     ]
 
+    qemu_platform_runbook: QemuPlatformSchema
+    host_node: RemoteNode
+    libvirt_conn_str: str
+
     def __init__(self, runbook: schema.Platform) -> None:
         super().__init__(runbook=runbook)
 
@@ -59,10 +66,25 @@ class QemuPlatform(Platform):
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         libvirt_events_thread.init()
 
+        self.qemu_platform_runbook = self.runbook.get_extended_runbook(
+            QemuPlatformSchema, type_name=QEMU
+        )
+        self.libvirt_conn_str = "qemu:///system"
+
+        if self.qemu_platform_runbook.is_host_remote():
+            self.host_node = RemoteNode(schema.Node(name="qemu-host"), 0, "qemu-host")
+            self.host_node.set_connection_info(
+                address=self.qemu_platform_runbook.host.address,
+                username=self.qemu_platform_runbook.host.username,
+                private_key_file=self.qemu_platform_runbook.host.private_key_file
+            )
+
+            self.libvirt_conn_str = self._get_libvirt_conn_string()
+
     def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
         self._configure_environment(environment, log)
 
-        with libvirt.open("qemu:///system") as qemu_conn:
+        with libvirt.open(self.libvirt_conn_str) as qemu_conn:
             return self._configure_node_capabilities(environment, log, qemu_conn)
 
     def _deploy_environment(self, environment: Environment, log: Logger) -> None:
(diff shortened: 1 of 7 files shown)


squirrelsc:
@cwize1 Can you have a look on this change too?

lisa/tools/firewall.py
```diff
@@ -60,6 +60,28 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return False
 
+    def start_forwarding(self, port: int, dst_ip: str, dst_port: str) -> None:
+        self.run(
```
squirrelsc, line 64: Does it need to check the exit code?

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -14,16 +14,30 @@ class CloudInitSchema:
...
+    # specified host by connecting remotely to the libvirt instance running on it.
+    # If None, the local machine is used as host.
+    host: Optional[LibvirtHost] = None
+
     # The timeout length for how long to wait for the OS to boot and request an IP
     # address from the libvirt DHCP server.
     # Specified in seconds. Default: 30s.
     network_boot_timeout: Optional[float] = None
 
+    def is_host_remote(self) -> bool:
```
squirrelsc: How about `is_local_only`?

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -14,16 +14,30 @@ class CloudInitSchema:
...
+    private_key_file: str = ""
 
 # QEMU orchestrator's global configuration options.
 @dataclass_json()
 @dataclass
 class QemuPlatformSchema:
+    # An optional remote host for the VMs. All test VMs will be spawned on the
+    # specified host by connecting remotely to the libvirt instance running on it.
+    # If None, the local machine is used as host.
+    host: Optional[LibvirtHost] = None
```
squirrelsc: `hosts`, and it should support multiple hosts as a list.

lisa/sut_orchestrator/qemu/schema.py
squirrelsc: Another way to make it simpler, to add `__post_init__`, and covert the local as an item in the host. So, the all logic can be consistent in the platform.

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: What's the use-case for multiple hosts?

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: Any reason to prefer the inverted form?

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -45,6 +48,10 @@ class QemuPlatform(Platform):
         SerialConsole,
     ]
 
+    qemu_platform_runbook: QemuPlatformSchema
```
cwize1: This is creating static variables on the class. If you want to pre-declare member variables, it should be done in the `__init__` function.

lisa/sut_orchestrator/qemu/schema.py
squirrelsc: It makes the platform scalable for testing. In our experience with Hyper-V platform in LISAv2, there are cases. 1) it needs more than one host to test faster, 2) some scenarios need to check live migrations, move vm from one machine to another one, and some tests need to test the traffic across hosts.

lisa/sut_orchestrator/qemu/schema.py
squirrelsc: When it supports multiple hosts (see below comments), it doesn't prevent to combine remote and local in the list together.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -490,6 +563,14 @@ def _create_node_os_disk(
     ) -> None:
         node_context = get_node_context(node)
 
+        if self.qemu_platform_runbook.is_host_remote():
```
cwize1: You should set the `host_node` to the same value as `local_node` when running on the local machine. That way, this function can use identical code for the local and remote cases.

squirrelsc: Once the node is created, it doesn't need to differentiate remote or local node. Use the `Node` type to get consistent behavior. The only difference is the remote node need to `set_connection_info`.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -423,16 +471,26 @@ def _fill_nodes_metadata(
                 environment, log, qemu_conn, node, timeout
             )
 
+            node_addr = address
+            node_port = 22
+            if self.qemu_platform_runbook.is_host_remote():
+                self.host_node.tools[Iptables].start_forwarding(10022, address, 22)
```
cwize1, line 476: This doesn't seem like it'll support testing multiple VMs on the same host.

What I recommend you do is allow the user to set the libvirt network that the VM connects to. Then the user can setup their own network bridge on the host, create a libvirt "network" that points to that bridge, and then set tell LISA to use that "network" for the VMs. Then the test runner will be able to access the VMs directly without needing to deal with port forwarding.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -259,6 +280,11 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
                 raise LisaException(f"file does not exist: {qemu_node_runbook.qcow2}")
 
             vm_disks_dir = os.path.dirname(qemu_node_runbook.qcow2)
+            if self.qemu_platform_runbook.is_host_remote():
+                vm_disks_dir = os.path.join(
+                    "/home", self.qemu_platform_runbook.host.username, vm_name_prefix
```
cwize1: In general it is a bad idea to place VM resources in a user's home directory, as it can cause permissions issues. Perhaps this directory should be a runbook parameter?

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -281,11 +307,29 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
...
+                host = self.qemu_platform_runbook.host
+                new_os_disk_base_file_path = os.path.join(
+                    "/home",
+                    host.username,
+                    vm_name_prefix,
+                    os.path.basename(qemu_node_runbook.qcow2)
+                )
+
+                local_node.execute(
+                    f"scp -i {host.private_key_file} "
```
cwize1: You can use `host_node.shell.copy` instead of using `scp`.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -660,3 +741,12 @@ def _try_get_node_ip_address(
         addr = addrs[0]["addr"]
         assert isinstance(addr, str)
         return addr
+
+    def _get_libvirt_conn_string(self) -> str:
+        hypervisor = "qemu"
+        host = self.qemu_platform_runbook.host
+
+        host_addr = host.address if host is not None else ""
+        transport = "+tcp" if host is not None else ""
```
cwize1: It looks like the `tcp` protocol is unencrypted. Is it possible to use `ssh` instead?

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: It's great to make it configurable. In the mean time, it's useful, if there is a default value. So the platform is easier to use.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -281,11 +307,29 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
                 vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
             )
             node_context.os_disk_base_file_path = qemu_node_runbook.qcow2
+            if self.qemu_platform_runbook.is_host_remote():
+                host = self.qemu_platform_runbook.host
+                new_os_disk_base_file_path = os.path.join(
```
cwize1, line 312: For the most part, LISA seems to use the `Path` class instead of strings and `os.path,join`.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -281,11 +307,29 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
...
+                    "/home",
+                    host.username,
+                    vm_name_prefix,
+                    os.path.basename(qemu_node_runbook.qcow2)
+                )
+
+                local_node.execute(
```
cwize1: There shouldn't be anything in  `_configure_nodes` that will require cleanup, such as creating files. Just make a note of where everything will be placed and then do the actual work in `_create_nodes`.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -360,16 +404,20 @@ def _delete_nodes(
             except Exception as ex:
                 log.warning(f"console log delete failed. {ex}")
 
-            # Delete the files created for the VM.
-            try:
-                os.remove(node_context.os_disk_file_path)
-            except Exception as ex:
-                log.warning(f"OS disk delete failed. {ex}")
+            if self.qemu_platform_runbook.is_host_remote():
+                vm_dir = os.path.dirname(node_context.os_disk_file_path)
```
cwize1: We should avoid having different logic for local and remote hosts. Either both local and remote hosts should create sub-directories for each VM or neither should.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -360,16 +404,20 @@ def _delete_nodes(
...
+                self.host_node.shell.remove(PurePath(vm_dir), True)
+            else:
+                # Delete the files created for the VM.
+                try:
+                    os.remove(node_context.os_disk_file_path)
```
cwize1: If `host_node` is set to `local_host` for the local case, then this can become `host_node.shell.remove`.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -360,16 +404,20 @@ def _delete_nodes(
...
                 log.warning(f"console log delete failed. {ex}")
 
-            # Delete the files created for the VM.
-            try:
-                os.remove(node_context.os_disk_file_path)
-            except Exception as ex:
-                log.warning(f"OS disk delete failed. {ex}")
+            if self.qemu_platform_runbook.is_host_remote():
+                vm_dir = os.path.dirname(node_context.os_disk_file_path)
+                self.host_node.shell.remove(PurePath(vm_dir), True)
```
cwize1: All exceptions should be caught and ignored for each operation during cleanup. That is, the cleanup operation should do its best to remove as much as possible.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -462,11 +520,26 @@ def _create_node_cloud_init_iso(
...
         )
 
+        if self.qemu_platform_runbook.is_host_remote():
+            host = self.qemu_platform_runbook.host
+            local_node.execute(
+                f"scp -i {host.private_key_file} "
+                f"{iso_path} "
+                f"scp://{host.address}/{node_context.cloud_init_file_path}"
+            )
+            tmp_dir.cleanup()
```
cwize1, line 539: The `tmp_dir.cleanup()` should be run in a `finally` block, to ensure it is always removed even if there are unexpected exceptions.

For example,
```python3
try:
    tmp_dir = None
    if self.qemu_platform_runbook.is_host_remote():
        tmp_dir = tempfile.TemporaryDirectory()
        ...
finally:
    if tmp_dir:
        tmp_dir.cleanup()

```

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -660,3 +741,12 @@ def _try_get_node_ip_address(
         addr = addrs[0]["addr"]
         assert isinstance(addr, str)
         return addr
+
+    def _get_libvirt_conn_string(self) -> str:
+        hypervisor = "qemu"
+        host = self.qemu_platform_runbook.host
+
+        host_addr = host.address if host is not None else ""
```
cwize1: You can simplify `if host is not None` to `if host`.

lisa/sut_orchestrator/qemu/platform.py
cwize1: `_delete_nodes` shouldn't be generating any values. All the values required by `_delete_nodes` should be generated in `_configure_nodes` and then stored.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -59,29 +66,41 @@ def supported_features(cls) -> List[Type[Feature]]:
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         libvirt_events_thread.init()
 
+        self.qemu_platform_runbook = self.runbook.get_extended_runbook(
+            QemuPlatformSchema, type_name=QEMU
+        )
+        self.libvirt_conn_str = "qemu:///system"
+
+        if self.qemu_platform_runbook.is_host_remote():
+            self.host_node = RemoteNode(schema.Node(name="qemu-host"), 0, "qemu-host")
```
cwize1: You are welcome to also set the `local_node` as an member variable if it would help simplify the code.

lisa/sut_orchestrator/qemu/platform.py
cwize1: For example, you could create a new node variable called `os_disk_source_file_path` and only set that variable if the file needs to be copied to the host. Then `_create_nodes` can copy the file if that variable is not `None`.

cwize1 (changes requested):
.

lisa/sut_orchestrator/qemu/platform.py
cwize1: `/srv/` (or `/var/`) might make for a better default.

lisa/sut_orchestrator/qemu/schema.py
cwize1: This might be a lot of work to support. So, feel free to place a restriction that `hosts` can only contain `1` item for now. Then in the future, we can do the work to add support for multiple hosts.

squirrelsc: > This might be a lot of work to support. So, feel free to place a restriction that `hosts` can only contain `1` item for now. Then in the future, we can do the work to add support for multiple hosts.

It sounds a plan, let's make the schema as final version firstly.

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: Made it configurable in the runbook and using `/var/tmp` as default.

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: I will investigate it. `tcp` seemed the simplest to use so I just went ahead with it to get things working quickly...

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -59,29 +62,45 @@ def supported_features(cls) -> List[Type[Feature]]:
...
```
cwize1: RemoteNode has a `parent_logger` parameter that you should be setting, You probably need to move this code into `_prepare_environment`/`_configure_environment`.

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -14,16 +14,35 @@ class CloudInitSchema:
...
+@dataclass
+class LibvirtHost:
+    address: Optional[str] = None
+    username: Optional[str] = None
+    private_key_file: Optional[str] = None
+
+    # The directory where lisa will store VM related files (such as disk images).
+    # This directory must already exist and the test user should have write permission
+    # to it.
+    lisa_working_dir: Optional[str] = "/var/tmp"
```
cwize1: Is this value actually optional? Same thing for other variables.

lisa/sut_orchestrator/qemu/platform.py
cwize1: You should still pre-declare these variables in the `__init__` function, so that type checking can work correctly. For example:
```python3
        super().__init__(runbook=runbook)
        self.host_node: Node
```

lisa/sut_orchestrator/qemu/platform.py
cwize1: Actually, it looks `local_node` isn't even being used anymore.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -280,12 +304,21 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
             node_context.cloud_init_file_path = os.path.join(
                 vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
             )
-            node_context.os_disk_base_file_path = qemu_node_runbook.qcow2
+
+            node_context.os_disk_base_file_path = os.path.join(
+                vm_disks_dir, os.path.basename(qemu_node_runbook.qcow2)
+            )
+            self.host_node.shell.copy(
```
cwize1: The OS image can potentially be multiple gigabytes in size. So, we should avoid copying it in the local host case, to improve performance.

cwize1 (changes requested):
.

squirrelsc:
BTW, please update document for the new schema.

lisa/sut_orchestrator/qemu/platform.py
pupacha: Also, please assign `local_node` to `self.host_node` here itself in the else part.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -65,21 +68,29 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
             QemuPlatformSchema, type_name=QEMU
         )
-        self.libvirt_conn_str = "qemu:///system"
 
-        if self.qemu_platform_runbook.is_host_remote():
-            self.host_node = RemoteNode(schema.Node(name="qemu-host"), 0, "qemu-host")
+    def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        host = self.qemu_platform_runbook.host
+        if host.is_remote():
+            assert host.address
```
cwize1, line 81: This is external data. So, you should throw a proper exception (`LisaException`) with a proper error message.

(FYI: `assert` is just a debug and type system declaration. It is valid for the compiler to remove it, if it wants to.)

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -33,16 +38,13 @@ class QemuPlatformSchema:
     # An optional remote host for the VMs. All test VMs will be spawned on the
     # specified host by connecting remotely to the libvirt instance running on it.
     # If None, the local machine is used as host.
-    host: Optional[LibvirtHost] = LibvirtHost()
+    host: LibvirtHost = LibvirtHost()
```
cwize1: This must remain `Optional` for back-compat.

lisa/util/shell.py
```diff
@@ -584,7 +584,10 @@ def remove(self, path: PurePath, recursive: bool = False) -> None:
                        (will fail if that's the case and this flag is off)
         """
         assert isinstance(path, Path), f"actual: {type(path)}"
-        path.rmdir()
+        if path.is_dir():
```
cwize1, line 587: While you are at it, also properly handle the `recursive` parameter?

lisa/tools/firewall.py
```diff
@@ -60,27 +60,31 @@ def command(self) -> str:
...
         )
+        result.assert_exit_code()
 
-    def stop_forwarding(self, port: int, dst_ip: str, dst_port: str) -> None:
-        self.run(
+    def stop_forwarding(self, port: int, dst_ip: str, dst_port: int) -> None:
+        result = self.run(
             f"-D FORWARD -o virbr0 -p tcp -d {dst_ip} --dport {dst_port} -j ACCEPT",
-            sudo=True
+            sudo=True,
```
cwize1, line 79: Alternatively, you can add `expected_error_code=0` to the `run` call instead of using `assert_exit_code()`.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -65,21 +68,29 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+            assert host.username
+            assert host.private_key_file
+
+            self.host_node = RemoteNode(
+                runbook=schema.Node(name="qemu-host"),
+                index=0,
```
cwize1: You should probably set the `index` to `-1` to avoid confusion with the actual test nodes.

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: Default value should ensure backwards compat?

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -360,16 +404,11 @@ def _delete_nodes(
...
+                vm_dir = os.path.dirname(node_context.os_disk_file_path)
+                self.host_node.shell.remove(Path(vm_dir), True)
             except Exception as ex:
-                log.warning(f"OS disk delete failed. {ex}")
-
-            try:
-                os.remove(node_context.cloud_init_file_path)
-            except Exception as ex:
-                log.warning(f"cloud-init ISO file delete failed. {ex}")
+                log.warning(f"Working directory delete failed. {ex}")
```
cwize1: The term "Working directory" is a little too generic. Maybe "VM files directory"?

cwize1 (changes requested):
.

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: Moved it to host now...

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: Added schema support for multiple hosts.

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: Updated.

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: It is backward compatible.

lisa/util/shell.py
anirudhrb: Done.

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: This doesn't _generate_ any values. It is simply _extracting_ the path to the parent folder from the values already stored in `node_context`.

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: Confusion where? In the logs? Isn't the name enough?

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: Done

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -280,12 +317,21 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
             node_context.cloud_init_file_path = os.path.join(
                 vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
             )
-            node_context.os_disk_base_file_path = qemu_node_runbook.qcow2
+
+            if self.host_node.is_remote:
```
squirrelsc, line 321: 1. the `is_remote` is incorrect here, and always be True, because it's the method itself. The `is_remote()` is right.
2. I'm not sure if the "remote" check is useful here. Can remote and local use the same folder structure?

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: We can use the consistent way for local and remote hosts. It will simplify the code logic. The only difference is the localhost doesn't need connection info.

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: 1. `is_remote` is correct based on usages in other files. `is_remote()` throws the error "bool is not callable".
2. This is an optimization. We avoid copying OS disk image for local node. That's why remote check is useful here.

lisa/sut_orchestrator/qemu/platform.py
cwize1: What do you mean by "the consistent way"?

anirudhrb: In case of port forwarding, we do have special logic for remote hosts (set up iptable rules for forwarding). So consistent way is not possible.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -258,7 +292,10 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
             if not os.path.exists(qemu_node_runbook.qcow2):
                 raise LisaException(f"file does not exist: {qemu_node_runbook.qcow2}")
 
-            vm_disks_dir = os.path.dirname(qemu_node_runbook.qcow2)
+            vm_disks_dir = os.path.join(
+                self.qemu_platform_runbook.hosts[0].lisa_working_dir, vm_name_prefix
+            )
+            self.host_node.shell.mkdir(Path(vm_disks_dir))
```
cwize1: The line should be moved out of `_configure_nodes` as well.

lisa/sut_orchestrator/qemu/platform.py
cwize1: `is_remote` is marked with `@property`. So, it behaves like a C# style property.

lisa/sut_orchestrator/qemu/platform.py
cwize1: At least in the other code I've seen, nodes that aren't part of the test environment get an index of `-1`.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -59,29 +65,59 @@ def supported_features(cls) -> List[Type[Feature]]:
...
     def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        if len(self.qemu_platform_runbook.hosts) > 1:
+            log.warning(
+                "Multiple hosts are currently not supported. "
+                "Only the first host will be used."
+            )
+
+        host = self.qemu_platform_runbook.hosts[0]
+        if host.is_remote():
+            assert host.address
```
cwize1, line 81: Why is this one still using `assert`?

anirudhrb: `host.is_remote()` already ensures `host.address is not None`. This is just to make the type checker happy.

lisa/tools/qemu_img.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class QemuImg(Tool):
+    @property
+    def command(self) -> str:
+        return "qemu-img"
+
+    def create_diff_qcow2(self, output_img_path: str, backing_img_path: str) -> None:
+        params = f"create -F qcow2 -f qcow2 -b {backing_img_path} {output_img_path}"
+        self.run(params, True)
```
cwize1: Set `expected_exit_code=0`

lisa/sut_orchestrator/qemu/platform.py
cwize1: Ah, because this is what `is_remote` is checking.

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: Sorry, I'm confused on the LibVirtHost, I thought its type is LibVirtHost. BTW, what's the reason it needs a LibVirtHost schema? Can it reuse the RemoteNode schema?

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: Anyway, I think it can make the logic simpler, if the port forwarding or other approaches are used on local too.
If to use port forwarding, one way is to set a start port range from high end like 30000, and map the ports one by one incrementally, if a port is unused.

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: This `is_remote` is used only once, and the logic is simple as one line. It doesn't worth to use a method.

lisa/sut_orchestrator/qemu/platform.py
cwize1: We probably shouldn't use port forward on local, as people can run these tests directly on their dev box and we don't want to be opening random inbound ports if it isn't necessary.

lisa/sut_orchestrator/qemu/platform.py
cwize1: Please at least try this, as we are already using `ssh` so it should just work. However, if there is unexpected complexity, then feel free to fix this in a subsequent task.

lisa/sut_orchestrator/qemu/platform.py
cwize1: I still think using libvirt networks is the correct way to solve this problem. Though I realize this might be a non-trivial change. So, I am happy if you delay this to a subsequent task.

cwize1 (changes requested):
.

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: It improves readability though. `if host.address is not None:` is not as obvious as `if host.is_remote():`.

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: I need to figure out the bridged network thing. I had tried it at first but it didn't work. I am inclined to doing it in a follow up PR.

Just to understand, in what cases do we need multiple VMs on the same host? When I do implement it how do I test it?

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: I tried it out. It's not straightforward. There is no way to tell libvirt to use a specific ssh key. It picks up the config from ssh config file (e.g. `~/.ssh/config`). So we would need to modify that config programmatically and also clean it up later...

Does it really matter that the connection is unencrypted? Since this would only be used in a test environment...

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: I think there is some utility in having a separate schema because we could have libvirt specific fields in it. Right now we have the `lisa_working_dir` property that is not part of `RemoteNode`. In the future we could have a `supported_hypervisors` property for example to indicate the hypervisors (qemu, ch etc) that the host supports.

lisa/sut_orchestrator/qemu/platform.py
cwize1: You can create a simple test function with the following annotation:
```python3
@TestCaseMetadata(
    description="",
    requirement=node_requirement(
        node=schema.NodeSpace(
            node_count=2,
        )
    ),
)
```

There are a bunch of "Microsoft suite" tests that use multiple nodes, I think mainly around testing networking. I am using it to test multi-node Kubernetes clusters.

squirrelsc: @cwize1 The field `schema.Environment.topology` is used to differentiate different network topology, but it's not used so far. The value is always subnet. Please let me know, if you need to test different topology, we can discuss the requirement, and think about how to support it.

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: The Qemu can inherit the RemoteNode, so it's easy to obtain the general supports of node. The Node has working_path, and it can be overwritten to read from the schema.

lisa/sut_orchestrator/qemu/platform.py
cwize1: I think I found the docs on how to specify the SSH key: https://libvirt.org/uri.html#URI_remote. You can add a `?keyfile=` option to the end of the URL. Also, it probably make sense to use `+libssh2` instead of `+ssh` to improve performance.

It is generally a good idea to encrypt everything nowadays regardless of what it is being used for or where it is being run.

lisa/sut_orchestrator/qemu/platform.py
cwize1: @squirrelsc Unfortunately, unlike with Hyper-V which has a fairly unified networking API, libvirt/QEMU doesn't provide much assistance for networking. Unless you are doing something super simple (e.g. using the default NAT network), then a developer has to manually create the network outside of libvirt and then point the libvirt VM at that network. Typically, either Linux kernel bridges or OVS (Open vSwitch) are used but both have their challenges. OVS is a big piece of software that is complicated to use. And every Linux distro has their own networking manager API that you have to us to provision networks, such as kernel bridges. I don't think it is worth trying to pull that complexity into LISA.

squirrelsc: Thank you for the thoughts. IMO, it depends on test scenarios, and doesn't need to fully automate all things. The lab environments is complex, and it's acceptable for preconfigured steps. If some servers are setup with some topologies, LISA just needs to use it, and keep it unchanged. The LISA needs to be aware of the topology, and assign test cases by its requirements. If the different topology tests is needed from a test case, the platform checks the settings (from runbook) of each server, and then allocate VMs from matched servers. The ADO agents supports it by "capabilities". The capabilities just a couple of key/value pairs, it's very similar like what I want to do.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -360,16 +414,11 @@ def _delete_nodes(
...
-            # Delete the files created for the VM.
-            try:
-                os.remove(node_context.os_disk_file_path)
-            except Exception as ex:
-                log.warning(f"OS disk delete failed. {ex}")
-
             try:
-                os.remove(node_context.cloud_init_file_path)
+                vm_dir = os.path.dirname(node_context.os_disk_file_path)
+                self.host_node.shell.remove(Path(vm_dir), True)
```
cwize1: You can use `vm_disks_dir` here now.

lisa/tools/qemu_img.py
```diff
@@ -0,0 +1,18 @@
...
+class QemuImg(Tool):
+    @property
+    def command(self) -> str:
+        return "qemu-img"
+
+    def create_diff_qcow2(self, output_img_path: str, backing_img_path: str) -> None:
+        params = f"create -F qcow2 -f qcow2 -b {backing_img_path} {output_img_path}"
+        self.run(
+            params,
+            expected_exit_code=0,
```
cwize1, line 17: You should probably set `force_run=True` here. (Tools by default will cache and reuse the results of identical commands.)

lisa/sut_orchestrator/qemu/schema.py
```diff
@@ -15,10 +15,32 @@ class CloudInitSchema:
...
+@dataclass
+class LibvirtHost:
+    address: Optional[str] = None
+    username: Optional[str] = None
+    private_key_file: Optional[str] = None
+
+    # The directory where lisa will store VM related files (such as disk images).
+    # This directory must already exist and the test user should have write permission
+    # to it.
+    lisa_working_dir: str = "/var/tmp"
```
cwize1, line 28: Is it possible to leave the default as the base VM image's directory for the local case?

cwize1 (changes requested):
.

lisa/sut_orchestrator/qemu/schema.py
anirudhrb: I guess it could be done if needed. Is it for backward compatibility?

lisa/sut_orchestrator/qemu/platform.py
anirudhrb: Changed it to ssh. libssh didn't work. It said "libssh transport support was not enabled". Looks like it is a compile time flag that needs to be enabled while building libvirt. (Same for libssh2)

lisa/sut_orchestrator/qemu/schema.py
cwize1: When running on a local dev box, it is nice to be able to easily see all the files created, particularly for cases where you keep the environment after the test completes/fails (e.g. while debugging).

cwize1 (approved):

squirrelsc (approved):

PR #1761, opened 2022-02-22
CPU offline-online storage and network workload testcases

dsrivastavv:


microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +61,69 @@ class CPUSuite(TestSuite):
...
+        priority=3,
+        requirement=simple_requirement(
+            min_core_count=32,
+        ),
+    )
+    def verify_cpu_offline_storage_workload(self, log: Logger, node: Node) -> None:
+        # run fio process asynchronously on the node and give it 10 seconds to start
+        fio_storage_workload_script = CustomScriptBuilder(
+            Path(__file__).parent.joinpath("scripts"), ["fio_storage_worload.sh"]
+        )
```
LiliDeng: use fio tool

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +61,69 @@ class CPUSuite(TestSuite):
...
+        client = cast(RemoteNode, environment.nodes[1])
+
+        # run iperf3 process asynchronously on the server and client,
+        # and give it 10 seconds to start
+        # server.tools[Iperf3].run_as_server_async()
+        fio_network_client_workload_script = CustomScriptBuilder(
+            Path(__file__).parent.joinpath("scripts"),
+            ["fio_network_client_workload.sh"],
+        )
+        client_iperf3_process = client.tools[Iperf3].run_script_async(
```
LiliDeng: use iperf3 tool

lisa/executable.py
```diff
@@ -278,6 +278,32 @@ def run(
             expected_exit_code_failure_message=expected_exit_code_failure_message,
         )
 
+    def run_script_async(
```
squirrelsc: What's the reason to add it in tools? Follow the `withscript.py` to use the `CustomScriptBuilder` to define, and use `run_async` of `CustomScript`.

microsoft/testsuites/cpu/scripts/fio_network_client_workload.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+targetIPAddress=$1
```
squirrelsc: What's the reason, it needs a script? The iperf3 script can be started from LISA.

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +61,69 @@ class CPUSuite(TestSuite):
...
+            fio_network_client_workload_script,
+            parameters=f"{server.internal_address}",
+            shell=True,
+        )
+        time.sleep(10)
```
squirrelsc: don't hardcode sleep

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +61,69 @@ class CPUSuite(TestSuite):
...
+        requirement=simple_requirement(
+            min_core_count=32,
+        ),
+    )
+    def verify_cpu_offline_storage_workload(self, log: Logger, node: Node) -> None:
+        # run fio process asynchronously on the node and give it 10 seconds to start
+        fio_storage_workload_script = CustomScriptBuilder(
+            Path(__file__).parent.joinpath("scripts"), ["fio_storage_worload.sh"]
+        )
+        process = node.tools[Fio].run_script_async(fio_storage_workload_script)
```
squirrelsc: When is the process ended?

lisa/executable.py
dsrivastavv: The reason I added in it tools is to avoid setup instruction in the script. If we call the script from tool, it will take care of installation and then call the script

microsoft/testsuites/cpu/scripts/fio_network_client_workload.sh
dsrivastavv: `iperf3` needs to be run in a loop asynchronously. Currently in Lisa, I don't think there is a way to run this entire script with loop asynchronously. We can only run the statement "iperf3 -c $targetIPAddress -t 60 -P 32 >> workload.json" asynchronously

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: The program needs some setup time before we can run tests. This is something that was present in the v2 tests

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: I will add code to explicitly kill the process after we complete the test. The test needs to be run when a storage workload is running.

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: `iperf3` needs to be run in a loop asynchronously. Currently in Lisa, I don't think there is a way to run this entire script with loop asynchronously. We can only run the statement "iperf3 -c $targetIPAddress -t 60 -P 32 >> workload.json" asynchronously

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: Please refer to my earlier comment

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +61,69 @@ class CPUSuite(TestSuite):
...
+            The cpu hotplug steps are same as `verify_cpu_hot_plug` test case.
+            """,
+        priority=3,
+        requirement=simple_requirement(
+            min_core_count=32,
+        ),
+    )
+    def verify_cpu_offline_storage_workload(self, log: Logger, node: Node) -> None:
+        # run fio process asynchronously on the node and give it 10 seconds to start
+        fio_storage_workload_script = CustomScriptBuilder(
```
squirrelsc: It's like a metadata of the script, so, it can be in `__init__` of the test suite. The script won't be copied, until it's called from `node.tools`.

lisa/executable.py
squirrelsc: The script can be called itself, I don't get why it needs to be called from the tool. It looks the script depends on the tool here, so add dependency in the `CustomScriptBuilder`, so when it's used the tool will be installed.

microsoft/testsuites/cpu/scripts/fio_network_client_workload.sh
squirrelsc: It looks in this case, the `iperf3` doesn't need to run 10 times here. We can just make it run one time.

microsoft/testsuites/cpu/cpusuite.py
squirrelsc: we can enrich the logic in `iperf3.run_as_client_async` to check if the the networking throughput is up, and then return. If it just needs to wait the process complete starts, it can put a 0.5 sleep in the `run_as_client_async` and comment it.

lisa/executable.py
dsrivastavv: This makes sense and should be cleaner. I will update!

microsoft/testsuites/cpu/cpusuite.py
squirrelsc: When you add the logic to kill the process, put it in `final`. So, even there is an exception, it can be killed too.

microsoft/testsuites/cpu/cpusuite.py
LiliDeng: we can launch iperf3 from the tool asynchronously, if not, please add it into the tool.

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +61,69 @@ class CPUSuite(TestSuite):
...
+    )
+    def verify_cpu_offline_network_workload(
+        self, log: Logger, environment: Environment
+    ) -> None:
+        server = cast(RemoteNode, environment.nodes[0])
+        client = cast(RemoteNode, environment.nodes[1])
+
+        # run iperf3 process asynchronously on the server and client,
+        # and give it 10 seconds to start
+        # server.tools[Iperf3].run_as_server_async()
```
LiliDeng: uncomment it

dsrivastavv: I missed it. Thanks for catching

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: So for fio we need to add atleast 30 seconds of sleep. This is because the tests are not actually started until it creates file which is approx 1GB in size. Should we add this sleep in fio or terstcase?

squirrelsc: Sleep 30 seconds is terrible. It's not enough sometimes.
1. Reduce the test data size to 100M. It will speed up the test case.
2. check if there is stdout from the process. See if it's possible to check the fio starts to run.

lisa/tools/iperf3.py
```diff
@@ -251,6 +252,19 @@ def run_as_client(
...
+    def wait_start_client(self, process: Process, timeout: int = 300) -> bool:
+        # IPerf output emits lines of the following form when it is running
+        # 132.00-133.00 sec   167 MBytes  1.40 Gbits/sec    5    626 KBytes
+        # check if stdout buffers contain "bits/sec" to determine if running
+        start_time = time.time()
+        while time.time() - start_time < timeout:
+            if "bits/sec" in process._stdout_writer._buffer:
+                self._log.debug(process._stdout_writer._buffer)
+                return True
+            time.sleep(0.1)
```
squirrelsc: If it takes more than 10 seconds, sleep 1 second is ok. There may be perf impact if check it too frequently.

lisa/tools/iperf3.py
```diff
@@ -251,6 +252,19 @@ def run_as_client(
             expected_exit_code_failure_message="fail to lanuch iperf3 client",
         )
 
+    def wait_start_client(self, process: Process, timeout: int = 300) -> bool:
```
squirrelsc: `wait_warmup`?

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +58,89 @@ class CPUSuite(TestSuite):
...
+            server.tools[Iperf3].run_as_server_async()
+            client_iperf_process = client.tools[Iperf3].run_as_client_async(
+                server.internal_address, run_time_seconds=300
+            )
+
+            # wait for iperf3 to start
+            is_iperf_started = client.tools[Iperf3].wait_start_client(
+                client_iperf_process
+            )
+            assert_that(is_iperf_started, "Unable to start iperf3 process").is_true()
```
squirrelsc: It hides the details. From the logic, it means the warmup is timed out. Considering what happened if timed out, we can raise exception once it's timed out in the wait method directly.

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +58,89 @@ class CPUSuite(TestSuite):
...
+
+            # verify that the iperf3 was running when hotplug was triggered
+            assert_that(
+                client_iperf_process.is_running(),
+                "Network workload was not running during CPUhotplug",
+            ).is_true()
+        finally:
+            # kill fio process
+            server.tools[Kill].by_name("iperf3")
+            server.tools[Kill].by_name("iperf3")
```
squirrelsc: why is it killed twice? typo to client?

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +58,89 @@ class CPUSuite(TestSuite):
...
+            node.tools[Kill].by_name("fio")
+
+    @TestCaseMetadata(
+        description="""
+            This test will check cpu hotplug with network workload.
+            The cpu hotplug steps are same as `verify_cpu_hot_plug` test case.
+            """,
+        priority=3,
+        requirement=simple_requirement(
+            min_core_count=32,
```
squirrelsc: explain in the description, why it needs min 32 cores.

lisa/tools/fio.py
```diff
@@ -107,20 +111,87 @@ def launch(
...
+        )
+        process = self.run_async(
+            cmd,
+            force_run=True,
+            sudo=True,
+            cwd=cwd,
+        )
+        return process
+
+    def wait_start(self, process: Process, timeout: int = 300) -> bool:
```
squirrelsc: Keep consistent like `wait_warmup`.

lisa/tools/fio.py
```diff
@@ -107,20 +111,87 @@ def launch(
...
+        # FIO output emits lines of the form when it is running
+        # (f=10): [M(10)][24.9%][r=70.0MiB/s,w=75.0MiB/s][r=70,w=75 IOPS][eta 03m:46s]
+        # check if stdout buffers contain the string "eta" to determine if it is running
+        start_time = time.time()
+        while time.time() - start_time < timeout:
+            if "eta" in process._stdout_writer._buffer:
```
squirrelsc: The private method shouldn't be accessed outside. Please add a method called `stdout_contains` in `Process`.

lisa/tools/iperf3.py
dsrivastavv: This function is specific to the client. Is `wait_warmup_client` good?

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: Ah missed this. One should be for iperf running on client

lisa/tools/fio.py
dsrivastavv: I can put this method in `Process` but we would still need to access the `_buffer` object which is internal to the spur library. Another issue here is that `word in string` test is not very general and might not be applicable in other scenarios.

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: @LiliDeng Can you please tell me why `verify_cpu_hot_plug` has 32 core requirement?

LiliDeng: Just want to check hot plug works well with multiple cores.

lisa/tools/iperf3.py
squirrelsc: yes, how about `wait_client_warmup`?

lisa/tools/fio.py
squirrelsc: The buffer may be huge, and I'm not sure if it's format is bytes or str. So I don't want it to be copied for each comparison, contains is pretty lightweight. Let's see if there are more requirements in future. We can update it to support regexp if there are.

microsoft/testsuites/cpu/cpusuite.py
dsrivastavv: Gotcha! I think we can remove this requirement then in these testcases

lisa/tools/fio.py
dsrivastavv: I found a bug where the buffer was getting cleared faster than we were reading it. 0.1 second interval worked well but this problem came up when we moved to 1 sec interval. I have modified LogWriter to include a flush timeout and then read buffer faster than this timeout

squirrelsc: That's my concern, the buffer may be updated very fast, it's not a reliable way. It may need to add a new log handler on each process to save stdout to a string, and do search.

dsrivastavv: Let's discuss it offline to find the best way forward.

lisa/util/process.py
```diff
@@ -232,6 +239,22 @@ def is_running(self) -> bool:
             self._running = self._process.is_running()
         return self._running
 
+    def wait_warmup(self, keyword: str, timeout: int = 300) -> None:
```
squirrelsc: `wait_output` is better. It shouldn't limit it's usage by the name.

lisa/util/process.py
```diff
@@ -232,6 +239,22 @@ def is_running(self) -> bool:
...
+        # check if stdout buffers contain the string "keyword" to determine if
+        # it is running
+        start_time = time.time()
+        while time.time() - start_time < timeout:
+            if keyword in self._log_buffer.getvalue():
+                return
+
+            # LogWriter only flushes if "\n" is written, so we need to flush
+            # manually.
+            self._stdout_writer.flush()
```
squirrelsc, line 249: It should be flush before the check, so it doesn't need to wait extra 1 second.

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +58,87 @@ class CPUSuite(TestSuite):
...
+                overwrite=True,
+                time_based=True,
+            )
+
+            # FIO output emits lines of the following form when it is running
+            # (f=10): [M(10)][24.9%][r=70.0MiB/s,w=75.0MiB/s]
+            # [r=70,w=75 IOPS][eta 03m:46s]
+            # check if stdout buffers contain the string "eta" to
+            # determine if it is running
+            fio_process.wait_warmup("eta")
```
squirrelsc: this can be called in launch_async. It's more convenient to other places.

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +58,87 @@ class CPUSuite(TestSuite):
...
+            # run iperf3 process asynchronously on the server and client
+            server.tools[Iperf3].run_as_server_async()
+            client_iperf_process = client.tools[Iperf3].run_as_client_async(
+                server.internal_address, run_time_seconds=300
+            )
+
+            # IPerf output emits lines of the following form when it is running
+            # 132.00-133.00 sec   167 MBytes  1.40 Gbits/sec    5    626 KBytes
+            # check if stdout buffers contain "bits/sec" to determine if running
+            client_iperf_process.wait_warmup("bits/sec")
```
squirrelsc: move into `run_as_client_async`

microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -51,3 +58,87 @@ class CPUSuite(TestSuite):
...
+        self, log: Logger, environment: Environment
+    ) -> None:
+        server = cast(RemoteNode, environment.nodes[0])
+        client = cast(RemoteNode, environment.nodes[1])
+
+        try:
+            # run iperf3 process asynchronously on the server and client
+            server.tools[Iperf3].run_as_server_async()
+            client_iperf_process = client.tools[Iperf3].run_as_client_async(
+                server.internal_address, run_time_seconds=300
```
squirrelsc: it doesn't need to set timeout here. Keep it running.

dsrivastavv: Sorry I didn't get this. The run_time here is the total time for iperf to run. Do you mean I should set this to infinity?

squirrelsc: Yes, set it to infinity. So if the following logic takes place more than 300 seconds, the test logic is still validated. There is an overall timeout on test case level, it will prevent the test case keeps running.

squirrelsc (approved):

PR #1762, opened 2022-02-22
Add "ARM" in CpuType

lubaihua33:


LiliDeng (dismissed):

lisa/tools/lscpu.py
```diff
@@ -192,6 +192,8 @@ def get_cpu_type(self, force_run: bool = False) -> CpuType:
             return CpuType.AMD
         elif "GenuineIntel" in result.stdout:
             return CpuType.Intel
+        elif "ARM" in result.stdout:
+            return CpuType.ARM
```
squirrelsc, line 197: Add `else` to return an exception with the `result.stdout`.

squirrelsc (approved):

PR #1763, opened 2022-02-22
Migrate POWER-HIBERNATE-CLOCK-SYNC

lesscodingmorehappiness:


squirrelsc:
Can you use `git rebase origin/main` to get the commits liner? And remember to use `git fetch` to download latest remote changes.

lisa/base_tools/sed.py
```diff
@@ -53,3 +53,18 @@ def append(
             shell=True,
         )
         result.assert_exit_code(message=result.stdout)
+
+    def substitute_or_append(
```
squirrelsc, line 60: It looks not bring much value to the method. The `is_substitute` is still a flag, which needs to be calculated outside of method. Can you include the check logic in this method? Ideally, the regexp should be used by search and substitute both.

lisa/tools/fdisk.py
```diff
@@ -82,6 +82,16 @@ def delete_partitions(self, disk_name: str) -> None:
...
+    
+    def change_partition_type(self, disk_name: str, type_code: str) -> None:
+        # t -> change a partition type
+        cmd_res = self.node.execute(
+            f"(echo n; echo p; echo 1; echo ; echo ; echo t; echo {type_code}; echo p; echo w) | "
+            f"{self.command} -w always -W always {disk_name}",
+            shell=True,
+            sudo=True,
+        )
+        cmd_res.assert_exit_code()
```
squirrelsc: It can be merged into above line, and always provide the error message.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+            # readlink -f /dev/disk/azure/scsi1/lun0
+            # /dev/sdc
+            cmd_result = self._node.execute(
+                f"readlink -f {disk}", shell=True, sudo=True
+            )
+            disk_array[int(disk.split("/")[-1].replace("lun", ""))] = cmd_result.stdout
+        if disk_array and disk_array[0] == data_disk_name:
+            log.info("Data disk has already been added")
+        else:
+            disks.add_data_disk(1, schema.DiskType.StandardSSDLRS, 1024)
```
LiliDeng: this can be set in simple_requirement, add the disk during deployment.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+
+        dmesg = node.tools[Dmesg]
+        root_msg = re.findall('root=', dmesg.get_output(force_run=True), re.IGNORECASE)
+        env_setup_result = [res for res in root_msg if 'resume' in res]
+        log.debug("Env Setup Result: " + ''.join(env_setup_result))
```
LiliDeng, line 216: use one pattern to find the line which contains kernel command line and resume.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+            # Configuration for hibernation
+            node.execute(
+                "umask 022"
+            )
+
+            # GetGuestGeneration
+            exit_code = node.execute(
+                "ls /sys/firmware/efi/"
+            ).exit_code
+            os_generation = 2 if exit_code == 0 else 1
```
LiliDeng: node.tools[VmGeneration].get_generation()

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+
+            node.tools[SwapOn].run(swap_partition, shell=True, sudo=True)
+            node.tools[SwapOn].run("-s", shell=True, sudo=True)
+
+            # Seems there may not be 'sw' in blkid, so use partition name to check
+            partition_info = node.tools[Blkid].get_partition_info_by_name(swap_partition)
+            sw_uuid = "UUID=" + partition_info.uuid
+            log.debug(f"Found the Swap space disk UUID: {sw_uuid}")
+            if not sw_uuid:
+                raise SkippedException("Swap space disk UUID is empty. Abort the test.")
```
LiliDeng: LisaException?

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+            # So it will be mounted automatically during reboot
+            node.execute(
+                "chmod 766 /etc/fstab", shell = True, sudo = True
+            )
+            node.tools[Echo].write_to_file(
+                f"{sw_uuid} none swap sw 0 0",
+                "/etc/fstab",
+                sudo = True,
+                append = True
+            )
```
LiliDeng, line 320: need recover /etc/fstab after testing

lisa/tools/swap.py
```diff
@@ -14,6 +14,14 @@ def command(self) -> str:
     def _check_exists(self) -> bool:
         return True
 
+    def set_up_swap_space(self, partition: str) -> None:
+        cmd_res = self.node.execute(
+            f"mkswap {partition}",
+            shell=True,
+            sudo=True
+        )
+        cmd_res.assert_exit_code()
```
squirrelsc: See the comments in fdisk. Merge it into the execute.

lisa/tools/swap.py
```diff
@@ -14,6 +14,14 @@ def command(self) -> str:
     def _check_exists(self) -> bool:
         return True
 
+    def set_up_swap_space(self, partition: str) -> None:
```
squirrelsc: how about `create_swap`?

lisa/tools/swap.py
```diff
@@ -14,6 +14,14 @@ def command(self) -> str:
...
+        cmd_res = self.node.execute(
+            f"mkswap {partition}",
+            shell=True,
```
squirrelsc: It doesn't need `shell=True`

lisa/tools/swap.py
squirrelsc: `partition_name` is clearer. But where the actual disk space is from?

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+from lisa.features import Disk, StartStop
+from lisa.operating_system import Redhat, Suse, Ubuntu
+from lisa.testsuite import TestCaseMetadata
+from lisa.tools import Blkid, Dmesg, Echo, Fdisk, Sed, Swap, SwapOn
+from lisa.util import LisaException, SkippedException, find_patterns_in_lines
+from lisa.util.logger import Logger
+from lisa.util.shell import wait_tcp_port_ready
+
+
+def check_vm_feature_support_status(min_support_version: str, node: Node) -> bool:
```
squirrelsc: Make them private method, if they are not used outside of this file (module)

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+    if not min_support_version:
+        raise SkippedException("No min support kernel version provided!")
```
squirrelsc: It looks abnormally called this method, it should raise LisaException, not skip the test case.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+            ).exit_code
+            os_generation = 2 if exit_code == 0 else 1
+            
+            if isinstance(node.os, Redhat):
+                min_support_kernel = "4.18.0-202"
+                check_vm_support_res = check_vm_feature_support_status(min_support_kernel)
+                current_kernel_version = node.execute(
+                    "uname -r", shell = True, sudo = True
+                )
+                assert_that(
```
squirrelsc: If it needs to skip the test case, use "SkippedException". The assert failure is failed result, not skipped.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+    return True
+
+def get_year(node: Node) -> str:
+    hard_ware_clock = node.execute(
+        "hwclock",
+        shell=True,
+        sudo=True
+    ).stdout
+    if isinstance(node.os, Ubuntu) and node.os.information.version == '16.4.0':
+        year = re.split('\s+', hard_ware_clock)[3]
```
squirrelsc: 1. provide example for each pattern
2. use re.compile like other examples.
3. use match to capture, instead of use array list, it's not easy to maintain.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+    else:
+        # hwclock format: '2022-02-04 11:38:06.949136+0000'
+        year = re.split('-', hard_ware_clock)[0]
+    return year
+
+def check_is_file(node: Node, path: str) -> bool:
+    cmd_res = node.execute(
```
squirrelsc: use `os.shell.is_dir`

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,587 @@
...
+        shell=True, 
+        sudo=True
+    )
+    if cmd_res.stdout == 'yes':
+        return True
+    else:
+        return False
+
+def found_sys_log(node: Node, message: str) -> bool:
+    if check_is_file(node, '/var/log/messages'):
```
squirrelsc: It's about exists or not, not "is file". The name is confusing. Use `os.shell.exists` to check them.

lisa/tools/swap.py
lesscodingmorehappiness: Current logic is adding data disk.
As discussed we may implement Matt and Divyansh's change later

microsoft/testsuites/power/power.py
lesscodingmorehappiness: 1. added
2. and 3: do you mean writing it this way:
`hwclock = '2022-02-04 11:38:06.949136+0000'`
`pattern = re.compile("^\d{4}")`
`year = re.match(pattern, hwclock)`
`print(year[0])`

microsoft/testsuites/power/power.py
lesscodingmorehappiness: is_dir is for directory? should I add new functioin like is_file in shell.py?

microsoft/testsuites/power/power.py
lesscodingmorehappiness: a noob question: is this 'os' from python lib or lisa node.os?

microsoft/testsuites/power/power.py
lesscodingmorehappiness: simple_requirement will always use some default setting for the data disk? does that default setting fit all test cases? if not, is there any way to customize the setting, or we should use this disk.add_data_disk function

microsoft/testsuites/power/power.py
lesscodingmorehappiness: do we also need to restore other files? e.g. /etc/default/grub.d/50-cloudimg-settings.cfg

microsoft/testsuites/power/power.py
squirrelsc: The simple_requirement or TestCaseRequirement add disk when deploying, it's different with add_data_disk in the test case. Use TestCaseRequirement to add settings. It's complex but more flexible.

microsoft/testsuites/power/power.py
squirrelsc: It's better to. Make sure the recovered environment is still working.

microsoft/testsuites/power/power.py
squirrelsc: Similar. the `util/__init__.py` has some implementations of matching, it can be reused.

microsoft/testsuites/power/power.py
squirrelsc: node.os. It's cross platform and local/remote.

lisa/node.py
```diff
@@ -448,6 +448,23 @@ def get_working_path(self) -> PurePath:
 
         return self.get_pure_path(result.stdout)
 
+    def find_partition_with_freespace(self, size_in_gb: int) -> str:
```
squirrelsc, line 451: What's the reason it's copied here? The method exists on node already.

lisa/tools/fdisk.py
```diff
@@ -83,6 +83,30 @@ def delete_partitions(self, disk_name: str) -> None:
...
+    #         f"(echo n; echo p; echo 1; echo ; echo ; echo t;"
+    #         " echo {type_code}; echo p; echo w) | "
+    #         f"{self.command} -w always -W always {disk_name}",
+    #         shell=True,
+    #         sudo=True,
+    #         expected_exit_code=0,
+    #         expected_exit_code_failure_message="Faied to change the partition type."
+    #     )
+
+    def change_partition_type(self, disk_name: str, type_code: str) -> None:
```
squirrelsc, line 86: What's the reason to comment out the original one, and copy it?

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,623 @@
...
+    return True
+
+
+def _get_year(node: Node) -> str:
+    hard_ware_clock = node.execute(
+        "hwclock",
+        shell=True,
+        sudo=True
+    ).stdout
+    pattern = re.compile("^\d{4}")
```
squirrelsc: move the pattern out of the method, so, it's compiled once only.

lisa/tools/fdisk.py
lesscodingmorehappiness: Was modifying for debugging, forgot to remove the commented codes. Removing now.

lisa/node.py
lesscodingmorehappiness: This may be a new pushed one so it was not rebased before I pushed fixes? I'll do rebase again before pushing

squirrelsc:
Some commits have the same name. Please reorganize them, to make every commit is meaningful.

lesscodingmorehappiness:
> Some commits have the same name. Please reorganize them, to make every commit is meaningful.

should I squash them to be a single commit?

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,675 @@
...
+from lisa.util import (
+    LisaException,
+    SkippedException,
+    find_patterns_in_lines,
+    get_matched_str,
+)
+from lisa.util.logger import Logger
+from lisa.util.shell import wait_tcp_port_ready
+
+_YEAR_PATTERN = re.compile("^\d{4}")
```
squirrelsc: missed examples above.

squirrelsc:
> > Some commits have the same name. Please reorganize them, to make every commit is meaningful.
> 
> should I squash them to be a single commit?

It depends. Below recommendations can be referred.

1. Keep each commit as small as possible, but it should be meaningful and complete. For example, renaming a method, should make sure all place are changed. And it's worth to be in a separated commit.
2. Use `git add -p/git reset -p` to split changes in a file to two.
3. Use `git rebase -i origin/main` and "b" to insert changes between commits. It can be used to split one commit to multiple.

lisa/tools/swap.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from math import e
```
squirrelsc: What's it? Is it misimported?

lesscodingmorehappiness: it's brought by rebase I think. removed now.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+)
+from lisa.util.logger import Logger
+from lisa.util.shell import wait_tcp_port_ready
+
+# matching 4-digit year between whitespace
+# or at the beginning of a line
+# example: 2022 from '2022-02-04 11:38:06.949136+0000'
+# or 2018 from 'Fri 07 Sep 2018 11:26:52 AM CEST .838868 seconds'
+# will be matched
+_YEAR_PATTERN = re.compile("(\s\d{4}\s)|(^\d{4})")
```
squirrelsc: Try this to capture the 4 digits of year only.

`(?:\s(\d{4})\s)|(?:^(\d{4}))`

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+            log.info("VM resume completed.")
+        else:
+            raise LisaException("VM resume did not finish.")
+
+        # Note that if you use NTP, the hardware clock is
+        # automatically synchronized to the system clock every 11 minutes,
+        # and this command is useful only at boot time
+        # to get a reasonable initial system time.
+        log.info("Waiting for RTC re-sync in 12 minutes...")
+        time.sleep(720)
```
squirrelsc: Don't hard code sleep in code. Try to use some signal. In this case, it should trigger to the time sync, instead of waiting it. Wait 10 minutes in each test case is terrible.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+            " in every 15 seconds until 10 minutes timeout.")
+
+        # Verify the VM status
+        # Can not find if VM hibernation completion
+        # or not as soon as it disconnects the network. Assume it is in timeout.
+        timeout = 600
+        start_time = time.time()
+        vm_stopped = False
+        while (time.time() - start_time < timeout):
+            time.sleep(15)
```
squirrelsc: in the loop, the sleep should be 1 or 2 seconds.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+                         dmesg.get_output(force_run=True), re.IGNORECASE)
+        if res:
+            log.error("Found Call Trace or Fatal error in dmesg")
+            # This is linux-next,
+            # so there is high chance to get call trace from other issue.
+            # For now, only print the error.
+        else:
+            log.info("Not found Call Trace and Fatal error in dmesg")
+        log.info("Waiting 60-second for logging sync")
+        time.sleep(60)
```
squirrelsc: not to hard code sleep.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+        # to get a reasonable initial system time.
+        log.info("Waiting for RTC re-sync in 12 minutes...")
+        time.sleep(720)
+
+        log.info("Capturing the RTC timestmap")
+        after_time_stamp = _get_year(node)
+        log.debug(f"Got after_time_stamp: {after_time_stamp}")
+
+        # Verify the kernel panic, call trace or fatal error
+        res = re.findall('(call trace|fatal error)',
```
squirrelsc, line 676: @LiliDeng  should we include "fatal error" in `dmesg.check_kernel_errors`? The logic can be merged, if it is.

microsoft/testsuites/power/power.py
lesscodingmorehappiness: using '|' will make get_matched_str return a tuple not a str because it's actually using 're.findall'. 
So I'll change to use re.search in get_year function. Or do you think it's better to add a new function (e.g. get_first_matched_str) in init.py?

squirrelsc: Try this one? It captures the first 4 digits

`^(?:[\w\W]*?)(\d{4})(?:[-\s])`

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+                _entry3 = node.tools[Cat].read_with_filter(
+                    "/etc/default/grub",
+                    '^GRUB_TIMEOUT=30',
+                    sudo=True
+                )
+                assert_that(_entry3).described_as(
+                    f"{_entry3} - Missing config update in grub file"
+                ).is_true()
+
+            else:
```
squirrelsc: use `elif` for Ubuntu, and raise LisaException for other distros.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+                    "Successfully updated 50-cloudimg-settings.cfg file"
+                    " with all three entries.\n"
+                    "Setup Hibernate Kernel Completed.")
+
+        #################################
+        #      SETUP HIBERNATE DONE     #
+        #################################
+
+        # Reboot VM to apply swap setup changes
+        log.info("Rebooting VM")
```
squirrelsc: It doesn't need a info level log. Check if there is debug log for reboot already. If not, add a debug level log here.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+
+        node.execute_async(
+            'echo disk > /sys/power/state',
+            shell=True,
+            sudo=True
+        )
+
+        log.info(
+            "Sent hibernate command to the VM and continue checking its status"
+            " in every 15 seconds until 10 minutes timeout.")
```
squirrelsc: Do you know why it needs to do like this way? Keep sending command for 10 minutes?

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+        if vm_stopped:
+            log.info(
+                "Verified VM status is stopped after hibernation command sent")
+        else:
+            raise LisaException(
+                "Failed to change VM status to be stopped"
+                " after hibernation command sent")
+
+        # Resume the VM
+        start_stop.start()
```
squirrelsc, line 642: How about just reboot? Why it needs to shutdown and then start?

microsoft/testsuites/power/power.py
lesscodingmorehappiness: change to a loop to check every 10 sec and timeout in 12mins

microsoft/testsuites/power/power.py
lesscodingmorehappiness: changed to a loop check and timeout in 60s

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+        if after_time_stamp == str(controller_time_stamp):
+            log.info(
+                "Successfully verified the system date changed back to correct date")
+        else:
+            raise LisaException(
+                "Expected VM time changed back to the correct one after sync-up,"
+                f" but found {after_time_stamp}")
+
+        # Recover modified files
+        _recover_files(node, log, modified_files)
```
squirrelsc: this needs in the "final" section of a try...final.... Put above logic in a method, so it can be a short logic as above.

```python
try:
    _test_hibernation(...)
final:
    _recover_files(...)
```

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+        node.public_port,
+        log=log,
+        timeout=timeout,
+    )
+    return is_ready
+
+
+def _recover_files(node: Node, log: Logger, files_and_contents: dict) -> None:
+    for file in files_and_contents:
+        node.tools[Echo].write_to_file(
```
squirrelsc: Some file may be big. It's better to make a backup at the same path and recover it. It's faster.

microsoft/testsuites/power/power.py
lesscodingmorehappiness: no the command will be sent only once, we just check vm status periodically and timeout in 10 mins

microsoft/testsuites/power/power.py
lesscodingmorehappiness: shutdown is the default mode(i.e. in /sys/power/disk file) and it works when other modes could not work for some reason.
any special reason that we should change to reboot mode?

microsoft/testsuites/power/power.py
lesscodingmorehappiness: what's the prefer method we use to backup? is it 
`mv <file path> <file path>.bak`
and for restore
`rm <file path>`
`mv <file path>.bak <file path`

besides, the sed functions will create .bak files which will replace the existing one. Should we put backup files in somewhere else. e.g. /tmp ?

squirrelsc: Your approach is right. It's better to use a unique backup file name for the current tests, so it won't be overwritten by other stuffs suddenly.

lisa/tools/fdisk.py
```diff
@@ -83,6 +83,18 @@ def delete_partitions(self, disk_name: str) -> None:
                 sudo=True,
             )
 
+    def change_partition_type(self, disk_name: str, type_code: str) -> None:
+        # t -> change a partition type
+        self.node.execute(
+            f"(echo n; echo p; echo 1; echo ; echo ; echo t;"
+            " echo {type_code}; echo p; echo w) | "
+            f"{self.command} -w always -W always {disk_name}",
```
LiliDeng, line 91: -w -W not supported in older fdisk version

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+# will be matched
+_YEAR_PATTERN = re.compile("(\s\d{4}\s)|(^\d{4})")
+
+
+def _check_vm_feature_support_status(min_support_version: str, node: Node) -> bool:
+    if not min_support_version:
+        raise LisaException("No min support kernel version provided!")
+
+    kernel_version = node.tools[Uname].get_linux_information().kernel_version
+    if kernel_version > min_support_version:
```
LiliDeng, line 47: >=?

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+        return True
+    elif kernel_version < min_support_version:
+        return False
+
+    return True
+
+
+def _get_year(node: Node) -> str:
+    hard_ware_clock = node.execute(
```
LiliDeng: there is a tool hwclock, you can also wrapper this in that tool.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,680 @@
...
+        else:
+            # Configuration for hibernation
+            node.execute(
+                "umask 022"
+            )
+
+            # GetGuestGeneration
+            os_generation = node.tools[VmGeneration].get_generation()
+
+            if isinstance(node.os, Redhat):
```
LiliDeng: we can skip the case in the very beginning if current distro/kernel doesn't support hibernation at all.

microsoft/testsuites/power/power.py
squirrelsc: How about 5 seconds? It saves up to 5 seconds more.

microsoft/testsuites/power/power.py
lesscodingmorehappiness: moved the check to the beginning of the test

lisa/tools/fdisk.py
lesscodingmorehappiness: should we just remove them? in our test case it's new disk and new partition seems wipe signature is not 'necessary'?

squirrelsc:
It looks the change is merged in another PR. Feel free to reopen it, if not fully done.

PR #1764, opened 2022-02-23
Add case verify_vmbus_heartbeat_properties.

LiliDeng:


microsoft/testsuites/core/lsvmbus.py
```diff
@@ -104,3 +107,149 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
...
+                    "server_monitor_pending",
+                    "client_monitor_conn_id",
+                    "client_monitor_latency",
+                    "client_monitor_pending",
+                ]
+            )
+            vmbus_channel_files.append(["latency", "pending"])
+        for file in vmbus_driver_files:
+            node.execute(
+                f"ls -lt {device_path}/{file}",
```
squirrelsc: Can we add a tool `Ls`, which can be used to `list`, `exists`, `is_dir`?

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -104,3 +107,149 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
...
+            )
+        cmd_result = node.execute(
+            f"ls -d {device_path}/channels/*", shell=True, sudo=True
+        )
+        channel = cmd_result.stdout
+        for file in vmbus_channel_files:
+            node.execute(
+                f"ls -lt {channel}/{file}",
+                sudo=True,
+                shell=True,
```
squirrelsc: shell is not needed in this kind of commands.

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -104,3 +107,149 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
...
+                expected_exit_code=0,
+                expected_exit_code_failure_message=(f"{channel}/{file} doesn't exist"),
+            )
+        in_out_files = [
```
squirrelsc: move this up near other lists.

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -104,3 +107,149 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
...
+        origin_interrupts = node.tools[Cat].read(
+            f"{channel}/interrupts", force_run=True, sudo=True
+        )
+        origin_events = node.tools[Cat].read(
+            f"{channel}/events", force_run=True, sudo=True
+        )
+        timeout = 60
+        timer = create_timer()
+        while timeout > timer.elapsed(False):
+            current_interrupts = node.tools[Cat].read(
```
squirrelsc: Is it possible some interrupts are not increased?

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -104,3 +107,149 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
...
+                shell=True,
+                expected_exit_code=0,
+                expected_exit_code_failure_message=(
+                    f"{device_path}/{file} doesn't exist"
+                ),
+            )
+        cmd_result = node.execute(
+            f"ls -d {device_path}/channels/*", shell=True, sudo=True
+        )
+        channel = cmd_result.stdout
```
squirrelsc: Does it mean, is there only one channel? What's the value here, can you add an example?

lisa/tools/ls.py
```diff
@@ -0,0 +1,26 @@
...
+    @property
+    def command(self) -> str:
+        return "ls"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def path_exists(self, path: str) -> bool:
+        cmd_result = self.run(path, sudo=True, force_run=True)
```
squirrelsc: the sudo should be an argument, and false is default value.

microsoft/testsuites/core/lsvmbus.py
LiliDeng: added

microsoft/testsuites/core/lsvmbus.py
LiliDeng: update in tool Ls

microsoft/testsuites/core/lsvmbus.py
LiliDeng: updated

lisa/tools/ls.py
```diff
@@ -0,0 +1,26 @@
...
+        return 0 == cmd_result.exit_code
+
+    def list_dir(self, path: str) -> List[str]:
+        cmd_result = self.node.execute(
+            f"{self.command} -d {path}/*", sudo=True, shell=True
```
squirrelsc: no shell, and if it needs to support `*`, add the search_pattern as an argument and default is `*`.

microsoft/testsuites/core/lsvmbus.py
LiliDeng: tested many times, I didn't see it failed in this step.

lisa/tools/ls.py
squirrelsc: force_run should be an argument too.

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -104,3 +108,127 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
...
+            )
+            out_file_value = node.tools[Cat].read(
+                f"{device_path}/out_{in_out_file}", force_run=True, sudo=True
+            )
+            assert_that(in_file_value).described_as(
+                f"value of {device_path}/in_{in_out_file} is not equal to"
+                f" {device_path}/out_{in_out_file}"
+            ).is_equal_to(out_file_value)
+        for channel in channels:
+            origin_interrupts = node.tools[Cat].read(
```
squirrelsc: Get original values for all channels before the loop. So it gives more time to other channels to increase. So the overall waiting time should be much shorter.

microsoft/testsuites/core/lsvmbus.py
LiliDeng: I used different size, just see one folder under channels, but now I iterate all folders if it has more than 1

lisa/tools/ls.py
LiliDeng: It is not related with search, without *, it only lists current folder, not sub folder/files

squirrelsc: Sometime, it needs to list current folder only. How about add an argument `recursive`?

LiliDeng: recursive is a different concept, it needs all recursive folders under current path. I'll think about it.

squirrelsc (approved):

PR #1765, opened 2022-02-23
Add dump_variables transformer

pupacha:
Transformer that dumps given variables to a yaml.

Sample runbook
```YAML
transformer:
  - type: dump_variables
     file_path: <some_file.yml>
     variables:
        - var1
        - var2
        - var3
```

### DIFF
--- a/lisa/transformers/dump_variables.py
+++ b/lisa/transformers/dump_variables.py
@@ -11,7 +11,6 @@ from dataclasses_json import dataclass_json
 from lisa import schema
 from lisa.transformer import Transformer
 
-
 DUMP_VARIABLES = "dump_variables"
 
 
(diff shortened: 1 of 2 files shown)


lisa/transformers/dump_variables.py
```diff
@@ -0,0 +1,53 @@
...
+from lisa.util import field_metadata
+
+DUMP_VARIABLES = "dump_variables"
+
+
+@dataclass_json
+@dataclass
+class DumpVariablesTransformerSchema(schema.Transformer):
+    # items(variables) to be dumped as yaml
+    items: Dict[Any, Any] = field(default_factory=dict)
```
squirrelsc: It can be a str list, the names of variables are enough.

lisa/transformers/dump_variables.py
```diff
@@ -0,0 +1,53 @@
...
+    file_path: str = field(
+        default="",
```
squirrelsc: Give it a default value, so it's easier to use.

squirrelsc (dismissed):

squirrelsc (approved):

PR #1766, opened 2022-02-24
util: fix path generator to take low order part of microseconds

glima:
'%f', on strftime(), reads as "Microsecond as a decimal number,
zero-padded to 6 digits". If we ever want to be minimally safer
against collisions on consecutive runs, let's stick to the low order
part of that, not the higher, which has a high chance of clashing,
under the same second. Still not clash proof, but much better.

### DIFF
--- a/lisa/util/__init__.py
+++ b/lisa/util/__init__.py
@@ -254,7 +254,7 @@ def get_datetime_path(current: Optional[datetime] = None) -> str:
     if current is None:
         current = datetime.now()
     date = get_date_str(current)
-    time = current.utcnow().strftime("%H%M%S-%f")[:-3]
+    time = current.utcnow().strftime("%H%M%S-%f")[-3:]
     return f"{date}-{time}"
 
 


lisa/util/__init__.py
```diff
@@ -254,7 +254,7 @@ def get_datetime_path(current: Optional[datetime] = None) -> str:
     if current is None:
         current = datetime.now()
     date = get_date_str(current)
-    time = current.utcnow().strftime("%H%M%S-%f")[:-3]
+    time = current.utcnow().strftime("%H%M%S-%f")[-3:]
```
squirrelsc, line 257: The path is used for internal paths of a test run too. The timestamp in the path is used to align with other paths, and in the log. So, the last 3 digits make it disordered. For the path conflict, I will expose a argument as "--log-path" or "-l" next week.

glima: ? Can you elaborate more about "The timestamp in the path is used to align with other paths, and in the log. So, the last 3 digits make it disordered", please?

This is what we have:
```
lisa-v3/runtime/runs/20220107 on  main [$⇡] took 10ms ❯ ls
20220107-001203-969  20220107-001757-855  20220107-002227-014  20220107-002440-612  20220107-003003-731
20220107-001724-896  20220107-002217-141  20220107-002335-264  20220107-002703-272
```

And this is what we still will get with the patch, but with more entropy on those last 3 digits. What's the harm?

glima: I mean nothing forbids you to come up with that extra work, later, but this is no harm to what we've got today and orthogonal.

squirrelsc: If you see inside of the log folder, there are environments, and test result folders. They use this method to create the folder names. This timestamps in the folder can align with the time stamps in the log. It helps on troubleshooting concurrency issues.

glima:
```
lisa-v3/runtime/runs/20220107/20220107-003003-731 on  main [$⇡] took 24ms ❯ tree
.
├── environments
│   └── 20220107-003011-145-generated_0
│       ├── arm_template.json
│       ├── arm_template_parameters.json
│       └── node-0
│           ├── 20220107-003120-103
│           │   └── serial_console.log
│           ├── 20220107-003121-125
│           │   └── serial_console.log
│           ├── 20220107-003122-190
│           │   └── serial_console.log
│           ├── 20220107-003143-330_captured_started
│           │   ├── modinfo-hv_netvsc.txt
│           │   ├── os-release.txt
│           │   ├── uname.txt
│           │   └── uptime.txt
│           └── 20220107-003240-361
│               ├── serial_console.log
│               └── serial_console.png
├── lisa-20220107-003003-731.log
├── lisa.html
└── tests
    └── 20220107-003120-102-smoke_test
        └── 20220107-003120-102-smoke_test.log

```
I see what you mean now. But then those are independent of the overall parent dir already. How about we have an argument on that function, telling to use first/last 3 digits, and we always have the run id have it with > entropy, while the inner ones get it with less (since they won't clash anyway, they are always local, nothing to do with run_id)?

glima: And I should say we'll never get sub-second test cases, I don't see any even for our hello world runs (so the ordering still would be there, with my approach).

squirrelsc: LISA supports run test cases in parallel, but it won't happen if there is only one environment. We run the same smoke_test cases with 30 concurrencies.
Get first/last 3 digits is not a way to solve the problem on remote path conflicts thoroughly. Please wait no more than two weeks to get the configurable path.

glima: OK, thanks. I'll close this when that lands, if it fixes our issue.

glima: Thanks for the other PR, it should cover this. Closing.

PR #1767, opened 2022-02-25
add GPU pytorch test cases, variable schema change and refactoring 

squirrelsc:
see commits for details

### DIFF
--- a/lisa/variable.py
+++ b/lisa/variable.py
@@ -241,7 +241,16 @@ def add_secrets_from_pairs(
         if raw_pair.lower().startswith("s:"):
             is_secret = True
             raw_pair = raw_pair[2:]
-        key, value = raw_pair.split(":", 1)
+        try:
+            key, value = raw_pair.split(":", 1)
+        except Exception as identifier:
+            raise LisaException(
+                f"failed on parsing variable '{raw_pair}'. The right format is "
+                f"like name:value. If there is whitespace in the value, quote "
+                f'the whole string like "name:value has whitespace". If It\'s a '
+                'secret variable, follow the formant "s:name:value"'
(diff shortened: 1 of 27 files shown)


lisa/variable.py
```diff
@@ -213,7 +214,25 @@ def _load_from_file(
         raise LisaException("variable file must be dict")
 
     for key, raw_value in raw_variables.items():
-        results.update(load_from_variable_entry(key, raw_value, is_secret=is_secret))
+        try:
+            entry = convert_to_variable_entry(raw_value, is_secret=is_secret)
+            raw_value = entry.value
+            is_secret = is_secret or entry.is_secret
+            is_case_visible = entry.is_case_visible
+            mask = entry.mask
```
pupacha, line 222: How about making scheme of variables from file similar to how it is in runbook ?
Currently file variables is supported as following. 
```YAML
var1: value1
var2:
  value: value2
  is_secret: true
```
We can make it consistent with runbook schema
```YAML
- name: var1
  value: value1
  is_secret: true
- name: var2
  value: value2
  is_secret: false
```

squirrelsc: In the variable file, the current schema is still supported. It's not changed this time. If it needs to have the second format, it can be a standard runbook, and use include to use it.

pupacha: I meant, let us change file's variable schema similar to runbook's variable schema for consistency.

squirrelsc: The file format was designed to compatible with LISAv2. LISAv2 uses xml, but it's like <name>value</name> format. It's a simpler format, than LISAv3 used. So I prefer to remove support for file more than use the same format as runbook. It's duplicate with the included runbook.

PR #1774, opened 2022-03-03
QEMU orchestrator: Add support for list of user-data files.

cwize1:
Allow multiple cloud-init user-data files to be specified in
the runbook. In addition, implement rudimentary merging logic
for combining properties from different user-data files.

### DIFF
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -371,7 +371,7 @@ class QemuPlatform(Platform):
 
             # Create required directories and copy the required files to the host
             # node.
-            self.host_node.shell.mkdir(Path(node_context.vm_disks_dir))
+            self.host_node.shell.mkdir(Path(node_context.vm_disks_dir), exist_ok=True)
             if node_context.os_disk_source_file_path:
                 self.host_node.shell.copy(
                     Path(node_context.os_disk_source_file_path),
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -293,17 +293,25 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
...
-                        qemu_node_runbook.cloud_init.extra_user_data
-                    )
-                )
-                with open(extra_user_data_file_path, "r") as file:
-                    node_context.extra_cloud_init_user_data = yaml.safe_load(file)
+            )
+            if extra_user_data:
+                node_context.extra_cloud_init_user_data = []
+
+                if isinstance(extra_user_data, str):
```
squirrelsc, line 303: It looks like a bool from Ln 296, when will it be a string?

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -430,6 +438,9 @@ def _fill_nodes_metadata(
                 private_key_file=self.runbook.admin_private_key_file,
             )
 
+            # Ensure cloud-init completes its setup.
+            node.execute("cloud-init status --wait", sudo=True, expected_exit_code=0)
```
squirrelsc: for each `expected_exit_code`, it's better to specify message too. It helps on troubleshooting.

squirrelsc (dismissed):

squirrelsc (approved):

PR #1775, opened 2022-03-03
QEMU orchestrator: Add support for data disks.

cwize1:


lisa/schema.py
```diff
@@ -478,6 +478,10 @@ def _generate_min_capability(self, capability: Any) -> Any:
             min_value.data_disk_count = search_space.generate_min_capability_countspace(
                 self.data_disk_count, capability.data_disk_count
             )
+        if self.data_disk_size or capability.data_disk_size:
```
squirrelsc, line 481: @LiliDeng please check if this change will impact Azure data disk tests or not.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -359,6 +359,24 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
...
+                    node_space.disk.data_disk_count, int
+                ), f"actual: {type(node_space.disk.data_disk_count)}"
+                assert isinstance(
+                    node_space.disk.data_disk_size, int
+                ), f"actual: {type(node_space.disk.data_disk_size)}"
+
+                for i in range(node_space.disk.data_disk_count):
+                    data_disk = DataDiskContext()
+                    data_disk.file_path = os.path.join(
+                        vm_disks_dir, f"{node_context.vm_name}-data-{i}.qcow2"
```
squirrelsc, line 374: It needs to make sure the vm_name is global unique. As my understand, it's a random string, which across environments and test runs. It still has a chance to conflict. It's better to create a folder for each environment and use the folder to detect conflicts.
Once each environment is unique, it's easy to avoid conflicts inside of an environment.

cwize1: Each VM is already given their own directory.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -131,11 +136,19 @@ def _configure_node_capabilities(
...
         host_capabilities = self._get_host_capabilities(qemu_conn, log)
+        nodes_capabilities = self._create_node_capabilities(host_capabilities)
 
         nodes_requirement = []
         for node_space in environment.runbook.nodes_requirement:
-            node_capabilities = self._create_node_capabilities(log, node_space)
-            nodes_requirement.append(node_capabilities)
+            # Check that the general node capabilities are compatible with this node's
+            # specific requirements.
+            if not node_space.check(nodes_capabilities):
```
cwize1, line 145: I think I finally figured out how node capabilities/requirements are supposed to be handled. Does this look correct?

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -221,17 +232,6 @@ def _check_host_capabilities(
         total_required_memory_mib = 0
 
         for node_requirements in nodes_requirements:
-            # Ensure host has enough CPU cores for the VM.
-            # Note: The CPU scheduler can easily handle overprovisioning of CPU
-            # cores.
-            assert isinstance(node_requirements.core_count, int)
-            if node_requirements.core_count > host_capabilities.core_count:
```
cwize1, line 228: With the change to how node requirements/capabilities are handled, this logic could be moved to `_create_node_capabilities`.

lisa/sut_orchestrator/qemu/platform.py
cwize1: Also, this is just following the existing logic used by the QEMU orchestrator.

lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -359,6 +359,24 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
                             yaml.safe_load(file)
                         )
 
+            # Configure data disks.
+            if node_space.disk:
+                assert isinstance(
+                    node_space.disk.data_disk_count, int
```
cwize1, line 365: I originally tried to use `node.capability` here. But the `data_disk_count` was an `IntRange` instead of an `int`. Is this expected behavior?

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: Yes, it's right. And the capability is extendable, if there are special things in qemu to enforce like vm_size in Azure.

lisa/sut_orchestrator/qemu/platform.py
squirrelsc: The capability should be set by platform. Its default value is unlimited capacity from 0 to max.

The capability type can be rewritten by platform, like `AzureDiskOptionSettings`. But it needs to replace the default one by platform like `_convert_to_azure_node_space`. It's pretty complicated, and undocumented. Feel free to let me know, if there are more questions.

I'm planning to write some resource planner, it may be helpful for lab resource management. But don't rely on it, I don't have time allocated...

squirrelsc (approved):

PR #1785, opened 2022-03-08
DPDK multiprocess test initial refactors

mcgov:
- generalize 'run concurrent' function
- add ability to compile sample applications to testpmd tool
- create multiprocess test that tests both new items

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -4,7 +4,7 @@
 import re
 import time
 from collections import deque
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, List, Tuple, Union
 
 from assertpy import assert_that, fail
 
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -132,6 +132,46 @@ def verify_dpdk_ovs(
...
+        _init_hugepages(node)
+
+        # setup and run mp_server application
+        examples_path = test_kit.testpmd.dpdk_build_path.joinpath("examples")
+        server_app_path = examples_path.joinpath("dpdk-mp_server")
+        kit_cmd_pair = {test_kit: f"{server_app_path} --help"}
+        output: Dict[DpdkTestResources, str] = dict()
+        task_manager = _start_testpmd_concurrent(kit_cmd_pair, 5, log, output)
+
+        # TODO: Just enough to test the changes, rest of test incoming.
```
squirrelsc, line 171: Is it a draft PR?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -132,6 +132,46 @@ def verify_dpdk_ovs(
...
+
+        task_manager.wait_for_all_workers()
```
squirrelsc, line 173: I created a new wrapper called `run_in_parallel`. It works well for the simple concurrent tasks. Check code for examples.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I was hoping to check the initial refactors in before adding the rest of the test logic, so the test is sort of just a stub to verify that everything builds and runs.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I need to do stuff while concurrent things are running usually. Maybe I could extend run_in_parallel and add an async version?

squirrelsc: If so, you can start them like run_async one by one, and then verify. Finally, just wait them finished or kill them. The `run_in_parallel` is to block until all done.

mcgov: That's sort of what my current wrapper is doing, though I already put in some work to make the parallel util useful for me without changing how run_in_parallel works for other callers. I'll add the commit here and you can see what I mean.

mcgov: testing it out rn to make sure everything works still :)

mcgov: The general gist would be changing parallel to do something like this:

```
def run_in_parallel_async(
    tasks: List[Callable[[], T_RESULT]],
    callback: Callable[[T_RESULT], None],
    log: Optional[Logger] = None,
) -> TaskManager[T_RESULT]:

    task_manager = TaskManager(max_workers=len(tasks), callback=callback)
    for index, task in enumerate(tasks):
        task_manager.submit_task(Task(task_id=index, task=task, parent_logger=log))
    return task_manager


def run_in_parallel(
    tasks: List[Callable[[], T_RESULT]], log: Optional[Logger] = None
) -> List[T_RESULT]:
    """
    The simple version of concurrency task. It wait all task complete
    """
    results: List[T_RESULT] = []

    def simple_collect_result(result: T_RESULT) -> None:
        """
        Because the task is wait for all completed, and then call back the result, so
        the results are ordered.
        """
        results.append(result)

    task_manager = run_in_parallel_async(tasks, simple_collect_result, log)
    task_manager.wait_for_all_workers()
    return results
```

mcgov: I would prefer a fancier `collect_result` function and would like the wrapper to not busy wait for the results, so this gives me what I need. Otherwise I can just keep using the wrapper I had before, though I like the new version though now that I've written it 😂

squirrelsc: ok, please add the async version. Please also add comments to explain it runs all methods at the same time.

mcgov:
These refactors turned into a bunch of changes. Going to raise smaller PRs to get them in.

PR #1787, opened 2022-03-09
Update lisa installation instruction on Linux

dsrivastavv:


docs/installation_linux.rst
```diff
@@ -53,16 +57,16 @@ Poetry is used to manage Python dependencies of LISA.
 
 .. code:: bash
 
-   curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3 -
+   curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3 - --version 1.1.5
```
squirrelsc: Let's use latest version of peotry.

docs/installation_linux.rst
```diff
@@ -53,16 +57,16 @@ Poetry is used to manage Python dependencies of LISA.
...
 After running this, you should see
 ``Add export PATH="/home/YOURUSERNAME/.local/bin:$PATH" to your shell configuration file``
-on the console. Then do
+message on the console. Follow the message and add the necessary exports to ``$HOME/.profile`` file. Then do
 
 .. code:: bash
 
    source $HOME/.profile
-   poetry install
+   make setup
```
squirrelsc, line 69: It may not work on windows.

docs/installation_linux.rst
```diff
@@ -1,18 +1,24 @@
...
-bits <https://www.python.org/>`__. The latest version of Python 3 is
-recommended. If you found LISA is not compatible with higher version
+LISA has been tested to work with `Python 3.8 64-bit
+<https://www.python.org/>`__. The latest version of Python 3 is
+recommended. If you find that LISA is not compatible with higher version
 Python, `please file an
 issue <https://github.com/microsoft/lisa/issues/new>`__.
 
-Run following commands to install Python 3.8 in Ubuntu 20.04.
+Run following commands to install Python 3.8:
```
squirrelsc, line 21: What's the reason to remove Ubuntu? The below commands work on Ubuntu only.

squirrelsc: You can explain that other distros can install the similar packages.

docs/installation_linux.rst
dsrivastavv: The CI pipelines use this version of poetry, and I used the same because to ensure that the setup instructions run smoothly

docs/installation_linux.rst
dsrivastavv: This file is specific to Linux installation. The poetry commands have some flags set and the overall command seemed a bit complex. So, I used this short version instead

docs/installation_linux.rst
dsrivastavv: We had mentioned ubuntu multiple times in the command. I removed them and instead added the line 
"We will guide you through the installation of LISA on Linux. The following commands assume Ubuntu 18.04 as the underlying Linux distro." before the commands

docs/installation_linux.rst
squirrelsc: It doesn't need to follow what we used internally. It's considered stability and other reasons (maybe just no time to upgrade it). We may upgrade the poetry in pipelines later.

docs/run_test/run_locally.rst
```diff
@@ -0,0 +1,33 @@
+Running Tests Locally
```
squirrelsc: 1. The title should be like "Run tests in Local VM". It's more accurate. The LISA can run in local without a VM.
2. Link it to some page, creating an orphan page is not a good practice.

dsrivastavv: My bad! I thought the index was automatically created from directory structure by sphinx.

squirrelsc (approved):

PR #1789, opened 2022-03-10
kill: name posix signals

mcgov:
Add constants to name the signals used in the Kill tool.

### DIFF
--- a/lisa/tools/kill.py
+++ b/lisa/tools/kill.py
@@ -7,6 +7,11 @@ from .pgrep import Pgrep
 
 
 class Kill(Tool):
+
+    SIGINT = 2
+    SIGTERM = 15
+    SIGKILL = 9
+
     @property
     def command(self) -> str:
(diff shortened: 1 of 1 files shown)


dsrivastavv (approved):

squirrelsc (approved):

lisa/tools/kill.py
```diff
@@ -15,12 +20,12 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return False
 
-    def by_name(self, process_name: str, signum: int = 9) -> None:
+    def by_name(self, process_name: str, signum: int = SIGKILL) -> None:
```
squirrelsc, line 23: This may not be right in Windows, but we cannot do that more so far.

mcgov: Yep, it will need a windows version once we need it there. Can I resolve this?

PR #1790, opened 2022-03-10
add parallel run_in_parallel_async

mcgov:
Add a function to run tasks async without busy waiting, caller provides thread callback, tasks, and logger.
Does not change interface for the regular run_in_parallel.


mcgov:
This enables me to refactor code within dpdksuite to use this interface, part of a larger code cleaning in that class to remove some old cruft.

squirrelsc (approved):

PR #1792, opened 2022-03-11
DPDK: Add multiprocess test from v2

mcgov:
- Adds test using the multiprocess client-server app

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -24,7 +24,7 @@ from lisa import (
 from lisa.features import NetworkInterface, Sriov
 from lisa.nic import NicInfo, Nics
 from lisa.testsuite import simple_requirement
-from lisa.tools import Dmesg, Echo, Git, Ip, Lsmod, Lspci, Make, Modprobe, Mount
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -152,6 +152,92 @@ def verify_dpdk_nff_go(
...
+        # APP: -n : allow one client to connect
+        server_proc = node.execute_async(
+            (
+                f"{server_app_path} -l 1-2 -n 4 "
+                f"-b {node.nics.get_nic_by_index(0).pci_slot} -- -p 3 -n 1"
+            ),
+            sudo=True,
+            shell=True,
+        )
+        time.sleep(1)  # give a moment for server to start
```
squirrelsc: Is it really needed? The blind sleep makes trouble sometimes. Divyansh added a process stdout, it can help on waiting something happens.

BTW, it's better to put comments above the code, not behind.

mcgov: That should work! It will give a nicer error message if the server fails to init too.

squirrelsc (approved):

PR #1794, opened 2022-03-14
Add nested KVM NTTTCP L1 Nat testcase 

dsrivastavv:


lisa/base_tools/wget.py
```diff
@@ -43,7 +44,7 @@ def get(
         # So that here can use the corresponding path format.
         if file_path:
             # create folder when it doesn't exist
-            self.node.execute(f"mkdir -p {file_path}", shell=True, sudo=sudo)
+            self.node.shell.mkdir(PurePosixPath(file_path), exist_ok=True)
```
squirrelsc, line 47: That makes the mkdir failed, if it's a sudo path. Can you implement a mkdir tool to support it better?

lisa/tools/dnsmasq.py
```diff
@@ -0,0 +1,39 @@
...
+        kill = self.node.tools[Kill]
+        kill.by_name("dnsmasq")
+        # setup dnsmasq on interface `nic_name` and listen on `nic_address`
+        # assign dhcp address in `dhcp_range`
+        cmd = (
+            f"--strict-order --except-interface=lo --interface={nic_name} "
+            f"--listen-address={nic_address} --bind-interfaces "
+            f"--dhcp-range={dhcp_range} --conf-file='' "
+        )
+        cmd += f"--pid-file=/var/run/qemu-dnsmasq-{nic_name}.pid "
```
squirrelsc: It can be written like below, it regards like one string. You can write them in the same line. The "black" will auto format it. Like above logic, just extend it to six lines, not over use "+=", if there is no logic on each line.

```python
        cmd += f"--pid-file=/var/run/qemu-dnsmasq-{nic_name}.pid "
            f"--dhcp-leasefile=/var/run/qemu-dnsmasq-{nic_name}.leases "
            "--dhcp-no-override "
```

lisa/nic.py
```diff
@@ -325,6 +335,7 @@ def _get_node_nic_info(self) -> None:
         ).is_greater_than(0)
 
     def _get_default_nic(self) -> None:
+        print(self.nic_names)
```
squirrelsc: remove the test log.

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -172,6 +181,258 @@ def perf_nested_kvm_ntttcp_private_bridge(
...
+        node_l2.tools[Echo].write_to_file(
+            "dhclient ens4", PurePosixPath("/etc/rc.local"), append=True, sudo=True
+        )
+        node_l2.execute("chmod +x /etc/rc.local", sudo=True)
+
+        # reboot l2 vm
+        node_l2.execute_async("reboot", sudo=True)
+
+        # route traffic on `eth0` port`host_fwd_port` on l1 vm to l2 vm
+        node_l1.tools[Iptables].run(
```
squirrelsc, line 407: Can they be added into iptable tool?

lisa/base_tools/wget.py
dsrivastavv: It does handles sudo case:
```
assert self._inner_shell
try:
    self._inner_shell.mkdir(
        path_str, mode=mode, parents=parents, exist_ok=exist_ok
    )
except PermissionError:
    self._inner_shell.run(command=["sudo", "mkdir", "-p", path_str])
```

microsoft/testsuites/performance/kvmperf.py
dsrivastavv: These rules are very specific to current use case. I don't think these are something that could be general enough to make it a function

lisa/base_tools/wget.py
dsrivastavv: Can you please look at this commit: https://github.com/microsoft/lisa/pull/1794/commits/22f6bfbc67ea9434986f16eebcc45c0bafa66698?

squirrelsc (dismissed):

squirrelsc (approved):

PR #1797, opened 2022-03-14
Support tdnf for CBL-Mariner.

cwize1:
Some CBL-Mariner derivatives only have `tdnf` instead of `dnf`. So, this change allows
package installation on CBL-Mariner using either `tdnf` or `dnf` using whichever is
available.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -1293,6 +1293,7 @@ class CBLMariner(RPMDistro):
     def _dnf_tool(self) -> str:
         return self._dnf_tool_name
 
+
 @dataclass
 # `zypper lr` repolist is of the form
 # `<id>|<alias>|<name>|<enabled>|<gpg_check>|<refresh>`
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -1278,6 +1278,20 @@ class CBLMariner(RPMDistro):
...
+
+    def _initialize_package_installation(self) -> None:
+        result = self._node.execute("command -v dnf", no_info_log=True)
+        if result.exit_code == 0:
+            self._dnf_tool_name = "dnf"
+            return
+
+        self._dnf_tool_name = "tdnf"
+
+    def _dnf_tool(self) -> str:
```
squirrelsc, line 1293: It's a private method name, where will it be used?

cwize1: It's an inherited method from the Fedora class.

squirrelsc (approved):

PR #1798, opened 2022-03-14
Fixes for QEMU relating to logging directory change.

cwize1:
1. Ensure the environment's log directory is created before QEMU orchestrator
connects to any nodes.
    
2. Ensure QEMU host node logs sits under the envionment's log directory.
    
3. Ensure QEMU host node is always called "qemu-host" even if it is the
local machine.
    
4. Expose a test suite's `part_path` variable to test functions so that it
can be used when calling `local_node_connect()`.

### DIFF
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -75,6 +75,9 @@ class QemuPlatform(Platform):
         )
 
     def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        # Ensure environment log directory is created before connecting to any nodes.
+        _ = environment.log_path
+
         if len(self.qemu_platform_runbook.hosts) > 1:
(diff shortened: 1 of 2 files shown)


lisa/environment.py
```diff
@@ -366,6 +368,14 @@ def _validate_single_default(
             has_default = True
         return has_default
 
+    # Sometimes the nodes are created before the environment log or working directories
+    # are accessed for the first time and therefore the nodes create these directories.
+    # However, the 'create_folder' function will throw an exception if a directory
+    # already exists.
+    def _ensure_dirs_created(self) -> None:
```
squirrelsc: The log folder should be created in the flow. And the working path should be created explicitly when first using. The working folder is not used in most case, so it won't be created for each run. Can you show me more details?

squirrelsc (approved):

PR #1806, opened 2022-03-15
QEMU: Bug fixes and improvements for storage directories.

cwize1:
1. Since the VM disks directory is shared between all nodes, only try to delete it
once. This avoid unnecessary warnings in the logs.

2. Place the QEMU console log in the node's local log directory. Currently, it is
placed as a peer of the base QCOW2 image file.

### DIFF
--- a/lisa/sut_orchestrator/qemu/platform.py
+++ b/lisa/sut_orchestrator/qemu/platform.py
@@ -58,6 +58,7 @@ class QemuPlatform(Platform):
         self.libvirt_conn_str: str
         self.qemu_platform_runbook: QemuPlatformSchema
         self.host_node: Node
+        self.vm_disks_dir: str
 
     @classmethod
     def type_name(cls) -> str:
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/qemu/platform.py
```diff
@@ -58,6 +58,7 @@ def __init__(self, runbook: schema.Platform) -> None:
         self.libvirt_conn_str: str
         self.qemu_platform_runbook: QemuPlatformSchema
         self.host_node: Node
+        self.vm_disks_dir: str
```
squirrelsc, line 61: The static typing can be specified on first assignment, line Ln 281 in this case. Only if the first place uses the variable, it needs to declare like this way as a workaround.

cwize1: Adding it here makes life easier for the IDE language servers, which makes the IDE language integration more reliable.

squirrelsc (approved):

PR #1809, opened 2022-03-16
Environment: support to mark env or node as dirty.

squirrelsc:


lisa/environment.py
```diff
@@ -345,6 +350,10 @@ def get_information(self) -> Dict[str, str]:
 
         return final_information
 
+    def mark_dirty(self) -> None:
+        self.log.debug("mark environment to dirty")
```
dsrivastavv, line 357: Can we add environment name to improve logging information?

squirrelsc: Because this log is written to `self.log`. It has the "env[env_name]" leading the line already.

dsrivastavv (dismissed):

PR #1810, opened 2022-03-17
Add platform for Cloud Hypervisor (via libvirt)

anirudhrb:
LISA currently has a `qemu` platform that orchestrates the tests on a QEMU VM. It interacts with QEMU via libvirt. Most of this code can be used for Cloud Hypervisor too by changing the libvirt connection URI scheme from `qemu://` to `ch://`.

Extract all the re-usable code from `qemu` platform into a base class called `BaseLibvirtPlatform`. Both the `qemu` and the new cloud-hypervisor platform inherit from this base class. `BaseLibvirtPlatform` is an abstract class and cannot be used as a platform on its own. It must be extended. `BaseLibvirtPlatform` will use `BaseLibvirtPlatformSchema` and `BaseLibvirtNodeSchema` and both can be extended if required.

```
Platform
  |
  |==> BaseLibvirtPlatform
    |
    |==> QemuPlatform
    |
    |==> CloudHypervisorPlatform
    |
    |==> SomeOtherLibvirtBasedPlatform
```

All the common handling will be done in the base class and derived classes can override wherever any special handling is required based on the hypervisor used.

The following things need special handling for Cloud Hypervisor:
1. The Cloud Hypervisor driver for libvirt doesn’t support any flags during domain creation. So LISA’s `domain.createWithFlags` call throws an error. For Cloud Hypervisor no flags should be sent.  `domain.undefineFlags` should be handled along similar lines.
2. Raw disk images should be used because [qcow2 support will be dropped from cloud-hypervisor](https://github.com/cloud-hypervisor/cloud-hypervisor/issues/1985).
3.  Domain XML should specify virtio devices.

NOTE: The upstream version of libvirt has limited support for Cloud Hypervisor so [cloud-hypervisor/libvirt](https://github.com/cloud-hypervisor/libvirt) should be used.

Sample runbook:
```yaml
extension:            
  - /home/anirudh/src/lisa/microsoft/testsuites    
testcase:                                        
  - criteria:                                       
      name: "cpu_count_check" 
variable:                                     
  - name: keep_environment                 
    value: "no"                      
  - name: qcow2                   
    value: "/home/anirudh/imgs/focal-server-cloudimg-amd64.img"    
  - name: extra_user_data                                  
    value: ""                                              
  - name: admin_private_key_file               
    value: "/home/anirudh/imgs/id_rsa"         
notifier:                             
  - type: console                           
platform:                                   
  - type: cloud-hypervisor                  
    admin_username: anirudh    
    cloud-hypervisor:     
      network_boot_timeout: 60    
      hosts:    
        - address: "10.77.0.7"    
          username: "anirudh"    
    requirement:    
      cloud-hypervisor:    
        disk_img: $(qcow2)    
        disk_img_format: "qcow2"    
        firmware: "/home/anirudh/hypervisor-fw"    
        cloud_init:                              
```

Cc: @vyadavmsft @pupacha


anirudhrb:
@squirrelsc @cwize1 please provide your feedback on this approach. Suggestions with alternative approaches are also welcome.

anirudhrb:
@squirrelsc @cwize1 please provide your feedback on this approach. Suggestions for a completely new approach are also welcome.

cwize1:
> @squirrelsc @cwize1 please provide your feedback on this approach. Suggestions with alternative approaches are also welcome.

The design looks good to me.

squirrelsc:
@anirudhrb Overall, it's good to me, and I'm happy to see it happens. Do you know how other libvirt based implemented? I want to make sure we can make a more reusable libvirt base class.

lisa/mixin_modules.py
```diff
@@ -21,9 +21,12 @@
 import lisa.sut_orchestrator.ready  # noqa: F401
 
 if platform.system() == "Linux":
-    import lisa.sut_orchestrator.qemu.context  # noqa: F401
     import lisa.sut_orchestrator.qemu.platform  # noqa: F401
     import lisa.sut_orchestrator.qemu.schema  # noqa: F401
+    import lisa.sut_orchestrator.libvirt.context  # noqa: F401
+    import lisa.sut_orchestrator.libvirt.platform  # noqa: F401
+    import lisa.sut_orchestrator.libvirt.schema  # noqa: F401
+    import lisa.sut_orchestrator.cloud_hypervisor.platform  # noqa: F401
```
squirrelsc: Can we have a libvirt folder to put all of them in it? Because the libvirt is not an independent platform.

lisa/sut_orchestrator/cloud_hypervisor/platform.py
```diff
@@ -0,0 +1,84 @@
...
+class CloudHypervisorPlatform(BaseLibvirtPlatform):
+    @classmethod
+    def type_name(cls) -> str:
+        return CLOUD_HYPERVISOR
+
+    @classmethod
+    def supported_features(cls) -> List[Type[Feature]]:
+        return BaseLibvirtPlatform._supported_features
+
+    def hypervisor_name(self) -> str:
```
squirrelsc: how about name it as `protocol_name` or `schema_name`? It explicitly explains where it's used.

lisa/sut_orchestrator/cloud_hypervisor/platform.py
```diff
@@ -0,0 +1,84 @@
...
+        node_context.console_logger.attach(
+            libvirt_conn, domain, node_context.console_log_file_path
+        )
+
+    # Create the OS disk.
+    def _create_node_os_disk(
+        self, environment: Environment, log: Logger, node: Node
+    ) -> None:
+        node_context = get_node_context(node)
+        self.host_node.tools[QemuImg].convert(
```
squirrelsc: is it possible the input is raw already?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -0,0 +1,896 @@
...
+        self._init_libvirt_conn_string()
+        self._configure_environment(environment, log)
+
+        with libvirt.open(self.libvirt_conn_str) as qemu_conn:
+            return self._configure_node_capabilities(environment, log, qemu_conn)
+
+    def _deploy_environment(self, environment: Environment, log: Logger) -> None:
+        self._deploy_nodes(environment, log)
+
+    def _delete_environment(self, environment: Environment, log: Logger) -> None:
```
squirrelsc, line 133: Because this class has subclasses, if a method is intended to be called in this class only, not inheritable, use `__` (private) before them and move them after the other `_` (protected) methods. It's in LISA code guidelines, and it's common practice in Python.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -0,0 +1,896 @@
...
+        with libvirt.open(self.libvirt_conn_str) as qemu_conn:
```
squirrelsc: rename qemu, or just remove it in this case.

lisa/mixin_modules.py
anirudhrb: You mean it should not be inside `sut_orchestrator` folder? So, something like `lisa/libvirt/context.py`?

anirudhrb:
> @anirudhrb Overall, it's good to me, and I'm happy to see it happens. Do you know how other libvirt based implemented? I want to make sure we can make a more reusable libvirt base class.

Thanks! I didn't quite understand the question.

squirrelsc:
> > @anirudhrb Overall, it's good to me, and I'm happy to see it happens. Do you know how other libvirt based implemented? I want to make sure we can make a more reusable libvirt base class.
> 
> Thanks! I didn't quite understand the question.

In current PR, it supports cloud-hypervisor and qemu. The base libvirt platform exposes enough interface for their different behavior. But if other libvirt based platforms need to be supported in future, is the current interface enough? It doesn't need to support other platforms for now, but it needs to understand the impact in future, and try the best to construct a sharable interface to prevent big churn in future. When the code is used widely, it's almost impossible to do big refactoring without regression.

lisa/mixin_modules.py
squirrelsc: I mean create a libvirt folder only, and put libvirt, cloud-hypervisor and qemu all in it.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -0,0 +1,896 @@
...
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        libvirt_events_thread.init()
+
+        self.qemu_platform_runbook = self.runbook.get_extended_runbook(
+            self.__platform_runbook_type(), type_name=type(self).type_name()
+        )
+
+    def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        # Ensure environment log directory is created before connecting to any nodes.
+        _ = environment.log_path
```
squirrelsc, line 88: This line doesn't need anymore, I removed the log_path checking. You can do a favor to remove it. @cwize1

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -0,0 +1,896 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import io
```
squirrelsc: It's hard to review the code change, when a file is renamed. It's good to have rename in a separated commit, and then update. It doesn't need to operate the commits too much. Please use "git rebase -i origin/main" and add "b" before this commit. It can stop and add a new commit. In that commit, you can rename files like this, then commit it and continue with "git rebase --continue". There may be conflict, but you can accept all coming changes, because there is no code change in last step.

lisa/mixin_modules.py
anirudhrb: Okay, so there will be `lisa/sut_orchestrator/libvirt` and all files will be directly in it?

squirrelsc: yes

anirudhrb:
> > > @anirudhrb Overall, it's good to me, and I'm happy to see it happens. Do you know how other libvirt based implemented? I want to make sure we can make a more reusable libvirt base class.
> > 
> > 
> > Thanks! I didn't quite understand the question.
> 
> In current PR, it supports cloud-hypervisor and qemu. The base libvirt platform exposes enough interface for their different behavior. But if other libvirt based platforms need to be supported in future, is the current interface enough? It doesn't need to support other platforms for now, but it needs to understand the impact in future, and try the best to construct a sharable interface to prevent big churn in future. When the code is used widely, it's almost impossible to do big refactoring without regression.

I don't have extensive knowledge of the internals of libvirt and the other hypervisors supported.

AFAIK, Libvirt exposes common interfaces for all hypervisors. The current APIs should mostly work for other hypervisors too. In case there is some differences the new platform can override that method and handle it there. For example, in this PR, the Cloud Hypervisor platform doesn't support the flags while creating VM (`domain.createWithFlags`). So, I have overridden the start code and did it without the flags. Other than a couple more instances like this most of the existing QEMU platform code worked for Cloud Hypervisor too.

In the future I don't expect any huge refactoring to be required to support new hypervisors via libvirt. It will either be overriding a method in the platform or extracting some part into a new method in the base platform and then overriding it.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: But the PR will still show it as a rename. Will you then review commits separately instead?

I will try to do this.

squirrelsc: There is a view to see commit one by one. It would be better.

anirudhrb: Done. The last commit now has only the rename and the import fixes because of the rename.

cwize1:
> > > > @anirudhrb Overall, it's good to me, and I'm happy to see it happens. Do you know how other libvirt based implemented? I want to make sure we can make a more reusable libvirt base class.
> > > 
> > > 
> > > Thanks! I didn't quite understand the question.
> > 
> > 
> > In current PR, it supports cloud-hypervisor and qemu. The base libvirt platform exposes enough interface for their different behavior. But if other libvirt based platforms need to be supported in future, is the current interface enough? It doesn't need to support other platforms for now, but it needs to understand the impact in future, and try the best to construct a sharable interface to prevent big churn in future. When the code is used widely, it's almost impossible to do big refactoring without regression.
> 
> I don't have extensive knowledge of the internals of libvirt and the other hypervisors supported.
> 
> AFAIK, Libvirt exposes common interfaces for all hypervisors. The current APIs should mostly work for other hypervisors too. In case there is some differences the new platform can override that method and handle it there. For example, in this PR, the Cloud Hypervisor platform doesn't support the flags while creating VM (`domain.createWithFlags`). So, I have overridden the start code and did it without the flags. Other than a couple more instances like this most of the existing QEMU platform code worked for Cloud Hypervisor too.
> 
> In the future I don't expect any huge refactoring to be required to support new hypervisors via libvirt. It will either be overriding a method in the platform or extracting some part into a new method in the base platform and then overriding it.

It'll be pretty difficult to predict what differences there will be, if we ever add support for other Hypervisors. So, IMHO it is probably better to keep things simple for now.

cwize1:
Is this change ready for review? It is still marked as draft.

lisa/sut_orchestrator/cloud_hypervisor/platform.py
anirudhrb: Will rename this

anirudhrb:
> Is this change ready for review? It is still marked as draft.

Now it is.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -47,16 +48,48 @@ class QemuPlatformSchema:
...
+# QEMU orchestrator's per-node configuration options.
+# This ensures backward compatibility with existing runbooks that specify the
+# qcow2 property.
+@dataclass_json()
+@dataclass
+class QemuNodeSchema(BaseLibvirtNodeSchema):
+    # The disk image to use for the node.
+    # The file must use the qcow2 file format and should not be changed during test
+    # execution.
+    qcow2: str = ""
```
pupacha, line 81: `qcow2` field is no more required. `disk_img` field of `BaseLibvirtNodeSchema` class already handles this.

anirudhrb: This is for backward compatibility with existing qemu runbooks.

squirrelsc:
@anirudhrb The change is ok to me, but I need @cwize1 to approve it firstly.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -298,99 +306,137 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
...
-
-                for i in range(node_space.disk.data_disk_count):
-                    data_disk = DataDiskContext()
-                    data_disk.file_path = os.path.join(
-                        self.vm_disks_dir, f"{node_context.vm_name}-data-{i}.qcow2"
-                    )
-                    data_disk.size_gib = node_space.disk.data_disk_size
-
-                    node_context.data_disks.append(data_disk)
+        node_context.os_disk_base_file_fmt = DiskImageFormat(
```
cwize1, line 365: Out of interest, what happens if `disk_img_format` is not a valid value?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -841,9 +920,12 @@ def _try_get_node_ip_address(
         assert isinstance(addr, str)
         return addr
 
+    def hypervisor_name(self) -> str:
```
cwize1: Did you intend this to be "public"? Add underscore prefix?

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -0,0 +1,167 @@
...
+        node_context = get_node_context(node)
+        if node_context.os_disk_base_file_fmt == DiskImageFormat.QCOW2:
+            self.host_node.tools[QemuImg].convert(
+                "qcow2",
+                node_context.os_disk_base_file_path,
+                "raw",
+                node_context.os_disk_file_path,
+            )
+        else:
+            node_context.os_disk_file_path = node_context.os_disk_base_file_path
```
cwize1: I think you should be copying the file here for tests that have multiple nodes.

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: It throws an error. Something like:
```
ValueError: 'xyz' is not a valid DiskImageFormat
```

lisa/sut_orchestrator/cloud_hypervisor/platform.py
anirudhrb: Renamed it.

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -0,0 +1,173 @@
...
+        node_context = get_node_context(node)
+        if node_context.os_disk_base_file_fmt == DiskImageFormat.QCOW2:
+            self.host_node.tools[QemuImg].convert(
+                "qcow2",
+                node_context.os_disk_base_file_path,
+                "raw",
+                node_context.os_disk_file_path,
+            )
+        else:
+            self.host_node.shell.spawn(
```
cwize1: This function doesn't wait for the process to complete. It returns a process object.

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/ch_platform.py
squirrelsc: Why call spawn here? The node.execute is the right layer to call.

anirudhrb: Changed to execute().

cwize1 (approved):

squirrelsc (approved):

PR #1811, opened 2022-03-17
Migrate case verify_lis_preinstall_disk_size

LiliDeng:


lisa/sut_orchestrator/azure/tools.py
```diff
@@ -193,6 +195,34 @@ def can_install(self) -> bool:
 
         return False
 
+    def download(self) -> PurePath:
+        if not self.node.shell.exists(self.node.working_path.joinpath("LISISO")):
+            wget_tool = self.node.tools[Wget]
+            lis_path = wget_tool.get("https://aka.ms/lis", str(self.node.working_path))
+
+            result = self.node.execute(
```
squirrelsc: use the tar tool

LiliDeng: update

lisa/tools/fallocate.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def create_file(self, length_in_bytes: int, file_path: str) -> None:
+        self.run(
+            f"-l {length_in_bytes} {file_path}",
+            shell=True,
+            sudo=True,
+            force_run=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"fail to run {self.command} -l {length_in_bytes} {file_path}"
```
squirrelsc: The command line is dumped in exception already. This line can be more specific like "fail to create file by fallocate"

lisa/tools/rpm.py
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from typing import cast
```
squirrelsc: This should be in base_tools folder. We can use it in operating system classes later.

lisa/tools/stat.py
```diff
@@ -0,0 +1,65 @@
...
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        package_name = "coreutils"
+        posix_os.install_packages(package_name)
+        return self._check_exists()
+
+    def get_block_size_fs(self, file: str) -> int:
```
squirrelsc: What's the last "fs" for? Is it "file system", but the input argument is "file"? Anyway the name should be like `get_fs_block_size`.

LiliDeng: file system

lisa/tools/fallocate.py
LiliDeng: update

lisa/tools/rpm.py
LiliDeng: update

lisa/tools/stat.py
LiliDeng: update

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -364,3 +374,28 @@ def enable_ssh_on_windows(
             f"-CommandId 'RunPowerShellScript' "
             f"-Parameter @{{'PublicKey'='{public_key_data}'}}"
         )
+
+
+class Rpm(Tool):
```
squirrelsc: move it to base_tools folder, so it can be used by operating system in future. Refer to other base_tools. And I think the rpm is not installable for all distros, is it right?

https://github.com/microsoft/lisa/tree/main/lisa/base_tools

squirrelsc: Remember to refer it in tools folder like others in base_tools.

LiliDeng: update

squirrelsc (approved):

PR #1812, opened 2022-03-17
Add netperf PPS testcase

dsrivastavv:


microsoft/testsuites/performance/common.py
```diff
@@ -160,9 +160,20 @@ def perf_tcp_latency(
...
-def perf_tcp_pps(environment: Environment, test_type: str) -> None:
-    client = cast(RemoteNode, environment.nodes[0])
-    server = cast(RemoteNode, environment.nodes[1])
+def perf_tcp_pps(
+    environment: Environment,
+    test_type: str,
+    server: Optional[RemoteNode] = None,
+    client: Optional[RemoteNode] = None,
+) -> None:
+    if server is None:
```
squirrelsc: Explain more in which scenario it's set outside, and when one of them are set. If it's not right to set one of them outside, it needs to check either both of them None or both of them are set.

microsoft/testsuites/performance/kvmperf.py
```diff
@@ -210,229 +209,176 @@ def perf_nested_kvm_ntttcp_different_l1_nat(
...
                 nested_image_password,
                 self._SERVER_HOST_FWD_PORT,
                 nested_image_url,
-                image_name=self._SERVER_IMAGE,
-                taps=[self._SERVER_TAP],
-                stop_existing_vm=True,
-                name="server",
+                self._SERVER_IP_ADDR,
+                self._BR_NAME,
+                self._BR_NETWORK,
```
squirrelsc: When there are many arguments, it's better to specify names of them. It's easier to read.

microsoft/testsuites/performance/common.py
```diff
@@ -160,9 +160,23 @@ def perf_tcp_latency(
...
+    environment: Environment,
+    test_type: str,
+    server: Optional[RemoteNode] = None,
+    client: Optional[RemoteNode] = None,
+) -> None:
+    # Either server and client are set explicitly or we use
+    # the first two nodes from the environment. We never combine
+    # the two options.
+    if server is not None or client is not None:
+        assert server is not None
```
squirrelsc: Add message to let user know, when this situation happens.

squirrelsc (approved):

PR #1814, opened 2022-03-17
Add verify_cpu_offlined_channel_add testcase

dsrivastavv:
Migrate HANDLE-OFFLINED-CPU-TO-VMBUS-CHANNEL testcase

### DIFF
--- a/microsoft/testsuites/cpu/common.py
+++ b/microsoft/testsuites/cpu/common.py
@@ -84,6 +84,24 @@ def get_idle_cpus(node: Node) -> List[str]:
     return idle_cpu
 
 
+def set_cpu_state_serial(
+    log: Logger, node: Node, idle_cpu: List[str], state: str
+) -> None:
+    for target_cpu in idle_cpu:
+        log.debug(f"setting cpu{target_cpu} to {state}.")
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/cpu/common.py
```diff
@@ -84,7 +84,39 @@ def get_idle_cpus(node: Node) -> List[str]:
...
+            raise BadEnvironmentStateException(
+                (
+                    f"expected cpu{target_cpu} state: {CPUState.OFFLINE}(offline), "
+                    f"actual state: {CPUState.ONLINE}(online)."
+                    f"the test failed leaving cpu{target_cpu} in a bad state."
+                ),
+            )
+
+
+def set_idle_cpu_online(log: Logger, node: Node, idle_cpu: List[str]) -> None:
```
squirrelsc: They can be merged into one method like 
```python
def set_idle_cpu(log: Logger, node: Node, idle_cpu: List[str], is_set_online:bool) -> None:
    ....
    status = set_cpu_state(node, target_cpu, is_set_online)
    if status != is_set_online:
        ...
```

squirrelsc (approved):

PR #1816, opened 2022-03-18
Test Migration: INFINIBAND-IBM-MPI-2VM

kamalca:
New test name is verify_ibm_mpi.

Adds function to install IBM MPI then runs a ping pong test using the installed MPI.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -7,7 +7,7 @@ from typing import Any, Dict, List
 
 from assertpy import assert_that
 
-from lisa.base_tools import Sed, Uname, Wget
+from lisa.base_tools import Cat, Sed, Uname, Wget
 from lisa.feature import Feature
 from lisa.operating_system import CentOs, Redhat, Ubuntu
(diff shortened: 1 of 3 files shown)


lisa/features/infiniband.py
```diff
@@ -334,3 +343,27 @@ def install_open_mpi(self) -> None:
...
+            executable=True,
+        )
+        node.execute(
+            f"{script_path} -i silent",
+            sudo=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Failed to install IBM MPI.",
+        )
+        node.execute(
+            f"export MPI_IB_PKEY={self.get_pkey()}; make -j $(nproc)",
```
kamalca: This is not the ideal way to do this. I am looking for a work around for the environment being wiped when running with sudo.

squirrelsc: I created a fix in another PR, will merge it tomorrow.

squirrelsc (approved):

PR #1818, opened 2022-03-18
DPDK: add skip for core count requirement

mcgov:
DPDK requires at least 8 cores to run, add a function to validate
the node has enough cores to support the test.


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -817,6 +829,8 @@ def initialize_node_resources(
     pmd: str,
     sample_apps: Union[List[str], None] = None,
 ) -> DpdkTestResources:
+
+    validate_node_requirements(node)
```
squirrelsc, line 833: It can be done by configuring in requirement of metadata. It can help filter vm sizes before environment is deployed, so it's much faster, and can find testable vm sizes automatically if it's not specified.

PR #1820, opened 2022-03-19
Mitigate hibernation test cases.

LiliDeng:


lisa/features/hibernation.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class Hibernation(Feature):
+    @classmethod
+    def name(cls) -> str:
+        return FEATURE_NAME_HIBERNATION
+
+    @classmethod
+    def enabled(cls) -> bool:
+        return True
```
squirrelsc, line 16: Can we set it to false by default, I don't know how it affects the usage. But disable it by default should be better.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1559,6 +1561,7 @@ def _resource_sku_to_capability(  # noqa: C901
             [
                 schema.FeatureSettings.create(features.StartStop.name()),
                 schema.FeatureSettings.create(features.SerialConsole.name()),
+                schema.FeatureSettings.create(features.Hibernation.name()),
```
squirrelsc, line 1569: Does it mean all vm size support it?

LiliDeng: from the spec, it has some recommended sizes, I will figure out whether any API expose this.

squirrelsc: It's ok, if there is no API about it. We can just list recommended vm families.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,113 @@
...
+    def verify_hibernation_synthetic_network(self, node: Node, log: Logger) -> None:
+        self._verify_hibernation(node, log)
+
+    @TestCaseMetadata(
+        description="""
+            to do
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            min_nic_count=3,
```
squirrelsc: Why a hibernation test case needs three nics?

LiliDeng: hmm, last time there is a bug when enable sriov, after resume, one vf changed into two, I just want to make sure there is no VF count decrease or increase, just my own thoughts.

LiliDeng: now I changed into one nic.

microsoft/testsuites/power/power.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def _verify_hibernation(
+        self, environment: Environment, log: Logger, workload: str = ""
+    ) -> None:
+        client_node = cast(RemoteNode, environment.nodes[0])
+        if len(environment.nodes) >= 2:
+            server_node = cast(RemoteNode, environment.nodes[1])
+        node_nic = Nics(client_node)
+        node_nic.initialize()
+        if workload == "network":
```
squirrelsc: Is this logic shared among test cases? If not, put them into each test case. Even they are shared, they should be in separated method, instead of the "if/else" in the hibernation method.

microsoft/testsuites/power/power.py
squirrelsc: Good. If you thought there is a bug, and it needs a regression. Please create a separated test case and specify in the description.

lisa/tools/date.py
```diff
@@ -25,3 +25,15 @@ def current(self, no_error_log: bool = True) -> datetime:
                 f"'Date' return non-zero exit code: {command_result.stderr}"
             )
         return parser().parse(command_result.stdout)
+
+    def set(self, time_str: str, format: str = "") -> datetime:
```
squirrelsc: Let it pass in the Python format datetime object. This code format it here, unless we need to test the date format ourselves.

microsoft/testsuites/power/common.py
```diff
@@ -0,0 +1,75 @@
...
+from lisa.features import StartStop
+from lisa.nic import Nics
+from lisa.tools import HibernationSetup, Kill
+from lisa.util import LisaException
+from lisa.util.perf_timer import create_timer
+from lisa.util.shell import wait_tcp_port_ready
+
+
+def verify_hibernation(node: RemoteNode, log: Logger) -> None:
+    node_nic = Nics(node)
```
squirrelsc: use `node.nics`

squirrelsc (dismissed):

squirrelsc (dismissed):

lisa/tools/hibernation_setup.py
```diff
@@ -0,0 +1,66 @@
...
+from lisa.executable import Tool
+from lisa.util import find_patterns_in_lines
+
+from .dmesg import Dmesg
+from .git import Git
+from .make import Make
+
+
+class HibernationSetup(Tool):
+    repo = "https://github.com/microsoft/hibernation-setup-tool"
```
dsrivastavv: [nit] these can be private variables

lisa/tools/hibernation_setup.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git, Make]
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
```
dsrivastavv: [nit] can you please move this function to the end of class since it's a private function

lisa/tools/hibernation_setup.py
```diff
@@ -0,0 +1,66 @@
...
+    def check_exit(self) -> int:
+        dmesg = self.node.tools[Dmesg]
+        dmesg_output = dmesg.get_output(force_run=True)
+        matched_lines = find_patterns_in_lines(dmesg_output, [self.exit_pattern])
+        if not matched_lines:
+            return 0
+        return len(matched_lines[0])
+
+    def hibernate(self) -> None:
+        self.node.execute_async("systemctl hibernate", sudo=True)
```
dsrivastavv: Can we add systemctl as a tool?

lisa/tools/hibernation_setup.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def check_entry(self) -> int:
+        dmesg = self.node.tools[Dmesg]
+        dmesg_output = dmesg.get_output(force_run=True)
+        matched_lines = find_patterns_in_lines(dmesg_output, [self.entry_pattern])
+        if not matched_lines:
+            return 0
+        return len(matched_lines[0])
+
+    def check_exit(self) -> int:
```
dsrivastavv, line 44: check_entry and check_exit are pretty much same with except the matched pattern. Maybe we can add a _check function(self, pattern) and call it from check_entry and check_exit

lisa/tools/hwclock.py
```diff
@@ -25,3 +29,20 @@ def install(self) -> bool:
...
+
+    def set_datetime(self, time: datetime) -> None:
+        cmd_result = self.run(
+            f"--set --date='{time}'", shell=True, sudo=True, force_run=True
+        )
+        cmd_result.assert_exit_code()
+
+    @retry(exceptions=LisaException, tries=20, delay=0.5)  # type: ignore
+    def get(self, no_error_log: bool = True) -> datetime:
+        command_result = self.run(
```
dsrivastavv, line 45: We can use the exit code check parameter in `run` function

lisa/tools/hibernation_setup.py
LiliDeng: fixed

lisa/tools/hibernation_setup.py
LiliDeng: fixed

lisa/tools/hibernation_setup.py
LiliDeng: fixed

lisa/tools/hibernation_setup.py
LiliDeng: fixed

lisa/tools/hwclock.py
LiliDeng: fixed

lisa/tools/hibernation_setup.py
LiliDeng: fixed

squirrelsc (approved):

dsrivastavv (approved):

PR #1829, opened 2022-03-23
Transformer supports cleanup phase.

squirrelsc:


squirrelsc:
@pupacha I updated the dump variable transformer. It defaults to log path now. I think it may be a break change to your pipelines. The "enabled" supports variables now.

@lizzha The "cleanup" phase transformers run even the lisa is failed. So the VM which is created by transformer can be deleted in this phase. You can specify two delete transformers too, if it needs to be deleted earlier, instead of waiting to the end. Let me know, if there is any problem. BTW, the "enabled" of transformer can be used to change the flow of transfomers.

lisa/runner.py
```diff
@@ -367,3 +366,15 @@ def _start_loop(self) -> None:
...
+
+    def _cleanup(self) -> None:
+        try:
+            for runner in self._runners:
+                runner.close()
+        except Exception as identifier:
+            self._log.warn(f"error on close runner: {identifier}")
+
+        try:
+            transformer.run(self._runbook_builder, constants.TRANSFORMER_PHASE_CLEANUP)
```
dsrivastavv, line 378: [nit] should we add a transformer.close() instead?

squirrelsc: The `transformer` here is a module name, it's not an object. In this method, it run transformers one by one. Once a transformer is completed, it's discarded.

PR #1830, opened 2022-03-23
process fix regression on sudo without shell

squirrelsc:


squirrelsc:
@cwize1 Thank you for finding this, and sorry for the break.

cwize1 (approved):

PR #1831, opened 2022-03-23
NICs: use ordered dict for node.nics, add clearer 'get' functions for…

mcgov:
Use ordered dict for node.nics since get_nic_by_index is not guaranteed to return nics in any specific order.
Additionally add functions for DPDK tests to explicitly get the default SRIOV nic and secondary SRIOV nics. 
These functions have checks to make sure we are returning NICs with SRIOV.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -4,6 +4,7 @@
 
 import os
 import re
+from collections import OrderedDict
 from pathlib import PurePosixPath
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
@@ -130,7 +131,7 @@ class Nics(InitializableMixin):
     def __init__(self, node: "Node"):
         super().__init__()
         self._node = node
-        self.nics: Dict[str, NicInfo] = dict()
+        self.nics: Dict[str, NicInfo] = OrderedDict()
 
     def __str__(self) -> str:
         _str = ""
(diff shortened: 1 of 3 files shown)


lisa/nic.py
```diff
@@ -151,14 +152,42 @@ def get_unpaired_devices(self) -> List[str]:
         return [x.upper for x in self.nics.values() if not x.lower]
 
     def get_upper_nics(self) -> List[str]:
-        return [x.upper for x in self.nics.values()]
+        return list(self.nics.keys())
+
+    def get_paired_devices(self) -> List[str]:
```
squirrelsc, line 157: how about `get_upper_with_secondary_nic`? The secondary mean the "master" nic has a "slave" nic. If it's confusing, how about `get_upper_with_lower`?

lisa/nic.py
```diff
@@ -151,14 +152,42 @@ def get_unpaired_devices(self) -> List[str]:
...
+        return [x.upper for x in self.nics.values() if x.lower]
 
     def get_lower_nics(self) -> List[str]:
         return [x.lower for x in self.nics.values() if x.lower]
 
     def get_device_slots(self) -> List[str]:
         return [x.pci_slot for x in self.nics.values() if x.pci_slot]
 
+    def get_secondary_sriov_nic(self) -> NicInfo:
```
squirrelsc, line 166: We use lower and upper so far, the sriov is confusing here.

lisa/nic.py
```diff
@@ -151,14 +152,42 @@ def get_unpaired_devices(self) -> List[str]:
...
+        # helper for DPDK tests where we want to pick
+        # one of the nics which isn't handling the SSH connection
+
+        default_nic_name = self.default_nic
```
squirrelsc, line 170: If we can sort the nics correctly, I think the two methods are not necessarily. They make the class confusing, and hard to choose.

mcgov:
Will ditch this branch and open a new one with just the ordereddict change

PR #1832, opened 2022-03-23
DPDK Perf tests

mcgov:
Add DPDK Perf tests

- Add DpdkPerf module with tests for dual-core, multicore, and multiqueue for both netvsc and failsafe pmd
- Fix multiqueue/multicore functions and logic to allow multicore without multiqueue. 
- Multiqueue automatically uses max cores available or enough for each queue and port, add override to use less (or more) if needed.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkperf.py
+++ b/microsoft/testsuites/dpdk/dpdkperf.py
@@ -1,0 +1,242 @@
+from typing import Any, Dict, Tuple
+
+from assertpy import assert_that
+
+from lisa import (
+    Logger,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    notifier,
+    simple_requirement,
+)
+from lisa.environment import Environment
+from lisa.features import Sriov
+from lisa.messages import NetworkPPSPerformanceMessage, create_message
+from lisa.tools import Lscpu
+from lisa.util import constants
+from microsoft.testsuites.dpdk.dpdkutil import (
+    DpdkTestResources,
+    verify_dpdk_send_receive,
+    verify_dpdk_send_receive_multi_txrx_queue,
+)
+
+
+@TestSuiteMetadata(
+    area="dpdk",
+    category="performance",
+    description="""
+    This test suite is to validate DPDK performance
+    """,
+)
+class DpdkPerformance(TestSuite):
+    TIMEOUT = 12000
+
+    @TestCaseMetadata(
+        description="""
+        DPDK Performance: failsafe mode, minimal core count
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            min_count=2, network_interface=Sriov(), min_nic_count=2
+        ),
+    )
+    def perf_dpdk_failsafe_pmd_dual_core(
+        self,
+        environment: Environment,
+        log: Logger,
+        variables: Dict[str, Any],
+    ) -> None:
+        self._run_dpdk_perf_test("failsafe", environment, log, variables)
+
+    @TestCaseMetadata(
+        description="""
+        DPDK Performance: failsafe mode, maximal core count, default queue settings
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            min_count=2, network_interface=Sriov(), min_nic_count=2
+        ),
+    )
+    def perf_dpdk_failsafe_pmd_multi_core(
+        self,
+        environment: Environment,
+        log: Logger,
+        variables: Dict[str, Any],
+    ) -> None:
+
+        core_counts = [
+            n.tools[Lscpu].get_core_count() for n in environment.nodes.list()
+        ]
+
+        assert_that(core_counts).described_as(
+            "Nodes contain different core counts, DPDK Suite expects sender "
+            "and receiver to have same core count."
+        ).contains_only(core_counts[0])
+
+        self._run_dpdk_perf_test(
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -0,0 +1,242 @@
...
+        variables: Dict[str, Any],
+    ) -> None:
+        self._run_dpdk_perf_test("failsafe", environment, log, variables)
+
+    @TestCaseMetadata(
+        description="""
+        This test case gathers performance data for dpdk send-recv-fwd: failsafe pmd
+        Uses all available cores and default queues
+        """,
+        priority=2,
```
squirrelsc: Let's put perf tests on p3 by default. Unless it's fast and high impact.

microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -0,0 +1,242 @@
...
+    @TestCaseMetadata(
+        description="""
+        This test case gathers performance data for dpdk send-recv-fwd: failsafe pmd
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_count=2, network_interface=Sriov(), min_nic_count=2
+        ),
+    )
+    def perf_dpdk_failsafe_dual_core(
```
squirrelsc: Should it limit the environment has at least two cores?

microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -0,0 +1,242 @@
...
+    @TestCaseMetadata(
+        description="""
+        This test case gathers performance data for dpdk send-recv-fwd: netvsc pmd
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            min_count=2, network_interface=Sriov(), min_nic_count=2
+        ),
+    )
+    def perf_dpdk_netvsc_dual_core(
```
squirrelsc: The requirement shows "Sriov", but the test case name has "netvsc" as synthetic. Is it expected? If it's synthetic, the name "perf_dpdk_synthetic_dual_core" is better.

microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -0,0 +1,242 @@
...
+        self,
+        environment: Environment,
+        log: Logger,
+        variables: Dict[str, Any],
+    ) -> None:
+        self._run_dpdk_perf_test("netvsc", environment, log, variables)
+
+    @TestCaseMetadata(
+        description="""
+        This test case gathers performance data for dpdk send-recv-fwd: netvsc pmd
```
squirrelsc: The description is the same as above. They should describe differences among test cases.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -191,21 +192,43 @@ def generate_testpmd_command(
...
-            core_args = f"-l 0-{cores_to_use-1}"
-        else:
-            core_args = "-l 0-1"
+
+            # use either as many cores as we can or one for each queue and port
+            cores_to_use = min(cores_available, cores_to_use + txq + rxq)
+
+        # force set number of cores if the override argument is present.
+        if use_cores:
+            cores_to_use = use_cores
```
squirrelsc: The two names `cores_to_use` and `use_cores` are close here.  Can they be named better?

microsoft/testsuites/dpdk/dpdkperf.py
mcgov: Nope, DPDK will bind to the cores you tell it to. Since they're committed to doing dpdk stuff you would want your system to have more than just those cores.

microsoft/testsuites/dpdk/dpdkperf.py
mcgov: The naming conventions are confusing. I used the same ones as V2 but they're confusing. Maybe we should just change them here. Failsafe bonds to the synthetic interface which is paired with the VF so it can fail over. 'Netvsc' bonds to the VF itself but can experience interruptions if the VF has an issue.

Maybe we can change to 'failsafe' and 'vf_direct' ?

mcgov: testpmd uses those terms as arguments but we can still change the test names to be more clear

microsoft/testsuites/dpdk/dpdkperf.py
squirrelsc: Does it work with one core vm sizes?

microsoft/testsuites/dpdk/dpdkperf.py
mcgov: Ah that's right, those are the names of the DPDK PMD's. Maybe we should leave them actually, even though when you dig through the code it can be confusing because there are tons of references to the actual driver hv_netvsc.

mcgov: https://doc.dpdk.org/guides-20.11/nics/netvsc.html
https://doc.dpdk.org/guides-20.11/nics/fail_safe.html

squirrelsc: I see, thank you for the documents! let's use consistent names like dpdk.

mcgov: I'll change the test names to failsafe_pmd and netvsc_pmd instead of netvsc and failsafe

squirrelsc (approved):

PR #1834, opened 2022-03-24
hibernation: temp disable to prevent failure.

squirrelsc:


squirrelsc:
@LiliDeng some sub doesn't have the preview feature enabled, so it causes errors. It needs to follow GPU feature to add the snippets only when the setting is enabled. But GPU feature is more complicated, because it handles different drivers. So it doesn't need subclasses in hibernation.

PR #1835, opened 2022-03-24
migrate verify_client_active_interval and verify_no_pre_exist_users

LiliDeng:


lisa/base_tools/cat.py
```diff
@@ -30,9 +30,13 @@ def read_with_filter(
         grep_string: str,
         force_run: bool = False,
         sudo: bool = False,
+        include: bool = True,
```
squirrelsc, line 33: What means include here?

squirrelsc: use `invert_match` like grep explanation. The default value is False.

lisa/tools/ssh.py
```diff
@@ -47,7 +49,7 @@ def enable_public_key(self, public_key: str) -> None:
     def add_known_host(self, ip: str) -> None:
         self.node.execute(f"ssh-keyscan -H {ip} >> ~/.ssh/known_hosts", shell=True)
 
-    def get_sshd_config_path(self) -> str:
+    def get_default_sshd_config_path(self) -> str:
```
squirrelsc, line 52: Why changes the name? If it needs to support different paths, we can add a method to set path into the ssh class. The ssh class is hold by the node, so we can save value in it.

lisa/tools/ssh.py
```diff
@@ -60,8 +62,30 @@ def get_sshd_config_path(self) -> str:
...
         sed = self.node.tools[Sed]
         sed.append(f"MaxSessions {count}", config_path, sudo=True)
         service = self.node.tools[Service]
         service.restart_service("sshd")
+
+    def get(self, section: str) -> str:
+        config_path = self.get_default_sshd_config_path()
+        setting = ""
+        # remove all lines starting with #
+        cmd_result_default_config = self.node.execute(
```
squirrelsc: use Python regexp

lisa/tools/ssh.py
```diff
@@ -60,8 +62,30 @@ def get_sshd_config_path(self) -> str:
...
+            f"grep -vxE '[[:blank:]]*([#;].*)?' {config_path}", sudo=True
+        )
+        setting += cmd_result_default_config.stdout
```
squirrelsc: `setting =` is enough, there is no logic to append value before this line.

lisa/tools/ssh.py
```diff
@@ -60,8 +62,30 @@ def get_sshd_config_path(self) -> str:
...
+        if isinstance(self.node.os, Ubuntu):
+            sshd_config = "/etc/ssh/sshd_config.d/50-cloudimg-settings.conf"
+            path_exist = self.node.execute(f"ls -lt {sshd_config}", sudo=True)
+            if path_exist.exit_code == 0:
+                cmd_result_extra_config = self.node.execute(
+                    f"grep -vxE '[[:blank:]]*([#;].*)?' {sshd_config}", sudo=True
+                )
+                setting += cmd_result_extra_config.stdout
+        pattern = re.compile(rf".*{section}\s+(?P<value>.*)", re.M)
```
squirrelsc: Maybe we can ask for the whole name matches to prevent error. So the regexp should be like `"^{section}\s+(?P<value>.*)"`.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -781,3 +781,44 @@ def verify_boot_error_fail_warnings(self, node: Node) -> None:
...
+        2. Pass with warning if not find it.
+        3. Pass with warning if the value is not between 0 and 180.
+        """,
+        priority=1,
+    )
+    def verify_client_active_interval(self, node: Node) -> None:
+        ssh = node.tools[Ssh]
+        value = ssh.get("ClientAliveInterval")
+        if not value:
+            raise PassedException("not find ClientAliveInterval in sshd_config")
```
squirrelsc: Why those passed? If so, what's value of this test case?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -781,3 +781,44 @@ def verify_boot_error_fail_warnings(self, node: Node) -> None:
...
+        current_user = node.tools[Whoami].get_username()
+        cat = node.tools[Cat]
+        passwd_outputs = cat.read_with_filter(
+            "/etc/shadow", current_user, True, True, False
+        )
+        for passwd_raw_output in passwd_outputs.splitlines():
+            user_passwd = passwd_raw_output.split(":")[1]
+            user_name = passwd_raw_output.split(":")[0]
+            if not ("*" in user_passwd or "!" in user_passwd):
+                raise LisaException(f"password of user {user_name} is not deleted.")
```
squirrelsc: Describe it from positive way to say how to fix it, like 

password of user {user_name} should be deleted in /etc/shadow.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -781,3 +781,45 @@ def verify_boot_error_fail_warnings(self, node: Node) -> None:
...
+    )
+    def verify_no_pre_exist_users(self, node: Node) -> None:
+        current_user = node.tools[Whoami].get_username()
+        cat = node.tools[Cat]
+        passwd_outputs = cat.read_with_filter(
+            "/etc/shadow", current_user, True, True, True
+        )
+        for passwd_raw_output in passwd_outputs.splitlines():
+            user_passwd = passwd_raw_output.split(":")[1]
+            user_name = passwd_raw_output.split(":")[0]
```
squirrelsc: It can be done by one line like below,

user_name, user_passwd = passwd_raw_output.split(":")

squirrelsc (approved):

PR #1840, opened 2022-03-28
Enable hibernation feature.

LiliDeng:


lisa/features/hibernation.py
```diff
@@ -1,12 +1,28 @@
...
 
+from lisa import schema
 from lisa.feature import Feature
 
 FEATURE_NAME_HIBERNATION = "Hibernation"
 
 
+class HibernationSettings(schema.FeatureSettings):
+    type: str = FEATURE_NAME_HIBERNATION
+    install_by_platform: bool = True
```
squirrelsc: Because we don't support other installation in code for now, it doesn't need to provide the `install_by_platform` flag. Other than that, it should add `enabled`, and set to false by default. So even a vm size support hibernation, if it's not specified, it won't be added.

lisa/features/hibernation.py
```diff
@@ -18,3 +34,17 @@ def enabled(cls) -> bool:
     @classmethod
     def can_disable(cls) -> bool:
         return True
+
+    @classmethod
+    def _install_by_platform(cls, *args: Any, **kwargs: Any) -> None:
+        raise NotImplementedError()
+
+    @classmethod
+    def _get_resource(cls, resources: Any, type_name: str) -> Any:
```
squirrelsc: It's used in Azure implementation, so it shouldn't be a method in base Hibernation class. From the name and content, it's hard to understand what kind of resource it returns.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1052,6 +1052,9 @@ def _create_deployment_parameters(
         ):
             arm_parameters.use_availability_sets = True
 
+        if self._azure_runbook.enable_hibernation:
```
squirrelsc: With the new implementation, the arm parameter is not necessary now. The on_before_deployment can update the arm template without any parameters.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -979,12 +979,25 @@ def _select_vm_size(
...
 class Hibernation(AzureFeatureMixin, features.Hibernation):
-    @classmethod
-    def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
-        # arm_parameters: AzureArmParameter = kwargs.pop("arm_parameters")
-        # arm_parameters.enable_hibernation = True
-        ...
+    _hibernation_properties = """
+        {
+            "additionalCapabilities": {
+                "hibernationEnabled": "[parameters('enable_hibernation')]"
```
squirrelsc: It doesn't need the parameter now. If the feature is enabled, always add with value "true". Otherwise, the whole node shouldn't be added.

lisa/features/hibernation.py
```diff
@@ -1,12 +1,17 @@
...
 
 from lisa.feature import Feature
 
 FEATURE_NAME_HIBERNATION = "Hibernation"
 
 
 class Hibernation(Feature):
+    @classmethod
+    def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
+        cls._install_by_platform(*args, **kwargs)
```
squirrelsc: It doesn't need to follow GPU feature, the GPU feature is a reference.

This method can do some basic check, for example, only the feature setting is enabled, then call the `_enable_hibernation` (the `_install` is for GPU driver, hibernation doesn't need to `install`).

lisa/node.py
```diff
@@ -311,6 +311,10 @@ def mark_dirty(self) -> None:
         self.log.debug("mark node to dirty")
         self._is_dirty = True
 
+    def reconnect(self) -> None:
```
squirrelsc: When it needs to call initialize? The initialize will be called on first command. So the code just needs to close.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -639,6 +638,16 @@ def delete_file_share(
     share_service_client.delete_share(file_share_name)
 
 
+def get_resource(resources: Any, type_name: str) -> Any:
```
squirrelsc: This method is not clear. The parameters and return are `Any`. And there is no explanation what it do. What kind of "resource" is it?

squirrelsc:
Don't put many fixes in the one PR. We should make small fix get into faster.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -838,6 +838,6 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
             "/etc/shadow", current_user, True, True, True
         )
         for passwd_raw_output in passwd_outputs.splitlines():
-            user_name, user_passwd = passwd_raw_output.split(":")
+            user_name, user_passwd = passwd_raw_output.split(":")[0:2]
```
squirrelsc: Use regexp or add more comments.

microsoft/testsuites/performance/common.py
```diff
@@ -246,6 +247,14 @@ def perf_ntttcp(
         )
         for ntttcp in [client_ntttcp, server_ntttcp]:
             ntttcp.setup_system(udp_mode)
+        client.reboot()
```
squirrelsc: Add comments, why it needs reboot?

microsoft/testsuites/performance/common.py
```diff
@@ -246,6 +247,14 @@ def perf_ntttcp(
...
+        wait_tcp_port_ready(
```
squirrelsc: The reboot has logic to wait already.

squirrelsc:
@LiliDeng  is this still useful?

LiliDeng:
> @LiliDeng is this still useful?

I will split the PR, then make this one for hibernation cases only.

LiliDeng:
> @LiliDeng is this still useful?

I will split the PR, then make this one for hibernation cases only.

lisa/sut_orchestrator/azure/common.py
LiliDeng: update

squirrelsc: `find_by_name`, it's like the comments described. And it doesn't matter whether it's resource.

squirrelsc: Actually, it can be one line in Python.

`next(x for x in resources if x["type"] == type_name)`

lisa/sut_orchestrator/azure/features.py
```diff
@@ -969,12 +971,29 @@ def _select_vm_size(
...
+            raise SkippedException(
+                "Hibernation cannot be enabled on Virtual Machines created in an"
+                " Availability Set."
+            )
+        template: Any = kwargs.get("template")
+        log = cast(Logger, kwargs.get("log"))
+        log.debug("updating arm template to support vm hibernation.")
+        resources = template["resources"]
+        virtual_machines = get_resource(resources, "Microsoft.Compute/virtualMachines")
+        virtual_machines["properties"].update(json.loads(cls._hibernation_properties))
```
squirrelsc, line 999: the `update` may delete other settings. Can it change to `add`?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -969,12 +971,29 @@ def _select_vm_size(
...
+        """
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
+
+    @classmethod
+    def _enable_hibernation(cls, *args: Any, **kwargs: Any) -> None:
+        parameters: Any = kwargs.get("arm_parameters")
+        if parameters.use_availability_sets:
```
squirrelsc, line 989: When will it be created? If the availability is used by default, how to test hibernation?

LiliDeng: When specify the availability set in runbook, then use_availability_sets is true, by default use_availability_sets is set as false. 
should we remove the availability set, if it is specified in runbook.
Or we can have a pipeline to run hibernation to make sure the coverage.

lisa/sut_orchestrator/azure/features.py
LiliDeng: virtual_machines["properties"] is dict type

lisa/sut_orchestrator/azure/common.py
LiliDeng: update

squirrelsc (dismissed):

squirrelsc (approved):

PR #1841, opened 2022-03-28
Migrate Test: INFINIBAND-MVAPICH-MPI-2VM

kamalca:
New test case verify_mvapich_mpi covers functionality from LISAv2 test.

Leaving this a draft until test issue is resolved.
Blocked because tests are reporting only one process even when many processes are running, causing core tests (such as pingpong test) to fail.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -208,6 +208,8 @@ class Infiniband(Feature):
             "gcc-c++",
             "glibc.i686",
             "libgcc.i686",
+            "byacc",
+            "libevent",
         ]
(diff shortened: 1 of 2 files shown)


squirrelsc:
Is it a draft?

squirrelsc (dismissed):

squirrelsc:
Why only one process is reported? What can we do from LISA side?

kamalca:
The tests skipped due to only one process being detected does not affect the zero exit code of the command. Therefore, this test case will still pass as is without needing to disable the tests being skipped.

squirrelsc (approved):

PR #1845, opened 2022-03-29
Add 'excludeInverse' test case criteria select action.

cwize1:
This is useful for when you want to filter the set of tests to the intersection of two
include filters.


squirrelsc:
Which scenario is it needed? If it needs to intersect two conditions on different criteria, they can be specified together in the same criteria section. If it needs to intersect two conditions on the same field, for example test case name, it can be done in a regular expression.

squirrelsc:
@cwize1 as we discussed, it looks this change is not necessary. Can you clarify or close it?

PR #1847, opened 2022-03-30
Service tool exit code handling changes

mcgov:
CentOS and RHEL have some verisons which set service.rdma to refuse manual start/stop commands. This PR:
- Adds an optional exit code ignore in addition to 0 for success.
- Adds a constant to service for a systemd exit code that the setup for testpmd will encounter nonfatally on some platforms.

### DIFF
--- a/lisa/tools/service.py
+++ b/lisa/tools/service.py
@@ -2,10 +2,15 @@
 # Licensed under the MIT license.
 from typing import Type
 
-from lisa.executable import Tool
+from lisa.executable import ExecutableResult, Tool
 
 
 class Service(Tool):
+    # exit codes for systemd are documented:
+    # https://manpages.debian.org/buster/systemd/systemd.exec.5.en.html
+
+    SYSTEMD_EXIT_NOPERMISSION = 4
+
     @property
     def command(self) -> str:
         return "systemctl"
@@ -26,8 +31,8 @@ class Service(Tool):
         self._internal_tool = self.node.tools[service_type]
         return True
 
-    def restart_service(self, name: str) -> None:
-        self._internal_tool.restart_service(name)  # type: ignore
+    def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
+        self._internal_tool.restart_service(name, ignore_exit_code)  # type: ignore
 
     def stop_service(self, name: str) -> None:
         self._internal_tool.stop_service(name)  # type: ignore
@@ -72,9 +77,11 @@ class ServiceInternal(Tool):
             cmd_result = self.run(f"{name} stop", shell=True, sudo=True, force_run=True)
             cmd_result.assert_exit_code()
 
-    def restart_service(self, name: str) -> None:
+    def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
         cmd_result = self.run(f"{name} restart", shell=True, sudo=True, force_run=True)
-        cmd_result.assert_exit_code()
+        # optionally ignore exit code if it matches our expected non-zero value
+
+        _check_error_codes(cmd_result, ignore_exit_code)
 
 
 class Systemctl(Tool):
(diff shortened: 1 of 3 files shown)


lisa/tools/service.py
```diff
@@ -26,8 +31,8 @@ def _check_exists(self) -> bool:
         self._internal_tool = self.node.tools[service_type]
         return True
 
-    def restart_service(self, name: str) -> None:
-        self._internal_tool.restart_service(name)  # type: ignore
+    def restart_service(self, name: str, ignore_error_code: int = 0) -> None:
```
squirrelsc, line 34: `ignore_error` is shorter and similar.

lisa/tools/service.py
```diff
@@ -72,9 +77,11 @@ def stop_service(self, name: str) -> None:
             cmd_result = self.run(f"{name} stop", shell=True, sudo=True, force_run=True)
             cmd_result.assert_exit_code()
 
-    def restart_service(self, name: str) -> None:
+    def restart_service(self, name: str, ignore_error_code: int = 0) -> None:
         cmd_result = self.run(f"{name} restart", shell=True, sudo=True, force_run=True)
-        cmd_result.assert_exit_code()
+        # optionally ignore exit code if it matches our expected non-zero value
+        if not (ignore_error_code and cmd_result.exit_code == ignore_error_code):
```
dsrivastavv: How about we skip `assert_exit_code` check and do something like `assert_that(cmd_result.exit_code, message).is_in(0, ignore_error_code)`? We could customize the message to say that exit code was not in the list of expected values

lisa/tools/service.py
```diff
@@ -91,9 +98,10 @@ def stop_service(self, name: str) -> None:
             cmd_result = self.run(f"stop {name}", shell=True, sudo=True, force_run=True)
             cmd_result.assert_exit_code()
 
-    def restart_service(self, name: str) -> None:
+    def restart_service(self, name: str, ignore_error_code: int = 0) -> None:
         cmd_result = self.run(f"restart {name}", shell=True, sudo=True, force_run=True)
-        cmd_result.assert_exit_code()
+        if not (ignore_error_code and cmd_result.exit_code == ignore_error_code):
```
squirrelsc: This logic is not easy to understand. How about like this way?

1. update the method in ExecutableResult.assert_exit_code.
2. Change `expected_exit_code` to accept int, List[int]. The default value is 0.
3. If the exit code in this list, the exit code is ignored.

mcgov: @somil55 I took Chi's approach and modified assert_exit_code to allow for a list of acceptable codes. I'm guessing other instances of this kind of issue will pop up in the future so I like the generalized solution. 

Pushed a new commit with the changes and I've tested it with RHEL7.9

lisa/tools/service.py
```diff
@@ -26,8 +31,8 @@ def _check_exists(self) -> bool:
         self._internal_tool = self.node.tools[service_type]
         return True
 
-    def restart_service(self, name: str) -> None:
-        self._internal_tool.restart_service(name)  # type: ignore
+    def restart_service(self, name: str, ignore_error: int = 0) -> None:
```
squirrelsc: `ignored_exit_code`?

lisa/tools/service.py
```diff
@@ -120,3 +127,11 @@ def _check_service_running(self, name: str) -> bool:
         return (
             "could not be found" not in cmd_result.stdout and 0 == cmd_result.exit_code
         )
+
+
+def _check_error_codes(cmd_result: ExecutableResult, error_code: int = 0) -> None:
+    if error_code:
```
squirrelsc: I guess it's ok to include two `0`. So direct use `[0, error_code]` is simple and ok.

mcgov:
testing now, will update when ready

mcgov:
Ready for review, tests on RH7.9 passed

squirrelsc (approved):

PR #1850, opened 2022-03-31
runbook: add debug runbook 

squirrelsc:
Sample for "examples":
`lisa -r .\microsoft\runbook\debug.yml -v case:bye`

Sample for standard tier tests
`lisa -r .\microsoft\runbook\azure.yml -v case:smoke_test -v tier:debug`

Sample for other runbooks.
`lisa -r .\microsoft\runbook\debug.yml -v origin:.\microsoft\runbook\azure.yml -v case:bye`

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -298,7 +298,7 @@ class Variable:
     )
 
     name: str = field(default="")
-    value: Union[str, bool, int, Dict[Any, Any], List[Any]] = field(default="")
+    value: Union[str, bool, int, Dict[Any, Any], List[Any], None] = field(default="")
     # True means this variable can be used in test cases.
     is_case_visible: bool = False
     mask: str = ""
@@ -1120,16 +1120,17 @@ class Criteria:
     name: Optional[str] = None
     area: Optional[str] = None
     category: Optional[str] = None
-    # the runbook is complex to convert, so manual overwrite it in __post_init__.
     priority: Optional[Union[int, List[int]]] = field(
         default=None,
         metadata=field_metadata(
-            validate=ListableValidator(int, validate.Range(min=0, max=4))
+            validate=ListableValidator(int, validate.Range(min=0, max=4)),
+            allow_none=True,
         ),
     )
     # tags is a simple way to include test cases within same topic.
     tags: Optional[Union[str, List[str]]] = field(
-        default=None, metadata=field_metadata(validate=ListableValidator(str))
+        default=None,
+        metadata=field_metadata(validate=ListableValidator(str), allow_none=True),
     )
 
 
--- a/lisa/variable.py
+++ b/lisa/variable.py
@@ -324,7 +324,7 @@ def load_from_variable_entry(
 
     assert isinstance(name, str), f"actual: {type(name)}"
     results: Dict[str, VariableEntry] = {}
-    if type(raw_value) in [str, int, bool, float, list, dict]:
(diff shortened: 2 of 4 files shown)


lisa/variable.py
```diff
@@ -18,6 +18,8 @@
 _ENV_START = "LISA_"
 _SECRET_ENV_START = "S_LISA_"
 
+_none_type = type(None)
```
cwize1: Is this really needed? Why not just inline `type(None)`?

microsoft/runbook/debug.yml
```diff
@@ -0,0 +1,21 @@
+# this runbook overwrites test criteria and uses to debug test cases by name pattern.
+include:
+  - path: $(origin)
+variable:
+  - name: retry
+    value: 0
+  - name: times
+    value: 1
+  - name: origin
+    value: ../../examples/runbook/hello_world.yml
```
cwize1, line 10: Why is this pointing at the example runbook?

lisa/schema.py
```diff
@@ -298,7 +298,7 @@ class Variable:
     )
 
     name: str = field(default="")
-    value: Union[str, bool, int, Dict[Any, Any], List[Any]] = field(default="")
+    value: Union[str, bool, int, Dict[Any, Any], List[Any], None] = field(default="")
```
dsrivastavv, line 301: What happens if we do not specify a variable in runbook? Would it be treated as `None` after this change?

microsoft/runbook/debug.yml
```diff
@@ -0,0 +1,21 @@
+# this runbook overwrites test criteria and uses to debug test cases by name pattern.
+include:
+  - path: $(origin)
```
cwize1, line 3: Would it make sense for this runbook to be instead setup as an alternative tier?

cwize1 (dismissed):

lisa/variable.py
squirrelsc: It's a small perf difference, I guess Python doesn't run `type(None)` only once, if I write it inline.

microsoft/runbook/debug.yml
squirrelsc: That provides the similar experience like `lisa` does. User can specify a case name from the examples. It's easy to learn how to use the debug runbook.

lisa/schema.py
squirrelsc: If a variable is not specified, but used. There is an exception raised. So the variable is required in command line or env var, if it's not specified.

microsoft/runbook/debug.yml
squirrelsc: Good point, add one more runbook as "debug" tier. This one is kept, because if someone use the totally different test suites than microsoft, the top-level runbook can work. For example, the "examples" runbook doesn't include the Microsoft test suites, and it can use the top-level debug runbook.

lisa/variable.py
cwize1: You'll probably find that it will run slower as it will be harder for the compiler to understand the code and then optimize it. At any rate, this seems like a micro-optimization.

cwize1 (dismissed):

lisa/variable.py
squirrelsc: Agree, it's not possible to be bottleneck in Python. Change to type(None) to improve readability of the code.

PR #1851, opened 2022-04-01
CH orchestrator: copy firmware file to host node

anirudhrb:
The firmware file specified in the runbook should first be copied to the host node (if it is remote) before using. This is similar to how the OS disk image file is handled.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/ch_platform.py
+++ b/lisa/sut_orchestrator/libvirt/ch_platform.py
@@ -1,7 +1,9 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import os
 import xml.etree.ElementTree as ET  # noqa: N817
+from pathlib import Path
 from typing import List, Type
 
 import libvirt  # type: ignore
@@ -53,7 +55,35 @@ class CloudHypervisorPlatform(BaseLibvirtPlatform):
 
         assert isinstance(node_runbook, CloudHypervisorNodeSchema)
         node_context = get_node_context(node)
-        node_context.firmware_path = node_runbook.firmware
+        if self.host_node.is_remote:
+            node_context.firmware_source_path = node_runbook.firmware
+            node_context.firmware_path = os.path.join(
+                self.vm_disks_dir, os.path.basename(node_runbook.firmware)
+            )
+        else:
+            node_context.firmware_path = node_runbook.firmware
+
+    def _create_node(
+        self,
+        node: Node,
+        node_context: NodeContext,
+        environment: Environment,
+        log: Logger,
+        lv_conn: libvirt.virConnect,
+    ) -> None:
+        if node_context.firmware_source_path:
+            self.host_node.shell.copy(
+                Path(node_context.firmware_source_path),
+                Path(node_context.firmware_path),
+            )
+
+        super()._create_node(
+            node,
+            node_context,
+            environment,
+            log,
+            lv_conn,
+        )
 
     def _create_node_domain_xml(
         self, environment: Environment, log: Logger, node: Node
--- a/lisa/sut_orchestrator/libvirt/context.py
+++ b/lisa/sut_orchestrator/libvirt/context.py
@@ -25,6 +25,7 @@ class EnvironmentContext:
 @dataclass
 class NodeContext:
     vm_name: str = ""
+    firmware_source_path: str = ""
     firmware_path: str = ""
     cloud_init_file_path: str = ""
     os_disk_source_file_path: Optional[str] = None
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -442,33 +442,48 @@ class BaseLibvirtPlatform(Platform):
 
         for node in environment.nodes.list():
             node_context = get_node_context(node)
-
-            # Create required directories and copy the required files to the host
-            # node.
-            if node_context.os_disk_source_file_path:
-                self.host_node.shell.copy(
-                    Path(node_context.os_disk_source_file_path),
-                    Path(node_context.os_disk_base_file_path),
-                )
-
-            # Create cloud-init ISO file.
-            self._create_node_cloud_init_iso(environment, log, node)
-
-            # Create OS disk from the provided image.
-            self._create_node_os_disk(environment, log, node)
-
-            # Create data disks
-            self._create_node_data_disks(node)
-
-            # Create libvirt domain (i.e. VM).
-            xml = self._create_node_domain_xml(environment, log, node)
-            domain = lv_conn.defineXML(xml)
-
-            self._create_domain_and_attach_logger(
-                lv_conn,
-                domain,
+            self._create_node(
+                node,
                 node_context,
+                environment,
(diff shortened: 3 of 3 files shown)


squirrelsc:
@cwize1 @anirudhrb does it apply to qemu and others?

cwize1:
> @cwize1 @anirudhrb does it apply to qemu and others?

Typically with QEMU, the firmware file is distributed as a package within the Linux distro. It seems a little odd that CH requires the firmware to be manually copied over.

vyadavmsft:
The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware  as per instruction from CH wiki.

cwize1:
> The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.

How is CH itself being installed?

vyadavmsft:
> > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> 
> How is CH itself being installed?

https://github.com/cloud-hypervisor/cloud-hypervisor  you can build from source or copy binary from release .

cwize1:
> > > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> > 
> > 
> > How is CH itself being installed?
> 
> https://github.com/cloud-hypervisor/cloud-hypervisor you can build from source or copy binary from release .

Why can't the firmware be copied during the same step that installs CH?

vyadavmsft:
> > > > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> > > 
> > > 
> > > How is CH itself being installed?
> > 
> > 
> > https://github.com/cloud-hypervisor/cloud-hypervisor you can build from source or copy binary from release .
> 
> Why can't the firmware be copied during the same step that installs CH?

As far as i know install is not handled through lisa for CH and being done at pipeline level.

squirrelsc:
> As far as i know install is not handled through lisa for CH and being done at pipeline level.

I asked Pavan to move the logic into a transformer. So it's configurable in runbook, after it's implemented.

pupacha:
> > > > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> > > 
> > > 
> > > How is CH itself being installed?
> > 
> > 
> > https://github.com/cloud-hypervisor/cloud-hypervisor you can build from source or copy binary from release .
> 
> Why can't the firmware be copied during the same step that installs CH?

Idea was to let lisa copy all the required input files (base image file, firmware, etc) to remote host into the working dir (default /var/tmp). So, it would be consistent to let lisa do the copying.

cwize1:
> > > > > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> > > > 
> > > > 
> > > > How is CH itself being installed?
> > > 
> > > 
> > > https://github.com/cloud-hypervisor/cloud-hypervisor you can build from source or copy binary from release .
> > 
> > 
> > Why can't the firmware be copied during the same step that installs CH?
> 
> Idea was to let lisa copy all the required input files (base image file, firmware, etc) to remote host into the working dir (default /var/tmp). So, it would be consistent to let lisa do the copying.

Isn't the CH installation itself also a "required input file"?

vyadavmsft:
> > > > > > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> > > > > 
> > > > > 
> > > > > How is CH itself being installed?
> > > > 
> > > > 
> > > > https://github.com/cloud-hypervisor/cloud-hypervisor you can build from source or copy binary from release .
> > > 
> > > 
> > > Why can't the firmware be copied during the same step that installs CH?
> > 
> > 
> > Idea was to let lisa copy all the required input files (base image file, firmware, etc) to remote host into the working dir (default /var/tmp). So, it would be consistent to let lisa do the copying.
> 
> Isn't the CH installation itself also a "required input file"?

CH installation is required dependencies just like qemu platform has qemu package installation. but base image file/fw files needs to be copied to remote host. Am i missing something here ?

cwize1:
> > > > > > > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> > > > > > 
> > > > > > 
> > > > > > How is CH itself being installed?
> > > > > 
> > > > > 
> > > > > https://github.com/cloud-hypervisor/cloud-hypervisor you can build from source or copy binary from release .
> > > > 
> > > > 
> > > > Why can't the firmware be copied during the same step that installs CH?
> > > 
> > > 
> > > Idea was to let lisa copy all the required input files (base image file, firmware, etc) to remote host into the working dir (default /var/tmp). So, it would be consistent to let lisa do the copying.
> > 
> > 
> > Isn't the CH installation itself also a "required input file"?
> 
> CH installation is required dependencies just like qemu platform has qemu package installation. but base image file/fw files needs to be copied to remote host. Am i missing something here ?

In QEMU, the firmware files are also considered a dependency. Even if CH is being manually installed, why can't the firmware files be also manually installed before the test runs?

anirudhrb:
> > > > > > > > The reason for copying firmware for CH is that it is not installed from package and there is no package available yet to install, so need to copy firmware as per instruction from CH wiki.
> > > > > > > 
> > > > > > > 
> > > > > > > How is CH itself being installed?
> > > > > > 
> > > > > > 
> > > > > > https://github.com/cloud-hypervisor/cloud-hypervisor you can build from source or copy binary from release .
> > > > > 
> > > > > 
> > > > > Why can't the firmware be copied during the same step that installs CH?
> > > > 
> > > > 
> > > > Idea was to let lisa copy all the required input files (base image file, firmware, etc) to remote host into the working dir (default /var/tmp). So, it would be consistent to let lisa do the copying.
> > > 
> > > 
> > > Isn't the CH installation itself also a "required input file"?
> > 
> > 
> > CH installation is required dependencies just like qemu platform has qemu package installation. but base image file/fw files needs to be copied to remote host. Am i missing something here ?
> 
> In QEMU, the firmware files are also considered a dependency. Even if CH is being manually installed, why can't the firmware files be also manually installed before the test runs?

The difference is that we don't need to specify QEMU firmware explicitly in the domain XML. Libvirt/Qemu is able to figure out the location of the firmware file and use it. But in case of CH, no such thing happens. The firmware path needs to be explicitly specified in the domain XML. And it should be configurable via runbook because there is no "standard" location for it.

While copying the file can be done outside lisa, I thought it would be a good idea to do it in lisa for consistency. All the other file paths specified in the runbook are local paths that need to be copied to the remote host node. It would be confusing to have just this one property take a path on the remote node.

vyadavmsft:
@squirrelsc Is there any comment from your side to address ?

squirrelsc:
> @squirrelsc Is there any comment from your side to address ?

No, please got Chris sign off first.

cwize1 (approved):

squirrelsc (approved):

PR #1859, opened 2022-04-05
Add hyperv nested testcases

dsrivastavv:
Migrate following testcases from LISAV2: 

1. AZURE-NESTED-HYPERV-NTTTCP-DIFFERENT-L1-NAT
2. HYPERV-WINDOWS-NESTED-HYPERV-STORAGE-MULTIDISK
3. HYPERV-WINDOWS-NESTED-HYPERV-STORAGE-SINGLE-DISK

More information on these testcases can be found in https://github.com/microsoft/lisa/blob/master/XML/TestCases/NestedVmTests.xml

### DIFF
--- a/lisa/base_tools/wget.py
+++ b/lisa/base_tools/wget.py
@@ -1,8 +1,12 @@
 import re
 from pathlib import PurePosixPath
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Optional, Type
 
 from lisa.executable import Tool
+from lisa.tools.ls import Ls
+from lisa.tools.mkdir import Mkdir
+from lisa.tools.powershell import PowerShell
+from lisa.tools.rm import Rm
 from lisa.util import LisaException, is_valid_url
 
 if TYPE_CHECKING:
@@ -90,3 +94,50 @@ class Wget(Tool):
                 f"Internet is not accessible, exception occurred with wget {e}"
             )
         return False
+
+    @classmethod
+    def _windows_tool(cls) -> Optional[Type[Tool]]:
+        return WindowsWget
+
+
+class WindowsWget(Wget):
+    @property
+    def command(self) -> str:
+        return ""
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def get(
+        self,
+        url: str,
+        file_path: str = "",
+        filename: str = "",
+        overwrite: bool = True,
+        executable: bool = False,
+        sudo: bool = False,
+        force_run: bool = False,
+    ) -> str:
+        ls = self.node.tools[Ls]
+
+        fullpath = f"{file_path}\\{filename}"
+        # return if file exists and not overwrite
+        if ls.path_exists(file_path, sudo=sudo) and not overwrite:
+            self._log.debug(
+                f"File {fullpath} already exists and rewrite is set to False"
+            )
+
+        # create directory if it doesn't exist
+        self.node.tools[Mkdir].create_directory(file_path, sudo=sudo)
+
+        # TODO: add support for executables
+        # remove existing file if present and download
+        self.node.tools[Rm].remove_file(fullpath, sudo=sudo)
+        self.node.tools[PowerShell].run_cmdlet(
+            f"$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest -Uri '{url}'"
+            f" -OutFile '{fullpath}'",
+            sudo=sudo,
+            force_run=force_run,
+        )
+
+        return fullpath
--- a/lisa/tools/lagscope.py
+++ b/lisa/tools/lagscope.py
@@ -232,10 +232,11 @@ class Lagscope(Tool, KillableMixin):
 
     def get_average(self, result: ExecutableResult) -> Decimal:
         matched_results = self._average_pattern.match(result.stdout)
-        assert (
-            matched_results
-        ), "not found matched average latency statistics from lagscope results."
-        return Decimal(matched_results.group("average_latency_us"))
+        if matched_results:
+            return Decimal(matched_results.group("average_latency_us"))
+        else:
+            self._log.debug(f"no average latency found in {result.stdout}")
+            return Decimal(-1.0)
 
     def create_latency_performance_messages(
         self, result: ExecutableResult, environment: "Environment", test_case_name: str
--- a/lisa/tools/ls.py
+++ b/lisa/tools/ls.py
@@ -1,9 +1,10 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-from typing import List
+from typing import List, Optional, Type
 
 from lisa.executable import Tool
+from lisa.tools.powershell import PowerShell
 
 
 class Ls(Tool):
@@ -15,10 +16,12 @@ class Ls(Tool):
     def can_install(self) -> bool:
         return False
 
-    def path_exists(
-        self, path: str, sudo: bool = False, force_run: bool = False
-    ) -> bool:
-        cmd_result = self.run(path, sudo=sudo, force_run=force_run)
+    def path_exists(self, path: str, sudo: bool = False) -> bool:
+        cmd_result = self.run(
+            path,
+            force_run=True,
+            sudo=sudo,
+        )
         return 0 == cmd_result.exit_code
 
     def list_dir(self, path: str, sudo: bool = False) -> List[str]:
@@ -28,3 +31,27 @@ class Ls(Tool):
             shell=True,
         )
         return cmd_result.stdout.split()
+
+    @classmethod
+    def _windows_tool(cls) -> Optional[Type[Tool]]:
+        return WindowsLs
+
+
+class WindowsLs(Ls):
+    @property
+    def command(self) -> str:
+        return ""
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def path_exists(self, path: str, sudo: bool = False) -> bool:
+        output = self.node.tools[PowerShell].run_cmdlet(
+            f"Test-Path {path}",
+            force_run=True,
+            sudo=sudo,
+        )
+        return output.strip() == "True"
+
+    def list_dir(self, path: str, sudo: bool = False) -> List[str]:
+        raise NotImplementedError()
--- a/lisa/tools/mdadm.py
+++ b/lisa/tools/mdadm.py
@@ -1,9 +1,11 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-from typing import List, cast
+from typing import List, Optional, Type, cast
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
+from lisa.tools.hyperv import HyperV
+from lisa.tools.powershell import PowerShell
 
 
 class Mdadm(Tool):
@@ -14,11 +16,6 @@ class Mdadm(Tool):
     @property
     def can_install(self) -> bool:
         return True
-
-    def _install(self) -> bool:
-        posix_os: Posix = cast(Posix, self.node.os)
-        posix_os.install_packages("mdadm")
-        return self._check_exists()
 
     def create_raid(
         self,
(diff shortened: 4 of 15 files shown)


lisa/tools/powershell.py
```diff
@@ -24,7 +24,9 @@ def run_cmdlet(
         sudo: bool = False,
         fail_on_error: bool = True,
     ) -> str:
-        result = self.run(cmdlet, force_run=force_run, sudo=sudo)
+        result = self.node.execute(
```
squirrelsc: Consider there may be `"` in the command already. So it needs to replace `"` with escape char.

lisa/tools/reboot.py
```diff
@@ -127,3 +133,34 @@ def reboot(self, time_out: int = 300) -> None:
...
+            f"checking if SSH port {remote_node.public_port} is reachable "
+            f"on {remote_node.name}..."
+        )
+
+        # wait for ssh port to be ready
+        connected, _ = wait_tcp_port_ready(
+            address=remote_node.public_address,
+            port=remote_node.public_port,
+            log=self._log,
+            timeout=3,
```
squirrelsc: Does the timeout is enough to wait it boot up? How to make sure the node is rebooted actually, instead of stuck before reboot? Refer to Linux to check if the uptime is changed.

lisa/util/process.py
```diff
@@ -308,6 +295,37 @@ def _filter_sudo_result(self, raw_input: str) -> str:
...
+            process_result.output = process_result.output.replace(
+                "0;Administrator: C:\Windows\system32\conhost.exe", ""  # noqa: W605
+            )
+
+            # Unknown \x07 unicode character is returned by PowerShell
+            process_result.output = process_result.output.replace("\x07", "")
+
+        # Windows contains error in stdout, so we need to
+        # appropriately handle it.
+        if (
```
squirrelsc: I feel those changes are not reliable, it shouldn't be put in the lib level. If anything doesn't follow this pattern it's hard to fix. For example, some log show with Error, but it's considered as correct.
Let's move the check to each command. You can add some assert_output in ExecutableResult for convenient.

microsoft/testsuites/nested/common.py
```diff
@@ -94,6 +95,131 @@ def connect_nested_vm(
...
+        public_address=host.public_address,
+        username=guest_username,
+        password=guest_password,
+        public_port=port,
+        port=port,
+    )
+
+    # wait for nested vm ssh connection to be ready
+    try_connect(
+        ConnectionInfo(
```
squirrelsc: You can create a connection info object. It can be used field_to_dict to convert it to dict, and use it when `set_connection_info`

microsoft/testsuites/nested/common.py
```diff
@@ -94,6 +95,131 @@ def connect_nested_vm(
...
+    switch_name: str = "nestedvmswitch",
+    nat_name: str = "nestedvmnat",
+) -> None:
+    # Delete nested vm image
+    image_folder_path = "C:\\Users\\lisatest"
+    file_path = f"{image_folder_path}\\{image_name}"
+    host.tools[WindowsFileManager].delete_file(file_path)
+
+    # Delete VM
+    host.tools[HyperV].delete_vm(name)
```
squirrelsc: Assign `host.tools[HyperV]` to a variable, it looks clearer.

microsoft/testsuites/performance/nestedperf.py
```diff
@@ -79,6 +84,30 @@ def perf_nested_kvm_storage_singledisk(
...
         self._storage_perf_qemu(node, environment, variables, setup_raid=False)
 
+    @TestCaseMetadata(
+        description="""
+        This test case is to validate performance of nested VM in hyper-v
+        using fio tool with single l1 data disk attached to the l2 VM.
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            supported_os=[Windows],
```
squirrelsc: Add the Ready as well, so it's easy for testing.

microsoft/testsuites/performance/nestedperf.py
```diff
@@ -99,6 +128,31 @@ def perf_nested_kvm_storage_multidisk(
     ) -> None:
         self._storage_perf_qemu(node, environment, variables)
 
+    @TestCaseMetadata(
```
squirrelsc, line 142: Move hyperv test cases together.

lisa/tools/windows_file_manager.py
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+from lisa.operating_system import Windows
+from lisa.tools import PowerShell
+
+
+class WindowsFileManager(Tool):
```
squirrelsc: As discussed, avoid creating standalone tool for Windows only.

lisa/util/process.py
dsrivastavv: I will push this change and reboot related change in another commit.

lisa/base_tools/wget.py
```diff
@@ -90,3 +93,45 @@ def verify_internet_access(self) -> bool:
...
         return False
+
+    @classmethod
+    def _windows_tool(cls) -> Optional[Type[Tool]]:
+        return WindowsWget
+
+
+class WindowsWget(Wget):
+    @property
+    def command(self) -> str:
```
squirrelsc, line 105: It doesn't need to overwrite this to empty. Add `_check_exists` and always return True.

lisa/tools/ls.py
```diff
@@ -28,3 +31,21 @@ def list_dir(self, path: str, sudo: bool = False) -> List[str]:
...
+    def _windows_tool(cls) -> Optional[Type[Tool]]:
+        return WindowsLs
+
+
+class WindowsLs(Ls):
+    @property
+    def command(self) -> str:
+        return ""
+
+    def path_exists(self, path: str, sudo: bool = False) -> bool:
```
squirrelsc, line 48: Add other methods too, but raise NotImplementedError.

lisa/tools/mdadm.py
```diff
@@ -47,3 +44,92 @@ def stop_raid(
...
+        return self._check_exists()
+
+
+class WindowsMdadm(Mdadm):
+    @property
+    def command(self) -> str:
+        return "powershell"
+
+    @property
+    def can_install(self) -> bool:
```
squirrelsc: It shouldn't be installable, right? Override _check_exists instead of the can_install. That make the logic meaningful.

lisa/util/process.py
squirrelsc: The reboot can be improved next time, but this logic should be moved out from this change. The intermediate can be "not supported" but shouldn't be "incorrect".

dsrivastavv: Figured out a way to make reboot robust. Will psuh the changes with NAT testcase

squirrelsc: I have a PR, with this fix, the process can return exit code now. The key thing is to disable tty for Windows.
https://github.com/microsoft/lisa/pull/1873

microsoft/testsuites/nested/common.py
dsrivastavv: I assign it to a variable if it's used more than once in a function

lisa/tools/lagscope.py
```diff
@@ -232,10 +232,10 @@ def run_as_client(
...
     def get_average(self, result: ExecutableResult) -> Decimal:
         matched_results = self._average_pattern.match(result.stdout)
-        assert (
-            matched_results
-        ), "not found matched average latency statistics from lagscope results."
-        return Decimal(matched_results.group("average_latency_us"))
+        if matched_results:
+            return Decimal(matched_results.group("average_latency_us"))
+        else:
+            return Decimal(-1.0)
```
squirrelsc, line 239: Add debug log to explain why the average_latency_us is not matched, and -1 returned.

lisa/tools/uptime.py
```diff
@@ -26,3 +29,36 @@ def since_time(self, no_error_log: bool = True) -> datetime:
...
+
+        # get the last boot time
+        result = powershell.run_cmdlet(
+            "Get-CimInstance -ClassName win32_operatingsystem | select lastbootuptime",
+            force_run=True,
+            timeout=timeout,
+        )
+
+        # parse the last boot time from regex
+        match = re.search(self.DATETIME_STRING_REGEX, result)
```
squirrelsc: The datetime format may be different, so the check may return nothing. Try to output the value only, and call `parse().parse` directly.

microsoft/testsuites/nested/common.py
squirrelsc: Yes, it should be, but it's written 3 times in this method.

lisa/base_tools/wget.py
```diff
@@ -90,3 +93,48 @@ def verify_internet_access(self) -> bool:
...
+        if ls.path_exists(file_path, sudo=sudo) and not overwrite:
+            self._log.debug(
+                f"File {fullpath} already exists and rewrite is set to False"
+            )
+
+        # TODO: add support for executables
+        # remove existing file if present and download
+        self.node.tools[Rm].remove_file(fullpath, sudo=sudo)
+        self.node.tools[PowerShell].run_cmdlet(
+            f"$ProgressPreference = 'SilentlyContinue'; Invoke-WebRequest -Uri '{url}'"
```
squirrelsc, line 137: Is the `"$ProgressPreference = 'SilentlyContinue` the same as "ErrorAction" parameter? If so, use ErrorAction is clear.

dsrivastavv: This is different from error action. Error action suppress errors, but this command suppress display of download progress.

microsoft/testsuites/nested/common.py
dsrivastavv: I didn't quite get this comment. We use information from different places to create this `ConnectionInfo` object.

squirrelsc: The Ln 141 uses the variables as well. I mean, you can use the same connection info there by field_to_dict method. It saves lines of code.

dsrivastavv: I am not sure if we can use it directly. `set_connection_info` requires a field called `public_port` which is not present in `ConnectionInfo` object. So, directly applying `fields_to_dict` would not work. We will need to anyways set it explicitly in dictionary before passing which I feel is making code more complicated. Please let me know if you feel otherwise

lisa/tools/uptime.py
dsrivastavv: Updated regex to capture entire string without matching a date format explicitly

microsoft/testsuites/nested/common.py
squirrelsc: It's ok, there are default value. Either port and public_port is set, it will be used by another one.

dsrivastavv: Modified in latest commit

lisa/tools/powershell.py
```diff
@@ -23,8 +23,13 @@ def run_cmdlet(
         force_run: bool = False,
         sudo: bool = False,
         fail_on_error: bool = True,
+        timeout: int = 600,
     ) -> str:
-        result = self.run(cmdlet, force_run=force_run, sudo=sudo)
+        # cmdlet can have `"` characters, so we need to escape them
+        cmdlet = cmdlet.replace('"', '\\"')
```
squirrelsc: Does the `\` work in PowerShell? Should it be "`"?

lisa/tools/mdadm.py
```diff
@@ -47,3 +44,86 @@ def stop_raid(
...
+                force_run=True,
+            )
+
+    def _create_pool(self, pool_name: str) -> None:
+        # delete pool if exists
+        self._delete_pool(pool_name)
+
+        # create pool
+        self.node.tools[PowerShell].run_cmdlet(
+            "$disks = Get-PhysicalDisk -CanPool  $true; New-StoragePool "
```
squirrelsc, line 125: Is the pool the same as raid?

microsoft/testsuites/nested/common.py
```diff
@@ -72,28 +76,108 @@ def connect_nested_vm(
...
 
+def hyperv_remove_nested_vm(
+    host: RemoteNode,
+    name: str = "L2-VM",
+    image_name: str = HYPERV_NESTED_VM_IMAGE_NAME,
+    switch_name: str = "nestedvmswitch",
+    nat_name: str = "nestedvmnat",
+) -> None:
+    image_name = f"{name}_{image_name}"
+    image_folder_path = "C:\\Users\\lisatest"
```
squirrelsc: The user name may be changed, it's not reliable. Try to create a folder under system drive.

lisa/tools/mdadm.py
dsrivastavv: Pool is basically a set of disks.  We first create pool and then setup raid0 enabled disk from it

lisa/tools/powershell.py
dsrivastavv: Something like `powershell "echo \"hello\""` worked fine for me in command prompt

lisa/tools/mdadm.py
squirrelsc: Which command creates the raid disk? I cannot find command specified raid.

lisa/tools/powershell.py
squirrelsc: The Powershell escape is complex, it depends on what results it needs. We can see if there is error, how to fix it in future. the `\"` just remove the quote, it may not be what needs.

```powershell
powershell "echo \"hello\""
```
returns
```cmd
hello
```

```powershell
powershell "echo hel`\"lo"
```
returns
```cmd
hel"lo
```

lisa/tools/mdadm.py
dsrivastavv: `New-VirtualDisk` is the function name. The entire logic as defined in v2 is here: https://github.com/microsoft/lisa/blob/24061a4288a02bd223beb878344d481bc0d77e3e/Testscripts/Windows/NESTED-HYPERV-STORAGE-PERF.ps1#L47-L66

lisa/tools/powershell.py
dsrivastavv: Sure! This seems more reliable escape character. I will make the change

lisa/tools/mdadm.py
squirrelsc: The Windows Storage pool is not the RAID technology, but they solve similar problems. Please check the test logic, if the storage pool is ok for the test logic.

dsrivastavv: We need to add parameter `-NumberOfColumns 2` to strip it across two disks. I will add it in the next commit

vyadavmsft:
Can we get description of PR what is trying to be done ? its a good practice ! how are we adding nested vm testcase i thought there is no hyper-v support ?

dsrivastavv:
> Can we get description of PR what is trying to be done ? its a good practice ! how are we adding nested vm testcase I thought there is no hyper-v support ?

This PR is migrating some hyper-v nested testcases. Basically, the L1 VM is windows and L2-VM is a linux distro. This PR is adding hyperv as a tool, so we can spin up L2 VMs. I added testcase name to the info but please feel free to look at each commit name for a better understanding.

However, we still do not have support for hyper-v platform. These tests are meant for azure and we use hyper-v as a tool

squirrelsc (approved):

PR #1860, opened 2022-04-05
QEMU: Add watchdog over VM delete.

cwize1:
There seems to be a rare issue where the process deadlocks while trying to delete
a QEMU VM. This issue is rare enough that getting a repo on a dev box seems to be
very difficult.

This change adds a 60 second watchdog timer over the VM delete operation. Usually,
a VM delete takes less than 1 second. So, this should be plenty to avoid false
positives. When the watchdog fires, all the Python threads' callstacks are printed
and then the process is hard stopped. This will hopefully provide insight into
why this issue is occurring.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -1,14 +1,17 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import faulthandler
 import io
 import os
 import random
 import string
+import sys
 import tempfile
 import time
 import xml.etree.ElementTree as ET  # noqa: N817
 from pathlib import Path
+from threading import Timer
 from typing import Any, List, Optional, Tuple, Type, cast
 
 import libvirt  # type: ignore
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -503,6 +510,9 @@ def _stop_and_delete_vm_flags(
     ) -> None:
         node_context = get_node_context(node)
 
+        watchdog = Timer(60.0, self._watchdog_callback)
```
squirrelsc, line 513: Does it work like timeout? LISA uses the `func_timeout` package to manage timeout. Search for `func_set_timeout` and `FunctionTimeOut` about how it's used.

cwize1: I doubt the problem I am seeing can be recovered from, due to the use of shared/global resources in libvirt. Also, I want to see the callstacks of all threads, in case it reveals the cause of the deadlock. So, using `func_timeout` probably doesn't make sense here.

romoh (approved):

squirrelsc (approved):

PR #1862, opened 2022-04-06
add perf tests for huge vm

mcgov:
Creates new test 'perf_dpdk_[pmd]_multicore_hugevm' that has a core count requirement of at least 48 cores.
Modifies the other previous tests to require 4 or 8 cores depending on their type. 
dual_core is supposed to be minimal core count so 4 is the absolute minimum. 
multicore and multiqueue require at least 8 (minimum recommended for dpdk)

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkperf.py
+++ b/microsoft/testsuites/dpdk/dpdkperf.py
@@ -47,7 +47,7 @@ class DpdkPerformance(TestSuite):
         log: Logger,
         variables: Dict[str, Any],
     ) -> None:
-        self._validate_core_counts_are_equal(environment)
+
         self._run_dpdk_perf_test("failsafe", environment, log, variables)
 
     @TestCaseMetadata(
@@ -69,10 +69,8 @@ class DpdkPerformance(TestSuite):
         variables: Dict[str, Any],
     ) -> None:
 
-        core_count = self._validate_core_counts_are_equal(environment)
-
         self._run_dpdk_perf_test(
-            "failsafe", environment, log, variables, use_cores=core_count
+            "failsafe", environment, log, variables, use_max_cores=True
         )
 
     @TestCaseMetadata(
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -146,13 +202,43 @@ def perf_dpdk_netvsc_pmd_multi_core(
...
+        ),
+    )
+    def perf_dpdk_netvsc_pmd_multi_core_huge_vm(
+        self,
+        environment: Environment,
+        log: Logger,
+        variables: Dict[str, Any],
+    ) -> None:
+
+        core_counts = [
```
squirrelsc: Add a method to share this logic like `validate_cores_equal`

microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -164,6 +250,25 @@ def perf_dpdk_netvsc_pmd_multi_queue(
...
+        ),
+    )
+    def perf_dpdk_netvsc_pmd_multi_queue_huge_vm(
+        self,
+        environment: Environment,
+        log: Logger,
+        variables: Dict[str, Any],
+    ) -> None:
+
+        self._run_dpdk_perf_test("netvsc", environment, log, variables, use_queues=True)
```
squirrelsc, line 236: Why some cases check the core count, but some of them not?

squirrelsc: If it's needed, can this check move into `_run_dpdk_perf_test`?

mcgov: There's a default number of cores for the tests that line up with the minimum required cores for those tests. For the test that use larger amounts of cores the 'run concurrent' function takes in a single argument and assumes the VMs are the same size. I wanted to double check we aren't passing a core argument that DPDK can't fulfill on one of the VMs. Honestly now that we have the minimum core count argument on the tests themselves this check is probably redundant.

squirrelsc: For the statement below, it looks like "dpdk expects" all core should be the same. So I don't understand why not check the core count in `_run_dpdk_perf_test`, but only check in some of test cases.


Nodes contain different core counts, DPDK Suite expects sender and receiver to have same core count.

mcgov: scrooched it to the run function

squirrelsc (approved):

PR #1863, opened 2022-04-06
Reboot VM after tasksmax change to make sure it takes effect

LiliDeng:


lisa/tools/ntttcp.py
```diff
@@ -431,6 +431,7 @@ def _install(self) -> bool:
         return self._check_exists()
 
     def _set_tasks_max(self) -> None:
+        set = False
```
squirrelsc: `need_reboot`

squirrelsc (approved):

PR #1864, opened 2022-04-06
Set run_infinite as False when run time seconds is set.

LiliDeng:


lisa/operating_system.py
```diff
@@ -705,13 +706,15 @@ def add_repository(self, repo: str, key_location: str = "") -> None:
             expected_exit_code_failure_message="fail to add repository",
         )
 
+    @retry(tries=60, delay=5)  # type: ignore
```
squirrelsc: Let's try 10 times firstly, 60 times not only 60 * 5. Because the install command also takes long time. If it's too long, it's too long when troubleshooting.

lisa/tools/iperf3.py
```diff
@@ -222,10 +222,11 @@ def run_as_client_async(
             f"{self.command} {cmd}", shell=True, sudo=True
         )
 
-        # IPerf output emits lines of the following form when it is running
-        # 132.00-133.00 sec   167 MBytes  1.40 Gbits/sec    5    626 KBytes
-        # check if stdout buffers contain "bits/sec" to determine if running
-        process.wait_output("bits/sec")
+        if not output_json and not log_file:
```
squirrelsc: Add comments to explain this logic. Why it doesn't need to wait, if the two conditions met.

lisa/tools/iperf3.py
```diff
@@ -259,10 +261,12 @@ def run_as_client(
             client_ip,
             ip_version,
             udp_mode,
+            run_infinite,
         )
         return process.wait_result(
             expected_exit_code=0,
             expected_exit_code_failure_message="fail to launch iperf3 client",
+            timeout=run_time_seconds + 1,
```
squirrelsc: Many app use timeout=0 or -1 to control if it needs to wait or not. It needs only one variable to control the timeout. And it's weird to +1 here. Let's change to use 0 as infinite, and change the run_as_client_async to convert "0" to "inf".

LiliDeng: infinite may cause the whole running enters into dead loop?

lisa/tools/ntttcp.py
```diff
@@ -201,6 +201,7 @@ def run_as_server_async(
         process = self.node.execute_async(
             f"ulimit -n 204800 && {self.command} {cmd}", shell=True, sudo=True
         )
+        process.wait_output("threads created")
```
squirrelsc: Let's use one PR for each fix. So it will be merged earlier.

lisa/tools/iperf3.py
squirrelsc: That should depend on how it's called, and the caller should fix it. If there are two arguments to control the same thing, it's easy to set the conflict variables, and get unexpected result.

lisa/operating_system.py
LiliDeng: split it into another PR

lisa/tools/ntttcp.py
LiliDeng: split it into another PR

lisa/tools/iperf3.py

lisa/tools/iperf3.py
LiliDeng: with iperf3 run with infinite, won't wait for the result, I update it into run_time_seconds

lisa/tools/iperf3.py
```diff
@@ -222,10 +224,22 @@ def run_as_client_async(
...
-        # 132.00-133.00 sec   167 MBytes  1.40 Gbits/sec    5    626 KBytes
-        # check if stdout buffers contain "bits/sec" to determine if running
-        process.wait_output("bits/sec")
+        if log_file:
+            timeout = 60
+            timer = create_timer()
+            while timeout > timer.elapsed(False):
+                cat = self.node.tools[Cat]
+                iperf_log = cat.read(log_file, sudo=True, force_run=True)
+                if "Connecting to host " in iperf_log:
```
squirrelsc, line 235: Can it wait for wait_output with a timeout?

lisa/tools/iperf3.py
squirrelsc: If you check all place with infinite=False, the run_time_seconds are not 0. So it's safe to use 0 to present "infinite".

lisa/tools/iperf3.py
LiliDeng: while timeout > timer.elapsed(False): is a timeout?

squirrelsc: Is there any output to console to check?

LiliDeng: when specify the log_file, all output write into the log_file

squirrelsc: I see.

1. Add sleep(1) to prevent it call too frequently.
2. Add argument "no_debug_log" to cat.read not to output to console to reduce useless log.

lisa/tools/iperf3.py
LiliDeng: in run_as_client, infinite set as False

squirrelsc: In the run_as_client, the run seconds are set to 10 seconds, it's the same as "infinite==false".

LiliDeng: current logic is 
```
        # set runtime to infinite if run_infinite is True
        if run_infinite:
            run_time = "inf"
        else:
            run_time = str(run_time_seconds)
```

lisa/tools/iperf3.py
LiliDeng: update

lisa/tools/iperf3.py
squirrelsc: The condition can be changed to `if run_time_seconds<= 0:`

squirrelsc (approved):

lisa/tools/iperf3.py
LiliDeng: how about this comments, I remove run_infinite parameter, set run_time_seconds as 0, when run_time_seconds is not passed or passed into a negative value, use run infinite.

PR #1870, opened 2022-04-07
Fix resize case, currently resize case always fail for no capability.

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1573,6 +1573,7 @@ def _resource_sku_to_capability(  # noqa: C901
                 schema.FeatureSettings.create(features.StartStop.name()),
                 schema.FeatureSettings.create(features.SerialConsole.name()),
                 schema.FeatureSettings.create(features.Hibernation.name()),
+                schema.FeatureSettings.create(features.Resize.name()),
```
squirrelsc, line 1593: My understanding is not all vm size can be resize. Is there a list of it?

LiliDeng: probably no, but when resize a vm, I know there is an api to know the vm sizes list which can be resized based on current vm size. if no sizes listed, then can skip the case, I will check.

microsoft/testsuites/core/vm_hot_resize.py
```diff
@@ -37,14 +39,30 @@ class VmHotResize(TestSuite):
...
+                expected_vm_capability = None
+                expected_vm_capability = resize.resize()
+                break
+            except Exception as e:
+                if "cannot find current vm size in eligible list" in str(e):
+                    retry = retry + 1
+                elif "OperationNotAllowed" in str(e):
+                    retry = retry + 1
+                else:
+                    log.debug("unexpected exception found")
```
squirrelsc: For unknown exception, use `raise identifier` to reraise it.

microsoft/testsuites/core/vm_hot_resize.py
```diff
@@ -37,14 +39,30 @@ class VmHotResize(TestSuite):
...
         resize = node.features[Resize]
-        expected_vm_capability = resize.resize()
+        retry = 1
+        maxretry = 10
+        while retry < maxretry:
+            try:
+                expected_vm_capability = None
+                expected_vm_capability = resize.resize()
+                break
+            except Exception as e:
```
squirrelsc: In other places, we use `identifier` as the name.

microsoft/testsuites/core/vm_hot_resize.py
```diff
@@ -37,14 +39,30 @@ class VmHotResize(TestSuite):
...
     )
-    def verify_vm_hot_resize(self, node: Node) -> None:
+    def verify_vm_hot_resize(self, log: Logger, node: Node) -> None:
         resize = node.features[Resize]
-        expected_vm_capability = resize.resize()
+        retry = 1
+        maxretry = 10
+        while retry < maxretry:
+            try:
+                expected_vm_capability = None
```
squirrelsc: It's not easy to understand.

```python
expected_vm_capability: Optional[schema.NodeSpace] = None
try:
    expected_vm_capability = resize.resize()
```

microsoft/testsuites/core/vm_hot_resize.py
LiliDeng: update

microsoft/testsuites/core/vm_hot_resize.py
LiliDeng: update

microsoft/testsuites/core/vm_hot_resize.py
LiliDeng: update

squirrelsc (approved):

PR #1873, opened 2022-04-07
Fix bugs on Windows, and some refactoring.

squirrelsc:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1123,6 +1130,11 @@ def _create_node_runbook(
             azure_node_runbook.name = truncate_keep_prefix(
                 f"{name_prefix}-n{index}", 50
             )
+            # It's used as computer name only. Windows doesn't support name more
+            # than 15 chars
+            azure_node_runbook.short_name = truncate_keep_prefix(
+                azure_node_runbook.name, 15
```
dsrivastavv, line 1136: Will this name be different for each VM in a resource group?

dsrivastavv: Seems to work fine

dsrivastavv (approved):

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Yes, it takes last part, it contains n0, n1, different from every node.

PR #1878, opened 2022-04-09
Fix verify_deployment_provision_ephemeral_managed_disk

LiliDeng:
Current case failed for below error, I try 4 cores, the same failure.

```
Provisioning.verify_deployment_provision_ephemeral_managed_disk: FAILED   deployment failed. LisaException: NotSupported: OS disk of Ephemeral VM with size greater than 14 GB is not allowed for VM size Standard_DS2_v2 when the DiffDiskPlacement is ResourceDisk. Please refer to https://aka.ms/Ephemeral for more details.
```

### DIFF
--- a/microsoft/testsuites/core/provisioning.py
+++ b/microsoft/testsuites/core/provisioning.py
@@ -112,6 +112,7 @@ class Provisioning(TestSuite):
         priority=1,
         requirement=simple_requirement(
             environment_status=EnvironmentStatus.Deployed,
+            min_core_count=8,
             disk=DiskEphemeral(),
             supported_features=[SerialConsole],
         ),


microsoft/testsuites/core/provisioning.py
```diff
@@ -112,6 +112,7 @@ def verify_deployment_provision_standard_ssd_disk(
         priority=1,
         requirement=simple_requirement(
             environment_status=EnvironmentStatus.Deployed,
+            min_core_count=8,
```
squirrelsc, line 115: The error message is not related to core count? What's the root cause?

LiliDeng: related with os disk and temp storage of using vm size

squirrelsc: Please check if there is another information about the storage size. The core count is not accurate to control the storage size, and it's misleading.

LiliDeng: we can find the info from vmsize, but need figure out how to get the os disk size.

PR #1879, opened 2022-04-11
Add SecurityProfile feature and TVM cases

LiliDeng:


lisa/operating_system.py
```diff
@@ -683,19 +683,20 @@ def get_repositories(self) -> List[RepositoryInfo]:
...
-                url=key_location,
-                file_path=str(self._node.working_path),
-            )
-            self._node.execute(
-                cmd=f"apt-key add {key_file_path}",
-                sudo=True,
-                expected_exit_code=0,
-                expected_exit_code_failure_message="fail to add apt key",
-            )
+    def add_repository(self, repo: str, keys_location: List[str] = []) -> None:
```
squirrelsc: Make all method supports the same parameters, so it's easy to be data driven.

lisa/features/security_profile.py
```diff
@@ -0,0 +1,62 @@
...
+
+FEATURE_NAME_SECURITY_PROFILE = "Security_Profile"
+
+
+@dataclass_json()
+@dataclass()
+class SecurityProfileSettings(schema.FeatureSettings):
+    type: str = FEATURE_NAME_SECURITY_PROFILE
+    secure_boot_enabled: bool = False
+    vtpm_enabled: bool = False
```
squirrelsc: The two variables are set to true together always, so it doesn't need to split to two variables.

microsoft/testsuites/tvm/tvmsuite.py
```diff
@@ -0,0 +1,84 @@
...
+                ],
+            )
+        elif isinstance(node.os, Suse):
+            node.os.add_repository(
+                "https://packages.microsoft.com/yumrepos/azurecore/",
+                "packages-microsoft-com-azurecore",
+            )
+        else:
+            raise SkippedException(f"current os {node.os.name} doesn't support tvm")
+        node.os._initialize_package_installation()
```
squirrelsc: This doesn't need to be called here. It should be called in os, if it's needed.

lisa/features/security_profile.py
LiliDeng: I am not 100% sure, since from the portal, we can just enable one of them.

lisa/operating_system.py
LiliDeng: for suse and centos, I didn't see they use keys.

squirrelsc: They can have the same arguments, and not used. So the different os can use the same arg dict like we did for perf notifier.

lisa/features/security_profile.py
squirrelsc: We can put them together firstly. If it's needed, then split it. We're not create the TVM feature, it's for integration testing.

lisa/operating_system.py
```diff
@@ -705,6 +706,7 @@ def add_repository(self, repo: str, key_location: str = "") -> None:
             expected_exit_code=0,
             expected_exit_code_failure_message="fail to add repository",
         )
+        self._initialize_package_installation()
```
squirrelsc: I remember the apt will update repos after add.

lisa/operating_system.py

lisa/features/security_profile.py
LiliDeng: update

lisa/operating_system.py
LiliDeng: tested and removed

squirrelsc (approved):

PR #1880, opened 2022-04-12
Add hyperv_drm testcase

saurabh-sengar:
Signed-off-by: Saurabh Sengar <ssengar@microsoft.com>

### DIFF
--- a/microsoft/testsuites/display/drm.py
+++ b/microsoft/testsuites/display/drm.py
@@ -1,0 +1,74 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from typing import Any
+
+from assertpy import assert_that
+
+from lisa import (
+    Logger,
+    Node,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    simple_requirement,
+)
+from lisa.base_tools import Cat
+from lisa.sut_orchestrator import AZURE, READY
+from lisa.tools import Lsmod, Uname
+from lisa.util import SkippedException
+
+
+@TestSuiteMetadata(
+    area="drm",
+    category="functional",
+    description="""
+    This test suite uses to verify drm driver sanity.
+    """,
+    requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+)
+class Drm(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This case is to check whether the hyperv_drm driver registered successfully.
+        Once driver is registered successfully it should appear in `lsmod` output.
+
+        Steps,
+        1. lsmod
+        2. Check if hyperv_drm exist in the list.
+        """,
+        priority=2,
+        use_new_environment=True,
+    )
+    def verify_drm_driver(self, node: Node, log: Logger) -> None:
+        lsmod = node.tools[Lsmod]
+        assert_that(lsmod.module_exists("hyperv_drm")).described_as(
+            "hyperv_drm module is absent"
+        ).is_equal_to(True)
+
+    @TestCaseMetadata(
+        description="""
+        This case is to check whether the dri node is populated correctly.
+        If hyperv_drm driver is bind correctly it should populate dri node.
+        This dri node can be find at following sysfs entry : /sys/kernel/debug/dri/0.
+        The dri node name (/sys/kernel/debug/dri/0/name) should contain `hyperv_drm`.
+
+        Step,
+        1. Cat /sys/kernel/debug/dri/0/name.
+        2. Verify it contains hyperv_drm string in it.
+        """,
+        priority=2,
+    )
+    def verify_dri_node(self, node: Node, log: Logger) -> None:
+        cat = node.tools[Cat]
+
+        dri_path = "/sys/kernel/debug/dri/0/name"
+        dri_name = cat.read(dri_path, sudo=True, force_run=True)
+        assert_that(dri_name).described_as(
+            "dri node not populated for hyperv_drm"
+        ).matches("hyperv_drm")
+
+    def before_case(self, log: Logger, **kwargs: Any) -> None:
+        node = kwargs["node"]
+        kernel_version = node.tools[Uname].get_linux_information().kernel_version
+        if kernel_version <= "5.13.0":
+            raise SkippedException("DRM hyperv driver is supported from 5.14.x kernel.")


microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,68 @@
...
+)
+class DRM_Sanity(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This case is to check whether the drm driver registered successfully.
+
+        Step,
+        1. Check hyperv_drm driver is inserted correctly.
+        """,
+        priority=0,
```
squirrelsc: By default, all test cases should be P2. There is only one P0 test case, it measures test case can boot up or not. Let me know the impact of the failure, we can decide if it needs to be P1.

microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,68 @@
...
+    """,
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class DRM_Sanity(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This case is to check whether the drm driver registered successfully.
+
+        Step,
+        1. Check hyperv_drm driver is inserted correctly.
```
squirrelsc: Give more details, how it's checked. The test metadata will be abstracted out of the code. It needs to help understand details.

microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,68 @@
...
+    )
+    def verify_drm_driver(self, node: Node, log: Logger) -> None:
+        lsmod = node.tools[Lsmod]
+        if lsmod.module_exists("hyperv_drm"):
+            log.info(f"hyperv_drm module present")
+            ret  = "true"
+        else:
+            log.error(f"hyperv_drm module absent")
+            ret = "false"
+        assert_that(ret).is_equal_to("true")
```
squirrelsc: Give the error message, once it's failed, it should let people know what's expected, and how to fix it.

microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,68 @@
...
+        priority=0,
+    )
+    def verify_dri_node(self, node: Node, log: Logger) -> None:
+        cat = node.tools[Cat]
+
+        dri_path = "/sys/kernel/debug/dri/0/name";
+        dri_name = cat.read(dri_path, sudo=True, force_run=True)
+        assert_that(dri_name).matches("hyperv_drm")
+
+    def before_case(self, log: Logger, **kwargs: Any) -> None:
```
squirrelsc: It doesn't need to be copied into test cases. It's for example purpose.

microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,68 @@
...
+        1. Check hyperv_drm driver is inserted correctly.
+        """,
+        priority=0,
+        use_new_environment=True,
+    )
+    def verify_drm_driver(self, node: Node, log: Logger) -> None:
+        lsmod = node.tools[Lsmod]
+        if lsmod.module_exists("hyperv_drm"):
+            log.info(f"hyperv_drm module present")
+            ret  = "true"
```
squirrelsc: You can assert the output of "lsmod.module_exists", it doesn't need the if/else to set the ret.

squirrelsc:
@saurabh-sengar Thank you for contributing the test case, please read the guideline to understand comments on the PR.
https://mslisa.readthedocs.io/en/main/write_test/guidelines.html#troubleshooting-excellence

microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,68 @@
...
+from lisa.base_tools import Cat
+from lisa.tools import Lsmod
+
+@TestSuiteMetadata(
+    area="drm",
+    category="functional",
+    description="""
+    This test suite uses to verify drm driver sanity.
+    """,
+    requirement=simple_requirement(unsupported_os=[]),
```
squirrelsc: 1. If a test case can be run on Windows, this line is needed. Otherwise, it doesn't.
2. It looks this test case applies to Azure only, please add platform restriction.

microsoft/testsuites/display/drm_sanity.py
saurabh-sengar: will change it to P2

microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,68 @@
...
+)
+class DRM_Sanity(TestSuite):
```
squirrelsc: Follow guideline to name classes.

microsoft/testsuites/display/drm_sanity.py
saurabh-sengar: ok

microsoft/testsuites/display/drm_sanity.py
saurabh-sengar: ok

microsoft/testsuites/display/drm_sanity.py
saurabh-sengar: I see, will correct

microsoft/testsuites/display/drm_sanity.py
saurabh-sengar: thanks, good suggestion

microsoft/testsuites/display/drm_sanity.py
saurabh-sengar: yes this is only for linux, will add restrictions

microsoft/testsuites/display/drm_sanity.py
saurabh-sengar: ok

microsoft/testsuites/display/drm_sanity.py
```diff
@@ -0,0 +1,63 @@
...
+
+@TestSuiteMetadata(
+    area="drm",
+    category="functional",
+    description="""
+    This test suite uses to verify drm driver sanity.
+    """,
+    requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+)
+class drmSanity(TestSuite):
```
squirrelsc: The name can be like `Drm`, it's right for Python. And if there is no more test to added, it doesn't need to add "Sanity" with the name.

microsoft/testsuites/display/drm.py
```diff
@@ -0,0 +1,69 @@
...
+        cat = node.tools[Cat]
+
+        dri_path = "/sys/kernel/debug/dri/0/name";
+        dri_name = cat.read(dri_path, sudo=True, force_run=True)
+        assert_that(dri_name).described_as('dri node not populated for hyperv_drm').matches("hyperv_drm")
+
+    def before_case(self, node: Node, log: Logger, **kwargs: Any) -> None:
+        kernel_version = node.tools[Uname].get_linux_information().kernel_version
+        if kernel_version <= "5.14.0":
+            raise SkippedException("DRM hyperv driver is supported from 5.14.x kernel.")
```
squirrelsc, line 74: If it's supported *from* 5.14, should the condition be `if kernel_version < "5.14.0":`?

saurabh-sengar: I observed that even kernel_version < "5.14.0" is not fully correct, as it skips "5.14.0-rc*" versions which should have supported. I decided to skip 5.14.0 all together for consistency , but revisiting it I think  kernel_version <= "5.13.0" will be the best condition.

squirrelsc (dismissed):

squirrelsc (dismissed):

squirrelsc (approved):

PR #1882, opened 2022-04-13
Resolve image version from "latest" in prepare_environment

dsrivastavv:
Fix issue where TestResult message has "latest" in image name rather than the actual version

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -473,6 +473,14 @@ class AzurePlatform(Platform):
                         f"cap: {environment.runbook.nodes_requirement}"
                     )
                     break
+
+            for req in nodes_requirement:
+                node_runbook = req.get_extended_runbook(AzureNodeSchema, AZURE)
+                if node_runbook.location and node_runbook.marketplace:
+                    # resolve Latest to specified version
+                    node_runbook.marketplace = self._parse_marketplace_image(
+                        node_runbook.location, node_runbook.marketplace
+                    )
         return is_success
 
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -363,6 +363,11 @@ def _prepare_environment(  # noqa: C901
                     else:
                         existing_location = node_runbook.location
 
+                # resolve Latest to specified version
+                node_runbook.marketplace = self._parse_marketplace_image(
```
squirrelsc: It should be called, only if the marketplace is not empty.

dsrivastavv: yes. Just saw the test failures on this. Both location and marketplace need to be specified

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -363,6 +363,12 @@ def _prepare_environment(  # noqa: C901
                     else:
                         existing_location = node_runbook.location
 
+                    if node_runbook.marketplace:
```
squirrelsc: I think it again, it should be put before the last return, and in the `if environment.runbook.nodes_requirement:`. There are some different scenarios. That's the right place to resolve the image name. You can add a new for loop there.

dsrivastavv: Do you mean just before `return is_success` on line 482?

squirrelsc: yes

squirrelsc (approved):

PR #1884, opened 2022-04-14
QEMU: Fix for deadlock on VM delete.

cwize1:
It seems that libvirt can sometimes deadlock if two threads
simultaneously try to close the console log stream and delete
(undefine) the VM. I am not sure if this is intended behavior or not.

This change changes the node shutdown flow, so that the console log is
fully closed before attempting to undefine the VM. Hopefully this will
prevent the deadlock.

The watchdog timer has been kept in case there are other ways this
deadlock can occur.


cwize1:
@anirudhrb You may wish to review.

squirrelsc (approved):

romoh (approved):

anirudhrb (approved):

PR #1889, opened 2022-04-18
Add JUnit notifier.

cwize1:
1. Add timestamp to all notifier messages.

2. Add stacktrace to test result failure messages.

3. On test result messages, report relative function name (in addition to the fully
qualified name), and fully qualified test suite name.

4. Add new notifier messages for test suite status.

5. Add a new notifier type for outputting test results in the JUnit format.

### DIFF
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -15,7 +15,7 @@ if TYPE_CHECKING:
 @dataclass
 class MessageBase:
     type: str = "Base"
-    message_time: datetime = datetime.min
+    time: datetime = datetime.min
     elapsed: float = 0
 
 
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -15,6 +15,7 @@ if TYPE_CHECKING:
 @dataclass
 class MessageBase:
     type: str = "Base"
+    message_time: datetime = datetime.min
     elapsed: float = 0
 
 
@@ -33,6 +34,7 @@ TestRunStatus = Enum(
 class TestRunMessage(MessageBase):
     type: str = "TestRun"
     status: TestRunStatus = TestRunStatus.INITIALIZING
+    runbook_name: str = ""
     test_project: str = ""
     test_pass: str = ""
     tags: Optional[List[str]] = None
--- a/lisa/notifiers/env_stats.py
+++ b/lisa/notifiers/env_stats.py
@@ -82,7 +82,7 @@ class EnvironmentStats(notifier.Notifier):
         result_info = self._test_results.get(test_result.id_, None)
         if not result_info:
             result_info = TestResultInformation(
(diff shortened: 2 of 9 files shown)


lisa/notifiers/env_stats.py
```diff
@@ -82,7 +82,7 @@ def _process_test_result_message(self, test_result: TestResultMessage) -> None:
         result_info = self._test_results.get(test_result.id_, None)
         if not result_info:
             result_info = TestResultInformation(
-                id=test_result.id_, name=test_result.name
+                id=test_result.id_, name=test_result.full_name
```
cwize1, line 85: FYI: `name` was renamed to `full_name` so that `name` could be used for relative name instead of the full name.

lisa/testsuite.py
```diff
@@ -40,6 +41,16 @@
 )
 from lisa.util.perf_timer import Timer, create_timer
 
+TestSuiteStatus = Enum(
```
squirrelsc: It doesn't worth to create suite message in LISA. In the default lisa runner, each test case is scheduled individually. So you can consider the test suite will be start as many as the test case method. So it's status is same as test result status.
If Junit needs this, we can use test result message to simulate it.

lisa/messages.py
```diff
@@ -15,6 +15,7 @@
 @dataclass
 class MessageBase:
     type: str = "Base"
+    message_time: datetime = datetime.min
```
squirrelsc: How about name it `time`? It's in message already.

lisa/notifiers/junit.py
```diff
@@ -0,0 +1,189 @@
...
+    def _received_test_suite(self, message: TestSuiteMessage) -> None:
+        if message.status == TestSuiteStatus.INITIALIZING:
+            self._test_suite_started(message)
+
+        elif message.status == TestSuiteStatus.COMPLETED:
+            self._test_suite_completed(message)
+
+    # Handle a test case message.
+    def _received_test_result(self, message: TestResultMessage) -> None:
+        if message.status in [TestStatus.PASSED, TestStatus.FAILED, TestStatus.SKIPPED]:
```
squirrelsc: use `message.is_completed`

lisa/testsuite.py
```diff
@@ -165,7 +201,7 @@ def set_status(
             self.status = new_status
             if new_status == TestStatus.RUNNING:
                 self._timer = create_timer()
-            self._send_result_message()
+            self._send_result_message(stacktrace)
```
squirrelsc: Can it just set into the test result, instead of pass as a parameter? It's no different situation like `message`.

squirrelsc (approved):

PR #1890, opened 2022-04-19
add optional filter function for callers using git.get_tag

mcgov:
Tag naming conventions vary from project to project, add an optional function parameter to git.get_tags to let callers define which types of tags to pick in a more flexible way. Filter can now intake a function that ingests and returns a list of strings while retaining the version number sorted order from the git tag result.

This commit also adds:
- a nicer format for error messages to bubble up the settings for this call after a failure.
- a use of the new function in DPDK suite which was used to test.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -444,9 +444,12 @@ class DpdkTestpmd(Tool):
                 dir_name=self._dpdk_repo_path_name,
             )
             if not self._dpdk_branch:
+
                 # dpdk stopped using a default branch
                 # if a branch is not specified, get latest version tag.
-                self._dpdk_branch = git_tool.get_tag(self.dpdk_path)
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -444,9 +444,15 @@ def _install(self) -> bool:
                 dir_name=self._dpdk_repo_path_name,
             )
             if not self._dpdk_branch:
+
+                def filter_for_version_tags(tags: List[str]) -> List[str]:
+                    return [x for x in tags if x.startswith("v")]
```
squirrelsc: Can it just pass in a regexp?

squirrelsc: Sorry, I think the string of regexp is simpler. So the caller just need to pass in the string, not `re.compile(...`

squirrelsc (approved):

PR #1894, opened 2022-04-20
Make env bad if tcp connection is failed, and other fixes.

squirrelsc:


lisa/testsuite.py
```diff
@@ -146,7 +147,9 @@ def handle_exception(
                 TestStatus.PASSED,
                 f"{phase}warning: {exception}",
             )
-        elif isinstance(exception, BadEnvironmentStateException):
+        elif isinstance(exception, BadEnvironmentStateException) or isinstance(
```
sharsonia, line 150: so we would still mark test case as FAILED with line 154?

squirrelsc: Yes, only SkippedException and PassedException should make test case skipped or passed. Otherwise, it should be failed.

sharsonia: Ok but actually we never got a chance to try to run test case itself. I think we should have one exception which can differentiate real test failures from the bad environment failures. Then it will be easier to project correct data related to test failures too. Just my thought

squirrelsc: Because we reuse the environment to run test cases, it's hard to determine an environment is in good state before starting a new test case. The bad environment is one of approach to filter out obvious bad environments. In other cases, we have to make sure every test case recover environment or mark an environment is dirty. Unless the environment is not reused, it's hard to make sure there is no side effects among them.

sharsonia: ok, what I am trying to say is that once we figured that it is the environment which is a problem then the test status can be marked as "ATTEMPTED" or some new category that suggest that the test encountered bad environment.
This way we can clearly distinguish a real failed test scenario from the bad environment issues

squirrelsc: That may hide problems from previous test cases. Once they are marked as Attempted, that means it's a known issue.

sharsonia: Well then lets have a new category for these kind of issues and mark status as that instead of ATTEMPTED.
The point is bad environment shouldn't reflect a "FAILED" test otherwise we are amplifying the dashboards/metrics for real test case failures.
Issues with environment should be looked upon/investigated differently then the test case issues

squirrelsc: The FAILED is used to track all unknown issues, it includes test case failure (current or previous), test framework failures. Our goal is to know all failures and triage them (fix or mark it attempted). For example, if we "know" how an environment is broken, we can fix it by mark it dirty to prevent it's used by next test case.
It's not practical to create another type of failure to track failures from previous test cases. Firstly, it means each test case needs to check preconditions to make sure it works (we should do, but not enforced today). Secondly, we need to create another type of failure, but it follows the same triage process.

lisa/testsuite.py
```diff
@@ -146,7 +147,9 @@ def handle_exception(
...
+            exception, TcpConnetionException
```
dsrivastavv: [nit] *TcpConne[c]tionException

sharsonia: Why separate commit to fix the the typo when the typo was introduced as part of commit which is not yet merged?
Just trying to understand why we need extra commit and not resolve the issue in the commit introducing the problem itself?

sharsonia: > Why separate commit to fix the the typo when the typo was introduced as part of commit which is not yet merged? Just trying to understand why we need extra commit and not resolve the issue in the commit introducing the problem itself?

oh nvm, I see there are multiple places in the original code itself that have the typo

sharsonia (commented):
[nit]There is a typo in the commit header (smot[*k]e_test).

sharsonia (approved):

PR #1896, opened 2022-04-20
[tools][modinfo]: Fix filename pattern match

sharsonia:


sharsonia:
[lisa-20220420-204904-753.log](https://github.com/microsoft/lisa/files/8525594/lisa-20220420-204904-753.log)

squirrelsc (approved):

PR #1900, opened 2022-04-21
Skip lis tests when lis driver doesn't support install on current distro

LiliDeng:


lisa/features/gpu.py
```diff
@@ -122,8 +122,7 @@ def install_compute_sdk(self, version: str = "") -> None:
         self._install_gpu_dep()
         try:
             # install LIS driver if required and not already installed.
-            if LisDriver.can_install:
-                self._node.tools[LisDriver]
+            self._node.tools[LisDriver]
         except Exception as identifier:
             self._log.debug(
                 "LisDriver is not installed. It might not be required. " f"{identifier}"
```
squirrelsc: remove the `" f"`. It doesn't need in one line.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -187,12 +192,14 @@ def command(self) -> str:
 
     @property
     def can_install(self) -> bool:
+        return True
+
+    def is_support(self) -> bool:
```
squirrelsc: Why not use `can_install`?

lisa/features/gpu.py
```diff
@@ -122,8 +122,7 @@ def install_compute_sdk(self, version: str = "") -> None:
         self._install_gpu_dep()
         try:
             # install LIS driver if required and not already installed.
-            if LisDriver.can_install:
```
squirrelsc, line 125: The `can_install` is called by LISA, it doesn't need to be called explicitly.

lisa/sut_orchestrator/azure/tools.py
LiliDeng: it returns an object, always return true

lisa/features/gpu.py
LiliDeng: hmm, yes, let me retry

lisa/features/gpu.py
LiliDeng: removed

lisa/features/gpu.py
LiliDeng: updated

LiliDeng:
either we skip test here, or add a failure type in db

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -192,8 +197,10 @@ def can_install(self) -> bool:
             and self.node.os.information.version < "7.8.0"
         ):
             return True
-
-        return False
+        raise SkippedException(
```
squirrelsc: Don't skip test case in the tool, catch it and use Skipped in test case.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -121,97 +121,90 @@ def _verify_lis_preinstall_disk_size(
...
-                    )
-                    fallocate.create_file(
-                        single_partition_file_size, "/lib/modules/file.out"
-                    )
-                if boot_partition != root_partition:
-                    single_partition_file_size_boot = (
-                        boot_part_avail_space - boot_part_required_space
+                fallocate.create_file(single_partition_file_size_boot, "/boot/file.out")
+            result = lisdriver.install_from_iso()
+            if test_type == "negative":
```
squirrelsc, line 198: `if test_type == "negative" and result.exit_code == 0:`

microsoft/testsuites/lis/lissuite.py
```diff
@@ -121,97 +121,90 @@ def _verify_lis_preinstall_disk_size(
...
-                self._clean_up_files(node)
-                if test_type == "positive":
-                    lisdriver.uninstall_from_iso()
-            else:
-                raise SkippedException(
-                    f"current distro {node.os.name} doesn't support install LIS"
-                )
+                else:
+                    log.debug(f"fail to install lis for reason {result.stdout}")
+            self._clean_up_files(node)
```
squirrelsc: This line may not be called, if exception raised above. call it in after_case

lisa/sut_orchestrator/azure/tools.py

microsoft/testsuites/lis/lissuite.py
LiliDeng: updated

squirrelsc (approved):

PR #1901, opened 2022-04-21
Fix sshconnection error coming from paramiko exec_command

dsrivastavv:
paramiko_client is also throwing SSHException if the tcp ports are not ready. Putting it inside retry makes logic robust

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -150,7 +150,6 @@ def try_connect(connection_info: ConnectionInfo, ssh_timeout: int = 300) -> Any:
 
     # wait for ssh port to be ready
     timeout_start = time.time()
-    is_ready = False
     while time.time() < timeout_start + ssh_timeout:
         try:
             paramiko_client.connect(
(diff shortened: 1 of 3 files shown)


lisa/util/shell.py
```diff
@@ -161,42 +160,37 @@ def try_connect(connection_info: ConnectionInfo, ssh_timeout: int = 300) -> Any:
                 key_filename=connection_info.private_key_file,
                 banner_timeout=10,
             )
+
+            stdin, stdout, _ = paramiko_client.exec_command("cmd\n")
```
squirrelsc, line 164: While line do you see the exception actually being? The paramiko_client is initialized in the loop. I think the key thing is to capture the `ConnectionResetError` exception. Is it right?

squirrelsc: This method is to use low level API (paramiko) to check the os type, and then use SpurShell to connect with the right shell type (windows or linux).

dsrivastavv: This `paramiko_client.exec_command("cmd\n")` causes exceptions. I have seen it throwing both `SSHException` and `ConnectionResetError` error

squirrelsc: ok, we can go with this change, please rebase to latest main. Next time, you can create a new branch from main, and cherry pick it from original branch.

dsrivastavv: Sure! Let me verify it for few other nested testcases before merging this

squirrelsc (dismissed):

squirrelsc:
The other two commits should be in another PR. Once a PR is approved, it's better to get it in, instead of adding new changes continuously.

squirrelsc (approved):

dsrivastavv:
> The other two commits should be in another PR. Once a PR is approved, it's better to get it in, instead of adding new changes continuously.

Sorry this was work in progress. I had marked it in the PR title. Will open a draft PR next time onwards

PR #1903, opened 2022-04-23
Fix repo not exist for focal, impish.

LiliDeng:


microsoft/testsuites/tvm/tvmsuite.py
```diff
@@ -88,6 +88,10 @@ def _add_azure_core_repo(self, node: Node) -> None:
                 repo_url = "https://packages.microsoft.com/repos/azurecore-debian"
             else:
                 codename = node.os.information.codename
+                # there is no available repo for distro which higher than bionic, use
+                # bionic for temp solution
+                if codename not in ["trusty", "xenial", "bionic"]:
```
squirrelsc: Use the version number to check, and then get codename from the version number. It's easier to read.

squirrelsc (approved):

PR #1904, opened 2022-04-23
Fix verify_disk_with_nobarrier.

LiliDeng:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -710,7 +710,7 @@ def get_raw_data_disks(self) -> List[str]:
         pattern = re.compile(r"/dev/disk/azure/scsi[0-9]/lun[0-9][0-9]?", re.M)
         # /dev/disk/azure/scsi1/lun0
         cmd_result = self._node.execute(
-            "ls -d /dev/disk/azure/scsi*/*", shell=True, sudo=True
+            "ls -d /dev/disk/azure/scsi1/*", shell=True, sudo=True
```
squirrelsc, line 715: What's problem with original pattern? I saw Divyansh just changed the regexp few days ago. This change includes scsi0 as well. Can you update code comments to explain why filter only scsi1, not scsi*?

LiliDeng: in some distro, the os and temp disk display under /dev/disk/azure/scsi0, for data disks, they are under /dev/disk/azure/scsi1

squirrelsc: I think the change from Divyansh is to include scsi0, but with this change, scsi0 is excluded. How to handle it?

LiliDeng: @somil55 hi, why you include /dev/disk/azure/scsi0?

dsrivastavv: There were disks which followed scsi0 indexing as well.

LiliDeng: I found the disks under scsi0 are os and temp disk, do you have an image which has data disks under /dev/disk/azure/scsi0

dsrivastavv: Sorry I don't remember the image and VM size explicitly.

LiliDeng: @squirrelsc @somil55 this method is to get the data disks from /dev/disk/azure/scsi* folder
refer
https://docs.microsoft.com/en-us/troubleshoot/azure/virtual-machines/troubleshoot-device-names-problems#identify-disk-luns

lisa/tools/mkfs.py
```diff
@@ -52,8 +52,8 @@ def format_disk(self, disk_name: str, file_system: FileSystem) -> None:
             raise LisaException(f"Unrecognized file system {file_system}.")
 
     def _install(self) -> bool:
-        self.node.tools[Mkfsxfs]
-        return self._check_exists()
+        # the installation is completed in format_disk based on file_system
+        return True
```
squirrelsc, line 56: Can it return `self._check_exists()`?

squirrelsc: The commit comment isn't accurate as well. It says "add missing install method". Here is not add something, it deletes logics.

LiliDeng: previously we don't have _install method, yesterday Diviyansh's PR was in which contains this method, I think it should return true, and the installation is completed in format_disk method.

squirrelsc (approved):

PR #1905, opened 2022-04-24
Fix verify_vmbus_heartbeat_properties

LiliDeng:


lisa/tools/python.py
```diff
@@ -18,6 +18,18 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
+    def create_symblink(self) -> None:
```
squirrelsc, line 21: This should be in ln tool, not in Python. If it's needed in Python, it should be created in `_initialize` of Python.

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -124,7 +124,7 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
             supported_platform_type=[AZURE],
         ),
     )
-    def verify_vmbus_heartbeat_properties(self, node: Node) -> None:
+    def verify_vmbus_heartbeat_properties(self, node: Node) -> None:  # noqa: C901
```
squirrelsc, line 127: try to abstract a method, not overuse noqa on methods.

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -182,54 +180,78 @@ def verify_vmbus_heartbeat_properties(self, node: Node) -> None:
                 "client_monitor_latency",
                 "client_monitor_pending",
             ]
-            vmbus_channel_files += ["latency", "pending"]
+            vmbus_channel_files += ["pending"]
         for file in vmbus_driver_files:
             if not ls.path_exists(f"{device_path}/{file}", sudo=True):
                 raise LisaException(f"{device_path}/{file} doesn't exist")
+        # for old distro, there is no channels folder, so skip testing
```
squirrelsc, line 187: Can we filter old distros by distro versions? It may mischecks on new distros, if it doesnt' have the channel folder by a bug.

Issue #1907, opened 2022-04-25
lisa run getting exception in synthetic_add_max_nics_one_by_one_after_provision_validation test case


souradeep100:
While executing  getting following exception:

```
[...]

PR #1908, opened 2022-04-25
Fix network test cases issues.

LiliDeng:
Include fix for #1907

### DIFF
--- a/lisa/tools/iperf3.py
+++ b/lisa/tools/iperf3.py
@@ -75,6 +75,10 @@ class Iperf3(Tool):
         r" Gbits/sec.*receiver([\w\W]*?))",
         re.MULTILINE,
     )
+    _json_pattern = re.compile(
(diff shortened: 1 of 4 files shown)


microsoft/testsuites/network/synthetic.py
```diff
@@ -39,7 +39,7 @@ class Synthetic(TestSuite):
         """,
         priority=2,
         requirement=simple_requirement(
-            min_nic_count=8,
+            nic_count=8,
```
squirrelsc: why not use min_nic_count?

LiliDeng: hmm, seems I misunderstand the issue #1907, will push another fix

squirrelsc (dismissed):

raghavthind2005 (dismissed):

squirrelsc (approved):

PR #1909, opened 2022-04-25
DPDK: use package manager by default

mcgov:
Swap back to using the package manager as the default installation source for DPDK for basic functionality testing.

For our broad validation tests we should be using the package manager to pull in matching dependencies and a certified version of DPDK rather than attempting to special case each version and distro.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -23,8 +23,6 @@ from lisa.util import perf_timer
 from lisa.util.parallel import TaskManager, run_in_parallel, run_in_parallel_async
 from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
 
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -186,7 +186,7 @@ def initialize_node_resources(
     pmd: str,
     sample_apps: Union[List[str], None] = None,
 ) -> DpdkTestResources:
-    dpdk_source = variables.get("dpdk_source", DPDK_STABLE_GIT)
+    dpdk_source = variables.get("dpdk_source", "package_manager")
```
squirrelsc, line 187: It looks there are two DPDK_STABLE_GIT, but not be used any more. Can they be removed?

mcgov: indeed they can! will remove

squirrelsc (approved):

PR #1911, opened 2022-04-25
add 'backports' option to install_packages

mcgov:
Debian based systems require backported packages to be installed explicitly, add an option to install_packages to select the backport repository as the source. 

On RPM based distros this will be a no-op, they provide backports as part of the regular repositories, though it is probably possible to filter for a backport based on their version number on RHEL.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -298,9 +298,10 @@ class Posix(OperatingSystem, BaseClassMixin):
         packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]],
         signed: bool = True,
         timeout: int = 600,
+        backports: bool = False,
     ) -> None:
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -298,9 +298,10 @@ def install_packages(
         packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]],
         signed: bool = True,
         timeout: int = 600,
+        backports: bool = False,
```
squirrelsc: It looks the backports applies to Debian only. Please rename it to `extra` or `additional_arguments` and support it in all os.

mcgov:
testing after changes currently, hold merge

mcgov:
tests look good, opening for review

squirrelsc (approved):

PR #1913, opened 2022-04-26
Add modprobe module_exists function

mcgov:
Add 'check if module exists' function to modprobe.

-Modprobe allows you to combine --dry-run and -q to check if a module exists and will load on the system. This commit enables use of this feature by adding a 'dry run' boolean and provides a named function to check if a module exists on the system.

-Adding this because testing manual binding for ib drivers is super annoying when trying to build from source and run DPDK on every distro and version. Rather than exploding or special casing every driver we can check if one exists and load it only if it's present (and needs to be loaded).

### DIFF
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -63,18 +63,33 @@ class Modprobe(Tool):
     def load(
         self,
         modules: Union[str, List[str]],
-    ) -> None:
+        dry_run: bool = False,
+    ) -> bool:
+
         if isinstance(modules, list):
             modules_str = "-a " + " ".join(modules)
         else:
             modules_str = modules
-        self.run(
-            f"{modules_str}",
+        command = f"{modules_str}"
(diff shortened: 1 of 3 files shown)


lisa/tools/modprobe.py
```diff
@@ -63,18 +63,32 @@ def remove(self, mod_names: List[str], ignore_error: bool = False) -> None:
...
             force_run=True,
             sudo=True,
+        )
+        if dry_run:
+            return result.exit_code == 0
+
+        result.assert_exit_code(
             expected_exit_code=0,
-            expected_exit_code_failure_message=f"Fail to load module[s]: {modules_str}",
+            message=f"Fail to load module[s]: {modules_str}",
```
squirrelsc: Is it worth to output the detail failure message? You can add an argument on `assert_exit_code` to specify if print raw output like `with_output`.

squirrelsc (approved):

PR #1915, opened 2022-04-27
FIX: update rhel regex to work with better optional patch numbers

mcgov:


lisa/operating_system.py
```diff
@@ -1006,12 +1006,12 @@ class RPMDistro(Linux):
     # microsoft-azure-rhel8-eus  Microsoft Azure RPMs for RHEL8 Extended Update Support
     _rpm_repository_info_pattern = re.compile(r"(?P<id>\S+)\s+(?P<name>\S.*\S)\s*")
 
-    # ex: dpdk-20.11-3.el8.x86_64
+    # ex: dpdk-20.11-3.el8.x86_64 or dpdk-18.11.8-1.el7_8.x86_64
```
squirrelsc, line 1009: It's better to put two lines, so we know they are two examples.

squirrelsc (approved):

PR #1917, opened 2022-04-27
Fix regression issue when some vm sizes don't have vCPUsAvailable capability.

LiliDeng:


lisa/features/gpu.py
```diff
@@ -98,7 +98,11 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
         settings = cast(GpuSettings, kwargs.get("settings"))
 
         # default to install by platform, unless it's specified to skip.
-        if (settings.is_enabled) and settings.install_by_platform:
+        if (
```
squirrelsc: Drop the `is_enabled` from the PR. I'm looking for another fix in Azure platform. So it doesn't need to add in each feature.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1506,14 +1506,11 @@ def _resource_sku_to_capability(  # noqa: C901
         ](is_allow_set=True, items=[])
         for sku_capability in resource_sku.capabilities:
             name = sku_capability.name
+            vcpus = 0
             if name == "vCPUsAvailable":
                 node_space.core_count = int(sku_capability.value)
```
squirrelsc: If some VMs don't have the `vCPUsAvailable`, how the `node_space.core_count` set correctly? Should it be set by vCPUs?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1506,14 +1506,11 @@ def _resource_sku_to_capability(  # noqa: C901
...
-                if resource_sku.family in ["standardLSv2Family"]:
-                    # refer https://docs.microsoft.com/en-us/azure/virtual-machines/lsv2-series # noqa: E501
-                    # NVMe disk count = vCPU / 8
-                    nvme = NvmeSettings()
-                    nvme.disk_count = int(node_space.core_count / 8)
-                    node_space.features.add(nvme)
+            if name == "vCPUs":
```
squirrelsc: use `elif`. It's clearer, like a `switch`.

lisa/features/gpu.py
LiliDeng: removed

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: For some old vm sizes which don't have the concept of active and total vcpus, we should use vcpus

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: update

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: The code logic can be more clear. It can have two variables, to save both value, and then pick one of them by priority like this. It makes the code more readable, because the core_count is set in one place.
```python
        vcpus = 0            
        vcpus_available = 0

        ...

            if name == "vCPUsAvailable":
                vcpus_available = int(sku_capability.value)
            elif name == "vCPUs":
                vcpus = int(sku_capability.value)

            ...

        if vcpus_avaiable:
            node_space.core_count = vcpus_available
        else:
            node_space.core_count = vcpus
```

mcgov:
Thank you Lili! I appreciate the quick attention to this fix

squirrelsc (approved):

PR #1918, opened 2022-04-27
Add ARM64 repo url for Ubuntu in verify_repository_installed case

lubaihua33:
In ARM64 Ubuntu, the repo url is http://ports.ubuntu.com/ubuntu-ports.


squirrelsc:
What's the reason to close it?

PR #1922, opened 2022-04-28
Dump excluded case names

ShiningCreek:
Dump excluded case names for post run statistical analysis.

### DIFF
--- a/Libraries/CommonFunctions.psm1
+++ b/Libraries/CommonFunctions.psm1
@@ -109,6 +109,9 @@ Function Select-TestCases($TestXMLs, $TestCategory, $TestArea, $TestNames, $Test
     }
     if ($ExcludeTests) {
         $excludedTestsArray = @($ExcludeTests.Trim(', ').Split(',').Trim())
+        $excludedCaseLog = "ExcludedCases.log"
+        $null = New-Item -Path $LogDir -Name $excludedCaseLog -ItemType "file"
+        $excludedCaseLogPath = Join-Path -Path $LogDir -ChildPath $excludedCaseLog
     }
 
     # Filter test cases based on the criteria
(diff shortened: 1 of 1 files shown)


Libraries/CommonFunctions.psm1
```diff
@@ -111,6 +111,10 @@ Function Select-TestCases($TestXMLs, $TestCategory, $TestArea, $TestNames, $Test
         $excludedTestsArray = @($ExcludeTests.Trim(', ').Split(',').Trim())
     }
 
+    $excludedCaseLog = "ExcludedCases.log"
+    $null = New-Item -Path $LogDir -Name $excludedCaseLog -ItemType "file"
+    $excludedCaaseLogPath = Join-Path -Path $LogDir -ChildPath $excludedCaseLog
```
LiliDeng: excludedCaseLogPath

LiliDeng:
there is a unit test fail
```
 Describing Test if LISAv2-Framework Run-LISAv2 fails to parse report results on Azure
VERBOSE: Loading module from path 'D:\a\1\s\TestControllers\AzureController.psm1'.
04/28/2022 04:44:04 : [ERROR] EXCEPTION: Could not find a part of the path 'D:\a\1\s\fake_path\ExcludedCases.log'.
04/28/2022 04:44:04 : [ERROR] Source: Line 115 in script .\Libraries\CommonFunctions.psm1.
```

LiliDeng:
I fixed the unit test issue. We can collect exclude test cases from existing log. Why we need another file to store them?
```
04/28/2022 09:10:06 : [INFO ] Excluded Test  : VERIFY-DPDK-OVS [Exact match]
04/28/2022 09:10:06 : [INFO ] Excluded Test  : VERIFY-DPDK-NFF-GO [Exact match]
```

ShiningCreek:
The concerns are if test log is corrupted or log pattern is changed, getting case names by parsing test log will fail to get correct result.

LiliDeng (approved):

PR #1925, opened 2022-04-28
[DPDK] Skip build dependencies for package manager installation

mcgov:
Manual installation of ninja and meson is buggy across versions, skip this process for package manager installation since it's not needed anyway.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -587,7 +587,11 @@ class DpdkTestpmd(Tool):
         node = self.node
         ubuntu = node.os
         if not isinstance(ubuntu, Ubuntu):
-            return
+            fail(
+                "_install_ubuntu_dependencies was called on node "
+                f"which was not Ubuntu: {node.os.information.full_version}"
+            )
+            return  # appease the type checker
         ubuntu.add_repository("ppa:canonical-server/server-backports")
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -576,6 +576,8 @@ def _install_dependencies(self) -> None:
                     list(self._ubuntu_packages_1804),
                     extra_args=self._debian_backports_args,
                 )
+                if self._dpdk_source == "package_manager":
```
squirrelsc: 1. How about define a constant for "package_manager"?
2. The skipped logic is long. How about add some private method like `_install_ubuntu_build_dependencies`? So the logic will be clearer. Either install by package manager or by built. The logic is like,
```python
if self._dpdk_source != PACKAGE_MANAGER:
    self._install_ubuntu_build_dependencies()
```

mcgov: Excellent! I will break it up

mcgov: This is so much cleaner I should have done this a long time ago.

squirrelsc: I learned that not too long ago. The function is not only to reuse code, it provides meaningful logic as well.

squirrelsc (dismissed):

squirrelsc (approved):

PR #1928, opened 2022-04-29
Fix device statistics test

sharsonia:


lisa/tools/ethtool.py
```diff
@@ -438,6 +438,25 @@ def _parse_rx_hash_level(self, interface: str, protocol: str, raw_str: str) -> N
...
+        self._parse_statistics_info(interface, device_statistics_raw)
+
+    def _parse_statistics_info(self, interface: str, raw_str: str) -> None:
+        statistics: Dict[str, int] = {}
+
+        items = find_groups_in_lines(raw_str, self._statistics_pattern)
+        statistics = {x["name"]: int(x["value"]) for x in items}
+
+        self.interface = interface
+        self.statistics = statistics
```
squirrelsc: `items`

sharsonia: > `items`

sorry didn't get what you are trying to say?

squirrelsc: The variable name should be changed to `items`. Because it's in statistics already, the same name as the class doesn't make sense.

squirrelsc: I mean it doesn't need to repeat the `statistics` in both class and member names.

sharsonia: > I mean it doesn't need to repeat the `statistics` in both class and member names.

hmm well thats how other classes also have in ethtool. And if I use items here then when I am traversing this dict, I would again use items in networksettings.py. So it would be like items.items(), which would be confusing

squirrelsc (approved):

PR #1931, opened 2022-05-01
Check hv_netvsc built-in from config file.

LiliDeng:


microsoft/testsuites/network/networksettings.py
```diff
@@ -656,21 +655,17 @@ def after_case(self, log: Logger, **kwargs: Any) -> None:
...
-        build_in_pattern = re.compile(r"builtin", re.M)
-
         uname_tool = node.tools[Uname]
         kernel_version = uname_tool.get_linux_information().kernel_version
 
-        modinfo = node.tools[Modinfo]
-        netvsc_module = modinfo.get_filename("hv_netvsc")
-        matched = find_patterns_in_lines(netvsc_module, [build_in_pattern])
-        if not (matched[0]):
+        config_path = f"/boot/config-{kernel_version}"
```
squirrelsc, line 661: What's the reason to change the original check approach? Can you add more detail in git commit? Because this commit changed the behavior, and the old behavior is removed. It needs to be documented in git commit history.

LiliDeng: update commit comments

squirrelsc (approved):

PR #1936, opened 2022-05-03
Manually add NVMe support

kamalca:
Manually add NVMe support for New VM SKU E96bs_v5 in the generated template.

I don't think there is a good way to detect NVMe support at this time. This change to the template may not be required in the future if NVMe capable SKUs use NVMe be default. It is unknown whether SCSI will remain a supported deployment option for these SKUs.

I am making this change so that I can start testing the New SKU.


kamalca:
Master is probably not the appropriate place for these changes. I will use another branch for testing

PR #1937, opened 2022-05-04
Fix KVM private bridge testcase

dsrivastavv:


dsrivastavv:
@LiliDeng I am getting following error while running ntttcp in l2 vm. Seems like it gets stuck after running ntttcp few times. Any idea what could be causing this?

```
2022-05-04 01:18:10.802[32020][DEBUG] lisa.server[0].cmd[3350] cmd: ['sudo', 'sh', '-c', 'ulimit -n 204800 && ntttcp  -P 64 -t 10 -W 1 -C 1 -b 64k --show-nic-packets ens3  -e  --show-dev-interrupts Hypervisor callback interrupts '], cwd: None, shell: True, sudo: True, posix: True, remote: True
2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout NTTTCP for Linux 1.4.0
2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout ---------------------------------------------------------
2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout 01:18:13 INFO: no role specified. use receiver role
2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout 01:18:13 INFO: 65 threads created
2022-05-04 01:18:11.844[32020][DEBUG] lisa.client[0].cmd[9498] cmd: ['sh', '-c', 'lagscope -s192.168.53.14  -t10 '], cwd: None, shell: True, sudo: False, posix: True, remote: True
2022-05-04 01:18:11.891[30076][DEBUG] lisa.client[0].cmd[9498].stdout lagscope 1.0.1
2022-05-04 01:18:11.891[30076][DEBUG] lisa.client[0].cmd[9498].stdout ---------------------------------------------------------
2022-05-04 01:18:11.906[32020][DEBUG] lisa.client[0].cmd[6694] cmd: ['sudo', 'sh', '-c', 'ulimit -n 204800 && ntttcp  -s192.168.53.14 -P 64 -n 320 -t 10 -W 1 -C 1 -b 64k --show-nic-packets ens3  --show-dev-interrupts Hypervisor callback interrupts '], cwd: None, shell: True, sudo: True, posix: True, remote: True
2022-05-04 01:18:11.906[30076][DEBUG] lisa.client[0].cmd[9498].stdout 01:18:14 INFO: failed to bind socket[3] to a local port: [0.0.0.0:25001]. errno = 98. Ignored
2022-05-04 01:18:11.922[30076][DEBUG] lisa.client[0].cmd[9498].stdout 01:18:14 INFO: New connection: local:48578 [socket:3] --> 192.168.53.14:6001
2022-05-04 01:18:11.969[29060][DEBUG] lisa.server[0].cmd[3350].stdout 01:18:14 INFO: Test cycle time negotiated is: 12 seconds
2022-05-04 01:18:12.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout NTTTCP for Linux 1.4.0
2022-05-04 01:18:12.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout ---------------------------------------------------------
2022-05-04 01:18:12.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:14 INFO: Test cycle time negotiated is: 12 seconds
2022-05-04 01:18:12.985[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
2022-05-04 01:18:12.985[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
2022-05-04 01:18:12.985[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
2022-05-04 01:18:13.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
2022-05-04 01:18:13.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
2022-05-04 01:18:13.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
```

I have verified that both VM's are accessible from each other

lisa/tools/ip.py
```diff
@@ -74,6 +75,13 @@ def restart_device(self, nic_name: str) -> None:
             ),
         )
 
+    def setup_nic_at_boot(self, ip: str, nic_name: str) -> None:
```
squirrelsc: The method name is confusing. What's setup on the nic? And what does mean "at boot"?

lisa/tools/rc_local.py
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import PurePosixPath
+
+from lisa.executable import Tool
+from lisa.tools.echo import Echo
+
+
+class RcLocal(Tool):
```
squirrelsc: Because it's a virtual tool. Add some comments to explain what it uses.

squirrelsc: But I'm not sure if there is better name. The rc.local is a file name, it's used to save startup scripts. The tool name can be related to the task instead of file name.

microsoft/testsuites/performance/nestedperf.py
```diff
@@ -252,10 +257,11 @@ def perf_nested_kvm_ntttcp_private_bridge(
             )
         finally:
             # stop running QEMU instances
-            node.tools[Qemu].delete_vm()
+            # node.tools[Qemu].delete_vm()
```
squirrelsc: If it's not needed, remove the code.

lisa/tools/ip.py
squirrelsc: Maybe it doesn't need a new method, just add an argument `persistent` on original method.

microsoft/testsuites/performance/nestedperf.py
dsrivastavv: Forgot to remove comment. Will fix it in next push

dsrivastavv:
> @LiliDeng I am getting following error while running ntttcp in l2 vm. Seems like it gets stuck after running ntttcp few times. Any idea what could be causing this?
> 
> ```
> 2022-05-04 01:18:10.802[32020][DEBUG] lisa.server[0].cmd[3350] cmd: ['sudo', 'sh', '-c', 'ulimit -n 204800 && ntttcp  -P 64 -t 10 -W 1 -C 1 -b 64k --show-nic-packets ens3  -e  --show-dev-interrupts Hypervisor callback interrupts '], cwd: None, shell: True, sudo: True, posix: True, remote: True
> 2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout NTTTCP for Linux 1.4.0
> 2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout ---------------------------------------------------------
> 2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout 01:18:13 INFO: no role specified. use receiver role
> 2022-05-04 01:18:10.844[29060][DEBUG] lisa.server[0].cmd[3350].stdout 01:18:13 INFO: 65 threads created
> 2022-05-04 01:18:11.844[32020][DEBUG] lisa.client[0].cmd[9498] cmd: ['sh', '-c', 'lagscope -s192.168.53.14  -t10 '], cwd: None, shell: True, sudo: False, posix: True, remote: True
> 2022-05-04 01:18:11.891[30076][DEBUG] lisa.client[0].cmd[9498].stdout lagscope 1.0.1
> 2022-05-04 01:18:11.891[30076][DEBUG] lisa.client[0].cmd[9498].stdout ---------------------------------------------------------
> 2022-05-04 01:18:11.906[32020][DEBUG] lisa.client[0].cmd[6694] cmd: ['sudo', 'sh', '-c', 'ulimit -n 204800 && ntttcp  -s192.168.53.14 -P 64 -n 320 -t 10 -W 1 -C 1 -b 64k --show-nic-packets ens3  --show-dev-interrupts Hypervisor callback interrupts '], cwd: None, shell: True, sudo: True, posix: True, remote: True
> 2022-05-04 01:18:11.906[30076][DEBUG] lisa.client[0].cmd[9498].stdout 01:18:14 INFO: failed to bind socket[3] to a local port: [0.0.0.0:25001]. errno = 98. Ignored
> 2022-05-04 01:18:11.922[30076][DEBUG] lisa.client[0].cmd[9498].stdout 01:18:14 INFO: New connection: local:48578 [socket:3] --> 192.168.53.14:6001
> 2022-05-04 01:18:11.969[29060][DEBUG] lisa.server[0].cmd[3350].stdout 01:18:14 INFO: Test cycle time negotiated is: 12 seconds
> 2022-05-04 01:18:12.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout NTTTCP for Linux 1.4.0
> 2022-05-04 01:18:12.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout ---------------------------------------------------------
> 2022-05-04 01:18:12.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:14 INFO: Test cycle time negotiated is: 12 seconds
> 2022-05-04 01:18:12.985[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
> 2022-05-04 01:18:12.985[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
> 2022-05-04 01:18:12.985[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
> 2022-05-04 01:18:13.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
> 2022-05-04 01:18:13.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
> 2022-05-04 01:18:13.000[18016][DEBUG] lisa.client[0].cmd[6694].stdout 01:18:15 ERR : pthread_create() create thread failed. errno = 11
> ```
> 
> I have verified that both VM's are accessible from each other

I have verified that the following TCP setting work fine:
```
NTTTCP_TCP_CONCURRENCY = [1,2,4,8,16,32,64,128,256,512,1024,2048,4096]
```

@LiliDeng @lubaihua33  Should this value be a function of cpu core count?

LiliDeng:
> 

it failed to create thread; we need [reboot the vm after set max user ](https://github.com/microsoft/lisa/blob/main/lisa/tools/ntttcp.py#L441)

dsrivastavv:
> > 
> 
> it failed to create thread; we need [reboot the vm after set max user ](https://github.com/microsoft/lisa/blob/main/lisa/tools/ntttcp.py#L441)

Reboot is already there in the code, right?

LiliDeng:
> > > 
> > 
> > 
> > it failed to create thread; we need [reboot the vm after set max user ](https://github.com/microsoft/lisa/blob/main/lisa/tools/ntttcp.py#L441)
> 
> Reboot is already there in the code, right?

correct

squirrelsc:
> > > > 
> > > 
> > > 
> > > it failed to create thread; we need [reboot the vm after set max user ](https://github.com/microsoft/lisa/blob/main/lisa/tools/ntttcp.py#L441)
> > 
> > 
> > Reboot is already there in the code, right?
> 
> correct

@LiliDeng Let's try to answer the question in one shot completely. So, if the reboot is included already, do you know why Divyansh met this error?

dsrivastavv:
@squirrelsc I think we can get this PR in without waiting for Lili to fix ntttcp. Fixing NTTTCP can be done independently

lisa/tools/startup.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import PurePosixPath
+
+from lisa.executable import Tool
+from lisa.tools.echo import Echo
+
+
+class Startup(Tool):
```
squirrelsc: How about `StartConfiguration`? Startup is like to control "start" of a machine.

lisa/tools/ip.py
```diff
@@ -73,6 +80,8 @@ def restart_device(self, nic_name: str) -> None:
                 f"fail to restart [down then up] the nic {nic_name}"
             ),
         )
+        if persist:
+            self.node.tools[Startup].add_command(f"ip link set dev {nic_name} up")
```
squirrelsc: What's the reason to add it for persist? My understanding is that the nics will be up by default.

dsrivastavv: We need to put nic in up state after adding the ip address. I think it's better to explicitly call it rather than replying on distro to do it. This is something that was followed in v2 as well.

squirrelsc: Let's check if it's supported by distro by default. If yes, it doesn't need this logic. The unnecessary steps bring other problems.

dsrivastavv: Sure! I will validate this and let you know

LiliDeng:
@divyanara 
First make sure below value is set, and reboot, then it will take effect, I think perf_ntttcp already take care of it, if you are sure the target machines are correct.
image

Second launch the L2 VM with more cpu, I changed the core from 2 => 8, not sure the minimal workable cpu count, you can have a try.
image

I saw you used `--show-nic-packets ens3`, I think it should be ens4.

lisa/tools/ip.py
dsrivastavv: We need to restart the nic, otherwise ip route is not updated:
```
yyy@ubuntu:~$ ip route
default via 10.0.2.2 dev ens3 proto dhcp src 10.0.2.15 metric 100
10.0.2.0/24 dev ens3 proto kernel scope link src 10.0.2.15
10.0.2.2 dev ens3 proto dhcp scope link src 10.0.2.15 metric 100
yyy@ubuntu:~$ sudo ip link set dev ens4 down
yyy@ubuntu:~$ sudo ip link set dev ens4 up
yyy@ubuntu:~$ sudo ip route
default via 10.0.2.2 dev ens3 proto dhcp src 10.0.2.15 metric 100
10.0.2.0/24 dev ens3 proto kernel scope link src 10.0.2.15
10.0.2.2 dev ens3 proto dhcp scope link src 10.0.2.15 metric 100
192.168.53.0/24 dev ens4 proto kernel scope link src 192.168.53.14
```

squirrelsc: ok. BTW, if this test case is failed to recover the nodes, the node should be marked as dirty.

dsrivastavv: Sure! I will up update this

dsrivastavv:
> @divyanara First make sure below value is set, and reboot, then it will take effect, I think perf_ntttcp already take care of it, if you are sure the target machines are correct. image
> 
> Second launch the L2 VM with more cpu, I changed the core from 2 => 8, not sure the minimal workable cpu count, you can have a try. image
> 
> I saw you used `--show-nic-packets ens3`, I think it should be ens4.

Seems like 8 core is not enough. I verified that UserTasksMax is correctly set. Let me check if higher values can work

dsrivastavv:
@LiliDeng What worked for me eventually was 16 cores and 4096MB memory configuration

lisa/tools/start_configuration.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+
+    def _check_exists(self) -> bool:
+        # check if /etc/rc.local exists
+        return self.node.shell.exists(PurePosixPath("/etc/rc.local"))
+
+    def _install(self) -> bool:
+        # create rc.local file at /etc/rc.local
+        # with shbang #!/bin/sh
+        self.node.tools[Echo].write_to_file(
```
squirrelsc, line 38: Forgot to comment this part. How about the same VM is used twice? This logic may cause duplicate content in the file. Before creating it, it needs to check if the file exists or not.

microsoft/testsuites/nested/common.py
```diff
@@ -165,17 +169,20 @@ def hyperv_remove_nested_vm(
...
-    hyperv.delete_nat_networking(switch_name, nat_name)
+        # delete nat network
+        hyperv.delete_nat_networking(switch_name, nat_name)
 
-    # enable port forwarding
-    hyperv.delete_port_forwarding(nat_name)
+        # enable port forwarding
+        hyperv.delete_port_forwarding(nat_name)
+    except Exception as e:
+        raise BadEnvironmentStateException(f"Failed to cleanup environment: {e}")
```
squirrelsc: This method is called in final already. So, if it raises another exception, it will overwrite the original exception. So if there is any error, just dump an error log and mark the environment is dirty. Not to raise bad env exception.

microsoft/testsuites/performance/nestedperf.py
```diff
@@ -234,28 +237,38 @@ def perf_nested_kvm_ntttcp_private_bridge(
...
+            try:
+                # stop running QEMU instances
+                node.tools[Qemu].delete_vm()
 
-            # clear bridge and taps
-            node.tools[Ip].delete_interface(self._BR_NAME)
+                # clear bridge and taps
+                node.tools[Ip].delete_interface(self._BR_NAME)
+            except Exception as e:
+                raise BadEnvironmentStateException(
```
squirrelsc: Similar as above comments, mark it to dirty, instead of raise exception.

lisa/tools/start_configuration.py
dsrivastavv: This already happens in ` _check_exists`. The control won't reach here if ` _check_exists` return `True`

microsoft/testsuites/nested/common.py
dsrivastavv: Sorry I didn't get this part. Did you recently add a features to mark environment as dirty? I have used `BadEnvironmentStateException` pretty much everywhere else in my code.

squirrelsc: Call `node.mark_dirty()` or `env.mark_dirty()`. 

The mark dirty happens whatever test results, pass or fail. It doesn't affect test result, just let LISA to recycle this environment, not to use by further test runs. It's used like here, the test logic is completed (pass or fail), but the clean up phase is failed.

The BadEnvironmentException happens on the test case failed due to bad environment status, and the environment should be recycled as well. It shouldn't overwrite the original failure, because raise an exception when an exception happens will hide the original exception.

dsrivastavv: Got it! Thanks for clarifying

squirrelsc (approved):

PR #1939, opened 2022-05-04
Fix generation in hyperv platform

dsrivastavv:
`image_info.hyper_v_generation` can sometime contain value "V1/V2" instead of "1/2"

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1199,7 +1199,12 @@ class AzurePlatform(Platform):
             image_info = self._get_image_info(
                 azure_node_runbook.location, azure_node_runbook.marketplace
             )
-            azure_node_runbook.hyperv_generation = image_info.hyper_v_generation
+
+            # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
+            azure_node_runbook.hyperv_generation = int(
+                image_info.hyper_v_generation.strip("V")
+            )
+
             # retrieve the os type for arm template.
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1337,6 +1340,15 @@ def _deploy(
                 plugin_manager.hook.azure_deploy_failed(error_message=error_message)
                 raise LisaException(error_message)
 
+    def _parse_hyperv_generation(self, generation: Optional[str]) -> int:
+        if generation == "V1":
```
squirrelsc: That can be done by one line like `int(generation.strip("V"))`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1199,7 +1199,12 @@ def _create_node_runbook(
             image_info = self._get_image_info(
                 azure_node_runbook.location, azure_node_runbook.marketplace
             )
-            azure_node_runbook.hyperv_generation = image_info.hyper_v_generation
+
+            # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
+            azure_node_runbook.hyperv_generation = image_info.hyper_v_generation.strip(
```
squirrelsc: use `int()` to convert it to int, so, it's consistent with the schema.

dsrivastavv: This was already getting handled downstream, so I didn't put it here explicitly

squirrelsc (approved):

PR #1961, opened 2022-05-10
Add '+' in the match pattern of ntp output for Mariner distro

lubaihua33:
In Mariner image, the output of “ntpq -c rl 127.0.0.1 | grep offset= | awk -F "=" '{print $3}' | awk '{print $1}'” is like:
+0.170273, 

We need to add '+' in the match pattern.


squirrelsc (approved):

LiliDeng:
@lubaihua33 I didn't notice your PR here, my changes https://github.com/microsoft/lisa/pull/1989/files contains this fix and merged. could you please kindly close this one?

PR #1966, opened 2022-05-12
Hibernate a VM only support from API.

LiliDeng:


lisa/features/startstop.py
```diff
@@ -16,7 +16,7 @@ def can_disable(cls) -> bool:
         # no reason to disable it, it can not be used
         return False
 
-    def _stop(self, wait: bool = True) -> None:
+    def _stop(self, wait: bool = True, hibernate: bool = False) -> None:
```
squirrelsc: There may be more stop state to support. Make it an Enum for further compatible.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -86,14 +86,22 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
 
-    def _execute(self, wait: bool, operator: str) -> None:
+    def _execute(self, wait: bool, operator: str, hibernate: bool = False) -> None:
```
squirrelsc: Pass it as an extra kwargs like 

```python
def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
    ...
    operation = operator_method(
        resource_group_name=self._resource_group_name,
        vm_name=self._vm_name,
        **kwargs,
        )
    ...
```

lisa/features/startstop.py
LiliDeng: you mean make an enum, and put hibernate in it?

lisa/sut_orchestrator/azure/features.py
LiliDeng: update

lisa/features/startstop.py
```diff
@@ -16,29 +18,29 @@ def can_disable(cls) -> bool:
         # no reason to disable it, it can not be used
         return False
 
-    def _stop(self, wait: bool = True) -> None:
+    def _stop(self, wait: bool = True, **kwargs: Any) -> None:
```
squirrelsc: Hmmm, it doesn't need to add it on each level. For the _stop, just use the real arguments as last time, so it's easy to know which parameter supported.

lisa/features/startstop.py
```diff
@@ -16,29 +18,29 @@ def can_disable(cls) -> bool:
...
         raise NotImplementedError()
 
-    def _start(self, wait: bool = True) -> None:
+    def _start(self, wait: bool = True, **kwargs: Any) -> None:
```
squirrelsc: It doesn't need to add on _start/_restart.

lisa/features/startstop.py
```diff
@@ -16,29 +18,29 @@ def can_disable(cls) -> bool:
...
-    def _restart(self, wait: bool = True) -> None:
+    def _restart(self, wait: bool = True, **kwargs: Any) -> None:
         raise NotImplementedError()
 
     def enabled(self) -> bool:
         # most platform support shutdown
         return True
 
-    def stop(self, wait: bool = True) -> None:
+    def stop(self, wait: bool = True, **kwargs: Any) -> None:
```
squirrelsc: As above, use the real arguments on stop only.

lisa/features/startstop.py
squirrelsc: Yes, add an enum, and use an extendable name like "state". The value includes hibernate and shutdown.

LiliDeng: added

lisa/features/startstop.py
```diff
@@ -1,10 +1,19 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+from enum import Enum
 
 from lisa.feature import Feature
 
 FEATURE_NAME_STARTSTOP = "StartStop"
 
+DeallocateState = Enum(
```
squirrelsc: Use the pattern like below, it supports prompt on names. And not use "deallocate", it's an Azure name.

```python
class StopState(str, Enum):
    Hibernate = "hibernate"
    Shutdown= "shutdown"
```

lisa/tools/hibernation_setup.py
```diff
@@ -18,6 +18,10 @@ class HibernationSetup(Tool):
     _repo = "https://github.com/microsoft/hibernation-setup-tool"
     _entry_pattern = re.compile(r"^(.*hibernation entry.*)$", re.MULTILINE)
     _exit_pattern = re.compile(r"^(.*hibernation exit.*)$", re.MULTILINE)
+    _received_pattern = re.compile(
```
dsrivastavv, line 24: Can you please add an example positive matching string?

lisa/features/startstop.py
LiliDeng: updated

lisa/tools/hibernation_setup.py
LiliDeng: added

squirrelsc (approved):

PR #1971, opened 2022-05-16
Libvirt: Add Installer transformers for Qemu, CH & Libvirt

pupacha:
Transformer schemas

```YAML 
transformer:
  - type: qemu_installer
    connection:
      address: $(address)
      username: $(username)
    installer:
      type: package
    libvirt:
      # type: package

      # type: source
      # repo: https://github.com/cloud-hypervisor/libvirt
      # ref: origin/ch

  - type: cloudhypervisor_installer
    connection:
      address: $(address)
      username: $(username)
    installer:
      # type: binary

      # type: package

      # type: source
      # repo: https://github.com/cloud-hypervisor/cloud-hypervisor
      # ref: origin/main

    libvirt:
      # type: package

      # type: source
      # repo: https://github.com/cloud-hypervisor/libvirt
      # ref: origin/ch
```

cc: @anirudhrb @vyadavmsft

### DIFF
--- a/lisa/sut_orchestrator/libvirt/transformers.py
+++ b/lisa/sut_orchestrator/libvirt/transformers.py
@@ -10,10 +10,14 @@ from dataclasses_json import dataclass_json
 from lisa import schema
 from lisa.node import Node, quick_connect
 from lisa.operating_system import CBLMariner, Linux, Ubuntu
-from lisa.tools import Echo, Git, Wget
+from lisa.tools import Git, Wget
 from lisa.transformer import Transformer
-from lisa.transformers.kernel_source_installer import _get_code_path
-from lisa.util import field_metadata, filter_ansi_escape, subclasses
+from lisa.util import (
+    UnsupportedDistroException,
+    field_metadata,
+    filter_ansi_escape,
+    subclasses,
+)
 from lisa.util.logger import Logger
 
 
@@ -25,29 +29,10 @@ class BaseInstallerSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
 
 @dataclass_json()
 @dataclass
-class PackageInstallerSchema(BaseInstallerSchema):
-    ...
-
-
-@dataclass_json()
-@dataclass
-class ReleaseInstallerSchema(BaseInstallerSchema):
-    ...
-
-
-@dataclass_json()
-@dataclass
 class SourceInstallerSchema(BaseInstallerSchema):
     # source code repo
     repo: str = ""
     ref: str = ""
-    # where to clone the repo
-    path: str = field(
-        default="",
-        metadata=field_metadata(
-            required=True,
-        ),
-    )
 
 
 @dataclass_json
@@ -60,6 +45,10 @@ class InstallerTransformerSchema(schema.Transformer):
     # installer's parameters.
     installer: Optional[BaseInstallerSchema] = field(
         default=None, metadata=field_metadata(required=True)
+    )
+    # libvirt installer parameters
+    libvirt: Optional[BaseInstallerSchema] = field(
+        default=None, metadata=field_metadata(required=False)
     )
 
 
@@ -83,40 +72,16 @@ class BaseInstaller(subclasses.BaseClassWithRunbookMixin):
         raise NotImplementedError()
 
 
-class LibvirtInstallerTransformer(Transformer):
-    @classmethod
-    def type_name(cls) -> str:
-        return "libvirt_installer"
+class QemuInstaller(BaseInstaller):
+    ...
 
-    @classmethod
-    def type_schema(cls) -> Type[schema.TypedSchema]:
-        return InstallerTransformerSchema
 
-    @property
-    def _output_names(self) -> List[str]:
-        return []
+class CloudHypervisorInstaller(BaseInstaller):
+    ...
 
-    def _internal_run(self) -> Dict[str, Any]:
-        runbook: InstallerTransformerSchema = self.runbook
-        assert runbook.connection, "connection must be defined."
-        assert runbook.installer, "installer must be defined."
 
-        node = quick_connect(runbook.connection, "libvirt_installer_node")
-
-        factory = subclasses.Factory[BaseInstaller](BaseInstaller)
-        installer = factory.create_by_runbook(
-            runbook=runbook.installer,
-            node=node,
-            log=self._log,
-        )
-        libvirt_version = installer.install()
-        self._log.info(f"installed libvirt version: {libvirt_version}")
-
-        if isinstance(node.os, Ubuntu):
-            node.execute("systemctl disable apparmor", shell=True, sudo=True)
-        node.execute("systemctl enable libvirtd", shell=True, sudo=True)
-        node.reboot()
-        return {}
+class LibvirtInstaller(BaseInstaller):
+    ...
 
 
 class QemuInstallerTransformer(Transformer):
@@ -139,7 +104,7 @@ class QemuInstallerTransformer(Transformer):
 
         node = quick_connect(runbook.connection, "qemu_installer_node")
 
-        factory = subclasses.Factory[BaseInstaller](BaseInstaller)
+        factory = subclasses.Factory[QemuInstaller](QemuInstaller)
         installer = factory.create_by_runbook(
             runbook=runbook.installer,
             node=node,
@@ -147,8 +112,23 @@ class QemuInstallerTransformer(Transformer):
         )
         qemu_version = installer.install()
         self._log.info(f"installed qemu version: {qemu_version}")
-
         node.reboot()
+
+        if runbook.libvirt:
+            libvirt_factory = subclasses.Factory[LibvirtInstaller](LibvirtInstaller)
+            libvirt_installer = libvirt_factory.create_by_runbook(
+                runbook=runbook.libvirt,
+                node=node,
+                log=self._log,
+            )
+            libvirt_version = libvirt_installer.install()
+            self._log.info(f"installed libvirt version: {libvirt_version}")
+
+            if isinstance(node.os, Ubuntu):
+                node.execute("systemctl disable apparmor", shell=True, sudo=True)
+            node.execute("systemctl enable libvirtd", shell=True, sudo=True)
+            node.reboot()
+
         return {}
 
 
@@ -172,7 +152,7 @@ class CloudHypervisorInstallerTransformer(Transformer):
 
         node = quick_connect(runbook.connection, "cloudhypervisor_installer_node")
 
-        factory = subclasses.Factory[BaseInstaller](BaseInstaller)
+        factory = subclasses.Factory[CloudHypervisorInstaller](CloudHypervisorInstaller)
         installer = factory.create_by_runbook(
             runbook=runbook.installer,
             node=node,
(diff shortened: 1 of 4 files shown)


lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,504 @@
...
+        linux.install_packages(dep_packages_list)
+        result = self._node.execute("pip3 install meson", shell=True, sudo=True)
+        assert not result.exit_code, "Failed to install meson"
+
+    def install(self) -> str:
+        runbook: SourceInstallerSchema = self.runbook
+
+        self._log.info("Installing dependencies for Libvirt...")
+        self._install_dependencies()
+
```
vyadavmsft: Please get rid of all unnecessary new lines.

vyadavmsft (dismissed):

lisa/sut_orchestrator/libvirt/transformers.py
vyadavmsft: @anirudhrb take a look at CH changes.

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+        self._log = log
+
+    def validate(self) -> None:
+        raise NotImplementedError()
+
+    def install(self) -> str:
+        raise NotImplementedError()
+
+
+class LibvirtInstallerTransformer(Transformer):
```
squirrelsc: Can it be used independently? My understanding is the libvirt is not an e2e virtualization tool, so there is no scenario only to install it. If so, this logic should be shared under other installers.

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+
+@dataclass_json()
+@dataclass
+class PackageInstallerSchema(BaseInstallerSchema):
+    ...
+
+
+@dataclass_json()
+@dataclass
+class ReleaseInstallerSchema(BaseInstallerSchema):
```
squirrelsc: If the schema is the same as base, it doesn't need to define a new one.

lisa/tools/qemu_img.py
```diff
@@ -1,14 +1,35 @@
...
 
 
 class QemuImg(Tool):
     @property
     def command(self) -> str:
         return "qemu-img"
 
+    @property
+    def can_install(self) -> bool:
+        for _os in [CBLMariner, Ubuntu]:
```
squirrelsc, line 17: `isinstance` is better, because it supports derived system too.

lisa/tools/qemu_img.py
```diff
@@ -1,14 +1,35 @@
...
+        return False
+
+    def _install(self) -> bool:
+        linux: Linux = cast(Linux, self.node.os)
+        if isinstance(self.node.os, CBLMariner):
+            linux.install_packages("qemu-img")
+        elif isinstance(self.node.os, CBLMariner):
+            linux.install_packages("qemu-utils")
+        else:
+            raise LisaException("Missing QemuImg tool install impl for {linux} os")
```
squirrelsc: Raise a NotImplementError is clearer.

lisa/tools/qemu_img.py
pupacha: doing `isinstance` inside the loop.

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: Sure, will include `libvirt` installation mode as field inside qemu/ch installers.
```YAML
- type: qemu_installer
  libvirt:
    - type: <libvirt_installer>
     .
     .
```

squirrelsc: How about just make the libvirt installation logic like a function, or in base installer, and call it in real installers? It doesn't need to be an installer. If it needs extra parameters, that can be in the base installer.

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+        return []
+
+    def _internal_run(self) -> Dict[str, Any]:
+        runbook: InstallerTransformerSchema = self.runbook
+        assert runbook.connection, "connection must be defined."
+        assert runbook.installer, "installer must be defined."
+
+        node = quick_connect(runbook.connection, "cloudhypervisor_installer_node")
+
+        factory = subclasses.Factory[BaseInstaller](BaseInstaller)
```
squirrelsc: That may be confused, which installers can work with this transformeron the same BaseInstaller. It can derive to two installers like "CloudHypervisorInstaller" and "QemuInstaller". And the transformer only accepts corresponding installers. The type_name can be simplified by this way.

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+class QemuPackageInstaller(BaseInstaller):
+
+    __distro_package_mapping = {
+        Ubuntu.__name__: ["qemu-kvm"],
+        CBLMariner.__name__: ["qemu-kvm"],
+    }
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "qemu_package"
```
squirrelsc: `package` is enough, if there is one more layer like `QemuInstaller`.

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: Thanks for suggesting this. Will apply it.

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: Having it as just a function, eliminates the flexibility of using different libvirt installation mode. Having an installer lets us use diff combinations. For example
1.
```YAML
- type: qemu_installer
  libvirt:
    - type: pacakge
      .
      .
````

2. 
```YAML
- type: qemu_installer
  libvirt:
    - type: source
      repo: <repo>
      ref: <ref>
      path: <path>
````

File: lisa/sut_orchestrator/libvirt/transformers.py (hunk 7 above)
Comment (pupacha) [reply lisa/sut_orchestrator/libvirt/transformers.py]: Changed.

File: lisa/sut_orchestrator/libvirt/transformers.py [hunk 8]
```diff
@@ -0,0 +1,523 @@
...
+            "setcap cap_net_admin+ep /usr/local/bin/cloud-hypervisor",
+            sudo=True,
+        )
+        return _get_package_version(self._node, "cloud-hypervisor")
+
+
+def _get_source_code(
+    runbook: SourceInstallerSchema, node: Node, default_name: str, log: Logger
+) -> PurePath:
+    code_path = _get_code_path(runbook.path, node, default_name)
```
anirudhrb: Why not just clone it in lisa working dir?

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+
+    echo = node.tools[Echo]
+    echo_result = echo.run(str(code_path), shell=True)
+    code_path = node.get_pure_path(echo_result.stdout)
+
+    if not node.shell.exists(code_path):
+        # create and give permission on code folder
+        node.execute(f"mkdir -p {code_path}")
+        node.execute(f"chmod -R 777 {code_path}")
```
anirudhrb: Why is this chmod needed?

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+def _get_source_code(
+    runbook: SourceInstallerSchema, node: Node, default_name: str, log: Logger
+) -> PurePath:
+    code_path = _get_code_path(runbook.path, node, default_name)
+
+    echo = node.tools[Echo]
+    echo_result = echo.run(str(code_path), shell=True)
+    code_path = node.get_pure_path(echo_result.stdout)
```
anirudhrb: What does `echo`ing something and reading it back achieve?

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+    @classmethod
+    def type_name(cls) -> str:
+        return "package"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return BaseInstallerSchema
+
+    def validate(self) -> None:
+        assert type(self._node.os).__name__ in self.__distro_package_mapping, (
```
squirrelsc: It can raise unsupportedOsException

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+        code_path = _get_source_code(runbook, self._node, self.type_name(), self._log)
+        self._log.info("Building source code of Cloudhypervisor...")
+        self._build_and_install(code_path)
+        return _get_package_version(self._node, "cloud-hypervisor")
+
+
+class CloudHypervisorReleaseInstaller(CloudHypervisorInstaller):
+    @classmethod
+    def type_name(cls) -> str:
+        return "latest_release"
```
squirrelsc: It doesn't need to specify "latest", maybe it needs to support different versions later. How about name it "binary"?

lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -0,0 +1,523 @@
...
+
+    if not node.shell.exists(code_path):
+        # create and give permission on code folder
+        node.execute(f"mkdir -p {code_path}")
+        node.execute(f"chmod -R 777 {code_path}")
+
+    log.debug(f"cloning code from {runbook.repo} to {code_path}...")
+    git = node.tools[Git]
+    code_path = git.clone(url=runbook.repo, cwd=code_path)
+    git.fetch(cwd=code_path)
```
squirrelsc: It doesn't need to fetch, after clone, the code is latest.

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: Sure, will use the lisa working dir. It makes things simple.

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: Will remove. It is redundant.

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: `echo` helps resolve the env variable if provided as input to `path` field. Ex: `$HOME/lisa` will be resolved to `/home/user/lisa`

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: will remove.

lisa/sut_orchestrator/libvirt/transformers.py
pupacha: Sure, will rename it to "release_binary", but current code supports to get only the latest release binary.

squirrelsc: The `binary` should be enough. If it's necessary, it can be extended in future.

pupacha: changed.

squirrelsc (approved):

PR #1975, opened 2022-05-16
Fix for JUnit notifier.

cwize1:
Fix a bug where the JUnit notifier will hit an exception if the test environment
fails to initialize.

### DIFF
--- a/lisa/notifiers/junit.py
+++ b/lisa/notifiers/junit.py
@@ -141,7 +141,9 @@ class JUnit(Notifier):
 
     # Test case completed message.
     def _test_case_completed(self, message: TestResultMessage) -> None:
-        testsuite_info = self._testsuites_info[message.suite_full_name]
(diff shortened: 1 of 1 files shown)


lisa/notifiers/junit.py
```diff
@@ -141,7 +141,9 @@ def _test_case_running(self, message: TestResultMessage) -> None:
 
     # Test case completed message.
     def _test_case_completed(self, message: TestResultMessage) -> None:
-        testsuite_info = self._testsuites_info[message.suite_full_name]
+        testsuite_info = self._testsuites_info.get(message.suite_full_name)
```
squirrelsc, line 144: When will it happen?

squirrelsc, line 144: When will it happen?

cwize1: When there is an error during Environment initialization. So, none of the test cases ever enter the running state.

squirrelsc (approved):

PR #1976, opened 2022-05-17
Add testcase for running kvm-unit-tests

anirudhrb:
The kvm-unit-tests are a set of tests which can be used to verify virtualized CPUs (like KVM and others), emulated CPUs (like TCG from QEMU) or even real CPU behavior. They are maintained by the KVM community in this repo: https://gitlab.com/kvm-unit-tests/kvm-unit-tests.

This PR adds a test case that runs all the tests from the above repo. The entire design is similar to the design used by xfstests.
Implemented a `Tool` called `KvmUnitTests` clones the repo and builds it.  `KvmUnitTestSuite` invokes that tool and analyzes the results of the test run to determine if there are any unexpected failures.

### DIFF
--- a/lisa/tools/kvm_unit_tests.py
+++ b/lisa/tools/kvm_unit_tests.py
@@ -2,19 +2,18 @@
 # Licensed under the MIT license.
 import string
 from pathlib import Path, PurePath
-from typing import List, Optional, Type, cast
+from typing import List, Type, cast
 
 from lisa.executable import Tool
-from lisa.operating_system import Debian, Posix, Redhat, Suse
-from lisa.tools import Cat, Echo
-from lisa.util import LisaException, find_patterns_in_lines
+from lisa.operating_system import Posix
 
 from .git import Git
 from .make import Make
 
+
 class KvmUnitTests(Tool):
-    test_runner_cmd: Optional[Path] = None
-    repo_root: Optional[Path] = None
+    test_runner_cmd: str
+    repo_root: PurePath
 
     repo = "https://gitlab.com/kvm-unit-tests/kvm-unit-tests.git"
     deps = [
@@ -60,38 +59,32 @@ class KvmUnitTests(Tool):
 
         # run ./configure in the repo
         configure_path = self.repo_root.joinpath("configure")
-        self.node.execute(configure_path, cwd=self.repo_root,
-                expected_exit_code=0);
+        self.node.execute(str(configure_path), cwd=self.repo_root, expected_exit_code=0)
 
         # run make in the repo
         make.make("", self.repo_root)
 
-        self.test_runner_cmd = str(self.repo_root.joinpath("run_tests.sh"));
+        self.test_runner_cmd = str(self.repo_root.joinpath("run_tests.sh"))
         print("Finished installation\n")
-        return True 
+        return True
 
     def run_tests(self, timeout: int) -> List[str]:
         result = self.run(
-            "",
-            timeout=timeout,
-            sudo=True,
-            force_run=True,
-            cwd=self.repo_root
+            "", timeout=timeout, sudo=True, force_run=True, cwd=self.repo_root
         )
 
-        lines = result.stdout.split('\n')
+        lines = result.stdout.split("\n")
         failures: List[str] = []
         for line in lines:
-            line = ''.join(filter(lambda c: c in string.printable, line))
-            if 'FAIL' in line:
-                test_name = line.split(' ')[1]
-                failures.append(test_name);
+            line = "".join(filter(lambda c: c in string.printable, line))
+            if "FAIL" in line:
+                test_name = line.split(" ")[1]
+                failures.append(test_name)
         return failures
 
     def save_logs(self, test_names: List[str], log_path: Path) -> None:
         for test_name in test_names:
             self.node.shell.copy_back(
                 self.repo_root / "logs" / f"{test_name}.log",
-                log_path / f"{test_name}.log"
+                log_path / f"{test_name}.log",
             )
-
--- a/lisa/tools/kvm_unit_tests.py
+++ b/lisa/tools/kvm_unit_tests.py
@@ -1,0 +1,97 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import string
+from pathlib import Path, PurePath
+from typing import List, Optional, Type, cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Debian, Posix, Redhat, Suse
+from lisa.tools import Cat, Echo
+from lisa.util import LisaException, find_patterns_in_lines
+
+from .git import Git
+from .make import Make
+
+class KvmUnitTests(Tool):
+    test_runner_cmd: Optional[Path] = None
+    repo_root: Optional[Path] = None
+
+    repo = "https://gitlab.com/kvm-unit-tests/kvm-unit-tests.git"
+    deps = [
+        "gcc",
+        "make",
+        "binutils",
+        "qemu-kvm",
+    ]
+
+    @property
+    def command(self) -> str:
+        return str(self.test_runner_cmd)
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git, Make]
+
+    @property
+    def exists(self) -> bool:
+        return self.test_runner_cmd is not None
+
+    def _install_dep(self) -> None:
+        posix_os: Posix = cast(Posix, self.node.os)
+        tool_path = self.get_tool_path()
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
+
+        # install dependency packages
+        for package in list(self.deps):
+            if posix_os.is_package_in_repo(package):
+                posix_os.install_packages(package)
+
+    def _install(self) -> bool:
+        print("Installing kvm-unit-tests")
+        self._install_dep()
+        tool_path = self.get_tool_path()
+        make = self.node.tools[Make]
+        self.repo_root = tool_path.joinpath("kvm-unit-tests")
+
+        # run ./configure in the repo
+        configure_path = self.repo_root.joinpath("configure")
+        self.node.execute(configure_path, cwd=self.repo_root,
+                expected_exit_code=0);
+
+        # run make in the repo
+        make.make("", self.repo_root)
+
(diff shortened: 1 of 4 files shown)


lisa/tools/kvm_unit_tests.py
```diff
@@ -0,0 +1,95 @@
...
+from typing import List, Type, cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+
+from .git import Git
+from .make import Make
+
+
+class KvmUnitTests(Tool):
```
squirrelsc: This tool is not used in other test suites ideally. So put it into the same folder as `microsoft/testsuites/kvm`.

microsoft/testsuites/kvm/kvm_unit_tests.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import Path
+
```
vyadavmsft, line 4: we can get rid of  new lines here

lisa/tools/kvm_unit_tests.py
```diff
@@ -0,0 +1,95 @@
...
+        # run ./configure in the repo
+        configure_path = self.repo_root.joinpath("configure")
+        self.node.execute(str(configure_path), cwd=self.repo_root, expected_exit_code=0)
+
+        # run make in the repo
+        make.make("", self.repo_root)
+
+        self.test_runner_cmd = str(self.repo_root.joinpath("run_tests.sh"))
+        self._log.info("Finished building kvm-unit-tests")
+        return True
```
dsrivastavv: You can overload check_exists method (or use default one) to verify if this tool is installed. Instead of returning `True` here, simply call check_exists.

lisa/tools/kvm_unit_tests.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git, Make]
+
+    @property
+    def exists(self) -> bool:
+        return True if self.test_runner_cmd else False
+
+    def _install_dep(self) -> None:
```
dsrivastavv: [nit] move private methods at the end of class

microsoft/testsuites/kvm/kvm_unit_tests.py
```diff
@@ -0,0 +1,55 @@
...
+        # cause.
+        expected_failures = [
+            "pmu_lbr",
+            "svm_pause_filter",
+            "vmx",
+            "ept",
+            "debug",
+        ]
+
+        failures = node.tools[KvmUnitTests].run_tests(self.TIME_OUT)
```
dsrivastavv: You might consider moving the timeout as a constant in the tool. Looks like the tool is the bottleneck, so it is better to make it explicit in the tool.

lisa/tools/kvm_unit_tests.py
anirudhrb: Shouldn't we have all tools in the "tools" directory?

lisa/tools/kvm_unit_tests.py
anirudhrb: What's the point of doing it that way? Here we need to return true if the installation succeeded. The `make` command succeeded so we can directly return True.

microsoft/testsuites/kvm/kvm_unit_tests.py
anirudhrb: I think time-outs should be decided by the test case and not implicitly in the `Tool`.

lisa/tools/kvm_unit_tests.py
dsrivastavv: Installation passing doesn't ensure that command is added to the `PATH` variable. Calling default check exists will help do an additional check that it can be called from ssh session.

microsoft/testsuites/kvm/kvm_unit_tests.py
dsrivastavv: `KvmUnitTests` doesn't have any parameter that would modify it's behaviour when called from different tests. The timeout here is inherent to the tool. I would agree with your suggestion if say the tool ran differently based on some parameter for different tests, in which case timeout should be decided by test.

lisa/tools/kvm_unit_tests.py
anirudhrb: This command is not added to PATH. The command here is a script called `run_tests.sh` in the kvm-unit-tests repo. We invoke it directly by giving a full path to it. The "installation" in this case just involves building (`make`) the code for generating the unit test binaries. So, installation has succeeded if `make` has succeeded.

lisa/tools/kvm_unit_tests.py
anirudhrb: Done.

microsoft/testsuites/kvm/kvm_unit_tests.py
anirudhrb: Done.

microsoft/testsuites/kvm/kvm_unit_tests.py
anirudhrb: `make fix` is automatically adding this.

lisa/tools/kvm_unit_tests.py
anirudhrb: Done.

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
```diff
@@ -0,0 +1,97 @@
...
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
+
+        # install dependency packages
+        for package in list(self.deps):
+            if posix_os.is_package_in_repo(package):
+                posix_os.install_packages(package)
+
+    def _install(self) -> bool:
+        self._log.info("Building kvm-unit-tests")
```
squirrelsc: Avoid using info level log in tool.

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
```diff
@@ -0,0 +1,97 @@
...
+        self._install_dep()
+        tool_path = self.get_tool_path()
```
squirrelsc: If there is no different version for the tool, it can be setup in a global path, so if the environment is used in second time, the tool won't be installed again.

lisa/tools/kvm_unit_tests.py
squirrelsc: The current implementation may not be able to detect the tool is installed already. It means, if the environment is used second time, the tool may be installed again. That's why it needs the _check_exists. It's not a good practice to set the command in `_install`, because if it's not called, the command is not set. The command should be assigned in `_initialize`, if it's possible.

And echo Divyansh's point, the `_check_exists` should be called after installation. There may be unknown reason to fail the installation, we should know it as early as possible.

microsoft/testsuites/kvm/kvm_unit_tests.py
```diff
@@ -0,0 +1,49 @@
...
+    """,
+)
+class KvmUnitTestSuite(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            Runs the kvm-unit-tests suite for Azure VMs.
+        """,
+        priority=3,
+    )
+    def kvm_unit_tests_for_azure_vm(
```
squirrelsc: The test case should be able to run on any place, and use the standard name as other tests: `verify_kvm_unit_tests`

microsoft/testsuites/kvm/kvm_unit_tests.py
```diff
@@ -0,0 +1,49 @@
...
+            "vmx",
+            "ept",
+            "debug",
+        ]
+
+        failures = node.tools[KvmUnitTests].run_tests()
+        if failures:
+            log.info(f"Failed tests: {failures}")
+
+        node.tools[KvmUnitTests].save_logs(failures, log_path)
```
squirrelsc: Because this line is called after `run_tests`. It doesn't need to be a separated line, just call it in `run_tests`. So every time, the raw log will be copied back.

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
```diff
@@ -0,0 +1,97 @@
...
+from typing import List, Type, cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.tools import Git, Make
+
+
+class KvmUnitTests(Tool):
+    # These tests take some time to finish executing. The default
+    # timeout of 600 is not sufficient.
```
squirrelsc, line 14: The timeout of test case is 600 seconds as well. It may need to extend in test case level too.

lisa/tools/kvm_unit_tests.py
anirudhrb: Done

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
anirudhrb: Which global path are you talking about? Is there a pre-defined path in lisa for such global things?

microsoft/testsuites/kvm/kvm_unit_tests.py
anirudhrb: The `expected_failures` list is actually "tuned" for Azure VMs. There might be other failures (or no failures at all) in other environments. That's why I named it this way.

microsoft/testsuites/kvm/kvm_unit_tests.py
anirudhrb: Done.

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
squirrelsc: get it like `self.get_tool_path(global=True)`. It won't create a path in working folder directly instead of each run folder.

microsoft/testsuites/kvm/kvm_unit_tests.py
squirrelsc: By default, test cases are for Azure, because they are in microsoft folder. But it may be used in other platforms as well. When we need that, we can just make the expected failures configurable. The test case name is a breaking change to pipelines, we should make it right.

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
anirudhrb: Done.

squirrelsc (approved):

PR #1983, opened 2022-05-19
Fix feature requirement regression

kamalca:
Takes the union of the feature requirements before taking the intersection of the other platform and node requirements.

This fixes errors where node features are said to not be found in platform capabilities.

### DIFF
--- a/lisa/runners/lisa_runner.py
+++ b/lisa/runners/lisa_runner.py
@@ -610,8 +610,8 @@ class LisaRunner(BaseRunner):
 
         # if platform defined requirement, replace the requirement from
         # test case.
-        platform_requirement = self._create_platform_requirement()
         for test_result in test_results:
+            platform_requirement = self._create_platform_requirement()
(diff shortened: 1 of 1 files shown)


lisa/runners/lisa_runner.py
```diff
@@ -635,6 +635,40 @@ def _merge_test_requirements(
                         original_node_requirement = schema.load_by_type(
                             schema.NodeSpace, node_requirement_data
                         )
+
+                        # Manage the union of the platform requirements and the node
+                        # requirements before taking the intersection of
+                        # the rest of the requirements.
+                        platform_requirement.features = search_space.SetSpace(
```
squirrelsc, line 642: The platform_requirement is used for all test cases in this method. It doesn't work, if the test run contains test cases with different feature requirements. But it's a good way to fix it. So, you can call the `_create_platform_requirement` in this loop, instead of calling it out of loop.

squirrelsc (approved):

PR #1984, opened 2022-05-20
Fix lisav3 test case issues.

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1573,6 +1573,15 @@ def _resource_sku_to_capability(  # noqa: C901
                 # will revert it till bug fixed.
                 if "standardDCSv2Family" == resource_sku.family:
```
squirrelsc: Because there are multiple items now, to reduce code lines, it should be like `if esource_sku.family in ["standardDCSv2Family", "standardNCSv2Family"]:`.

microsoft/testsuites/performance/common.py
```diff
@@ -393,6 +394,9 @@ def perf_iperf(
                     )
                 )
                 current_server_port += 1
+                time.sleep(0.5)
+            # add sleep to avoid server is not ready when client connects to it.
+            time.sleep(5)
```
squirrelsc: It can be fixed by use `wait_output` on process. 5 seconds is too long, and hard sleep isn't stable.

LiliDeng: the log is redirected into json file till the that iteration test done. wait_output doesn't work here.

squirrelsc: Is it possible to check the json file? BTW, we should use small PR to make other changes effective early.

LiliDeng: the json file will generate in the last, I will remove this commit from this PR.

squirrelsc (approved):

PR #1987, opened 2022-05-23
Fix gpu and hibernation feature broken issue.

LiliDeng:


lisa/features/hibernation.py
```diff
@@ -55,5 +51,8 @@ def can_disable(cls) -> bool:
     def _enable_hibernation(cls, *args: Any, **kwargs: Any) -> None:
         raise NotImplementedError()
 
+    def enabled(self) -> bool:
```
squirrelsc, line 54: Please fix nvme and security_profile as well.

lisa/schema.py
```diff
@@ -353,9 +353,15 @@ def __hash__(self) -> int:
         return hash(self._get_key())
 
     @staticmethod
-    def create(type: str) -> "FeatureSettings":
+    def create(
+        type: str, extended_schemas: Optional[Dict[Any, Any]] = None
+    ) -> "FeatureSettings":
         # If a feature has no setting, it will return the default settings.
-        return FeatureSettings(type=type)
+        if extended_schemas:
```
squirrelsc, line 360: It doesn't need the check, just call like below.

```python
feature = FeatureSettings(type=type, extended_schemas=extended_schemas)
```

lisa/features/hibernation.py
LiliDeng: fixed

lisa/schema.py
LiliDeng: if remove the check, will encounter below exception.
```
raise UndefinedParameterError(error_message)
dataclasses_json.undefined.UndefinedParameterError: Received input field with same name as catch-all field: 'extended_schemas': 'None'
```

squirrelsc (approved):

PR #1990, opened 2022-05-23
Fix feature not detected errors

dsrivastavv:


squirrelsc (dismissed):

dsrivastavv:
@squirrelsc  I am seeing errors in `has_resource_disk` feature. Holding merging this PR until that is fixed

dsrivastavv:
@squirrelsc Added a fix for intersection issue with AzureDiskOptionSetting and verified against earlier commit

lisa/runners/lisa_runner.py
```diff
@@ -703,7 +708,11 @@ def _create_platform_requirement(self) -> Optional[schema.NodeSpace]:
         # fill in required fields as max capability. So it can be
         # used as a capability in next steps to merge with test requirement.
         if not platform_requirement.disk:
-            platform_requirement.disk = schema.DiskOptionSettings()
+            if self.platform.type_name() == constants.PLATFORM_AZURE:
```
squirrelsc: We shouldn't create a reference from generic code to platform specified. The AzureDiskOptionSettings() should be overridden in azure platform, not in runner.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -669,6 +669,18 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
                     value.data_disk_size = self._get_disk_size_from_iops(
                         value.data_disk_iops, disk_type_iops
                     )
+        elif method_name == RequirementMethod.intersect:
+            # intersect two int range
+            assert isinstance(self.data_disk_iops, search_space.IntRange)
```
squirrelsc: It doesn't need to required both of them are not None. If one of them is None, another one should be effective. That's how intersect works with None type.

lisa/runners/lisa_runner.py
```diff
@@ -635,6 +636,10 @@ def _merge_test_requirements(
                         original_node_requirement = schema.load_by_type(
                             schema.NodeSpace, node_requirement_data
                         )
+                        original_node_requirement.disk = node_requirement.disk
```
squirrelsc: Why are the two lines needed?

dsrivastavv: This was essentially the issue. `original_node_requirement` when loaded by type converts from AzureDiskSettings to DiskSettings, removing the field `has_resource_disk`. This is fixing the issue by reassigning the correct value.

lisa/runners/lisa_runner.py
dsrivastavv: Any suggestions on how to handle this? I need this to be platform specific otherwise `intersect` doesn't work

lisa/runners/lisa_runner.py
squirrelsc: When it's loaded from type, the unsupported parts should be in extended schemas. And it will be loaded back when pass the prepare environment of platform.

lisa/runners/lisa_runner.py
squirrelsc: It should be handled in this method. What exact "doesn't work" mean here? We can discuss it tomorrow.
https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/platform_.py#L2071

lisa/schema.py
```diff
@@ -380,10 +381,16 @@ def _get_key(self) -> str:
...
     def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
         # default FeatureSetting is a place holder, nothing to do.
-        extended_schemas = None
-        if hasattr(capability, "extended_schemas"):
-            extended_schemas = capability.extended_schemas
-        return FeatureSettings.create(self.type, extended_schemas=extended_schemas)
+        value = FeatureSettings.create(self.type)
+
+        # intersect the extended schemas
+        if method_name == "intersect" and capability is not None:
```
squirrelsc: Use the constant, instead of string

lisa/schema.py
```diff
@@ -380,10 +381,16 @@ def _get_key(self) -> str:
...
+            value.extended_schemas = deep_update_dict(
```
squirrelsc: Maybe it still needs to check if `extended_schemas` attribute exists or not. Please test it.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -669,6 +669,39 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
                     value.data_disk_size = self._get_disk_size_from_iops(
                         value.data_disk_iops, disk_type_iops
                     )
+        elif method_name == RequirementMethod.intersect:
+            # both req and cap cannot be none.
+            if self.data_disk_iops is not None:
```
squirrelsc: Try method `countspace_intersect`

lisa/sut_orchestrator/azure/features.py
```diff
@@ -669,6 +669,39 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
...
+                assert isinstance(
+                    self.data_disk_iops, search_space.IntRange
```
squirrelsc: It's not true, the value can be int as well.

lisa/schema.py
dsrivastavv: Didn't find any issues as such but I have added a check to be on a safe side

squirrelsc (approved):

PR #1991, opened 2022-05-23
Set DPDK core requirement to 8

mcgov:
Set DPDK core count requirement to 8 (the recommended amount).


squirrelsc (approved):

squirrelsc:
It's ok to raise the requirements. But for key performance testing, it needs to run in exclusive VM. It means the VM is dedicated on a host, so there is no side effect from usage of other VMs.

PR #1993, opened 2022-05-23
DPDK: check supported OS versions in a single place

mcgov:
Refactor: dpdk suite isn't supported on all platforms, refactor old checks into a single function check for DPDK support and skip others.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -9,7 +9,7 @@ from semver import VersionInfo
 
 from lisa.executable import Tool
 from lisa.nic import NicInfo
-from lisa.operating_system import Debian, Fedora, Redhat, Ubuntu
+from lisa.operating_system import Debian, Fedora, Redhat, Suse, Ubuntu
 from lisa.tools import Echo, Git, Lscpu, Lspci, Modprobe, Service, Tar, Wget
 from lisa.util import LisaException, SkippedException, UnsupportedDistroException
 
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -568,8 +568,34 @@ def _load_drivers_for_dpdk(self) -> None:
...
+                supported = node.os.information.version >= "18.4.0"
+            else:
+                supported = node.os.information.version >= "10.0.0"
+        elif isinstance(node.os, Redhat):
+            supported = node.os.information.version >= "7.5.0"
+        elif isinstance(node.os, Suse):
+            supported = node.os.information.version >= "15.0.0"
+        else:
+            # this OS is not supported
+            raise SkippedException(
```
squirrelsc: Use the UnsupportedOsException, it returns more details.

mcgov: updated to throw unsupporteddistro on unknown OS's. 
Unsupported but known should still throw Skipped, this check is intended to replace all the spots we skip 16.04 debian 9 etc and other EOL versions.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -568,8 +568,34 @@ def _load_drivers_for_dpdk(self) -> None:
...
+        elif isinstance(node.os, Suse):
+            supported = node.os.information.version >= "15.0.0"
+        else:
+            # this OS is not supported
+            raise UnsupportedDistroException(
+                node.os, "This OS is not supported by the DPDK test suite for Azure."
+            )
+
+        if not supported:
+            raise SkippedException(
```
squirrelsc: This is similar, it should raise UnsupportedOsException. If this case should be skipped, the unsupported os exception should be converted to skipped exception in test case.

mcgov: Can I change UnsupportedDistroException to inherit from SkippedException instead of LisaException? It doesn't seem right that running a test on an unsupported platform results in a Fail currently.

squirrelsc: No, we cannot assume all UnsupportedDistroException should be skipped. In general, the tools shouldn't raise SkippedException, because the tool shouldn't decide to skip a test case. The test case should do. The tool should describe what happens with accurate message.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -209,7 +210,13 @@ def initialize_node_resources(
     testpmd.set_dpdk_source(dpdk_source)
     testpmd.set_dpdk_branch(dpdk_branch)
     testpmd.add_sample_apps_to_build_list(sample_apps)
-    testpmd.install()
+    try:
+        testpmd.install()
+    except UnsupportedDistroException as err:
+        # forward message from distro exception, add OS Version info that is removed.
+        raise SkippedException(
```
squirrelsc: It can totally use the exception message from `UnsupportedDistroException` like `raise SkippedException(err)`.

mcgov: Oh cool! Alright will switch

squirrelsc:
Please fix the failure and merge into one commit.

squirrelsc (approved):

PR #1994, opened 2022-05-24
DPDK: force source install on tests that require it

mcgov:
Add a function to force source install on tests that require a source build.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -108,7 +108,8 @@ class Dpdk(TestSuite):
     def verify_dpdk_ovs(
         self, node: Node, log: Logger, variables: Dict[str, Any]
     ) -> None:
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -571,3 +572,7 @@ def verify_uio_binding(
                 "/dev/uio0 still exists after driver unload"
             ),
         )
+
+    def _force_dpdk_default_source(self, variables: Dict[str, Any]) -> None:
+        if not variables.get("dpdk_source"):
```
squirrelsc: It may need to write like `variables.get("dpdk_source", None):`, in case this variable doesn't exist in runbook.

mcgov: Ah good catch! Will switch

squirrelsc (approved):

PR #1995, opened 2022-05-24
Fio &  Ntttcp perf changes

SRIKKANTH:
Changes for test time duration and min CPU count.

### DIFF
--- a/lisa/tools/ntttcp.py
+++ b/lisa/tools/ntttcp.py
@@ -171,7 +171,7 @@ class Ntttcp(Tool):
     def run_as_server_async(
         self,
         nic_name: str,
-        run_time_seconds: int = 10,
+        run_time_seconds: int = 300,
         ports_count: int = 64,
         buffer_size: int = 64,
         cool_down_time_seconds: int = 1,
@@ -212,7 +212,7 @@ class Ntttcp(Tool):
     def run_as_server(
         self,
         nic_name: str,
-        run_time_seconds: int = 10,
+        run_time_seconds: int = 300,
         ports_count: int = 64,
         buffer_size: int = 64,
         cool_down_time_seconds: int = 1,
@@ -265,7 +265,7 @@ class Ntttcp(Tool):
         nic_name: str,
         server_ip: str,
         threads_count: int,
-        run_time_seconds: int = 10,
+        run_time_seconds: int = 300,
         ports_count: int = 64,
         buffer_size: int = 64,
         cool_down_time_seconds: int = 1,
--- a/lisa/tools/ntttcp.py
+++ b/lisa/tools/ntttcp.py
@@ -171,7 +171,7 @@ class Ntttcp(Tool):
     def run_as_server_async(
         self,
         nic_name: str,
-        run_time_seconds: int = 300,
+        run_time_seconds: int = 10,
         ports_count: int = 64,
         buffer_size: int = 64,
         cool_down_time_seconds: int = 1,
@@ -212,7 +212,7 @@ class Ntttcp(Tool):
     def run_as_server(
         self,
         nic_name: str,
-        run_time_seconds: int = 300,
+        run_time_seconds: int = 10,
         ports_count: int = 64,
         buffer_size: int = 64,
         cool_down_time_seconds: int = 1,
@@ -265,7 +265,7 @@ class Ntttcp(Tool):
         nic_name: str,
         server_ip: str,
         threads_count: int,
-        run_time_seconds: int = 300,
+        run_time_seconds: int = 10,
         ports_count: int = 64,
         buffer_size: int = 64,
         cool_down_time_seconds: int = 1,
--- a/microsoft/testsuites/performance/common.py
+++ b/microsoft/testsuites/performance/common.py
@@ -47,7 +47,7 @@ def perf_disk(
     test_name: str = "",
     num_jobs: Optional[List[int]] = None,
     block_size: int = 4,
-    time: int = 120,
+    time: int = 300,
     size_gb: int = 0,
     numjob: int = 0,
     overwrite: bool = False,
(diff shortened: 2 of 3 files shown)


lisa/tools/ntttcp.py
```diff
@@ -171,7 +171,7 @@ def get_throughput(self, stdout: str) -> str:
     def run_as_server_async(
         self,
         nic_name: str,
-        run_time_seconds: int = 10,
+        run_time_seconds: int = 300,
```
squirrelsc: Keep the default value to 10 seconds. It's used not for perf testing, also functional. Extend it will make test cases take unnecessary longer time.

microsoft/testsuites/performance/common.py
```diff
@@ -47,7 +47,7 @@ def perf_disk(
     test_name: str = "",
     num_jobs: Optional[List[int]] = None,
     block_size: int = 4,
-    time: int = 120,
+    time: int = 300,
```
squirrelsc, line 50: Same as comments in ntttcp.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -60,7 +60,7 @@ class StoragePerformance(TestSuite):
         priority=3,
         timeout=TIME_OUT,
         requirement=simple_requirement(
-            min_core_count=72,
+            min_core_count=2,
```
squirrelsc, line 63: The core requirements avoid side effects from other VMs on the same host. What's the reason to change it to 2?

SRIKKANTH: To test new SKUs which can have vCPUs < 72. Some times even 4.

squirrelsc: If it's not an exclusive VM on a dedicated host, it doesn't worth to test performance on it, because the performance is effected by collocated other VMs. So, we run performance tests on exclusive VMs. There should be a feature to filter the exclusive VMs, but haven't implemented yet.

lisa/tools/ntttcp.py
SRIKKANTH: Changed back to 10s

microsoft/testsuites/performance/common.py
SRIKKANTH: The file is specific to performance tests unlike earlier ntttcp changes where my changes were at tool level. I think its okay to change the time here.

microsoft/testsuites/performance/storageperf.py
SRIKKANTH: I agree with your comment about dedicated hosts and colocation effect. But there are scenarios where we need to test new VM SKUs which are as small as 4. In cases like these irrespective of how many VMs are on the same host, we should be able to test the VM IO perf. Its not fair to exclude these VMs because they might collocated with other VMs. Its up to the user to deal with/investigate performance issues in case of colocation effects. IMHO LISAv3 should not decide this.

microsoft/testsuites/performance/common.py
squirrelsc: The number is well tuned for perf measurement, it's balanced between cost and identifying issues. This value can be used for all performance purposes tests. If it needs to extend longer in stress or longhaul tests, it can be extended in test case level.

microsoft/testsuites/performance/storageperf.py
dsrivastavv: I have seen fio failing on lower core counts. @SRIKKANTH Can you please test on a 2 core and 4 core machine and check if this works?

squirrelsc: In LISAv2 or LISAv3, we don't test on shared VMs. LISA is not only a test framework but includes Azure test standard as well. The perf measurement on shared VMs is not actionable or automated. If it needs to measure performance on lower core count, we can create a separated case, it collects the data only, but never failed.

SRIKKANTH: @squirrelsc I don't think its true in case of LISAv2. 
@LiliDeng Can you confirm? 
@somil55 I could run FIO on 2&4 core VMs without any issues. I can provide you a recent job if you want.

squirrelsc: V2 doesn't block it in code, but in practice, we only run and collect data on designed perf VM sizes, and view the data in report. If the perf is dropped, it's definitely need to be fixed. If we run test on all VMs, it's hard to know where the perf issue is from, so it's not actionable.
In V3, it's limited on test code, and in future, we should add an automated way to measure the perf, instead of viewing the data. If we allow test on shared VMs, it will produce noise, and not automatable.

SRIKKANTH: @squirrelsc In that case how we are supposed to user v3 for IO perf validation in VM SKU validation where we need to test all the VM sizes <>=72 vCPUs.

squirrelsc: The current logic can test vm size, which core count equals or is greater than 72. But it's not a good solution, a feature called "ExclusiveVM" should be implemented to pickup the VMs. It's not implemented yet. So the 72 core is temp solution.

SRIKKANTH: I'm still not clear why we cant use v3 for non exclusive VMs or how can we test non exclusive VMs. Should we maintain our own branch for new SKU validations?

squirrelsc: Again, the perf result on a shared VM is not actionable. We cannot use the test results on a shared VM to file a performance bug.

SRIKKANTH: Thats what we are doing so far in case of new VM SKU validation and everyone is ok with it.

squirrelsc:
Supported by #2033

PR #1996, opened 2022-05-24
Add case verify_stress_sriov_disable_enable.

LiliDeng:


microsoft/testsuites/network/stress.py
```diff
@@ -100,6 +103,27 @@ def verify_stress_sriov_iperf(self, environment: Environment) -> None:
...
+         network interface through sdk under stress.
+
+        Steps,
+        1. Do the basic sriov check.
+        2. Set enable_accelerated_networking as False to disable sriov.
+        3. Set enable_accelerated_networking as True to enable sriov.
+        4. Do the basic sriov check.
+        """,
+        priority=3,
+        timeout=7200,
```
squirrelsc: Does it need to take two hours?

microsoft/testsuites/network/stress.py
```diff
@@ -100,6 +103,27 @@ def verify_stress_sriov_iperf(self, environment: Environment) -> None:
...
+        requirement=simple_requirement(
+            network_interface=features.Sriov(),
+            supported_platform_type=[AZURE, READY],
```
squirrelsc: Remove Ready, because this test case used the feature, it cannot run on Ready environment.

microsoft/testsuites/network/stress.py
```diff
@@ -100,6 +103,27 @@ def verify_stress_sriov_iperf(self, environment: Environment) -> None:
...
+        ),
+    )
+    def verify_stress_sriov_disable_enable(self, environment: Environment) -> None:
+        sriov_disable_enable(environment, times=50)
```
squirrelsc, line 134: Does it need so many times?

microsoft/testsuites/network/stress.py
```diff
@@ -100,6 +103,27 @@ def verify_stress_sriov_iperf(self, environment: Environment) -> None:
         # 3. Do VF connection test.
         sriov_vf_connection_test(environment, vm_nics)
 
+    @TestCaseMetadata(
+        description="""
+        This case verify VM works well after disable and enable accelerated network in
+         network interface through sdk under stress.
+
+        Steps,
```
squirrelsc, line 118: It should specify its run times.

microsoft/testsuites/network/stress.py
LiliDeng: Run once, total time is 5589.970

microsoft/testsuites/network/stress.py
LiliDeng: I tested, 50 times can make the issue repro in 100%, Dexuan suggested to do this 100 times

squirrelsc: ok, specify it's a regression test and provide the patch's link in the test case metadata as a reference. There may be other reasons to cause issue, but we should provide what we know. And check if more cores (like 4) can get it run faster.

squirrelsc (approved):

PR #1998, opened 2022-05-25
Skip TVM test when VM is not gen2 type and add minimal requirement for os version.

LiliDeng:


microsoft/testsuites/tvm/tvmsuite.py
```diff
@@ -78,6 +81,20 @@ def verify_measuredboot_compatibility(self, node: Node) -> None:
...
+        vm_generation = node.tools[VmGeneration].get_generation()
+        if "1" == vm_generation:
+            raise SkippedException("TVM cases only support generation 2 VM.")
+        if (
+            (isinstance(node.os, Debian) and node.os.information.version < "11.0.0")
+            or (isinstance(node.os, Ubuntu) and node.os.information.version < "18.4.0")
+            or (isinstance(node.os, Redhat) and node.os.information.version < "8.3.0")
+            or (isinstance(node.os, Suse) and node.os.information.version < "15.2.0")
+        ):
+            raise SkippedException(
```
squirrelsc, line 95: Can you try if below pattern works? It can returns a more readable and consistent error message.

```python
raise SkippedException(UnsupportedDistroException(node.os), f"TVM doesn't support this version."
            )
```

LiliDeng: here is the output, it doesn't show the version info.
```
skipped: (UnsupportedDistroException(<lisa.operating_system.Ubuntu object at 0x000001418C4DCF10>), "TVM doesn't support this version.") 
```

squirrelsc: How about below?

`raise SkippedException(UnsupportedDistroException(node.os, f"TVM doesn't support this version."))`

LiliDeng: this one is good enough.

squirrelsc (approved):

PR #2003, opened 2022-05-26
Retry to install packages with --allow-unauthenticated

LiliDeng:
In some old debian distro, will see this issue.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -322,7 +322,7 @@ class Posix(OperatingSystem, BaseClassMixin):
     def install_packages(
         self,
         packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]],
-        signed: bool = True,
+        signed: bool = False,
         timeout: int = 600,
         extra_args: Optional[List[str]] = None,
     ) -> None:


lisa/operating_system.py
```diff
@@ -794,6 +800,13 @@ def _install_packages(
         install_result = self._node.execute(
             command, shell=True, sudo=True, timeout=timeout
         )
+        if self._debian_package_install_unauthenticated_regex.search(
```
squirrelsc: If we allow this blindly, it can be added by default. So that would be succeeded at first try.

lisa/operating_system.py
```diff
@@ -794,6 +800,13 @@ def _install_packages(
...
+            install_result.stdout
+        ):
+            command += " --allow-unauthenticated"
```
dsrivastavv: Can we add it as parameter to the function?

LiliDeng: it was already a parameter signed, signed = False, then `--allow-unauthenticated` will be appended, but from current test results, there are some packages from the old distro (debian) need specify --allow-unauthenticated, that means we can't know when to use this beforehand.

dsrivastavv: Got it!

lisa/operating_system.py

squirrelsc (approved):

PR #2005, opened 2022-05-27
Update NicInfo only when mac match is found

dsrivastavv:
Fix error of initializing NicInfo when upper_nic is not set

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -333,9 +333,9 @@ class Nics(InitializableMixin):
                     upper_nic_mac = ip.get_mac(nic_name)
                     if upper_nic_mac == lower_nic_mac:
                         upper_nic = nic_name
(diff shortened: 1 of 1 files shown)


LiliDeng (approved):

lisa/nic.py
```diff
@@ -333,9 +333,9 @@ def _get_node_nic_info(self) -> None:
                     upper_nic_mac = ip.get_mac(nic_name)
                     if upper_nic_mac == lower_nic_mac:
                         upper_nic = nic_name
+                        nic_info = NicInfo(upper_nic, lower_nic, pci_slot)
+                        self.append(nic_info)
                         break
```
squirrelsc, line 338: Does it need to return exception, if it's not found?

dsrivastavv: No. It essentially means that SRIOV is not present.

squirrelsc (approved):

PR #2006, opened 2022-05-27
fix sriov test cases issues

LiliDeng:


lisa/tools/lspci.py
```diff
@@ -40,7 +40,7 @@
     constants.DEVICE_TYPE_GPU: ["3D controller", "VGA compatible controller"],
 }
 
-PATTERN_MODULE_IN_USE = re.compile(r"Kernel driver in use: (.*)\r", re.M)
+PATTERN_MODULE_IN_USE = re.compile(r"Kernel driver in use: ([A-Za-z0-9_-]*)", re.M)
```
squirrelsc, line 47: provide examples above, including the invalid examples cause this change.

squirrelsc (approved):

PR #2015, opened 2022-05-31
Check exception from console log in kdump test cases

lubaihua33:
Add 'check_exception' in SerialConsole.
Add exception check from console log if the VM can't be connected after kernel crash.

### DIFF
--- a/lisa/features/serial_console.py
+++ b/lisa/features/serial_console.py
@@ -32,6 +32,20 @@ class SerialConsole(Feature):
         ),
     ]
 
+    # blk_update_request: I/O error, dev sdc, sector 0
+    # ata1.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x0
+    # Failure: File system check of the root filesystem failed
+    # (initramfs)
+    initramfs_patterns: List[Pattern[str]] = [
+        re.compile(r"^(.*blk_update_request: I/O error.*)$", re.MULTILINE),
+        re.compile(r"^(.*exception.*)$", re.MULTILINE),
+        re.compile(
+            r"^(.*Failure: File system check of the root filesystem failed.*)$",
+            re.MULTILINE,
+        ),
+        re.compile(r"^(.*\(initramfs\).*)$", re.MULTILINE),
+    ]
+
     @classmethod
     def name(cls) -> str:
         return FEATURE_NAME_SERIAL_CONSOLE
@@ -135,3 +149,21 @@ class SerialConsole(Feature):
 
         if panics:
             raise LisaException(f"{stage} found panic in serial log: {panics}")
+
+    def check_initramfs(
+        self, saved_path: Optional[Path], stage: str = "", force_run: bool = False
+    ) -> None:
+        self._node.log.debug("checking initramfs in serial log...")
+        content: str = self.get_console_log(saved_path=saved_path, force_run=force_run)
+
+        initramfs_logs = [
+            x
+            for sublist in find_patterns_in_lines(content, self.initramfs_patterns)
+            for x in sublist
+            if x
+        ]
+
+        if initramfs_logs:
+            raise LisaException(
+                f"{stage} found initramfs in serial log: {initramfs_logs}"
+            )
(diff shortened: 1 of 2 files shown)


lisa/features/serial_console.py
```diff
@@ -32,6 +32,19 @@ class SerialConsole(Feature):
         ),
     ]
 
+    disk_exception_patterns: List[Pattern[str]] = [
```
squirrelsc: Add examples above.

lisa/features/serial_console.py
```diff
@@ -135,3 +148,30 @@ def check_panic(
 
         if panics:
             raise LisaException(f"{stage} found panic in serial log: {panics}")
+
+    def check_exception(
```
squirrelsc: When this method is needed? Why not add patterns in kernel panic?

lisa/features/serial_console.py
```diff
@@ -135,3 +148,30 @@ def check_panic(
...
+        self._node.log.debug("checking exception in serial log...")
+        content: str = self.get_console_log(saved_path=saved_path, force_run=force_run)
+        disk_exceptions = [
+            x
+            for sublist in find_patterns_in_lines(content, self.disk_exception_patterns)
+            for x in sublist
+            if x
+        ]
+
+        filesystem_exceptions = [
```
squirrelsc: The patterns can be merged, so one loop is enough.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -291,6 +305,10 @@ def _kdump_test(self, node: Node, log: Logger) -> None:
                 )
                 time.sleep(10)
         if not stdout:
+            serial_console = node.features[SerialConsole]
+            serial_console.check_exception(
```
squirrelsc: How about add patterns into kernel panic, and just check panic here?

lisa/features/serial_console.py
lubaihua33: This is not the kernel panic. It is a file system exception, sometimes may be caused by disk I/O error, or disk driver issue.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: This is not the same as a kernel panic. When the issue happened, the disk has an exception. Then the file system check of root filesystem failed. The system entered into 'initramfs'.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -290,6 +304,11 @@ def _kdump_test(self, node: Node, log: Logger) -> None:
                     f"{identifier.__class__.__name__}: {identifier}. Retry..."
                 )
                 time.sleep(10)
+            if not stdout:
+                serial_console = node.features[SerialConsole]
+                serial_console.check_rootfilesystem_exception(
```
squirrelsc: It may generate a lot of log. It needs to reduce log. For example, check the log every 1 minute, or use another loop before this one to check it only on first one or two minutes.

lisa/features/serial_console.py
squirrelsc: Does it make the system hang? If so, it's same as kernel panic.

lubaihua33: No, the system doesn't hang. It will enter into 'initramfs', which is usually caused by disk corruption or the file system is not mounted properly. In the kdump issue, there is disk corruption because the disk driver is not correct, which also causes the file system check failed.

When entering into 'initramfs',  If the disk can be repaired, such as using 'fsck' to repair it, then we can input 'reboot' in the 'initramfs' model, and the system can boot up normally.

squirrelsc: can we rename it more accurately than "exception"? For example, rename it like "check_initramfs"? There may be better names.

microsoft/testsuites/kdump/kdumpcrash.py
lubaihua33: Actually, it doesn't generate a lot of logs. 
If the VM can be connected, the 'stdout' should not be NULL. So it won't download the console log.
If the VM can't be connected in 300s (the timeout of 'try_connect' is 300s), it will download the console log. If there is initramfs, it will raise exception. If not, it will keep to try_connect. The kdump timeout is 800s, so it will download the console log 3 times at most.

squirrelsc (approved):

PR #2017, opened 2022-06-01
Azure: Fix deployment issues with custom rg name input

pupacha:
* short_name field is not assigned any value when node name is provided
in runbook. This change will have it assigned a value in all cases.
* Added explicit parameter for prefix for all instances of truncate_keep_prefix function calls.
`def truncate_keep_prefix(content: str, kept_len: int, prefix: str = "lisa-") -> str:`
  the value passed to `content` parameter of `truncate_keep_prefix` doesn't necessarily start with the default `prefix` parameter value `lisa-`. Hence, changed required instances of the function call to pass `content[:5]` as prefix. Otherwise, there's an expection of prefix mismatch.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1160,11 +1160,11 @@ class AzurePlatform(Platform):
             azure_node_runbook.name = truncate_keep_prefix(
                 f"{name_prefix}-n{index}", 50
             )
-            # It's used as computer name only. Windows doesn't support name more
-            # than 15 chars
-            azure_node_runbook.short_name = truncate_keep_prefix(
-                azure_node_runbook.name, 15
-            )
+        # It's used as computer name only. Windows doesn't support name more
+        # than 15 chars
+        azure_node_runbook.short_name = truncate_keep_prefix(
+            azure_node_runbook.name, 15
+        )
         if not azure_node_runbook.vm_size:
             raise LisaException("vm_size is not detected before deploy")
         if not azure_node_runbook.location:
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1154,16 +1154,18 @@ class AzurePlatform(Platform):
         )
 
         if not azure_node_runbook.name:
-            # the max length of vm name is 64 chars. Below logic takes last 40
-            # chars in resource group name and keep the leading "lisa-". So it's
-            # easy to identify it's a lisa created node.
-            azure_node_runbook.name = truncate_keep_prefix(
-                f"{name_prefix}-n{index}", 50
-            )
+            # the max length of vm name is 64 chars. Below logic takes last 45
+            # chars in resource group name and keep the leading 5 chars.
+            # name_prefix can contain any of customized (existing) or
+            # generated (starts with "lisa-") resource group name,
+            # so, pass the first 5 chars as prefix to truncate_keep_prefix
+            # to handle both cases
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -514,9 +514,8 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
             normalized_name = constants.NORMALIZE_PATTERN.sub("-", constants.RUN_NAME)
             # Take last chars to make sure the length is to exceed max 90 chars
             # allowed in resource group name.
-            resource_group_name = truncate_keep_prefix(
-                f"{normalized_name}-e{environment.id}", 80
-            )
+            rg_name = f"{normalized_name}-e{environment.id}"
```
squirrelsc: This part should be ok, the resource group name is generated, and it should start with "lisa-". The original logic can double check it. The normalized name should always start with "lisa-"

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1160,11 +1160,11 @@ def _create_node_runbook(
...
             )
-            # It's used as computer name only. Windows doesn't support name more
-            # than 15 chars
-            azure_node_runbook.short_name = truncate_keep_prefix(
-                azure_node_runbook.name, 15
-            )
+        # It's used as computer name only. Windows doesn't support name more
+        # than 15 chars
+        azure_node_runbook.short_name = truncate_keep_prefix(
+            azure_node_runbook.name, 15
```
squirrelsc: There is the place to keep first 5 chars. Other parts should keep as is.

lisa/sut_orchestrator/azure/platform_.py
pupacha: Yes, will revert this line change.

lisa/sut_orchestrator/azure/platform_.py
pupacha: and I believe it also needs to be changed at https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/platform_.py#L1160 as `name_prefix` is not always guaranteed to start with "lisa-"

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1157,14 +1157,13 @@ def _create_node_runbook(
...
-            azure_node_runbook.name = truncate_keep_prefix(
-                f"{name_prefix}-n{index}", 50
-            )
-            # It's used as computer name only. Windows doesn't support name more
-            # than 15 chars
-            azure_node_runbook.short_name = truncate_keep_prefix(
-                azure_node_runbook.name, 15
-            )
+            node_name = f"{name_prefix}-n{index}"
+            azure_node_runbook.name = truncate_keep_prefix(node_name, 50, node_name[:5])
```
squirrelsc, line 1164: Add comments above that, the leading "lisa-" is handled in previous logic. The name_prefix includes both customized and generated. So skip the validation.

pupacha: Added comment.

squirrelsc: The description doesn't specify the reason actually. There is validation to make sure either it's existing rg name, or start with "lisa-". Because the two conditions are handled before calling into this method, and it cannot be differentiated here. So, we just use first 5 chars to skip the head validation.

squirrelsc: If we don't specify clearly, the logic will be confused, and in future, it may be broken by others.

squirrelsc (approved):

PR #2022, opened 2022-06-01
Skip test when kernel doesn't enable cifs module, get current driver supported SMB version.

LiliDeng:


microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -369,6 +383,20 @@ def xfstesting_azure_file_share_validation(
     ) -> None:
         assert isinstance(environment.platform, AzurePlatform)
         node = cast(RemoteNode, environment.nodes[0])
+        if not _is_module_enabled(node, "CONFIG_CIFS"):
+            raise SkippedException(
```
squirrelsc, line 387: Raise with unsupported distro exception with error message.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -68,10 +60,32 @@ def _prepare_data_disk(
...
+        version = "3.1.1"
+    else:
+        version = "3.0"
+    return version
+
+
+def _is_module_enabled(node: Node, config_name: str) -> bool:
+    uname_tool = node.tools[Uname]
+    kernel_ver = uname_tool.get_linux_information().kernel_version
+    config_path = f"/boot/config-{kernel_ver}"
```
squirrelsc, line 74: There are many place to copy this logic, make it as a virtual tool like "kernel_config.py". It can provide method like, `is_enabled`, `is_builtin`.

PR #2025, opened 2022-06-01
Enable ACC feature in tests

dsrivastavv:
Enable ACC by setting following in testmetadata

```
   supported_features=[Acc,....],
   ....
),
```

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -874,7 +874,7 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
         if self.excluded_features:
             for feature in self.excluded_features:
                 cap_feature = self._find_feature_by_type(
-                    feature.type, capability.excluded_features
+                    feature.type, capability.features
                 )
                 if cap_feature:
                     result.add_reason(
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -24,7 +24,6 @@ from PIL import Image, UnidentifiedImageError
 from retry import retry
 
 from lisa import Environment, Logger, features, schema, search_space
-from lisa.features import NvmeSettings
 from lisa.features.gpu import ComputeSDK
 from lisa.features.resize import ResizeAction
 from lisa.node import Node, RemoteNode
@@ -488,16 +487,6 @@ class NetworkInterface(AzureFeatureMixin, features.NetworkInterface):
                 )
             )
         return all_nics
-
-
-class Nvme(AzureFeatureMixin, features.Nvme):
-    @classmethod
-    def settings_type(cls) -> Type[schema.FeatureSettings]:
-        return NvmeSettings
-
-    def _initialize(self, *args: Any, **kwargs: Any) -> None:
-        super()._initialize(*args, **kwargs)
-        self._initialize_information(self._node)
 
 
 # Tuple: (IOPS, Disk Size)
--- a/lisa/features/__init__.py
+++ b/lisa/features/__init__.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from .acc import ACC
 from .disks import (
     Disk,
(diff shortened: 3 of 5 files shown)


lisa/schema.py
```diff
@@ -565,6 +565,65 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
         return value
 
 
+@dataclass_json()
+@dataclass()
+class ACCOptionSettings(FeatureSettings):
```
squirrelsc: The feature setting should be in the feature file. The Disk and Network is special.

lisa/schema.py
```diff
@@ -565,6 +565,65 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
...
+    type: str = "ACC"
+    is_supported: Optional[bool] = None
```
squirrelsc: As we discussed, can we use unsupported_features? If so, it doesn't need to add this flag.

lisa/schema.py
```diff
@@ -724,6 +783,7 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
         default=search_space.IntRange(min=512),
         metadata=field_metadata(decoder=search_space.decode_count_space),
     )
+    acc: Optional[ACCOptionSettings] = None
```
squirrelsc: it should go with other features, not in the NodeSpace directly.

lisa/schema.py
```diff
@@ -565,6 +565,65 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
...
+
+        return value
+
+    def _check_acc(
+        self, requirement: Optional[bool], capability: Optional[bool]
+    ) -> search_space.ResultReason:
+        result = search_space.ResultReason()
+        # if requirement is none, capability can be either of True or False
+        # else requirement should match capability
+        if requirement is not None:
```
squirrelsc: The logic may not need to be so complex. But let's verify if the "is_supported" can be removed. If it can be removed, follow other features pattern, they would be simpler.

lisa/schema.py
dsrivastavv: The reason I used schema is because this will allow extending any other capabilities to ACC easily. Do you think we should stick to just using features?

squirrelsc: Let's try unsupported_features firstly. If there are other settings, we can add them later.

lisa/schema.py
```diff
@@ -874,7 +874,7 @@ def check(self, capability: Any) -> search_space.ResultReason:
         if self.excluded_features:
             for feature in self.excluded_features:
                 cap_feature = self._find_feature_by_type(
-                    feature.type, capability.excluded_features
```
squirrelsc, line 877: Why change here?

dsrivastavv: The logic seems wrong. We need to throw error if the unsupported feature in requirement is present in supported feature in capabilities.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -500,6 +500,12 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._initialize_information(self._node)
 
 
+class ACC(AzureFeatureMixin, features.ACC):
```
squirrelsc: If there is no special logic, it doesn't need to create it in platform.

dsrivastavv: I see a similar thing being done for `class Nvme(AzureFeatureMixin, features.Nvme):`, so I thought this is the norm we followed. Can you please clarify?

squirrelsc: It's optional. Can you try if it works, if so, we need to remove nvme as well.

dsrivastavv: Verified that removing NVME doesn't cause any issue

lisa/schema.py
dsrivastavv: This comment is not applicable since we removed settings

squirrelsc (approved):

PR #2031, opened 2022-06-02
Remove DPDK support for Oracle

kamalca:
Oracle is not an officially supported distro for DPDK

Also, a small change to remove the uname -r portion of the kernel-devel package name since it tends to cause failures on CentOS and some versions of RHEL

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -11,7 +11,12 @@ from lisa.executable import Tool
 from lisa.nic import NicInfo
 from lisa.operating_system import Debian, Fedora, Oracle, Redhat, Suse, Ubuntu
 from lisa.tools import Echo, Git, Lscpu, Lspci, Modprobe, Service, Tar, Wget
-from lisa.util import LisaException, SkippedException, UnsupportedDistroException
+from lisa.util import (
+    LisaException,
+    MissingPackagesException,
+    SkippedException,
+    UnsupportedDistroException,
+)
 
 PACKAGE_MANAGER_SOURCE = "package_manager"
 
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -81,7 +81,7 @@ def command(self) -> str:
 
     _redhat_packages = [
         "psmisc",
-        "kernel-devel-$(uname -r)",
+        "kernel-devel",
```
dsrivastavv: This might install kernel devel packages for a different kernel version. Have you checked if this scenario happens?

kamalca: For the cases where there is no `kernel-devel-$(uname -r)` package, it does install a package for a slightly different kernel build. For example, `4.18.0-240.10.1.el8_3.x86_64` installs `kernel-devel-4.18.0-348.7.1.el8_5.x86_64`. However, the test passes with the mismatch of kernel version.

I have not checked whether there are any cases where `kernel-devel-$(uname -r)` exists but `kernel-devel` does not install that package. From my experience with the InfiniBand test cases, it is better to trust the package manager to select the best kernel-devel package to install. The only case where this is not true is if the kernel has been updated but not rebooted yet so the kernel version might not match.

kamalca: If there are any problems with this, then we would need to try installing `kernel-devel-$(uname -r)` first then if that fails, then try installing `kernel-devel`.

squirrelsc (dismissed):

microsoft/testsuites/dpdk/dpdktestpmd.py
dsrivastavv: Agreed. Retrying without `uname -r` on failure would be more robust.

squirrelsc (approved):

dsrivastavv (approved):

PR #2032, opened 2022-06-02
Reload hv modules all in one command

kamalca:
Failures in the reload_hyperv_modules test case were due to the `ip link` and `dhclient` portion of the command which reset the network connection after the module has been reloaded. Since the errors weren't with reloading the module, we can drastically reduce the number of failures by only resetting the network setup once per test instead of once per reload.

### DIFF
--- a/microsoft/testsuites/core/hv_module.py
+++ b/microsoft/testsuites/core/hv_module.py
@@ -222,21 +222,22 @@ class HvModule(TestSuite):
                 "kernel and therefore can not be reloaded"
             )
 
-        for i in range(loop_count):
-            result = node.execute(
-                f"modprobe -r -v {module}; modprobe -v {module}; "
-                "ip link set eth0 down; ip link set eth0 up; dhclient -r; dhclient",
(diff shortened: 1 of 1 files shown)


kamalca:
I do understand that this might not be the best solution in terms of catching the failure as soon as it happens and having insight into why a failure occurred. However, any failures to reload the hv_netvsc driver would likely cause a kernel panic and not provide an error message response anyway. This also drastically reduces the time required to run the test as the majority of the time is spent running `dhclient` to reestablish a connection.

squirrelsc (approved):

squirrelsc:
> I do understand that this might not be the best solution in terms of catching the failure as soon as it happens and having insight into why a failure occurred. However, any failures to reload the hv_netvsc driver would likely cause a kernel panic and not provide an error message response anyway. This also drastically reduces the time required to run the test as the majority of the time is spent running `dhclient` to reestablish a connection.

If there is a chance on kernel panic. It should call serial_console.check_panic after timeout.

PR #2033, opened 2022-06-02
azure: support maximized cap on vm size

squirrelsc:
It's used with vm_size to ignore requirements on test cases. It can force run perf tests on any vm size.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -419,6 +419,14 @@ class AzurePlatform(Platform):
                             matched_cap = azure_cap
                             matched_score = matcher.ratio()
                     if matched_cap:
+                        # If max capability is set, use the max capability,
+                        # instead of the real capability. It needs to be in the
+                        # loop, to find supported locations.
+                        if node_runbook.maximize_capability:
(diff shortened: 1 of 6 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1831,13 +1837,9 @@ def _generate_mockup_capability(
...
         # all nodes support following features
+        all_features = self.supported_features()
         node_space.features.update(
-            [
-                schema.FeatureSettings.create(features.Nvme.name()),
-                schema.FeatureSettings.create(features.Gpu.name()),
-                schema.FeatureSettings.create(features.StartStop.name()),
-                schema.FeatureSettings.create(features.SerialConsole.name()),
-            ]
+            [schema.FeatureSettings.create(x.name()) for x in all_features]
```
dsrivastavv, line 1854: What happens if `unsupported_features` is set by the testcase? Seems like adding all features to capability will results in test case being skipped

squirrelsc: Good catch. It's to maximize capabilities as possible, but there are some conflicts among features, like availability set. It's not to support performance scenarios mainly. Not target to enable all cases.

PR #2034, opened 2022-06-03
kvm_unit_tests_tool: Fix permissions for copy_back

pupacha:
Since the tests are run as root, copy_back fails on Mariner distro as
the user doesn't have read access to logs. Hence, adding read
permissions.

cc: @anirudhrb

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -1133,7 +1133,7 @@ class RPMDistro(Linux):
         return False
 
     def _is_package_in_repo(self, package: str) -> bool:
-        command = f"{self._dnf_tool()} --showduplicates list {package} -y"
+        command = f"{self._dnf_tool()} list {package} -y"
         result = self._node.execute(command, sudo=True, shell=True)
         return 0 == result.exit_code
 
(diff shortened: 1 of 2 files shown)


anirudhrb (approved):
LGTM

lisa/operating_system.py
```diff
@@ -1133,7 +1133,7 @@ def _package_exists(self, package: str) -> bool:
         return False
 
     def _is_package_in_repo(self, package: str) -> bool:
-        command = f"{self._dnf_tool()} --showduplicates list {package} -y"
+        command = f"{self._dnf_tool()} list {package} -y"
```
squirrelsc, line 1136: @LiliDeng Can you review if this change has regression? So, it needs to explain parameters in comments. It can help others understand the usage of parameters immediately, instead of checking with original author.

LiliDeng: @pupacha hi, do you find any issue when use --showduplicates?

pupacha: @squirrelsc @LiliDeng Recently I added implementation of `_is_package_in_repo` for `RPMDistro` Class - https://github.com/microsoft/lisa/commit/d806c409c40bdc373ed16c90ae260a9d389cd881 .
But just noticed that `self._dnf_tool()` can return `tdnf` for CBL-Mariner and `tdnf` does not support `--showduplicates` param. There is no impact for removing this parameter as it just helps to display duplicate packages for other dnf tools.

squirrelsc (approved):

PR #2038, opened 2022-06-05
Fix hibernation cases reuse not matched env issue.

LiliDeng:


LiliDeng:
tested. this way doesn't work.

PR #2040, opened 2022-06-06
Fix EOL centos repo issue.

LiliDeng:


lisa/operating_system.py
```diff
@@ -1394,6 +1394,24 @@ def capture_system_information(self, saved_path: Path) -> None:
             saved_path / "centos-release.txt",
         )
 
+    def _initialize_package_installation(self) -> None:
+        # refer https://www.centos.org/centos-linux-eol/
```
squirrelsc, line 1398: As the document, it mentions only CentOS 8. How about 7.x?

LiliDeng: till now, I didn't find this issue on CentOS 7.

squirrelsc (approved):

PR #2042, opened 2022-06-06
Skip test when current distro doesn't support lis installation

LiliDeng:


microsoft/testsuites/lis/lissuite.py
```diff
@@ -50,7 +50,7 @@ def verify_lis_driver_version(self, node: Node, log: Logger) -> None:
         cat = node.tools[Cat]
         modinfo_tool = node.tools[Modinfo]
         wget_tool = node.tools[Wget]
-        node.tools[LisDriver]
+        self._install_lis(node)
```
squirrelsc: The method is to check lis supported or not, not to install it. So it should be called like `_check_lis_installable`.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -242,6 +233,19 @@ def _check_lis_version_hex(
             f" LIS version: {source_version_hex}, Actual LIS version: {version_hex}"
         )
 
+    def _install_lis(self, node: Node) -> LisDriver:
+        try:
+            lisdriver = node.tools[LisDriver]
+            return lisdriver
+        except Exception as identifier:
+            if "cannot find [lisdriver]" in str(identifier):
```
squirrelsc: It's better to wrap the error in the LisDriver, and then raise an unsupported distro exception.

microsoft/testsuites/lis/lissuite.py
LiliDeng: updated

microsoft/testsuites/lis/lissuite.py
LiliDeng: updated

squirrelsc (approved):

PR #2046, opened 2022-06-07
skip tests in the beginning for xfstests and add detailed assert info when no data disk found.

LiliDeng:


microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -522,3 +534,14 @@ def _execute_xfstests(
...
         xfstests.check_test_results(cmd_results.stdout, log_path)
+
+    def _install_xfstests(self, node: Node) -> Xfstests:
+        try:
+            xfstests = node.tools[Xfstests]
+            return xfstests
+        except UnsupportedDistroException as identifier:
+            raise SkippedException(identifier)
+
+    def _is_btrfs_supported(self, node: Node) -> None:
```
squirrelsc: Usually, if a method starts with "is_", it means the method returns a boolean. This method is to check some condition, it should start like "_check" or "_skip_btrfs".

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -522,3 +534,14 @@ def _execute_xfstests(
             timeout=self.TIME_OUT,
         )
         xfstests.check_test_results(cmd_results.stdout, log_path)
+
+    def _install_xfstests(self, node: Node) -> Xfstests:
```
squirrelsc, line 538: try before_case. It can raise an exception, and directly skip a case. It doesn't need to add one line in each test case.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: fixed.

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: the return value xfstests will be used in test case.

squirrelsc (approved):

PR #2047, opened 2022-06-08
NetworkController: add wait skip option for sriov toggling

mcgov:
Adding an option for tests to skip the added wait for sriov toggling. 

- DPDK tests for failover after sriov rescinding already check for the VF disable/reenable so the check messes with the timeout timing is redundant in that case (and generates a lot of extra logs).

### DIFF
--- a/lisa/features/network_interface.py
+++ b/lisa/features/network_interface.py
@@ -21,7 +21,7 @@ class NetworkInterface(Feature):
     def enabled(self) -> bool:
         return True
 
-    def switch_sriov(self, enable: bool) -> None:
+    def switch_sriov(self, enable: bool, wait: bool = True) -> None:
         raise NotImplementedError
 
     def is_enabled_sriov(self) -> bool:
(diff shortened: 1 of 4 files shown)


LiliDeng:
wait VF show up and disappear is necessary.

mcgov:
> wait VF show up and disappear is necessary.

The dpdk rescind tests do wait, but they check for the hotplug driver event log for the synthetic interface that is used for the test. The advantage is that a) I get the dmesg logs in my runlog so can clearly see if the interface was never added and b) I can check whether SRIOV was enabled for the interface I'm using in the test. Currently I have two logs, one of which is the node.nics class spamming sysfs and ip. It's hard to read through and I think the dmesg output is a better indicator of what is actually happening on the node.

The check now only looks for interface 0 and does it with that sysfs check that assumes there will be an upper/lower pair when the VF is added. I don't like this since it's possibly not portable outside of azure (and possibly not portable even in Azure on some SKUs if we ever offer bare metal).

I did write that check originally so it's my fault the NIC code does that, I want to refactor the NIC code to get rid of assumptions like that.

Generally though: we should allow users to toggle async and perform their own checks while they wait for the VF to be added.

squirrelsc (approved):

PR #2048, opened 2022-06-08
DPDK: fix rescind test timing and regex issues

mcgov:
A couple of changes to the rescind test:
- Shorten the max timeout.
- Reduce number of processes we wait on. Previously there was a mashup of a few timers to kill testpmd. We don't actually care which one kills testpmd and it's possible one or the other will fail. Since we only care that testpmd returns and gives us output we can just wait on testpmd.
- Fix the regexes to not assume the VF will be named the same thing after hotplug. We already filter the old hotplug messages during the search, so it's enough to look for the VF being associated with the preexisting synthetic interface.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -35,7 +35,7 @@ from microsoft.testsuites.dpdk.dpdkvpp import DpdkVpp
 
 VDEV_TYPE = "net_vdev_netvsc"
 MAX_RING_PING_LIMIT_NS = 200000
-DPDK_VF_REMOVAL_MAX_TEST_TIME = 60 * 10
+DPDK_VF_REMOVAL_MAX_TEST_TIME = 200
 
 
 @TestSuiteMetadata(
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -260,20 +260,14 @@ class DpdkTestpmd(Tool):
             cmd,
             sudo=True,
         )
-        self.timer_proc.wait_result()
         proc_result = testpmd_proc.wait_result()
         self._last_run_output = proc_result.stdout
         self.populate_performance_data()
         return proc_result.stdout
 
     def kill_previous_testpmd_command(self) -> None:
-        # kill testpmd early, then kill the timer proc that is still running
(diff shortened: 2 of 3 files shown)


squirrelsc:
> A couple of changes to the rescind test:
> 
> * Shorten the max timeout.
> * Reduce number of processes we wait on. Previously there was a mashup of a few timers to kill testpmd. We don't actually care which one kills testpmd and it's possible one or the other will fail. Since we only care that testpmd returns and gives us output we can just wait on testpmd.
> * Fix the regexes to not assume the VF will be named the same thing after hotplug. We already filter the old hotplug messages during the search, so it's enough to look for the VF being associated with the preexisting synthetic interface.

If there are three things in the PR, it's better to split to three commits. And the same descriptions should be in the commit. When we search the code in future, usually we are looking in commit messages, not on PR.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -35,7 +35,7 @@
 
 VDEV_TYPE = "net_vdev_netvsc"
 MAX_RING_PING_LIMIT_NS = 200000
-DPDK_VF_REMOVAL_MAX_TEST_TIME = 60 * 10
+DPDK_VF_REMOVAL_MAX_TEST_TIME = 200
```
squirrelsc, line 38: What's the difference to change the timeout value? As my understanding, the longer timeout can prevent unstable status, unless the timeout is misused.

mcgov: I can change it back, it seems that more often the VF is added again but testpmd decides to ignore the kill signal and ends up running the entire time anyway. Maybe there's a better way to set the timeout where we can still get the stdout though.

mcgov:
> > A couple of changes to the rescind test:
> > 
> > * Shorten the max timeout.
> > * Reduce number of processes we wait on. Previously there was a mashup of a few timers to kill testpmd. We don't actually care which one kills testpmd and it's possible one or the other will fail. Since we only care that testpmd returns and gives us output we can just wait on testpmd.
> > * Fix the regexes to not assume the VF will be named the same thing after hotplug. We already filter the old hotplug messages during the search, so it's enough to look for the VF being associated with the preexisting synthetic interface.
> 
> If there are three things in the PR, it's better to split to three commits. And the same descriptions should be in the commit. When we search the code in future, usually we are looking in commit messages, not on PR.

I'll break it up.

PR #2050, opened 2022-06-08
DPDK: cleanup async timeout and testpmd kill

mcgov:
Remove the double wait in testpmd and shorten the timeout in the 
 event testpmd fails to terminate.

1. Testpmd could die early so it's dumb to wait in that case.
2. There's a second timer for rescind that could kill testpmd instead.
3. We don't care what the exit code for either timer is.
4. shell.kill() seems to be buggy so attempting to kill could
    result in the test timing out even though testpmd terminated
    and gave us output to process.

It's better to just wait for testpmd.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 
 import re
+from functools import partial
 from pathlib import PurePosixPath
 from typing import List, Pattern, Tuple, Type, Union
 
@@ -10,14 +11,29 @@ from semver import VersionInfo
 
 from lisa.executable import Tool
 from lisa.nic import NicInfo
+from lisa.node import Node
 from lisa.operating_system import Debian, Fedora, Oracle, Redhat, Suse, Ubuntu
-from lisa.tools import Echo, Git, Lscpu, Lspci, Modprobe, Service, Tar, Unzip, Wget
+from lisa.tools import (
+    Echo,
+    Git,
+    Kill,
+    Lscpu,
+    Lspci,
+    Modprobe,
+    Pidof,
+    Service,
+    Tar,
+    Unzip,
+    Wget,
+)
 from lisa.util import (
     LisaException,
     MissingPackagesException,
     SkippedException,
     UnsupportedDistroException,
 )
+from lisa.util.parallel import TaskManager, run_in_parallel_async
+from lisa.util.perf_timer import create_timer
 
 PACKAGE_MANAGER_SOURCE = "package_manager"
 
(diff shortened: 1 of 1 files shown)


ArjunSharda (dismissed):

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -276,8 +275,9 @@ def kill_previous_testpmd_command(self) -> None:
             "Timer process was not initialized before "
             "calling kill_previous_testpmd_command"
         ).is_not_none()
-        self.node.execute(f"killall -s INT {self.command}", sudo=True, shell=True)
-        self.timer_proc.kill()
+        self.node.execute(
+            f"killall -s INT {self.command}", sudo=True, shell=True, timeout=40
```
squirrelsc: Because the previous processes are not useful. It can be killed by 9, which responds fast.

mcgov: What do you mean? I'm not sure. It's still useful to have them since not all tests kill testpmd early, we just don't need to wait for them and don't care what their exist codes are. testpmd likes to be killed with INT so we send that signal instead of KILL or TERM

mcgov: Now that you mention it trying both would make sense.

mcgov: I totally missed that kill optionally allows other signals with timeout
how does `kill -s INT testpmd --timeout 5000 TERM --timeout 10000 KILL` sound?

mcgov: Ah nevermind, not supported on all versions of kill

squirrelsc: Kill by 9 doesn't wait for any response, so it's fast. Send message to process let it finalize. Because this place is to clean up before start, so it doesn't need to wait. But it's better to kill after each test run as well. The test environment may be used for next test case, the next test case may not be dpdk, so the leaving process affects next one.

mcgov: I had it in my head that INT was needed but it seems like KILL actually is fine as long as we don't care about the exit code of testpmd. I'll swap to KILL.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -264,32 +264,49 @@ def generate_testpmd_command(
...
     def run_for_n_seconds(self, cmd: str, timeout: int) -> str:
         self._last_run_timeout = timeout
         self.node.log.info(f"{self.node.name} running: {cmd}")
-        self.timer_proc = self.node.execute_async(
-            f"sleep {timeout} && killall -s INT {cmd.split()[0]}",
+
+        # testpmd exits cleanly when killed with INT.
+        # setup some timers to kill it after timeout
+        self.node.execute_async(
+            f"sleep {timeout} && {self.generate_testpmd_killer()}",
```
dsrivastavv: Can you please check if we can do something like this:
1. Use timeout on the test case
2. Add the kill in the `after_case` method?

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -264,32 +264,49 @@ def generate_testpmd_command(
...
             sudo=True,
         )
-        self.timer_proc.wait_result()
         proc_result = testpmd_proc.wait_result()
         self._last_run_output = proc_result.stdout
         self.populate_performance_data()
         return proc_result.stdout
 
+    def check_testpmd_is_running(self) -> str:
+        pid_result = self.node.execute(f"pidof {self.command}")
```
dsrivastavv: We can create a tool for `pidof` command

mcgov: We should but I don't want to in this PR. I only see one other spot using pidof so it will still be easy to swap this use over later.
We should open like, a running issue thread with tools we need to create.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Worth checking, I think that I had issues with the shell.kill command that timeout will call but maybe that assumption has changed.

squirrelsc: If sleep before kill, it may make trouble. The kill is called after certain seconds, it may kill the wanted processes at that time.

microsoft/testsuites/dpdk/dpdktestpmd.py
dsrivastavv: I generally try to avoid `node.execute` unless it's something very specific to the testcase. `pidof` and your logic of killing process robustly and finding processes seem like something that could be used in future tests.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Ah I remember, using wait_result(timeout=20) results in:
`2022-06-09 00:50:08.481[2972][DEBUG] lisa.env[generated_0].node[0].cmd[2820] failed on killing process: return code: 1
output: b''
stderr output: b'kill: (10366): Operation not permitted\n'`

and the test ends up failing because stdout from the process isn't available in the result object after the failure.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I'm glad you brought this up I forgot a) there's already a kill tool and b) I added stuff to it to use here and forgot to make the second change 😂

mcgov: ok waaaay cleaner now

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: node.shell.kill seems broken :\ maybe we can point it to the kill tool instead?

squirrelsc: Yes, the `node.shell.kill` doesn't support sudo. It may be the reason.

mcgov:
@squirrelsc @somil55  updated, this is much cleaner! Thank you for feedback :)

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: We'll fix that in another PR

mcgov:
@squirrelsc Super clean code looking really good, should we merge?

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: @somil55 Ah I forgot to clarify, the scheme here isn't to clean up. It's so we can stop the program and collect the test results after a variable amount of time. There is still a max timeout on the test case itself that will kill everything if things stall for some reason.

squirrelsc (approved):

PR #2057, opened 2022-06-09
Add '-runAsSudo' when running command 'dmesg'

lubaihua33:


LiliDeng:
please use lisav3 to run hibernation, let know any blocker issue when use lisav3,

LiliDeng (approved):

PR #2058, opened 2022-06-09
pidof: Add tool for use in kill tool

mcgov:
pgrep tool can fail, pidof is universal and should be more stable.
Add the tool and first use case in 'kill' tool.

### DIFF
--- a/lisa/tools/kill.py
+++ b/lisa/tools/kill.py
@@ -1,9 +1,10 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+
 from lisa.executable import Tool
 
-from .pgrep import Pgrep
+from .pidof import Pidof
 
 
 class Kill(Tool):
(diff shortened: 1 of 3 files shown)


lisa/tools/kill.py
```diff
@@ -21,9 +22,18 @@ def can_install(self) -> bool:
         return False
 
     def by_name(self, process_name: str, signum: int = SIGKILL) -> None:
-        running_processes = self.node.tools[Pgrep].get_processes(process_name)
-        for process in running_processes:
-            self.by_pid(process.id, signum)
+        pids = self.node.tools[Pidof].get_pids(process_name)
+        for pid in pids:
+            self.by_pid(pid, signum)
+        if not pids:
```
squirrelsc: try this grammar, but the pids shouldn't be None.

```python
for ...:
    ...
else:
    ...

File: lisa/tools/kill.py [hunk 2]
```diff
@@ -21,9 +22,18 @@ def can_install(self) -> bool:
...
+            self.node.log.debug(
```
squirrelsc: Use the log object of self, instead of self.node. It affects the leading chars in this place. And the different log object maybe different on handlers.

lisa/tools/kill.py
```diff
@@ -21,9 +22,18 @@ def can_install(self) -> bool:
...
+                f"Pid for {process_name} not found! attempt kill by name..."
+            )
+            # some versions of kill may not support kill by name
```
squirrelsc: If so, it shouldn't be in the "not pids" section. What if the distro supports pids but not support kill by pid?

mcgov: I'm not sure what you mean, do any distros not support killing by pid? I thought that was a baked-in concept of *nix. The sticky part afaik is that kill can be either an executable or a shell builtin and they might have different behavior.

mcgov: I guess my question is 'what should I change'

mcgov: I'll reword

squirrelsc (approved):

PR #2059, opened 2022-06-09
Add ltp lite test

dsrivastavv:
This PR migrates LTP lite testcases from LISAv2


lisa/tools/git.py
```diff
@@ -91,6 +91,9 @@ def checkout(
             # create a temp branch to checkout tag or commit.
             checkout_branch = f"{constants.RUN_ID}"
 
+        # mark directory safe
+        self._mark_safe(cwd)
```
squirrelsc: When is it needed? Why it's in `checkout` only? If the code is checked out multiple times, it shouldn't run multiple times. So this command should be have a flag to run only once.

lisa/tools/mkswap.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+
+
+class MkSwap(Tool):
+    @property
+    def command(self) -> str:
+        return "mkswap"
```
squirrelsc: The new tool should be added into `__init__.py`, so it's easy to reference.

lisa/tools/__init__.py
```diff
@@ -143,6 +148,7 @@
     "NFSClient",
     "NFSServer",
     "Nm",
+    "Nproc",
```
squirrelsc, line 153: Try to use `git add -p` to split one file change into multiple commits.

lisa/tools/chmod.py
```diff
@@ -0,0 +1,15 @@
...
+
+from lisa.executable import Tool
+
+
+class Chmod(Tool):
+    @property
+    def command(self) -> str:
+        return "chmod"
+
+    def update_folder_permission(
```
squirrelsc: remove "permission" from the name, the chmod is all about change the permission.

lisa/messages.py
```diff
@@ -177,6 +181,15 @@ class NetworkUDPPerformanceMessage(PerfMessage):
     packet_size_kbytes: Decimal = Decimal(0)
 
 
+@dataclass
+class LTPResultMessage(PerfMessage):
```
squirrelsc, line 185: LTP is not a perf message, it should inherit from community test message.

lisa/messages.py
```diff
@@ -52,8 +52,12 @@ class TransportProtocol(str, Enum):
     Udp = "UDP"
 
 
+class CommunityTestMessage(MessageBase):
```
squirrelsc: There should be a `TestResultMessageBase(MessageBase)`, and the CommunityTestMessage and TestResultMessage derive from it. Lisa use the TestResultMessage, but the community test cases dispatch the CommunityTestMessage.

lisa/messages.py
```diff
@@ -52,8 +52,12 @@ class TransportProtocol(str, Enum):
...
+    type: str = "CommunityTest"
+
+
 @dataclass
-class PerfMessage(MessageBase):
+class PerfMessage(CommunityTestMessage):
```
squirrelsc: Perf message is not related to CommunityTestMessage.

lisa/messages.py
squirrelsc: It doesn't need to create a message type for each community test. It's not easy to maintain. We can use a field like `tool` to identify which community test it is.

lisa/tools/git.py
dsrivastavv: I received an error while trying to checkout to the given tag and this is to fix the error. Let me see if this can be moved to the `clone` function

lisa/messages.py
dsrivastavv: So are we planning to move fio and ntttcp testcases to CommunityTestMessage as well? Or we do not change anything with fio and ntttcp testcases.

lisa/messages.py
dsrivastavv: How do we pass case specific information? Should I add something like a `metadata` field which contains a dictionary with additional information?

lisa/tools/mkswap.py
dsrivastavv: oops. missed it

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,272 @@
...
+            rm.remove_file(self.LTP_OUTPUT_PATH, sudo=True)
+
+        # add logging and output file parameter
+        if drive_name:
+            parameters = (
+                f"-p -q -l {self.LTP_RESULT_PATH} -o {self.LTP_OUTPUT_PATH} -z"
+                f" {drive_name} "
+            )
+        else:
+            parameters = f"-p -q -l {self.LTP_RESULT_PATH} -o {self.LTP_OUTPUT_PATH} "
```
LiliDeng:
```
parameters = f"-p -q -l {self.LTP_RESULT_PATH} -o {self.LTP_OUTPUT_PATH} "
if drive_name:
   parameters += f" -z {drive_name}"
```

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -0,0 +1,73 @@
...
+            test_list = self.LTP_LITE_TESTS
+
+        # get comma seperated list of tests to skip
+        if skip_tests:
+            skip_test_list = skip_tests.split(",")
+        else:
+            skip_test_list = []
+
+        # run ltp lite tests
+        drive_name = "/dev/sdc"
```
LiliDeng: don't use hard code

lisa/messages.py
squirrelsc: That's different things. The ntttcp and fio are performance test cases, not functional test cases. The perf test cases supposed to pass every time, and we collect perf results. If it's failed, we should fix them. Because we care about the perf results.
For the community functional tests, one LISA test contains multiple one. It should report as one LISA test result and multiple community test results.

lisa/messages.py
squirrelsc: If it's special field, it can be put into the information as dict. For below ones, see each comment.

```python
    ltp_version: str = "" # it can be a test_version field in community test message.
    architecture: str = "" # it can be added in the information of base test result message.
    ltp_test_name: str = "" # it's the name field of base test result message.
    ltp_test_result: str = "" # it's the status field of base test result message.
    ltp_test_exit_value: int = 0 # if it's the way to return error message of ltp test case, put it in the message field.
```

lisa/tools/free.py
```diff
@@ -0,0 +1,23 @@
...
+    @property
+    def command(self) -> str:
+        return "free"
+
+    def get_swap_size(self) -> int:
+        # Return total swap size
+        # Example output:
+        #         total used free
+        # Swap:   0     0    0
+        out = self.run("-m").stdout
```
LiliDeng: do we need set force_run as True by default?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # parse results
+        result_messages = self._parse_results(
+            lisa_test_name, environment, self.LTP_RESULT_PATH
+        )
+
+        return result_messages
+
+    def _install(self) -> bool:
+        assert isinstance(self.node.os, Posix), f"{self.node.os} is not supported"
```
LiliDeng, line 124: we can make ltp as a tool like xfstests.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -0,0 +1,73 @@
...
+from lisa import TestCaseMetadata, TestSuite, TestSuiteMetadata, schema, search_space
+from lisa.environment import Environment
+from lisa.node import Node
+from lisa.testsuite import simple_requirement
+from microsoft.testsuites.ltp.ltp import Ltp
+
+
+@TestSuiteMetadata(
+    area="ltp",
+    category="functional",
```
LiliDeng: it is community category?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,272 @@
...
+                    "ntp",
+                    "git-core",
+                    "db48-utils",
+                    "libaio-devel",
+                    "libattr1",
+                    "libcap-progs",
+                    "libdb-4_8",
+                    "perl-BerkeleyDB",
+                ]
+            )
```
LiliDeng, line 232: how about other distros?

dsrivastavv: I have tested on the major classes: Fedore, Suse, Debian and Mariner. This was also the class of OS supported by v2.

microsoft/testsuites/ltp/ltpsuite.py
dsrivastavv: Are there any scenarios where "/dev/sdc" would be OS or resource disk?

LiliDeng: the disk labels are changeable if there is a reboot in this env, then /dev/sdc can be any disk, os disk, resource disk, data disk

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+        # assert that all tests passed
+        failed_tests = []
+        for result in results:
+            if result.status == TestStatus.FAILED:
+                failed_tests.append(result.name)
+
+            # create test result message
+            community_message = CommunityTestMessage()
```
squirrelsc: as discussed, use the create_message to handle node and environment together.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -0,0 +1,94 @@
...
+            community_message.name = result.name
+            community_message.suite_name = self.__class__.__name__
```
squirrelsc: The suite name should be from community test. If it doesn't have, keep it empty.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # run ltp lite tests
+        drive_name = "/dev/sdc"
+        results = node.tools[Ltp].run_test(
+            "ltp-lite", environment, test_list, skip_test_list, drive_name=drive_name
+        )
+
+        # assert that all tests passed
+        failed_tests = []
+        for result in results:
```
squirrelsc: This part should be integrated in the Ltp tool

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,325 @@
...
+                    "950000",
+                    PurePosixPath("/sys/fs/cgroup/cpu/user.slice/cpu.rt_runtime_us"),
+                    sudo=True,
+                )
+
+        # Minimum 4M swap space is needed by some mmp test
+        if self.node.tools[Free].get_swap_size() < 4:
+            self.node.execute("dd if=/dev/zero of=/tmp/swap bs=1M count=1024")
+            self.node.tools[MkSwap].run("/tmp/swap")
+            self.node.tools[SwapOn].run("/tmp/swap")
```
LiliDeng: make sure to restore them back to original status, it may have impact for other cases.

microsoft/testsuites/ltp/ltpsuite.py
dsrivastavv: Should the `CommunityTestMessage` be also sent by the ltp tool?

squirrelsc: yes, the test message should be sent by the tool.

dsrivastavv:
@squirrelsc We have few output and result files generated by the ltp tests that need to be downloaded in the runtime logs. What would be the best folder to put these in? I was thinking of `tests\{run_id}-ltp_lite` folder

microsoft/testsuites/ltp/ltp.py
dsrivastavv: I will add the code for removing swap but I think there might be a bigger issue. I found following warning on LTP github repo:
```
Don't run them on production systems. Growfiles, doio, and iogen in particular stress the I/O capabilities of systems and while they should not cause problems on properly functioning systems, they are intended to find (or cause) problems.
```

I think we should mark the nodes as dirty which run ltp tests. Any bugs in the ltp which don't lead to clean exit of tests could cause problems with lisa testing. @squirrelsc @LiliDeng suggestions?

squirrelsc: > I will add the code for removing swap but I think there might be a bigger issue. I found following warning on LTP github repo:
> 
> ```
> Don't run them on production systems. Growfiles, doio, and iogen in particular stress the I/O capabilities of systems and while they should not cause problems on properly functioning systems, they are intended to find (or cause) problems.
> ```
> 
> I think we should mark the nodes as dirty which run ltp tests. Any bugs in the ltp which don't lead to clean exit of tests could cause problems with lisa testing. @squirrelsc @LiliDeng suggestions?

If it's not recoverable, mark the node to dirty.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,355 @@
...
+        # assert that all tests passed
+        failed_tests = []
+        for result in results:
+            if result.status == TestStatus.FAILED:
+                failed_tests.append(result.name)
+
+            # create test result message
+            info: Dict[str, Any] = {}
+            info["information"] = {}
+            info["information"]["architecture"] = result.architecture
```
squirrelsc: The architecture should be common for all community test cases. Use the value from kernel information of os, so it can be shared cross test cases.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -0,0 +1,79 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+
+from logging import Logger
+from typing import Any, Dict
+
+from lisa import TestCaseMetadata, TestSuite, TestSuiteMetadata, schema, search_space
+from lisa.node import Node
```
squirrelsc: The Node and simple_requirement should be imported from the lisa. If they are not in `lisa/__init__.py`, they should be. We hope test case writers can get all from `lisa.*`.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,355 @@
...
+
+    # Machine Architecture: x86_64
+    _RESULT_LTP_ARCH_REGEX = re.compile(r"Machine Architecture: (.*)\s+")
+
+    LTP_DIR_NAME = "ltp"
+    LTP_TESTS_GIT_TAG = "20190930"
+    LTP_GIT_URL = "https://github.com/linux-test-project/ltp.git"
+    BUILD_REQUIRED_DISK_SIZE_IN_GB = 2
+    LTP_RESULT_PATH = "/opt/ltp/ltp-results.log"
+    LTP_OUTPUT_PATH = "/opt/ltp/ltp-output.log"
```
squirrelsc, line 56: If the ltp test cases run twice on a machine, the previous one may be overwritten. So use log_path of test case to put logs, and they can be uploaded to other log places.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,355 @@
...
+        cat = self.node.tools[Cat]
+        with open(f"{log_path}/ltp-output.log", "w") as f:
+            f.write(
+                cat.read(
+                    self.LTP_OUTPUT_PATH, sudo=True, force_run=True, no_debug_log=True
+                )
+            )
+
+        # write results to working_path
+        with open(f"{log_path}/ltp-results.log", "w") as f:
```
squirrelsc: the working_path is different with log_path. The working_path put temp files, which are not considered as log, it might be used for troubleshooting. But the log_path might be copied to some places to save. So the log path should put log only, and make it's a reasonable size.

dsrivastavv: I want to copy the results and output so that they can be saved as additional logs for future reference.

dsrivastavv: I will update the comment if it's confusing

microsoft/testsuites/ltp/ltp.py
dsrivastavv: The `log_path` is a local log storage whereas `LTP_RESULT_PATH` and `LTP_OUTPUT_PATH` point to remote node location. For example, "log_path = runtime\log\20220614\20220614-083040-542\tests\20220614-083134-777-ltp_lite". Also, if the ltp test case is run again, previous results and output are first deleted. Can you please clarify if this resolves your concern?

microsoft/testsuites/ltp/ltp.py
squirrelsc: use node.os.shell.copy_back. It can copy a file back to lisa.

dsrivastavv: fixed. Much faster than using `cat`

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,357 @@
...
+        sysctl.run("-p")
+
+        # define regular stable releases in order to avoid unstable builds
+        # https://github.com/linux-test-project/ltp/tags
+        # 'ltp_version_git_tag' is passed from Test Definition in
+        # .\XML\TestCases\CommunityTests.xml. 'ltp_version_git_tag' default
+        # value is defined in .\XML\Other\ReplaceableTestParameters.xml. You can
+        # run the ltp test with any tag using LISAv2's Custom Parameters feature.
+        build_dir = self.node.find_partition_with_freespace(
+            self.BUILD_REQUIRED_DISK_SIZE_IN_GB
```
LiliDeng, line 288: I run the test, and it picked up disk /dev/sdb1 as build_dir, it is expected?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,357 @@
...
+        sysctl.write("vm.dirty_ratio", "10")
+        sysctl.write("vm.dirty_background_ratio", "5")
+        sysctl.run("-p")
+
+        # define regular stable releases in order to avoid unstable builds
+        # https://github.com/linux-test-project/ltp/tags
+        # 'ltp_version_git_tag' is passed from Test Definition in
+        # .\XML\TestCases\CommunityTests.xml. 'ltp_version_git_tag' default
+        # value is defined in .\XML\Other\ReplaceableTestParameters.xml. You can
+        # run the ltp test with any tag using LISAv2's Custom Parameters feature.
```
LiliDeng: this is for lisav2, I don't think it is useful for lisav3?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -0,0 +1,357 @@
...
+            PurePosixPath(self.LTP_OUTPUT_PATH), PurePath(log_path) / "ltp-output.log"
+        )
+
+        # write results to log path
+        self.node.shell.copy_back(
+            PurePosixPath(self.LTP_RESULT_PATH), PurePath(log_path) / "ltp-results.log"
+        )
+
+        # parse results
+        results = self._parse_results(self.LTP_RESULT_PATH)
```
squirrelsc: The file is copied back in previous lines, it doesn't need to read from remote again.

microsoft/testsuites/ltp/ltp.py
dsrivastavv: Yes. The test can run on any disk except os disk

microsoft/testsuites/ltp/ltp.py
dsrivastavv: fixed

microsoft/testsuites/ltp/ltp.py
dsrivastavv: fixed

LiliDeng (approved):

squirrelsc (approved):

PR #2061, opened 2022-06-09
add actionable node.nics exception messages

mcgov:
node.nics get_nic_by_index doesn't catch indexerror or keyerror currently
This makes for confusing output. Add a try except to give a better description of the error in this case.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -185,7 +185,23 @@ class Nics(InitializableMixin):
         # when there are more than one nic on the system
         number_of_nics = len(self.get_upper_nics())
         assert_that(number_of_nics).is_greater_than(0)
-        return self.nics[self.get_upper_nics()[index]]
+        try:
+            nic_name = self.get_upper_nics()[index]
+        except IndexError:
+            raise LisaException(
+                f"Attempted get_upper_nics()[{index}], only "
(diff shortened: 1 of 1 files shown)


mcgov:
specifically to help with this pattern:
```
    def verify_xdp_multiple_nics(self, node: Node) -> None:
        xdpdump = get_xdpdump(node)
        for i in range(3):
            nic_info = node.nics.get_nic_by_index(i)
            output = xdpdump.test_by_ping(nic_name=nic_info.upper)

            self._verify_xdpdump_result(output)

```

lisa/nic.py
```diff
@@ -185,7 +185,21 @@ def get_nic_by_index(self, index: int = -1) -> NicInfo:
         # when there are more than one nic on the system
         number_of_nics = len(self.get_upper_nics())
         assert_that(number_of_nics).is_greater_than(0)
-        return self.nics[self.get_upper_nics()[index]]
+        try:
```
squirrelsc, line 188: It looks the first one capture the exception on first line, and second capture on second line. To improve readability, it's better to try/except for each line. Ideally, the try part should be as short as possible.

mcgov: @squirrelsc does the current version look good?

squirrelsc (approved):

PR #2064, opened 2022-06-13
restore env after testing

LiliDeng:
@mcgov check_testpmd_is_running my understanding is that when there is pids returned then indicates testpmd is running otherwise there is no testpmd running, correct?

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -17,7 +17,7 @@ from lisa import (
 )
 from lisa.features import Sriov
 from lisa.testsuite import simple_requirement
-from lisa.tools import Echo, Git, Ip, Kill, Lsmod, Make, Modprobe
+from lisa.tools import Echo, Git, Ip, Kill, Lsmod, Make, Modprobe, Service
 from microsoft.testsuites.dpdk.dpdknffgo import DpdkNffGo
 from microsoft.testsuites.dpdk.dpdkovs import DpdkOvs
 from microsoft.testsuites.dpdk.dpdkutil import (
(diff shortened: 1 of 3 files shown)


squirrelsc (approved):

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -609,3 +609,11 @@ def verify_uio_binding(
     def _force_dpdk_default_source(self, variables: Dict[str, Any]) -> None:
         if not variables.get("dpdk_source", None):
             variables["dpdk_source"] = "https://dpdk.org/git/dpdk-stable"
+
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
```
mcgov, line 613: Would we be able to get away with just a reboot? I'm not sure if the modprobe and driver binds persist, since we don't know the state on failure it might be easier to just reset.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -302,7 +302,7 @@ def run_for_n_seconds(self, cmd: str, timeout: int) -> str:
 
     def check_testpmd_is_running(self) -> bool:
         pids = self.node.tools[Pidof].get_pids(self.command, sudo=True)
-        return len(pids) == 0
+        return len(pids) > 0
```
mcgov, line 305: Woah 100% this is needed

microsoft/testsuites/dpdk/dpdksuite.py
LiliDeng: reboot also works.

squirrelsc: reboot takes longer time. But if some distros require reboot, we have to add it.

LiliDeng: I'd like to merge it now.

PR #2065, opened 2022-06-13
Make '(initramfs)' log the only indication of the initramfs status

lubaihua33:
There might be "blk_update_request: I/O error." or ".*exception.*" lines in the log for some images, but they might not cause the system to enter into initramfs. Only there is "(initramfs)" line, the system will enter into initramfs. 

We separate initramfs_logs into two variables, filesystem_exception_logs collects the exception logs to show them in the result's 'Message', initramfs_logs is the indication of initramfs.

### DIFF
--- a/lisa/features/serial_console.py
+++ b/lisa/features/serial_console.py
@@ -35,15 +35,19 @@ class SerialConsole(Feature):
     # blk_update_request: I/O error, dev sdc, sector 0
     # ata1.00: exception Emask 0x0 SAct 0x0 SErr 0x0 action 0x0
     # Failure: File system check of the root filesystem failed
-    # (initramfs)
-    initramfs_patterns: List[Pattern[str]] = [
+    # We can use these patterns to show more information in result's 'Message'
+    filesystem_exception_patterns: List[Pattern[str]] = [
         re.compile(r"^(.*blk_update_request: I/O error.*)$", re.MULTILINE),
         re.compile(r"^(.*exception.*)$", re.MULTILINE),
         re.compile(
             r"^(.*Failure: File system check of the root filesystem failed.*)$",
             re.MULTILINE,
         ),
-        re.compile(r"^(.*\(initramfs\).*)$", re.MULTILINE),
+    ]
+
+    # (initramfs)
+    initramfs_patterns: List[Pattern[str]] = [
+        re.compile(r"^(.*\(initramfs\))", re.MULTILINE),
     ]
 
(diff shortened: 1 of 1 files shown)


lisa/features/serial_console.py
```diff
@@ -156,6 +159,15 @@ def check_initramfs(
         self._node.log.debug("checking initramfs in serial log...")
         content: str = self.get_console_log(saved_path=saved_path, force_run=force_run)
 
+        filesystem_exception_logs = [
```
dsrivastavv, line 163: [nit] maybe split this into multiple statements

lisa/features/serial_console.py
```diff
@@ -35,15 +35,18 @@ class SerialConsole(Feature):
...
         re.compile(r"^(.*blk_update_request: I/O error.*)$", re.MULTILINE),
         re.compile(r"^(.*exception.*)$", re.MULTILINE),
         re.compile(
             r"^(.*Failure: File system check of the root filesystem failed.*)$",
             re.MULTILINE,
         ),
-        re.compile(r"^(.*\(initramfs\).*)$", re.MULTILINE),
+    ]
+
+    initramfs_patterns: List[Pattern[str]] = [
```
dsrivastavv, line 49: Can you please add an example pattern matching this regex?

squirrelsc (dismissed):

lisa/features/serial_console.py
lubaihua33: I think we don't need to split it into multiple statements.

lisa/features/serial_console.py
lubaihua33: Have added. Thank you.

LiliDeng (approved):

lisa/features/serial_console.py
```diff
@@ -35,15 +35,19 @@ class SerialConsole(Feature):
...
         re.compile(r"^(.*blk_update_request: I/O error.*)$", re.MULTILINE),
         re.compile(r"^(.*exception.*)$", re.MULTILINE),
         re.compile(
             r"^(.*Failure: File system check of the root filesystem failed.*)$",
             re.MULTILINE,
         ),
-        re.compile(r"^(.*\(initramfs\).*)$", re.MULTILINE),
+    ]
+
+    # (initramfs)
```
LiliDeng, line 48: if more lines here, that would be good.

squirrelsc (approved):

PR #2066, opened 2022-06-13
try to fix centos 8 repo issue by setting skip_if_unavailable as true.

LiliDeng:


lisa/operating_system.py
```diff
@@ -1319,7 +1319,7 @@ def _package_exists(self, package: str) -> bool:
     def _is_package_in_repo(self, package: str) -> bool:
         if self._first_time_installation:
             self._initialize_package_installation()
-            self._first_time_installation = True
+            self._first_time_installation = False
```
squirrelsc, line 1322: Why change it to false? This is to avoid call repo updates many times.

LiliDeng: I think last time, I changed this, but find, we should set it as false? if it is true, then repo updates many times?

squirrelsc (approved):

PR #2067, opened 2022-06-13
Fix failure of kexec tool installation from source code for Debian 9

lubaihua33:


lisa/tools/gcc.py
```diff
@@ -24,5 +24,8 @@ def compile(self, filename: str, output_name: str = "") -> None:
 
     def _install(self) -> bool:
         posix_os: Posix = cast(Posix, self.node.os)
-        posix_os.install_packages("gcc")
+        if isinstance(self.node.os, Debian):
+            posix_os.install_packages("gcc", signed=False)
```
LiliDeng: have you run the latest code, it used signed = false by default now.

lubaihua33: Thanks for the comment. I deleted the change of gcc.py.

squirrelsc (dismissed):

lisa/tools/find.py
```diff
@@ -22,6 +22,7 @@ def find_files(
         start_path: PurePath,
         name_pattern: str = "",
         path_pattern: str = "",
+        find_type: str = "",
```
dsrivastavv: `type` should be enough. We already have `find` prefix in function name

dsrivastavv: Also, this should be Optional[str] datatype

lisa/tools/kdump.py
```diff
@@ -98,12 +98,15 @@ def _install_from_src(self) -> None:
...
+        find_tool = self.node.tools[Find]
+        kexec_source_folder = find_tool.find_files(
+            tool_path, name_pattern="kexec-tools*", find_type="d"
+        )
+        code_path = tool_path.joinpath(kexec_source_folder[0])
         self.node.tools[Gcc]
         make = self.node.tools[Make]
-        self.node.execute("./configure", cwd=code_path).assert_exit_code()
-        make.make_install(cwd=code_path)
+        self.node.execute("./configure", cwd=code_path, sudo=True).assert_exit_code()
```
dsrivastavv: You can use `expected_exit_code` and `expected_exit_code_failure_message` arguments in `node.execute` rather than `assert_exit_code()`

lubaihua33: Have added.

lisa/tools/find.py
lubaihua33: Have changed to 'type'.  The datatype str is enough I think.

lisa/tools/kdump.py
```diff
@@ -98,15 +98,30 @@ def _install_from_src(self) -> None:
...
         make = self.node.tools[Make]
-        self.node.execute("./configure", cwd=code_path).assert_exit_code()
-        make.make_install(cwd=code_path)
         self.node.execute(
-            "yes | cp -f /usr/local/sbin/kexec /sbin/", sudo=True, shell=True
-        ).assert_exit_code()
+            "./configure",
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                "It is failed to compile the kexec-tools source code"
```
LiliDeng: fail to run configure

lisa/tools/kdump.py
```diff
@@ -98,15 +98,30 @@ def _install_from_src(self) -> None:
         kexec_tar = wget.get(self._kexec_repo, str(tool_path))
         tar = self.node.tools[Tar]
         tar.extract(kexec_tar, str(tool_path))
-        kexec_source = tar.get_root_folder(kexec_tar)
-        code_path = tool_path.joinpath(kexec_source)
+        find_tool = self.node.tools[Find]
+        kexec_source_folder = find_tool.find_files(
+            tool_path, name_pattern="kexec-tools*", type="d"
+        )
+        code_path = tool_path.joinpath(kexec_source_folder[0])
```
LiliDeng, line 105: the tar ball has a fixed name, we can use it directly?

lisa/tools/kdump.py
lubaihua33: Have changed.

lisa/tools/kdump.py
lubaihua33: We don't need to care about the name because it may vary. Using 'd' type will get the folder after unzipping more accurately.

LiliDeng (approved):

squirrelsc (approved):

PR #2068, opened 2022-06-13
fix get hv_vmbus version issue, the raw output contains "\r\n"

LiliDeng:


lisa/tools/modinfo.py
```diff
@@ -59,7 +59,11 @@ def get_version(
             no_info_log=True,
             no_error_log=True,
         )
-        return get_matched_str(output, self._version_pattern)
+        for line in output.splitlines():
```
squirrelsc: Can it be fixed by add "re.M" on the pattern?

lisa/tools/modinfo.py
```diff
@@ -59,7 +59,11 @@ def get_version(
...
+            version = get_matched_str(line, self._version_pattern)
+            if len(version) > 0:
+                return version
+        return ""
```
dsrivastavv: If no version is found, should it throw an error? or is this behaviour normal for some distros?

LiliDeng: some modules don't have version

squirrelsc (approved):

lisa/tools/modinfo.py
LiliDeng: yes, fixed by adding re.M, updated and tested

PR #2069, opened 2022-06-13
DPDK on Debian: buster 5.4 missing ib drivers

mcgov:
Buster doesn't have an ib config 'out of box' on older versions and recommends using the latest kernel for backports.

Add a check for buster and update if the kernel version doesn't ship a configuration for ib and rdma.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -587,8 +587,25 @@ class DpdkTestpmd(Tool):
         else:
             mellanox_drivers = ["mlx5_core", "mlx5_ib"]
         modprobe = self.node.tools[Modprobe]
-        if isinstance(self.node.os, Debian):
-            modprobe.load("rdma_cm")
+        if isinstance(self.node.os, Debian) and not isinstance(self.node.os, Ubuntu):
+            # NOTE: debian buster doesn't include rdma and ib drivers
+            # on 5.4 specifically for linux-image-cloud:
+            # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1012639
+            # for backports on this release we should update the kernel to latest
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -62,6 +62,10 @@ class DpdkTestpmd(Tool):
         r"dpdk-(?P<major>[0-9]+)\.(?P<minor>[0-9]+)"
     )
 
+    _debian_get_arch_from_kernel_version_regex = re.compile(
```
squirrelsc: Add examples above. And there is hardware_platform, which is arch, in kernel_information already.

mcgov: I think I'm actually looking for the 'prerelease' or 'build' field in the semver. It's used for the package naming scheme for linux-image-[build version]. I'll poke around and see if it's picked up by the kernel tool.

squirrelsc (dismissed):

squirrelsc (dismissed):

squirrelsc (approved):

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -587,8 +587,25 @@ def _load_drivers_for_dpdk(self) -> None:
         else:
             mellanox_drivers = ["mlx5_core", "mlx5_ib"]
         modprobe = self.node.tools[Modprobe]
-        if isinstance(self.node.os, Debian):
-            modprobe.load("rdma_cm")
+        if isinstance(self.node.os, Debian) and not isinstance(self.node.os, Ubuntu):
+            # NOTE: debian buster doesn't include rdma and ib drivers
```
LiliDeng, line 591: @mcgov hi, this change will skip testing run on ubuntu.

PR #2070, opened 2022-06-13
Update dpdk_nff_go Distro checks

kamalca:
Previous check allowed Ubuntu 16.04 to run because Ubuntu is a subclass of Debian and version > 10.0.0

Unsupported distros are now skipped instead of failing

Small refactoring

### DIFF
--- a/microsoft/testsuites/dpdk/dpdknffgo.py
+++ b/microsoft/testsuites/dpdk/dpdknffgo.py
@@ -45,19 +45,28 @@ class DpdkNffGo(Tool):
     @property
     def can_install(self) -> bool:
         # nff-go only implemented for debain in lisav2
-        return isinstance(self.node.os, Debian)
+        # but this will be tested for during _install function
+        return True
 
     def _install(self) -> bool:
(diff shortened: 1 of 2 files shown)


mcgov (dismissed):
lgtm

microsoft/testsuites/dpdk/dpdknffgo.py
```diff
@@ -45,19 +45,28 @@ def dependencies(self) -> List[Type[Tool]]:
...
-        os = self.node.os
+        node = self.node
+        os = node.os
         version = os.information.version
-        if not (
-            (isinstance(os, Ubuntu) and version >= "18.4.0")
-            or (isinstance(os, Debian) and version >= "10.0.0")
-        ):
+        if isinstance(os, Ubuntu):
+            if not version >= "18.4.0":
```
squirrelsc: nit. Usually, we should avoid using "not" in condition, unless the logic is clearer. In this case `version < "18.4.0"` is better.

squirrelsc (approved):

PR #2071, opened 2022-06-13
fix some failures in dpdk ring ping

kamalca:
Fixed time out error and git sort error

### DIFF
--- a/lisa/tools/git.py
+++ b/lisa/tools/git.py
@@ -6,6 +6,7 @@ import re
 from typing import List
 
 from assertpy import assert_that
+from semver import VersionInfo
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix, Suse
@@ -26,6 +27,7 @@ class Git(Tool):
     CERTIFICATE_ISSUE_PATTERN = re.compile(
         r"server certificate verification failed", re.M
     )
+    VERSION_PATTERN = re.compile(r"[0-9]+\.[0-9]+\.[0-9]+")
 
     @property
     def command(self) -> str:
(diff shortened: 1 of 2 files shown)


lisa/tools/git.py
```diff
@@ -36,15 +36,19 @@ def can_install(self) -> bool:
         return True
 
     def _install(self) -> bool:
+        package_name = self.command
         if isinstance(self.node.os, Suse):
-            self.node.os.install_packages("git-core")
+            package_name = "git_core"
```
LiliDeng: git-core?

mcgov: This was a typo from my version of the branch I think. Let's make sure we test on Suse before merging.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -542,7 +542,7 @@ def _install(self) -> bool:
         node.execute(
             "ninja",
             cwd=self.dpdk_build_path,
-            timeout=1200,
+            timeout=1800,
```
mcgov, line 545: Can you share the logs to the tests that are failing after 20? 30 minutes seems very long

mcgov: Edit, 30 minutes is needed for the build in some cases

squirrelsc (dismissed):

lisa/tools/git.py
```diff
@@ -215,3 +227,8 @@ def get_tag(
             return tags[-1]
         else:
             return tags[0]
+
+    def get_version(self) -> VersionInfo:
```
kamalca, line 231: @squirrelsc Would you prefer me to add this functionality to Tool instead of Git? I'm not sure whether `--version` is considered universal to all Linux commands, but this code seems pretty generic.

squirrelsc: Let's see if there are more requirements. The implementation is not common, even the return value. The VersionInfo need a sematic version string, but some tools don't follow it.

squirrelsc (approved):

PR #2072, opened 2022-06-14
DPDK bugfix: don't skip Ubuntu

mcgov:
IF condition previously excluded Ubuntu and would result in UnsupportedDistro being thrown.
Add explicit case for Ubuntu before Debian to jump out of the IF

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -587,7 +587,10 @@ class DpdkTestpmd(Tool):
         else:
             mellanox_drivers = ["mlx5_core", "mlx5_ib"]
         modprobe = self.node.tools[Modprobe]
-        if isinstance(self.node.os, Debian) and not isinstance(self.node.os, Ubuntu):
+        if isinstance(self.node.os, Ubuntu):
+            # Ubuntu shouldn't need any special casing, skip to loading rdma/ib
+            pass
+        elif isinstance(self.node.os, Debian):
             # NOTE: debian buster doesn't include rdma and ib drivers
             # on 5.4 specifically for linux-image-cloud:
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -587,7 +587,9 @@ def _load_drivers_for_dpdk(self) -> None:
         else:
             mellanox_drivers = ["mlx5_core", "mlx5_ib"]
         modprobe = self.node.tools[Modprobe]
-        if isinstance(self.node.os, Debian) and not isinstance(self.node.os, Ubuntu):
+        if isinstance(self.node.os, Ubuntu):
+            pass
```
squirrelsc, line 592: Add comments explain like Ubuntu doesn't need prerequisite setup.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -587,7 +587,9 @@ def _load_drivers_for_dpdk(self) -> None:
...
+        elif isinstance(self.node.os, Debian):
```
LiliDeng, line 593: does previous load module no need any more?
image

mcgov: I don't think we need to manually load it as long as the correct packages are installed and ib_uverbs loads. 
I want to remove as many of these special cases as possible, if we find they were needed we can add it back.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: It does, it just doesn't need any special casing up at the top of the function. I'll comment.

squirrelsc (approved):

PR #2075, opened 2022-06-16
Revert: QEMU: Ensure secure-boot is disabled.

cwize1:
Turns out the "QEMU: Ensure secure-boot is disabled" change doesn't actually do
anything.


cwize1:
It turns out the problem I was experiencing was a bug in the OVMF (UEFI) firmware that only occurs when a VM is allocated more than 2GiB of memory.

In my testing of the previous change, I wasn't consistent with my testing. So, I thought the change fixed the problem when in reality I was just running a test that used 2 GiB VMs instead of 3 GiB VMs.

LiliDeng (approved):

squirrelsc (approved):

PR #2076, opened 2022-06-17
Fix repo check test case

LiliDeng:
- update pattern to match repolist output.
- change repo id into lower for comparison.
- rhui repo only show up in RHEL.


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -673,8 +672,8 @@ def verify_repository_installed(self, node: Node) -> None:
                     "Updates repository should be present",
                 ).is_true()
 
-            # verify that atleast five repositories are present in Redhat
-            if isinstance(node.os, Redhat):
+            # verify that at least five repositories are present in Redhat
+            if isinstance(node.os, Redhat) and "Red Hat" == node.os.information.vendor:
```
squirrelsc: If it needs to check for only redhat, the logic is `if type(node.os) == Redhat`.

LiliDeng: update

squirrelsc (approved):

PR #2077, opened 2022-06-19
Fix docker installation failure on RHEL 8

LiliDeng:


lisa/operating_system.py
```diff
@@ -1404,10 +1398,9 @@ def _initialize_package_installation(self) -> None:
             #  issue
             cmd_results = self._node.execute("yum repolist -v", sudo=True)
             if 0 != cmd_results.exit_code:
-                cmd_results = self._node.execute(
-                    "yum-config-manager --save --setopt=skip_if_unavailable=true",
-                    sudo=True,
-                )
+                from lisa.tools import YumConfigManager
```
squirrelsc: Avoid importing inline. Move the tool to base_tools, which is used in operating_system. And then move the import line to the top of the file.

lisa/base_tools/__init__.py
```diff
@@ -3,5 +3,6 @@
 from .sed import Sed
 from .uname import Uname
 from .wget import Wget
+from .yum_config_manager import YumConfigManager
 
-__all__ = ["Uname", "Sed", "Wget", "Cat", "Rpm"]
```
squirrelsc, line 7: Add it in the `__init__` under tools like other base tools, so it's easy to be used as normal tools.

lisa/base_tools/yum_config_manager.py
```diff
@@ -0,0 +1,48 @@
...
+    @property
+    def command(self) -> str:
+        return "yum-config-manager"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        from lisa.operating_system import Posix
```
squirrelsc: Instead of adding the import for type, you can add `# type: ignore` at the line, which need the type.

lisa/base_tools/yum_config_manager.py
```diff
@@ -0,0 +1,45 @@
+from lisa.executable import Tool
+from lisa.operating_system import Posix
```
squirrelsc: This may cause cycle loop. You can add `# type: ignore` on `posix_os.install_packages(package_name)`

squirrelsc: This may cause cycle loop. You can add `# type: ignore` on `posix_os.install_packages(package_name)`

lisa/base_tools/yum_config_manager.py
```diff
@@ -0,0 +1,49 @@
...
+        return "yum-config-manager"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        posix_os: Posix = self.node.os  # type: ignore
+        package_name = "yum-utils"
+        posix_os.install_packages(package_name)
```
squirrelsc: Try to add `# type: ignore` on this line, there is no need on `from lisa.operating_system import Posix` and `posix_os: Posix = self.node.os`

LiliDeng: update

squirrelsc (approved):

PR #2079, opened 2022-06-20
force sync after deleting file

LiliDeng:


squirrelsc (approved):

squirrelsc:
The commit message is not clear enough. How the "force sync" is proceeded? Which file is deleted?

PR #2080, opened 2022-06-20
Support jump boxes for development environment.

squirrelsc:
Runbook example,

```yaml
dev:
  mock_tcp_ping: True
  jump_boxes:
    - private_key_file: $(admin_private_key_file)
      address: 20.236.23.33
    - private_key_file: $(admin_private_key_file)
      address: 20.236.23.40
```

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -5,6 +5,7 @@ import copy
 from dataclasses import dataclass, field
 from enum import Enum
 from functools import partial
+from pathlib import Path
 from typing import Any, Dict, List, Optional, Type, TypeVar, Union, cast
 
 from dataclasses_json import (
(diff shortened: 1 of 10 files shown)


lisa/schema.py
```diff
@@ -1309,6 +1310,67 @@ def type_name(cls) -> str:
...
+        add_secret(self.username, PATTERN_HEADTAIL)
+        add_secret(self.password)
+        add_secret(self.private_key_file)
+
+        if not self.password and not self.private_key_file:
+            raise LisaException(
+                "at least one of password or private_key_file need to be set when "
+                "connecting"
+            )
+        elif not self.private_key_file:
```
mcgov, line 1337: Would this disallow use of an empty password? It's terrible practice to use empty passwords and I am onboard with just not supporting that. If it's an undocumented requirement that's a side effect of an internal library we should make sure to warn at runtime and also document it.

mcgov (commented):
Super excited to have jumpbox support! Only note is a question of updating documentation for password requirements

lisa/schema.py
squirrelsc: If there is no password and no private key, the exception raises on Ln1333. This `elif` is to handle non-empty password. The next `else` of Ln 1341 to handle non-empty private key. So with them all, the empty password is not allowed.

lisa/util/shell.py
```diff
@@ -499,6 +527,83 @@ def _purepath_to_str(
...
+            if is_trace_enabled:
+                jb_logger.debug(f"creating connection from source: {runbook} ")
+            client = paramiko.SSHClient()
+            client.set_missing_host_key_policy(paramiko.MissingHostKeyPolicy())
+            client.connect(
+                hostname=runbook.address,
+                port=runbook.port,
+                username=runbook.username,
+                password=runbook.password,
+                key_filename=runbook.private_key_file,
```
dsrivastavv, line 548: Does all jumpbox need to have same key?

dsrivastavv:
What was the need for adding this feature? Is this required for some testcase?

lisa/util/shell.py
squirrelsc: It doesn't need. The key can be configured for each jumpbox like address or port.

squirrelsc:
> What was the need for adding this feature? Is this required for some testcase?

It's not required by test cases. I got feedback that some developers work from home, and need to bypass corp VPN. The jumpboxes are supported by ssh_config, but LISA doesn't use the ssh_config. So we need to implement it in LISA to support jumpbox scenarios.

PR #2082, opened 2022-06-21
fix netperf compile issue when gcc version >= 10

LiliDeng:


lisa/tools/gcc.py
```diff
@@ -22,6 +22,14 @@ def compile(self, filename: str, output_name: str = "") -> None:
         else:
             self.run(filename)
 
+    def version(self) -> int:
```
squirrelsc: change it like "verbose-noun", for example "get_version", because it's a method.

lisa/tools/gcc.py
```diff
@@ -22,6 +22,14 @@ def compile(self, filename: str, output_name: str = "") -> None:
...
+        output = self.run(
+            "-dumpversion",
+            expected_exit_code=0,
+            expected_exit_code_failure_message="fail to get gcc dumpversion",
+        )
+        return int(output.stdout.split(".")[0])
```
squirrelsc: how about return the "VersionInfo" type? Maybe the minor version is needed in future.

lisa/tools/gcc.py
LiliDeng: update

lisa/tools/gcc.py
LiliDeng: update

squirrelsc (approved):

Issue #1804, opened 2022-03-15
Move exit code check to run function


dsrivastavv:
Earlier:
https://github.com/microsoft/lisa/blob/236211f7ff36af67ece2d3384b322593df64b64c/lisa/tools/lsmod.py#L38-L48

New format:
https://github.com/microsoft/lisa/blob/8229cc9b89954de6f5407e35a138681fd2ef0118/lisa/tools/modprobe.py#L53-L62

snosratiershad:
@somil55, should this change applied wildly on everywhere we check exit codes?

dsrivastavv:
@snosratiershad yes

PR #2083, opened 2022-06-21
refactor: move exit code check to excepted_exit_code parameter

snosratiershad:
As discussed in [this issue](https://github.com/microsoft/lisa/issues/1804) I moved syntax on where development logic wouldn't affect.

### DIFF
--- a/lisa/base_tools/wget.py
+++ b/lisa/base_tools/wget.py
@@ -81,10 +81,13 @@ class Wget(Tool):
                 f" stdout: {command_result.stdout}"
             )
         actual_file_path = self.node.execute(
-            f"ls {download_file_path}", shell=True, sudo=sudo
+            f"ls {download_file_path}",
+            shell=True,
+            sudo=sudo,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"File {actual_file_path} "
+                                                "doesn't exist."
         )
-        if actual_file_path.exit_code != 0:
-            raise LisaException(f"File {actual_file_path} doesn't exist.")
         if executable:
             self.node.execute(f"chmod +x {actual_file_path}", sudo=sudo)
 
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -237,11 +237,7 @@ class LisDriver(Tool):
 
     def _install(self) -> bool:
         result = self.install_from_iso()
-        if result.exit_code != 0:
-            raise LisaException(
-                f"Unable to install the LIS RPMs! exit_code: {result.exit_code}"
-                f"stderr: {result.stderr}"
-            )
+        result.assert_exit_code(0, message="Unable to install the LIS RPMs!")
         self.node.reboot(360)
         return True
 
--- a/lisa/tools/date.py
+++ b/lisa/tools/date.py
@@ -19,9 +19,10 @@ class Date(Tool):
 
     def current(self, no_error_log: bool = True) -> datetime:
         # always force run to get current date time.
-        command_result = self.run(no_error_log=no_error_log, force_run=True, timeout=10)
-        if command_result.exit_code != 0:
-            raise LisaException(
-                f"'Date' return non-zero exit code: {command_result.stderr}"
-            )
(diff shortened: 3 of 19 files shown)


lisa/sut_orchestrator/azure/tools.py
```diff
@@ -237,11 +237,7 @@ def _check_exists(self) -> bool:
 
     def _install(self) -> bool:
         result = self.install_from_iso()
-        if result.exit_code != 0:
-            raise LisaException(
-                f"Unable to install the LIS RPMs! exit_code: {result.exit_code}"
-                f"stderr: {result.stderr}"
```
squirrelsc, line 243: This is not easy to move, because the stderr is not presented in the exception by default. The exception message is more obvious than log, so it's worth to include the stderr. But the default `assert_exit_code` cannot include it, because there may be a lot of stderr/stdout, and doesn't help. So when it needs the stderr, we cannot use the `assert_exit_code`.

lisa/tools/date.py
```diff
@@ -19,9 +19,10 @@ def _check_exists(self) -> bool:
 
     def current(self, no_error_log: bool = True) -> datetime:
         # always force run to get current date time.
-        command_result = self.run(no_error_log=no_error_log, force_run=True, timeout=10)
-        if command_result.exit_code != 0:
-            raise LisaException(
-                f"'Date' return non-zero exit code: {command_result.stderr}"
```
squirrelsc, line 25: as above comments.

lisa/tools/dhclient.py
```diff
@@ -67,10 +67,12 @@ def renew(self, interface: Optional[str] = None) -> None:
                 f"-r {interface} && dhclient {interface}",
                 shell=True,
                 sudo=True,
+                expected_exit_code=0
```
squirrelsc, line 70: Not sure if it brings a regression. In current logic, we don't care the result of renew.

snosratiershad: You are right, it's my mistake. going to revert.

lisa/sut_orchestrator/azure/tools.py
snosratiershad: Should I add a string formatting with `stderr` and `exit_code` when presenting `excepted_exit_code_failure_message`? for example:
```python
excepted_exit_code_failure_message = excepted_exit_code_failure_message % {"stderr": self.stderr, "exit_code": self.exit_code}
```
I saw something like it before, in https://github.com/pallets/click/blob/main/src/click/decorators.py line:452

squirrelsc: The stdout/stderr shouldn't be included by default, because the output may be very long, not readable. Before we find a better way, it's worth to handle separately.

lisa/tools/dhclient.py
squirrelsc: It's not reverted correctly. The `f` should be kept, and the `expected_exit_code=0` should be removed

lisa/base_tools/wget.py
```diff
@@ -81,10 +81,13 @@ def get(
                 f" stdout: {command_result.stdout}"
             )
         actual_file_path = self.node.execute(
-            f"ls {download_file_path}", shell=True, sudo=sudo
+            f"ls {download_file_path}",
+            shell=True,
+            sudo=sudo,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"File {actual_file_path} "
+                                                "doesn't exist."
```
dsrivastavv, line 89: Can you please setup `black` and `isort`? And then run `make check` to verify code formatting

snosratiershad:
@squirrelsc @somil55 I going to fix them all in another branch and then rebase it with PR. sorry for late response.

squirrelsc:
> @squirrelsc @somil55 I going to fix them all in another branch and then rebase it with PR. sorry for late response.

No problem, please make sure the same PR is used. So it's easy to track changes.

dsrivastavv:
@snosratiershad Any updates on this PR?

dsrivastavv:
Closing this due to no activity. @snosratiershad Please feel free to open this again once you push changes

PR #2084, opened 2022-06-21
Add unique id to community test message

dsrivastavv:


lisa/testsuite.py
```diff
@@ -555,6 +555,7 @@ def start(
             case_kwargs.update({"log_path": case_log_path})
             case_kwargs.update({"working_path": case_working_path})
             case_kwargs.update({"part_path": case_part_path})
+            case_kwargs.update({"case_result": case_result})
```
squirrelsc: Rename it to `result` in the list.

dsrivastavv: I chose to stick with complete name because using just `result` is a bit confusing, particularly in the ltp.run_test_case function. Let me know if you feel strongly about this change.

squirrelsc: The names in the list is hard to use actually, because there is no intelligence prompt in IDE. So we should make the names simpler. The "case_result" is not a good name, either "test_case_result", "test_result" or "result". Shorter is better here.

squirrelsc: The "case_" in the code is to differentiate with "suite_". They don't need to be the same.

squirrelsc (approved):

PR #2085, opened 2022-06-22
clean up env after running xfstests and fix log path of xfstests

LiliDeng:


lisa/tools/xfstests.py
```diff
@@ -277,26 +279,31 @@ def check_test_results(self, raw_message: str, log_path: Path) -> None:
             fail_info += find_patterns_in_lines(
                 raw_message, [re.compile(f".*{fail_case}.*$", re.MULTILINE)]
             )[0][0]
-        self.save_xfstests_log(fail_cases.split(), log_path)
+        self.save_xfstests_log(fail_cases.split(), log_path, test_type)
+        self.node.execute(f"rm -rf {self.get_tool_path()}", sudo=True)
```
squirrelsc: Make sure the test is runnable, because the tool may not be installed again in next test case.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -484,6 +486,17 @@ def xfstesting_azure_file_share_validation(
...
 
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
+        node: Node = kwargs.pop("node")
+        for path in [
+            "/dev/mapper/delay-test",
+            "/dev/mapper/huge-test",
+            "/dev/mapper/huge-test-zero",
+        ]:
+            if 0 == node.execute(f"ls -lt {path}", sudo=True).exit_code:
+                log.debug(f"dmsetup remove {path} to clean up env")
```
squirrelsc: This line is not necessary, the next line can output more details.

lisa/tools/xfstests.py
LiliDeng: remove this

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: remove this

squirrelsc (approved):

PR #2088, opened 2022-06-22
Infiniband: Try kernel-devel-$(uname -r) first

kamalca:


lisa/features/infiniband.py
```diff
@@ -224,11 +223,24 @@ def setup_rdma(self) -> None:
...
+                    "kernel-devel-$(uname -r) not found. Trying kernel-devel"
+                )
+                node.os.install_packages("kernel-devel")
+        elif isinstance(node.os, Ubuntu) and node.os.information.version >= "18.4.0":
             node.os.install_packages(list(ubuntu_required_packages))
         else:
-            raise LisaException(f"Unsupported distro: {node.os.name} is not supported.")
+            raise UnsupportedDistroException(
+                node.os,
+                f"Unsupported distro: {node.os.name} {node.os.information.version} is "
```
squirrelsc: The UnsupportedDistroException will give distro versions, so it doesn't need to be in the error message. The error message should describe it from business logic, at least mention like install required packages failed, or the distro is not supported by infiniband.

lisa/features/infiniband.py
```diff
@@ -224,11 +223,24 @@ def setup_rdma(self) -> None:
...
-        elif isinstance(node.os, Ubuntu):
+
+            try:
+                node.os.install_packages("kernel-devel-$(uname -r)")
+            except MissingPackagesException:
+                node.log.debug(
+                    "kernel-devel-$(uname -r) not found. Trying kernel-devel"
+                )
+                node.os.install_packages("kernel-devel")
+        elif isinstance(node.os, Ubuntu) and node.os.information.version >= "18.4.0":
```
dsrivastavv, line 240: Is Debian OS class not supported?

squirrelsc (approved):

lisa/features/infiniband.py
kamalca: At this time, I have only implemented images supported by the [HPC team](https://github.com/Azure/azhpc-images)

PR #2090, opened 2022-06-23
Add docker installation from repo for debian backport images

dsrivastavv:


lisa/tools/docker.py
```diff
@@ -47,7 +47,7 @@ def run_container(
         docker_run_output: str,
     ) -> None:
         self.run(
-            f"run --name {container_name} " f"{image_name} 1> {docker_run_output} 2>&1",
+            f"run --name {container_name} {image_name} 1> {docker_run_output} 2>&1",
```
LiliDeng: is it "1>" is covered in "2>&1" ?

dsrivastavv: This was just a formatting change automatically done by black. But I can take a look

lisa/tools/docker.py
```diff
@@ -65,7 +65,14 @@ def start(self, name: str = "docker") -> None:
...
         if isinstance(self.node.os, Debian):
-            self.node.os.install_packages("docker.io")
+            try:
+                self.node.os.install_packages("docker.io")
+            except Exception as e:
+                self._log.error(
+                    f"Failed to install docker.io: {e}, trying to install docker from"
+                    " repo"
+                )
+                self._install_from_repo()
```
squirrelsc, line 75: How about call this method when it's Debian. and install packages for other distros? That will be more clear.

lisa/tools/docker.py
```diff
@@ -65,7 +65,14 @@ def start(self, name: str = "docker") -> None:
     def _install(self) -> bool:
         service_name = "docker"
         if isinstance(self.node.os, Debian):
-            self.node.os.install_packages("docker.io")
+            try:
```
LiliDeng, line 68: ubuntu distro also use the same logic, right?

lisa/tools/docker.py
dsrivastavv: This is not the recommended installation method. Most of the Debian distribution install using `docker.io` package.

lisa/tools/docker.py
dsrivastavv: Most ubuntu versions will work with "docker.io" package installation

lisa/tools/docker.py
dsrivastavv: I think `run --name {container_name} {image_name} > {docker_run_output} 2>&1` should be enough

squirrelsc (dismissed):

lisa/tools/docker.py
dsrivastavv: fixed. Above is working fine

squirrelsc (approved):

PR #2091, opened 2022-06-24
Change dump path and crashkernel size if the memeory size is big

lubaihua33:
For Redhat, If the system memory size is bigger than 1T, the default size of /var/crash  maybe not be enough to store the dump file, need to change the dump path

System memory is big, we also need to increase the crashkernel size and the timeout time.  If the system memory size is bigger than 1T, we set the crashkernel as 2G, the timeout time is 1200,  if the system memory size is bigger than 6T, the timeout time is 2000. Those are experience values.

### DIFF
--- a/microsoft/testsuites/kdump/kdumpcrash.py
+++ b/microsoft/testsuites/kdump/kdumpcrash.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 import time
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from random import randint
 from typing import cast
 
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -231,15 +232,46 @@ def _check_supported(self, node: Node) -> None:
...
             self._check_supported(node)
         except UnsupportedDistroException as identifier:
             raise SkippedException(identifier)
 
         kdump = node.tools[KdumpBase]
+
+        memory_size = node.execute(
+            "free -h | grep Mem | awk '{print $2}'", shell=True, sudo=True
+        )
+        if "Ti" in memory_size.stdout and float(memory_size.stdout.strip("Ti")) > 1:
```
squirrelsc, line 264: Why not always use the resource disk if it exists?

lubaihua33: Yes, this is a good way. 

I tried to use another location to save the vmcore on Ubuntu and Suse. The dump is all failed. I was trying to figure out if it needs another configuration or if it has some issues when it changes to another location. After understanding the reason, we can change to using the resource disk by default.

I just added this way on Redhat, which can work well on Redhat.

squirrelsc (approved):

PR #2092, opened 2022-06-24
support to specify os disk size

LiliDeng:
WIP, need find out a public Shared Image Gallery to test.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -19,6 +19,7 @@ from typing import Any, Dict, List, Optional, Tuple, Type, Union, cast
 from azure.core.exceptions import HttpResponseError
 from azure.identity import DefaultAzureCredential
 from azure.mgmt.compute.models import (  # type: ignore
+    GalleryImageVersion,
     PurchasePlan,
     ResourceSku,
     RunCommandInput,
@@ -74,6 +75,7 @@ from .common import (
     AzureVmPurchasePlanSchema,
     DataDiskCreateOption,
     DataDiskSchema,
+    SharedImageGallerySchema,
     check_or_create_resource_group,
     check_or_create_storage_account,
     get_compute_client,
@@ -1163,6 +1165,7 @@ class AzurePlatform(Platform):
             AzureNodeSchema, type_name=AZURE
         )
 
+        os_disk_size = 30
         if not azure_node_runbook.name:
             # the max length of vm name is 64 chars. Below logic takes last 45
             # chars in resource group name and keep the leading 5 chars.
@@ -1191,10 +1194,17 @@ class AzurePlatform(Platform):
             azure_node_runbook.vhd = self._get_deployable_vhd_path(
                 azure_node_runbook.vhd, azure_node_runbook.location, log
             )
+            os_disk_size = max(
+                os_disk_size, self._get_vhd_os_disk_size(azure_node_runbook.vhd)
+            )
             azure_node_runbook.marketplace = None
             azure_node_runbook.shared_gallery = None
         elif azure_node_runbook.shared_gallery:
             azure_node_runbook.marketplace = None
+            os_disk_size = max(
+                os_disk_size,
+                self._get_sig_os_disk_size(azure_node_runbook.shared_gallery),
+            )
         elif azure_node_runbook.marketplace:
             # marketplace value is already set in runbook
             pass
@@ -1211,7 +1221,9 @@ class AzurePlatform(Platform):
             image_info = self._get_image_info(
                 azure_node_runbook.location, azure_node_runbook.marketplace
             )
-
+            os_disk_size = max(
+                os_disk_size, image_info.os_disk_image.additional_properties["sizeInGb"]
+            )
             # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
             if image_info.hyper_v_generation:
                 azure_node_runbook.hyperv_generation = int(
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1163,6 +1164,7 @@ def _create_node_runbook(
             AzureNodeSchema, type_name=AZURE
         )
 
+        os_disk_size = 0
```
squirrelsc: Set it to default value, instead of 0.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1237,6 +1241,10 @@ def _create_node_runbook(
                     plan_publisher=plan_publisher,
                 )
 
+        azure_node_runbook.osdisk_size_in_gb = os_disk_size
```
squirrelsc, line 1251: The logic is not straightforward. It should make the os_disk_size right, then assign to another variable. BTW, if the default value is set to 10, does it need to check here again?

```python
        if os_disk_size < 10:
            os_disk_size = 10
        azure_node_runbook.osdisk_size_in_gb = os_disk_size
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1191,10 +1193,12 @@ def _create_node_runbook(
             azure_node_runbook.vhd = self._get_deployable_vhd_path(
                 azure_node_runbook.vhd, azure_node_runbook.location, log
             )
+            os_disk_size = self._get_blob_size(azure_node_runbook.vhd)
```
squirrelsc: How about `os_disk_size = max(os_disk_size, self._get_blob_size(azure_node_runbook.vhd))`?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2089,6 +2097,55 @@ def _enable_ssh_on_windows(self, node: Node) -> None:
         log.debug("SSH script result:")
         log.dump_json(logging.DEBUG, result)
 
+    def _get_blob_size(self, blob_url: str) -> int:
```
squirrelsc: Make the name more accurate, like `_get_vhd_os_disk_size`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2089,6 +2097,55 @@ def _enable_ssh_on_windows(self, node: Node) -> None:
...
+        vhd_blob = BlobClient.from_blob_url(
+            blob_url=blob_url,
+            credential=self.credential,
+        )
+        properties = vhd_blob.get_blob_properties()
+        assert properties.size
+        return int(properties.size / 1024 / 1024 / 1024)
+
+    def _get_sig_size(self, sharedImage: SharedImageGallerySchema) -> int:
```
squirrelsc: It worth to add a method to return the image info.

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: it will check every image/vhd size, if it is too small, will set it as 10 or 30, otherwise use its default size, I tried if the os disk size is 30, when set it as 10 in template, threw exception during deployment. so just hard code doesn't work.

mcgov:
Not sure about how the requirements work, will this allow us to specify requirement: min_disk_size_gb=50 or something similar?
I am so glad we are getting this feature, thank you @LiliDeng for working on it!!

LiliDeng:
> Not sure about how the requirements work, will this allow us to specify requirement: min_disk_size_gb=50 or something similar? I am so glad we are getting this feature, thank you @LiliDeng for working on it!!

by default mariner image is 2 GB, it will be out of space when install the frist package, this is the main reason I created the PR.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2089,6 +2101,72 @@ def _enable_ssh_on_windows(self, node: Node) -> None:
...
+                )
+                if not time:
+                    time = gallery_image.publishing_profile.published_date
+                if gallery_image.publishing_profile.published_date > time:
+                    time = gallery_image.publishing_profile.published_date
+                    found_image = gallery_image
+        return found_image
+
+    def _get_sig_size(self, shared_image: SharedImageGallerySchema) -> int:
+        if not shared_image.resource_group_name:
```
squirrelsc: This logic should be in the _get_sig_info, because it looks sharable.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2089,6 +2101,72 @@ def _enable_ssh_on_windows(self, node: Node) -> None:
...
+                    expand="ReplicationStatus",
+                )
+                if not time:
+                    time = gallery_image.publishing_profile.published_date
+                if gallery_image.publishing_profile.published_date > time:
+                    time = gallery_image.publishing_profile.published_date
+                    found_image = gallery_image
+        return found_image
+
+    def _get_sig_size(self, shared_image: SharedImageGallerySchema) -> int:
```
squirrelsc: As vhd, it should be like `_get_sig_os_disk_size`

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py

squirrelsc (approved):

PR #2094, opened 2022-06-27
add ignorable pattern for dmesg of RHEL9

LiliDeng:


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -192,6 +192,12 @@ class AzureImageStandard(TestSuite):
             r"^(.*WARNING ExtHandler ExtHandler cgroups v2 mounted at.*)$", re.M
         ),
         re.compile(r"^(.*dataloss warning file.*)$", re.M),
+        re.compile(
+            r"(.*temp-disk-dataloss-warning.*Deactivated successfully.*)$",
+            re.M,
+        ),
+        re.compile(r"(.*because of a failed condition check.*)", re.M),
+        re.compile(r"(.*all trigger condition checks failed.*)", re.M),
```
squirrelsc: Are they ignorable? Which component produce them, and is it possible to have serious problems here? I'm asking here because the patterns look general.

LiliDeng: here are the raw messages, I think they can be ignored. 
```
systemd[1]: Arbitrary Executable File Formats File System Automount Point was skipped because of a failed condition check (ConditionPathExists=/proc/sys/fs/binfmt_misc)
systemd[1]: Load Kernel Modules was skipped because all trigger condition checks failed.
systemd[1]: Repartition Root Disk was skipped because all trigger condition checks failed.
```

squirrelsc: So, you can take another part, which is not confusing, like "Arbitrary Executable File Formats File System Automount Point was skipped"

LiliDeng: then I need add many patterns how about 
```was skipped because all trigger condition checks failed```
```was skipped because of a failed condition check```

squirrelsc: How about "was skipped because all trigger condition checks failed"? The key thing is to make sure it won't ignore any serious problem, and easy to understand.

LiliDeng: updated and tested

squirrelsc (approved):

PR #2095, opened 2022-06-27
Handle error in kill if pid is not found

dsrivastavv:
This can happen if parent process is killed, automatically killing child processes

### DIFF
--- a/lisa/tools/kill.py
+++ b/lisa/tools/kill.py
@@ -3,6 +3,7 @@
 
 
 from lisa.executable import Tool
+from lisa.util import LisaException
 from lisa.util.constants import SIGKILL
 
 from .pidof import Pidof
@@ -17,7 +18,9 @@ class Kill(Tool):
     def can_install(self) -> bool:
         return False
 
-    def by_name(self, process_name: str, signum: int = SIGKILL) -> None:
+    def by_name(
+        self, process_name: str, signum: int = SIGKILL, ignore_not_exist: bool = False
+    ) -> None:
 
         # attempt kill by name first
         kill_by_name = self.run(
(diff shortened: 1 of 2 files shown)


lisa/tools/kill.py
```diff
@@ -40,12 +41,15 @@ def by_name(self, process_name: str, signum: int = SIGKILL) -> None:
...
             f"-{signum} {pid}",
             shell=True,
             sudo=True,
             force_run=True,
-            expected_exit_code=0,
-            expected_exit_code_failure_message="fail to run "
-            f"{self.command} -{signum} {pid}",
         )
+
+        if result.exit_code != 0:
```
mcgov, line 51: There's a wrinkle here where `kill` can be either a shell builtin or an executable and the output isn't guaranteed to be consistent. I suppose we will find out which systems don't give us this output once we start using it but... 

It might be better to add a test case for `kill` to `core` to verify this kill tool behaves as we expect and add that test to the tier 1 tests.

We will find out immediately which distros are our edge cases and be able to identify new distros with issues without having to dig through other test results.

dsrivastavv: Is there a way to detect this? For example, does `which kill` help?

squirrelsc: We can add a flag, like ignore_error. When it's true, ignore exit code. If false, raise exception, if it's failed.

mcgov: I think which kill can help, but I'm more concerned about relying on stdout spew.
I think it's fine to merge this but we should also add a small test to identify platforms which don't use this output spew. 
We can do a little work to make a test to check that this works on all platforms OR we can spend a bunch of time looking at why a bunch of tests failed one by one, does that make sense?

dsrivastavv: As discussed, let's merge it to remove current regressions in test. We will add a test to check platforms which don't throw out this error.

lisa/tools/kill.py
```diff
@@ -35,13 +36,20 @@ def by_name(self, process_name: str, signum: int = SIGKILL) -> None:
...
+            if ignore_exit_code and "No such process" in result.stdout:
```
squirrelsc: Remove the check of "No such process". Just use the ignore_exit_code result.

dsrivastavv: This could lead to the test passing on other errors from kill command. We will not mark those nodes for recovery and other tests could potentially run on it

squirrelsc: The words and logic should be consistent. If it needs to ignore the "no such process" only, the variable name should be change like "ignore_not_exist".

squirrelsc (approved):

PR #2096, opened 2022-06-27
Fix NTP package error on redhat > 8

dsrivastavv:
The logic was incorrect. Redhat 8 and greater does not have ntp package

### DIFF
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -191,13 +191,14 @@ class Ltp(Tool):
             self.node.os.install_packages(
                 ["libaio-devel", "libattr", "libcap-devel", "libdb"]
             )
+
+            # db4-utils and ntp are not available in Redhat >= 8.0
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/ltp/ltp.py
```diff
@@ -195,9 +195,7 @@ def _install(self) -> bool:
                 isinstance(self.node.os, Redhat)
                 and self.node.os.information.release >= "8.0"
             ):
-                self.node.os.install_packages(["db4-utils"])
-            else:
-                self.node.os.install_packages(["ntp"])
+                self.node.os.install_packages(["db4-utils", "ntp"])
```
squirrelsc, line 201: As we discussed, make the logic straightforward, or add more comments to explain the expected behavior.

dsrivastavv: I have added the comments to explain this. For the if-else, the current flow should be understandle with comments

squirrelsc (approved):

PR #2098, opened 2022-06-28
Fix docker installation on centos < 8

dsrivastavv:


squirrelsc:
Do you mean <= 7 instead of < 7?

squirrelsc (dismissed):

dsrivastavv:
> Do you mean <= 7 instead of < 7?

```
This is <= 7. The logic is something like:
If redhat >= 8.0:
    install_method_1
else:
    install_method_2
```

squirrelsc:
> > Do you mean <= 7 instead of < 7?
> 
> ```
> This is <= 7. The logic is something like:
> If redhat >= 8.0:
>     install_method_1
> else:
>     install_method_2
> ```

Ok, please make commit message accurate.

squirrelsc (approved):

PR #2099, opened 2022-06-28
DPDK: Mark netvsc pmd tests dirty

kamalca:
Netvsc pmd tests make changes to the interfaces that cause problems in later tests.

Also skip multiprocess test on ConnectX3 hardware due to unsupported functionality.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -253,6 +253,9 @@ def initialize_node_resources(
 
     # netvsc pmd requires uio_hv_generic to be loaded before use
     if pmd == "netvsc":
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -253,6 +253,7 @@ def initialize_node_resources(
 
     # netvsc pmd requires uio_hv_generic to be loaded before use
     if pmd == "netvsc":
+        node.mark_dirty()
```
squirrelsc, line 258: The "mark dirty" should be used after the main test logic, and the environment is dirty, cannot be reused. But here the dirty is used when initializing the environment. It won't affect the current test. Please add more comments, if it's intent.

squirrelsc (approved):

PR #2100, opened 2022-06-28
Test additional Python versions

tonybaloney:
The documentation states that Python 3.8 is the tested version, but newer versions are recommended.

This PR adds the newer versions to the test matrix.

pyproject.toml has also been updated as it binds the python version to only the 3.8 series.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -9,19 +9,20 @@ on:
       - main
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.os }} with python ${{ matrix.python-version }}
     strategy:
       matrix:
         os: [ubuntu-20.04, windows-2019]
+        python-version: ["3.8", "3.9", "3.10", "3.11-dev"]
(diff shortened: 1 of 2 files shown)


squirrelsc:
They're waiting near one hour, is it caused by too many workflows in the pipeline?

squirrelsc:
try to reopen the pr to trigger checks.

tonybaloney:
Check that the "Approve and Run" button is not showing toward the bottom of this PR

squirrelsc (approved):

squirrelsc:
> Check that the "Approve and Run" button is not showing toward the bottom of this PR

You are right, the button doesn't show on my page. Do you know how to fix it?

tonybaloney:
@squirrelsc I think Github actions was broken. One way to fix this is to go and cancel the first actions run from https://github.com/microsoft/lisa/actions/runs/2579496054

LiliDeng:
once it merges, does every job will be in queue like this?

squirrelsc:
> once it merges, does every job will be in queue like this?

If we cannot figure out the problem, we shouldn't merge it.

PR #2102, opened 2022-06-29
DPDK: GPU and HPC SKUs not supported

kamalca:
GPU and HPC hardware is incompatible with DPDK tests and not a DPDK use case.


squirrelsc (approved):

mcgov (approved):
Let's merge this but can we also add the perf tests in dpdkperf? It can be a second PR.

PR #2103, opened 2022-06-30
Add coreutils timeout tool

mcgov:
Timeout tool for scheduling a command with timeout. Optional parameter to send a kill signal after the first signal.


lisa/tools/timeout.py
```diff
@@ -0,0 +1,30 @@
...
+    @property
+    def command(self) -> str:
+        return "timeout"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def run_with_timeout(
+        self, command: str, signal: str, timeout: int, kill_timeout: int = 0
```
squirrelsc: Give the signal a default value.

squirrelsc:
I'm thinking about how the tool can be integrated with existing tools deeper. But so far it can be merged to use firstly. Current tools have the timeout underlying, but the timeout is controlled by force kill. We may consider supporting signal for graceful kill.

mcgov:
I have a use for it coming in the next PR :)

squirrelsc:
> 

ok, add a default value for signal, which can gracefully kill.

mcgov:
Refactored the SIG constants to util.constants now that they're shared between Kill and Timeout. Set default signal to SIGTERM

lisa/tools/timeout.py
```diff
@@ -11,7 +12,7 @@ def can_install(self) -> bool:
         return False
 
     def run_with_timeout(
-        self, command: str, signal: str, timeout: int, kill_timeout: int = 0
+        self, command: str, timeout: int, signal: int = SIGTERM, kill_timeout: int = 0
```
squirrelsc, line 15: How about put SIGINT as default? Is it softer?

mcgov: I think TERM is the 'nice' way to ask for a kill normally. Testpmd likes INT in interactive mode but it's less common (I think).

squirrelsc (approved):

PR #2105, opened 2022-06-30
DPDK: Force stable source for 18.04

mcgov:
Ubuntu 18.04 is used for golden image tests and the packaged
version of DPDK is 17.11. That version lacks the net_vdev_netvsc
pmd used for intuitive handling of hyper-v guests.

Rather than skipping, force a source build on that platform using 19.11
as the default.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -19,7 +19,7 @@ from lisa import (
 )
 from lisa.features import NetworkInterface
 from lisa.nic import NicInfo
-from lisa.operating_system import OperatingSystem
+from lisa.operating_system import OperatingSystem, Ubuntu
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -131,6 +131,21 @@ def _enable_hugepages(node: Node) -> None:
     )
 
 
+def _force_dpdk_stable_lkg_for_ubuntu_18_04(
```
squirrelsc: The method may be extended to other distros. So change the name like `_set_source_by_distro`

squirrelsc (approved):

mcgov:
why are the checks broken

squirrelsc:
> why are the checks broken

triggered by close and open PR.

PR #2106, opened 2022-06-30
DPDK Perf: Unsupported GPU, HPC

kamalca:
GPU and HPC distros are unsupported by DPDK perf tests. These SKUs are not a use case for DPDK.

Building on #2102


squirrelsc (approved):

mcgov:
Can we add these requirements to the perf tests as well, it can be a separate PR.

mcgov (approved):

PR #2107, opened 2022-07-02
handle the situation when images were deleted

LiliDeng:
current exception is 
```
    new_marketplace.version = versioned_images[-1].name
IndexError: list index out of range
```

sample image
```
suse sles-15-sp2 gen1 latest
```

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1766,6 +1766,11 @@ class AzurePlatform(Platform):
                     offer=marketplace.offer,
                     skus=marketplace.sku,
                 )
+            if 0 == len(versioned_images):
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1766,6 +1766,11 @@ def _parse_marketplace_image(
                     offer=marketplace.offer,
                     skus=marketplace.sku,
                 )
+            if 0 == len(versioned_images):
+                raise LisaException(
+                    f"fail to list versions of image {marketplace.publisher} "
```
squirrelsc: Make the "fail" accurate, like "cannot find any version of ..."

squirrelsc (approved):

PR #2109, opened 2022-07-04
Notify results of kvm-unit-tests using CommunityTestMessage

anirudhrb:
Enhance parsing logic to also extract passed and skipped tests and notify the result of each test by sending a `CommunityTestMessage`.

### DIFF
--- a/microsoft/testsuites/kvm/kvm_unit_tests.py
+++ b/microsoft/testsuites/kvm/kvm_unit_tests.py
@@ -2,9 +2,15 @@
 # Licensed under the MIT license.
 from pathlib import Path
 
-from assertpy.assertpy import assert_that
-
-from lisa import Logger, Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa import (
+    Environment,
+    Logger,
+    Node,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+)
+from lisa.testsuite import TestResult
 from lisa.tools import Lscpu
 from lisa.util import SkippedException
 from microsoft.testsuites.kvm.kvm_unit_tests_tool import KvmUnitTests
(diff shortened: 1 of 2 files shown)


anirudhrb:
Cc: @vyadavmsft @pupacha

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
```diff
@@ -47,15 +72,61 @@ def run_tests(self, failure_logs_path: Path) -> List[str]:
...
+                result.status = TestStatus.PASSED
+            elif status == "FAIL":
+                if result.name in self.EXPECTED_FAILURES:
+                    result.status = TestStatus.ATTEMPTED
+                else:
+                    result.status = TestStatus.FAILED
+            elif status == "SKIP":
+                result.status = TestStatus.SKIPPED
+            else:
+                continue
```
dsrivastavv: Is there any other status expected? If not, can you please throw an error?

microsoft/testsuites/kvm/kvm_unit_tests.py
```diff
@@ -25,28 +31,17 @@ class KvmUnitTestSuite(TestSuite):
...
-        failures = node.tools[KvmUnitTests].run_tests(log_path)
-        if failures:
-            log.info(f"Failed tests: {failures}")
-
-        unexpected_failures = list(
-            filter(lambda x: x not in expected_failures, failures)
-        )
-
-        assert_that(unexpected_failures).is_empty()
+        node.tools[KvmUnitTests].run_tests(result, environment, log_path)
```
dsrivastavv, line 47: Do we need to fail `verify_kvm_unit_tests` if `run_tests` return a non empty list?

anirudhrb: `run_tests` doesn't return anything anymore. It takes of the assertion too.

microsoft/testsuites/kvm/kvm_unit_tests_tool.py
anirudhrb: No other status is expected. If the execution reaches this line, it means that something other than test status was printed by the script. In my testing I've never seen the script print anything else other than test status. But there's no guarantee that the script won't be modified in the future to print some additional info. So, we should not error out. We should simply ignore such lines and continue parsing the lines that do indicate a test status.

dsrivastavv: Isn't this issue handled in line 113-115 which ignores any lines not following pattern: "name status"? We need to ensure that any changes in the script can be detected and adding it to logs might not be enough since they can easily get ignored. 

However, sending an error here could lead to later test result community messages not being sent. How about we throw an error right after all the community messages are sent?

Also, does each of these result line follow a specific regex pattern? If so, it would be more robust to parse from regex. You can look at the example here: https://github.com/microsoft/lisa/blob/43398b083687a5b1743e2814b03f5d8457157d37/microsoft/testsuites/ltp/ltp.py#L329-L345

anirudhrb: Sure, I can use a regex pattern but that still leaves the question of what to do with lines that don't match the pattern. 

> We need to ensure that any changes in the script can be detected

No, we don't need to detect all changes in the script. Some changes are harmless for us. If the script starts printing some added info, we can simply ignore that. For example, let's say tomorrow the script prints a summary line at the end. We don't want LISA to throw an error for that because the summary line doesn't match the regex. We can simply ignore that line because we don't need it.

The type of change we _do_ want to detect is if the format of the test status line ever changes. This detection is already done. If the test status line format changes, none of the lines will match the regex (or in the current code all lines will fall to the `else` case) and the results will be empty. There is code above that checks if the `results` list is empty and throws an exception. This exception will alert us.

anirudhrb: Added regex based parsing.

dsrivastavv (approved):

squirrelsc (approved):

PR #2113, opened 2022-07-06
Add python as dependency for lsvmbus

dsrivastavv:
Running lsvmbus fails if `python` is not installed or `python` command doesn't exist

### DIFF
--- a/lisa/tools/lsvmbus.py
+++ b/lisa/tools/lsvmbus.py
@@ -1,9 +1,10 @@
 import re
-from typing import Any, List
+from typing import Any, List, Type
 
 from lisa.base_tools.wget import Wget
 from lisa.executable import Tool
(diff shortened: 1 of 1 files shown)


LiliDeng:
thanks @somil55 , seems we work on the same issue https://github.com/microsoft/lisa/pull/2111/commits/726f5f074b705248ea1db8146536d707d842d300

dsrivastavv:
@LiliDeng I think we can abandon this PR. Can you please move the dependency part to your VM?

LiliDeng:
> @LiliDeng I think we can abandon this PR. Can you please move the dependency part to your VM?

thank you, I will test it when take python as dependency, actually, when I debug it, it didn't enter into dependency logic at all.

LiliDeng:
thanks, fix contained in PR https://github.com/microsoft/lisa/pull/2111 and merged, so close this one.

PR #2116, opened 2022-07-06
Force run wget for docker installation

dsrivastavv:
LISA was using cached results

### DIFF
--- a/lisa/tools/docker.py
+++ b/lisa/tools/docker.py
@@ -83,6 +83,7 @@ class Docker(Tool):
                 filename="get-docker.sh",
                 file_path="./",
                 executable=True,
+                force_run=True,
             )
             self.node.execute("./get-docker.sh", sudo=True)
         # RHEL 8 and its derivatives don't support docker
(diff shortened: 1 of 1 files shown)


lisa/tools/docker.py
```diff
@@ -83,6 +83,7 @@ def _install(self) -> bool:
                 filename="get-docker.sh",
                 file_path="./",
                 executable=True,
+                force_run=True,
```
squirrelsc, line 86: Why doesn't the cached result work here?

dsrivastavv: I am not entirely sure. Maybe the file was deleted but the command cache was still present. This error is not reproducible and there were around 3 instances of this issue. I feel it could be happening because multiple tests ran on same VM

LiliDeng: @somil55 @squirrelsc I have met this issue today, and fixed in #2136

dsrivastavv: @LiliDeng Were you able to get a repro of this?

LiliDeng: @somil55 I repro the issue when run verify_dpdk_sriov_rescind_failover_send_only and verify_dpdk_ovs in sequence. the second case run will delete the downloaded file since overwrite is set True as default, and at that time force run is set as force, wget will use cache result, so I think the RCA of your issue is the same.

dsrivastavv: Thanks for confirming Lili! I will close out this PR

dsrivastavv:
Fixed in #2136

PR #2120, opened 2022-07-08
Fixe some recent errors for fio testcases

dsrivastavv:


lisa/tools/fio.py
```diff
@@ -61,10 +61,11 @@ def can_install(self) -> bool:
 
     def install(self) -> bool:
         posix_os: Posix = cast(Posix, self.node.os)
-        if posix_os.package_exists("fio"):
-            posix_os.install_packages("fio")
-        else:
+        try:
             self._install_from_src()
+        except Exception as e:
```
squirrelsc, line 66: What's the reason to install from source code firstly? The latest code may not work with old distros.

dsrivastavv: The earlier logic was incorrect.  It is checking if fio is installed with `posix_os.package_exists`. But I found that this instead runs command `['sudo', 'yum', 'list', 'installed', 'fio']` which is just checking if fio is available in the system. If not, it moves to compiling from source which fails on some distro because C11 atomics support is not found. So, in these cases we need to do `yum install fio`. I reorganised the logic so that we rely on compiling first to have more consistency in fio version, but fall back to other version with `yum install` if that fails

squirrelsc (approved):

dsrivastavv:
@squirrelsc Atomicwrites 1.4.0 doesn't seem to be available on pypi. The only version available is 1.4.1: https://pypi.org/project/atomicwrites/#history. I think we will need to update poetry lock file. The pipeline on main is failing for windows as well. My local poetry installation has some issues. Can you please update poetry lock file?

squirrelsc:
> @squirrelsc Atomicwrites 1.4.0 doesn't seem to be available on pypi. The only version available is 1.4.1: https://pypi.org/project/atomicwrites/#history. I think we will need to update poetry lock file. The pipeline on main is failing for windows as well. My local poetry installation has some issues. Can you please update poetry lock file?

Fixed, and merged.

PR #2125, opened 2022-07-12
mark env(s) as bad if node(s) cannot be accessible.

LiliDeng:


lisa/runners/lisa_runner.py
```diff
@@ -352,6 +352,16 @@ def _run_test_task(
         # on next test cases, close the connection after each test run. It will
         # be connected on the next command automatically.
         environment.nodes.close()
+        # Try to connect node(s), if cannot access node(s) of this environment,
+        # set the current environment as Bad. So that this environment won't be reused.
+        if not environment.nodes.is_connectable():
```
squirrelsc: call `initialize` should be enough, it doesn't need to wrap a magic command.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -590,6 +590,19 @@ def check_or_create_resource_group(
                     resource_group_name, {"location": location}
                 )
 
+            timeout = 30
+            timeout_timer = create_timer()
+            while timeout_timer.elapsed(False) < timeout:
+                az_shared_rg_exists = rm_client.resource_groups.check_existence(
```
squirrelsc: Is it fixed by wait longer? The create_or_update is a synchronized API as before.

microsoft/testsuites/network/common.py
```diff
@@ -205,7 +205,7 @@ def cleanup_iperf3(environment: Environment) -> None:
 
 def remove_extra_nics(environment: Environment) -> None:
     for node in environment.nodes.list():
-        node = cast(RemoteNode, environment.nodes[0])
+        node = cast(RemoteNode, node)
```
squirrelsc: It's better to move the easy fixes in a separated PR.

LiliDeng:
force rebase by accident.

PR #2126, opened 2022-07-12
make env as bad when nodes not accessible

LiliDeng:


lisa/runners/lisa_runner.py
```diff
@@ -352,19 +352,21 @@ def _run_test_task(
...
-        except Exception as identifier:
-            self._log.debug(f"exception {identifier}")
-            environment.status = EnvironmentStatus.Bad
-            self._log.debug(
-                f"set environment '{environment.name}' as bad, "
-                f"because after test case '{test_result.name}', "
-                f"node(s) cannot be accessible."
-            )
-        environment.nodes.close()
+        if test_result.is_completed:
```
squirrelsc: Is this check necessary? If the logic goes here, the test result should be completed. This line is harmless to logic, but it's dead code, confusing people.

LiliDeng:
@squirrelsc 
I do not quite understand logic in 
https://github.com/microsoft/lisa/blob/main/lisa/runners/lisa_runner.py#L368 why when set keep failed and case failed, environment status is EnvironmentStatus.Deleted?
https://github.com/microsoft/lisa/blob/main/lisa/runners/lisa_runner.py#L390 whey env is deployed/connected or it is prepared status and env is in use, execute delete_environment

during my testing, I find when env is marked as BAD, it is not deleted actually.

lisa/node.py
```diff
@@ -255,6 +255,7 @@ def close(self) -> None:
             self._shell.close()
         if self._nics:
             self._nics = None
+        self._is_initialized = False
```
squirrelsc: sorry for confusing, I double checked the logic, the `node.initialize` shouldn't be called second time. So it needs to add an extra method in `Nodes` to connect all nodes in parallel (like `Nodes.initialize`). The method can be named like `test_connections`.

squirrelsc:
> @squirrelsc I do not quite understand logic in https://github.com/microsoft/lisa/blob/main/lisa/runners/lisa_runner.py#L368 why when set keep failed and case failed, environment status is EnvironmentStatus.Deleted? 
This is to keep environment, if the runbook specified "keep_environment" as failed. So it keeps environment when the test result is failed.


> https://github.com/microsoft/lisa/blob/main/lisa/runners/lisa_runner.py#L390 whey env is deployed/connected or it is prepared status and env is in use, execute delete_environment
Your change on this line is right, the bad environment should go to first `if` clause. The current logic is not clear. You can change the condition like:
```python
if environment.status == EnvironmentStatus.Prepared and not environment.is_in_use:
    # The prepared only environment doesn't need to be deleted. It may cause platform fail to delete non-existing environment.
    environment.status = EnvironmentStatus.Deleted
else:
    # previous `if` clause logic.
```
> 
> during my testing, I find when env is marked as BAD, it is not deleted actually.

lisa/runners/lisa_runner.py
```diff
@@ -352,11 +352,20 @@ def _run_test_task(
...
+        # Try to connect node(s), if cannot access node(s) of this environment,
+        # set the current environment as Bad. So that this environment won't be reused.
+        if not environment.nodes.test_connections():
+            environment.status = EnvironmentStatus.Bad
+            self._log.debug(
+                f"set environment '{environment.name}' as bad, "
+                f"because after test case '{test_result.name}', "
+                f"node(s) cannot be accessible."
+            )
+            environment.nodes.close()
```
squirrelsc: This line shouldn't be in the if clause. The connections should be closed, if the test returns True.

lisa/runners/lisa_runner.py
```diff
@@ -392,6 +401,7 @@ def _delete_environment_task(
             in [
```
squirrelsc: As the previous comments, please make this logic clearer.

selftests/runners/test_lisa_runner.py
```diff
@@ -238,8 +238,8 @@ def test_fit_a_predefined_env(self) -> None:
 
         self.verify_test_results(
             expected_test_order=["mock_ut1", "mock_ut2", "mock_ut3"],
-            expected_envs=["", "customized_0", "customized_0"],
-            expected_status=[TestStatus.SKIPPED, TestStatus.PASSED, TestStatus.PASSED],
+            expected_envs=["", "customized_0", ""],
+            expected_status=[TestStatus.SKIPPED, TestStatus.PASSED, TestStatus.SKIPPED],
```
squirrelsc: The code logic should be updated to support the UTs. The UT is used to verify the core logic. Every change should be thought carefully.

lisa/runners/lisa_runner.py
LiliDeng: I tried this, but when set failed to keep_environment and case failed, the env still be deleted
```
        if environment.status == EnvironmentStatus.Prepared and not environment.is_in_use:
            # The prepared only environment doesn't need to be deleted. It may cause platform fail to delete non-existing environment.
            environment.status = EnvironmentStatus.Deleted
        else:
            try:
                self.platform.delete_environment(environment)
                self._log.debug(
                    f"error on deleting environment '{environment.name}': {identifier}"
                )
```

squirrelsc: ok, it looks if the state is deleted already, it should go with first logic too. Let me know, if there is other cases.

lisa/runners/lisa_runner.py

selftests/runners/test_lisa_runner.py
LiliDeng: hmm, I am looking into it, now the env will be marked as Bad, since it is not real env which can be ssh.

lisa/runners/lisa_runner.py
LiliDeng: I add EnvironmentStatus.Deleted situation into if, so far, test results are expected.

lisa/runners/lisa_runner.py
```diff
@@ -388,22 +397,19 @@ def _delete_environment_task(
...
         if (
-            environment.status
-            in [
-                EnvironmentStatus.Deployed,
-                EnvironmentStatus.Connected,
-            ]
-        ) or (
-            environment.status == EnvironmentStatus.Prepared and environment.is_in_use
+            environment.status == EnvironmentStatus.Deleted
+            or environment.status == EnvironmentStatus.Prepared
```
squirrelsc: The deleted should be included without other conditions.
```python
  environment.status == EnvironmentStatus.Deleted
            or (environment.status == EnvironmentStatus.Prepared
            and not environment.is_in_use)
```

selftests/runners/test_lisa_runner.py
squirrelsc: use the `is_unittest` to skip connection test. There are some examples in code.

squirrelsc (approved):

PR #2128, opened 2022-07-12
Doc: add test case diagram

squirrelsc:


LiliDeng:
do we need add before_case and after_case introduction?

squirrelsc:
> do we need add before_case and after_case introduction?

I will add some

docs/write_test/write_case.rst
```diff
@@ -52,23 +52,25 @@ can save everyone’s time.
...
 
-The LISA test is composed of :ref:`write_test/write_case:metadata`,
-:ref:`write_test/write_case:test case body` and
-:ref:`write_test/write_case:setup and clean-up`.
+A typical test case includes definition, setup, run, and validation.
 
-Metadata
-~~~~~~~~
+   .. figure:: ../img/sample.png
+      :alt: sample test case
```
dsrivastavv, line 58: Should we add `cleanup` step in the diagram as well?

squirrelsc: I will add some descriptions in after_case. It needs to keep the sample simple, so it's easy to follow.

dsrivastavv (approved):

LiliDeng (approved):

PR #2132, opened 2022-07-14
support xdp on mariner

LiliDeng:


lisa/tools/tcpdump.py
```diff
@@ -71,6 +71,7 @@ def parse(self, packet_filename: str = "tcp_dump.pcap") -> List[IpPacket]:
             expected_exit_code_failure_message=f"error on parse packet file: "
             f"{packet_filename}",
             force_run=True,
+            sudo=True,
```
squirrelsc, line 74: Why the `parse` method need sudo?

LiliDeng: in mariner, I saw no permission issue for tcp_dump.pcap

squirrelsc (approved):

PR #2137, opened 2022-07-17
use one single place for xfstests tool to avoid space is full issue

LiliDeng:


lisa/tools/xfstests.py
```diff
@@ -114,7 +114,7 @@ class Xfstests(Tool):
 
     @property
     def command(self) -> str:
-        return "xfstests"
+        return str(self._code_path)
```
squirrelsc: It looks is the folder "/xfstest-dev" does it work as the command here?

LiliDeng: command is not used in xfstests at all, change into 'f"ls -lt {self._code_path}"'?

squirrelsc: ok, so it looks like used for exists check only. If so, the code is ok. But you can add some comments to explain the weird logic.

LiliDeng: updated

lisa/tools/xfstests.py
```diff
@@ -114,7 +114,8 @@ class Xfstests(Tool):
 
     @property
     def command(self) -> str:
-        return "xfstests"
+        # command is not used in xfstests
```
squirrelsc: "command is not used in xfstests" it's not true. The command is not directly to be called. But it's used to check if the command installed or not.

LiliDeng: in this tool, _check_exists is overwrite, I don't think command is used.

squirrelsc: If it's not used. set it a value like "mockup" and add the comments. It isn't confused.

LiliDeng: update

squirrelsc (approved):

PR #2138, opened 2022-07-18
fix ErrorCode:AuthorizationPermissionMismatch issue

LiliDeng:
current code throws below exception when invoke get_blob_properties, probably because they are v1 type, not sure the RCA, but updated code works.
```"HttpResponseError: Operation returned an invalid status 'This request is not authorized to perform this operation using this permission.' ErrorCode:AuthorizationPermissionMismatch"```

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -86,6 +86,7 @@ from .common import (
     get_or_create_storage_container,
     get_resource_management_client,
     get_storage_account_name,
+    get_storage_client,
     global_credential_access_lock,
     wait_copy_blob,
     wait_operation,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lisa/sut_orchestrator/azure/platform_.py [hunk 1]
```diff
@@ -2136,27 +2147,39 @@ def _enable_ssh_on_windows(self, node: Node) -> None:
         log.dump_json(logging.DEBUG, result)
 
     def _get_vhd_os_disk_size(self, blob_url: str) -> int:
-        vhd_blob = BlobClient.from_blob_url(
-            blob_url=blob_url,
+        matched = STORAGE_CONTAINER_BLOB_PATTERN.match(blob_url)
```
squirrelsc, line 2150: The from blob url should wrap this logic already. When does not it work?

LiliDeng: any feedback?

squirrelsc: Can you paste the error again? I want to double check if from_blob_url work with this scenario or not.

LiliDeng: deployment failed. HttpResponseError: Operation returned an invalid status 'This request is not authorized to perform this operation using this permission.'
ErrorCode:AuthorizationPermissionMismatch

squirrelsc (approved):

PR #2141, opened 2022-07-18
DPDK: fix check for source built binary

mcgov:
Previous check looked for any installed binary, since a package manager could have been used for a previous test we need to check for the source build specifically.
The currently running test will know to use the built binary, but we need to mark the environment as dirty for the next test in this case.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -469,9 +469,10 @@ class DpdkTestpmd(Tool):
         self.dpdk_path = self.node.working_path.joinpath(self._dpdk_repo_path_name)
 
         if self.find_testpmd_binary(
-            assert_on_fail=False
+            assert_on_fail=False, check_path="/usr/local/bin"
         ):  # tools are already installed
             return True
+
         git_tool = node.tools[Git]
         echo_tool = node.tools[Echo]
 


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -469,9 +469,14 @@ def _install(self) -> bool:
...
 
         if self.find_testpmd_binary(
-            assert_on_fail=False
+            assert_on_fail=False, check_path="/usr/local/bin"
         ):  # tools are already installed
             return True
+        else:
+            # we are building dpdk on a node where it was installed
+            # as a package, this will make it unclear which version to use
+            # for the next test, so mark the environment dirty.
```
squirrelsc: If it's installed from source code, it can be installed into a global tool path. It can be got in next test case too. So the tool can be shared across test cases. Even not, it doesn't need to make an env is dirty. The tools can be built and installed again in another folder.

mcgov: It should be fine, I'll remove. Trying to avoid weird failures because of bad state.

squirrelsc: We should understand the bad state, and avoid overusing the "dirty" feature. The dirty feature will slow down the test runs, because it needs to deploy more environments.

mcgov:
@LiliDeng Can you review and merge if this looks good? It will complement the other PR that's open from you.

squirrelsc (approved):

PR #2142, opened 2022-07-19
fix no attribute _dpdk_version_info issue

LiliDeng:


mcgov (dismissed):
This is good, I think our hitting this is due to environment reuse. This combined with the other PR to check for source installation specifically (and ignore the package manager version) when using a source install should fix that.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -145,6 +145,19 @@ def set_dpdk_branch(self, dpdk_branch: str) -> None:
         self._dpdk_branch = dpdk_branch
 
     def get_dpdk_version(self) -> VersionInfo:
+        if not hasattr(self, "_dpdk_version_info"):
```
squirrelsc: I'm not sure when the attribute is not here. But it looks the `_install` can set it. The install progress is pretty complex, and not sure if the following install steps is the same as current logic. Anyway, if the fix works well, please resolve my comments and merge it.

LiliDeng: this happened when the following case reuse this tool, at that time, _install is not invoked, so _dpdk_version_info is not initialized.
Yes, this is complex when reinitialized _dpdk_version_info, any suggestions to simplify it? @mcgov

squirrelsc: I see, the reason is the DpdkTestpmd is not initialized with a standard way, so it's behavior is weird. (testpmd = DpdkTestpmd(node)). Now, LISA supports to initialize with parameters, please try to update the logic to use `node..tools.get()` to replace it.

LiliDeng: sorry, I didn't understand, I need implement initialize with parameters, then node..tools.get() is for ?

squirrelsc: Please search the code for "testpmd = DpdkTestpmd(node)". It's not a right way to initialize a tool, so the tool object is not hold by node. The historical reason is this tool need pass parameters when initializing. Now, the node.tools.get() can pass in parameters, and can be used to create a tool like normal way. Once it's changed like this way, the variable will be available always.

LiliDeng: when pass like this way `node.tools.get(DpdkTestpmd, dpdk_source=dpdk_source, dpdk_branch=dpdk_branch)`, I can see the values in https://github.com/microsoft/lisa/blob/main/lisa/executable.py#L583 during debug. after that, they got lost, I can't add extra parameters for _initialize of tool DpdkTestpmd

squirrelsc: It needs to receive those values from the __init__ method of the tool. Check how it's used by subclass factory.

LiliDeng: update

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -392,6 +390,19 @@ def add_sample_apps_to_build_list(self, apps: Union[List[str], None]) -> None:
         else:
             self._sample_apps_to_build = []
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._dpdk_source = kwargs.pop("dpdk_source")
```
squirrelsc: Add a default value, so in some case, if the field is missing, the default value will be used.

squirrelsc (dismissed):

microsoft/testsuites/dpdk/dpdktestpmd.py
LiliDeng: update

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -392,6 +390,23 @@ def add_sample_apps_to_build_list(self, apps: Union[List[str], None]) -> None:
         else:
             self._sample_apps_to_build = []
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        dpdk_source = kwargs.pop("dpdk_source")
```
squirrelsc: I mean change it like,

```python
dpdk_source = kwargs.pop("dpdk_source", PACKAGE_MANAGER_SOURCE)
```

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -392,6 +390,15 @@ def add_sample_apps_to_build_list(self, apps: Union[List[str], None]) -> None:
         else:
             self._sample_apps_to_build = []
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self._dpdk_source = kwargs.pop("dpdk_source", PACKAGE_MANAGER_SOURCE)
+        self._dpdk_branch = kwargs.pop("dpdk_branch")
```
squirrelsc: write it like `kwargs.pop("dpdk_branch", "")`, give it a default value.

LiliDeng: update

squirrelsc: is it pushed?

LiliDeng: by default, dpdk_source is given PACKAGE_MANAGER_SOURCE, dpdk_branch is not used if install dpdk from package manager

squirrelsc: It's about argument design. Because it's not used, the dpdk_branch is able to be not passed when calling it. But if there is no default value given, there will be an error. So the default value is to make the argument is optional.

LiliDeng: update :)

squirrelsc (approved):

PR #2143, opened 2022-07-19
DPDK ring_ping: change metric

kamalca:
For increased test reliability, we decided to reduce the bar for ring ping performance. Further discussion may need to occur surrounding the performance expectations of this test.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -57,7 +57,7 @@ class Dpdk(TestSuite):
     # regex for parsing ring ping output for the final line,
     # grabbing the max latency of 99.999% of data in nanoseconds.
     # ex: percentile 99.999 = 12302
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -57,7 +57,7 @@ class Dpdk(TestSuite):
     # regex for parsing ring ping output for the final line,
     # grabbing the max latency of 99.999% of data in nanoseconds.
```
mcgov, line 58: This comment needs adjusting to explain the change.

mcgov (changes requested):
Can we refactor the guts of the verify_dpdk_ring_ping logic into dpdkutil as well, leaving the performance check as the last thing to check in the test suite. This will prep for adding the stricter perf version.

LiliDeng:
any update? @kamalca

LiliDeng (approved):

mcgov (approved):

PR #2149, opened 2022-07-21
Add Python tests for newer versions

tonybaloney:
Copy of #2100


squirrelsc:
See an error in check tab 

The workflow is not valid. .github/workflows/continuous-integration-workflow.yml (Line: 56, Col: 13): Unexpected symbol: '"3'. Located at position 26 within expression: matrix.python-version == "3.8" .github/workflows/continuous-integration-workflow.yml (Line: 60, Col: 13): Unexpected symbol: '"3'. Located at position 26 within expression: matrix.python-version == "3.8"

tonybaloney:
This PR shows some backward incompatible changes with the way data classes work. It will now raise a `ValueError` if you set the default value of a field to a mutable type.

This has always been problematic in Python, leading to unexpected behaviours, so the new error is just enforcing something which was always there.

I'll include some changes to the source to use zero-arg lambda's to comply with this without breaking compatibility to older versions of LISA

squirrelsc:
@tonybaloney Thank you for the changes, they are really adding value to LISA. Can you split it to two PRs? One is about more test matrix, another is to fix the mutable types. So we can get the second change merged quicker. And we encourage to create meaningful commit, and rebase to main. So if it's possible, Please try to use "git rebase -i" and "git add/reset -p" to edit each commit, and merge the history in the PR. Thank you again!

tonybaloney:
@squirrelsc this is a good idea. 3.11 seems to require more changes, so I will split into 2. One testing up to 3.10 and the other expanding beyond with mutable fields in a factory

LiliDeng:
@tonybaloney hi, Anthony, any update :)

LiliDeng:
@tonybaloney it has been long time since your last update, close it for now, feel free to reopen it when you have time to update it, thank you.

PR #2151, opened 2022-07-21
Libvirt: Implement _get_environment_information

pupacha:
Add following fields as information.
* Host Distro
* Host Kernel Version
* Libvirt Version
* VMM (qemu/ch) Version

### DIFF
--- a/lisa/sut_orchestrator/libvirt/ch_platform.py
+++ b/lisa/sut_orchestrator/libvirt/ch_platform.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 
 import os
+import re
 import xml.etree.ElementTree as ET  # noqa: N817
 from pathlib import Path
 from typing import List, Type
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -189,3 +189,13 @@ def _create_node_os_disk(
                 expected_exit_code=0,
                 expected_exit_code_failure_message="Failed to copy os disk image",
             )
+
+    def _get_vmm_version(self) -> str:
+        result = ""
+        if self.host_node:
+            result = self.host_node.execute(
```
squirrelsc: In LISA, if the output is not huge (over 200 lines), it's better to filter by regexp instead of grep. With this way, the LISA log contains the raw output. When the output is changed, the regexp is easier to troubleshoot than setup a new environment to get the raw output again.

pupacha: Modified to use regex.

squirrelsc (approved):

PR #2153, opened 2022-07-22
Fix kernel_installer

pupacha:
check for source only if the runbook has source field.

### DIFF
--- a/lisa/transformers/kernel_installer.py
+++ b/lisa/transformers/kernel_installer.py
@@ -135,7 +135,10 @@ class KernelInstallerTransformer(Transformer):
             posix = cast(Posix, node.os)
             posix.replace_boot_kernel(installed_kernel_version)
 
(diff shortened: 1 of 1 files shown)


LiliDeng (approved):

lisa/transformers/kernel_installer.py
```diff
@@ -135,7 +135,10 @@ def _internal_run(self) -> Dict[str, Any]:
             posix = cast(Posix, node.os)
             posix.replace_boot_kernel(installed_kernel_version)
 
-        if installer.runbook.source == "linux-image-azure-fde":
+        if (
```
squirrelsc, line 138: It can be merged with above `if`. So the logic is clearer to handle different behavior of fde/non-fde.

pupacha: Sure, Raised PR: https://github.com/microsoft/lisa/pull/2154

PR #2154, opened 2022-07-22
kernel_installer: Refactor logic of fde/non-fde check

pupacha:


LiliDeng (approved):

lisa/transformers/kernel_installer.py
```diff
@@ -128,39 +128,34 @@ def _internal_run(self) -> Dict[str, Any]:
...
-                ),
-                sudo=True,
-            )
-            node.execute(
-                f"cp {efi_file} /boot/efi/EFI/ubuntu/grubx64.efi",
-                sudo=True,
-                shell=True,
-            )
+        if hasattr(installer.runbook, "source"):
+            if installer.runbook.source != "linux-image-azure-fde":
```
squirrelsc, line 132: It doesn't need to be changed, but it's better to use positive conditions by default, if there is special reason. Use `if installer.runbook.source == "linux-image-azure-fde":` first, it will be a little bit better to understand.

PR #2156, opened 2022-07-23
Fix vm hot resize if new size allocation failed

dsrivastavv:
Retry if allocation fails due to non availability

### DIFF
--- a/microsoft/testsuites/core/vm_hot_resize.py
+++ b/microsoft/testsuites/core/vm_hot_resize.py
@@ -50,9 +50,12 @@ class VmHotResize(TestSuite):
                 expected_vm_capability = resize.resize()
                 break
             except Exception as identifier:
-                if "cannot find current vm size in eligible list" in str(identifier):
-                    retry = retry + 1
-                elif "OperationNotAllowed" in str(identifier):
+                if (
+                    "cannot find current vm size in eligible list" in str(identifier)
+                    or "OperationNotAllowed" in str(identifier)
+                    or "Allocation failed" in str(identifier)
+                    or "AllocationFailed" in str(identifier)
+                ):
                     retry = retry + 1
                 else:
                     raise identifier


dsrivastavv:
Let's keep this until @mcgov can verify this on his runs

microsoft/testsuites/core/vm_hot_resize.py
```diff
@@ -54,6 +54,8 @@ def verify_vm_hot_resize(self, log: Logger, node: Node) -> None:
                     retry = retry + 1
                 elif "OperationNotAllowed" in str(identifier):
                     retry = retry + 1
+                elif "Allocation failed" in str(identifier):
```
mcgov: I think this is fine to merge (it's a trivial 2 line change) but I'd like to see a comment added that explains why 10 retries is enough. I don't know the internals of NodeSpace so I'm not sure how it organizes by VM family. The issue with TiP is we're going to be on a node that only supports a specific family of sizes. Will 10 retries be enough to find one that the TiP node will support?

mcgov (changes requested):

microsoft/testsuites/core/vm_hot_resize.py
squirrelsc: The three `in` operators can be merged into one `if`, like

```python
                if "cannot find current vm size in eligible list" in str(identifier) or "OperationNotAllowed" in str(identifier) or "Allocation failed" in str(identifier):
                    retry = retry + 1
                else
                    raise identifier
```

dsrivastavv: I think the intent of 10 retries here is to wait for enough time for the requested nodes to become available. This is based on few of the recent failures I checked. For Tip specifically, do you think the error was happening because the requested resize vm was not available in the family of cluster it supports? or we just needed to wait for it to become free?

mcgov: Ah, ok so three questions:

- If hot-resize guarantees the node will not be live-migrated than just resizing it should be TiP safe, so is it accurate that hot resizing will guarantee the VM is not moved off the host node?

- If the problem is just that there weren't enough cores couldn't we test making the VM smaller instead of larger?

- Does running this test mean we need to mark the environment as dirty if the size is not reset?

dsrivastavv: 1. I think this should be true if the host node supports the new vm size but this might not always be true. I will confirm it. 
2. The v2 test was specifically targeting to a higher core count because it could lead to more issues, particularly with storage and networking when new cores are added
3. We are not making any changes other than resize. So if resize fails, nothing happened. However, if the vm is unreachable after this operation it could mean that it is stuck in some intermediate state and we should mark it as dirty. Further, I don't think VM metadata is updated in LISA after resize happens and it could schedule tests which were supposed to run on older vm size. This case should also be handled by marking node as dirty.

mcgov: I'm fine with merging as-is, if there are further issues with node state not being reset after the test we'll open bugs for them 🙂

mcgov (dismissed):

microsoft/testsuites/core/vm_hot_resize.py
mcgov: (feel free to resolve)

LiliDeng:
there are different allocation failures from current db records, could you please double confirm, whether you missed some? thanks. @somil55

dsrivastavv:
> there are different allocation failures from current db records, could you please double confirm, whether you missed some? thanks. @somil55

Added based on your suggestion

LiliDeng (approved):

PR #2158, opened 2022-07-26
increase wait time for sriov enable check

mcgov:
SRIOV enable/disable can take longer than a couple of minutes in some cases. Increase the polling interval to increase the potential minimum timeout (timing is not guaranteed with this retry decorator).

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -446,7 +446,7 @@ class NetworkInterface(AzureFeatureMixin, features.NetworkInterface):
         modprobe_tool = self._node.tools[Modprobe]
         modprobe_tool.reload(["hv_netvsc"])
 
-    @retry(tries=60, delay=2)
+    @retry(tries=60, delay=10)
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -446,7 +446,7 @@ def reload_module(self) -> None:
         modprobe_tool = self._node.tools[Modprobe]
         modprobe_tool.reload(["hv_netvsc"])
 
-    @retry(tries=60, delay=2)
+    @retry(tries=60, delay=10)
```
squirrelsc, line 449: 60 times * 10 seconds. It means 10 minutes. Is it too long?

mcgov: I would rather have a stable test with a slow worst-case than a uniformly fast test that is unstable.

mcgov: But no, I think 10 is a safe maximum.

squirrelsc (approved):

PR #2159, opened 2022-07-26
Reduce minimum core count requirement for storage tests

mcgov:
Storage tests currently require 72 cores for some premium storage performance tests. Reduce this to allow running them on all VM size families.

### DIFF
--- a/microsoft/testsuites/performance/storageperf.py
+++ b/microsoft/testsuites/performance/storageperf.py
@@ -51,7 +51,7 @@ class StoragePerformance(TestSuite):
         priority=3,
         timeout=TIME_OUT,
         requirement=simple_requirement(
-            min_core_count=32,
+            min_core_count=64,
             disk=schema.DiskOptionSettings(
                 disk_type=schema.DiskType.PremiumSSDLRS,
                 data_disk_iops=search_space.IntRange(min=5000),
@@ -69,7 +69,7 @@ class StoragePerformance(TestSuite):
         priority=3,
         timeout=TIME_OUT,
         requirement=simple_requirement(
-            min_core_count=32,
+            min_core_count=64,
             disk=schema.DiskOptionSettings(
                 disk_type=schema.DiskType.PremiumSSDLRS,
                 data_disk_iops=search_space.IntRange(min=5000),
@@ -89,7 +89,7 @@ class StoragePerformance(TestSuite):
         priority=3,
         timeout=TIME_OUT,
         requirement=simple_requirement(
-            min_core_count=32,
+            min_core_count=64,
             disk=schema.DiskOptionSettings(
                 disk_type=schema.DiskType.PremiumSSDLRS,
                 data_disk_iops=search_space.IntRange(min=5000),
--- a/microsoft/testsuites/performance/storageperf.py
+++ b/microsoft/testsuites/performance/storageperf.py
@@ -51,7 +51,7 @@ class StoragePerformance(TestSuite):
         priority=3,
         timeout=TIME_OUT,
         requirement=simple_requirement(
-            min_core_count=72,
+            min_core_count=32,
             disk=schema.DiskOptionSettings(
                 disk_type=schema.DiskType.PremiumSSDLRS,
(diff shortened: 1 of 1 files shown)


squirrelsc:
> Storage tests currently require 72 cores for some premium storage performance tests. Reduce this to allow running them on all VM size families.

As discussed before, if it needs to run perf tests on non-designed vm sizes, it needs to use force argument, instead of change the test case requirement. Please check with Srikanth on details.

mcgov:
> > Storage tests currently require 72 cores for some premium storage performance tests. Reduce this to allow running them on all VM size families.
> 
> As discussed before, if it needs to run perf tests on non-designed vm sizes, it needs to use force argument, instead of change the test case requirement. Please check with Srikanth on details.

For sure, I'm not trying to run with less than a full node. We have a ton of DS_v3 quota but the maximum size on that family (and DS_v2) is 64 cores. Would you be fine with switching to 64? I am already running them with the override_vmsize since it's in an availability set.

Clarifying: we don't have enough quota of the 72 core series to run all azure host builds at once. Requiring that SKU for all perf tests is blocking us having stable testing.

squirrelsc:
> > > Storage tests currently require 72 cores for some premium storage performance tests. Reduce this to allow running them on all VM size families.
> > 
> > 
> > As discussed before, if it needs to run perf tests on non-designed vm sizes, it needs to use force argument, instead of change the test case requirement. Please check with Srikanth on details.
> 
> For sure, I'm not trying to run with less than a full node. We have a ton of DS_v3 quota but the maximum size on that family (and DS_v2) is 64 cores. Would you be fine with switching to 64? I am already running them with the override_vmsize since it's in an availability set.
> 
> Clarifying: we don't have enough quota of the 72 core series to run all azure host builds at once. Requiring that SKU for all perf tests is blocking us having stable testing.

I hope we have a feature called like ExclusiveVM, it can return all exclusive VMs, which are not share resources with other VMs. And those perf test cases should be run on them by default, instead of keeping change the core requirement. Before it's implemented, please try to use `maximize_capability` to ignore the test requirements.

mcgov:
Swapped to 64 core minimum, I'm fine with using the exclusive requirement once it's ready but in the meantime I need to unblock these perf tests.

squirrelsc:
> Swapped to 64 core minimum, I'm fine with using the exclusive requirement once it's ready but in the meantime I need to unblock these perf tests.

In mean time, you can try to use `maximize_capability: true` at the azure node requirement. It can override this kind of test case requirement, instead of have a code change. Please check with Srikanth or me, how it's used.

mcgov:
What is bad about running on 64 core nodes? I don't get what the issue is.

squirrelsc:
> What is bad about running on 64 core nodes? I don't get what the issue is.

Those test cases are perf cases. When we test them with distros, we hope them run on exclusive VMs, instead of a shared resource, so the perf results won't be affected by other VMs. That's why we need the exclusive feature to identify which one is the exclusive VM. To get the core count smaller, it's easy to fall into a shared VM for general perf testing. Even the exclusive feature is added, the `maximize_capability` should be added for other vm sizes, which is not exclusive.

mcgov:
Why even have the requirement checking for perf tests if we just turn it off?

mcgov:
Am I misunderstanding maximize_capability, does it selectively disable one requirement but keeps the others? It seems strange to disable all the requirement checking for perf tests, I don't really want to deal with breaking test assumptions. If the test will run with lower values they aren't really 'requirements', it seems like the test should WARN but still run.

squirrelsc:
> Am I misunderstanding maximize_capability, does it selectively disable one requirement but keeps the others? It seems strange to disable all the requirement checking for perf tests, I don't really want to deal with breaking test assumptions. If the test will run with lower values they aren't really 'requirements', it seems like the test should WARN but still run.

The setting ignores all test case requirements. The perf test cases are used to measure perf on designated vm sizes. Test on other VM sizes is not stable due to side effects from collocated VMs. So it means to "override" the test requirements.

LiliDeng:
@mcgov is it still valid?

PR #2162, opened 2022-07-26
xfstests - send community test results

LiliDeng:


lisa/tools/xfstests.py
```diff
@@ -1,17 +1,30 @@
...
 from lisa.executable import Tool
+from lisa.messages import CommunityTestMessage, TestStatus, create_test_result_message
 from lisa.operating_system import CBLMariner, Debian, Posix, Redhat, Suse, Ubuntu
 from lisa.tools import Cat, Echo
 from lisa.util import LisaException, UnsupportedDistroException, find_patterns_in_lines
 
 from .git import Git
 from .make import Make
 
+if TYPE_CHECKING:
```
squirrelsc: Move this tool to testsuites/xfstests, because it causes some cross reference, and it's actually not be used by multiple test suites.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -548,6 +568,9 @@ def _execute_xfstests(
             cwd=xfstests.get_xfstests_path(),
             timeout=self.TIME_OUT,
         )
+        xfstests.create_send_community_test_msg(
```
squirrelsc: This method should be called in the xfstest tool, not in test case level.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -548,6 +568,9 @@ def _execute_xfstests(
             cwd=xfstests.get_xfstests_path(),
```
squirrelsc: I just found this is the actual xfstests command. It should be wrapped in the xfstests, not compose a command in test case level.

lisa/tools/xfstests.py
LiliDeng: moved

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: update

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: update

squirrelsc (approved):

PR #2164, opened 2022-07-26
Add feature to send command to serial console

dsrivastavv:


microsoft/testsuites/core/serial_console.py
```diff
@@ -0,0 +1,38 @@
+from lisa import (
+    Logger,
+    Node,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    simple_requirement,
+)
+from lisa.features import SerialConsole
+from lisa.sut_orchestrator.azure.features import SerialConsole as AzureSerialConsole
```
squirrelsc: The method should be defined in general serialconsole too, it's not an Azure only feature.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -108,10 +121,126 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
             wait_operation(operation, failure_identity="Start/Stop")
 
 
+class FixedSerialPortsOperations(SerialPortsOperations):  # type: ignore
```
squirrelsc, line 124: Add comments to explain why it's here, and it doesn't need the old code comments to get it simpler.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -145,6 +274,116 @@ def _get_console_log(self, saved_path: Optional[Path]) -> bytes:
...
+        cmd: str,
+        timeout: int = 200,
+    ) -> str:
+
+        # create asyncio loop
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+
+        retry = 0
+        for retry in range(3):
```
squirrelsc: Add comments to explain why it needs retry.

microsoft/testsuites/core/serial_console.py
```diff
@@ -0,0 +1,38 @@
...
+        priority=1,
+        requirement=simple_requirement(
+            supported_features=[SerialConsole],
+        ),
+    )
+    def verify_serial_console(self, log: Logger, node: Node) -> None:
+        command = "ls"
+        serial_console = node.features[SerialConsole]
+        assert isinstance(serial_console, AzureSerialConsole)
+        output = serial_console.run(command)
```
squirrelsc: Does it actually run command? I feel it's not easy, because it needs to login firstly.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -108,10 +121,126 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
...
+    RESOURCE_PROVIDER_NAMESPACE = "Microsoft.Compute"
+    PARENT_RESOURCE_TYPE = "virtualMachines"
+    DEFAULT_SERIAL_PORT_ID = 0
+
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._initialize_information(self._node)
+        self._initialize_serial_console(id=self.DEFAULT_SERIAL_PORT_ID)
+
+    def run(self, cmd: str) -> str:
```
squirrelsc: It's not easy to use like this way, because most scenarios it needs to login, and the console may output other system logs unpredictable. So let's create a connection and return in this iteration. So it returns a writer and reader. The writer is used to send content to serial console, and the reader is used to read back.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -145,6 +274,116 @@ def _get_console_log(self, saved_path: Optional[Path]) -> bytes:
...
+        )
+        token = platform.credential.get_token("").token
+        serial_port_connection_str = (
+            f"{connection.connection_string}?authorization={token}"
+        )
+
+        return serial_port_connection_str
+
+    @staticmethod
+    async def _run_command_async(connection_str: str, command: str) -> str:
```
squirrelsc: Move it out of the class, instead of marking it static method. That would be clearer.

microsoft/testsuites/core/serial_console.py
dsrivastavv: This essentially writes the command to the serial console and we just check if that it is present in the serial console output

lisa/sut_orchestrator/azure/features.py
dsrivastavv: This was my initial thought as well. However, the websocket connection is very unreliable and each time we make a new connection, output buffer resets, and user would need to read everything from beginning. The things that worked for me is to create a websocket connection, flush out the current output, send command to the websocket and read back the output.  If we just return a reader and writer, all the connection management would become an issue for users.

microsoft/testsuites/core/serial_console.py
squirrelsc: The input and output is not 1:1 mapping. We should assume the web socket stable at least when it's assigned. We can control underlying when to deliver a new connection or reuse existing one.

lisa/sut_orchestrator/azure/features.py
squirrelsc: The input and output is not 1:1 mapping. We should assume the web socket stable at least when it's assigned. We can control underlying when to deliver a new connection or reuse existing one.

microsoft/testsuites/core/serial_console.py
squirrelsc: It's just to write and echo back. If it uses "ls", it's confusing, people thought a command run. If it's really needed, use a mockup string like "echo back", so it can self explained.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -108,10 +121,211 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
...
+                    )
+                )
+                retry += 1
+                continue
+            except Exception as e:
+                self._log.debug(f"Failed to run command: {e}")
+                retry += 1
+                continue
+
+        raise LisaException("Failed to write command to serial console")
```
squirrelsc: save last exception, and raise it here.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -108,10 +121,211 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
...
+            self._ws = await websockets.connect(connection_str)  # type: ignore
+
+        return self._ws
+
+    async def _write_async(self, cmd: str) -> None:
+        # connect to websocket and send command
+        ws = await self._get_connection_async()
+        await ws.send(cmd)
+
+    async def _read_async(self) -> str:
```
squirrelsc: Because the async method is not to expose, so it doesn't need to use async here. The self._asyncio_loop should be in the method, and make those method synchronized.

microsoft/testsuites/core/serial_console.py
```diff
@@ -0,0 +1,38 @@
...
+        ),
+    )
+    def verify_serial_console(self, log: Logger, node: Node) -> None:
+        command = "echo back"
+        serial_console = node.features[SerialConsole]
+        _ = serial_console.read()
+        serial_console.write(command)
+        output = serial_console.read()
+
+        assert output == command
```
squirrelsc: The "assert" doesn't return expected and actual values, when it's failed. use assertpy and add description to help troubleshooting.

lisa/sut_orchestrator/azure/features.py
dsrivastavv: I didn't understand this comment. Do you mean remove all async functions and synchronously call websocket libraryt?

squirrelsc: See when the method it's called, there is a "_asyncio_loop" always. It means the method doesn't need to be "async", The async should be wrapped in the method itself. If the websocket support sync calls, it should be used. Because in LISA, we don't use the async/await.

dsrivastavv: Got it! I think saw some sync examples with websocket api. Let me check those.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -108,10 +121,209 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
...
+                self._log.debug(f"Failed to run command: {e}")
+                last_raised_excption = e
+
+            retry += 1
+
+        raise LisaException(
+            f"Failed to write command to serial console: {last_raised_excption}"
+        )
+
+    def read(self, timeout: int = 120) -> str:
```
squirrelsc: The timeout is not passed in, so it's not useful actually.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -108,10 +121,209 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
...
+        retry = 0
+        for retry in range(3):
```
squirrelsc: It looks the whole method is for retry, the "retry" decorator can be used directly.

squirrelsc (approved):

PR #2167, opened 2022-07-27
pass test result id into perf message

LiliDeng:


lisa/tools/iperf3.py
```diff
@@ -408,6 +412,7 @@ def create_iperf_udp_performance_message(
         other_fields["send_buffer_size"] = Decimal(buffer_length)
         other_fields["connections_num"] = connections_num
         other_fields["protocol_type"] = TransportProtocol.Udp
+        other_fields["test_result_id"] = result.id_
```
squirrelsc: This field is a common field for all perf results, so add it in the `create_perf_message`, instead of every usage. It's easy to miss.

lisa/tools/iperf3.py
```diff
@@ -340,6 +343,7 @@ def create_iperf_udp_performance_message(
         connections_num: int,
         environment: "Environment",
         test_case_name: str,
+        result: "TestResult",
```
squirrelsc: The environment should be attached to the result already. We can try to refine it better to remove environment from the message. So update the `create_perf_message` to accept test result, and retrieve environment, test case name from the test result.

lisa/tools/iperf3.py

lisa/tools/iperf3.py

microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -265,16 +277,20 @@ def perf_dpdk_netvsc_pmd_multi_queue(
     def perf_dpdk_netvsc_pmd_multi_queue_huge_vm(
         self,
         environment: Environment,
```
squirrelsc: The environment should be removed.

LiliDeng: update

squirrelsc (approved):

PR #2169, opened 2022-07-29
Add mask if variable or vhd contains sas token

lubaihua33:
If the value of vhd parameter contains sas token, the token will be printed in the log before this fix.

### DIFF
--- a/lisa/secret.py
+++ b/lisa/secret.py
@@ -16,8 +16,14 @@ PATTERN_FILENAME = (
     re.compile(r"^[^.]*?[\\/]?(.)[^\\/]*?(.[.]?[^.]*)$"),
     r"\1***\2",
 )
+# https://xx.core.windows.net/vhds/CentOS.vhd?sp=r&st=xx%2012:10:45&5;CA%aEMpls3D
+# replace as https://xx.core.windows.net/vhds/CentOS.vhd***
+PATTERN_URL = (
+    re.compile(r"(https?://([-\w]+\.)+[-\w]+(/[-./\w]*)?\??)([\w]+=[%&-:;=\w]*)?$"),
+    r"\1***",
+)
 
-patterns = {"guid": PATTERN_GUID, "headtail": PATTERN_HEADTAIL}
+patterns = {"guid": PATTERN_GUID, "headtail": PATTERN_HEADTAIL, "url": PATTERN_URL}
 
 
 def replace(
@@ -55,13 +61,19 @@ def add_secret(
     sub: str = "******",
 ) -> None:
     global _secret_list
-    if origin and origin not in _secret_set:
+    if origin:
         if not isinstance(origin, str):
             origin = str(origin)
-        _secret_set.add(origin)
-        _secret_list.append((origin, replace(origin, sub=sub, mask=mask)))
-        # deal with longer first, in case it's broken by shorter
-        _secret_list = sorted(_secret_list, reverse=True, key=lambda x: len(x[0]))
+        if origin in _secret_set:
+            for index, secret in enumerate(_secret_list):
+                if origin == secret[0]:
+                    _secret_list[index] = (origin, replace(origin, sub=sub, mask=mask))
+                    break
+        else:
+            _secret_set.add(origin)
+            _secret_list.append((origin, replace(origin, sub=sub, mask=mask)))
+            # deal with longer first, in case it's broken by shorter
+            _secret_list = sorted(_secret_list, reverse=True, key=lambda x: len(x[0]))
 
 
 def mask(input: str) -> str:
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -26,7 +26,7 @@ from lisa import schema
 from lisa.environment import Environment, load_environments
 from lisa.feature import Features
 from lisa.node import Node, RemoteNode
-from lisa.secret import PATTERN_HEADTAIL, add_secret
+from lisa.secret import PATTERN_HEADTAIL, PATTERN_URL, add_secret
 from lisa.util import LisaException, constants, field_metadata, strip_strs
 from lisa.util.logger import Logger
 from lisa.util.parallel import check_cancelled
@@ -162,6 +162,8 @@ class AzureNodeSchema:
                 "disk_type",
             ],
         )
+        # If vhd contains sas token, need add mask
+        add_secret(self.vhd, PATTERN_URL)
 
     @property
     def marketplace(self) -> Optional[AzureVmMarketplaceSchema]:


LiliDeng (changes requested):
@lubaihua33 I still see plain text in below part.
```parsed runbook: ```

**Comment by lubaihua33** [comment]
> @lubaihua33 I still see plain text in below part. `parsed runbook: `

@LiliDeng, I have updated the code, please check.

**Review by LiliDeng — DISMISSED** [review]

File: lisa/parameter_parser/runbook.py [hunk 1]
```diff
@@ -172,9 +172,12 @@ def _validate_and_load(data: Any) -> schema.Runbook:
 
         assert _schema
         runbook = cast(schema.Runbook, _schema.load(data))
+        runbook_dict = runbook.to_dict()  # type: ignore
+        # Check if runbook contains secret. If so, need add mask
+        secret.check_secret(runbook_dict)
```
squirrelsc: It doesn't need so complex, just add sas token to secret values.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -162,6 +162,8 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
                 "disk_type",
             ],
         )
+        # If vhd contains sas url, need add mask
+        add_secret_for_sasurl(self.vhd)
```
squirrelsc: why not just call secret.add_secret?

lisa/secret.py
```diff
@@ -16,6 +16,12 @@
     re.compile(r"^[^.]*?[\\/]?(.)[^\\/]*?(.[.]?[^.]*)$"),
     r"\1***\2",
 )
+# https://xx.core.windows.net/vhds/CentOS.vhd?sp=r&st=xx%2012:10:45&5;CA%aEMpls3D
+# replace as https://xx.core.windows.net/vhds/CentOS.vhd***
+PATTERN_SASURL = (
```
squirrelsc: The pattern doesn't need to be in the secret lib, it's Azure specified.

lisa/secret.py
```diff
@@ -69,3 +75,20 @@ def mask(input: str) -> str:
         if secret[0] in input:
             input = input.replace(secret[0], secret[1])
     return input
+
+
+def add_secret_for_sasurl(value: str) -> None:
```
squirrelsc: Move this method into azure/common.py

lisa/secret.py
```diff
@@ -69,3 +75,20 @@ def mask(input: str) -> str:
...
+    sasurl_match = PATTERN_SASURL[0].findall(value)
+    if sasurl_match:
+        add_secret(value, PATTERN_SASURL)
+
+
+def check_secret(obj: Any) -> None:
```
squirrelsc: This method is not necessary. The log will be filtered automatically.

lisa/secret.py
lubaihua33: Have moved this pattern to azure/common.py

lisa/secret.py
lubaihua33: I posted a new update. I changed the method name to check_sasurl. In this method, it checks if there is a value that has sasurl pattern, if so, then add secret for this value.

lisa/variable.py
```diff
@@ -10,6 +10,7 @@
 import yaml
 
 from lisa import schema, secret
+from lisa.sut_orchestrator.azure.common import PATTERN_SASURL
```
squirrelsc: This change doesn't need to change the variable.py. It's a lower level lib, shouldn't import azure libs. The secrets should be marked when it's first used in azure or other filles. If it's hard to know, mark it on first imported by special runbook variable, env variable, or command line variable. All of them support to mark a value to secret. The runbook supports format, and it's not must have.

lubaihua33: I posted a new update as we discussed. Please help to review again.

lisa/secret.py
```diff
@@ -17,7 +17,14 @@
     r"\1***\2",
 )
 
-patterns = {"guid": PATTERN_GUID, "headtail": PATTERN_HEADTAIL}
+# https://xx.core.windows.net/vhds/CentOS.vhd?sp=r&st=xx%2012:10:45&5;CA%aEMpls3D
+# replace as https://xx.core.windows.net/vhds/CentOS.vhd***
+PATTERN_URL = (
+    re.compile(r"(https?://([-\w]+\.)+[-\w]+(/[-./\w]*)?)(\?[\w]+=[%&-:;=\w]*)$"),
```
squirrelsc: The pattern should support normal url too, so it doesn't need to check if it's a normal url or not.

`(https?://([-\w]+\.)+[-\w]+(/[-./\w]*)?\??)([\w]+=[%&-:;=\w]*)?$`

lisa/variable.py
```diff
@@ -300,7 +300,13 @@ def add_secrets_from_pairs(
                 'secret variable, follow the formant "s:name:value"'
                 f"The raw error: {identifier}"
             )
-        _add_variable(key, value, results, is_secret=is_secret)
+        match = secret.PATTERN_URL[0].findall(str(value))
```
squirrelsc: as we discussed, it doesn't need to handle urls here. The pattern should be replaced when add_secret again.

lisa/secret.py
lubaihua33: If so, the normal url will also print as "https://xx.core.windows.net/vhds/CentOS.vhd***", but the "***" is not we want.

lubaihua33: If we use this pattern and don't check if it's a normal url or not, just `add_secret`, I have tried this. There are two kinds of print when we set different masks for PATTERN_URL[1]:
1.
```
PATTERN_URL = (
    re.compile(r"(https?://([-\w]+\.)+[-\w]+(/[-./\w]*)?\??)([\w]+=[%&-:;=\w]*)?$"),
    r"\1***",
)
```
then it will print as "https://xx.core.windows.net/vhds/CentOS.vhd***" even for the normal url.
2.
```
PATTERN_URL = (
    re.compile(r"(https?://([-\w]+\.)+[-\w]+(/[-./\w]*)?\??)([\w]+=[%&-:;=\w]*)?$"),
    r"\1",
)
```
then in "replace" function in secret.py, it has the below check:
```
        result = mask.sub(configured_sub, origin)
        if result == origin:
            # failed and fallback
            result = sub
```
After `mask.sub`, the origin and the result are the same, then the normal url is masked as `sub` which is "******“.

I think these 2 prints are not friendly.

lisa/variable.py
lubaihua33: I changed add_secret as below:
```
def add_secret(
    origin: Any,
    mask: Optional[Union[Pattern[str], Tuple[Pattern[str], str]]] = None,
    sub: str = "******",
) -> None:
    global _secret_list
    if origin:
        if not isinstance(origin, str):
            origin = str(origin)
        if origin in _secret_set:
            for i in range(len(_secret_list)):
                if origin == _secret_list[i][0]:
                    _secret_list[i] = (origin, replace(origin, sub=sub, mask=mask))
                    break
        else:
            _secret_set.add(origin)
            _secret_list.append((origin, replace(origin, sub=sub, mask=mask)))
            # deal with longer first, in case it's broken by shorter
            _secret_list = sorted(_secret_list, reverse=True, key=lambda x: len(x[0]))
```
This works well. 
After calling ` __post_init__`  of `AzureNodeSchema`, the vhd url mask is replaced as the new one in _secret_list. I can see this when debugging.  

But after  ` __post_init__` , it no longer prints vhd parameter info. So we still couldn't see the vhd file name, but just "******".

squirrelsc: it's ok. The first priority is to protect the secret, secondly, it can show the original value help troubleshooting. Both requirements meet.

lisa/secret.py
squirrelsc: The vhd*** is acceptable, and we know if there is no "?", it means no parameters. And it doesn't need to break the current code logic. The "replace" in secret is dangerous, it may cause the mask failed in some cases.

lisa/variable.py
lubaihua33: OK, I posted a new update, please help to review it again.

squirrelsc (approved):

PR #2172, opened 2022-08-01
Enable Deployment of CVM

kamalca:
security_type can be TrustedLaunch or ConfidentialVm.
secure_boot_enabled and v_tpm_enabled can both be True or False.
These values can all be set through the runbook.


kamalca:
Hello @squirrelsc, I have gotten CVM deployment to work. I wanted to sync with you again about design choices.
This enables all three CVM related settings to be set through the runbook. I tried to leave the functionality of the existing security_profile feature intact. I was not sure how to best handle when runbook settings and testcase settings conflict. As written, the runbook settings will overwrite any testcase settings for security_profile.

lisa/features/security_profile.py
```diff
@@ -16,13 +16,18 @@
 @dataclass()
 class SecurityProfileSettings(schema.FeatureSettings):
     type: str = FEATURE_NAME_SECURITY_PROFILE
+    security_type: str = ""
```
squirrelsc: Can we use only one enum for the three fields? The default value is "disabled", and it can be "boot" or "full". The string is easy to make mistake.

squirrelsc:
> Hello @squirrelsc, I have gotten CVM deployment to work. I wanted to sync with you again about design choices. This enables all three CVM related settings to be set through the runbook. I tried to leave the functionality of the existing security_profile feature intact. I was not sure how to best handle when runbook settings and testcase settings conflict. As written, the runbook settings will overwrite any testcase settings for security_profile.

Overall, it's ok to me, except use less parameters. I think there is no test case conflict, because the secure_boot should be included in CVM. The parameter can be a NodeSpace, so the test case level can set it to secure boot or CVM. So the test case can be run on both settings.

lisa/features/security_profile.py
kamalca: I confirmed with Johnson that CVM validation does not require the ability to boot CVM without secure boot. I also don't think there are any cases where we would need TrustedLaunch without vTpm or secureBoot. Therefore, I have reduced the three fields to one enum with the options "none" (default), "boot", and "cvm". This also removes the possibility for runbook and test requirements to conflict (as you mentioned below).

lisa/features/security_profile.py
```diff
@@ -32,7 +33,11 @@ class SecurityProfile(Feature):
     @classmethod
     def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
         settings = cast(SecurityProfileSettings, kwargs.get("settings"))
-        if settings.secure_boot_enabled:
+        runbook: Any = cast(Any, kwargs.get("arm_parameters")).nodes[0]
+        if runbook.security_profile != constants.SECURITY_PROFILE_NONE:
+            settings.security_profile = runbook.security_profile
+            cls._enable_secure_boot(*args, **kwargs)
+        elif settings.security_profile != constants.SECURITY_PROFILE_NONE:
```
dsrivastavv: Please throw an error if there is no match

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1334,11 +1334,39 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
         template: Any = kwargs.get("template")
         log = cast(Logger, kwargs.get("log"))
-        log.debug("updating arm template to support secure boot.")
         resources = template["resources"]
         virtual_machines = find_by_name(resources, "Microsoft.Compute/virtualMachines")
-        virtual_machines["properties"].update(json.loads(cls._both_enabled_properties))
+
+        if constants.SECURITY_PROFILE_BOOT == settings.security_profile:
+            log.debug("Updating arm template to support secure boot.")
+            security_type = "TrustedLaunch"
```
dsrivastavv, line 1444: [nit] put these strings in constant file

lisa/features/security_profile.py
```diff
@@ -16,13 +17,13 @@
 @dataclass()
 class SecurityProfileSettings(schema.FeatureSettings):
     type: str = FEATURE_NAME_SECURITY_PROFILE
-    secure_boot_enabled: bool = False
+    security_profile: str = ""
```
squirrelsc: It's still a string, search DiskType in schema.py. So, other values will be failed. And it needs a default value as "none".

lisa/features/security_profile.py
```diff
@@ -32,7 +33,11 @@ class SecurityProfile(Feature):
     @classmethod
     def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
         settings = cast(SecurityProfileSettings, kwargs.get("settings"))
-        if settings.secure_boot_enabled:
+        runbook: Any = cast(Any, kwargs.get("arm_parameters")).nodes[0]
```
squirrelsc: It's the arm_parameters, not runbook. The "settings" is the right place to check the field, although they are the same source.

LiliDeng:
@kamalca is any restriction for the VM sizes which support TVM? currently, if the size supports v2, then the SecurityProfile feature will be added.  if there is the restriction, we need add similar code like below in platform.py.
```
            elif name == "HyperVGenerations":
                if "V2" in str(sku_capability.value):
                    node_space.features.add(
                        schema.FeatureSettings.create(features.SecurityProfile.name())
                    )
```

lisa/features/security_profile.py
kamalca: Is there a better location for setting SecurityProfileSettings based on "arm_parameters"?

Is your suggestion to just rename the variable `runbook` to `arm_parameters`?

squirrelsc: I mean it should use the setting to check the field. It doesn't need the arm_template.

kamalca:
@squirrelsc I don't think this is quite right yet. I'd like your input on what I have so far. Thank you :)

lisa/features/security_profile.py
```diff
@@ -1,38 +1,78 @@
...
-from lisa import schema
+from lisa import schema, search_space
 from lisa.feature import Feature
+from lisa.util import LisaException, constants, field_metadata
 
 FEATURE_NAME_SECURITY_PROFILE = "Security_Profile"
 
 
+class SecurityProfileType(str, Enum):
+    Standard = constants.SECURITY_PROFILE_NONE
```
squirrelsc, line 18: The standard doesn't need. If some vm sizes don't support any, the feature won't be here.

lisa/features/security_profile.py
```diff
@@ -1,38 +1,78 @@
...
     def _get_key(self) -> str:
-        return f"{self.type}/{self.secure_boot_enabled}"
+        return f"{self.type}/{self.security_profile}"
 
     def _generate_min_capability(self, capability: Any) -> Any:
+        assert isinstance(
+            capability, SecurityProfileSettings
+        ), f"actual: {type(capability)}"
+
+        self.security_profile.intersection_update(capability.security_profile)
```
squirrelsc: It needs,
1. create a deep copy of self.
2. call get min capability of setspace, and then return the generated min capability.
3. return the copied instance

lisa/features/security_profile.py
squirrelsc: Is it changed to use settings? Besides that, it needs to update runbook of each node to use the deployed type of CVM.

lisa/features/security_profile.py
```diff
@@ -55,4 +95,9 @@ def enabled(self) -> bool:
         return True
 
 
-SecureBootEnabled = partial(SecurityProfileSettings, secure_boot_enabled=True)
+SecureBootEnabled = partial(
```
squirrelsc, line 108: Create a new one called CvmEnabled

lisa/features/security_profile.py
```diff
@@ -55,4 +95,9 @@ def enabled(self) -> bool:
...
+    SecurityProfileSettings,
+    security_profile=search_space.SetSpace(
+        True, [SecurityProfileType.Boot, SecurityProfileType.CVM]
```
squirrelsc, line 111: Leave only Boot here, because it's used as the test case requirement, not capability.

lisa/features/security_profile.py
squirrelsc: It needs to rewrite the `check` to make sure the `boot` is supported, even it's `CVM`

lisa/features/security_profile.py
```diff
@@ -1,38 +1,78 @@
...
         return self
 
 
 class SecurityProfile(Feature):
     @classmethod
     def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
         settings = cast(SecurityProfileSettings, kwargs.get("settings"))
-        if settings.secure_boot_enabled:
+        runbook: Any = cast(Any, kwargs.get("arm_parameters")).nodes[0]
+        try:
```
squirrelsc: Sorry, I just found the previous bug. The `on_before_deployment` should be in the Azure feature level, not general here.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -110,6 +111,19 @@ class AzureNodeSchema:
         default=None, metadata=field_metadata(data_key="shared_gallery")
     )
     vhd: str = ""
+    security_profile: Optional[SecurityProfileType] = field(
```
squirrelsc: Not add it into the node space. It's a general feature, doesn't like disk or network.

lisa/features/security_profile.py
kamalca: Here is the case I was considering when I wrote this:

Test case requires SecureBoot
Runbook specifies CVM

Is it okay to run the test case with CVM security profile?

squirrelsc: Yes, it sounds the test case should be run, and it needs to make the check method correct.

lisa/features/security_profile.py
kamalca: What does this change?

squirrelsc: The lisa/features is platform independent. The arm template is azure only. So the change should be in Azure folder.

lisa/sut_orchestrator/azure/common.py
kamalca: Should it be in AzurePlatformSchema?
I'm having a hard time figuring out how to get it from AzurePlatformSchema to settings or arm_parameters

squirrelsc: It should be added automatically into features of a capability.

kamalca: I don't understand.

1. Which schema should the security_profile be in
2. Where should security_profile be in the runbook
3. How does the field security_profile get turned into a SecurityProfileSettings

squirrelsc: 1. The schema itself is in right place already (features/security_profile)
2. In runbook, it's a settings schema. Schema is Settings, the SecurityProfile is the feature class, not schema. I have a try, it  should be added like below, but it doesn't work. I will fix it.
3. It's not valid, the settings will convert to a feature in Features class.

image

squirrelsc: PS: it should work as I verified. Let me know if there is any problem.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -117,6 +117,7 @@ class AzureNodeSchema:
         default=None, metadata=field_metadata(data_key="shared_gallery")
     )
     vhd: str = ""
+    osdisk_size_in_gb: int = 30
```
squirrelsc: What's this, is it merged incorrectly?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1409,24 +1410,62 @@ def create_setting(
...
 
-        return None
+        return SecurityProfileSettings(
+            security_profile=search_space.SetSpace(True, capabilities)
+        )
+
+    @classmethod
+    def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
+        settings = cast(SecurityProfileSettings, kwargs.get("settings"))
+        if SecurityProfileType.Standard not in settings.security_profile:
```
squirrelsc: The setting should be only one value, not a list. It just compares if the security_profile is Standard or not.

lisa/util/constants.py
```diff
@@ -102,6 +102,12 @@
 ENVIRONMENT_KEEP_NO = "no"
 ENVIRONMENT_KEEP_FAILED = "failed"
 
+SECURITY_PROFILE_NONE = "none"
+SECURITY_PROFILE_BOOT = "boot"
+SECURITY_PROFILE_CVM = "cvm"
+SECURITY_PROFILE_BOOT_SETTING = "TrustedLaunch"
```
squirrelsc: The two values belongs to Azure, so they shouldn't be LISA level consts.

squirrelsc: And it looks it's used only once, it doesn't need to be constants.

lisa/sut_orchestrator/azure/features.py
kamalca: @squirrelsc Disagrees with this suggestion. Marking as resolved.

squirrelsc (dismissed):

lisa/features/security_profile.py
kamalca: Closing this comment as it is outdated

squirrelsc (approved):

PR #2176, opened 2022-08-02
copy vhd into the testing location when source vhd location is different

LiliDeng:
just a draft, 3 things need to be fixed.
- when reuse rg, this logic still be executed, need fix.
- need shorten the copied vhd, currently it is too long.
- also if the vhd has been copied, if the sas url is different, it will still be copied again.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -4,6 +4,7 @@
 import re
 import sys
 from dataclasses import InitVar, dataclass, field
+from datetime import datetime, timedelta
 from threading import Lock
 from time import sleep
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
@@ -17,7 +18,13 @@ from azure.mgmt.storage.models import (  # type: ignore
     Sku,
     StorageAccountCreateParameters,
 )
-from azure.storage.blob import BlobServiceClient, ContainerClient
+from azure.storage.blob import (
+    AccountSasPermissions,
+    BlobServiceClient,
+    ContainerClient,
+    ResourceTypes,
+    generate_account_sas,
+)
 from azure.storage.fileshare import ShareServiceClient  # type: ignore
 from dataclasses_json import dataclass_json
 from marshmallow import validate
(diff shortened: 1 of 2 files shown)


squirrelsc:
Can you provide the reason of this change in the commit description?

squirrelsc:
> * also if the vhd has been copied, if the sas url is different, it will still be copied again.
     Is it expected behavior, or something you need to fix?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -488,13 +495,38 @@ def get_storage_credential(
     return {"account_name": account_name, "account_key": key.value}
 
 
-def get_or_create_storage_container(
+def generate_sas(
```
squirrelsc: `generate_sas_token` as the returned value name.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2016,8 +2050,17 @@ def _get_deployable_vhd_path(
         """
         matches = SAS_URL_PATTERN.match(vhd_path)
         if not matches:
-            return vhd_path
-        log.debug("found the vhd is a sas url, it may need to be copied.")
+            vhd_path = self._check_and_generate_sas_url(vhd_path, location, log)
+            matches = SAS_URL_PATTERN.match(vhd_path)
+            if not matches:
+                return vhd_path
+            log.debug(
```
squirrelsc: Why not direct copy without sas token?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2084,6 +2127,26 @@ def _get_deployable_vhd_path(
 
         return full_vhd_path
 
+    def _get_storage_details(self, vhd_path: str) -> Any:
```
squirrelsc: `_get_vhd_details`?

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: I tired, but didn't work, the exception is 
```
azure.core.exceptions.ClientAuthenticationError: Server failed to authenticate the request. Please refer to the information in the www-authenticate header.
RequestId:ccc36585-601e-000e-2959-a6a9db000000
Time:2022-08-02T10:17:54.9250330Z
ErrorCode:CannotVerifyCopySource
Content: <?xml version="1.0" encoding="utf-8"?><Error><Code>CannotVerifyCopySource</Code><Message>Server failed to authenticate the request. Please refer to the information in the www-authenticate header.
RequestId:ccc36585-601e-000e-2959-a6a9db000000
Time:2022-08-02T10:17:54.9250330Z</Message></Error>
```

LiliDeng: probably I need find out another way, for this method start_copy_from_url, it needs to be ` The source blob must either be public or must be authenticated via a shared access signature. `

image

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/platform_.py

LiliDeng:
@squirrelsc during my testing, I found if set concurrency as 2, then there are 2 sas url generated, vhd will be copied twice, can this design be accepted?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1937,8 +1969,17 @@ def _get_deployable_vhd_path(
         """
         matches = SAS_URL_PATTERN.match(vhd_path)
         if not matches:
-            return vhd_path
-        log.debug("found the vhd is a sas url, it may need to be copied.")
+            vhd_path = self._generate_sas_token(vhd_path, location, log)
```
squirrelsc: This should be split into two method. It doesn't need to check if it's sas token again. 1) Check if it's in the same location. If yes, return the vhd path. 2) Generate sas token, and go with next logics.

squirrelsc (approved):

PR #2179, opened 2022-08-03
DPDK: Cleanup queue and core count selection logic

mcgov:
Cleaning up perf test and queue/core selection logic.

Noting here that the multicore/hugevm tests should eventually support multiple nics. Currently we do not make use of multiple nics because it is more complicated to configure them. A future PR will have to implement this support.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkperf.py
+++ b/microsoft/testsuites/dpdk/dpdkperf.py
@@ -47,7 +47,7 @@ class DpdkPerformance(TestSuite):
             supported_features=[IsolatedResource],
         ),
     )
-    def perf_dpdk_failsafe_pmd_dual_core(
+    def perf_dpdk_failsafe_pmd_minimal(
         self,
         result: TestResult,
         log: Logger,
@@ -58,7 +58,7 @@ class DpdkPerformance(TestSuite):
 
     @TestCaseMetadata(
         description="""
-        DPDK Performance: failsafe mode, maximal core count, default queue settings
+        DPDK Performance: failsafe mode, multiple service cores
         """,
         priority=3,
         requirement=simple_requirement(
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -86,8 +86,8 @@ def perf_dpdk_failsafe_pmd_multi_core(
         requirement=simple_requirement(
             min_count=2,
             network_interface=Sriov(),
-            min_nic_count=2,
-            min_core_count=48,
+            min_nic_count=8,
```
dsrivastavv, line 89: Why is there different `min_nic_count` in this (8) vs earlier(3) test?

mcgov: Ah good eye, just a bug. The requirements should be the same for both sets (failsafe/netvsc)

mcgov: The implication of the core assignment change is cores are now assigned as queues * nics + nics
Maybe it's worth renaming the tests, but the changes in the tests are just modifying each of those parameters. 
'Multicore' tests increase N for nics, multiqueue increase N for queues, huge_vm increase both.

mcgov: Ok I've written two explanations and I think I actually want to make a code change.
We have a matrix of
queues X cores X nics X vmsize
and I was thinking of making the core size implicit based on the amount of nics. I think I want to change this to make core count not implicit.

microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -76,41 +76,43 @@ def perf_dpdk_failsafe_pmd_multi_core(
...
+        DPDK Performance: failsafe mode, multiple service cores, max nics
         """,
         priority=3,
         requirement=simple_requirement(
             min_count=2,
             network_interface=Sriov(),
-            min_nic_count=2,
-            min_core_count=48,
+            min_nic_count=8,
+            min_core_count=64,
```
squirrelsc: I'm almost done the IsolatedResource feature (called exclusive VM). Do you mind change all "min_core_count=64" to this feature? The source is from https://docs.microsoft.com/en-us/azure/security/fundamentals/isolation-choices#compute-isolation. And if new VM sizes are not here, it can be added separatedly.

mcgov: Ah I see this was done for me.

microsoft/testsuites/dpdk/dpdkperf.py
mcgov: Fixed up

mcgov:
@LiliDeng  or @squirrelsc  can I get another eye on this? Should be ready to merge.

mcgov:
Identified that the issue I noticed during testing will be fixed by PR https://github.com/microsoft/lisa/pull/2204

squirrelsc (approved):

PR #2181, opened 2022-08-03
Change CommunityTest name to SubTest

dsrivastavv:
The SubTest framework in LISA can be used for any test which needs to send out muliple subtest results

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -9,7 +9,7 @@ from assertpy.assertpy import assert_that
 
 from lisa import Environment, notifier
 from lisa.executable import Tool
-from lisa.messages import CommunityTestMessage, TestStatus, create_test_result_message
+from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
 from lisa.operating_system import CBLMariner
 from lisa.testsuite import TestResult
 from lisa.tools import Docker, Echo, Git, Whoami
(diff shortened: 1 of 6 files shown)


microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -148,7 +148,7 @@ def _send_community_test_msg(
         test_status: TestStatus,
     ) -> None:
         community_msg = create_test_result_message(
```
squirrelsc: change those names too, search "community" to try best to change names.

dsrivastavv: The reason I skipped changing these is because these actually are "community" tests. Our reason for updating naming from community to subtest is to accommodate any test framework which follow a similar pattern in future, and doesn't fall in the category of "community" test. It's the similar case for `xfs` and `ltp` too. Please let me know if you think otherwise

squirrelsc: The "community" is the original concept, which is replaced by "sub tests". So, it doesn't matter it's community test or not. The reason to send subtests message, is we won't dump test results for each "subtests". Yes, it is community tests, and it's also sub tests. We don't care about whether it's community tests now. Less concept, more clear.

microsoft/testsuites/xdp/functional.py
```diff
@@ -429,7 +429,7 @@ def verify_xdp_remove_add_vf(self, node: Node, log: Logger) -> None:
         """,
         priority=3,
     )
-    def verify_xdp_community_test(self, node: Node) -> None:
+    def verify_xdp_subtest(self, node: Node) -> None:
```
squirrelsc: This doesn't need to be changed. Because it's really "community" tests, not our customized tests. Only when we talk the test cases inside of LISA, there are test results and subtest results.

dsrivastavv: Missed changing it back. fixed it!

squirrelsc (approved):

PR #2183, opened 2022-08-04
Azure: move feature detections from platform to feature.

squirrelsc:
Please review by commits, in case there is any regression.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -1449,6 +1449,36 @@ class ACC(AzureFeatureMixin, features.ACC):
         return None
 
 
+class NestedVirtualization(AzureFeatureMixin, features.NestedVirtualization):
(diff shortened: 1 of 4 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -77,6 +77,12 @@ def _initialize_information(self, node: Node) -> None:
 
 
 class StartStop(AzureFeatureMixin, features.StartStop):
+    @classmethod
+    def check_supported(
```
dsrivastavv: Can we rename it something else? `check_supported` seems like a method name that should return a bool. Maybe something like `get_feature_if_supported`?

squirrelsc: good suggestion, will rename to `create_setting`

dsrivastavv (approved):

PR #2184, opened 2022-08-06
Azure: support to specify multiple locations.

squirrelsc:
If there are multiple locations, it can be split by "," like "eastus,westus2".


dsrivastavv:
@squirrelsc Do we need to update docs to reflect this change?

squirrelsc:
> @squirrelsc Do we need to update docs to reflect this change?

Sure, I will update it.

dsrivastavv (approved):

PR #2189, opened 2022-08-09
Fix ping permission issue and ssh is not active when retry many times to check nic status.

LiliDeng:


lisa/tools/ping.py
```diff
@@ -29,7 +37,7 @@ def ping_async(
         if package_size:
             args += f" -s {package_size}"
 
-        return self.run_async(args, force_run=True)
+        return self.run_async(args, force_run=True, sudo=sudo)
```
squirrelsc, line 40: keep arguments consistent between sync and async v3ersion. if the sudo must be True, remove it from async method, and set it always True.

lisa/tools/ping.py
```diff
@@ -48,7 +57,17 @@ def ping(
             count=count,
             interval=interval,
             package_size=package_size,
+            sudo=sudo,
         ).wait_result()
+        if not sudo and self._no_permission_pattern.findall(result.stdout):
```
squirrelsc, line 65: Sorry, I haven't noticed, this logic called ping twice. Is there any reason to call it second time? Can you add comments in code?

LiliDeng: For some distro, it needs sudo, on these distros, will retry with sudo.

squirrelsc: Please add it in code comments.

squirrelsc (approved):

PR #2191, opened 2022-08-09
update dpdk tests use IsolatedResource feature

LiliDeng:


microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -67,6 +68,7 @@ def perf_dpdk_failsafe_pmd_dual_core(
             min_nic_count=2,
             min_core_count=8,
```
squirrelsc: remove the core count requirement.

squirrelsc (approved):

PR #2195, opened 2022-08-10
restore nics status into origin one to avoid side-effect for following cases.

LiliDeng:
- use new env for cases which will change nics
- store nics status before testing, restore them after testing, not put it into test case cleanup, since it will affect following cases.

### DIFF
--- a/lisa/features/network_interface.py
+++ b/lisa/features/network_interface.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 from functools import partial
-from typing import Type
+from typing import Any, Type
 
 from lisa import schema
 from lisa.feature import Feature
@@ -40,6 +40,10 @@ class NetworkInterface(Feature):
 
     def get_nic_count(self, is_sriov_enabled: bool = True) -> int:
         raise NotImplementedError
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.origin_extra_synthetic_nics_count: int = 0
+        self.origin_extra_sriov_nics_count: int = 0
 
 
 Sriov = partial(NetworkInterfaceOptionSettings, data_path=schema.NetworkDataPath.Sriov)
--- a/microsoft/testsuites/network/common.py
+++ b/microsoft/testsuites/network/common.py
@@ -222,3 +222,18 @@ def sriov_disable_enable(environment: Environment, times: int = 4) -> None:
         if sriov_is_enabled:
             sriov_basic_test(environment, vm_nics)
         network_interface_feature.switch_sriov(enable=(not sriov_is_enabled))
+
+
+def restore_extra_nics(environment: Environment) -> None:
+    remove_extra_nics(environment)
+    # restore nics info into previous status
+    for node in environment.nodes.list():
+        network_interface_feature = node.features[NetworkInterface]
+        network_interface_feature.attach_nics(
+            network_interface_feature.origin_extra_sriov_nics_count,
+            enable_accelerated_networking=True,
+        )
+        network_interface_feature.attach_nics(
+            network_interface_feature.origin_extra_synthetic_nics_count,
+            enable_accelerated_networking=False,
+        )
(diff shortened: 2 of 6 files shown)


lisa/features/network_interface.py
```diff
@@ -38,7 +38,9 @@ def remove_extra_nics(self) -> None:
     def reload_module(self) -> None:
         raise NotImplementedError
 
-    def get_nic_count(self, is_sriov_enabled: bool = True) -> int:
+    def get_nic_count(
+        self, is_sriov_enabled: bool = True, include_primary: bool = True
```
squirrelsc: Can it just be got by minus 1? So it doesn't need to add the extra argument.

microsoft/testsuites/network/common.py
```diff
@@ -222,3 +222,33 @@ def sriov_disable_enable(environment: Environment, times: int = 4) -> None:
         if sriov_is_enabled:
             sriov_basic_test(environment, vm_nics)
         network_interface_feature.switch_sriov(enable=(not sriov_is_enabled))
+
+
+def get_extra_nics(environment: Environment) -> Dict[str, int]:
```
squirrelsc: It needs to maintain the original_nic_count in networkadapter feature, instead of count and save it in test case level.

microsoft/testsuites/network/common.py
```diff
@@ -222,3 +222,33 @@ def sriov_disable_enable(environment: Environment, times: int = 4) -> None:
...
+        )
+        sriov_nic_count = network_interface_feature.get_nic_count(
+            is_sriov_enabled=True, include_primary=False
+        )
+        nics_info[f"{node.name}_synthetic"] = synthetic_nic_count
+        nics_info[f"{node.name}_sriov"] = sriov_nic_count
+    return nics_info
+
+
+def restore_extra_nics(environment: Environment, nics_info: Dict[str, int]) -> None:
```
squirrelsc: The "remove" should be called in "restore".

lisa/features/network_interface.py
LiliDeng: we are not sure, the primary one is sriov or synthetic?

squirrelsc: As the logic, it always finds a primary. What's different if it's sriov or synthetic? In this place, it looks like just need a number.

LiliDeng: If primary is sriov, others are synthetic, this method will return all sriov nic count, should we minus 1 or not? 
The code will get extra sriov nics and extra synthetic nics, then restore them after testing (which need update nics.)

LiliDeng:
will update PR tomorrow

microsoft/testsuites/network/common.py
LiliDeng: update

lisa/features/network_interface.py
LiliDeng: update

microsoft/testsuites/network/common.py
LiliDeng: update

LiliDeng:
@squirrelsc anything need update?

lisa/features/network_interface.py
```diff
@@ -41,6 +41,9 @@ def reload_module(self) -> None:
     def get_nic_count(self, is_sriov_enabled: bool = True) -> int:
         raise NotImplementedError
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.origin_extra_nics: Dict[str, int] = {}
```
squirrelsc: 1. It doesn't need to be a dict, use two variables to represent two numbers. 
2. I don't understand the "origin extra" well. Does it mean "origin" (created with VM) and "secondary" (not primary) nics? Please name it like secondary or something else, and specify what origin means in some comments.
3. How about other platforms than Azure? Can they use the same way to calculate the count?

squirrelsc (approved):

PR #2196, opened 2022-08-10
Move exit code check to run function

### LINKED ISSUE #1804: Move exit code check to run function

adityagesh:
Fixes #1804

### DIFF
--- a/lisa/base_tools/cat.py
+++ b/lisa/base_tools/cat.py
@@ -21,7 +21,11 @@ class Cat(Tool):
     ) -> str:
         # Run `cat <file>`
         result = self.run(
-            file, force_run=force_run, sudo=sudo, shell=True, no_debug_log=no_debug_log
+            file,
+            force_run=force_run,
+            sudo=sudo,
+            shell=True,
+            no_debug_log=no_debug_log,
         )
         result.assert_exit_code(message=f"Error : {result.stdout}")
         return result.stdout
--- a/lisa/base_tools/wget.py
+++ b/lisa/base_tools/wget.py
@@ -82,10 +82,13 @@ class Wget(Tool):
                 f" stdout: {command_result.stdout}"
             )
         actual_file_path = self.node.execute(
-            f"ls {download_file_path}", shell=True, sudo=sudo
+            f"ls {download_file_path}",
+            shell=True,
+            sudo=sudo,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="File path does not exist, "
+            f"{download_file_path}",
         )
-        if actual_file_path.exit_code != 0:
-            raise LisaException(f"File {actual_file_path} doesn't exist.")
         if executable:
             self.node.execute(f"chmod +x {actual_file_path}", sudo=sudo)
 
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -65,8 +65,7 @@ class Waagent(Tool):
         # the vm need to be exported clearly, it needs to remove the current
         # user with below command:
         # self.run("-deprovision+user --force", sudo=True)
-        result = self.run("-deprovision --force", sudo=True)
-        result.assert_exit_code()
+        self.run("-deprovision --force", sudo=True, expected_exit_code=0)
 
     def get_configuration(self) -> Dict[str, str]:
         if isinstance(self.node.os, CoreOs):
@@ -237,11 +236,11 @@ class LisDriver(Tool):
 
     def _install(self) -> bool:
         result = self.install_from_iso()
-        if result.exit_code != 0:
-            raise LisaException(
-                f"Unable to install the LIS RPMs! exit_code: {result.exit_code}"
-                f"stderr: {result.stderr}"
-            )
+        result.assert_exit_code(
+            0,
+            f"Unable to install the LIS RPMs! exit_code: {result.exit_code}"
+            f"stderr: {result.stderr}",
+        )
         self.node.reboot(360)
         return True
 
--- a/lisa/tools/date.py
+++ b/lisa/tools/date.py
@@ -6,7 +6,6 @@ from datetime import datetime
 from dateutil.parser import parser
 
 from lisa.executable import Tool
-from lisa.util import LisaException
 
 
 class Date(Tool):
@@ -20,8 +19,7 @@ class Date(Tool):
     def current(self, no_error_log: bool = True) -> datetime:
         # always force run to get current date time.
         command_result = self.run(no_error_log=no_error_log, force_run=True, timeout=10)
-        if command_result.exit_code != 0:
-            raise LisaException(
-                f"'Date' return non-zero exit code: {command_result.stderr}"
-            )
+        command_result.assert_exit_code(
+            0, f"'Date' return non-zero exit code: {command_result.stderr}"
+        )
         return parser().parse(command_result.stdout)
--- a/lisa/tools/dhclient.py
+++ b/lisa/tools/dhclient.py
@@ -7,7 +7,7 @@ from typing import Optional
 from lisa.base_tools import Cat
 from lisa.executable import Tool
 from lisa.operating_system import Debian, Fedora, Suse
-from lisa.util import LisaException, UnsupportedDistroException, find_group_in_lines
+from lisa.util import UnsupportedDistroException, find_group_in_lines
 
 
 class Dhclient(Tool):
@@ -70,7 +70,6 @@ class Dhclient(Tool):
             )
         else:
             result = self.run("-r && dhclient", shell=True, sudo=True)
-        if result.exit_code != 0:
-            raise LisaException(
-                f"dhclient renew return non-zero exit code: {result.stdout}"
-            )
+        result.assert_exit_code(
+            0, f"dhclient renew return non-zero exit code: {result.stdout}"
+        )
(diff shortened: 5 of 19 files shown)


lisa/base_tools/wget.py
```diff
@@ -84,8 +84,7 @@ def get(
         actual_file_path = self.node.execute(
             f"ls {download_file_path}", shell=True, sudo=sudo
         )
-        if actual_file_path.exit_code != 0:
-            raise LisaException(f"File {actual_file_path} doesn't exist.")
+        actual_file_path.assert_exit_code(0, f"File {actual_file_path} doesn't exist.")
```
dsrivastavv: Thanks for helping with this. This needs to be moved into `node.execute`. Please look at the following arguments in the `node.execute` function: `expected_exit_code`, `expected_exit_code_failure_message`. Also, please look at my comment in #1804. Similar comments apply to other changes in this PR.

dsrivastavv: For example, this is what should go here:
```
actual_file_path = self.node.execute(
             f"ls {download_file_path}", shell=True, sudo=sudo,
             expected_exit_code=0,
             expected_exit_code_failure_message=f"File path doesn't exist"
         )
```

adityagesh: Wouldn't it be more difficult to debug without the 'actual_file_path' displayed in the log?

Should the same pattern be followed for the messages that includes result.stderr and result.stdout? ([ex](https://github.com/microsoft/lisa/pull/2196/files#diff-f8bd130bf4a96ebae2870059912174d37c0e836c66aab2a1a6410aea462b4c43R37))

dsrivastavv: Maybe you can add `download_file_path` to the logs. That should be enough. We also have some logs thrown from `node.execute` which should be enough to debug this issue.

lisa/tools/lsvmbus.py
```diff
@@ -205,12 +205,15 @@ def get_device_channels(self, force_run: bool = False) -> List[VmBusDevice]:
...
-                        f"get unexpected non-zero exit code {result.exit_code} "
-                        f"when run {self.command} -vv."
-                    )
+                result = self.run(
+                    "-vv",
+                    force_run=force_run,
+                    shell=True,
+                    sudo=True,
+                    expected_exit_code=0,
+                    expected_exit_code_failure_message="get unexpected non-zero "
```
dsrivastavv: [nit] *got

lisa/tools/uptime.py
```diff
@@ -22,12 +21,14 @@ def _check_exists(self) -> bool:
...
-            raise LisaException(
-                f"get unexpected non-zero exit code {command_result.exit_code} "
-                f"when run {self.command} -s."
-            )
+        command_result = self.run(
+            "-s",
+            force_run=True,
+            no_error_log=no_error_log,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="get unexpected non-zero exit code"
```
dsrivastavv: Can you please check if we need this message? When self.run fails it would log the command and parameters. So this message might be redundant

lisa/base_tools/wget.py
dsrivastavv: As discussed, please follow @squirrelsc suggestion related to stderr/stdout logging in PR #2083

dsrivastavv:
We "rebase and merge" to the main branch. So we need to ensure that each commit is meaningful and self contained, since each commit goes as-it-is to the main branch. Can you please ensure it for this PR? Please feel free to reach out if you need any help/clarification with this.

lisa/base_tools/cat.py
```diff
@@ -21,9 +21,13 @@ def read(
...
         result = self.run(
-            file, force_run=force_run, sudo=sudo, shell=True, no_debug_log=no_debug_log
+            file,
+            force_run=force_run,
+            sudo=sudo,
+            shell=True,
+            no_debug_log=no_debug_log,
+            expected_exit_code=0,
         )
-        result.assert_exit_code(message=f"Error : {result.stdout}")
```
dsrivastavv, line 26: As discussed this stdout/stderr needs to be handled separately. Other changes look good to me

adityagesh: reverted

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -223,7 +222,13 @@ def install_from_iso(self) -> ExecutableResult:
 
     def uninstall_from_iso(self) -> ExecutableResult:
         lis_folder_path = self.download()
-        return self.node.execute("./uninstall.sh", cwd=lis_folder_path, sudo=True)
+        return self.node.execute(
+            "./uninstall.sh",
+            cwd=lis_folder_path,
+            sudo=True,
+            expected_exit_code=0,
```
squirrelsc: Don't add assert here without test.

lisa/tools/date.py
```diff
@@ -19,9 +18,7 @@ def _check_exists(self) -> bool:
 
     def current(self, no_error_log: bool = True) -> datetime:
         # always force run to get current date time.
-        command_result = self.run(no_error_log=no_error_log, force_run=True, timeout=10)
-        if command_result.exit_code != 0:
-            raise LisaException(
-                f"'Date' return non-zero exit code: {command_result.stderr}"
-            )
+        command_result = self.run(
+            no_error_log=no_error_log, force_run=True, timeout=10, expected_exit_code=0
```
squirrelsc: Use separated assert, because it needs stderr.

lisa/tools/dhclient.py
```diff
@@ -69,8 +69,11 @@ def renew(self, interface: Optional[str] = None) -> None:
...
             )
         else:
-            result = self.run("-r && dhclient", shell=True, sudo=True)
-        if result.exit_code != 0:
-            raise LisaException(
-                f"dhclient renew return non-zero exit code: {result.stdout}"
+            result = self.run(
+                "-r && dhclient",
+                shell=True,
+                sudo=True,
```
squirrelsc: Remove the unnecessary comma at the end, so all arguments can be in the same line as before.

lisa/tools/dmesg.py
```diff
@@ -88,6 +86,6 @@ def _run(self, force_run: bool = False) -> ExecutableResult:
         result = self.run(force_run=force_run, no_error_log=True)
         if result.exit_code != 0:
             # may need sudo
-            result = self.run(sudo=True, force_run=force_run)
+            result = self.run(sudo=True, force_run=force_run, expected_exit_code=0)
```
squirrelsc: Remove the extra assert code.

lisa/tools/dmesg.py
```diff
@@ -43,8 +43,6 @@ def check_kernel_errors(
         throw_error: bool = True,
     ) -> str:
         command_output = self._run(force_run=force_run)
-        if command_output.exit_code != 0:
```
squirrelsc, line 46: It's lost.

lisa/transformers/kernel_installer.py
```diff
@@ -214,11 +214,10 @@ def install(self) -> str:
...
         ubuntu.wait_running_package_process()
         result = node.execute(f'add-apt-repository -y "{repo_entry}"', sudo=True)
-        if result.exit_code != 0:
-            result.assert_exit_code(
-                message="failed on add repo\n"
-                + "\n".join(ubuntu.get_apt_error(result.stdout))
-            )
+
+        result.assert_exit_code(
+            0, "failed on add repo\n" + "\n".join(ubuntu.get_apt_error(result.stdout))
```
squirrelsc: remove the extra `" + "`, when the message is merged into one line.

microsoft/testsuites/lis/lissuite.py
```diff
@@ -131,9 +131,7 @@ def _verify_lis_preinstall_disk_size(
         stat = node.tools[Stat]
         df = node.tools[Df]
         rpm = node.tools[Rpm]
-        cmd_result = lisdriver.uninstall_from_iso()
-        if 0 != cmd_result.exit_code:
-            raise LisaException("fail to uninstall lis")
+        lisdriver.uninstall_from_iso()
```
squirrelsc: it's lost.

lisa/tools/dmesg.py
adityagesh: The check is moved to _run function
https://github.com/microsoft/lisa/pull/2196/files#diff-a66e4e010a83a2d730960fa1fa9f07818ac40a963321177890c49367022e462bR89


Should it be reverted?

lisa/tools/dmesg.py
adityagesh: Moved assert from check_kernel_errors
https://github.com/microsoft/lisa/pull/2196/files#r944091083

lisa/tools/dmesg.py
squirrelsc: If it's moved into _run, it means all calls to _run will assert the exit code. It changes the current behavior.

squirrelsc (approved):

PR #2198, opened 2022-08-11
Add RepoURLNotExistException.

LiliDeng:


lisa/util/__init__.py
```diff
@@ -112,6 +112,23 @@ def __str__(self) -> str:
         return message
 
 
+class RepoURLNotExistException(LisaException):
```
squirrelsc: It means repo does not exists, not url doesn't exist. So the simpler name is `RepoNotExsistException`.

lisa/util/__init__.py
```diff
@@ -112,6 +112,23 @@ def __str__(self) -> str:
...
+    This exception is used to indicate that a repo URL not existing issue
+    """
+
+    def __init__(self, os: "OperatingSystem", message: str = "") -> None:
+        self.name = os.name
+        self.version = os.information.full_version
+        self._extended_message = message
+
+    def __str__(self) -> str:
+        message = f"Repo URL not existing in '{self.version}'"
```
squirrelsc: remove the "URL".

lisa/util/__init__.py
squirrelsc: It can be used in other place like git repo and more.

lisa/operating_system.py
```diff
@@ -1520,9 +1527,13 @@ def add_repository(
...
     def _initialize_package_installation(self) -> None:
         self.wait_running_process("zypper")
-        self._node.execute(
+        output = self._node.execute(
             "zypper --non-interactive --gpg-auto-import-keys refresh", sudo=True
-        )
+        ).stdout
+        if self._no_repo_defined.search(output):
+            raise LisaException(
+                f"There are no enabled repositories defined in {self._node.os.name}"
```
squirrelsc: Is the os name enough to troubleshoot? Is it need the distro version?

LiliDeng: I'll check.
BTW, two possible reasons here
- one is the disto is byos one, no repo is by design
- the other issue is that register repo has issue

lisa/util/__init__.py

lisa/util/__init__.py
LiliDeng: removed

lisa/operating_system.py
LiliDeng: add distro version, then it can provide more info

lisa/operating_system.py
```diff
@@ -1520,9 +1527,14 @@ def add_repository(
 
     def _initialize_package_installation(self) -> None:
         self.wait_running_process("zypper")
-        self._node.execute(
+        output = self._node.execute(
             "zypper --non-interactive --gpg-auto-import-keys refresh", sudo=True
-        )
+        ).stdout
+        if self._no_repo_defined.search(output):
+            raise LisaException(
```
squirrelsc, line 1534: Why not use the RepoNotExistException here?

LiliDeng: I thought they are different issues.
For ubuntu one, the repo definition is there, but repo server is not there.
For sles one, it is either a byos image or encounter repo register issue, there is not any repo listed out if run `zypper lr`

squirrelsc (approved):

PR #2200, opened 2022-08-11
Add additional methods to Free tool, add first use in DPDK tool

mcgov:
Free tool previously could only check for swap info.
- Add a skeleton to check any field in the output.
- Add public methods to check for free memory in mebibytes or gibibytes.
- Add use of the tool in DPDK to check for memory availability before requesting hugepages

### DIFF
--- a/lisa/tools/free.py
+++ b/lisa/tools/free.py
@@ -4,20 +4,78 @@
 from lisa.executable import Tool
 from lisa.util import LisaException
 
+# example output
+#               total        used        free      shared  buff/cache   available
+# Mem:        16310304      441732    15438492        2448      430080    15600880
+# Swap:        4194304           0     4194304
+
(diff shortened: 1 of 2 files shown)


lisa/tools/free.py
```diff
@@ -4,20 +4,78 @@
...
+
+    def get_swap_size(self) -> int:
+        # Return total swap size in Mebibytes
+        return self._get_field("Swap", "total", "MiB")
+
+    def get_free_memory_mb(self) -> int:
+        return self._get_field("Mem", "free", "MiB")
 
-        raise LisaException("Failed to get swap size")
+    def get_free_memory_gb(self) -> int:
```
squirrelsc, line 80: Is it needed? It's easily from mb by *1024.

PR #2204, opened 2022-08-12
DPDK Memory Request fixes

mcgov:
- add get_numa_node_count to fetch how many numa nodes are present in Lscpu
- add Free tool memory check and warnings for DPDK tests when allocating hugepages
- allocate memory per-numa node when multiple are present

### DIFF
--- a/lisa/tools/free.py
+++ b/lisa/tools/free.py
@@ -4,20 +4,72 @@
 from lisa.executable import Tool
 from lisa.util import LisaException
 
+# example output
+#               total        used        free      shared  buff/cache   available
+# Mem:        16310304      441732    15438492        2448      430080    15600880
+# Swap:        4194304           0     4194304
+
 
 class Free(Tool):
     @property
     def command(self) -> str:
         return "free"
 
-    def get_swap_size(self) -> int:
-        # Return total swap size
+    #     Usage:
+    #  free [options]
+
+    # Options:
+    #  -b, --bytes         show output in bytes
+    #      --kilo          show output in kilobytes
+    #      --mega          show output in megabytes
+    #      --giga          show output in gigabytes
+    #      --tera          show output in terabytes
+    #      --peta          show output in petabytes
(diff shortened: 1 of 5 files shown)


lisa/tools/free.py
```diff
@@ -4,20 +4,78 @@
...
+    def get_swap_size(self) -> int:
+        # Return total swap size in Mebibytes
+        return self._get_field("Swap", "total", "MiB")
+
+    def get_free_memory_mb(self) -> int:
+        return self._get_field("Mem", "free", "MiB")
 
-        raise LisaException("Failed to get swap size")
+    def get_free_memory_gb(self) -> int:
+        return self._get_field("Mem", "free", "GiB")
```
mcgov: Responding to previous review question from @squirrelsc, I usually try to just wrap the tool's behavior.

mcgov: Actually digging in, maybe it makes more sense to just use the default and then shift? We'll lose granularity below KiB but the behavior and style of flags has changed over time, while KibiBytes have always been the default.

mcgov: yeah I'm switching it, good note thank you 🙂

mcgov:
Noticed an issue with cores when NIC is assigned to a different numa node, working on a fix

lisa/tools/free.py
squirrelsc: It's ok to keep KB level. We're not wrapping the tool for general purpose use, and it's for test cases of LISA.

mcgov: It was reviewing the git history that did it for me, wanting to avoid flag issues across versions

LiliDeng:
During my testing, I always see /sys/class/net/*/device/numa_node return 0, I didn't see any regression issues bring in by this change. The logic is too complex to read, if you can add more comments in code, that would be helpful.

LiliDeng (approved):

PR #2205, opened 2022-08-12
Fixup redhat/fedora usage in testpmd, refactor dpdk stable location

mcgov:
- Cleanup use of Redhat/Fedora to point to Fedora
- Refactor DPDK_STABLE to a single location for the various uses.

Second commit to force update RHEL to latest kernel before installation to fix rdma-core mismatch issues causing test results to report 0 packets per second.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -1220,6 +1220,25 @@ class Fedora(RPMDistro):
             sed = self._node.tools[Sed]
             sed.substitute("$releasever", "8", "/etc/yum.repos.d/epel*.repo", sudo=True)
 
+    def _verify_package_result(self, result: ExecutableResult, packages: Any) -> None:
+        # yum returns exit_code=1 if DNF handled an error with installation.
+        # We do not want to fail if exit_code=1, but warn since something may
+        # potentially have gone wrong.
+        if result.exit_code == 1:
+            self._log.debug(f"DNF handled error with installation of {packages}")
+        elif result.exit_code == 0:
+            self._log.debug(f"{packages} is/are installed successfully.")
+        else:
+            raise LisaException(
+                f"Failed to install {packages}. exit_code: {result.exit_code}"
+            )
+
+    def group_install_packages(self, group_name: str) -> None:
+        # trigger to run _initialize_package_installation
+        self._get_package_list(group_name)
+        result = self._node.execute(f'yum -y groupinstall "{group_name}"', sudo=True)
+        self._verify_package_result(result, group_name)
+
     def _get_information(self) -> OsInformation:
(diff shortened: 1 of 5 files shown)


squirrelsc:
@LiliDeng Please review the PR, it stays 3 days.

lisa/operating_system.py
```diff
@@ -1220,6 +1220,25 @@ def install_epel(self) -> None:
...
+        # yum returns exit_code=1 if DNF handled an error with installation.
+        # We do not want to fail if exit_code=1, but warn since something may
+        # potentially have gone wrong.
+        if result.exit_code == 1:
+            self._log.debug(f"DNF handled error with installation of {packages}")
+        elif result.exit_code == 0:
+            self._log.debug(f"{packages} is/are installed successfully.")
+        else:
+            raise LisaException(
+                f"Failed to install {packages}. exit_code: {result.exit_code}"
```
dsrivastavv, line 1233: I think we should also log stderr and stout

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -677,7 +677,7 @@ def _install_dependencies(self) -> None:
             node.os.install_packages(
                 list(self._debian_packages), extra_args=self._debian_backports_args
             )
-        elif isinstance(node.os, Redhat):
+        elif isinstance(node.os, Fedora):
             self._install_redhat_dependencies()
```
dsrivastavv: [nit] redhat->fedora

lisa/operating_system.py
mcgov: This is the code as it exists currently, I'm just hoisting it to the superclass here. Can we make that a separate PR?

mcgov: I think too we usually don't log the full stdout and stderr for these because debug level will already log that, unless we've explicitly disabled logging there

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: oh nice good catch

mcgov: fixed :)

lisa/operating_system.py
squirrelsc: If there are a lot of log, it's not good to output in exception. But if it's possible, the key information should be extracted, and added in the exception. But it should be done in another PR, if it's in priority.

dsrivastavv: Sure! This can be done in another PR.

dsrivastavv (approved):

LiliDeng (approved):

PR #2207, opened 2022-08-14
'-i' from debian returns unknown platform, change into -m

LiliDeng:


lisa/operating_system.py
```diff
@@ -357,7 +357,7 @@ def update_packages(
 
     def capture_system_information(self, saved_path: Path) -> None:
         # avoid to involve node, it's ok if some command doesn't exist.
-        self._node.execute("uname -vrio").save_stdout_to_file(saved_path / "uname.txt")
+        self._node.execute("uname -vrmo").save_stdout_to_file(saved_path / "uname.txt")
```
squirrelsc, line 361: Please make sure it well tested on major distros.

LiliDeng: sure, will check tomorrow.

LiliDeng: I have tested below images.

image

squirrelsc (approved):

PR #2209, opened 2022-08-15
skip nfs udp on rhel 8.x

LiliDeng:


microsoft/testsuites/performance/storageperf.py
```diff
@@ -213,6 +213,18 @@ def _perf_nfs(
         ):
             raise SkippedException(f"{server_node.os.name} not supported")
 
+        # refer
```
squirrelsc: refer for what? Please summary for below logic, why rh 8 doesn't support udp mode.

LiliDeng: update

squirrelsc (approved):

PR #2211, opened 2022-08-15
NIC: add method to wait for sriov enabled in guest

mcgov:
We previously only allow a wait for SRIOV to be ready in the guest when toggling SRIOV. 
- Add a method to node.nic to perform this wait and reload info if needed.
- add a use in DPDK initialize_node_resources

This fixes a race resulting in erroneous 'SRIOV is not available' errors at the beginning of some tests.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -11,7 +11,8 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from assertpy import assert_that
 from retry import retry
 
-from lisa.tools import Echo, Ip
+import lisa.util.constants as constants
+from lisa.tools import Echo, Ip, Lspci
 from lisa.util import InitializableMixin, LisaException, find_groups_in_lines
 
 if TYPE_CHECKING:
@@ -268,8 +269,18 @@ class Nics(InitializableMixin):
         self.nics.clear()
         self._initialize()
 
-    @retry(tries=12, delay=10)
+    @retry(tries=15, delay=3, backoff=1.15)
     def wait_for_sriov_enabled(self) -> None:
+        lspci = self._node.tools[Lspci]
+
+        # check for VFs on the guest
+        vfs = lspci.get_devices_by_type(constants.DEVICE_TYPE_SRIOV)
+        assert_that(vfs).described_as(
+            "Could not identify any SRIOV NICs on the test node."
+        ).is_not_empty()
+
+        # check if the NIC driver has finished setting up the
+        # failsafe pair, reload if not
         if not self.get_lower_nics():
             self.reload()
         if not self.get_lower_nics():
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -9,6 +9,7 @@ from pathlib import PurePosixPath
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from assertpy import assert_that
+from retry import retry
 
 from lisa.tools import Echo, Ip
 from lisa.util import InitializableMixin, LisaException, find_groups_in_lines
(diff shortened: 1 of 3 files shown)


lisa/nic.py
```diff
@@ -267,6 +268,18 @@ def reload(self) -> None:
         self.nics.clear()
         self._initialize()
 
+    @retry(tries=6, delay=20)
```
squirrelsc: Make the delay shorter and tries more. So it won't wait 20 seconds for next try.

mcgov: Changed so it checks with a backoff to retain the 120 second total wait and quick checks at the start with fewer total ssh roundtrips so we don't hog all the bandwidth while checking.

squirrelsc: 1 second won't spend noticeable bandwidth, the only problem is it produces more log. But it saves up to 9 seconds for each run. How about change to at least 5 seconds?

LiliDeng (changes requested):
from current test results, there are two main reasons
- the kernel is too old to detect and show VF
- there is no VF nic at all
Probably firstly we detect there is (are) VF listed from lspci, after that check paired nics with retry?

mcgov:
> from current test results, there are two main reasons
> 
> * the kernel is too old to detect and show VF
> * there is no VF nic at all
>   Probably firstly we detect there is (are) VF listed from lspci, after that check paired nics with retry?

Yes, from digging in I noticed there are some kernels which just take like 10 - 60 seconds longer to detect the VF.

lisa/nic.py
mcgov: Oops, I thought I had pushed an update before adding that last comment.
Delay scheme, 15 tries, 3 seconds with 1.15x backoff 
calc_backoff(15,3,1.15)
3.4499999999999997
3.9674999999999994
4.562624999999999
5.247018749999998
6.034071562499997
6.939182296874995
7.980059641406244
9.17706858761718
10.553628875759756
12.136673207123719
13.957174188192276
16.050750316421116
18.458362863884282
21.22711729346692
24.411184887486957
max total wait: 167.15241747073347

My desire to add the backoff comes from testing where I saw concurrent tests running retries on nic.reload every second cause a bottleneck on the *orchestrator*, not the guest.

mcgov: Commit is updated now, sorry for confusion.

mcgov:
> from current test results, there are two main reasons
> 
> * the kernel is too old to detect and show VF
> * there is no VF nic at all
>   Probably firstly we detect there is (are) VF listed from lspci, after that check paired nics with retry?

Added check for VF before the failover pair detection :) Thank you for the note!

mcgov:
@LiliDeng adding a kernel version check outside of this PR since it's in dpdk/common.py

lisa/nic.py
```diff
@@ -267,6 +269,28 @@ def reload(self) -> None:
         self.nics.clear()
         self._initialize()
 
+    @retry(tries=15, delay=3, backoff=1.15)
+    def wait_for_sriov_enabled(self) -> None:
+        lspci = self._node.tools[Lspci]
+
+        # check for VFs on the guest
+        vfs = lspci.get_devices_by_type(constants.DEVICE_TYPE_SRIOV)
```
LiliDeng: lspci.get_devices_by_type(constants.DEVICE_TYPE_SRIOV, force_run=True)

lisa/nic.py
```diff
@@ -267,6 +269,28 @@ def reload(self) -> None:
...
+        assert_that(vfs).described_as(
+            "Could not identify any SRIOV NICs on the test node."
+        ).is_not_empty()
```
LiliDeng: better to check VF count than not empty

lisa/nic.py
```diff
@@ -267,6 +269,28 @@ def reload(self) -> None:
...
+        # failsafe pair, reload if not
+        if not self.get_lower_nics():
+            self.reload()
+        if not self.get_lower_nics():
+            assert_that(self.get_lower_nics()).described_as(
+                "Did not detect any upper/lower sriov paired nics!: "
+                f"upper: {self.get_upper_nics()} "
+                f"lower: {self.get_lower_nics()} "
+                f"unpaired: {self.get_unpaired_devices()}"
+            ).is_not_empty()
```
LiliDeng, line 293: show we judge each synthetic has paired VF nic?

mcgov: The command is a lot of spew since we combined it into a single one to avoid roundtrips, but I'll add the list of found VF devices to this. The pairing is just from sysfs at /sys/class/net/[ethname]/lower_[lowerethname]/device

lisa/nic.py
mcgov: Excellent catch

lisa/nic.py
mcgov: Sure, swapped

lisa/nic.py
mcgov: check new code, is that good? Added a __str__ for PciDevice so we can print the list of VFs

mcgov:
@LiliDeng Do you need any additional changes?

LiliDeng (approved):

PR #2213, opened 2022-08-16
Collect kernel latest commit id and print

SRIKKANTH:


squirrelsc (dismissed):

lisa/transformers/kernel_source_installer.py
```diff
@@ -330,6 +330,11 @@ def get_source_code(self) -> PurePath:
             self._log.info(f"checkout code from: '{runbook.ref}'")
             git.checkout(ref=runbook.ref, cwd=code_path)
 
+        result = self._node.execute(
+            "git log --pretty=format:%h | head -1", cwd=code_path, shell=True
```
LiliDeng: please use git tool, if no such method, please add one, thanks. how do you plan to pass it into db?

SRIKKANTH: Added 'list_commit_ids()' to git tool. In pipeline this info is collected and pushed to logs db as tag.

lisa/tools/git.py
```diff
@@ -167,6 +167,16 @@ def list_tags(self, cwd: pathlib.PurePath) -> List[str]:
...
 
+    def list_commit_ids(self, cwd: pathlib.PurePath) -> List[str]:
+        result = self.run(
+            "--no-pager log --pretty=format:%h",
+            shell=True,
+            cwd=cwd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Failed to fetch commit ids.",
+        )
+        return filter_ansi_escape(result.stdout).splitlines()
```
LiliDeng, line 179: please add force_run=True

LiliDeng (approved):

PR #2215, opened 2022-08-17
change cpu cases to use 16 core, and fix channel count reset logic

LiliDeng:
- if no core requirement setting, the case will be skipped since no idle cpu used for testing.
- after offline the idel cpu, the max channel count changed into cpu_count - idle_cpus count, only can set current channel between 1 and current max channel, otherwise, ethtool command will fail

### DIFF
--- a/microsoft/testsuites/cpu/cpusuite.py
+++ b/microsoft/testsuites/cpu/cpusuite.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 from __future__ import annotations
 
+import random
 from typing import cast
 
 from assertpy import assert_that
@@ -18,7 +19,7 @@ from lisa import (
 )
 from lisa.environment import Environment
 from lisa.node import RemoteNode
-from lisa.tools import Ethtool, Fio, Iperf3, Kill, Lscpu, Lsvmbus
+from lisa.tools import Ethtool, Fio, Iperf3, Kill, Lscpu, Lsvmbus, Modprobe
 from lisa.util import SkippedException
 from microsoft.testsuites.cpu.common import (
     CPUState,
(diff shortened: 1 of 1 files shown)


squirrelsc:
What's the reason it needs 32 cores? Should 2 cores be enough to have an idle cpu? The comments should be in test case metadata to let others know where the core requirement is from.

LiliDeng:
> 

I tried 16 core, the case still be skipped for no idle core to be tested

squirrelsc:
> > 
> 
> I tried 16 core, the case still be skipped for no idle core to be tested

Does it need an actual idle cpu? If so, any count of cores maybe failed. How about wait a moment for an idle cpu core?

LiliDeng:
> > > 
> > 
> > 
> > I tried 16 core, the case still be skipped for no idle core to be tested
> 
> Does it need an actual idle cpu? If so, any count of cores maybe failed. How about wait a moment for an idle cpu core?

the case is to verify taking idle cpu offline/online. we can't take working cpu offline.
newer kernel supports to update workload's target cpu into 0, then take remain ones (idle cpu) offline/online. but most of current distros don't support it, we have to rely on current idle cpus in system.

I will check whether wait for a while help or not.

LiliDeng:
> > > 
> > 
> > 
> > I tried 16 core, the case still be skipped for no idle core to be tested
> 
> Does it need an actual idle cpu? If so, any count of cores maybe failed. How about wait a moment for an idle cpu core?

no lucky if add wait, I try 2 cores, 16 cores, after 1 hour, still no idle cpu, I think the kernel will try to distribute workload into different cpus. that's why we see different target cpu distributed when run lsvmbus -vvv

squirrelsc (approved):

PR #2217, opened 2022-08-17
Reduce nfs FIO file size

dsrivastavv:
Reduce file size to 256MB to speedup tests


microsoft/testsuites/performance/storageperf.py
```diff
@@ -322,7 +322,7 @@ def _perf_premium_datadisks(
             disk_type=DiskType.premiumssd,
             numjob=core_count,
             block_size=block_size,
-            size_gb=8,
+            size_mb=8192,
```
LiliDeng, line 325: will it affect premium storage perf testing？

dsrivastavv: It should not affect it. It is the same size just specified in MB now rather than GB. Btw is this value passed to database?

LiliDeng: no, db doesn't store this value

LiliDeng (approved):

PR #2218, opened 2022-08-18
verify_boot_error_warnings refactor to use dmesg filters

mcgov:
- Break up test into syslog and dmesg flavors to split results by severity
- Add Dmesg tool option to filter by log level and facility
- refactor ignore line filtering into it's own shared function

### DIFF
--- a/lisa/tools/dmesg.py
+++ b/lisa/tools/dmesg.py
@@ -2,16 +2,59 @@
 # Licensed under the MIT license.
 
 import re
-from typing import List
+from typing import List, Optional
 
+from assertpy import assert_that
 from semver import VersionInfo
 
 from lisa.executable import Tool
 from lisa.util import LisaException
 from lisa.util.process import ExecutableResult
 
+# Supported log facilities:
+#     kern - kernel messages
+#     user - random user-level messages
+#     mail - mail system
+#   daemon - system daemons
+#     auth - security/authorization messages
+#   syslog - messages generated internally by syslogd
+#      lpr - line printer subsystem
+#     news - network news subsystem
+
+# Supported log levels (priorities):
+#    emerg - system is unusable
+#    alert - action must be taken immediately
+#     crit - critical conditions
+#      err - error conditions
+#     warn - warning conditions
+#   notice - normal but significant condition
+#     info - informational
+#    debug - debug-level messages
+
 
 class Dmesg(Tool):
+
+    __supported_log_facilities = [
+        "kern",
+        "user",
+        "mail",
+        "daemon",
+        "auth",
+        "syslog",
+        "lpr",
+        "news",
+    ]
+
+    __supported_log_levels = [
+        "emerg",
+        "alert",
+        "crit",
+        "err",
+        "warn",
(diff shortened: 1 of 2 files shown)


lisa/tools/dmesg.py
```diff
@@ -33,8 +76,38 @@ def command(self) -> str:
     def _check_exists(self) -> bool:
         return True
 
-    def get_output(self, force_run: bool = False) -> str:
-        command_output = self._run(force_run=force_run)
+    def get_output(
+        self,
+        force_run: bool = False,
+        log_facility: Optional[List[str]] = None,
```
squirrelsc: Use an enum like other places, so it's easier to use. One level is enough, it should include the higher levels.

lisa/tools/dmesg.py
```diff
@@ -33,8 +76,38 @@ def command(self) -> str:
     def _check_exists(self) -> bool:
         return True
 
-    def get_output(self, force_run: bool = False) -> str:
-        command_output = self._run(force_run=force_run)
+    def get_output(
+        self,
+        force_run: bool = False,
```
squirrelsc, line 94: force_run shouldn't be first argument, because it's not most important one.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -773,11 +799,37 @@ def verify_bash_history_is_empty(self, node: Node) -> None:
...
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+    )
+    def verify_boot_error_fail_warnings_dmesg(self, node: Node) -> None:
+        dmesg = node.tools[Dmesg]
+        log_output = dmesg.get_output(
+            force_run=True, log_level=["warn", "err", "emerg", "alert", "crit"]
+        )
+        found_results = self._process_log_lines(log_output)
```
squirrelsc, line 819: Does it need to filter in the warn level messages? I remembered we hope the kernel message is clean to avoid confuse users. @LiliDeng

mcgov: I'm kind of torn, WARN should mean something is going to go wrong eventually. If we're splitting by dmesg vs syslog it makes sense to include WARN in dmesg. I think we'll have an initial burst of messages to triage and then the dmesg one will end up being more stable (until we run on new SKUS or kernels etc).

squirrelsc: I mean, when we have the warn and above levels, the listed string may not be necessary. So if there is any warn and above message, the test case should fail.

lisa/tools/dmesg.py
mcgov: So much nicer, when I was testing it wasn't consistent whether selecting a single level automatically gave you the levels above (looking at Ubuntu 22.04) but I can fix it up to do this

mcgov: Still testing though

squirrelsc:
@LiliDeng can you check if the change is worth to be merged?

PR #2223, opened 2022-08-23
Add some enhancement in kdump test case for Redhat image

lubaihua33:
1. Redhat 9 no longer supports crashkerne=auto option, the auto_size case should be skipped for this image.
2. After configuring kdump and rebooting the system, a restart of the kdump service is needed to make sure the crash kernel is loaded.

### DIFF
--- a/microsoft/testsuites/kdump/kdumpcrash.py
+++ b/microsoft/testsuites/kdump/kdumpcrash.py
@@ -220,8 +220,12 @@ class KdumpCrash(TestSuite):
         # to https://wiki.centos.org/action/show/Sources?action=show&redirect=sources
         # In addition, we didn't see upstream kernel has the auto crashkernel feature.
         # It may be a patch owned by Redhat/Centos.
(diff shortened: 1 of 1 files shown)


lisa/tools/kdump.py
```diff
@@ -310,6 +310,14 @@ def enable_kdump_service(self) -> None:
         service = self.node.tools[Service]
         service.enable_service("kdump")
 
+    def restart_kdump_service(self) -> None:
+        """
+        This method restart the kdump service. If distro has a different kdump service
+        name, need override it.
+        """
+        service = self.node.tools[Service]
+        service.restart_service("kdump")
```
LiliDeng: how about enable it by default? will it be started automatically after reboot?

lubaihua33: Most of the time, it will be started automatically after a reboot. 

After the double check, restarting kdmp service should not be added here, which will hide the kdump inactive issue after a reboot. I have deleted this commit.

LiliDeng (approved):

PR #2226, opened 2022-08-29
Increase time for fio process

adityagesh:
Test case 'verify_cpu_offline_storage_workload' checks if fio process
exists after cpu hotplug.
The test fails when the number of cores in SKU is high since
the fio process exits before the cpu hotplug operation.

This increase in time does not affect the total duration of the
test as the tests kills the process.

### DIFF
--- a/microsoft/testsuites/cpu/cpusuite.py
+++ b/microsoft/testsuites/cpu/cpusuite.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import random
+import time
 from typing import cast
 
 from assertpy import assert_that
@@ -91,13 +92,15 @@ class CPUSuite(TestSuite):
         fio_data_size_in_gb = 1
         try:
             image_folder_path = node.find_partition_with_freespace(fio_data_size_in_gb)
+            # Each CPU takes ~5 seconds to toggle offline-online
+            fio_run_time = 300 + (node.tools[Lscpu].get_core_count() * 5)
             fio_process = node.tools[Fio].launch_async(
                 name="workload",
                 filename=f"{image_folder_path}/fiodata",
                 mode="readwrite",
                 iodepth=128,
                 numjob=10,
-                time=300,
+                time=fio_run_time,
                 block_size="1M",
                 size_gb=fio_data_size_in_gb,
                 group_reporting=False,
@@ -105,8 +108,14 @@ class CPUSuite(TestSuite):
                 time_based=True,
             )
 
+            # Added to find an optional runtime for fio_run_time
+            # Remove once test is stable
+            hot_plug_start_time = time.time()
+
             # verify cpu hotplug functionality
             verify_cpu_hot_plug(log, node)
+
+            log.debug(f"CPU Hotplug duration: {time.time() - hot_plug_start_time} s")
 
             # verify that the fio was running when hotplug was triggered
             assert_that(


microsoft/testsuites/cpu/cpusuite.py
```diff
@@ -91,13 +91,15 @@ def verify_cpu_offline_storage_workload(self, log: Logger, node: Node) -> None:
         fio_data_size_in_gb = 1
         try:
             image_folder_path = node.find_partition_with_freespace(fio_data_size_in_gb)
+            # Each CPU takes ~5 seconds to toggle offline-online
+            fio_run_time = 300 + (node.tools[Lscpu].get_core_count() * 5)
```
dsrivastavv, line 96: How is this timeout determined?

adityagesh: > **2022-08-29 05:48:24.770**[5944][DEBUG] lisa.suite[CPUSuite].case[verify_cpu_offline_storage_workload] set cpu820 from online to offline.
2022-08-29 05:48:24.770[5944][DEBUG] lisa.env[generated_0].node[0].cmd[938] cmd: ['sudo', 'sh', '-c', "echo '1' > /sys/devices/system/cpu/cpu820/online"], cwd: None, shell: True, sudo: True, posix: True, remote: True
2022-08-29 05:48:26.026[5944][DEBUG] lisa.env[generated_0].node[0].cmd[938] execution time: 1.255 sec, exit code: 0
2022-08-29 05:48:26.026[5944][DEBUG] lisa.env[generated_0].node[0].cmd[6099] cmd: ['sudo', 'sh', '-c', 'cat /sys/devices/system/cpu/cpu820/online'], cwd: None, shell: True, sudo: True, posix: True, remote: True
2022-08-29 05:48:26.057[2044][DEBUG] lisa.env[generated_0].node[0].cmd[6099].stdout 1
2022-08-29 05:48:26.073[5944][DEBUG] lisa.env[generated_0].node[0].cmd[6099] execution time: 0.046 sec, exit code: 0
2022-08-29 05:48:26.073[5944][DEBUG] lisa.suite[CPUSuite].case[verify_cpu_offline_storage_workload] set cpu820 from offline to online.
2022-08-29 05:48:26.073[5944][DEBUG] lisa.env[generated_0].node[0].cmd[6498] cmd: ['sudo', 'sh', '-c', "echo '0' > /sys/devices/system/cpu/cpu821/online"], cwd: None, shell: True, sudo: True, posix: True, remote: True
2022-08-29 05:48:27.324[5944][DEBUG] lisa.env[generated_0].node[0].cmd[6498] execution time: 1.250 sec, exit code: 0
2022-08-29 05:48:27.324[5944][DEBUG] lisa.env[generated_0].node[0].cmd[9717] cmd: ['sudo', 'sh', '-c', 'cat /sys/devices/system/cpu/cpu821/online'], cwd: None, shell: True, sudo: True, posix: True, remote: True
**2022-08-29 05:48:27.355**[5088][DEBUG] lisa.env[generated_0].node[0].cmd[9717].stdout 0


Please note the timestamp in the above logs. It takes ~5 seconds for each toggling each CPU. 
If the Kernel supports changing interrupt CPU for channel, then all CPU, except 0 will be toggled. Hence the calculation `node.tools[Lscpu].get_core_count() * 5`

300 seconds was the previous time for fio process, which is maintained as such so that VMs with less number of cores does not fail.

dsrivastavv: Is is certain to be same across different CPU architectures? I understand it's a bit difficult to identify a good timing but we need to ensure that we have a process to update it in future.

adityagesh: I am not certain of the different architectures. How can we find it?
It worked for different distro's with the same SKU

dsrivastavv: Let's sync offline on this

dsrivastavv: Let's add a logging here which gives maximum toggle time across all CPU. We can look at failures later to determine an optimal value

dsrivastavv (approved):

LiliDeng (approved):

PR #2228, opened 2022-08-29
networksettings: fix the ringbuffer_settings_change test case

sharsonia:


microsoft/testsuites/network/networksettings.py
```diff
@@ -120,14 +120,24 @@ def validate_ringbuffer_settings_change(self, node: Node) -> None:
...
+            # buf_size = roundup(buf_size, PAGE_SIZE);
+            #
+            # /* Now setup the send buffer. */
+            # buf_size = device_info->send_sections * device_info->send_section_size;
+            # buf_size = round_up(buf_size, PAGE_SIZE);
             assert_that(
                 int(actual_settings.current_ring_buffer_settings["RX"]),
                 "Changing RX Ringbuffer setting didn't succeed",
-            ).is_equal_to(expected_rx)
+            ).is_between(expected_rx - 5, expected_rx + 5)
```
squirrelsc, line 136: How is the "5" set? Please provide some explanation, and I fell the percentage is better than an absolute number.

sharsonia: > How is the "5" set? Please provide some explanation, and I fell the percentage is better than an absolute number.

I agree percentage is better but after talking to Haiyang as well, it looks like we don't know the exact variation that could be but it depend son slot size and/or PAGE size. From different runs for different configs until now, I observed it doesn't vary more than 2 but to be safe I put 5.

squirrelsc (approved):

PR #2232, opened 2022-08-31
Azure: select vm sizes by quota availability.

squirrelsc:
see details in commits.


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -392,30 +392,8 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
             # get allowed vm sizes. Either it's from the runbook defined vm
             # sizes, either from subscription supported.
```
sharsonia: minor: "Either it's from the runbook defined vm size, or from..."

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -394,19 +400,50 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
...
+            if all(isinstance(x, schema.NodeSpace) for x in caps):
+                cap_calculator: Dict[str, Tuple[int, int]] = {}
+                for index, cap in enumerate(caps):
+                    assert isinstance(cap, schema.NodeSpace), f"actual: {type(cap)}"
+                    azure_runbook = cap.get_extended_runbook(AzureNodeSchema, AZURE)
+                    vm_size = azure_runbook.vm_size
+                    if vm_size not in cap_calculator:
+                        cap_calculator[vm_size] = self._get_usage(location, vm_size)
+                    remaining, total = cap_calculator[vm_size]
+                    remaining += 1
```
sharsonia: Didnt understand this part. Why we are incrementing "remaining and checking if remaining is > total". Can't we just check if "remaining" meets the requirement or not. I think i am missing something here

squirrelsc: The logic counts remaining by vm_size, and make sure the whole env can be deployed, without quota issue. The "total" is the hard limitation of Azure. I should change the name from "total" to "limits".

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -442,18 +413,20 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
                 )
                 break
 
-        if not is_success and all_wait_for_resource:
-            # TODO raise for wait more resource
-            ...
+        if not is_success:
+            if any_wait_for_resource:
+                raise ResourceAwaitableException(
+                    "No available quota, try to deploy later."
```
dsrivastavv, line 420: As a result of this exception are we going to retry after some timeout or is this going to fail/skip with this error?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2196,6 +2164,179 @@ def _get_matched_capability(
...
+        usages_map: Dict[str, Any] = {value.name.value: value for value in usages}
+
+        # This method signature is used to generate cache. If pass in the log
+        # object, it makes the cache doesn't work. So create a logger in the
+        # method.
+        log = get_logger("azure")
+        location_info = self.get_location_info(location=location, log=log)
+        capabilities = location_info.capabilities
+
+        for vm_size, capability in capabilities.items():
```
dsrivastavv, line 2310: Can you please add comments explaining this logic?

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: In this PR, it will check the quota availability in prepare env phase. It has more chance to find another VM size, which supports the test requirement. If it doesn't find any, raise this exception out. In next task, the runner level will handle this exception, retry it later.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Understood, good catch! I wrote the wrong logic here. I should use it like remaining to minus.

PR #2233, opened 2022-09-01
change logic - expected half of cpus' interrupts increased

LiliDeng:


microsoft/testsuites/network/sriov.py
```diff
@@ -709,13 +713,18 @@ def verify_sriov_interrupts_change(self, environment: Environment) -> None:
...
+            unused_cpu = 0
             for cpu, init_interrupts_value in initial_pci_interrupts_by_cpus.items():
                 final_interrupts_value = final_pci_interrupts_by_cpus[cpu]
-                # 7. Compare interrupts changes, expected to see interrupts increased.
-                assert_that(final_interrupts_value).described_as(
-                    f"CPU {cpu} didn't have an increased interrupts count"
-                    " after iperf3 run!"
-                ).is_greater_than(init_interrupts_value)
+                # 7. Collect cpus which don't have interrupts count increased.
+                if final_interrupts_value <= init_interrupts_value:
```
dsrivastavv: This should be equal, right? Based on this logic it seems like there are scenarios where `final_interrupts_value` can be less than starting interrupt count

squirrelsc (dismissed):

microsoft/testsuites/network/sriov.py
LiliDeng: update

dsrivastavv (approved):

PR #2234, opened 2022-09-01
skip test if hv_network_direct module is not enabled

LiliDeng:


microsoft/testsuites/hpc/infinibandsuit.py
```diff
@@ -90,6 +90,11 @@ def verify_hpc_over_sriov(self, log: Logger, node: Node) -> None:
         requirement=simple_requirement(supported_features=[Infiniband]),
     )
     def verify_hpc_over_nd(self, log: Logger, node: Node) -> None:
+        if not node.tools[KernelConfig].is_enabled("CONFIG_HYPERV_INFINIBAND_ND"):
+            raise SkippedException(
+                "hv_network_direct module is not enabled in current distro "
```
squirrelsc: Add reason about why it's skipped, not failed.

squirrelsc: And wrap from Unsupported OS like you did before.

squirrelsc (approved):

PR #2235, opened 2022-09-01
Check console enabled log in /log/messages

adityagesh:
console enabled log may get flushed from dmesg.
Hence check for pattern in /var/log/messages.

Change assert statement for better error readability

### DIFF
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -732,7 +732,8 @@ class AzureImageStandard(TestSuite):
          in dmesg.
 
         Steps:
-        1. Get the kernel command line from dmesg output.
+        1. Get the kernel command line from /var/log/messages or
+            /var/log/syslog output.
         2. Check expected setting from kernel command line.
             2.1. Expected to see 'console=ttyAMA0' for aarch64.
             2.2. Expected to see 'console=ttyS0' for x86_64.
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -745,16 +745,27 @@ def verify_serial_console_is_enabled(self, node: Node) -> None:
             CpuArchitecture.X64: "ttyS0",
             CpuArchitecture.ARM64: "ttyAMA0",
         }
+        messages_log_file = "/var/log/messages"
```
LiliDeng: for centos/rhel, it is /var/log/syslog

adityagesh: updated

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -745,16 +745,30 @@ def verify_serial_console_is_enabled(self, node: Node) -> None:
...
         arch = lscpu.get_architecture()
         current_console_device = console_device[CpuArchitecture(arch)]
-        console_enabled_string = f"console [{current_console_device}] enabled"
-        dmesg = node.tools[Dmesg]
+        console_enabled_pattern = re.compile(
+            rf"^(.*console \[{current_console_device}\] enabled.*)$", re.M
+        )
+
+        cat = node.tools[Cat]
+        if node.shell.exists(node.get_pure_path("/var/log/messages")):
```
LiliDeng, line 758: shell.exists can't check file which need sudo permission.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -732,7 +732,7 @@ def verify_repository_installed(self, node: Node) -> None:  # noqa: C901
          in dmesg.
 
         Steps:
-        1. Get the kernel command line from dmesg output.
+        1. Get the kernel command line from /var/log/messages output.
```
LiliDeng: also need update this

adityagesh: updated

PR #2236, opened 2022-09-01
 fix CreateServicePrincipal.ps1 to fit different versions of Az Powershell module

lesscodingmorehappiness:
Users could use old or new versions of Az Powershell module


squirrelsc (approved):

squirrelsc:
There is conflict, and cannot be merged. Please change on the latest code.

PR #2237, opened 2022-09-02
Add case verify_vdso.

LiliDeng:


microsoft/testsuites/core/timesync.py
```diff
@@ -351,3 +352,33 @@ def timesync_chrony(self, node: Node) -> None:
...
+            2. Check output of 1, should not show the time related syscalls,
+             e.g. clock_gettime, gettimeofday
+        """,
+        priority=2,
+    )
+    def verify_vdso(self, node: Node) -> None:
+        time_pattern = re.compile("(time|gettimeofday|clock_gettime)", re.M)
+        strace = node.tools[Strace]
+        # 1. Run strace date 2>&1
+        output = strace.get("date 2>&1")
```
squirrelsc: How about other vdso commands? We should cover as much as possible.

microsoft/testsuites/core/timesync.py
```diff
@@ -351,3 +352,33 @@ def timesync_chrony(self, node: Node) -> None:
...
+        leading to performance bottleneck, Linux systems have a mechanism called vdso
+        which helps in gettimeoftheday and clock_gettime to be executed in userspace (no
+        syscall).
+
+        Steps:
+            1. Run strace date 2>&1
+            2. Check output of 1, should not show the time related syscalls,
+             e.g. clock_gettime, gettimeofday
+        """,
+        priority=2,
```
squirrelsc: It should be P1, if it doesn't spend more than 1 minute.

lisa/tools/strace.py
```diff
@@ -0,0 +1,28 @@
...
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+
+
+class Strace(Tool):
+    @property
+    def command(self) -> str:
+        return "strace"
+
+    def _check_exists(self) -> bool:
```
dsrivastavv: Do we need to override this? The default in parent class should handle this case

microsoft/testsuites/core/timesync.py
```diff
@@ -351,3 +352,33 @@ def timesync_chrony(self, node: Node) -> None:
         chrony.check_sources_and_stats()
         # 5. Check chrony tracking.
         chrony.check_tracking()
+
+    @TestCaseMetadata(
+        description="""
+        This test is to check gettime calls are not being redirected as system calls,
+        leading to performance bottleneck, Linux systems have a mechanism called vdso
```
dsrivastavv: [nit] bottleneck.

microsoft/testsuites/core/timesync.py
```diff
@@ -351,3 +352,33 @@ def timesync_chrony(self, node: Node) -> None:
...
+        which helps in gettimeoftheday and clock_gettime to be executed in userspace (no
+        syscall).
+
+        Steps:
+            1. Run strace date 2>&1
+            2. Check output of 1, should not show the time related syscalls,
+             e.g. clock_gettime, gettimeofday
```
dsrivastavv: Is the following exhaustive: time, gettimeofday, clock_gettime ? If so we can mention these directly here

microsoft/testsuites/core/timesync.py
```diff
@@ -351,3 +352,33 @@ def timesync_chrony(self, node: Node) -> None:
...
+        """,
+        priority=2,
+    )
+    def verify_vdso(self, node: Node) -> None:
+        time_pattern = re.compile("(time|gettimeofday|clock_gettime)", re.M)
```
dsrivastavv: [nit] can be a global variable in the class

microsoft/testsuites/core/timesync.py
LiliDeng: - https://github.com/nlynch-mentor/vdsotest
- https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/tree/tools/testing/selftests/vDSO

I found 2 repos which can test vdso. After trying, found the second one can't repro the bug. The first one is easy to compile and can repro the issue, also it contains other vdso commands, do you think it is feasible to use it?

squirrelsc: ok, we can use the first one. And add your explanation in suite description to explain why we don't use the kernel tests.

LiliDeng: update

microsoft/testsuites/core/timesync.py
LiliDeng: update

lisa/tools/strace.py
LiliDeng: remove

microsoft/testsuites/core/timesync.py
LiliDeng: update

microsoft/testsuites/core/timesync.py
LiliDeng: now change to run vdso benchmark, this code was removed

microsoft/testsuites/core/timesync.py
LiliDeng: now change to run vdso benchmark, this code was removed

squirrelsc (approved):

PR #2238, opened 2022-09-05
add case verify_dns_name_resolution

LiliDeng:


microsoft/testsuites/core/dns.py
```diff
@@ -0,0 +1,53 @@
...
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class Dns(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test case check DNS name resolution by ping bing.com.
+        """,
+        priority=1,
+    )
+    def verify_dns_name_resolution(self, node: Node) -> None:
```
squirrelsc, line 32: The test case behavior and name are inconsistent. 
1. The DNS can be tested without any other action, it's worth to have a case. 
2. The unattented-upgrade can be another test case. After upgraded, we can test the dns, reboot and other P0 scenarios.

LiliDeng: update

microsoft/testsuites/core/dns.py
```diff
@@ -0,0 +1,64 @@
...
+    This test suite covers DNS name resolution functionality.
+    """,
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class Dns(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test case check DNS name resolution by ping bing.com.
+        """,
+        priority=0,
```
squirrelsc: Make it P1

lisa/tools/ping.py
```diff
@@ -72,6 +76,10 @@ def ping(
                 sudo=True,
             ).wait_result()
         if not ignore_error:
+            if self._dns_name_resolution_issue.findall(result.stdout):
```
squirrelsc: It doesn't need to find the special error message, any failure of ping should be reported. Just change the L83 to return the stdout, if it's not too many lines.

microsoft/testsuites/core/dns.py
LiliDeng: update

lisa/tools/ping.py
LiliDeng: update

squirrelsc (approved):

PR #2240, opened 2022-09-06
Added cvm testcases, dmesg get_isolation_config()

Jose-Teuttli:
Added test that checks devices on guest
Added test that checks isolation config on guest
Added get_isolation_config function to dmesg tool

I am unsure if I am using the correct requirements, I currently only have deployed requirement for both test cases. I use dmesg and lsvmbus, but since these are tools LISA should ensure they are available on guest.

added testsuite under acc/ due to it being related to confidential computing but can be moved if asked.

Please let me know if anything needs to be changed, I tried to copy formatting of surrounding code.


lisa/tools/dmesg.py
```diff
@@ -81,6 +87,27 @@ def get_vmbus_version(self) -> VersionInfo:
                 return VersionInfo(int(major), int(minor))
         raise LisaException("No find matched vmbus version in dmesg")
 
+    def get_isolation_config(self) -> dict[str, str]:
```
squirrelsc: It's CVM only logic, not used by other test cases. So it doesn't need to add on dmesg level.

microsoft/testsuites/acc/cvm.py
```diff
@@ -0,0 +1,74 @@
...
+        This case verifies that lsvmbus only shows devices
+        that are allowed in a CVM guest
+
+        Steps:
+        1. Call lsvmbus
+        2. Iterate through list returned by lsvmbus to ensure all devices
+           listed are included in valid_class_ids
+        """,
+        priority=1,
+        requirement=simple_requirement(
```
squirrelsc, line 40: @LiliDeng I just merged Kameron's change about CVM. Please let Jose know how to test with feature enabled.

microsoft/testsuites/acc/cvm.py
```diff
@@ -0,0 +1,74 @@
...
+            environment_status=EnvironmentStatus.Deployed,
```
squirrelsc: This is not necessary, it's default value.

microsoft/testsuites/acc/cvm.py
```diff
@@ -0,0 +1,74 @@
...
+        description="""
+        This case verifies that lsvmbus only shows devices
+        that are allowed in a CVM guest
+
+        Steps:
+        1. Call lsvmbus
+        2. Iterate through list returned by lsvmbus to ensure all devices
+           listed are included in valid_class_ids
+        """,
+        priority=1,
```
squirrelsc, line 39: Should it be P1 test cases? What's the impact if the test case is failed? Please work with @LiliDeng to evaluate right priority.

LiliDeng: this case also applies to non-cvm, @Jose-Teuttli by any chance, you see any issues related with this case?

lisa/tools/dmesg.py
Jose-Teuttli: Alright will move this logic within cvm.py

microsoft/testsuites/acc/cvm.py
Jose-Teuttli: Alright will remove thank you

microsoft/testsuites/acc/cvm.py
Jose-Teuttli: I don't follow how this applies to non-cvm. All guests get offered devices, but CVM guest should reject devices not allowed in isolation. From what I understand, it is on the guest to reject these devices because the host will offer both valid and invalid devices. Which is seen in a line(s) such as the following:
[    0.563976] ubuntu kernel: hv_vmbus: Invalid offer 2 from the host supporting isolation
The devices allowed in isolation is taken from .allowed_in_isolation=true in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/hv/channel_mgmt.c

With regards to the priority, I am not confident. If the config test fails, then the guest is not properly configured as a CVM. If the devices test fails, then the guest has accepted an invalid device. However, I don't think there is anything stopping the CVM from working if it accepts an invalid device or has improper configuration, if it first passed the smoke test. However, if this were used in validating CVM SKU then it should be investigated.

Jose-Teuttli:
Once everything is working, I will squash all commits into just one for tree cleanliness. I still need to test with Kameron's CVM changes

microsoft/testsuites/acc/cvm.py
Jose-Teuttli: I was able to test with Kameron's changes successfully, the only caveat is that I needed to specify image, and sku with:

-v "marketplace_image:canonical 0001-com-ubuntu-confidential-vm-focal 20_04-lts-cvm latest" 
-v "vm_size:Standard_DC16ads_v5"

lisa/tools/dmesg.py
Jose-Teuttli: Moved logic within cvm.py

microsoft/testsuites/acc/cvm.py
Jose-Teuttli: removed

microsoft/testsuites/acc/cvm.py
squirrelsc: great

Jose-Teuttli:
Tested with Kameron's changes successfully. Only items left are 
1. Priority
2. @LiliDeng 's question regarding the case applying to non-cvm

1) If these tests fail on CVM, then this means something has been incorrectly configured. If the test has ran than it means the CVM was able to deploy despite it being mal configured, but it may not be providing the security a customer is expecting. I believe this falls under "an important feature doesn’t work with default settings; or the system can be used with limited functionality.", but I am not confident and willing to change if you think a different priority is better.

2) CVM guests have a list of allowed devices which is determined by .allowed_in_isolation=true in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/hv/channel_mgmt.c
This is tested because AFAIK the host offers the devices regardless of if the device is allowed or not on the guest. It is up to the guest to either accept/reject the valid/invalid devices.

microsoft/testsuites/acc/cvm.py
```diff
@@ -0,0 +1,85 @@
...
+        requirement=simple_requirement(
+            supported_features=[CvmEnabled()],
+        ),
+    )
+    def verify_isolation_config(self, log: Logger, node: Node) -> None:
+        dmesg_tool = node.tools[Dmesg]
+        dmesg_output = dmesg_tool.get_output()
+        isolation_config = re.search(self.__isolation_config_pattern, dmesg_output)
+        if isolation_config is not None:
+            config_a = isolation_config.group("config_a")
```
LiliDeng: we'd better to add the reason for config values need to be 0x1 and 0xba2 or kernel source code. if not are these values, what is the impact.

LiliDeng (dismissed):

microsoft/testsuites/acc/cvm.py
Jose-Teuttli: I agree, considering that support for other cvm types is being worked on, I expanded the testcase to
1) be easier to understand what values are being checked, and why they need to be such
2) make it easier to extend the testcase when other cvm types are added

Let me know if you think additional explanation is needed.

Jose-Teuttli: isolation_config_b struct is documented in kernel code @ https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/asm-generic/mshyperv.h#n37

PR #2241, opened 2022-09-07
IB: Support ConnectX-3 MLNX cards

kamalca:
Older cards such as ConnectX-3 require legacy LTS OFED drivers.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -10,7 +10,7 @@ from assertpy import assert_that
 from lisa.base_tools import Cat, Sed, Uname, Wget
 from lisa.feature import Feature
 from lisa.operating_system import CentOs, Redhat, Ubuntu
-from lisa.tools import Firewall, Ls, Make
+from lisa.tools import Firewall, Ls, Lspci, Make
 from lisa.tools.service import Service
 from lisa.tools.tar import Tar
 from lisa.util import (
@@ -74,6 +74,11 @@ class Infiniband(Feature):
     # example SKU: Standard_H16mr
     def is_over_nd(self) -> bool:
         raise NotImplementedError
+
+    def _is_legacy_device(self) -> bool:
+        lspci = self._node.tools[Lspci]
+        device_list = lspci.get_devices()
+        return any("ConnectX-3" in device.device_info for device in device_list)
 
     def get_ib_interfaces(self) -> List[IBDevice]:
         """Gets the list of Infiniband devices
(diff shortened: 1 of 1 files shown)


kamalca:
This change will still have errors for certain hardware / OS combinations. The LTS drivers for ConnectX-3 cards do not support newer OS distros.

lisa/features/infiniband.py
```diff
@@ -146,6 +149,8 @@ def _get_ib_device_names(self) -> List[str]:
     def _get_mofed_version(self) -> str:
         node = self._node
         default = "5.4-3.0.3.0"
+        if self._is_legacy_device():
+            return "4.9-5.1.0.0"
```
squirrelsc, line 155: Is it possible to parse out the version directly?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -563,6 +563,11 @@ def is_over_nd(self) -> bool:
         dmesg = self._node.tools[Dmesg]
         return "hvnd_try_bind_nic" in dmesg.get_output()
 
+    def _is_legacy_device(self) -> bool:
+        lspci = self._node.tools[Lspci]
```
squirrelsc: Why is this method implemented in Azure? Shouldn't be it independent with cloud?

squirrelsc:
> This change will still have errors for certain hardware / OS combinations. The LTS drivers for ConnectX-3 cards do not support newer OS distros.

What can we do on this?

lisa/sut_orchestrator/azure/features.py
kamalca: I wasn't sure whether this is cloud independent. Other clouds may use other legacy cards such as ConnectX-IB which I don't think are used in Azure. But I guess the same logic could still be used across all clouds. I can move this logic to infiniband.py.

lisa/features/infiniband.py
kamalca: This is not parsing the version on the VM, this is making a decision for which OFED drivers to install.

Most stable is 5.4 (LTS), old cards require 4.9 LTS, newest distros may require newer versions

kamalca:
> > This change will still have errors for certain hardware / OS combinations. The LTS drivers for ConnectX-3 cards do not support newer OS distros.
> 
> What can we do on this?

We can't control which distros NVIDIA supports with the 4.9 LTS drivers. But I could add some logic to raise an UnsupportedDistro / Skipped exception for these cases.

lisa/features/infiniband.py
```diff
@@ -305,15 +314,19 @@ def setup_rdma(self) -> None:
...
-                f"--skip-repo --skip-unsupported-devices-check --without-fw-update"
+                f"--skip-repo --without-fw-update"
             )
 
+        if not self._is_legacy_device():
+            extra_params += " --skip-unsupported-devices-check"
+
         node.execute(
-            f"./{mofed_folder}/mlnxofedinstall --add-kernel-support {extra_params}",
+            f"./{mofed_folder}/mlnxofedinstall --add-kernel-support " f"{extra_params}",
```
squirrelsc: This line doesn't need to be changed.

squirrelsc (approved):

PR #2242, opened 2022-09-07
Fix logic to connect to serial console websocket

dsrivastavv:
WIP

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -364,9 +364,11 @@ class SerialConsole(AzureFeatureMixin, features.SerialConsole):
             parent_resource=self._vm_name,
             serial_port=self._serial_port.name,
         )
-        token = platform.credential.get_token("").token
+        access_token = platform.credential.get_token(
+            "https://management.core.windows.net/.default"
+        ).token
         serial_port_connection_str = (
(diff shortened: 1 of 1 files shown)


lisa/util/logger.py
```diff
@@ -103,6 +103,7 @@ def __init__(self, logger: Logger, level: int):
         self._level = level
         self._log = logger
         self._buffer: str = ""
+        self.isatty = lambda: False
```
squirrelsc: Should it just return the value from underlying Logger stream?

pyproject.toml
```diff
@@ -20,6 +20,7 @@ azure-mgmt-marketplaceordering = {version = "^1.1.0", optional = true}
 azure-mgmt-network = {version = "^19.3.0", optional = true}
 azure-mgmt-resource = {version = "^21.0.0", optional = true}
 azure-mgmt-serialconsole = {version = "^1.0.0", optional = true}
+azure-cli-core = {version = "^2.40.0", optional = true}
```
squirrelsc: Add an internal PR to update packages to prevent broken, when this PR is merged.

squirrelsc (approved):

lisa/util/logger.py
dsrivastavv: This is not applicable anymore

PR #2244, opened 2022-09-09
Support merge feature requirement from runbook

squirrelsc:
If the type in runbook is complex like a set space, the original implementation doesn't handle feature req merging well. 

See commits for details.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -1414,11 +1414,11 @@ class CentOs(Redhat):
     def _initialize_package_installation(self) -> None:
         information = self._get_information()
         if 8 == information.version.major:
-            # refer https://www.centos.org/centos-linux-eol/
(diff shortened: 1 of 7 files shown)


lisa/operating_system.py
```diff
@@ -1417,7 +1417,7 @@ def _initialize_package_installation(self) -> None:
             # refer https://www.centos.org/centos-linux-eol/
             # CentOS 8 is EOL, old repo mirror was moved to vault.centos.org
             # CentOS-AppStream.repo, CentOS-Base.repo may contain non-existed repo
-            # use skip_if_unavailable to aviod installation issues bring in by above
+            # use skip_if_unavailable to avoid installation issues bring in by above
```
sharsonia: [minor] - If we are here correcting the spelling then might as well change this - 
"issue brought in by above"

PR #2245, opened 2022-09-11
fix install ofed driver on rhel8 and 9

LiliDeng:


lisa/features/infiniband.py
```diff
@@ -153,7 +153,9 @@ def _get_mofed_version(self) -> str:
         default = "5.4-3.0.3.0"
         if self._is_legacy_device():
             return "4.9-5.1.0.0"
-        if isinstance(node.os, Ubuntu) and str(node.os.information.release) == "20.04":
+        if (
+            isinstance(node.os, Ubuntu) and str(node.os.information.release) == "20.04"
```
squirrelsc: 1. the version should be compaired like `node.os.information.release == "20.4"`
2. To compare with higher version Ubuntu, how about use `>=`?

squirrelsc (approved):

PR #2249, opened 2022-09-13
Fix regression on merging features

squirrelsc:
When features are merged, the check is called to make sure the overlap of intersect is right scope on each req. But the feature is special, it needs some hacks on current infra to merge correctly.

With this fix,
1. The features will be concatenated into platform req in lisa runner.
2. The platform features is taken in node space intersect.
3. The features are actually merged in the azure platform.

### DIFF
--- a/lisa/runners/lisa_runner.py
+++ b/lisa/runners/lisa_runner.py
@@ -687,6 +687,35 @@ class LisaRunner(BaseRunner):
                             schema.NodeSpace, node_requirement_data
                         )
 
+                        # Manage the union of the platform requirements and the node
+                        # requirements before taking the intersection of
+                        # the rest of the requirements.
+                        platform_requirement.features = search_space.SetSpace(
+                            True,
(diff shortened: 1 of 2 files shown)


lisa/runners/lisa_runner.py
```diff
@@ -687,6 +687,35 @@ def _merge_test_requirements(
...
+                                else []
+                            )
+                            + (
+                                original_node_requirement.features.items
+                                if original_node_requirement.features
+                                else []
+                            ),
+                        )
+                        platform_requirement.excluded_features = search_space.SetSpace(
+                            False,
```
sharsonia, line 707: why False here? if I am reading it right, if there any excluded features then we would set it as true. isn't it?

squirrelsc: Sorry, just found the comments after merged.  The excluded features will be converted to True when get min capabilities.

PR #2250, opened 2022-09-14
networksettings: Distinguish the failure for AN enabled and not enabl…

sharsonia:
Distinguish the failure for AN enabled and not enabled in device statistics test.

When AN is enabled and the device_statistics test fails then the mlx driver code would need to be investigated. But for non-enabled cases, netvsc driver is at play.

### DIFF
--- a/microsoft/testsuites/network/networksettings.py
+++ b/microsoft/testsuites/network/networksettings.py
@@ -577,9 +577,11 @@ class NetworkSettings(TestSuite):
             per_tx_queue_packets: List[int] = []
             per_rx_queue_packets: List[int] = []
             nic = client_node.nics.get_nic(device_stats.interface)
+            an_enabled = "AN Disabled "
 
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/network/networksettings.py
```diff
@@ -577,9 +577,11 @@ def validate_device_statistics(self, environment: Environment, log: Logger) -> N
             per_tx_queue_packets: List[int] = []
             per_rx_queue_packets: List[int] = []
             nic = client_node.nics.get_nic(device_stats.interface)
+            an_enabled = ""
 
             # If AN is enabled on this interface then check the vf stats.
             if nic.lower:
+                an_enabled += "AN Enabled"
```
pupacha: _nit:_ Probably a space at the end would make output better readable.

LiliDeng (approved):

PR #2255, opened 2022-09-16
kata conformance automation

smit-gardhariya:
This is the development to automate the kata conformance test with Azure AKS cluster.

### DIFF
--- a/lisa/tools/lscpu.py
+++ b/lisa/tools/lscpu.py
@@ -55,7 +55,7 @@ class CPUInfo:
 class Lscpu(Tool):
     # CPU(s):              16
     # Total CPU(s):            2
-    __vcpu = re.compile(r"CPU\(s\):[ ]+([\d]+)\r?$", re.M)
+    __vcpu = re.compile(r"^(CPU|Total CPU)\(s\):[ ]+([\d]+)\r?$", re.M)
     # Thread(s) per core:  1
     #      Thread(s) per core:  1
     __thread_per_core = re.compile(r"^[ ]*Thread\(s\) per core:[ ]+([\d]+)\r?$", re.M)
@@ -129,7 +129,7 @@ class Lscpu(Tool):
             len(matched),
             f"cpu count should have exact one line, but got {matched}",
         ).is_equal_to(1)
-        self._core_count = int(matched[0])
+        self._core_count = int(matched[0][1])
 
         return self._core_count
 


anirudhrb:
Please run `make check` and fix the issues. `make fix` will fix all the auto-fixable ones.

pyproject.toml
```diff
@@ -54,6 +54,8 @@ sphinxemoji = {version="^0.1.8", optional=true}
 sphinx-copybutton = {version="^0.4.0", optional=true}
 rstcheck = {version="^3.3.1", optional=true}
 types-Pillow = {version="^8.3.3", optional=true}
+azure-mgmt-containerservice = "^20.3.0"
```
squirrelsc: Azure package should be optional as other azure packages.

pyproject.toml
```diff
@@ -54,6 +54,8 @@ sphinxemoji = {version="^0.1.8", optional=true}
...
+kubernetes = "^24.2.0"
```
squirrelsc: Because it's used by AKS test cases only, it should follow other azure package to be optional.

lisa/base_tools/wget.py
```diff
@@ -49,13 +49,13 @@ def get(
         # So that here can use the corresponding path format.
         if file_path:
             # create folder when it doesn't exist
-            self.node.shell.mkdir(PurePosixPath(file_path), exist_ok=True)
+            self.node.shell.mkdir(Path(file_path), exist_ok=True)
```
squirrelsc: Why makes this change? It maybe a remote path, and supports only posix, so it's PurePosixPath.

lisa/tools/go.py
```diff
@@ -0,0 +1,103 @@
...
+    def command(self) -> str:
+        return "go"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Wget, Tar, Rm]
```
squirrelsc: rm is the built-in tool, it doesn't need to be included here.

lisa/tools/go.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Suse):
```
squirrelsc: Means all Linux and bsd, the suse is not necessary, because the logic is the same.

microsoft/testsuites/kata_conformance/aks_infra.py
```diff
@@ -0,0 +1,135 @@
...
+        tenant_id: str,
+        client_id: str,
+        client_secret: str,
+    ):
+        self.log = log
+        self._credential_setup(subscription_id, tenant_id, client_id, client_secret)
+        self.kube_path = ""
+
+    def _credential_setup(self, subscription_id, tenant_id, client_id, client_secret):
+        if tenant_id:
```
squirrelsc: The Azure tools should be in Azure folder, and get credentials from Azure platform. Avoid to put it in test code to reduce dup logics.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
```diff
@@ -0,0 +1,117 @@
...
+            client_id=client_id,
+            client_secret=client_secret,
+            tenant_id=tenant_id,
+        )
+        aks.create_aks_infra(
+            kubernetes_version, worker_vm_size, node_count, azure_region, headers
+        )
+        log.info("Completed AKS Infra Creation")
+
+        node.tools[KataConformanceTests].run_tests(result, environment, log_path)
```
squirrelsc: How is the node related to the aks cluster?

microsoft/testsuites/kata_conformance/aks_infra.py
pupacha: @squirrelsc  Taking the credentials separately as case variables allows the testcase to decouple from Azure platform and can be run on bare metal (ready platform) in future.

smit-gardhariya: Also, conformance tests can be launched from anywhere but it will always be running on kubernetes cluster (AKS). As Pavan said it can be launched with ready platform as well.

I closed the PR mistakenly while syncing the fork. I am also pushing some new changes with new PR.

Thnx @squirrelsc for the review

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
smit-gardhariya: This Node will launch the kata conformance test case on AKS cluster. While creating AKS cluster, we are creating Kubernetes config file at ~/.kube/config. 

After building Kubernetes, we will use binary e2e.test to launch conformance test on AKS using above Kubernetes config file.

lisa/base_tools/wget.py
smit-gardhariya: This wont be the remote path. We are doing wget on node to get the source for installation of goLang having specific version.

I changed this as assert failure was happening at shell.py on line 666 as it expect Path instance and was getting PurePosixPath.

Path and PurePosixPath are child classes of PurePath.

smit-gardhariya: I will try to run again without this change by reverting it and will make sure.

PR #2258, opened 2022-09-17
fix tvm test case issues

LiliDeng:


lisa/operating_system.py
```diff
@@ -762,6 +762,8 @@ def add_repository(
             expected_exit_code=0,
             expected_exit_code_failure_message="fail to add repository",
         )
+        if not isinstance(self._node.os, Ubuntu):
```
squirrelsc: Is it to fix for debian only? If so,
1. Add comments to explain it.
2. Check it like `if type(self._node.os) == Debian)`.

squirrelsc (approved):

PR #2261, opened 2022-09-20
fix regression for lscpu

LiliDeng:


pupacha (approved):

lisa/tools/lscpu.py
```diff
@@ -55,7 +55,7 @@ def __repr__(self) -> str:
 class Lscpu(Tool):
     # CPU(s):              16
     # Total CPU(s):            2
```
squirrelsc, line 57: It's better to add negative cases and clearly explain it.

LiliDeng: addressed in #2264

PR #2262, opened 2022-09-20
Get vmbus version from messages or syslog in kdump test suite

lubaihua33:
For some distros, such as Mariner, there might be many audit messages in dmesg. The vmbus version information might get flushed. Hence check for pattern in /var/log/messages or /var/log/syslog.

### DIFF
--- a/lisa/tools/dmesg.py
+++ b/lisa/tools/dmesg.py
@@ -2,11 +2,13 @@
 # Licensed under the MIT license.
 
 import re
+from pathlib import PurePosixPath
 from typing import List
 
 from semver import VersionInfo
 
 from lisa.executable import Tool
+from lisa.tools import Cat
 from lisa.util import LisaException
 from lisa.util.process import ExecutableResult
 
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -203,10 +206,31 @@ def kdumpcrash_validate_large_memory_auto_size(
         self.crash_kernel = "auto"
         self._kdump_test(node, log_path, log)
 
+    def _get_vmbus_version(self, node: Node) -> VersionInfo:
+        vmbus_version_pattern = re.compile(
```
squirrelsc: Check with @LiliDeng what's the best way to get vmbus version with existing tools.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -22,7 +25,7 @@
 from lisa.features import SerialConsole
 from lisa.operating_system import Redhat
 from lisa.sut_orchestrator.azure.tools import Waagent
-from lisa.tools import Dmesg, Echo, KdumpBase, KernelConfig, Lscpu, Stat
+from lisa.tools import Cat, Echo, KdumpBase, KernelConfig, Lscpu, Stat
```
squirrelsc: Why not use dmesg?

squirrelsc: The reason should be added in the commit message.

lubaihua33: Added.

microsoft/testsuites/kdump/kdumpcrash.py
LiliDeng: Baihua is trying to fix when dmesg is full and flushed by many noisy message, at that time, it can't get vmbus version from dmesg.

LiliDeng: @lubaihua33 we can put the logic under dmesg tool?

lubaihua33: Sure, have done.

PR #2263, opened 2022-09-21
Kata Conformance testcase integration with LISA

smit-gardhariya:


microsoft/testsuites/cloud_hypervisor/ch_tests.py
```diff
@@ -96,6 +96,31 @@ def verify_cloud_hypervisor_live_migration_tests(
             result, environment, "integration-live-migration", hypervisor
         )
 
+    @TestCaseMetadata(
```
squirrelsc: The changes of cloud hypervisor test cases should be in another PR, so it's easier to review and clearer history.

microsoft/testsuites/cloud_hypervisor/ch_tests.py
```diff
@@ -96,6 +96,31 @@ def verify_cloud_hypervisor_live_migration_tests(
...
+        description="""
+            Runs cloud-hypervisor live migration tests.
+        """,
+        priority=3,
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                core_count=search_space.IntRange(min=16),
```
squirrelsc: 1. Why specify those requirements? The requirements should be minimized to maximize tested scope. If there are special reasons on cpu and memory, add it in description part.

lisa/tools/go.py
```diff
@@ -0,0 +1,103 @@
...
+from lisa.tools.rm import Rm
+from lisa.tools.tar import Tar
+from lisa.util import LisaException
+
+if TYPE_CHECKING:
+    from lisa.node import Node
+
+
+class Go(Tool):
+    def __init__(self, node: "Node") -> None:
```
squirrelsc: Not to override `__init__`, if there is no special requirement. Use `_initialize` to initialize variables.

lisa/tools/go.py
```diff
@@ -0,0 +1,103 @@
...
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Wget, Tar, Rm]
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Suse):
```
squirrelsc: again, The logic of Suse is the same as posix, so it doesn't need separate the logic out. Even the whole logic is not needed. Let Windows fails as is, because it's not widely supported in LISA.

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
```diff
@@ -0,0 +1,271 @@
...
+    def _install_dep(self) -> None:
+        git = self.node.tools[Git]
+        self._log.info("Cloning Repo : " + self.kubernetes_repo)
+        git.clone(self.kubernetes_repo, self.repo_root, timeout=3600)
+        self._log.info("Cloning Repo : " + self.kata_container_tests_repo)
+        git.clone(self.kata_container_tests_repo, self.repo_root, timeout=3600)
+
+        self._log.info("Installing GoLang")
+        go = self.node.tools[Go]
+        go.install_specific_version("1.19")
```
squirrelsc: If it needs to specify installed version, refer to `dpdktestpmd.py`. Use `__init__` to pass in the version like "1.19" or "". ("" means install from repo). Then use like `self.node.tools.get(Go, "1.19")` to initialize to special version like in `dpdkutil.py`.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
```diff
@@ -0,0 +1,117 @@
...
+    which uses infra of azure AKS as kubernetes cluster
+    """,
+)
+class KataConformanceTestSuite(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            Runs Kubernetes Kata Cotaniner Conformance test.
+        """,
+        priority=3,
+        timeout=36000,
```
squirrelsc: It means 10 hours? How long the test will take actually? If it's 10 hours, move it to P4.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
```diff
@@ -0,0 +1,117 @@
...
+from microsoft.testsuites.kata_conformance.aks_infra import AKSInfra
+from microsoft.testsuites.kata_conformance.kata_conformance_tests_tool import (
+    KataConformanceTests,
+)
+
+
+def validate_variable(variables, key, log, nullable=False):
+    log.debug(f"Checking for param : {key} in variables , nullable : {nullable}")
+    value = variables.get(key, None)
+    if (not nullable and value is None) or (
```
squirrelsc, line 28: In LISA, the principle is to set default values as much as possible, so the test cases is easy to run. We shouldn't ask users to provide every parameter to get test case work. So, if it's not nullable, it should use a default value, instead of fail. Unless it's credential values.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
```diff
@@ -0,0 +1,117 @@
...
+            log=log,
+            subscription_id=subscription_id,
+            client_id=client_id,
+            client_secret=client_secret,
+            tenant_id=tenant_id,
+        )
+        aks.create_aks_infra(
+            kubernetes_version, worker_vm_size, node_count, azure_region, headers
+        )
+        log.info("Completed AKS Infra Creation")
```
squirrelsc: The log lines should be in AKS class.

microsoft/testsuites/cloud_hypervisor/ch_tests.py
anirudhrb: +1

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -49,35 +51,84 @@ def run_tests(
...
+            assert_that(failures, f"Unexpected failures: {failures}").is_empty()
+        else:
+            temp = self.repo_root.joinpath("perf_mtr_report")
+            self.per_mtr_report_file = temp.joinpath("testcase_result.json")
+
+            self._build_ch_with_mshv(hypervisor=hypervisor)
+            perf_metrics_tests = self._list_perf_metrics_tests(hypervisor=hypervisor)
+            testcase_result = {}
+            for testcase in perf_metrics_tests:
+                self._log.info(f"Running testcase : {testcase}")
```
anirudhrb: The call to self.run below will print test case name anyway right?

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -158,3 +209,76 @@ def _send_subtest_msg(
...
 
         notifier.notify(subtest_msg)
+
+    def _build_ch_with_mshv(self, hypervisor="kvm") -> bool:
+        result = self.run(
+            f"build --hypervisor {hypervisor} --release",
+            timeout=self.TIME_OUT,
+            force_run=True,
+            cwd=self.repo_root,
+            no_info_log=False,  # print out result of each test
```
anirudhrb: Comment doesn't make sense here.

microsoft/testsuites/cloud_hypervisor/ch_tests.py
```diff
@@ -96,6 +96,31 @@ def verify_cloud_hypervisor_live_migration_tests(
             result, environment, "integration-live-migration", hypervisor
         )
 
+    @TestCaseMetadata(
+        description="""
+            Runs cloud-hypervisor live migration tests.
```
anirudhrb: Update the description

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -158,3 +209,76 @@ def _send_subtest_msg(
...
+    def _list_perf_metrics_tests(self, hypervisor="kvm") -> List[str]:
+
+        self._log.debug("Listing the performance test cases")
+        tests_list = []
+        result = self.run(
+            f"tests --hypervisor {hypervisor} --metrics -- -- --list-tests",
+            timeout=self.TIME_OUT,
+            force_run=True,
+            cwd=self.repo_root,
+            no_info_log=False,  # print out result of each test
```
anirudhrb: Comment doesn't make sense here. Please check all usages of no_info_log=False and keep it only if required.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -158,3 +209,76 @@ def _send_subtest_msg(
...
+            shell=True,
+            expected_exit_code=0,
+        )
+
+        stdout = result.stdout
+        regex = '\\"(.*)\\" \\('
+        pattern = re.compile(regex)
+        tests_list = pattern.findall(stdout)
+
+        self._log.debug(f"Testcase found : {tests_list}")
```
anirudhrb:
```suggestion
        self._log.debug(f"Testcases found : {tests_list}")
```

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -158,3 +209,76 @@ def _send_subtest_msg(
...
+            timeout=self.TIME_OUT,
+            force_run=True,
+            cwd=self.repo_root,
+            no_info_log=False,  # print out result of each test
+            shell=True,
+            expected_exit_code=0,
+        )
+
+        stdout = result.stdout
+        regex = '\\"(.*)\\" \\('
```
anirudhrb: Provide an example of the text you are trying to parse

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -158,3 +209,76 @@ def _send_subtest_msg(
...
+        cnt = 0
+        for line in output.split("\n"):
+            if line.find("git_human_readable") >= 0:
+                cnt -= 1
+                break
+            cnt += 1
+
+        result = "\n".join([i.strip() for i in output.split("\n")[cnt:-5]])
+        self._log.debug(f"Result from testcase stdout : {result}")
+        return result
```
anirudhrb: Provide an example output that you are trying to parse. For clarity.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -49,35 +51,84 @@ def run_tests(
...
+            temp = self.repo_root.joinpath("perf_mtr_report")
+            self.per_mtr_report_file = temp.joinpath("testcase_result.json")
+
+            self._build_ch_with_mshv(hypervisor=hypervisor)
+            perf_metrics_tests = self._list_perf_metrics_tests(hypervisor=hypervisor)
+            testcase_result = {}
+            for testcase in perf_metrics_tests:
+                self._log.info(f"Running testcase : {testcase}")
+                testcase_result[testcase] = {}
+                testcase_result[testcase]["status"] = "Started"
```
anirudhrb: Why not use `TestStatus` here instead of custom strings like "Started" and "Succeeded"?

microsoft/testsuites/kata_conformance/config_map.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kata-webhook
+data:
+  runtime_class: kata-qemu
```
anirudhrb, line 6: Why does it say qemu?

microsoft/testsuites/kata_conformance/create_pod.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: myapp-pod
+spec:
+  runtimeClassName: kata-qemu
```
anirudhrb, line 6: Why does it say qemu?

microsoft/testsuites/cloud_hypervisor/ch_tests.py
smit-gardhariya: Removed CH perf metric changes

microsoft/testsuites/cloud_hypervisor/ch_tests.py
smit-gardhariya: This we will manage in separate PR of CH. I will remove this as there is no specific requirements.

lisa/tools/go.py
smit-gardhariya: Made changes to accommodate version.

lisa/tools/go.py
smit-gardhariya: Removed Suse. Kept only Posix and failing on Windows

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
smit-gardhariya: Implemented

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
smit-gardhariya: For slower, system it takes around 8h. I changed it to 5h as of now. Made it P4

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
smit-gardhariya: Implemented defaults.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
smit-gardhariya: Log pushed to AKS class.

microsoft/testsuites/kata_conformance/aks_infra.py
```diff
@@ -0,0 +1,136 @@
...
+from azure.mgmt.containerservice import ContainerServiceClient
+from azure.mgmt.resource import ResourceManagementClient, SubscriptionClient
+
+from lisa.util import LisaException
+from lisa.util.logger import Logger
+
+global_credential_access_lock = Lock()
+
+
+class AKSInfra:
```
squirrelsc, line 18: It's discussed in previous PR.

1. Azure specified things should be in azure folder to share the logic. The credentials setup can be in a separated method in azure/common.py.
2. The ASK and k8s should be decoupled, if the test need to be in bare metal. I suppose it's not to trigger Azure test in bare metal, it's about to run kata in native k8s.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
```diff
@@ -0,0 +1,126 @@
...
+        subscription_id = validate_variable(
+            variables, "subscription_id", log, nullable=False
+        )
+        client_id = validate_variable(variables, "service_principal_client_id", log)
+        client_secret = validate_variable(variables, "service_principal_key", log)
+        tenant_id = validate_variable(variables, "service_principal_tenant_id", log)
+        kubernetes_version = validate_variable(
+            variables, "kubernetes_version", log, default="1.24.0"
+        )
+        worker_vm_size = validate_variable(
```
squirrelsc, line 80: I just realized the parameters are similar to VMs. So, it should be integrated deeper into LISA. Please use environment.topology to specify it's as "aks" topology. The most progresses are the same as VMs. Only when deploying, it needs to call the akc setup logic, instead of vm deploying logic. It can be supported by another arm_template or api based like this.

microsoft/testsuites/kata_conformance/aks_infra.py
```diff
@@ -0,0 +1,136 @@
...
+        self.log.info("Creating AKS Infra")
+        ts = datetime.datetime.now().strftime("%Y%m%d%H%M%S")
+        aks_cluster_name = f"LISA_AKS_Conf_{ts}"
+        self.resource_group_name = f"LISA_RG_Conf_{ts}"
+        self.log.info("Creating Resource Group : " + self.resource_group_name)
+
+        rg = self.rsc_mgmt_client.resource_groups.create_or_update(
+            self.resource_group_name, {"location": azure_region}
+        )
+        self.log.info("Resource Group Created : " + self.resource_group_name)
```
squirrelsc, line 82: use f string like `f"string: {variable}"`

lisa/util/shell.py
```diff
@@ -708,7 +712,8 @@ def stat(self, path: PurePath) -> os.stat_result:
         Outputs:
             os.stat_result: The status structure/class
         """
-        assert isinstance(path, Path), f"actual: {type(path)}"
+        path = Path(path)
+        # assert isinstance(path, Path), f"actual: {type(path)}"
```
squirrelsc, line 716: Just remove the commented line. The git can keep history.

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
```diff
@@ -0,0 +1,270 @@
...
+from lisa.util import LisaException
+
+
+@dataclass
+class KataConformanceTestResult:
+    name: str = ""
+    status: TestStatus = TestStatus.QUEUED
+
+
+class KataConformanceTests(Tool):
```
squirrelsc, line 25: Follow the name of test suite, like KataTest or KataContainersTest.

lisa/tools/go.py
```diff
@@ -0,0 +1,104 @@
...
+    def command(self) -> str:
+        return "go"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Wget, Tar, Rm]
```
squirrelsc, line 32: rm is an internal tool, it doesn't need to be added.

lisa/util/shell.py
```diff
@@ -663,7 +663,8 @@ def mkdir(
             parents: make parent directories as needed
             exist_ok: return with no error if target already present
         """
-        assert isinstance(path, Path), f"actual: {type(path)}"
+        path = Path(path)
```
squirrelsc, line 666: It's worth to be a small, separated PR. Let's get small and isolated changes in faster.

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
```diff
@@ -0,0 +1,270 @@
...
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git, Make, Go]
+
+    def run_tests(
+        self, test_result: TestResult, environment: Environment, failure_logs_path: Path
+    ) -> None:
+        echo = self.node.tools[Echo]
+        original_path = echo.run(
```
squirrelsc, line 55: The go path part should be handled by go itself. The Go tool should provide a method like `go_run`.

lisa/tools/go.py
```diff
@@ -0,0 +1,104 @@
...
+from lisa.tools.tar import Tar
+from lisa.util import LisaException
+
+
+class Go(Tool):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        go_version = kwargs.pop("go_version", None)
+        if go_version:
+            self._install_specific_version(go_version)
```
squirrelsc, line 20: Not to call install here. save the go_version to class, and call the source installation in the _install method.

lisa/tools/go.py
```diff
@@ -0,0 +1,104 @@
...
+        else:
+            raise LisaException(
+                "Doesn't support to install Go in Windows. "
+                "Make sure Go is installed and in PATH"
+            )
+        return self._check_exists()
+
+    def _install_specific_version(self, version: str) -> bool:
+        version_list = []
+        version_list.append("1.15")
```
squirrelsc, line 46: The logic looks awful, and I don't want to add new versions each time. It should directly download the go package. If it doesn't exist, raise the error with friendly message.

lisa/tools/go.py
```diff
@@ -0,0 +1,104 @@
...
+        # Add installation path to env variable PATH
+        echo = self.node.tools[Echo]
+        original_path = echo.run(
+            "$PATH",
+            shell=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="failure to grab $PATH via echo",
+        ).stdout
+        new_path = f"/usr/local/go/bin/:{original_path}"
+        self._log.debug("NewPath : " + str(new_path))
```
squirrelsc, line 85: After development, some noise debug level messages should be removed. This kind of information can be found in LISA lower level logs too.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
squirrelsc: Is it possible to reduce test time? It's ok to keep 10 hours, but as a normal test condition, it takes too long to complete if it takes more than 1 hour. Please check if it's possible to reduce running time.

anirudhrb:
Please update the title.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
smit-gardhariya: Kata test takes more than 1 hour as it includes AKS infra creation, Kubernetes Build, Kata webhook deployment on created AKS and then it launches the Conformance test which are 350+ in number.

microsoft/testsuites/kata_conformance/aks_infra.py
smit-gardhariya: sure, i will use it that way.

lisa/util/shell.py
smit-gardhariya: sure, i will do it.

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
smit-gardhariya: sure, i will use KataTest

lisa/tools/go.py
smit-gardhariya: sure

lisa/util/shell.py
smit-gardhariya: sure

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
smit-gardhariya: I tried setting Path in Go tool itself but it seems like we need to pass updated PATH every time we run command so i kept it here as well. You can see i am passing path while running command with self.run

lisa/tools/go.py
smit-gardhariya: sure, i will make that change.

lisa/tools/go.py
smit-gardhariya: Sure, i will make change and if any error occurs, I will throw error with proper message.

lisa/tools/go.py
smit-gardhariya: sure , will take care of this

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
squirrelsc: Do you mean the whole progress is around 1 hour in normal situation? If so, it can be P3. If it's more than 2 hours, it should be moved to P4.

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
squirrelsc: The PATH should be maintained in go tool, because it's part of the go tool. The go tool can provide a `go_run` method. The method can use the PATH as the env.

microsoft/testsuites/kata_conformance/kata_conformance_tests.py
smit-gardhariya: Yes it can take more than2 Hours. It depends on VM size from where we are launching the test. In this case, it would be the agent vm where pipeline is running. It will create AKS infra on Azure and try to run conformance test there but it can be launched from anywhere.

smit-gardhariya: Will make this P4

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
smit-gardhariya: This is removed to another PR having CH perf metrics changes only

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
smit-gardhariya: Taken care by CH perf metrics changes PR

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
smit-gardhariya: Taken care by CH perf metrics changes PR

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
smit-gardhariya: Taken care by CH perf metrics changes PR

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
smit-gardhariya: Taken care by CH perf metrics changes PR

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
smit-gardhariya: Taken care by CH perf metrics changes PR

microsoft/testsuites/kata_conformance/config_map.yaml
smit-gardhariya: this is the runtime class. I guess it can have any name. I followed the name mentioned in document shared by Vivek at earlier stage.

smit-gardhariya: This is the runtime class we create with runtime_class.yaml

microsoft/testsuites/kata_conformance/create_pod.yaml
smit-gardhariya: This is the runtime class we create with runtime_class.yaml. It can have any name i guess. I followed which is mentioned in the doc shared by Vivek at early stage.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
smit-gardhariya: separate PR raised.

microsoft/testsuites/cloud_hypervisor/ch_tests.py
smit-gardhariya: separate PR raised

microsoft/testsuites/kata_conformance/kata_conformance_tests_tool.py
smit-gardhariya: We need to set PATH everytime when we execute command.

Here we will not use GO command explicitly to run or build something. 

When we perform MAKE on Kubernetes repo, it expects GO command in PATH. That is why i am calling _install_dep() first and then building Kubernetes with MAKE in _install() method.

I will create new method to retrieve PATH with go tool.

squirrelsc: What I mean is the logic should be in the `go.py` instead of the upper layer tool. So the Go tool in `go.py` is easier to use.

LiliDeng:
@smit-gardhariya any update?

LiliDeng:
@smit-gardhariya do you plan to update this PR? Thanks.

LiliDeng:
@smit-gardhariya can we close this one, create a new PR if it is ready for review, thanks.

LiliDeng:
discussed with @smit-gardhariya , he will create new one once it is ready for reviewing.

PR #2271, opened 2022-09-26
Add a VM create stress test for the MSHV stack

anirudhrb:
Introduce a new test suite for testing the MSHV root partition. Add a test case that creates and destroys multiple VMs in parallel and repeats this multiple times to stress the virtualization stack.

### DIFF
--- a/lisa/tools/free.py
+++ b/lisa/tools/free.py
@@ -96,3 +96,6 @@ class Free(Tool):
         group = find_group_in_lines(output, self._mem_pattern)
         total_memory = group["total"]
         return total_memory
+
+    def log_memory_stats_mb(self) -> None:
+        self.run("-m", shell=True, force_run=True, no_info_log=False)
--- a/microsoft/testsuites/mshv/cloud_hypervisor_tool.py
+++ b/microsoft/testsuites/mshv/cloud_hypervisor_tool.py
@@ -1,0 +1,31 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+from lisa.util.process import Process
+
+
+class CloudHypervisor(Tool):
+    @property
+    def command(self) -> str:
+        return "cloud-hypervisor"
+
+    @property
+    def can_install(self) -> bool:
+        # cloud-hypervisor is already installed in MSHV dom0 image.
+        return False
+
+    def start_vm_async(
+        self,
+        kernel: str,
+        cpus: int,
+        memory_mb: int,
+        disk_path: str,
+        disk_readonly: bool = False,
+    ) -> Process:
+        opt_disk_readonly = "on" if disk_readonly else "off"
+        return self.run_async(
+            f'--kernel {kernel} --cpus boot={cpus} --memory size={memory_mb}M --disk "path={disk_path},readonly={opt_disk_readonly}" --net "tap=,mac=,ip=,mask="',  # noqa: E501
+            force_run=True,
+            shell=True,
+        )
--- a/microsoft/testsuites/mshv/mshv_root_tests.py
+++ b/microsoft/testsuites/mshv/mshv_root_tests.py
@@ -1,0 +1,193 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import time
+from pathlib import Path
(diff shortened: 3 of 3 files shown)


microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -0,0 +1,198 @@
...
+        self,
+        log: Logger,
+        node: Node,
+        variables: Dict[str, Any],
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+    ) -> None:
+        if self.CONFIG_VARIABLE in variables:
+            configs = variables[self.CONFIG_VARIABLE]
```
squirrelsc, line 85: Give each variable a default value, so, the test case can run without any configuration. And it's better to expand each variable, so it's easy to set some of them, instead of all each time.

microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -0,0 +1,198 @@
...
+                    node,
+                    log_path,
+                )
+                self._send_subtest_msg(
+                    result.id_,
+                    environment,
+                    test_name,
+                    TestStatus.PASSED,
+                )
+            except Exception:
```
squirrelsc: The exception instance should be sent by the error message. So, it's easy to troubleshoot.

microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -0,0 +1,198 @@
...
+    @TestCaseMetadata(
+        description="""
+            Stress the MSHV virt stack by repeatedly creating and destroying
+            multiple VMs in parallel. By default creates VMs with 1 vCPU and
+            1 GiB of RAM each. Number of VMs createdis equal to the number of
+            CPUs available on the host. By default, the test is repeated 25
+            times. All of these can be configured via the variable
+            "mshv_vm_create_stress_configs" in the runbook.
+        """,
+        priority=3,
```
squirrelsc: For stress tests, move it to P4.

lisa/tools/free.py
```diff
@@ -96,3 +96,6 @@ def get_total_memory(self) -> str:
         group = find_group_in_lines(output, self._mem_pattern)
         total_memory = group["total"]
         return total_memory
+
+    def log_memory_stats_mb(self) -> None:
```
squirrelsc, line 100: return the value as float type, so the logged value can be used in code.

microsoft/testsuites/mshv/cloud_hypervisor_tool.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class CloudHypervisor(Tool):
+    @property
+    def command(self) -> str:
+        return "cloud-hypervisor"
+
+    @property
+    def can_install(self) -> bool:
+        # cloud-hypervisor is already installed in MSHV dom0 image.
```
squirrelsc, line 15: We can use like this so far, but in future, the installation should be in this tool. If it's not detected, the install method will be called to install it.

lisa/tools/free.py
anirudhrb: It actually logs a bunch of values. For example:

```
              total        used        free      shared  buff/cache   available
Mem:          55765        2904       23381           1       29479       52212
Swap:             0           0           0
```

microsoft/testsuites/mshv/mshv_root_tests.py
anirudhrb: The test case can be run without any configuration as-is. I have now implemented a default value for each parameter so that some of them can be skipped.

microsoft/testsuites/mshv/mshv_root_tests.py
anirudhrb: Done.

microsoft/testsuites/mshv/cloud_hypervisor_tool.py
anirudhrb: cloud-hypervisor is always pre-installed in the MSHV root image so for now I have skipped the installation. If in the future this is made into a common tool, then we would need to implement the installation.

squirrelsc (approved):

PR #2273, opened 2022-09-26
Update to use Cache Ephemeral disk

dsrivastavv:
ResourceDisk size can be limited to less than marketplace image size leading to test failures

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -461,7 +461,7 @@ class DiskOptionSettings(FeatureSettings):
         ),
     )
     data_disk_size: search_space.CountSpace = field(
-        default_factory=partial(search_space.IntRange, min=1),
+        default_factory=partial(search_space.IntRange, min=0),
         metadata=field_metadata(
             allow_none=True, decoder=search_space.decode_count_space
         ),
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -306,7 +306,7 @@
                     "imageReference": "[if(not(empty(parameters('nodes')[copyIndex('vmCopy')]['vhd'])), lisa.getOsDiskVhd(parameters('nodes')[copyIndex('vmCopy')]['name']), if(not(empty(parameters('nodes')[copyIndex('vmCopy')]['shared_gallery'])), lisa.getOsDiskSharedGallery(parameters('nodes')[copyIndex('vmCopy')]['shared_gallery']), lisa.getOsDiskMarketplace(parameters('nodes')[copyIndex('vmCopy')])))]",
                     "osDisk": {
                         "name": "[concat(parameters('nodes')[copyIndex('vmCopy')]['name'], '-osDisk')]",
-                        "diffDiskSettings": "[if(equals(parameters('nodes')[copyIndex('vmCopy')]['disk_type'], 'Ephemeral'), json('{\"option\": \"local\", \"placement\": \"ResourceDisk\"}'), json('null'))]",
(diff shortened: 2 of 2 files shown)


dsrivastavv:
@LiliDeng Tested that this works on DS2V2. If we find further VM sizes that have less cache size than Image VHD size, we can remove `ephemeral` features from such VM sizes.

squirrelsc (dismissed):

LiliDeng:
which size you used? I always see case skip. 
use temp disk or os cache, both needs temp disk, cache disk >= os disk size, see below link.
https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks-faq

we need two cases when use temp disk or os cache disk for diffDiskSettings

LiliDeng (changes requested):
need further discussion for this change

squirrelsc:
> which size you used? I always see case skip. use temp disk or os cache, both needs temp disk, cache disk >= os disk size, see below link. https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks-faq
> 
> we need two cases when use temp disk or os cache disk for diffDiskSettings

As the example, the cache size is much larger than temp size disk. That's why it works. It's a quick workaround to make the test case works. If the diff disk setting is needed for testing in future, we can read both of them from sku properties.

https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks#size-requirements

LiliDeng:
> > which size you used? I always see case skip. use temp disk or os cache, both needs temp disk, cache disk >= os disk size, see below link. https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks-faq
> > we need two cases when use temp disk or os cache disk for diffDiskSettings
> 
> As the example, the cache size is much larger than temp size disk. That's why it works. It's a quick workaround to make the test case works. If the diff disk setting is needed for testing in future, we can read both of them from sku properties.
> 
> https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks#size-requirements

ok, I will sync with @somil55 since I always see case skip locally with this change.

dsrivastavv:
@LiliDeng Let's sync offline on this. We can merge the test case once you the changes work for you

lisa/sut_orchestrator/azure/features.py
```diff
@@ -956,7 +956,7 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
         # The Ephemeral doesn't support data disk, but it needs a value. And it
         # doesn't need to calculate on intersect
         value.data_disk_iops = 0
-        value.data_disk_size = 0
+        value.data_disk_size = 1
```
dsrivastavv: @squirrelsc We need to be consistent. The disk class below has min value 1 whereas in code above we were using 0. A value of 1 seems more practical which is why I have made the change here. Please let me know if you think otherwise.

```
data_disk_size: search_space.CountSpace = field(
        default_factory=partial(search_space.IntRange, min=1),
        metadata=field_metadata(
            allow_none=True, decoder=search_space.decode_count_space
        ),
    )
```

LiliDeng (dismissed):

lisa/sut_orchestrator/azure/features.py
squirrelsc: The below logic will assign a data_disk_size in all logic. the "0" is like to declare a variable with default value. 1 and 0 are the same, because it will be overwritten immediately. But 0 is a good default value, when people need to think about why it's 1.

dsrivastavv: Sure. Let's use default of 0 in that case.

dsrivastavv: updated

squirrelsc (approved):

PR #2274, opened 2022-09-27
add case verify_resource_disk_readme_file

LiliDeng:


microsoft/testsuites/core/storage.py
```diff
@@ -176,6 +176,51 @@ def verify_resource_disk_io(self, log: Logger, node: RemoteNode) -> None:
             "content read from file should be equal to content written to file",
         ).is_equal_to(original_text)
 
+    @TestCaseMetadata(
```
squirrelsc: Should this test case be in Azure standard test suite?

LiliDeng: updated

lisa/tools/lsblk.py
```diff
@@ -42,6 +43,7 @@ def __init__(
         used_blocks: int = 0,
         total_blocks: int = 0,
         percentage_blocks_used: int = 0,
+        fstype: str = "",
```
dsrivastavv, line 46: Can we set this to None instead?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -891,3 +897,74 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
+        ),
+    )
+    def verify_resource_disk_readme_file(self, log: Logger, node: RemoteNode) -> None:
+        resource_disk_mount_point = get_resource_disk_mount_point(log, node)
+
+        # verify that resource disk is mounted
+        # function returns successfully if disk matching mount point is present
+        node.features[Disk].get_partition_with_mount_point(resource_disk_mount_point)
+
+        # verify lost+found folder exists
```
dsrivastavv, line 928: The current test case name doesn't capture this? Should we create a new test for this?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -891,3 +897,74 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
+
+    @TestCaseMetadata(
+        description="""
+        This test will check that resource disk is formatted correctly.
+
+        Steps:
+        1. Get the mount point for the resource disk. If `/var/log/cloud-init.log`
+        file is present, mount location is `/mnt`, otherwise it is obtained from
+        `ResourceDisk.MountPoint` entry in `waagent.conf` configuration file.
+        2. Verify that resource disk file system type should not be 'ntfs'.
```
dsrivastavv, line 954: [nit] Can you please provide reason why this should not be NTFS?

squirrelsc (dismissed):

microsoft/testsuites/core/common.py
```diff
@@ -0,0 +1,25 @@
+from pathlib import PurePosixPath
+
+from lisa import Logger, RemoteNode
+from lisa.operating_system import CentOs
+from lisa.sut_orchestrator.azure.tools import Waagent
+
+
+def get_resource_disk_mount_point(
```
dsrivastavv, line 8: Maybe we can move this function to disk feature?

microsoft/testsuites/core/azure_image_standard.py
LiliDeng: no, not covered in current tests, to make things simple, we'd better to add a new case

microsoft/testsuites/core/azure_image_standard.py
LiliDeng: by default, the resource disk comes with ntfs, I added this case for one customer incident, there is hung caused by udevadm settle command, then the resource disk is not formatted correctly.

lisa/tools/lsblk.py
LiliDeng: why?

LiliDeng: self.fstype = fstype if fstype is not None else ""
any difference?

microsoft/testsuites/core/common.py
LiliDeng: we can work on it after this pr

microsoft/testsuites/core/azure_image_standard.py
LiliDeng: updated comments

lisa/tools/lsblk.py
dsrivastavv: I was think of something more like: fstype: Optional[str] =None. Though not sure if this could be possible here. Please feel free to close this if this is not possible

squirrelsc (approved):

PR #2279, opened 2022-09-27
Add Storage FIO testcases

smit-gardhariya:
This PR contains generic FIO testcase addition. Based on runbook params, FIO will be launched for storage perf test. We can pass list of params to cover more than one scenario. If nothing is passed, testcase will launch default testcases mentioned in json file


LiliDeng:
@smit-gardhariya please check perf_premium_datadisks_4k, whether they are duplicated, if not, you can refer this case with different parameters of fio.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +194,109 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
+                fio_result[test_name]["metrics"] = {}
+                log.error(e)
+
+            rm.remove_directory(str(path.joinpath(f"{filename}*")), sudo=True)
+
+        log.debug(fio_result)
+
+        # Saving the result
+        report_path = node.get_working_path().joinpath("fio_report")
+        report_file = report_path.joinpath("fio_test_result.json")
```
squirrelsc: Don't use a file to dump the test results, it's not extendable. Use standard way, send sub tests messages to dump the results.

smit-gardhariya:
> @microsoft-github-policy-service agree company="Microsoft"

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +194,109 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
         self._perf_nfs(result, protocol="udp")
 
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test data disk performance with 4K block size.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+    )
+    def perf_storage_generic_fio_test(self,
```
dsrivastavv: Can you please explain in comments how these tests are different from `perf_premium_datadisks_4k`?

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +194,109 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
+            test_name = testcase["testcase_name"]
+            fio_result[test_name] = {}
+            fio_result[test_name]["runtime_param"] = testcase
+
+            try:
+                path = node.get_working_path()
+
+                log.info(f"Executing the FIO testcase : {test_name}")
+
+                name = testcase.get("name", None)
```
dsrivastavv: These can be handled in a separate class encapsulating these variables

microsoft/testsuites/performance/default_fio_testcases.json
```diff
@@ -0,0 +1,116 @@
...
+            "direct": 0,
+            "size_mb": 1024,
+            "numjob": 4,
+            "time": 240,
+            "group_reporting": true
+        },
+        {
+            "name": "write",
+            "filename": "Write_Size_4G",
+            "testcase_name": "write_with_4gb",
```
anirudhrb: This seems a bit pointless. We can auto generate a name by using the other parameters.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +194,109 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
+    def perf_storage_generic_fio_test(self,
+        log: Logger,
+        node: Node,
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        testcases = variables.get("fio_testcase_list", None)
+        if (not testcases):
```
anirudhrb: redundant parentheses

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +194,109 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
     def perf_storage_over_nfs_synthetic_udp_4k(self, result: TestResult) -> None:
         self._perf_nfs(result, protocol="udp")
 
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test data disk performance with 4K block size.
```
anirudhrb: Update description

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: This test case is not hardcoded and run the test case as per runbook parameters. Even mode for running FIO test case is flexible and can be configured with parameters. This will provide flexibility to run FIO tests easily by just passing parameters in runbook as per need.

Whereas perf_premium_datadisks_4k seems doing same thing but with hardcoded parameters. It will launch FIO test for all 4 modes as perf_disk function run for 4 modes. Please let me know your thoughts on this.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Sure we can do this. I will look for this.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Sure Chi. I will do it.

microsoft/testsuites/performance/default_fio_testcases.json
```diff
@@ -0,0 +1,116 @@
+{
```
anirudhrb: We can use the logic of perf_premium_datadisks_4k as the default behavior instead of having this json file. Since perf_premium_datadisks_4k is already in use for Azure VM testing, I would say it is a very good default. This would also be in line with your point that the new test case is generic/configurable for version of perf_premium_datadisks_4k. In the absence of said configuration, it can fallback to perf_premium_datadisks_4k.

microsoft/testsuites/performance/storageperf.py
squirrelsc: I just realized it's another fio test. Please work with Lili to share code as much as possible.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +194,109 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
+        node: Node,
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        testcases = variables.get("fio_testcase_list", None)
+        if (not testcases):
+            testcases_file_path = os.path.join(
+                os.path.abspath(__file__).replace(os.path.basename(__file__), ""),
```
squirrelsc: `os.path.abspath(__file__).parent` is better. If it doesn't work try with pathlib. pathlib is shorter and intuitive.

squirrelsc: BTW, I feel it doesn't need a file. It can be put in code as dicts.

microsoft/testsuites/performance/default_fio_testcases.json
squirrelsc: That's fair. There are several things,

1. data driven test cases. If the parameters are only different things like this case. It's definitely should be data driven to share logics as much as possible. The shared data can be dict, or json, it doesn't matter.
2. For hard coded test cases names like `perf_premium_datadisks_4k`. It's necessary. When we test on different scenarios, we need standardized parameters to compare across multiple runs. It shouldn't need to ask every runner to choose from parameters.
3. For customizable parameters like this, it may be needed to handle occasional scenarios. But it shouldn't be overused to create incomparable results.

microsoft/testsuites/performance/storageperf.py
dsrivastavv: How about we modify `perf_premium_datadisks_4k` to work with current hardcoded setting if no parameters are specified, otherwise run it based on parameters? Also, does your testcase work with block size other than `4k`?

dsrivastavv: My concern is primarily to share as much code as possible. Please feel free to suggest any alternatives

anirudhrb: `perf_premium_datadisks_4k` has Azure specific naming as well as `requirements`. That's why we can't use it directly.
@smit-gardhariya Why not re-use the underlying `perf_disk` function? If no parameters are provided we can fallback to the configuration used by `perf_premium_datadisks_4k`.

smit-gardhariya: Sure @anirudhrb we can do it. I will make that change

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: leveraging use of existing perf_disk function

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Using perf_disk for all operation having leverage of passing params from runbook as well to run FIO with different configurations. perf_disk is running all FIO mode with all these configs if passed or perf_disk will be executed with default config.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Removed

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Done.

microsoft/testsuites/performance/default_fio_testcases.json
smit-gardhariya: To bring consistency in result, we are launching all test cases using perf_disk only now. If no param passed via runbook, we can launch some default params. This testcase is neither attached to specific scenario nor having specific requirements to run with.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: I used file earlier  so we can make add/change testcase without code change. But We are using perf_disk function now with default parameters. This is resolved now.

microsoft/testsuites/performance/default_fio_testcases.json
smit-gardhariya: It is resolved with new changes as we not have these fields now in runtime parameters

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: We have used perf_disk now to run testcase with default and runtime parameters. Please check

microsoft/testsuites/performance/storageperf.py
```diff
@@ -210,92 +206,68 @@ def perf_storage_generic_fio_test(self,
...
+                        test_result=result,
+                        test_name=test_name,
+                        num_jobs=num_jobs,
+                        block_size=block_size,
+                        time=time,
+                        size_mb=size_mb,
+                        overwrite=overwrite,
+                    )
+                except Exception as e:
+                    log.error(e)
```
pupacha: We should let the exception be raised. Else the testcase would get into PASSED state.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +189,86 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
+        self,
+        log: Logger,
+        node: Node,
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        testcases = variables.get("fio_testcase_list", None)
+        if not testcases or len(testcases) == 0:
```
pupacha, line 228: Instead of calling `per_disk` in both if & else, can we just fill the `testcases` variable with default values in `if` part and let the code run generic without `else` part.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: We need try block as we have list of testcases. I have put assert with latest commit to reflect correct status of lisa testcase.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: resolved.

pupacha (dismissed):

microsoft/testsuites/performance/common.py
```diff
@@ -40,11 +38,11 @@ def perf_disk(
...
     filename: str,
-    core_count: int,
-    disk_count: int,
-    disk_setup_type: DiskSetupType,
-    disk_type: DiskType,
     test_result: TestResult,
+    core_count: int = 0,
+    disk_count: int = 0,
+    disk_setup_type: Any = "",
+    disk_type: Any = "",
```
anirudhrb: Why the change in data types and ordering of the parameters?

smit-gardhariya: Datatype were specific to already implemented testcases. This is not needed by our generic testcase of FIO and these args are getting used to stamp with notifier message only.  So, We have to make it default to use the same function with our testcase.

Order has to be changed as default args cannot be written before mandatory args.

anirudhrb: Why not use `Optional` instead of changing type to `Any`?

microsoft/testsuites/performance/common.py
```diff
@@ -40,10 +40,10 @@ def perf_disk(
     start_iodepth: int,
     max_iodepth: int,
     filename: str,
-    core_count: int,
-    disk_count: int,
-    disk_setup_type: DiskSetupType,
-    disk_type: DiskType,
+    core_count: Optional[int],
```
squirrelsc: What's the reason to mark them optional? In the new change, they are assigned.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +189,86 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
+                    size_mb=size_mb,
+                    overwrite=overwrite,
+                    core_count=core_count,
+                    disk_count=1,
+                    disk_setup_type=None,
+                    disk_type=None,
+                )
+            except Exception as err:
+                failed_test_cases.append(testcase)
+                log.error(err)
```
squirrelsc: It doesn't need to log error message on test case level.

microsoft/testsuites/performance/common.py
squirrelsc: They are metrics, which are collected to database. So it's better to fill them in to get data more accurate.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +189,86 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
         self._perf_nfs(result, protocol="udp")
 
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test data disk performance.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+    )
+    def perf_storage_generic_fio_test(
```
LiliDeng, line 199: I didn't see any data disk requirement set in case requirement.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Sure Chi, i will remove it.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Yes, this is not specific to any disk type or network interface.

microsoft/testsuites/performance/common.py
smit-gardhariya: Yes in latest push we assigned core_count as per core and disk_count as 1 by default. We need to make disk_setup_type and disk_type as option so we can pass None to them and there will be minimal change in signature of perf_disk.

New testcase does not require disk specification. This is generic FIO test which can run as per passed params in runbook.

I will remove optional from core_count  and disk_count and will keep optional with disk_setup_type and disk_type otherwise flak8 check will fail.

smit-gardhariya:
As discussed with Lili, i have created new PR for better commit traceability.  Closing this PR.

New PR : https://github.com/microsoft/lisa/pull/2325

squirrelsc:
> As discussed with Lili, i have created new PR for better commit traceability. Closing this PR.
> 
> New PR : #2325

It doesn't need to create another PR to merge commits. The new PR lost all comments, it's hard to track changes.

PR #2280, opened 2022-09-27
Runner: Support to wait for await able resources on preparing and deploying phases.

squirrelsc:
See details in commits, and some other refactoring included.


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -421,7 +421,7 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
         if not is_success:
             if any_wait_for_resource:
                 raise ResourceAwaitableException(
-                    "No available quota, try to deploy later."
+                    "vm size", "No available quota, try to deploy later."
```
LiliDeng, line 424: do you want to update this?

PR #2284, opened 2022-09-28
Move the kernel config check to the first so it can determine if the distro supports kdump by introducing the KdumpBase tool

lubaihua33:
We move the kernel config check to the first, so we can determine if the distro supports kdump by introducing the KdumpBase tool.

### DIFF
--- a/lisa/tools/kdump.py
+++ b/lisa/tools/kdump.py
@@ -190,6 +190,7 @@ class KdumpBase(Tool):
 
     @classmethod
     def create(cls, node: "Node", *args: Any, **kwargs: Any) -> Tool:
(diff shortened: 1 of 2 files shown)


squirrelsc:
> We move the kernel config check to the first, so we can determine if the distro supports kdump by introducing the KdumpBase tool.

This should be included in commit comments too.

lubaihua33:
> > We move the kernel config check to the first, so we can determine if the distro supports kdump by introducing the KdumpBase tool.
> 
> This should be included in commit comments too.

Have added.

squirrelsc (approved):

PR #2285, opened 2022-09-28
Fix after case raid cleanup

dsrivastavv:
Raid cleanup fails if the server is running and references the raid disk

### DIFF
--- a/microsoft/testsuites/performance/storageperf.py
+++ b/microsoft/testsuites/performance/storageperf.py
@@ -3,7 +3,7 @@
 
 import inspect
 from pathlib import PurePosixPath
-from typing import Any, Dict, cast
+from typing import Any, Dict, List, cast
 
 from assertpy import assert_that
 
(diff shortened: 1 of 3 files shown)


squirrelsc (dismissed):

microsoft/testsuites/performance/storageperf.py
```diff
@@ -329,4 +329,11 @@ def _perf_premium_datadisks(
 
     def after_case(self, log: Logger, **kwargs: Any) -> None:
         node: Node = kwargs.pop("node")
+
+        # stop nfs server if it's running
+        nfsserver = node.tools[NFSServer]
```
LiliDeng: will other cases which don't use nfsserver be installed the tool here?

dsrivastavv: Nice catch. This should prompt installation. Maybe we can make this static function in the class.

dsrivastavv: This is not applicable anymore

dsrivastavv:
Tested that perf_storage_over_nfs_synthetic_udp_4k and perf_storage_over_nfs_synthetic_tcp_4k can run on shared node in either order

microsoft/testsuites/performance/storageperf.py
```diff
@@ -291,6 +291,11 @@ def _perf_nfs(
         for node in [server_node, client_node]:
             node.tools[Sysctl].write("fs.aio-max-nr", origin_value[node.name])
 
+        # clean up
```
squirrelsc: It should be in a final block to make sure it's always called. But not wrap above long content in the try/final block. create another method to put above content.

dsrivastavv: fixed

squirrelsc: Please take the comment " But not wrap above long content in the try/final block. create another method to put above content."

dsrivastavv: You mean the content in the current try block? The function `_perf_nfs` itself was created to encapsulate this.

squirrelsc: The try block is too long, and not clear. I double looked the code, the try can start from Ln257 `# setup nfs on server`. So it will be shorter 50%.

dsrivastavv: Actually, we need to start it from Line250. I missed deleting `server_raid_disk_mount_dir` as part of cleanup. Let me add this to cleanup as well.

squirrelsc: Because the try block is too long to read, so it's better to wrap the try part into a method like `_perf_nfs_core`, then one line leaves in try block.

squirrelsc (approved):

PR #2286, opened 2022-09-28
DPDK: restore min_core_count requirement

mcgov:
verify_build tests removed isolated resource but still need the min core count requirement.

Defensively add this arg back to the other DPDK tests which require it.


LiliDeng (approved):

squirrelsc:
> verify_build tests removed isolated resource but still need the min core count requirement.
> 
> Defensively add this arg back to the other DPDK tests which require it.

Any error found from the low core count? They should run on high end isolated VMs already.

mcgov:
Yep, the error was with verify_build_[failsafe|netvsc]. We had removed the requirement for those since they only run for 10 seconds but without the core requirement set it was running on a 4 core machine.

PR #2287, opened 2022-09-29
DPDK: fix broken OVS version check

mcgov:
Previously the test assumed an explicit version would be passed in, and would just fail if the DPDK and OVS versions didn't match up.

This commit switches to checking DPDK version first and then checking out the compatible OVS version to run the test.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkovs.py
+++ b/microsoft/testsuites/dpdk/dpdkovs.py
@@ -9,7 +9,7 @@ from semver import VersionInfo
 from lisa.executable import Tool
 from lisa.operating_system import Debian, Fedora
 from lisa.tools import Chown, Gcc, Git, Ip, Make, Modprobe, Uname, Whoami
-from lisa.util import UnsupportedDistroException
+from lisa.util import SkippedException, UnsupportedDistroException
 from microsoft.testsuites.dpdk.dpdktestpmd import DpdkTestpmd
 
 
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkovs.py
```diff
@@ -111,42 +111,49 @@ def _check_ovs_dpdk_compatibility(self, dpdk_tool: DpdkTestpmd) -> None:
...
-        # check if version too high (pairing is unknown)
-        elif self.ovs_version > "2.18.0":
-            raise NotImplementedError(
-                (
-                    f"Not implemented OVS version {self.ovs_version}. "
-                    "Please update the version requirement table in "
-                    "dpdkovs.py:check_ovs_dpdk_compatibilty with new data"
-                    "from https://docs.openvswitch.org/en/latest/faq/releases/"
+        else:
+            if dpdk_version < "19.11.0":
```
squirrelsc: Why low dpdk version skipped, but low ovs version failed?

microsoft/testsuites/dpdk/dpdkovs.py
```diff
@@ -111,42 +111,49 @@ def _check_ovs_dpdk_compatibility(self, dpdk_tool: DpdkTestpmd) -> None:
...
+                raise SkippedException(
+                    f"Dpdk version {dpdk_version} is not supported by this test."
+                )
+
+            if int(dpdk_version.major) not in dpdk_to_ovs_minimum_version.keys():
+                # we've already checked out latest OVS
+                # DPDK version is above 22, warn and proceed.
+                self.node.log.warn(
```
squirrelsc: info level should be enough, actually, nobody is looking for warning message.

microsoft/testsuites/dpdk/dpdkovs.py
```diff
@@ -111,42 +111,49 @@ def _check_ovs_dpdk_compatibility(self, dpdk_tool: DpdkTestpmd) -> None:
...
+                    "DPDK version is above the maximum in the version match"
+                    " table. Using latest OVS."
                 )
+                return
```
squirrelsc: Does it mean not to run following ovs enlistment code? So what happens, if return here?

microsoft/testsuites/dpdk/dpdkovs.py
mcgov: Hmm good point! I'll skip.

microsoft/testsuites/dpdk/dpdkovs.py
mcgov: OVS is already checking out latest at the top of the test, so if the DPDK version is also above the levels in our table (22 and beyond) it should be fine to just run with what's already checked out. I'll comment this a little better.
I don't really like that this table has to be updated constantly, but if it starts failing we'll see the note in the log and know that the test might need an update to the table.

mcgov: The alternative is pulling the rst table from the docs and parsing it though, so this seems cleaner than that.

microsoft/testsuites/dpdk/dpdkovs.py
mcgov: I forgot this wasn't even needed with this change. We check out latest, check the installed dpdk version, adjust OVS version lower if needed. OVS won't ever be too low of a version so we can just ditch that entirely.

squirrelsc (approved):

microsoft/testsuites/dpdk/dpdkovs.py
squirrelsc: Fantastic version compatibility handling!

PR #2288, opened 2022-09-30
[new case] verify_azure_file_share_nfs

LiliDeng:


pyproject.toml
```diff
@@ -54,6 +54,7 @@ sphinxemoji = {version="^0.1.8", optional=true}
 sphinx-copybutton = {version="^0.4.0", optional=true}
 rstcheck = {version="^3.3.1", optional=true}
 types-Pillow = {version="^8.3.3", optional=true}
+azure-mgmt-privatedns = "^1.0.0"
```
squirrelsc: 1. make it optional like other azure packages.
2. Add it in the internal repo to make sure once it's merged, it won't break CI runs.

microsoft/testsuites/core/storage.py
```diff
@@ -345,6 +359,119 @@ def hot_add_disk_parallel_premium_ssd(self, log: Logger, node: Node) -> None:
...
+        resource_group_name = information["resource_group_name"]
+        location = information["location"]
+        random_str = "".join(
+            random.choices(string.ascii_lowercase + string.digits, k=10)
+        )
+        storage_account_name = f"lisasc{random_str}"
+        file_share_name = f"lisa{random_str}fs"
+        log.debug(f"creating storage account {storage_account_name}")
+        # create storage account and file share
+        check_or_create_storage_account(
```
squirrelsc: It should be wrapped in a feature called like Nfs, instead of call Azure method in the test case. Although it's not related to vm size, but it can be attached to node as a feature.

pyproject.toml
LiliDeng: done

microsoft/testsuites/core/storage.py
LiliDeng: add Nfs feature

LiliDeng:
pass in the internal pipeline
```2022-10-09 02:21:01.454[5400][INFO] lisa.RootRunner                Storage.verify_azure_file_share_nfs: PASSED```

File: lisa/features/nfs.py [hunk 3]
```diff
@@ -0,0 +1,24 @@
...
+    def can_disable(cls) -> bool:
+        raise NotImplementedError()
+
+    def enabled(self) -> bool:
+        raise NotImplementedError()
+
+    def is_supported(self) -> bool:
+        raise NotImplementedError()
+
+    def create(self) -> str:
```
squirrelsc: 1. `create_share`, and does it need to expose a share name with a default value?
2. Does it need to be deleted too?

pyproject.toml
squirrelsc: didn't see internal PR created.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1601,3 +1611,93 @@ def create_setting(
...
+        super()._initialize(*args, **kwargs)
+        self._initialize_information(self._node)
+
+    def create(self) -> str:
+        platform: AzurePlatform = self._platform  # type: ignore
+        node_context = self._node.capability.get_extended_runbook(AzureNodeSchema)
+        location = node_context.location
+        resource_group_name = self._resource_group_name
+        random_str = "".join(
+            random.choices(string.ascii_lowercase + string.digits, k=10)
```
squirrelsc: `util/__init__.py` has an implementation: `generate_random_chars`, you can change it to use `choices`

microsoft/testsuites/core/storage.py
```diff
@@ -345,6 +345,42 @@ def hot_add_disk_parallel_premium_ssd(self, log: Logger, node: Node) -> None:
...
+    def verify_azure_file_share_nfs(self, log: Logger, node: Node) -> None:
+        nfs = node.features[Nfs]
+        server_shared_dir = nfs.create()
+        storage_account_name = server_shared_dir.split("/")[0]
+        mount_nfs = f"{storage_account_name}.file.core.windows.net"
+        mount_dir = f"/mount/{server_shared_dir}"
+        max_retry = 10
+        retry = 0
+        mount_success = False
+        # retry if the dns name can't be resolved after creating shortly
```
squirrelsc: If so, the retry should be limited to dns name detection logic only, instead of capture on the whole logic.

microsoft/testsuites/core/storage.py
```diff
@@ -345,6 +345,42 @@ def hot_add_disk_parallel_premium_ssd(self, log: Logger, node: Node) -> None:
...
+        while retry < max_retry:
```
squirrelsc: Use timeout, instead of counting, it's easier to measure by time.

microsoft/testsuites/core/storage.py
```diff
@@ -345,6 +345,42 @@ def hot_add_disk_parallel_premium_ssd(self, log: Logger, node: Node) -> None:
             log, node, DiskType.PremiumSSDLRS, self.DEFAULT_DISK_SIZE_IN_GB
         )
 
+    @TestCaseMetadata(
+        description="""
+        This test case will verify mount azure nfs 4.1 on guest successfully.
```
squirrelsc: Where is it specified to 4.1? Maybe it's upgraded to 4.2 later?

microsoft/testsuites/core/storage.py
LiliDeng: double check, it has permission denied issue when wait for private connection take effect.

lisa/sut_orchestrator/azure/features.py

lisa/features/nfs.py
LiliDeng: add delete part

microsoft/testsuites/core/storage.py
LiliDeng: updated to use timeout

microsoft/testsuites/core/storage.py
LiliDeng: now not specify the version, by default it is 4.1

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1601,3 +1618,117 @@ def create_setting(
...
+        self._initialize_information(self._node)
+        self.storage_account_name: str = ""
+        self.file_share_name: str = ""
+
+    def create_share(self) -> None:
+        platform: AzurePlatform = self._platform  # type: ignore
+        node_context = self._node.capability.get_extended_runbook(AzureNodeSchema)
+        location = node_context.location
+        resource_group_name = self._resource_group_name
+        random_str = generate_random_chars(string.ascii_lowercase + string.digits, 10)
```
squirrelsc, line 1641: The candidates are the same, just overwrite the length.

microsoft/testsuites/core/storage.py
```diff
@@ -345,6 +346,46 @@ def hot_add_disk_parallel_premium_ssd(self, log: Logger, node: Node) -> None:
...
+                    mount_success = True
+                    break
+                except Exception as identifier:
+                    time.sleep(5)
+                    log.debug(
+                        f"{identifier.__class__.__name__}: {identifier}. Retry..."
+                    )
+                    continue
+            if not mount_success:
+                raise LisaException("fail to mount azure nfs")
```
squirrelsc: Save the last identifier, and include in the Exception. There would be some raw detail information in error message.

microsoft/testsuites/core/storage.py
```diff
@@ -345,6 +346,46 @@ def hot_add_disk_parallel_premium_ssd(self, log: Logger, node: Node) -> None:
...
+                        mount_nfs,
+                        server_shared_dir,
+                        mount_dir,
+                    )
+                    mount_success = True
+                    break
+                except Exception as identifier:
+                    time.sleep(5)
+                    log.debug(
+                        f"{identifier.__class__.__name__}: {identifier}. Retry..."
```
squirrelsc: 1. It looks like happening in the `setup` method. So the retry should be in the setup method. Another way is to add retry decorator on the `setup` method. It can save this whole logic.
2. The message doesn't specify where it's from. So it needs to see like waiting for something effect.

lisa/sut_orchestrator/azure/features.py
LiliDeng: storage account and share name can't be upper case

microsoft/testsuites/core/storage.py
LiliDeng: updated

microsoft/testsuites/core/storage.py
LiliDeng: move retry into mount, and update the exception message

squirrelsc (approved):

PR #2290, opened 2022-10-01
networksettings: rewriting device_statistics logic

sharsonia:


microsoft/testsuites/network/networksettings.py
```diff
@@ -549,8 +551,10 @@ def validate_device_msg_level_change(self, node: Node, log: Logger) -> None:
             Steps:
             1. Get all the device's statistics.
             2. Validate device statistics lists per queue statistics as well.
-            3. Run traffic using iperf3 and validate traffic is evenly distributed
-                among different VMBUS channel queues.
+            3. Run traffic using iperf3 and check stats for each device.
+            4. if the same queue (say queue #0) is always inactive among 5 test runs,
+                and the number of connections is >= 4 (total #queues), test should fail
```
squirrelsc: 1. What's "the number of connections"? I cannot find it in the code logic. If there are concept used, the term should be consistent.
2. I feel there is no difference between run 1 or 5 times. If it's not enough to run iperf3 with 10 seconds, let's just extend it to 20 seconds.
3. Because I don't understand the connection or queue here, so I cannot understand the reason it's considered failed.

sharsonia: 1. The number of connections here means the number of VMBUS channels i.e. the logic holds good if there are at least 4 tx/rx queue counters. For eg - 
                            tx_queue_0_packets/rx_queue_0_packets
                            tx_queue_1_packets/rx_queue_1_packets
                            tx_queue_2_packets/rx_queue_2_packets
                            tx_queue_3_packets/rx_queue_3_packets

2. Run iperf 5 times so we can see if for those 5 runs, one of the queues is completely starved off of packets.

squirrelsc: 1. It should be called the count of channels, instead of connections. But what if the count of channels is lower than 4?
2. I'm not clear whether they are different between 5 times run and 1 run with longer time.

sharsonia: 1. I can change the wording. if count of channels is lower than 4 then as per Haiyang (will fwd you email), even if one queue is tmtimg/rcvng 0 packets, that might still be ok since there are less number of queues/channels.

2. The difference is that - if more number of packets are going through then there is higher chance for each queue to tx/tx packets as compared to if there are smaller number of packets. Yeah we can increase time but I think what we would want to do is check the stats at some interval and make sure that each queue is tx'ing/rxlng some packets.

microsoft/testsuites/network/networksettings.py
```diff
@@ -563,75 +567,117 @@ def validate_device_statistics(self, environment: Environment, log: Logger) -> N
...
+                    final_starved_rxq = [
+                        rxq for rxq in rxq_list if rxq in prev_rxq_list
+                    ]
+
+            if (
+                not prev_starved_queues_per_device
+                and (not final_starved_txq)
+                and (not final_starved_rxq)
+            ):
+                raise PassedException(
```
squirrelsc: Doesn't raise PassException, if it's fully passed. The Pass exception is used for "partial" passed or pass with warning.

microsoft/testsuites/network/networksettings.py
```diff
@@ -563,75 +567,117 @@ def validate_device_statistics(self, environment: Environment, log: Logger) -> N
...
+                    per_tx_queue_packets = [
+                        v
+                        for (k, v) in device_stats.counters.items()
+                        if self._tx_queue_stats_regex.search(k)
+                    ]
+
+                    per_rx_queue_packets = [
+                        v
+                        for (k, v) in device_stats.counters.items()
+                        if self._rx_queue_stats_regex.search(k)
```
squirrelsc: Don't use `self` in test suite. The test suite may be running in multiple test cases together.

microsoft/testsuites/network/networksettings.py
sharsonia: hmm I have to make the test pass form here. How do I do it explicitly if not with passed exception?

microsoft/testsuites/network/networksettings.py
```diff
@@ -563,75 +567,117 @@ def validate_device_statistics(self, environment: Environment, log: Logger) -> N
...
+                    ]
+
+                if len(per_tx_queue_packets) and len(per_rx_queue_packets) < 4:
```
squirrelsc: It can skip earlier when get into the test cases by checking queue counts. It doesn't need to run one round.

microsoft/testsuites/network/networksettings.py
```diff
@@ -563,75 +567,117 @@ def validate_device_statistics(self, environment: Environment, log: Logger) -> N
...
-            assert_that(
-                (max(per_rx_queue_pkt_percent) - min(per_rx_queue_pkt_percent)),
-                f"{an_enabled}Statistics show traffic is not evenly distributed"
-                " among rx queues",
-            ).is_less_than_or_equal_to(0.5)
+            counter += 1
+
+        assert_that(
+            final_starved_txq,
+            f"The tx stats for queue/queues {final_starved_txq} is/are 0.",
```
squirrelsc: Imaging if the test case failed by this assertion, and the user won't look into the code. What can she/he know about the failure, and how should it be fixed?

microsoft/testsuites/network/networksettings.py
squirrelsc: It can be return directly. But make sure the comments is clear enough. The test logic is not easy to understand, when it's returning as pass in the middle of code.

microsoft/testsuites/network/networksettings.py
sharsonia: This would need deeper investigation but for now we would know that any particular queue is being starved. Ok I can add message like Need further investigation in driver code.

microsoft/testsuites/network/networksettings.py
```diff
@@ -560,78 +532,144 @@ def validate_device_msg_level_change(self, node: Node, log: Logger) -> None:
...
+        #  {'name': 'tx_queue_1_bytes', 'value': '0'},
+        #  {'name': 'rx_queue_1_packets', 'value': '1108'},
+        #  {'name': 'rx_queue_1_bytes', 'value': '1415269'},
+        #  {'name': 'rx_queue_1_xdp_drop', 'value': '0'},
+        #  {'name': 'tx_queue_2_packets', 'value': '0'},]
+        #  {'name': 'cpu0_vf_rx_packets', 'value': '0'},]
+        #  {'name': 'cpu0_vf_rx_bytes', 'value': '0'},]
+        #  {'name': 'cpu0_vf_tx_packets', 'value': '0'},]
+        #  {'name': 'cpu0_vf_tx_bytes', 'value': '0'},]
+        tx_queue_stats_regex = re.compile(r"tx_queue_(?P<name>[\d]+)_packets")
```
squirrelsc: Why it's moved into the method? It means the reg expressions will be compiled every time.

sharsonia: So earlier it was in the Testsuite class but then you mentioned that we should not be using self to access it. Hmm not sure then how would I use the regex in the methods. Can only use them either with self or using class name.
Both would be same thing

squirrelsc: Sorry, my bad. I haven't realized they are regex. I thought they are some dicts to hold intermedia data.

squirrelsc: The regex should be outside of a method.

microsoft/testsuites/network/networksettings.py
```diff
@@ -549,14 +557,15 @@ def validate_device_msg_level_change(self, node: Node, log: Logger) -> None:
...
+                and require further investigation.
         """,
         priority=2,
-        requirement=node_requirement(
-            node=schema.NodeSpace(
-                node_count=2,
-            )
+        requirement=simple_requirement(
+            min_count=2,
+            min_core_count=4,
```
squirrelsc, line 570: Add test case description to explain it needs at least 4 cores and why.

microsoft/testsuites/network/networksettings.py
squirrelsc: Not seeing changes to make it clearer.

squirrelsc (dismissed):

squirrelsc (approved):

PR #2292, opened 2022-10-03
nohup: add option to run command with nohup

mcgov:
Notes:
not guaranteed that the python class will understand that the connection was dropped and return correctly. 
Nohup also redirects stdout/stderr to a seperate file to this will break the stdout/stderr collection in the python class.

Nohup should only be used for command strings that MUST execute and have the potential to drop the ssh connection.


lisa/util/process.py
```diff
@@ -180,17 +189,23 @@ def _process_command(
...
                 # expand variables in posix mode
                 update_envs.clear()
         else:
-            if sudo and self._is_posix:
-                command = f"sudo {command}"
+            if self._is_posix:
+                if sudo:
+                    command = f"sudo {command}"
+                if nohup:
+                    command = f"nohup {command}"
```
squirrelsc, line 208: Does it support kill now? If yes, it can be enabled by default without an argument. And is it tested to work with sudo and shell well?

mcgov: We should test on a wide set of OS's, I have not yet so this PR is just for review at the moment. 
The full documentation notes some shells implement nohup as a builtin, so it's possible (like kill) that there are multiple paths to call it if the gnucoreutil binary and the shell are both available. That implies it may also be possible that the binary is not present and only the shell builtin is present, which would break this logic.

My understanding was that you should always call nohup first if you're using it.

mcgov: > Does it support kill now? 

It should, nohup only catches SIGHUP not SIGKILL etc

mcgov: re testing: this PR needs substantial testing across OS's. It's possible nohup is implemented as a shell builtin and not a binary on some systems, this will break the current logic since nohup is always called first to ensure the calling process isn't killed on disconnect.

> Does it support kill now? And is it tested to work with sudo and shell well?

The full documentation notes some shells implement nohup as a builtin, so it's possible (like kill) that there are multiple paths to call it if the gnucoreutil binary and the shell are both available. That implies it may also be possible that the binary is not present and only the shell builtin is present, which would break this logic.

>If yes, (kill supported) it can be enabled by default without an argument. 

No, nohup redirects stdin and stdout to a file so it is not good for any process except on where you don't care about the output as much. Callers will have to collect the nohup.out file to get their stdout after reconnecting if they need it.

microsoft/testsuites/core/hv_module.py
```diff
@@ -224,6 +224,7 @@ def reload_hyperv_modules(self, case_name: str, log: Logger, node: Node) -> None
             "ip link set eth0 down; ip link set eth0 up; dhclient eth0",
             sudo=True,
             shell=True,
+            nohup=True,
```
squirrelsc: The following lines used the result.stdout. So it looks like only need to add nohup for Ln224 `"ip link set eth0 down; ip link set eth0 up; dhclient eth0",`.

mcgov: Hm, the issue is that this has a potential to drop the connection during these reloads, I don't own this test so maybe it's better if the owner applies the fix on this one.

mcgov:
There's another edge case this doesn't solve where if a connection is dropped and the program is still active and attempting to send data back via stdout it can receive a broken pipe signal. This doesn't fix that situation, just the situation when the OS sends SIGHUP after a connection drop.

lisa/tools/hostname.py
```diff
@@ -18,4 +18,4 @@ def can_install(self) -> bool:
         return False
 
     def get_hostname(self) -> str:
-        return self.run("", shell=True).stdout.strip()
+        return self.run("", shell=True, force_run=True).stdout.strip()
```
mcgov: need to remove this, unrelated to nohup

lisa/util/process.py
```diff
@@ -180,17 +189,23 @@ def _process_command(
                 split_command = ["cmd", "/c", command]
             else:
                 split_command = []
+                if nohup:
```
squirrelsc, line 192: The order is different with the else. `sudo` and `nohup`, which one should be the first one?

mcgov: I did this too, these look reversed but they're not. The first is 'push front' since we're appending to an empty list, the other is push back since we're prefixing the string with a new command.

mcgov: nohup should be first though (except in cases where nohup is not present as a binary)

mcgov:
quick note: testing this PR but the test I'm using needs the other fixes along with this one.
Is there a pipeline to test a PR like this with a chaosmonkey type run? Or just ad-hoc

squirrelsc:
> quick note: testing this PR but the test I'm using needs the other fixes along with this one. Is there a pipeline to test a PR like this with a chaosmonkey type run? Or just ad-hoc

If a pipeline supports to specify the github ref, it can be used to test the fix. "single test case" is a good one.

LiliDeng:
execute_async can do the same thing? currently, if enable nohup, the stdout is always 'appending output to nohup.out'

mcgov:
> execute_async can do the same thing? currently, if enable nohup, the stdout is always 'appending output to nohup.out'

I don't think it does, nohup has to be called on the target node to catch SIGHUP. I don't know if execute async does this on it's own.

nohup does break stdout and stderr by appending to a file instead. This is because, in the event the program is needed and the client disconnects, the output needs to go somewhere to prevent a SIGPIPE broken pipe signal from killing the process. 

Callers who need that output will have to collect it after, with this PR. I think that for callers who need the output we could write a nohup tool that uses this feature and also collects the output. That is a bigger task that I think is out of scope for this PR.

mcgov:
I've tested this with DPDK and SRIOV, running network today.

LiliDeng:
@mcgov any progress?

mcgov:
Need to rebase, this has been stable in my testing though. I think it's ready for final review and merge

mcgov:
Oh I forgot this needs some input from @squirrelsc  or @LiliDeng, for non-remote shells it's not needed. It's only needed for ssh connections. I'm not sure how to differentiate though.

squirrelsc:
> Oh I forgot this needs some input from @squirrelsc or @LiliDeng, for non-remote shells it's not needed. It's only needed for ssh connections. I'm not sure how to differentiate though.

It's ok, we can regard local command is the same way. So it won't make the logic complicated.

mcgov:
I think this is ready to merge, I've been testing it for a week and haven't noticed any issues in the sriov or dpdk tests. Can we merge this on a Monday and see if there are any regressions or something?

mcgov:
@squirrelsc want to try and merge this week?

squirrelsc (approved):

squirrelsc:
@LiliDeng Notice, it's a foundation change, may have regressions.

PR #2293, opened 2022-10-03
DPDK: fix regex iterator assumption, drop unreliable samples

mcgov:
Two small fixes:
- First and last samples have bad data, since we're looking for an order of magnitude difference they can mess up the data sometimes. Discard first and last sample before calculating the mean of the data.

- Fix an incorrect assumption about the boolean value of an empty iterator. Cast to list first, makes other logic cleaner.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -387,7 +387,8 @@ class DpdkTestpmd(Tool):
             )
         )
         cast_to_ints = list(map(int, matches))
-        return _discard_first_zeroes(cast_to_ints)
+        cast_to_ints = _discard_first_zeroes(cast_to_ints)
+        return _discard_first_and_last_sample(cast_to_ints)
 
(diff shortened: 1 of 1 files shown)


squirrelsc:
> Two small fixes:
> 
> * First and last samples have bad data, since we're looking for an order of magnitude difference they can mess up the data sometimes. Discard first and last sample before calculating the mean of the data.
> * Fix an incorrect assumption about the boolean value of an empty iterator. Cast to list first, makes other logic cleaner.

Please add it to commit message. It's easy to search in code history.

mcgov:
> > Two small fixes:
> > 
> > * First and last samples have bad data, since we're looking for an order of magnitude difference they can mess up the data sometimes. Discard first and last sample before calculating the mean of the data.
> > * Fix an incorrect assumption about the boolean value of an empty iterator. Cast to list first, makes other logic cleaner.
> 
> Please add it to commit message. It's easy to search in code history.

added :)

squirrelsc (approved):

PR #2294, opened 2022-10-03
SRIOV: allow old sriov switch behavior that doesn't kill connection

mcgov:
recent change in switch_sriov ended up killing testpmd during the rescind tests, which breaks their logic.
- Allow tests to skip calling node.close() when switching SRIOV, add use in dpdk sriov rescind tests.
- Add DPDK call to node.close() at the end of the test to cleanup.

### DIFF
--- a/lisa/features/network_interface.py
+++ b/lisa/features/network_interface.py
@@ -21,7 +21,9 @@ class NetworkInterface(Feature):
     def enabled(self) -> bool:
         return True
 
-    def switch_sriov(self, enable: bool, wait: bool = True) -> None:
+    def switch_sriov(
+        self, enable: bool, wait: bool = True, reset_connections: bool = True
+    ) -> None:
         raise NotImplementedError
 
     def is_enabled_sriov(self) -> bool:
--- a/lisa/sut_orchestrator/aws/features.py
+++ b/lisa/sut_orchestrator/aws/features.py
@@ -167,7 +167,9 @@ class NetworkInterface(AwsFeatureMixin, features.NetworkInterface):
             raise LisaException(f"fail to find primary nic for vm {self._node.name}")
         return nic
 
-    def switch_sriov(self, enable: bool, wait: bool = True) -> None:
+    def switch_sriov(
+        self, enable: bool, wait: bool = True, reset_connections: bool = True
+    ) -> None:
         aws_platform: AwsPlatform = self._platform  # type: ignore
         instance = boto3.resource("ec2").Instance(self._instance_id)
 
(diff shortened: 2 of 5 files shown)


squirrelsc:
> recent change in switch_sriov ended up killing testpmd during the rescind tests, which breaks their logic.
> 
> * Allow tests to skip calling node.close() when switching SRIOV, add use in dpdk sriov rescind tests.
> * Add DPDK call to node.close() at the end of the test to cleanup.

What's the error, if the ssh connection is reset every time?

mcgov:
> > recent change in switch_sriov ended up killing testpmd during the rescind tests, which breaks their logic.
> > 
> > * Allow tests to skip calling node.close() when switching SRIOV, add use in dpdk sriov rescind tests.
> > * Add DPDK call to node.close() at the end of the test to cleanup.
> 
> What's the error, if the ssh connection is reset every time?

The testpmd program dies mid-test so it can't record the failover occurring. The idea is to check the failover occurs *while DPDK is still active* so calling node.close and killing that process breaks the test.

squirrelsc:
> > What's the error, if the ssh connection is reset every time?
> 
> The testpmd program dies mid-test so it can't record the failover occurring. The idea is to check the failover occurs _while DPDK is still active_ so calling node.close and killing that process breaks the test.

Not sure if it's stable enough. Because the link may be broken, that's why the node.close is added to make it stable. If the output is needed, it may be redirected to some file, and make sure it's not lost.

mcgov:
> > > What's the error, if the ssh connection is reset every time?
> > 
> > 
> > The testpmd program dies mid-test so it can't record the failover occurring. The idea is to check the failover occurs _while DPDK is still active_ so calling node.close and killing that process breaks the test.
> 
> Not sure if it's stable enough. Because the link may be broken, that's why the node.close is added to make it stable. If the output is needed, it may be redirected to some file, and make sure it's not lost.

I don't think that will work, node.close is killing all the shells that are active, so wouldn't it still kill the other process?

squirrelsc (approved):

squirrelsc:
> > > > What's the error, if the ssh connection is reset every time?
> > > 
> > > 
> > > The testpmd program dies mid-test so it can't record the failover occurring. The idea is to check the failover occurs _while DPDK is still active_ so calling node.close and killing that process breaks the test.
> > 
> > 
> > Not sure if it's stable enough. Because the link may be broken, that's why the node.close is added to make it stable. If the output is needed, it may be redirected to some file, and make sure it's not lost.
> 
> I don't think that will work, node.close is killing all the shells that are active, so wouldn't it still kill the other process?

It can work with nohup change together.

PR #2296, opened 2022-10-03
VPP: use service tool

mcgov:
VPP: switch to using the service tool to restart the VPP service. This was causing issues on OS's which don't support the 'service servicename restart' syntax

Convert Unsupported Distro to Skip since we can install on any supported distros that VPP supports.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkvpp.py
+++ b/microsoft/testsuites/dpdk/dpdkvpp.py
@@ -8,7 +8,7 @@ from assertpy import assert_that
 from lisa.executable import Tool
 from lisa.operating_system import Debian, Fedora, Suse
 from lisa.tools import Gcc, Git, Make, Modprobe, Service
-from lisa.util import SkippedException
+from lisa.util import SkippedException, UnsupportedDistroException
 
 
 class DpdkVpp(Tool):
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -92,7 +87,7 @@ def _install(self) -> bool:
         elif isinstance(node.os, Fedora) or isinstance(node.os, Suse):
             pkg_type = "rpm"
         else:
-            raise UnsupportedDistroException(self.node.os)
+            raise SkippedException(self.node.os)
```
squirrelsc: The message won't be clear, if skip like this way. Try `SkippedException(UnsupportedDistroException(self.node.os, "<more clarify description here>"))`

squirrelsc: The SkippedException doesn't produce meaningful exception message by unsupported os. Please try the example to wrap the unsupported exception.

mcgov: Ohhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh I missed the wrapping, will fix

mcgov: will squish commits before merge.

squirrelsc (approved):

PR #2297, opened 2022-10-03
bugfix: modprobe: remove single quote

mcgov:
Bugfix: Single quoting messes with the shell, remove them


mcgov:
fixes an issue with https://github.com/microsoft/lisa/pull/2295
I missed this in testing because I just happened to not need the reset call during my test run. Further testing confirms this fix is needed.

squirrelsc (approved):

PR #2298, opened 2022-10-03
DPDK: node.close() before hv_netvsc reset in rescind kill

mcgov:
node.close() seems to be more reliable for killing the unresponsive process, so attempt node.close before resetting the network interface. after_case already performs this reset anyway so doing it twice it pointless if we can avoid it.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -214,7 +214,9 @@ class DpdkTestpmd(Tool):
                 vdev_flags = f"iface={nic.upper},force=1"
             else:
                 vdev_name = "net_failsafe"
-                vdev_flags = f"dev({nic.pci_slot}),dev(net_tap0,iface={nic.upper})"
+                vdev_flags = (
+                    f"dev({nic.pci_slot}),dev(net_tap0,iface={nic.upper},force=1)"
+                )
             if nic.bound_driver == "hv_netvsc":
                 vdev_info += f'--vdev="{vdev_name}{vdev_id},{vdev_flags}" '
             elif nic.bound_driver == "uio_hv_generic":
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -356,6 +356,10 @@ def kill_previous_testpmd_command(self) -> None:
             self.node.log.debug(
                 "Testpmd is not responding to signals, attempt reload of hv_netvsc."
             )
+            self.node.close()  # reset node connection, should kill testpmd
```
squirrelsc: 1. Rather than relies on close to end the testpmd, it's better to kill it explicitly. Considering we're enabling nohup, it may make mistake on assumptions on the SSH close behavior.
3. The comment should be in separated lines, when I learned programming.

mcgov: > 1. Rather than relies on close to end the testpmd, it's better to kill it explicitly.

This function does `kill` testpmd explicitly first (two lines above). These other calls are backups since testpmd is unresponsive sometimes. Adding these are the results of testing which shows `kill` is not sufficient sometimes. At this point in the test, the goal is just to kill testpmd and colllect it's output at any cost. These calls go in order of how big 'the hammer' used to kill testpmd is. 1. kill, 2. close all node connections (new) and reconnect, 3. reset the network driver entirely. Since the after_case call now resets the network driver it seems redundant to do it twice, so the smaller hammer of resetting connections seemed like a nice (and quick) alternative.

> Considering we're enabling nohup, it may make mistake on assumptions on the SSH close behavior.

Testpmd won't be run with nohup and there shouldn't be any other processes running with nohup when this call is made. nohup should be sort of rare, only for calls which could leave the node in a state with no network connection. Typical example use is when you have a long running process (linux kernel build) and are on an unreliable network connection.

> 2. The comment should be in separated lines, when I learned programming.

ok

squirrelsc: Good to know the detail. Could you add the key points of `close()` in comments to avoid confusion?

mcgov: Yep! Added some debug logging and extra comments to make this explicit. The previous version wasn't clear enough.

squirrelsc (approved):

PR #2299, opened 2022-10-04
modprobe: only reload network interfaces for network drivers

mcgov:
hv_netvsc needs a special case, since reloading it has the potential
 to leave the node without a network connection if things go wrong.

For most modules this isn't neccesary, so just special case netvsc.

### DIFF
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -23,7 +23,7 @@ class Modprobe(Tool):
         self.node.execute(
             "modprobe -r hv_netvsc; modprobe hv_netvsc; "
             "ip link set eth0 down; ip link set eth0 up;"
-            "dhclient -r; dhclient'",
+            "dhclient -r eth0; dhclient eth0",
             sudo=True,
             shell=True,
         )
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -16,6 +16,18 @@ class Modprobe(Tool):
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._command = "modprobe"
 
+    # hv_netvsc needs a special case, since reloading it has the potential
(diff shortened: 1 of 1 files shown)


squirrelsc (dismissed):

lisa/tools/modprobe.py
```diff
@@ -16,6 +16,18 @@ def _check_exists(self) -> bool:
...
         self._command = "modprobe"
 
+    # hv_netvsc needs a special case, since reloading it has the potential
+    # to leave the node without a network connection if things go wrong.
+    def _reload_hv_netvsc(self) -> None:
+        # These commands must be sent together, bundle them up as one line
+        self.node.execute(
+            "modprobe -r hv_netvsc; modprobe hv_netvsc; "
+            "ip link set eth0 down; ip link set eth0 up;"
+            "dhclient -r; dhclient'",
```
LiliDeng: extra single quotation marks after dhclient

LiliDeng: use 'dhclient -r eth0; dhclient eth0', it can avoid the vf nic getting ip address issue when use v5 size.

mcgov: Good catch, updated

lisa/tools/modprobe.py
```diff
@@ -97,15 +109,18 @@ def reload(
...
-                )
+
+                # hv_netvsc reload requires resetting the network interface
+                if mod_name == "hv_netvsc":
+                    # handle special case
+                    self._reload_hv_netvsc()
+                else:
+                    # execute the command for regular non-network modules
+                    self.node.execute(
+                        f"'modprobe -r {mod_name}; modprobe {mod_name};",
```
LiliDeng: extra single quotation marks before modprobe

LiliDeng (approved):

PR #2301, opened 2022-10-06
add parameter SkippedStorageAccountNames

LiliDeng:


squirrelsc (approved):

squirrelsc:
Where the V2 is needed?

LiliDeng:
> Where the V2 is needed?

it is used in scanning azure resource usage pipeline, pipeline is always stuck when scanning the storage account which has many files can not cleaned up, e.g. storage account for storing history marketplace image, storage account for storing logs.

LiliDeng:
> Where the V2 is needed?

it is used in scanning azure resource usage pipeline, pipeline is always stuck when scanning the storage account which has many files which are not cleaned up, e.g.

PR #2304, opened 2022-10-09
[new case] drm driver validation

LiliDeng:


microsoft/testsuites/display/modetest.py
```diff
@@ -0,0 +1,79 @@
...
+from typing import cast
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix, Redhat, Ubuntu
+from lisa.tools.git import Git
+from lisa.util import LisaException
+
+
+class Modetest(Tool):
+    repo = "https://github.com/grate-driver/libdrm"
```
squirrelsc, line 10: Let's stick to a long term supported branch to prevent breaking changes.

microsoft/testsuites/display/modetest.py
```diff
@@ -0,0 +1,79 @@
...
+                "libtool",
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/xorg-x11-util-macros-1.19.3-4.el9.noarch.rpm",  # noqa: E501
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/ninja-build-1.10.2-6.el9.x86_64.rpm",  # noqa: E501
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/meson-0.58.2-1.el9.noarch.rpm",  # noqa: E501
+            ]
+        else:
+            raise LisaException(
+                f"tool {self.command} can't be installed in distro {self.node.os.name}."
+            )
+        posix_os.install_packages(list(package_list))
```
squirrelsc: It's a list already, doesn't need convert to list.

microsoft/testsuites/display/modetest.py
```diff
@@ -0,0 +1,79 @@
...
+                "autoconf",
+                "automake",
+                "libpciaccess-devel.x86_64",
+                "libtool",
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/xorg-x11-util-macros-1.19.3-4.el9.noarch.rpm",  # noqa: E501
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/ninja-build-1.10.2-6.el9.x86_64.rpm",  # noqa: E501
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/meson-0.58.2-1.el9.noarch.rpm",  # noqa: E501
+            ]
+        else:
+            raise LisaException(
```
squirrelsc: Use Unsupported OS exception.

microsoft/testsuites/display/modetest.py
LiliDeng: I didn't see such branches, and no tags, the latest change for this rero happened on 4 years ago.

microsoft/testsuites/display/modetest.py
LiliDeng: without convert to list, this is the error
```Argument 1 to "install_packages" of "Posix" has incompatible type "List[str]"; expected "Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]]"```

Comment (squirrelsc) [reply microsoft/testsuites/display/modetest.py]: Please either update the install_packages to accept `List[str]` as a type.

**Review by squirrelsc — DISMISSED** [review]

File: lisa/operating_system.py [hunk 4]
```diff
@@ -323,7 +323,15 @@ def get_kernel_information(self, force_run: bool = False) -> KernelInformation:
 
     def install_packages(
         self,
-        packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]],
+        packages: Union[
+            str,
+            Tool,
+            Type[Tool],
+            List[str],
```
avylove: Is `List` the correct type here? Typically `List` would only be used if a functionality specific to `list` is required. If the only requirement is to be an argument to `str.join()`, then it would be `Iterable`. That said, I'd probably go with `Sequence` to provide more flexibility is the subclassed methods.

squirrelsc (dismissed):

lisa/operating_system.py
```diff
@@ -499,15 +509,24 @@ def _get_information(self) -> OsInformation:
...
-        self, packages: Union[str, Tool, Type[Tool], List[Union[str, Tool, Type[Tool]]]]
+        self,
+        packages: Union[
+            str,
+            Tool,
+            Type[Tool],
+            Sequence[str],
+            Sequence[Tool],
+            Sequence[Type[Tool]],
+            Sequence[Union[str, Tool, Type[Tool]]],
```
avylove, line 514: Doesn't the union cover the previous 3 lines?

lisa/operating_system.py
```diff
@@ -499,15 +509,24 @@ def _get_information(self) -> OsInformation:
...
-
-        assert isinstance(packages, list), f"actual:{type(packages)}"
-        for item in packages:
-            package_names.append(self.__resolve_package_name(item))
+        if isinstance(packages, Iterable):
+            if isinstance(packages, str) or isinstance(packages, Tool):
+                packages = [packages]
+            assert isinstance(packages, list), f"actual:{type(packages)}"
+            for item in packages:
+                package_names.append(self.__resolve_package_name(item))
```
avylove:
```suggestion
       if isinstance(packages, (str, Tool)):
            package_names = [self.__resolve_package_name(packages)]
       else:
           package_names = [self.__resolve_package_name(item) for item in packages]
```

microsoft/testsuites/display/modetest.py
```diff
@@ -0,0 +1,77 @@
...
+                "automake",
+                "libpciaccess-devel.x86_64",
+                "libtool",
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/xorg-x11-util-macros-1.19.3-4.el9.noarch.rpm",  # noqa: E501
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/ninja-build-1.10.2-6.el9.x86_64.rpm",  # noqa: E501
+                "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/meson-0.58.2-1.el9.noarch.rpm",  # noqa: E501
+            ]
+        else:
+            raise UnsupportedDistroException(self.node.os)
+        posix_os.install_packages(package_list)
```
avylove: package_list shouldn't be a list if you're not planing on modifying it, but you also don't need to use a variable if all you're doing is passing it to one function.
Does mypy complain if you do it like this? I would think the conditional should give mypy context to know the object will have the `install_packages()` method.

```suggestion
        if isinstance(self.node.os, Redhat):
            self.node.os.install_packages(
                (
                    "git",
                     "make",
                     "autoconf",
                     "automake",
                     "libpciaccess-devel.x86_64",
                     "libtool",
                     "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/xorg-x11-util-macros-1.19.3-4.el9.noarch.rpm",  # noqa: E501
                     "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/ninja-build-1.10.2-6.el9.x86_64.rpm",  # noqa: E501
                     "http://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/meson-0.58.2-1.el9.noarch.rpm",  # noqa: E501
                )
            )
        else:
            raise UnsupportedDistroException(self.node.os)
```

microsoft/testsuites/display/modetest.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _install_from_src(self) -> None:
+        self._install_dep_packages()
+        tool_path = self.get_tool_path()
+        git = self.node.tools[Git]
+        git.clone(self.repo, tool_path)
```
avylove: No need to use a variable since this is a simple lookup and you're only using it once

```suggestion
        self.node.tools[Git].clone(self.repo, tool_path)
```

microsoft/testsuites/display/modetest.py
```diff
@@ -0,0 +1,77 @@
...
+    def is_status_connected(self, driver_name: str) -> bool:
+        cmd_result = self.run(f"-M {driver_name}", sudo=True, shell=True)
+        # output segment
+        # Connectors:
+        # id encoder status          name             size (mm)       modes   encoders
+        # 31 35      connected       Virtual-1        0x0             24      35
+        for line in cmd_result.stdout.splitlines():
+            if "connected" in line:
+                return True
+        return False
```
avylove:
```suggestion
        return any("connected" in line for line in cmd_result.stdout.splitlines())
```

microsoft/testsuites/display/drm.py
```diff
@@ -67,6 +68,41 @@ def verify_dri_node(self, node: Node, log: Logger) -> None:
...
+
+        Step,
+        1. Get dmesg output.
+        2. Check no 'Unable to send packet via vmbus' shown up in dmesg output.
+        """,
+        priority=2,
+    )
+    def verify_no_error_output(self, node: Node, log: Logger) -> None:
+        dmesg = node.tools[Dmesg]
+        assert_that(dmesg.get_output()).described_as(
```
avylove: No need to use a variable for a lookup that is used once

```suggestion
        assert_that(node.tools[Dmesg].get_output()).described_as(
```

microsoft/testsuites/display/drm.py
```diff
@@ -67,6 +68,41 @@ def verify_dri_node(self, node: Node, log: Logger) -> None:
...
+
+        Step,
+        1. Install tool modetest.
+        2. Verify the status return from modetest is connected.
+        """,
+        priority=2,
+    )
+    def verify_connection_status(self, node: Node, log: Logger) -> None:
+        modetest = node.tools[Modetest]
+        is_status_connected = modetest.is_status_connected("hyperv_drm")
```
avylove: Same as above

```suggestion
        is_status_connected = node.tools[Modetest].is_status_connected("hyperv_drm")
```

lisa/operating_system.py
squirrelsc: The mypy doesn't think it's right. When passing one type list, mypy raise errors. So add them all here for convenience.

avylove: I just tried it and mypy didn't error. Are you still getting an error?

squirrelsc: That's cool, so we can remove Ln517-519. @LiliDeng

lisa/operating_system.py
LiliDeng: @avylove here is the mypy error, if change like this
```Item "type" of "Union[Type[Tool], Sequence[Union[str, Tool, Type[Tool]]]]" has no attribute "__iter__" (not iterable)```

squirrelsc (approved):

PR #2305, opened 2022-10-10
libvirt: save libvirtd logs during environment delete

anirudhrb:
libvirtd logs are useful for debugging libvirt related issues that might happen while executing testcases. So, retreive the logs using journalctl and write them to the host node's log path. Also introduce a new tool for working with the `journalctl` command.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -26,7 +26,7 @@ from lisa.feature import Feature
 from lisa.node import Node, RemoteNode, local_node_connect
 from lisa.operating_system import CBLMariner
 from lisa.platform_ import Platform
-from lisa.tools import Iptables, Ls, QemuImg, Uname
+from lisa.tools import Iptables, Journalctl, Ls, QemuImg, Uname
 from lisa.util import LisaException, constants, get_public_key_data
 from lisa.util.logger import Logger, filter_ansi_escape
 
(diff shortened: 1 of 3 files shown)


pupacha (dismissed):

LiliDeng (dismissed):

lisa/tools/journalctl.py
```diff
@@ -0,0 +1,23 @@
...
+class Journalctl(Tool):
+    @property
+    def command(self) -> str:
+        return "journalctl"
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def logs_for_unit(self, unit_name: str) -> str:
+        result = self.run(
```
squirrelsc, line 16: This may generate a lot of log in LISA and flood other logs. Save the output to the remote working path of tool, and copy it back.

squirrelsc:
@cwize1 A straightforward change. In case any feedback here.

lisa/tools/journalctl.py
squirrelsc: You can pass in the test case local working path, so the copy back can be written in this tool for convenient.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -161,6 +161,10 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
         self._delete_nodes(environment, log)
```
cwize1, line 161: FYI: I believe `_delete_environment` is not called if `_deploy_environment` fails (throws an exception). For example, `_deploy_nodes` calls `_delete_nodes` in an `except` block. (Sorry, I should have remembered this for your previous change.)

lisa/tools/journalctl.py
cwize1: I believe an alternative might be to pass `no_info_log=True` and `no_debug_log=True` here.

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: `_delete_environment` is being called when `_deploy_environment` throws an exception:
https://github.com/microsoft/lisa/blob/e4fc638c02fc27ce51439712ac1568a33c9f7cdc/lisa/runners/lisa_runner.py#L269-L295

lisa/tools/journalctl.py
anirudhrb: no_info_log is True by default. Added `no_debug_log=True`.

cwize1 (approved):

squirrelsc (approved):

PR #2306, opened 2022-10-10
Add stacktrace for runner exception

dsrivastavv:
Add stacktrace for runner exception

### DIFF
--- a/lisa/runner.py
+++ b/lisa/runner.py
@@ -220,7 +220,7 @@ class RootRunner(Action):
 
             self._start_loop()
         except Exception as identifier:
-            self._log.info(f"canceling runner due to exception: {identifier}")
+            self._log.exception(f"canceling runner due to exception: {identifier}")
             cancel()
         finally:
             self._cleanup()


lisa/runner.py
```diff
@@ -220,7 +221,10 @@ async def start(self) -> None:
 
             self._start_loop()
         except Exception as identifier:
-            self._log.info(f"canceling runner due to exception: {identifier}")
+            self._log.info(
```
squirrelsc: Can you try if _log.exception works?

dsrivastavv: updated. Verified that exception works

lisa/runner.py
```diff
@@ -220,7 +220,7 @@ async def start(self) -> None:
 
             self._start_loop()
         except Exception as identifier:
-            self._log.info(f"canceling runner due to exception: {identifier}")
+            self._log.exception(f"canceling runner due to exception: {identifier}")
```
squirrelsc, line 223: I think the exception message should be printed by the `log.exception` already. It doesn't need to be included in the custom message again.

dsrivastavv: It is with the following format:
```
canceling runner due to exception: {identifier}
Stacktrace
{identifier}
```

I think we can leave it here since it provides more context to the message `canceling runner due to exception` without having to scroll down the entire stacktrace. Please let me know if you feel otherwise

LiliDeng: @somil55 any update?

squirrelsc:
fixed by #2318

PR #2307, opened 2022-10-11
Add systemctl to tools

rakshaa2000:
- Add `systemctl.py` to `lisa/tools`
- Add "Systemctl" to `lisa/tools/__init__.py`

### DIFF
--- a/lisa/tools/__init__.py
+++ b/lisa/tools/__init__.py
@@ -81,6 +81,7 @@ from .strace import Strace
 from .stress_ng import StressNg
 from .swap import Swap
 from .sysctl import Sysctl
+from .systemctl import Systemctl
 from .tar import Tar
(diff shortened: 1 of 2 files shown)


LiliDeng:
Thank you! We already has this tool, please refer to https://github.com/microsoft/lisa/blob/main/lisa/tools/service.py

LiliDeng:
Thank you! We already has this tool, please refer to https://github.com/microsoft/lisa/blob/main/lisa/tools/service.py

mcgov:
Some tools are metatools which wrap multiple binaries depending on the system, so before implementing a new tool it's best to text search for uses of the binary to see if it's already bundled somewhere. Occasionally you find people calling executables directly with node.execute, so you can fix their uses if the tool isn't already present and then test to make sure nothing is broken.

PR #2312, opened 2022-10-14
Update libvirt-python version

pupacha:
Current libvirt-python library had a memory leak bug which is fixed with following commit.
https://gitlab.com/libvirt/libvirt-python/-/commit/9716b459

The next release with the fix is in a month. So, this change is temporary and will be updated to latest after its release.

CC: @anirudhrb  @cwize1


pupacha:
PR checks on windows are failing because `make setup` fails to install libvirt-python from source. Throws "pkg-config binary is required to compile libvirt-python". There's already an issue filed with libvirt-python on this - https://gitlab.com/libvirt/libvirt-python/-/issues/7
@squirrelsc  @LiliDeng  Is there any way to bypass the libvirt-python installation on windows and install it only on linux ?
`platform = 'linux'` is already there in pyproject.yml but still windows tried to install it.

squirrelsc:
> PR checks on windows are failing because `make setup` fails to install libvirt-python from source. Throws "pkg-config binary is required to compile libvirt-python". There's already an issue filed with libvirt-python on this - https://gitlab.com/libvirt/libvirt-python/-/issues/7 @squirrelsc @LiliDeng Is there any way to bypass the libvirt-python installation on windows and install it only on linux ? `platform = 'linux'` is already there in pyproject.yml but still windows tried to install it.

Maybe it doesn't work with git installation. Can you try a previous version, or beta version?

pupacha:
Closing the PR. Will wait for the next release of libvirt-python as git source installation is failing for windows.

PR #2314, opened 2022-10-16
[fix test issues] extend texinfo installation time, only collect cmp interrupts to compare etc

LiliDeng:
details see commits message

### DIFF
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -25,6 +25,7 @@ from lisa.operating_system import (
     Debian,
     DebianRepositoryInfo,
     Fedora,
+    FreeBSD,
     Oracle,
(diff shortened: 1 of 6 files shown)


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -885,10 +886,17 @@ def verify_client_active_interval(self, node: Node) -> None:
     def verify_no_pre_exist_users(self, node: Node) -> None:
         current_user = node.tools[Whoami].get_username()
         cat = node.tools[Cat]
+        shadow_file = "/etc/shadow"
```
squirrelsc: This should be in the `else`

LiliDeng: updated

squirrelsc (approved):

PR #2315, opened 2022-10-17
[fix test issues] extend time to do unattended-upgrade

LiliDeng:


microsoft/testsuites/core/dns.py
```diff
@@ -48,13 +48,28 @@ def verify_dns_name_resolution_after_upgrade(self, node: Node) -> None:
             )
             if 0 != cmd_result.exit_code:
                 node.os.install_packages("unattended-upgrades")
+            if not isinstance(node.os, Ubuntu):
```
squirrelsc: Does it cover Debian only? If so, use `type(node.os) == Debian)`

microsoft/testsuites/core/dns.py
```diff
@@ -48,13 +48,28 @@ def verify_dns_name_resolution_after_upgrade(self, node: Node) -> None:
...
+                if 10 == node.os.information.version.major:
```
squirrelsc: 1. Use version compare pattern like `node.os.information.version >= 10.0.0`.
2. As above pattern, always assume higher version use the same mechanism, until we found it's wrong.
3. Always use an `else` to handle rest versions.

microsoft/testsuites/core/dns.py
```diff
@@ -48,13 +48,28 @@ def verify_dns_name_resolution_after_upgrade(self, node: Node) -> None:
...
+                    node.execute(
+                        "mkdir -p /var/cache/apt/archives/partial",
+                        sudo=True,
+                        shell=True,
+                        expected_exit_code=0,
+                        expected_exit_code_failure_message=(
+                            "fail to make folder /var/cache/apt/archives/partial"
+                        ),
+                    )
+                if 9 == node.os.information.version.major:
```
squirrelsc: `elif`

microsoft/testsuites/core/dns.py
LiliDeng: yes, updated.

microsoft/testsuites/core/dns.py
LiliDeng: updated

microsoft/testsuites/core/dns.py
LiliDeng: updated

squirrelsc (approved):

PR #2316, opened 2022-10-18
CH Perf Metrics Tests : Changes to add assert and replace stdout with stderr

smit-gardhariya:
These changes are made to make LISA test case failed is anyone of the CH Perf metrics are failing using ASSERT.
Also pushing stderr instead of stdout if testcase fails


LiliDeng (dismissed):

pupacha (changes requested):
We should also have the raw output of each Perf metrics testcase as a log file. This will help in debugging incase of failures.

squirrelsc (dismissed):

smit-gardhariya:
> We should also have the raw output of each Perf metrics testcase as a log file. This will help in debugging incase of failures.

Implemented

pupacha (approved):

LiliDeng (approved):

PR #2317, opened 2022-10-18
[fix issues] iperf3 run issues

LiliDeng:


lisa/tools/iperf3.py
```diff
@@ -131,6 +132,8 @@ def run_as_server_async(
         process = self.node.execute_async(
             f"{self.command} {cmd}", shell=True, sudo=True
         )
+        if 0 == len(self.node.tools[Pgrep].get_processes(self.command)):
```
squirrelsc: How about to call process.wait_result, if it happened. So there have more information get from results, like exitcode and stdout.

lisa/operating_system.py
```diff
@@ -806,11 +806,14 @@ def _install_packages(
         add_args = self._process_extra_package_args(extra_args)
         command = (
             f"DEBIAN_FRONTEND=noninteractive apt-get {add_args} "
-            f"-y install {' '.join(packages)}"
+            f"-y --force-yes install {' '.join(packages)}"
         )
         if not signed:
             command += " --allow-unauthenticated"
         self.wait_running_package_process()
+        self._node.execute(
```
squirrelsc: Does it need to be executed every time?

lisa/operating_system.py
```diff
@@ -806,11 +806,14 @@ def _install_packages(
         add_args = self._process_extra_package_args(extra_args)
         command = (
             f"DEBIAN_FRONTEND=noninteractive apt-get {add_args} "
-            f"-y install {' '.join(packages)}"
+            f"-y --force-yes install {' '.join(packages)}"
```
squirrelsc: Will it bypass the version conflicts? Some version conflicts may raise errors, so we can know them. If it's force installed, it will be hidden.

LiliDeng: I've tested the changes in this file, still encounter below issue in 1/20 rate, need update and figure out effective way.
```
failed. AssertionError: [Failed to install ['iperf3'], please check the package name and repo are correct or not.
E: Unable to locate package iperf3

get unexpected exit code on cmd ['sudo', 'sh', '-c', 'DEBIAN_FRONTEND=noninteractive apt-get  -y --force-yes install iperf3 --allow-unauthenticated']] Expected <[0]> to contain item <100>, but did not
```

lisa/tools/iperf3.py
LiliDeng: if call process.wait_result here, it is not run as async mode?

squirrelsc: Call it in the `if` block, so it's only called when the process doesn't exist, it means it's failed already. That should be very fast.

lisa/operating_system.py
squirrelsc: The message says, "Unable to locate package iperf3". It shouldn't be related to the force parameter.

LiliDeng: I'll remove these changes in this PR, will submit one to fix it individually.

lisa/tools/iperf3.py
LiliDeng: updated

LiliDeng: check process existing or not is not a good way, since the iperf3 may be launched multiple times. not sure check process is running or not is a good way or not, running testing now.

LiliDeng: just keep fix part for client, need figure out a way to fix issue when output is json format.

squirrelsc (approved):

PR #2320, opened 2022-10-19
Update link for linux poetry installation to canonical link

contre:
When installing on Linux, poetry recommends using the https://install.python-poetry.org link for downloading and installing poetry instead of the raw.githubusercontent.com link that is currently being used.

### DIFF
--- a/docs/installation_linux.rst
+++ b/docs/installation_linux.rst
@@ -57,7 +57,7 @@ Poetry is used to manage Python dependencies of LISA.
 
 .. code:: bash
 
-   curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3 -
+   curl -sSL https://install.python-poetry.org | python3 -
 
 After running this, you should see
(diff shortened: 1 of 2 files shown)


squirrelsc (dismissed):

docs/installation_linux.rst
```diff
@@ -57,7 +57,7 @@ Poetry is used to manage Python dependencies of LISA.
 
 .. code:: bash
 
-   curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python3 -
+   curl -sSL https://install.python-poetry.org | python3 -
```
dsrivastavv, line 60: Thanks Richard! We will also need to make this change in the CI. Can you please update the CI files?

dsrivastavv: File is here: https://github.com/microsoft/lisa/blob/main/.github/workflows/continuous-integration-workflow.yml

contre: Done.

dsrivastavv (approved):
LGTM!

squirrelsc (approved):

PR #2323, opened 2022-10-20
Add journalctl tool to LISA

rakshaa2000:
- Add `journalctl.py` into `lisa/tools`
- Add `"Journalctl"` to `lisa/tools/__init__.py`


anirudhrb:
There's already a journalctl tool: https://github.com/microsoft/lisa/blob/main/lisa/tools/journalctl.py

PR #2324, opened 2022-10-21
[fix regression] set value for nodes_raw to avoid regression issue

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1042,12 +1042,15 @@ def load_environment(
...
     # create mock environment from environments
     environment_runbook = schema.Environment()
+    if environment_runbook.nodes_raw is None:
+        environment_runbook.nodes_raw = []
 
     compute_client = get_compute_client(platform)
     vms = compute_client.virtual_machines.list(resource_group_name)
     for vm in vms:
         node_schema = schema.RemoteNode(name=vm.name)
         environment_runbook.nodes.append(node_schema)
```
squirrelsc: It nodes line should be removed, because it should be handled when reloading.

lisa/schema.py
```diff
@@ -1155,6 +1155,8 @@ def reload_requirements(self) -> None:
 
         if self.nodes_raw:
             for node_raw in self.nodes_raw:
+                if not isinstance(node_raw, dict):
```
squirrelsc, line 1158: to_dict is handled by load_by_type already.

LiliDeng: I saw it threw below exception here `node_raw[constants.TYPE]` when pass in as list.
```TypeError: 'RemoteNode' object is not subscriptable```

lisa/sut_orchestrator/azure/common.py
LiliDeng: removed

squirrelsc (approved):

PR #2325, opened 2022-10-21
FIO Testcase addition to LISA

smit-gardhariya:
Generic FIO Test case addition to LISA


LiliDeng:
What is the purpose of this case? 
By default, it will run testing against the os disk, if it is a perf testing, do we have baseline for it?

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +189,104 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
     def perf_storage_over_nfs_synthetic_udp_4k(self, result: TestResult) -> None:
         self._perf_nfs(result, protocol="udp")
 
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test data disk performance.
```
LiliDeng, line 194: it is not data disk, if run against azure vm, it is os disk, and it actually depends on where the testing run.

microsoft/testsuites/performance/common.py
```diff
@@ -42,8 +42,8 @@ def perf_disk(
     filename: str,
     core_count: int,
     disk_count: int,
-    disk_setup_type: DiskSetupType,
-    disk_type: DiskType,
+    disk_setup_type: Optional[DiskSetupType],
```
squirrelsc: 1. Give the default value here, instead of set None when calling. `disk_setup_type: DiskSetupType = DiskSetupType.unknown,`
2. Add unknown for both enums, and use it.

smit-gardhariya: Done

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: Yes this will run on OS disk of L2 guest o2 L2 host.

LiliDeng (approved):

microsoft/testsuites/performance/storageperf.py
```diff
@@ -187,6 +189,102 @@ def perf_storage_over_nfs_synthetic_tcp_4k(self, result: TestResult) -> None:
...
+        """,
```
squirrelsc, line 195: Please specify the schema and usage of test case in the description.

smit-gardhariya: Done with PR : https://github.com/microsoft/lisa/pull/2349

PR #2327, opened 2022-10-22
[fix test issue] fix 'unable to locate package' issue by waiting cloud-init finish init work

LiliDeng:
cloud-init will update /etc/apt/source.list file for new provisioned vm, fix install package issue by waiting for cloud-init finish init work.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -781,7 +781,23 @@ class Debian(Linux):
     def _initialize_package_installation(self) -> None:
         # wait running system package process.
         self.wait_running_package_process()
-
+        if type(self._node.os) == Ubuntu:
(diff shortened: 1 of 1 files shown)


squirrelsc (dismissed):

lisa/operating_system.py
```diff
@@ -781,7 +781,10 @@ def add_repository(
     def _initialize_package_installation(self) -> None:
         # wait running system package process.
         self.wait_running_package_process()
-
+        if type(self._node.os) == Ubuntu:
+            # wait till cloud-init finish to run the init work include updating
+            # /etc/apt/source.list file
```
squirrelsc: Add comments that, this applies to Azure only. It's ok for current logic, because we don't check its result.

PR #2328, opened 2022-10-23
[fix test issue] remove hid_hyperv from necessary modules checklist on azure platform

LiliDeng:


microsoft/testsuites/core/hv_module.py
```diff
@@ -122,6 +126,14 @@ def verify_initrd_modules(self, node: Node) -> None:
             ):
                 missing_modules.append(module)
 
+        environment = result.environment
```
squirrelsc: The environment can be got from parameters, it' doesn't need to get from result.

LiliDeng: update

squirrelsc (approved):

PR #2331, opened 2022-10-24
[fix issue] skip LIS driver installation testing when kernel version is unsupported

LiliDeng:
when run lis install case against openlogic centos 7_9 7.9.2022020700, found its release is 7.0.0, since it reads from /etc/os-release.
I used /etc/redhat-release now, it works well. @squirrelsc do you think any risk for this change?

### DIFF
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -14,6 +14,7 @@ from lisa.tools import Gcc, Modinfo, PowerShell, Uname
 from lisa.util import (
     LisaException,
     UnsupportedDistroException,
+    UnsupportedKernelException,
     find_patterns_in_lines,
     get_matched_str,
 )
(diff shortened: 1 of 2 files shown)


lisa/operating_system.py
```diff
@@ -1417,31 +1417,22 @@ def _is_package_in_repo(self, package: str) -> bool:
...
-            assert matches, f"cannot match version information from: {full_version}"
-            assert matches.group("vendor")
-            information = OsInformation(
-                version=self._parse_version(matches.group("version")),
-                vendor=matches.group("vendor"),
-                release=matches.group("version"),
-                codename=matches.group("codename"),
-                full_version=full_version,
-            )
+        # Parse /etc/redhat-release to support 6.x and 8.x. Refer to
```
squirrelsc: What's reason to make this change? The os-release should be standard, which works on new distros.

LiliDeng: @squirrelsc when run lis install case against openlogic centos 7_9 7.9.2022020700, the release is 7.0.0. read from os-release

squirrelsc: How about 8.x? If so, we can move 7.x to use redhat-release only.

LiliDeng: let me try

squirrelsc (approved):

PR #2332, opened 2022-10-24
SRIOV tests, verify at runtime SRIOV is active before start

mcgov:
Noticed some tests failing with an error that no SRIOV interface is present, in DPDK tests I verify it's on and wait for the interface if not. Adding two commits:

- Adding a line to verify SRIOV is enabled on Azure side.
- Add a line to wait for SRIOV on test node side.

### DIFF
--- a/microsoft/testsuites/network/common.py
+++ b/microsoft/testsuites/network/common.py
@@ -48,6 +48,7 @@ def initialize_nic_info(
             ).is_greater_than(0)
         node_nic_info = Nics(node)
         node_nic_info.initialize()
+        node_nic_info.wait_for_sriov_enabled()
         for _, node_nic in node_nic_info.nics.items():
             assert_that(node_nic.ip_addr).described_as(
                 f"This interface {node_nic.upper} does not have a IP address."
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/network/common.py
```diff
@@ -36,12 +36,19 @@ def initialize_nic_info(
     for node in environment.nodes.list():
         if is_sriov:
             network_interface_feature = node.features[NetworkInterface]
+            # verify SRIOV in environment is enabled, wait for enable if not.
+            if not network_interface_feature.is_enabled_sriov():
+                network_interface_feature.switch_sriov(
+                    enable=True, wait=True, reset_connections=True
+                )
+
```
LiliDeng, line 44: Do you have an example of which fails for this reason? I think we should work on the case which will change the AN for the VM.

mcgov: Let me see if I can find it, I noticed some SRIOV tests fast-failing because SRIOV was not detected immediately upon test entry. Since we're reusing environments it occurred to me that we should wait for the VF for some amount of time before starting the test, it may be that it was just enabled and we need to wait a minute or two for the VF to show up.

mcgov: I don't think this is actually needed. Going to close

PR #2333, opened 2022-10-24
alternative implementation of CreateServicePrincipal_linux.sh

mcarifio:
An alternative script that passes arguments via the command line and makes a different set of assumptions about defaults. Requires `jq` be installed.

### DIFF
--- a/microsoft/utils/create-service-principal.sh
+++ b/microsoft/utils/create-service-principal.sh
@@ -1,0 +1,85 @@
+#!/usr/bin/env bash
+# create-service-principal.sh [--subscription=${uuid}] [--id="<name>"] [--role=Contributor | Owner] --pathname=${pathname}
+# The default subscription is the one you're currently logged into.
+# The default id is ${USER}-penguinator-sp.
+# The default role is Owner.
+
+# http://redsymbol.net/articles/unofficial-bash-strict-mode/
+set -Eeuo pipefail
+# print arguments to stderr and exit with error status.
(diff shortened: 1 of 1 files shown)


mcarifio:
@microsoft-github-policy-service agree

squirrelsc:
@lesscodingmorehappiness Can you check if it's same functionality as PowerShell version?

lesscodingmorehappiness:
Hi @mcarifio thanks for the contribution. I tested your script on my VM and it should work as same functionality as our existing versions.
Just few things:
1. I think this script can be used for creating SP for penguinator usage but that's not the only use case, so I'd prefer use LISA$(uuid) as default ID. Also, this allows the same user to create multiple SPs if needed.
2. For similar reason I'd prefer we keep default role as contributor.
3. I think some ideas of this script are great, but may not be very straightforward? i.e. Someone has to read the script first so they know they can input to set ID, role and etc. It may be better to have a message when triggering the script or take the variables step by step like old scripts.
4. I also tested my script and it worked. I'm curious what's the error/failure you got. Can you please share here or via email so I can do some check? Thanks

mcarifio:
Yes, @lesscodingmorehappiness, this doesn't add much additional functionality. It assumes you're logged in and uses the current subscription as the default. Passing in arguments rather than reading from stdin makes it a little easier to reuse `create-service-principal.sh` in other scripts as well. But this kind of reuse may never have been intended. Up to you how best to use it.

lesscodingmorehappiness (approved):

PR #2334, opened 2022-10-24
Implement StartStop feature for Libvirt orchestrator.

cwize1:
As part of this change, the libvirt connection object is now kept around long-term. This is required so that re-attaching the console logger works correctly. (Though in reality, it was already effectively kept around because of the console logger.)


cwize1:
@anirudhrb You may wish to review.

squirrelsc (approved):

PR #2336, opened 2022-10-25
Add a suite of stress-ng tests

anirudhrb:
stress-ng provides various stressors to stess a system in various ways. These stressors are divided into classes like cpu, memory, network etc. Introduce a new suite to run each class of stressors as a test case.

### DIFF
--- a/lisa/tools/stress_ng.py
+++ b/lisa/tools/stress_ng.py
@@ -5,6 +5,7 @@ from typing import cast
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
+from lisa.util.process import Process
 
 from .gcc import Gcc
 from .git import Git
@@ -31,7 +32,7 @@ class StressNg(Tool):
             self._install_from_src()
         return self._check_exists()
 
-    def launch(
+    def launch_vm_stressor(
         self, num_workers: int = 0, vm_bytes: str = "", timeout_in_seconds: int = 0
     ) -> None:
         # --vm N, start N workers spinning on anonymous mmap
@@ -56,6 +57,21 @@ class StressNg(Tool):
 
         cmd += f" --timeout {timeout_in_seconds} "
         self.run(cmd, force_run=True, timeout=timeout_in_seconds)
+
+    def launch_job_async(self, job_file: str, sudo: bool = False) -> Process:
+        return self.run_async(f"--job {job_file}", force_run=True, sudo=sudo)
+
+    def launch_class_async(
+        self,
+        class_name: str,
+        num_workers: int = 0,
+        timeout_secs: int = 60,
+        sudo: bool = False,
+    ) -> Process:
+        return self.run_async(
+            f"--sequential {num_workers} --class {class_name} --timeout {timeout_secs}",
+            sudo=sudo,
+        )
 
     def _install_from_src(self) -> bool:
         tool_path = self.get_tool_path()
--- a/microsoft/testsuites/stress/stress_ng_suite.py
+++ b/microsoft/testsuites/stress/stress_ng_suite.py
@@ -1,0 +1,171 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from pathlib import Path, PurePath
+from typing import Any, Dict, List, cast
+
+from lisa import (
+    Environment,
+    RemoteNode,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    notifier,
+)
+from lisa.features import SerialConsole
+from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
+from lisa.testsuite import TestResult
+from lisa.tools import StressNg
+from lisa.util import SkippedException
+from lisa.util.process import Process
+
+
+@TestSuiteMetadata(
+    area="stress",
+    category="stress-ng",
+    description="""
+    A suite for running the various classes of stressors provided
+    by stress-ng.
+    """,
+)
+class StressNgTestSuite(TestSuite):
+    TIME_OUT = 3600
+    CONFIG_VARIABLE = "stress_ng_jobs"
+
+    @TestCaseMetadata(
+        description="""
+        Runs a stress-ng jobfile. The path to the jobfile must be specified using a
+        runbook variable named "stress_ng_jobs". For more info about jobfiles refer:
+        https://manpages.ubuntu.com/manpages/jammy/man1/stress-ng.1.html
+        """,
+        priority=4,
+    )
+    def verify_stress_ng_jobfile(
+        self,
+        variables: Dict[str, Any],
+        environment: Environment,
+        result: TestResult,
+    ) -> None:
+        if self.CONFIG_VARIABLE in variables:
+            jobs = variables[self.CONFIG_VARIABLE]
+            for job_file in jobs:
+                self._run_stress_ng_job(job_file, environment, result)
+        else:
+            raise SkippedException("No jobfile provided for stress-ng")
+
+    @TestCaseMetadata(
+        description="Runs stress-ng's 'cpu' class stressors for 60s each.",
+        priority=4,
+    )
+    def verify_stress_ng_cpu_stressors(
+        self,
+        environment: Environment,
+    ) -> None:
+        self._run_stressor_class(environment, "cpu")
+
+    @TestCaseMetadata(
+        description="Runs stress-ng's 'memory' class stressors for 60s each.",
+        priority=4,
+    )
+    def verify_stress_ng_memory_stressors(
+        self,
+        environment: Environment,
+    ) -> None:
+        self._run_stressor_class(environment, "memory")
+
+    @TestCaseMetadata(
+        description="Runs stress-ng's 'vm' class stressors for 60s each.",
+        priority=4,
+    )
+    def verify_stress_ng_vm_stressors(
+        self,
+        environment: Environment,
+    ) -> None:
+        self._run_stressor_class(environment, "vm")
+
+    @TestCaseMetadata(
+        description="Runs stress-ng's 'io' class stressors for 60s each.",
+        priority=4,
+    )
+    def verify_stress_ng_io_stressors(
+        self,
+        environment: Environment,
+    ) -> None:
+        self._run_stressor_class(environment, "io")
+
+    @TestCaseMetadata(
+        description="Runs stress-ng's 'network' class stressors for 60s each.",
+        priority=4,
+    )
+    def verify_stress_ng_network_stressors(
+        self,
+        environment: Environment,
+    ) -> None:
+        self._run_stressor_class(environment, "network")
+
+    def _run_stressor_class(self, environment: Environment, class_name: str) -> None:
+        nodes = [cast(RemoteNode, node) for node in environment.nodes.list()]
+        procs: List[Process] = []
+        try:
+            for node in nodes:
+                procs.append(node.tools[StressNg].launch_class_async(class_name))
+            for proc in procs:
+                proc.wait_result(timeout=self.TIME_OUT, expected_exit_code=0)
+        except Exception as e:
+            self._check_panic(nodes)
+            raise e
+
+    def _run_stress_ng_job(
+        self,
+        job_file: str,
+        environment: Environment,
+        test_result: TestResult,
+    ) -> None:
+        nodes = [cast(RemoteNode, node) for node in environment.nodes.list()]
+        procs: List[Process] = []
+        job_file_name = Path(job_file).name
+        test_status = TestStatus.QUEUED
+        test_msg = ""
+        try:
+            for node in nodes:
+                remote_working_dir = node.working_path / "stress_ng_jobs"
+                node.shell.mkdir(remote_working_dir, exist_ok=True)
+                job_file_dest = remote_working_dir / job_file_name
+                node.shell.copy(PurePath(job_file), job_file_dest)
+                procs.append(node.tools[StressNg].launch_job_async(str(job_file_dest)))
+            for proc in procs:
+                proc.wait_result(expected_exit_code=0)
+            test_status = TestStatus.PASSED
+        except Exception as e:
+            test_status = TestStatus.FAILED
+            test_msg = repr(e)
+        finally:
+            self._send_subtest_msg(
+                test_result.id_,
+                environment,
+                job_file_name,
+                test_status,
+                test_msg,
+            )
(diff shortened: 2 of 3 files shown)


lisa/tools/stress_ng.py
```diff
@@ -47,6 +48,20 @@ def launch(
...
 
+    def launch_job_async(self, job_file: str, sudo: bool = False) -> Process:
+        return self.run_async(f"--job {job_file}", force_run=True, sudo=sudo)
+
+    def launch_class_async(
+        self,
+        class_name: str,
+        num_workers: int = 0,
+        timeout_secs: int = 60,
+        sudo: bool = False,
```
LiliDeng, line 69: sudo is not used

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+@TestSuiteMetadata(
+    area="stress",
+    category="stress-ng",
+    description="""
+    A suite for running the various classes of stressors provided
+    by stress-ng.
+    """,
+)
+class StressNgTestSuite(TestSuite):
+    TIME_OUT = 3600
```
LiliDeng, line 31: default timeout is 3600 for each case

lisa/tools/stress_ng.py
```diff
@@ -47,6 +48,20 @@ def launch(
             cmd += f" --timeout {timeout_in_seconds} "
         self.run(cmd, force_run=True)
 
+    def launch_job_async(self, job_file: str, sudo: bool = False) -> Process:
+        return self.run_async(f"--job {job_file}", force_run=True, sudo=sudo)
+
+    def launch_class_async(
```
squirrelsc, line 64: Can you change the default `launch` method to be more accurate name? It looks there shouldn't have a default `launch` method, because the stress_ng handles different scenarios.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+    def verify_stress_ng_jobfile(
+        self,
+        log: Logger,
+        node: Node,
+        variables: Dict[str, Any],
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+    ) -> None:
+        if self.CONFIG_VARIABLE in variables:
```
squirrelsc, line 50: It can provide a default value, instead of skip the test case.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+    by stress-ng.
+    """,
+)
+class StressNgTestSuite(TestSuite):
+    TIME_OUT = 3600
+    CONFIG_VARIABLE = "stress_ng_jobs"
+
+    @TestCaseMetadata(
+        description="""
+        """,
```
squirrelsc, line 39: description is required.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+        timeout=TIME_OUT,
+        priority=4,
+    )
+    def verify_stress_ng_jobfile(
```
squirrelsc, line 42: Because there is no "validation" in this kind of perf test cases, start the name with "perf_", not "verify_"

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+            jobs = variables[self.CONFIG_VARIABLE]
+            for job_file in jobs:
+                self._run_stress_ng_job(job_file, environment, result)
+        else:
+            raise SkippedException("No jobfile provided for stress-ng")
+
+    @TestCaseMetadata(
+        description="Runs stress-ng's 'cpu' class stressors",
+        priority=4,
+        timeout=3600,
```
squirrelsc: Timeout used by LISA to force stop a case. But it makes the case doesn't have a right result. Make sure the test case end before the timeout. The time to run should be provided in description too.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+                remote_working_dir = node.working_path / "stress_ng_jobs"
+                node.shell.mkdir(remote_working_dir, exist_ok=True)
+                job_file_dest = remote_working_dir / job_file_name
+                node.shell.copy(PurePath(job_file), job_file_dest)
+                procs.append(node.tools[StressNg].launch_job_async(str(job_file_dest)))
+            for proc in procs:
+                proc.wait_result(expected_exit_code=0)
+            test_status = TestStatus.PASSED
+        except Exception:
+            test_status = TestStatus.FAILED
```
squirrelsc, line 139: Add failure reason as possible for failed test cases.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+    @TestCaseMetadata(
+        description="""
+        """,
+        timeout=TIME_OUT,
+        priority=4,
+    )
+    def verify_stress_ng_jobfile(
+        self,
+        log: Logger,
+        node: Node,
```
LiliDeng: remove unused parameters

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,165 @@
...
+    ) -> None:
+        self._run_stressor_class(environment, "network")
+
+    def _run_stressor_class(self, environment: Environment, class_name: str) -> None:
+        nodes = [cast(RemoteNode, node) for node in environment.nodes.list()]
+        procs: List[Process] = []
+        for node in nodes:
+            procs.append(node.tools[StressNg].launch_class_async(class_name))
+        for proc in procs:
+            proc.wait_result(timeout=3600, expected_exit_code=0)
```
LiliDeng: run in parallel instead of loop

anirudhrb: They will run in parallel in the target node since we are running async.

lisa/tools/stress_ng.py
anirudhrb: Sure, renamed it to a proper name.

lisa/tools/stress_ng.py
anirudhrb: Fixed.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: Removed the timeout property for the tests. Using this to wait for the stress-ng tool to finish executing.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: Removed now.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: Done.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: Done.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: Done.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: These tests are finished with 3600s.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,161 @@
...
+        test_result: TestResult,
+    ) -> None:
+        nodes = [cast(RemoteNode, node) for node in environment.nodes.list()]
+        procs: List[Process] = []
+        job_file_name = Path(job_file).name
+        test_status = TestStatus.QUEUED
+        test_msg = ""
+        try:
+            for node in nodes:
+                remote_working_dir = node.working_path / "stress_ng_jobs"
```
mcgov, line 130: I think we've been leaning towards using the Path, you can use remote_wroking_dir = node.working_path.join("stress_ng_jobs") and then avoid casting to a path afterwards.

Very excited to have a stress-ng suite! Thank you for this effort 🫡

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,161 @@
...
+    ) -> None:
+        self._run_stressor_class(environment, "network")
+
+    def _run_stressor_class(self, environment: Environment, class_name: str) -> None:
+        nodes = [cast(RemoteNode, node) for node in environment.nodes.list()]
+        procs: List[Process] = []
+        for node in nodes:
+            procs.append(node.tools[StressNg].launch_class_async(class_name))
+        for proc in procs:
+            proc.wait_result(timeout=self.TIME_OUT, expected_exit_code=0)
```
LiliDeng: should we store the perf results if they are perf testing, or any other check, I want to figure out the value of these cases.

anirudhrb: Well, now that I think about it, they are not really perf tests. We don't collect any metrics. According to the stress-ng man page: 

> ... it has never been intended to be used as a precise benchmark test suite, so do NOT use it in this manner.

The value of these tests right now is to induce a high load in the VM and see if anything bad (such as kernel panic) happens.

In the future, we plan to add detection of other kinds of stack traces in the kernel logs as well. The current `Dmesg` tool has the capability to detect such things, however we need more fine-grained detection (e.g. "Call Trace" matches all traces but in this case, we want to ignore OOMs since they are expected while stressing).

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: I removed the cast in the later line.  AFAIK, `node.working_path / "stress_ng_jobs"` is same as `node.working_path.joinpath("stress_ng_jobs")`.

anirudhrb: Actually, the cast to `PurePath` is needed below. Because `job_file` is a string coming from the runbook.

microsoft/testsuites/stress/stress_ng_suite.py
vyadavmsft: i would call they are stressor then perf. stress-ng is valuable tool for stress testing, usually used for long term testing.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,161 @@
...
```
LiliDeng: if test for 60 seconds, should we make the timeout as 3600?

microsoft/testsuites/stress/stress_ng_suite.py
LiliDeng: @anirudhrb we can pull the serial console log once VM is unresponsive in after_case.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: 60 seconds is per stressor. Each class (such as "cpu", "vm", "io" etc.) has multiple stressors in them and _each_ of them will run for 60 seconds. To give an example, the cpu stressors case runs for around 2000 seconds.

microsoft/testsuites/stress/stress_ng_suite.py
anirudhrb: How will we know the VM is unresponsive? Also, I believe the libvirt platform captures the serial logs of the nodes for each lisa run. Is this not the case for Azure platform?

LiliDeng (dismissed):

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,161 @@
...
+
+    @TestCaseMetadata(
+        description="Runs stress-ng's 'network' class stressors for 60s each.",
+        priority=4,
+    )
+    def verify_stress_ng_network_stressors(
+        self,
+        environment: Environment,
+    ) -> None:
+        self._run_stressor_class(environment, "network")
```
squirrelsc, line 103: Those test cases start process, but no verification, no data collected. What's they for?

anirudhrb: There is return code verification for the process. For now, we just want to see if any kernel panic (or something as bad) happens while stressing. Panic will cause the test to fail even without any explicit verification. In the future we will add more dmesg log validations.

squirrelsc: If so, use the serial console to check the kernel panic. That would give more detailed error information.

anirudhrb: Sure, we will use the serial console for verifying the logs. But this will be a future improvement. In libvirt platform serial console logs are captured automatically for all nodes. For now, these can be checked manually while investigating kernel panics.

squirrelsc: "Manually" is not a scalable way. Check panic is simple, refer to below example. If there are more patterns in libvirt, they can be added in future.

https://github.com/squirrelsc/LISA/blob/extension_prototype/microsoft/testsuites/core/provisioning.py#L258

anirudhrb: Added `check_panic`.

anirudhrb:
@LiliDeng @squirrelsc please review again.

microsoft/testsuites/stress/stress_ng_suite.py
```diff
@@ -0,0 +1,171 @@
...
+                environment,
+                job_file_name,
+                test_status,
+                test_msg,
+            )
+            self._check_panic(nodes)
+
+    def _check_panic(self, nodes: List[RemoteNode]) -> None:
+        for node in nodes:
+            node.features[SerialConsole].check_panic(saved_path=None, force_run=True)
```
LiliDeng, line 153: what will happen if save_path is None?

anirudhrb: If saved_path is not provided, it will save in a default location (`self._node.local_log_path / get_datetime_path()`). This handling is done in `get_console_log()` of serial_console.py.

LiliDeng (approved):

PR #2338, opened 2022-10-25
SRIOV: prefer /dev/random over /dev/zero

mcgov:
Johnson pointed out using random data might be better than 0 to avoid weird optimization issues. Swap /dev/zero to /dev/random (urandom might be slow if we run out of entropy)

### DIFF
--- a/microsoft/testsuites/network/common.py
+++ b/microsoft/testsuites/network/common.py
@@ -128,7 +128,7 @@ def sriov_vf_connection_test(
 
     dest_ssh.enable_public_key(source_ssh.generate_key_pairs())
     # generate 200Mb file
-    source_node.execute("dd if=/dev/zero of=large_file bs=200M count=1")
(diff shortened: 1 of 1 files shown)


LiliDeng:
thanks @mcgov , what is the impact for current implement (/dev/zero)?

mcgov:
> thanks @mcgov , what is the impact for current implement (/dev/zero)?

I had picked /dev/zero hoping it would be optimized by the filesystem so would be faster. Johnson pointed out that random data is probably better than all 0's since easily compressible data like 200mb of zeros might behave weirdly. /dev/random is better than /dev/urandom because it may be slower if we run out of entropy, and we don't need cryptographically secure random data for this test. Considering all that, /dev/random is the better choice.

The main fix was getting rid of the count=0 seek=200M argument since this results in a 0 sized file on some systems. Swapping from /dev/zero to /dev/random will be slightly less quick but faster than the old use of /dev/urandom

LiliDeng (approved):

PR #2341, opened 2022-10-26
Lagscope: Run server as async instead of '-D'

pupacha:
This way server logs are also printed and is useful for debugging

### DIFF
--- a/lisa/tools/lagscope.py
+++ b/lisa/tools/lagscope.py
@@ -125,26 +125,18 @@ class Lagscope(Tool, KillableMixin):
         for key in self._busy_pool_keys:
             sysctl.write(key, self._original_settings[key])
 
-    def run_as_server(self, ip: str = "", daemon: bool = True) -> None:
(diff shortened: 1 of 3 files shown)


lisa/tools/lagscope.py
```diff
@@ -125,26 +125,18 @@ def restore_busy_poll(self) -> None:
...
-            force_run=True,
-            sudo=True,
-            shell=True,
-            expected_exit_code=0,
-            expected_exit_code_failure_message=f"fail to launch cmd {self.command}"
-            f"{cmd}",
-        )
+        process = self.run_async(cmd, sudo=True, shell=True)
+        if not process.is_running():
+            raise LisaException(f"lagscope server failed to start: {result.stderr}")
```
LiliDeng: result is not defined.

pupacha: Missed to remove the variable. Removed it now.

LiliDeng (approved):

PR #2343, opened 2022-10-27
Fix ssh sessions issue in mshv_stress_vm_create test

anirudhrb:
The mshv_stress_vm_create test runs multiple instances of cloud-hypervisor using the CloudHypervisor tool. In doing so, multiple ssh sessions are created eventually exceeding the `MaxSessions` limit of `sshd`. This results in test failure with following message:

```
lisa.stderr Secsh channel 32 open FAILED: open failed: Connect failed
```

Fix this by setting the `MaxSessions` property in `/etc/ssh/sshd_config` to a high value that is unlikely to be exceeded.

### DIFF
--- a/microsoft/testsuites/mshv/mshv_root_tests.py
+++ b/microsoft/testsuites/mshv/mshv_root_tests.py
@@ -17,7 +17,7 @@ from lisa import (
 )
 from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
 from lisa.testsuite import TestResult
-from lisa.tools import Dmesg, Free, Ls, Lscpu, QemuImg, Wget
+from lisa.tools import Dmesg, Free, Ls, Lscpu, QemuImg, Ssh, Wget
 from lisa.util import SkippedException
 from microsoft.testsuites.mshv.cloud_hypervisor_tool import CloudHypervisor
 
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -191,3 +199,17 @@ def _send_subtest_msg(
         )
 
         notifier.notify(subtest_msg)
+
+    def _set_ssh_max_sessions(self, node: Node, max_sessions: int) -> None:
```
LiliDeng: there is a method set_max_session in ssh tool, please use it, thanks.

anirudhrb: Thanks! I have updated the PR.

microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -87,6 +88,13 @@ def verify_mshv_stress_vm_create(
             # fall back to defaults
             configs = [{}]
 
+        # This test can end up creating and a lot of ssh sessions and these kept active
+        # at the same time.
+        # In Ubuntu, the default limit is easily exceeded. So change the MaxSessions
+        # property in sshd_config to a high number that is unlikely to be exceeded.
+        node.tools[Ssh].set_max_session()
+        node.close()
```
LiliDeng: in set_max_session, it will restart sshd service, does it take effect? or we have to use node. Close()?

anirudhrb: Restarting sshd isn't sufficient. It doesn't take effect for existing connections. New connection needs to be established. Hence `node.close()`.  Similar usage is here: https://github.com/microsoft/lisa/blob/5f597d2b6c241278705f540b4a78fc2caa973e76/microsoft/testsuites/performance/common.py#L190-L195

squirrelsc: Can the set_max_session include this line directly?

LiliDeng (approved):

microsoft/testsuites/mshv/mshv_root_tests.py
anirudhrb: Done

PR #2344, opened 2022-10-27
[fix case issues] get dhcp timeout value from networkmanager config

LiliDeng:


lisa/tools/dhclient.py
```diff
@@ -44,15 +48,14 @@ def get_timeout(self) -> int:
...
-            group = find_group_in_lines(result.stdout, self._fedora_pattern)
-            if group and "default" not in group:
-                value = int(group["number"])
+            # the default value in fedora is 45
+            value = 45
+            result = self.node.tools[Journalctl].logs_for_unit("NetworkManager")
+            groups = find_groups_in_lines(result, self._fedora_pattern)
+            if len(groups) == 0:
+                raise LisaException("not find matched dhcp timeout line")
+            if value != int(groups[-1]["number"]):
```
squirrelsc: Why there are multiple items after regexp? if so, should we check each line, including both ipv4 and ipv6 on all nics?

LiliDeng: During testing, I only saw one image that has this kind of output, NetworkManager restart, so it has multiple line for the same nic.

```
Oct 28 01:42:06 lisa--315-e0-n0 NetworkManager[912]: <info>  [1666921326.6053] dhcp4 (eth0): activation: beginning transaction (timeout in 45 seconds)
Oct 28 01:42:13 lisa--315-e0-n0 NetworkManager[1198]: <info>  [1666921333.4400] dhcp4 (eth0): activation: beginning transaction (timeout in 45 seconds)
```

LiliDeng: find another better way to check config using `NetworkManager --print-config`, will update code and test

squirrelsc (approved):

PR #2347, opened 2022-10-27
Azure: implement extension feature

squirrelsc:


mcgov:
Very exciting to have this implemented, will the information about extensions be populated into the TestResult information? It should be easy to ID failures if they only happen with one extension enabled, as an example.

squirrelsc:
> Very exciting to have this implemented, will the information about extensions be populated into the TestResult information? It should be easy to ID failures if they only happen with one extension enabled, as an example.

Oh, this scenario is not addressed so far. Let me know when it's needed, and we can plan it. This change is to enable extension tests itself. A team needs to test add/update extensions.

PR #2348, opened 2022-10-27
Create a simple move file test to learn about LISA

contre:
Simple test to create a file in the home directory and then move to a data disk. Done as part of the ramp up training.

### DIFF
--- a/microsoft/testsuites/core/storage.py
+++ b/microsoft/testsuites/core/storage.py
@@ -11,6 +11,7 @@ from lisa import (
     TestCaseMetadata,
     TestSuite,
     TestSuiteMetadata,
+    search_space,
     simple_requirement,
 )
 from lisa.features import Disk, Nfs
@@ -24,7 +25,8 @@ from lisa.schema import DiskType
 from lisa.sut_orchestrator import AZURE
 from lisa.sut_orchestrator.azure.features import AzureDiskOptionSettings
 from lisa.sut_orchestrator.azure.tools import Waagent
-from lisa.tools import Blkid, Cat, Dmesg, Echo, Lsblk, NFSClient, Swap
+from lisa.tools import Blkid, Cat, Dmesg, Echo, Lsblk, Mount, NFSClient, Swap
+from lisa.tools.mv import Mv
 from lisa.util import BadEnvironmentStateException, LisaException, get_matched_str
 
 from .common import get_resource_disk_mount_point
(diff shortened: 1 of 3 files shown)


squirrelsc:
@contre It's great to see the mv tool! can you split it into a separated PR, so we can get it in?

microsoft/testsuites/core/storage.py
```diff
@@ -374,6 +376,63 @@ def verify_azure_file_share_nfs(self, log: Logger, node: Node) -> None:
         finally:
             nfs.delete_share()
             node.tools[NFSClient].stop(mount_dir)
+    
+    _file_path = "~/SAMPLE.txt"
```
dsrivastavv: These should be `CAPITALIZED_SNAKE_CASE`

microsoft/testsuites/core/storage.py
```diff
@@ -374,6 +376,63 @@ def verify_azure_file_share_nfs(self, log: Logger, node: Node) -> None:
...
+        3. Assert the file content in data disk. 
+        4. Assert the file on original place doesn't exist. 
+        """,
+        timeout=TIME_OUT,
+        priority=4,
+        requirement=simple_requirement(
+            disk=AzureDiskOptionSettings(
+                data_disk_count=search_space.IntRange(min=1),
+                data_disk_size=search_space.IntRange(min=20),
+            ),supported_platform_type=[AZURE],
```
dsrivastavv: We do not need to limit Platform to Azure, since no Azure dependent tools were used. This test seems like something that can run on any other cloud platforms as well.

microsoft/testsuites/core/storage.py
```diff
@@ -374,6 +376,63 @@ def verify_azure_file_share_nfs(self, log: Logger, node: Node) -> None:
...
+        
+        # Move file to data disk
+        node.tools[Mv].move_file(self._file_path, dest_path.as_posix(), sudo=True)
+
+        # read content from the file after moving
+        read_text = node.tools[Cat].read(dest_path.as_posix(), force_run=True, sudo=True)
+        
+        assert_that(
+            read_text,
+            f"content read from the file {dest_path}: '{read_text}' should be equal to content written to file: '{self._original_text}'",
```
dsrivastavv: [nit] We don't need to mention the compared quantities in assert_that message, and it should be displayed by assertion error message. We can remove `read_text` and `_original_text`

dsrivastavv:
Nice work! Please fix the pipelines as well

microsoft/testsuites/core/storage.py
contre: Should be addressed now

microsoft/testsuites/core/storage.py
contre: Removed the requirement

microsoft/testsuites/core/storage.py
contre: Removed as requested.

contre:
I will extract the Mv tool out into a standalone PR as requested,

dsrivastavv:
@contre We rebase and merge, so we need to ensure that each commit is meaningful as they are merged to `main` as-it-is. For example, this PR could be split into two commits:
1. Add mv tool
2. Add testcase for ....

contre:
> @contre We rebase and merge, so we need to ensure that each commit is meaningful as they are merged to `main` as-it-is. For example, this PR could be split into two commits:
> 
> 1. Add mv tool
> 2. Add testcase for ....

I decided to squash into one commit as this PR will not be merged. I'll create a separate PR to include the Mv tool that won't require cherry picking or anything of that nature.

dsrivastavv:
LGTM! Nice work

PR #2351, opened 2022-10-28
[improvement] move cloud-init wait into ubuntu class

LiliDeng:


lisa/operating_system.py
```diff
@@ -1061,6 +1044,15 @@ def _replace_default_entry(self, entry: str) -> None:
         cat = self._node.tools[Cat]
         cat.run("/etc/default/grub")
 
+    @retry(tries=10, delay=5)
```
squirrelsc: I think it doesn't need to add retry here, unless the `node.execute` raise exception. It rarely happens in LISA. This retry will work with the retry on Debian, so it will retry 10 times together.

lisa/transformers/kernel_installer.py
```diff
@@ -212,7 +212,7 @@ def install(self) -> str:
             f"deb {self.repo_url} {version_name} "
             f"restricted main multiverse universe"
         )
-        ubuntu.wait_running_package_process()
+        ubuntu._initialize_package_installation()
```
squirrelsc: 1. This part we need to think about how to refactor the code. Not to call a private method (starts with `_`) out of the class.
2. The original method name makes sense "wait running package process", because it needs to add repo. But it doesn't make sense to "initialize package installation", because no package will be installed after this line immediately.

lisa/operating_system.py
LiliDeng: remove

lisa/transformers/kernel_installer.py
LiliDeng: updated

lisa/transformers/kernel_installer.py
```diff
@@ -212,6 +212,7 @@ def install(self) -> str:
             f"deb {self.repo_url} {version_name} "
             f"restricted main multiverse universe"
         )
+        ubuntu.wait_cloud_init_finish()
```
squirrelsc: Should it be added into `wait_running_package_process`?

LiliDeng: then it will be run in every package installation.

squirrelsc: That's what we need to think about, why is the kernel installer special to wait cloud init? Why don't others need? Are we wrapping the functions in right way?

LiliDeng: kernel installer will update sources. List file by adding repo, if cloud-init does not finish the initialization work, the previous change will be overwritten.

squirrelsc: the os supports `add_repository`, can this be changed to use that method?

LiliDeng: updated

squirrelsc (approved):

PR #2356, opened 2022-10-29
[new case] verify_gdb

LiliDeng:


microsoft/testsuites/core/TestScripts/hello.c
```diff
@@ -0,0 +1,5 @@
+#include <stdio.h>
```
squirrelsc: use lower case for folders, it can be considered as test data, not test script. The c code is not script. So use `test_data` as folder name.

lisa/tools/gcc.py
```diff
@@ -26,11 +26,13 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
-    def compile(self, filename: str, output_name: str = "") -> None:
+    def compile(
+        self, filename: str, output_name: str = "", arguments: str = ""
+    ) -> None:
         if output_name:
-            self.run(f"{filename} -o {output_name}")
+            self.run(f"{arguments} {filename} -o {output_name}", shell=True)
```
squirrelsc: 1. It needs force_run arguments to make sure every time, it's recompiled.
2. It's going complex, so compose the command line, and call `run` once. Like,

```python
cmd = f"{arguments} {filename}"
if output_name:
    cmd += f"-o {output_name}"

self.run....
```

lisa/tools/gdb.py
```diff
@@ -0,0 +1,26 @@
...
+class Gdb(Tool):
+    @property
+    def command(self) -> str:
+        return "gdb"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def debug(self, filename: str, arguments: str = "-batch -ex 'run' -ex 'bt'") -> str:
```
squirrelsc: The `arguments` is not good to be default, because it doesn't like reusable or common. So just make arguments required.

microsoft/testsuites/core/gdb.py
```diff
@@ -0,0 +1,52 @@
...
+        priority=1,
+    )
+    def verify_gdb(self, node: Node) -> None:
+        if type(node.os) == CBLMariner:
+            node.os.install_packages(("binutils", "glibc-devel"))
+        self._copy_to_node(node, "hello.c")
+        node.tools[Gcc].compile(
+            f"{node.working_path}/hello.c", f"{node.working_path}/hello", "-g -ggdb"
+        )
+        output = node.tools[Gdb].debug(f"{node.working_path}/hello")
```
squirrelsc: This is to check if the VM will hang, how could it be detected?

microsoft/testsuites/core/gdb.py
```diff
@@ -0,0 +1,52 @@
...
+        matched = get_matched_str(output, self._gdb_output_pattern)
+        assert_that(matched).described_as(
+            f"dbg result doesn't matched expected value, actual output is {output}"
+        ).is_not_empty()
+
+    def _copy_to_node(self, node: Node, filename: str) -> None:
+        file_path = Path(os.path.dirname(__file__)) / "TestScripts" / filename
```
squirrelsc: This can be a const at class level, because the file is submitted, and never changed.

microsoft/testsuites/core/gdb.py
```diff
@@ -0,0 +1,52 @@
...
+        if not node.shell.exists(node.working_path / filename):
```
squirrelsc: Just copy as is, it's enough. So, the whole copy is only one line, it doesn't need a method.

microsoft/testsuites/core/gdb.py
```diff
@@ -0,0 +1,52 @@
...
+)
+class GDB(TestSuite):
+    # Hello World![Inferior 1 (process 1869) exited normally]
+    _gdb_output_pattern = re.compile(
+        r".*Hello World!\[Inferior .* \(process .*\) exited normally\].*"
+    )
+
+    @TestCaseMetadata(
+        description="""
+        This test case check gdb work well.
```
squirrelsc: More details, what error to be detected.

microsoft/testsuites/core/TestScripts/hello.c
LiliDeng: updated

lisa/tools/gcc.py
LiliDeng: updated

lisa/tools/gdb.py
LiliDeng: updated

microsoft/testsuites/core/gdb.py
LiliDeng: VM won't hung, just gdb output is unexpected in bad VM.
So check output has pattern 
```
    _gdb_output_pattern = re.compile(
        r".*Hello World!\[Inferior .* \(process .*\) exited normally\].*"
    )
```

microsoft/testsuites/core/gdb.py
LiliDeng: updated

microsoft/testsuites/core/gdb.py
LiliDeng: updated

microsoft/testsuites/core/gdb.py
LiliDeng: updated

microsoft/testsuites/core/gdb.py
LiliDeng: pass case output
```
Hello World![Inferior 1 (process 6753) exited normally]
No stack.
```

fail case output
```
Program received signal SIGSEGV, Segmentation fault.
0x0000000000000000 in ?? ()
#0  0x0000000000000000 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

both exit error code is 0

squirrelsc (approved):

PR #2358, opened 2022-10-31
DPDK: cleanup sriov enabled status in after-case

mcgov:
Make sure to clean up and re-enable SRIOV on the node in after_case. Defensive coding in the event the test fails while SRIOV is disabled.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -662,13 +662,13 @@ class Dpdk(TestSuite):
         environment: Environment = kwargs.pop("environment")
         for node in environment.nodes.list():
             interface = node.features[NetworkInterface]
+            if not interface.is_enabled_sriov():
+                log.debug("DPDK detected SRIOV was left disabled during cleanup.")
+                interface.switch_sriov(enable=True, wait=False, reset_connections=True)
             modprobe = node.tools[Modprobe]
             # cleanup driver changes
             if modprobe.module_exists("uio_hv_generic"):
                 node.tools[Service].stop_service("vpp")
                 modprobe.remove(["uio_hv_generic"])
-                node.close()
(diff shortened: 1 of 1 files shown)


LiliDeng:
I am interested in the root cause for when the AN disabled during testing, and why it was not set back after testing.

mcgov:
> I am interested in the root cause for when the AN disabled during testing, and why it was not set back after testing.

This is more of a defensive step, if there's an unexpected error during the test when sriov is disabled it would be good to check and re-enable it after the test. 

I've also started making sure to check and enable if needed at test start, since some of the SRIOV tests disable it and can potentially fail and leave it disabled. This is the second half to make sure my test doesn't unintentionally leave it disabled for other tests.

Since you pointed it out, I made a change to move the re-enable before the driver reloads, since re-enabling should be a 'safer' operation. Removed node.close call since the SRIOV re-enable will do this before the driver reloads are called.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -661,9 +661,14 @@ def _force_dpdk_default_source(self, variables: Dict[str, Any]) -> None:
...
+            interface = node.features[NetworkInterface]
+            if not interface.is_enabled_sriov():
+                log.debug("DPDK detected SRIOV was left disabled during cleanup.")
+                interface.switch_sriov(enable=True, wait=False, reset_connections=True)
             modprobe = node.tools[Modprobe]
+            # cleanup driver changes
             if modprobe.module_exists("uio_hv_generic"):
                 node.tools[Service].stop_service("vpp")
                 modprobe.remove(["uio_hv_generic"])
-                node.close()
```
LiliDeng: not remove this, since switch_sriov is not always executed?

mcgov: fixed, good catch!

LiliDeng (approved):

PR #2361, opened 2022-11-01
Create Mv tool for simple move operations

contre:
Simple tool to perform file move operations.

### DIFF
--- a/lisa/base_tools/__init__.py
+++ b/lisa/base_tools/__init__.py
@@ -1,8 +1,9 @@
 from .cat import Cat
+from .mv import Mv
 from .rpm import Rpm
 from .sed import Sed
 from .uname import Uname
 from .wget import Wget
 from .yum_config_manager import YumConfigManager
 
-__all__ = ["Uname", "Sed", "Wget", "Cat", "Rpm", "YumConfigManager"]
+__all__ = ["Uname", "Sed", "Wget", "Cat", "Rpm", "YumConfigManager", "Mv"]
--- a/lisa/base_tools/mv.py
+++ b/lisa/base_tools/mv.py
@@ -1,0 +1,26 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+
+
+class Mv(Tool):
+    @property
+    def command(self) -> str:
+        return "mv"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def move(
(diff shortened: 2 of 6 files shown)


lisa/tools/mv.py
```diff
@@ -0,0 +1,17 @@
...
+    @property
+    def command(self) -> str:
+        return "mv"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def move_file(self, src_path: str, dest_path: str, sudo: bool = False) -> None:
+        self.run(f"{src_path} {dest_path}", sudo=sudo, force_run=True, shell=True)
```
LiliDeng: If a file exists and uses force_run, the command will execute with error, it won't impact the run, since here we don't set expected error code. but the existing file won't be overwritten
We need to think about the usage of this method, we can provide an overwrite option, and use mv with -f option.

lisa/tools/mv.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Mv(Tool):
+    @property
+    def command(self) -> str:
+        return "mv"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
```
dsrivastavv: Can you also add the support to move folders?

lisa/tools/mv.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+
+
+class Mv(Tool):
```
dsrivastavv: There are some use of "mv" command directly. Can you please replace those calls with this tool?

lisa/tools/mv.py
contre: The mv command will move folders with the current design. It does not require anything special. I will go ahead and rename to just move.

lisa/tools/mv.py
contre: Done.

lisa/tools/mv.py
contre: I've gone ahead and added the overwrite functionality as described here and I've removed the force_run arg as well. Is this an acceptable change to address your concerns?

squirrelsc: force_run must be here to make sure it actually executed instead of just taking the cached output.

contre: I've returned the force_run variable, I misunderstood its use.

squirrelsc (approved):

PR #2362, opened 2022-11-01
Fix for verify_uio_binding test to check the prerequisite for enabling the uio_hv_generic driver

ellie-di:
This PR adds a check for "CONFIG_UIO_HV_GENERIC" kernel config before enabling its driver. In case this config is not enabled the test will fail gracefully and will print out the follow error message:

Dpdk.verify_uio_binding: FAILED   failed. LisaException: The kernel config CONFIG_UIO_HV_GENERIC is not set.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -20,7 +20,17 @@ from lisa import (
 from lisa.features import NetworkInterface
 from lisa.nic import NicInfo
 from lisa.operating_system import OperatingSystem, Ubuntu
-from lisa.tools import Dmesg, Echo, Free, Lscpu, Lsmod, Lspci, Modprobe, Mount, KernelConfig
+from lisa.tools import (
+    Dmesg,
+    Echo,
+    Free,
+    Lscpu,
+    Lsmod,
+    Lspci,
+    Modprobe,
+    Mount,
+    KernelConfig,
+)
 from lisa.tools.mkfs import FileSystem
 from lisa.util.parallel import TaskManager, run_in_parallel, run_in_parallel_async
 from microsoft.testsuites.dpdk.common import DPDK_STABLE_GIT_REPO, check_dpdk_support
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -17,6 +17,7 @@ from lisa import (
     UnsupportedDistroException,
     constants,
 )
+from lisa.base_tools.uname import Uname
 from lisa.features import NetworkInterface
 from lisa.nic import NicInfo
 from lisa.operating_system import OperatingSystem, Ubuntu
@@ -24,12 +25,12 @@ from lisa.tools import (
     Dmesg,
     Echo,
     Free,
+    KernelConfig,
     Lscpu,
     Lsmod,
     Lspci,
     Modprobe,
     Mount,
-    KernelConfig,
 )
 from lisa.tools.mkfs import FileSystem
 from lisa.util.parallel import TaskManager, run_in_parallel, run_in_parallel_async
@@ -230,14 +231,21 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
     echo = node.tools[Echo]
     lsmod = node.tools[Lsmod]
     modprobe = node.tools[Modprobe]
-    kernelcfg = node.tools[KernelConfig]
+    kconfig = node.tools[KernelConfig]
+    uname = node.tools[Uname]
 
     # check if kernel config for Hyper-V VMBus is enabled
     config = "CONFIG_UIO_HV_GENERIC"
-    if not kernelcfg.is_enabled(config):
-        raise LisaException(
-            f"The kernel config {config} is not set."
-        )
+    if not kconfig.is_enabled(config):
+        kversion = uname.get_linux_information().kernel_version_raw
+        if kversion < "4.10":
+            raise SkippedException(
+                f"The kernel config {config} found in Linux kernels >= 4.10"
+            )
+        else:
+            raise LisaException(
+                f"The kernel config {config} is not set in kernel version {kversion}."
+            )
 
     # enable if it is not already enabled
     if not lsmod.module_exists("uio_hv_generic", force_run=True):
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -20,7 +20,7 @@ from lisa import (
 from lisa.features import NetworkInterface
 from lisa.nic import NicInfo
 from lisa.operating_system import OperatingSystem, Ubuntu
-from lisa.tools import Dmesg, Echo, Free, Lscpu, Lsmod, Lspci, Modprobe, Mount
+from lisa.tools import Dmesg, Echo, Free, Lscpu, Lsmod, Lspci, Modprobe, Mount, kernel_config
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -20,7 +20,7 @@
 from lisa.features import NetworkInterface
 from lisa.nic import NicInfo
 from lisa.operating_system import OperatingSystem, Ubuntu
-from lisa.tools import Dmesg, Echo, Free, Lscpu, Lsmod, Lspci, Modprobe, Mount
+from lisa.tools import Dmesg, Echo, Free, Lscpu, Lsmod, Lspci, Modprobe, Mount, kernel_config
```
mcgov: I think you should be able to just import KernelConfig and call it directly below, there's some module magic in the lisa.tools module that enables it.

mcgov (changes requested):
one requested change, looks good to me otherwise though

ellie-di:
@LiliDeng Chi mentioned that debian 9 has recently removed from the maintain list of images and maybe that's the reason that I was not able to reproduce the issue with exact same signature that was reported. could you provide more info on that?

I think that this prerequisite check should exist in the test, regardless of debian 9 is removed from the list or not.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -220,6 +220,15 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
     echo = node.tools[Echo]
     lsmod = node.tools[Lsmod]
     modprobe = node.tools[Modprobe]
+    kernelcfg = node.tools[kernel_config.KernelConfig]
+
+    # check if kernel config for Hyper-V VMBus is enabled
+    config = "CONFIG_UIO_HV_GENERIC"
+    if not kernelcfg.is_enabled(config):
+        raise LisaException(
```
LiliDeng: use skipexception

LiliDeng:
> @LiliDeng Chi mentioned that debian 9 has recently removed from the maintain list of images and maybe that's the reason that I was not able to reproduce the issue with exact same signature that was reported. could you provide more info on that?
> 
> I think that this prerequisite check should exist in the test, regardless of debian 9 is removed from the list or not.

hi this image was removed from pipeline list, we can still make change, in case other distro has the same scenario, you can use image credativ debian 9 9.20210721.0 to verify your change.

microsoft/testsuites/dpdk/dpdkutil.py
ellie-di: why?

ellie-di:
> > @LiliDeng Chi mentioned that debian 9 has recently removed from the maintain list of images and maybe that's the reason that I was not able to reproduce the issue with exact same signature that was reported. could you provide more info on that?
> > I think that this prerequisite check should exist in the test, regardless of debian 9 is removed from the list or not.
> 
> hi this image was removed from pipeline list, we can still make change, in case other distro has the same scenario, you can use image credativ debian 9 9.20210721.0 to verify your change.

Thanks. To be clear, I was using "credativ debian 9 9.20210721.0". Do you mean I can still use "credativ debian 9 9.20210721.0" in the pipeline?

microsoft/testsuites/dpdk/dpdkutil.py
LiliDeng: debian has older kernel version like 4.9, it doesn't contain this module, it is expected, maybe we can also add minimal kernel version check here.
```
prompt: Generic driver for Hyper-V VMBus
type: tristate
depends on: [CONFIG_HYPERV](https://cateee.net/lkddb/web-lkddb/HYPERV.html)
defined in [drivers/uio/Kconfig](https://github.com/torvalds/linux/tree/master/drivers/uio/Kconfig)
found in Linux kernels: 4.10–4.20, 5.0–5.19, 6.0, 6.1-rc+HEAD
modules built: uio_hv_generic
```

LiliDeng:
> > > @LiliDeng Chi mentioned that debian 9 has recently removed from the maintain list of images and maybe that's the reason that I was not able to reproduce the issue with exact same signature that was reported. could you provide more info on that?
> > > I think that this prerequisite check should exist in the test, regardless of debian 9 is removed from the list or not.
> > 
> > 
> > hi this image was removed from pipeline list, we can still make change, in case other distro has the same scenario, you can use image credativ debian 9 9.20210721.0 to verify your change.
> 
> Thanks. To be clear, I was using "credativ debian 9 9.20210721.0". Do you mean I can still use "credativ debian 9 9.20210721.0" in the pipeline?

you can use credativ debian 9 9.20210721.0 to do the testing, locally or in single test case pipeline.

microsoft/testsuites/dpdk/dpdkutil.py
ellie-di: I was wondering the reason for "use skipexception"?

ellie-di: I think checking for kconfig is more generic than a minimal kernel version check.

LiliDeng: skipexception means current behavior is expected, we just skip this testing on current distro. is this clear?

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -220,6 +230,15 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
     echo = node.tools[Echo]
     lsmod = node.tools[Lsmod]
     modprobe = node.tools[Modprobe]
+    kernelcfg = node.tools[KernelConfig]
+
+    # check if kernel config for Hyper-V VMBus is enabled
+    config = "CONFIG_UIO_HV_GENERIC"
```
dsrivastavv, line 239: [nit] This should be a class level constant

ellie-di: this function looks like a util function, which class? Meanwhile, following existing coding style, a few lines above the `hv_uio_generic_uuid` could've been defined as a constant but is a literal variable inside the function, which is totally fine in my opinion.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -220,6 +231,22 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
...
     modprobe = node.tools[Modprobe]
+    kconfig = node.tools[KernelConfig]
+    uname = node.tools[Uname]
+
+    # check if kernel config for Hyper-V VMBus is enabled
+    config = "CONFIG_UIO_HV_GENERIC"
+    if not kconfig.is_enabled(config):
+        kversion = uname.get_linux_information().kernel_version_raw
+        if kversion < "4.10":
+            raise SkippedException(
```
LiliDeng: ```uname.get_linux_information().kernel_version_raw``` contains str like 4.9.0-16-amd64, it can't be used for comparison.
Change code like below, it works.
```
        kversion = uname.get_linux_information().kernel_version
        if kversion < "4.10.0":
```

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -220,6 +231,22 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
...
+                f"The kernel config {config} found in Linux kernels >= 4.10"
```
LiliDeng: raise `UnsupportedKernelException` is more suitable than `SkippedException`

mcgov (approved):
lgtm

LiliDeng (approved):

PR #2363, opened 2022-11-02
[fix issue] change the if condition when there is no MaxNetworkInterfaces capability for the VM size

LiliDeng:
when there is no MaxNetworkInterfaces capability, the node_space.network_interface.max_nic_count is a range, not empty or none
update for #2322

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1562,7 +1562,9 @@ class AzurePlatform(Platform):
 
         # for some new sizes, there is no MaxNetworkInterfaces capability
         # and we have to set a default value for max_nic_count
-        if not node_space.network_interface.max_nic_count:
+        if not node_space.network_interface.max_nic_count or not isinstance(
+            node_space.network_interface.max_nic_count, int
+        ):
             node_space.network_interface.max_nic_count = 1
 
         # some vm size do not have resource disk present


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1562,7 +1562,9 @@ def _resource_sku_to_capability(  # noqa: C901
 
         # for some new sizes, there is no MaxNetworkInterfaces capability
         # and we have to set a default value for max_nic_count
-        if not node_space.network_interface.max_nic_count:
+        if not node_space.network_interface.max_nic_count or not isinstance(
```
squirrelsc, line 1565: If it's a range (as the PR description), it should take max value in the range, instead of 1.

LiliDeng: it is "max_nic_count": {"min": 1, "max": 9223372036854775807, "max_inclusive": true}}

squirrelsc: ok, if so, does it mean we treat all unknown max as 1?

LiliDeng: yes, otherwise we have to hard code then one by one, from local run, I saw like 5 - 6 vm sizes don't have MaxNetworkInterfaces capability

squirrelsc (approved):

PR #2364, opened 2022-11-02
Update libvirt-python to latest and fix permission issue

pupacha:
* In `_create_node`, os disk file is copied only when it is not present on the libvirt host. Incase, the file already exists, there could be missing read permission issue. Hence, added read permission when os disk already exists in libvirt host. 

* Updated libvirt-python to latest.

cc: @anirudhrb

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -26,7 +26,7 @@ from lisa.feature import Feature
 from lisa.node import Node, RemoteNode, local_node_connect
 from lisa.operating_system import CBLMariner
 from lisa.platform_ import Platform
-from lisa.tools import Iptables, Journalctl, Ls, QemuImg, Uname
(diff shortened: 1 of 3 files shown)


squirrelsc:
@cwize1 Please help reviewing.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -522,7 +522,11 @@ def _create_node(
             source_exists = self.host_node.tools[Ls].path_exists(
                 path=node_context.os_disk_base_file_path, sudo=True
             )
-            if not source_exists:
+            if source_exists:
```
cwize1, line 525: If you are going to add an unrelated bug fix, at least mention the issue in the PR description.

cwize1 (approved):

lisa/sut_orchestrator/libvirt/platform.py
pupacha: Updated the PR description. This change is infact done as a separate commit.

squirrelsc (approved):

PR #2366, opened 2022-11-03
[fix issues] check guestregister finish run before running any zypper command on suse

LiliDeng:


lisa/operating_system.py
```diff
@@ -1610,6 +1610,17 @@ def add_repository(
 
     def _initialize_package_installation(self) -> None:
         self.wait_running_process("zypper")
+        from lisa.tools import Service
+
+
+        service = self._node.tools[Service]
+        if service.check_service_exists("guestregister"):
```
LiliDeng, line 1614: when I use like this way, it thows 'failed. NameError: name 'Service' is not defined', @squirrelsc any idea?
```
if TYPE_CHECKING:
    from lisa.tools import Service
```

squirrelsc: if a tool need to be referenced in os, it should be moved to base_tools. But if the tool is complex, it references other tools, they will be moved to base_tools too. For "service", there is no other dependent tools, so it should be simple.

lisa/tools/service.py
```diff
@@ -46,6 +46,9 @@ def check_service_status(self, name: str) -> bool:
     def check_service_exists(self, name: str) -> bool:
         return self._internal_tool._check_service_exists(name)  # type: ignore
 
+    def is_service_inactive(self, name: str) -> bool:
+        return self._internal_tool._is_service_inactive(name)  # type: ignore
```
squirrelsc, line 50: If there is no other implementation, just raise NotImplementError here, and overrides the method directly. Only if there are shared logic in base class, it needs a private method.

lisa/operating_system.py
LiliDeng: thanks. update

lisa/tools/service.py
LiliDeng: the service class is not base class of ServiceInternal and Systemctl, it was created to hide/wrapper implement of ServiceInternal and Systemctl.

squirrelsc (dismissed):

PR #2367, opened 2022-11-03
Use Aria tool for downloading image files

dsrivastavv:


lisa/tools/tar.py
```diff
@@ -6,13 +6,25 @@
...
 
 class Tar(Tool):
     @property
     def command(self) -> str:
         return "tar"
 
+    def _check_exists(self) -> bool:
+        if (
+            isinstance(self.node.os, Suse)
+            and not self.node.tools[Bzip2]._check_exists()
```
squirrelsc: The Bzip2 can be added as a dependency of Tar, if it's required.

squirrelsc: Avoid to call private method of other classes. If it happens, it means the best practices are not followed, or the class design should be changed.

lisa/tools/bzip2.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from lisa.executable import Tool
+
+
+class Bzip2(Tool):
+    @property
+    def command(self) -> str:
+        return "bzip2"
```
squirrelsc, line 10: Can bzip2 be installed on Suse?

lisa/tools/aria.py
```diff
@@ -0,0 +1,83 @@
+from pathlib import PurePosixPath
+from typing import TYPE_CHECKING
+
+from lisa.base_tools import Wget
+from lisa.executable import Tool
+from lisa.tools.make import Make
+from lisa.tools.tar import Tar
+
+if TYPE_CHECKING:
+    from lisa.operating_system import Posix
```
squirrelsc: the os reference won't bring cycle import for this tool. If so, just import as normal.

lisa/tools/bzip2.py
dsrivastavv: yes. I have added the install command.

lisa/tools/tar.py
dsrivastavv: This requirement is only for Suse. I checked on ubuntu and Redhat and it works fine

squirrelsc: When calling the `tools[Bzip2]`, the `_check_exists` is already called, so it's not useful to call `_check_exists`.

lisa/tools/aria.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def get(
+        self,
+        url: str,
+        file_path: str = "",
+        filename: str = "",
+        overwrite: bool = True,
+        sudo: bool = False,
+        force_run: bool = False,
+        num_connections: int = 2,
```
LiliDeng: use bigger number than 2 like 10 to speedup download

lisa/tools/aria.py
```diff
@@ -0,0 +1,83 @@
...
+        timeout: int = 600,
+    ) -> str:
+
+        if file_path:
+            # create folder when it doesn't exist
+            self.node.shell.mkdir(PurePosixPath(file_path), exist_ok=True)
```
LiliDeng: use mkdir tool than this, probably there is no permission when use shell mkdir

lisa/tools/aria.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            file_path = str(self.node.working_path)
+
+        download_path = f"{file_path}/{filename}"
+        download_pure_path = self.node.get_pure_path(download_path)
+        if overwrite and self.node.shell.exists(download_pure_path):
+            self.node.shell.remove(download_pure_path, recursive=True)
+            force_run = True
```
LiliDeng: seems this tool has `--allow-overwrite` this parameter, if so, please use this parameter.

lisa/tools/aria.py
dsrivastavv: This might be a function of number of cores. Let me set it to number of cores instead

lisa/tools/tar.py
dsrivastavv: Sure! I will remove this call

lisa/tools/aria.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get(
+        self,
+        url: str,
+        file_path: str = "",
+        filename: str = "",
+        overwrite: bool = True,
+        sudo: bool = False,
+        force_run: bool = False,
+        num_connections: Optional[int] = None,
```
LiliDeng, line 59: max-connection-per-server must be between 1 and 16.

lisa/tools/aria.py
```diff
@@ -0,0 +1,89 @@
...
+            self.node.tools[Mkdir].create_directory(file_path, sudo=sudo)
+        else:
+            file_path = str(self.node.working_path)
+
+        # set download path
+        download_path = f"{file_path}/{filename}"
+
+        # if num_connections is not specified, set to the number of cores on the node
+        if not num_connections:
+            num_connections = self.node.tools[Lscpu].get_core_count()
```
LiliDeng: max-connection-per-server must be between 1 and 16.
min(self.node.tools[Lscpu].get_core_count(), 16) ?

lisa/tools/aria.py
squirrelsc: The bzip2 is not changed.

lisa/tools/aria.py
dsrivastavv: fixed!

lisa/tools/aria.py
dsrivastavv: fixed!

lisa/tools/aria.py
```diff
@@ -0,0 +1,91 @@
...
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        posix_os: Posix = self.node.os  # type: ignore
+        try:
+            posix_os.install_packages(["aria2"])
+        except Exception as identifier:
+            self._log.debug(f"Failed to install aria2 from apt: {identifier}")
```
LiliDeng: apt is Debian only, right?

LiliDeng (approved):

PR #2368, opened 2022-11-04
[fix issue] correct the path of nijia

LiliDeng:
related with #2361

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -837,7 +837,7 @@ class DpdkTestpmd(Tool):
             dest_dir=str(cwd),
             sudo=True,
         )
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -837,7 +837,7 @@ def _install_ninja_and_meson(self) -> None:
             dest_dir=str(cwd),
             sudo=True,
         )
-        node.tools[Mv].move("ninja", "/usr/bin/ninja", overwrite=True, sudo=True)
+        node.tools[Mv].move(f"{cwd}/ninja", "/usr/bin/ninja", overwrite=True, sudo=True)
```
squirrelsc, line 840: cwd can be an argument of mv. It can be implied in src or dst.

LiliDeng: ok, will update

PR #2374, opened 2022-11-08
Fix to resolve file permissions issue with LTP_Lite test case

r-dailey:
Changed the file permissions command in the testcase to resolve it failing to move the logs from /opt/ltp.

### DIFF
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -125,7 +125,7 @@ class Ltp(Tool):
         self.run(parameters, sudo=True, force_run=True, timeout=12000)
 
         # to avoid no permission issue when copying back files
-        self.node.tools[Chmod].update_folder("/opt/ltp", "a+rwX", sudo=True)
+        self.node.tools[Chmod].update_folder("/opt", "a+rwX", sudo=True)
 
         # write output to log path
         self.node.shell.copy_back(


microsoft/testsuites/ltp/ltp.py
```diff
@@ -125,7 +125,7 @@ def run_test(
         self.run(parameters, sudo=True, force_run=True, timeout=12000)
 
         # to avoid no permission issue when copying back files
-        self.node.tools[Chmod].update_folder("/opt/ltp", "a+rwX", sudo=True)
+        self.node.tools[Chmod].update_folder("/opt", "a+rwX", sudo=True)
```
squirrelsc, line 128: It's not good to change permission of a system folder. How about creating the ltp folder first, and then change its permission? It may need to change following line to copy into that folder, instead of creating another layer of "ltp".

r-dailey: Ok I will change the solution to avoid changing the permissions of a system folder and test that.

LiliDeng: need change this part `a+rwX` for folder `/opt/ltp`

dsrivastavv, line 128: @r-dailey Do you know why it is failing because I see that the read permission on ltp folder is already set?

microsoft/runbook/azure.yml
```diff
@@ -24,7 +24,7 @@ variable:
     value: ""
     is_secret: true
   - name: admin_password
-    value: ""
+    value: "HelloWorld123!"
```
dsrivastavv: [nit] please remove

microsoft/testsuites/ltp/ltp.py
```diff
@@ -125,7 +125,8 @@ def run_test(
         self.run(parameters, sudo=True, force_run=True, timeout=12000)
 
         # to avoid no permission issue when copying back files
-        self.node.tools[Chmod].update_folder("/opt", "a+rwX", sudo=True)
+        self.node.tools[Chmod].update_folder("/opt/ltp", "a+rwX", sudo=True)
```
dsrivastavv: Do we need this now?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -125,7 +125,8 @@ def run_test(
...
+        self.node.tools[Chmod].update_folder("/home/lisatest", "a+rwX", sudo=True)
```
dsrivastavv: I think this should already be present. Can you please confirm?

dsrivastavv:
Please fix the commit history since we rebase and merge

microsoft/testsuites/ltp/ltp.py
r-dailey: Changed the log location to the working directory per a suggestion and that resolved the permissions issue.

r-dailey: It is failing because the read permission on the /opt folder is set to root only and the test does not run the copyback operation as root.

microsoft/runbook/azure.yml
r-dailey: Removed.

microsoft/testsuites/ltp/ltp.py
r-dailey: I will try without it and remove it if no longer neccessary.

microsoft/testsuites/ltp/ltp.py
r-dailey: What should already be present?

dsrivastavv: The read/write permission on the lisatest folder. Can you please confirm?

r-dailey: Oh I tried it without this at first and got a similar permissions error. I will retest without it to confirm though.

dsrivastavv: If this gives an error then it should be a bug. We use `lisatest` credential to access this folder and it's home folder should be accessible by itself.

r-dailey: I retested and it failed with a similar permissions error to before my change.

r-dailey: So should I file a new bug report for that or should I try to address it in this PR?

squirrelsc: What's the permission error here? Can you share from the log?

r-dailey: PermissionError: [Errno 13] Permission denied

squirrelsc: Which commands get this error? Is there other outputs from this command?
BTW, don't assume the user account is always "lisatest". It's configurable. It needs to use `~` to expand to home folder.

dsrivastavv: @squirrelsc The account LISA is using should have read/write access to this folder, right? And it should be something that should be handled by framework rather than individual tests?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -53,8 +53,8 @@ class Ltp(Tool):
     LTP_TESTS_GIT_TAG = "20190930"
     LTP_GIT_URL = "https://github.com/linux-test-project/ltp.git"
     BUILD_REQUIRED_DISK_SIZE_IN_GB = 2
-    LTP_RESULT_PATH = "/opt/ltp/ltp-results.log"
-    LTP_OUTPUT_PATH = "/opt/ltp/ltp-output.log"
+    LTP_RESULT_PATH = "/home/lisatest/ltp-results.log"
+    LTP_OUTPUT_PATH = "/home/lisatest/ltp-output.log"
```
dsrivastavv: Can you please use this from the username variable? `lisatest` name can change

squirrelsc: Users should have full permissions for their home folder.

microsoft/testsuites/ltp/ltp.py
r-dailey: The logs list the error as stemming from the copyback command on line 131 of the ltp.py file. I did not see any additional output from the command in the logs but I am somewhat new to this so I may have missed something.

squirrelsc: Can you double check if it causes by the source file not exist, or dest folder not exist?

microsoft/testsuites/ltp/ltp.py
r-dailey: Sure, Where is the variable store? Lisa.environment?

dsrivastavv: I think it should be present in `RemoteNode.connection_info`

microsoft/testsuites/ltp/ltp.py
r-dailey: for clarity you want me to check to see if the error is reproduced with a missing source or destination file?

microsoft/testsuites/ltp/ltp.py
r-dailey: Ok thank you

microsoft/testsuites/ltp/ltp.py
squirrelsc: Yes, I want to double check where the error message is from. My understanding is the home folder shouldn't have any permission issue to copy a file back.

r-dailey: With a missing source file the error changes to say "FileNotFoundError: [Errno 2] No such file" so it is not a missing file issue but definitely a permissions issue.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -125,7 +125,7 @@ def run_test(
         self.run(parameters, sudo=True, force_run=True, timeout=12000)
 
         # to avoid no permission issue when copying back files
-        self.node.tools[Chmod].update_folder("/opt/ltp", "a+rwX", sudo=True)
+        self.node.tools[Chmod].update_folder("/home/lisatest", "a+rwX", sudo=True)
```
LiliDeng: change "a+rwX" => "a+rwx" should be enough, can you please update this PR? @r-dailey

r-dailey: Sure I can change that and test it. I will update once that is done.

r-dailey: That did not resolve the permissions issue.

LiliDeng: @r-dailey this one should work, because /opt folder under root with insufficient permission, that is why we see permission denied issue when use the test user (non-root) to copy back file from vm into local machine.
```self.node.tools[Chmod].update_folder("/opt", "a+rwx", sudo=True)```

r-dailey: That is what I tried initially but @squirrelsc requested I do not use that form because that is modifying the permissions of a system folder and could influence other tests.

squirrelsc: > denied issue when use the test user (non-root) to copy back file from vm into local machine

Can we use test case working folder instead of /opt?

r-dailey: That is what the current changes in this PR are doing. I moved the log output location to the working directory, and I changed the permissions command to allow copyback to use that location as well. If the permissions should not need to be changed for that directory, then there may be a separate issue that needs to be addressed.

squirrelsc: You can change it to test case working folder, instead of the home root folder. I think there shouldn't have permissions issue. There are some test cases to copy files back from there.

r-dailey: Can you provide an example of the working folder?

LiliDeng: /opt folder is generated by compiling ltp. even we switch LTP_RESULT_PATH LTP_OUTPUT_PATH  to use user home folder, we still need change the permission, because mariner image default umask value is 027 and ltp run as sudo, most of linux distro umask value is 022. Now the most convenient way is to change permission for opt.

squirrelsc: Can it just change the subfolder in /opt, not change the whole folder?

LiliDeng: that is what current code does, it doesn't work.

squirrelsc: if so, we can give /opt more permission.

r-dailey: Changed to just modifying the permissions of /opt

LiliDeng (approved):

PR #2376, opened 2022-11-09
Adding a stressng test for cpu throttle

shruti25ratnam:
Adding cpu stress test launch on stressng tool.

### DIFF
--- a/lisa/tools/stress_ng.py
+++ b/lisa/tools/stress_ng.py
@@ -48,17 +48,17 @@ class StressNg(Tool):
         self.run(cmd, force_run=True)
 
     def launch_cpu(
-        self, num_cores: int = 0, timeout_in_seconds: int = 0
+        self, num_cores: int = 0, timeout_in_seconds: int = 3600
     ) -> None:
         # --cpu N, start N CPU workers
         # --timeout T, timeout after T seconds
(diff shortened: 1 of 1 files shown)


lisa/tools/stress_ng.py
```diff
@@ -47,6 +47,18 @@ def launch(
...
+        self, num_cores: int = 0, timeout_in_seconds: int = 0
+    ) -> None:
+        # --cpu N, start N CPU workers
+        # --timeout T, timeout after T seconds
+        cmd = ""
+        if num_cores:
+            cmd += f" --cpu {num_cores} "
+        if timeout_in_seconds:
+            cmd += f" --timeout {timeout_in_seconds} "
+            self.run(cmd, force_run=True, timeout=timeout_in_seconds)
```
LiliDeng:
```       
        if num_cores:
            cmd += f" --cpu {num_cores} "
        if timeout_in_seconds:
            cmd += f" --timeout {timeout_in_seconds} "
        self.run(cmd, force_run=True, timeout=timeout_in_seconds)
```

avylove: There are hanging spaces at the end of the strings

LiliDeng (approved):

PR #2377, opened 2022-11-09
[fix test issues] xfstests ext4 

LiliDeng:


microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -120,14 +120,18 @@ class Xfstesting(TestSuite):
...
+    # A regression test for b55c3cd102a6 ("ext4: add reserved GDT blocks check")
     # TODO: will figure out the detailed reason of every excluded case.
     EXCLUDED_TESTS = (
         "generic/211 generic/430 generic/431 generic/434 /xfs/438 xfs/490"
         + " btrfs/007 btrfs/178 btrfs/244 btrfs/262"
         + " xfs/030 xfs/032 xfs/050 xfs/052 xfs/106 xfs/107 xfs/122 xfs/132 xfs/138"
         + " xfs/144 xfs/148 xfs/175 xfs/191-input-validation xfs/289 xfs/293 xfs/424"
         + " xfs/432 xfs/500 xfs/508 xfs/512 xfs/514 xfs/515 xfs/516 xfs/518 xfs/521"
-        + " xfs/528 xfs/544 ext4/054"
+        + " xfs/528 xfs/544 ext4/054 ext4/058 ext4/059"
```
squirrelsc, line 134: We should have a systematical, and maintainable plan on community tests. Please come up a document and post in Github doc.

Principles,
1. The excluded tests should meet one of conditions: 1) test damages the environment, so the whole test case will fail. 2) the test case runs too long (need more details definition based on our Tier rules). 3) test case defect (fail or random failed), so it's not sufficient to verify something. we don't need to include all this kind of issues. Just include well known ones.
2. We should specify a version of community tests, instead of stick to main branch. If there is newer version kernel/distro need newer version tests, we can either 1) upgrade the version, or 2) specify new version for new kernel/distros.
3. The excluded/included list, the version should be configurable.
4. Community tests must send sub test message for each result with the error message if there is.

squirrelsc (approved):

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: Assigned to @somil55 a task.

PR #2378, opened 2022-11-09
TestSuite: support no timeout on test runs

squirrelsc:
There is a time limit on joining a thread, which is called from timeout lib. Remove it to support long haul testing over 7 days.


shruti25ratnam:
I have validated that this works. Thanks!

PR #2380, opened 2022-11-10
[fix issues] xfstests xfs file system case

LiliDeng:


microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -153,16 +153,15 @@ def can_install(self) -> bool:
     def dependencies(self) -> List[Type[Tool]]:
         return [Git, Make]
 
-    def run_test(self, test_type: str, timeout: int = 14400) -> str:
-        cmd_result = self.run(
-            f"-g {test_type}/quick -E exclude.txt",
+    def run_test(self, test_type: str, timeout: int = 14400) -> None:
+        self.run(
+            f"-g {test_type}/quick -E exclude.txt  > xfstest.log 2>&1",
```
squirrelsc, line 165: Copy this file back after the command finished. The parse test results should be called here to make sure the results are parsed every time too.

LiliDeng: update code to copy it back, in the following method, will parse results then copy back, after that delete this file before next testing.

squirrelsc: I mean the `run_test` should call the `check_test_results`. If you see how it's used, they are always called together. That make sure the file is always parsed.

squirrelsc (dismissed):

squirrelsc (dismissed):

PR #2383, opened 2022-11-11
Get the console log if the kdump test fails

lubaihua33:
We need to save the console log if the kdump test fails. In this commit, the get_console_log is added before throwing an exception.


LiliDeng:
is it enough to pull the serial console log when VM can't be connected?

lubaihua33:
> is it enough to pull the serial console log when VM can't be connected?

Yes, as we identify the known issue, we can add more checks for the console log.

LiliDeng (approved):

PR #2384, opened 2022-11-11
[fix issues][kdump] Set a smaller crashkernel for the VM with less memory

lubaihua33:
Some smaller VM sizes, such as Standard_B1ls, might have less than 1 G memory. The crash kernel size should be set to a smaller one. Redhat, Ubuntu, and Suse have different proposed settings for crash kernel. We combine their configuration to set an empirical value.

### DIFF
--- a/microsoft/testsuites/kdump/kdumpcrash.py
+++ b/microsoft/testsuites/kdump/kdumpcrash.py
@@ -265,7 +265,28 @@ class KdumpCrash(TestSuite):
         free = node.tools[Free]
         total_memory = free.get_total_memory()
 
-        if "T" in total_memory and float(total_memory.strip("T")) > 1:
+        # Ubuntu, Redhat and Suse have different proposed crashkernel settings
+        # Please see below refrences:
+        # Ubuntu: https://ubuntu.com/server/docs/kernel-crash-dump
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -265,7 +265,23 @@ def _kdump_test(self, node: Node, log_path: Path, log: Logger) -> None:
         free = node.tools[Free]
         total_memory = free.get_total_memory()
 
-        if "T" in total_memory and float(total_memory.strip("T")) > 1:
+        # Ubuntu Redhat and Suse have differnt proposed crashkernel settings
```
LiliDeng: Ubuntu, Redhat and Suse have different proposed crashkernel settings

LiliDeng: can we put guidance links here?

lubaihua33: Has added.

LiliDeng: > Ubuntu, Redhat and Suse have different proposed crashkernel settings

@lubaihua33 please fix the typo

LiliDeng (approved):

PR #2385, opened 2022-11-11
Add install step for ping tool

contre:
In the Ubuntu 22.10 minimal image, the ping tool is not installed by default so an install step was needed

### DIFF
--- a/lisa/tools/ping.py
+++ b/lisa/tools/ping.py
@@ -4,6 +4,8 @@ import re
 from typing import Optional
 
 from lisa.executable import Tool
+from lisa.operating_system import Debian
+from lisa.util import UnsupportedDistroException
 from lisa.util.process import Process
 
 INTERNET_PING_ADDRESS = "8.8.8.8"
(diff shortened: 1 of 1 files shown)


lisa/tools/ping.py
```diff
@@ -20,6 +21,18 @@ class Ping(Tool):
...
         return "ping"
 
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        assert isinstance(
+            self.node.os, Ubuntu
+        ), f"apt-add-repository: unsupported OS {self.node.os.name}"
```
LiliDeng: remove apt-add-repository:

lisa/tools/ping.py
```diff
@@ -20,6 +21,18 @@ class Ping(Tool):
     def command(self) -> str:
         return "ping"
 
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        assert isinstance(
```
squirrelsc: Don't assume it's Ubuntu, it can be any distro. Use if/else instead of assert.

squirrelsc: in else return the unsupported os exception.

contre: Done as requested.

squirrelsc (approved):

PR #2386, opened 2022-11-11
Add AddAptRepository tool as it is missing in Ubuntu 22.10 minimal

contre:
The add-apt-repository command is missing from the ubuntu 22.10 minimal image. Add a tool for this.

### DIFF
--- a/lisa/base_tools/apt_add_repository.py
+++ b/lisa/base_tools/apt_add_repository.py
@@ -1,0 +1,34 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from typing import TYPE_CHECKING, cast
+
+from lisa.executable import Tool
+
+if TYPE_CHECKING:
+    from lisa.operating_system import Ubuntu
+
+
+class AptAddRepository(Tool):
+    @property
+    def command(self) -> str:
+        return "apt-add-repository"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        ubuntu_os: Ubuntu = cast(Ubuntu, self.node.os)
+        package_name = "software-properties-common"
+        ubuntu_os.install_packages(package_name)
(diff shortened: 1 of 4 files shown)


lisa/base_tools/apt_add_repository.py
```diff
@@ -0,0 +1,34 @@
...
+    @property
+    def command(self) -> str:
+        return "apt-add-repository"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        from lisa.operating_system import Ubuntu
```
LiliDeng: put it on the top

lisa/base_tools/apt_add_repository.py
```diff
@@ -0,0 +1,34 @@
...
+
+        ubuntu_os: Ubuntu = cast(Ubuntu, self.node.os)
+        package_name = "software-properties-common"
+        ubuntu_os.install_packages(package_name)
+        return self._check_exists()
+
+    def add_repository(self, repo: str) -> None:
+        # create folder when it doesn't exist
```
LiliDeng: remove this comment

lisa/base_tools/apt_add_repository.py
squirrelsc: In base_tools, it cannot reference os. This line should be moved to above, and in type checking condition. From this place, use a line like below to workaround it.

```python
ubuntu: Ubuntu = self.node.os # type: ignore
```

contre: Thank you for the pointer, I was not sure how to do it because I found multiple examples throughout the repo.

lisa/base_tools/apt_add_repository.py
contre: Removed as requested. The dangers of copy paste programming.

LiliDeng (approved):

PR #2391, opened 2022-11-13
[fix issues] support suse on dpdk

LiliDeng:


lisa/operating_system.py
```diff
@@ -1552,6 +1552,21 @@ class Suse(Linux):
     )
     # Warning: There are no enabled repositories defined.
     _no_repo_defined = re.compile("There are no enabled repositories defined.", re.M)
+    # Name           : dpdk
+    # Version        : 19.11.10-150400.4.7.1
+    _suse_package_information_regex = re.compile(
```
squirrelsc, line 1557: use named group.

LiliDeng: fixed and tested

squirrelsc (approved):

PR #2394, opened 2022-11-14
Fix MSHV VM create stress test

anirudhrb:
VM creation is failing due to the "readonly=on" flag passed in cloud-hypervisor's disk config.

Don't use readonly disks for the VMs. Instead, create a copy of the disk image for each VM.

### DIFF
--- a/lisa/tools/cp.py
+++ b/lisa/tools/cp.py
@@ -1,0 +1,18 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from pathlib import PurePath
+
+from lisa.executable import Tool
+
+
+class Cp(Tool):
+    @property
+    def command(self) -> str:
+        return "cp"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def copy(self, src: PurePath, dest: PurePath) -> None:
+        self.run(f"{src} {dest}", force_run=True, expected_exit_code=0)
--- a/microsoft/testsuites/mshv/mshv_root_tests.py
+++ b/microsoft/testsuites/mshv/mshv_root_tests.py
@@ -17,7 +17,7 @@ from lisa import (
 )
 from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
 from lisa.testsuite import TestResult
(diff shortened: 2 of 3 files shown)


lisa/tools/cp.py
```diff
@@ -0,0 +1,18 @@
...
+    @property
+    def command(self) -> str:
+        return "cp"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def copy(self, src: PurePath, dest: PurePath) -> None:
+        self.run(f"{str(src)} {str(dest)}", expected_exit_code=0)
```
squirrelsc: add force_run for this method.

squirrelsc: it doesn't need `str()`, if the value is in string already. `f"{src} {dest}"`

squirrelsc:
@cwize1 please have a check.

cwize1:
> @cwize1 please have a check.

This doesn’t seem to be touching the libvirt orchestrator. Instead it is testing cloud hypervisor. So, I am not sure why I would need to review.

But regardless, LGTM

squirrelsc (dismissed):

microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -145,13 +145,14 @@ def _mshv_stress_vm_create(
             node.tools[Free].log_memory_stats_mb()
             procs = []
             for i in range(vm_count):
+                vm_disk_img_path = node.working_path / f"VM{i}_{self.DISK_IMG_NAME}"
+                node.tools[Cp].copy(disk_img_path, vm_disk_img_path)
```
pupacha, line 149: Are these vm disk files required after the testcase is finished? If not, we can clean them up and save some space on the Node.

lisa/tools/cp.py
anirudhrb: Done.

microsoft/testsuites/mshv/mshv_root_tests.py
anirudhrb: Done.

pupacha (approved):

squirrelsc (approved):

PR #2395, opened 2022-11-14
HyperV Module Reload use Nohup

mcgov:
- reload_hyperv_modules can fail when connection is reset, use nohup when calling the reload/interface reset/ ip renew calls

### DIFF
--- a/microsoft/testsuites/core/hv_module.py
+++ b/microsoft/testsuites/core/hv_module.py
@@ -18,7 +18,17 @@ from lisa import (
 from lisa.operating_system import Redhat
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
 from lisa.sut_orchestrator.azure.tools import LisDriver
-from lisa.tools import Find, KernelConfig, Lsinitrd, Lsmod, Modinfo, Modprobe, Uname
+from lisa.tools import (
+    Cat,
+    Find,
+    KernelConfig,
+    Lsinitrd,
+    Lsmod,
+    Modinfo,
+    Modprobe,
+    Uname,
+    Pgrep,
+)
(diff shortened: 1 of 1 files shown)


mcgov:
realizing this is not sufficient since the reset call may run for a long time and relies on the test output, so the test will need modification to pull the nohup output.

mcgov:
I just noticed this test is *only* running the reload test for hv_netvsc too

microsoft/testsuites/core/hv_module.py
```diff
@@ -234,24 +244,29 @@ def reload_hyperv_modules(self, case_name: str, log: Logger, node: Node) -> None
...
             ("for i in $(seq 1 %i); do " % loop_count)
             + f"modprobe -r -v {module}; modprobe -v {module}; "
             "done; sleep 1; "
             "ip link set eth0 down; ip link set eth0 up; dhclient eth0",
             sudo=True,
             shell=True,
+            nohup=True,
         )
+        pgrep = node.tools[Pgrep]
+        while pgrep.get_processes("nohup"):
```
squirrelsc, line 257: Is that possible to have multiple nohup processes? How about return the nohup process, if nohup=True?

mcgov: nohup won't be called by the OS as far as I know, it's for remote users running commands. It's possible there could be a lingering nohup process from another test if it hasn't been torn down correctly though. That's a sticky problem 🤔
Thank you for the note, I've marked this as a draft and I'm going to run some tests to verify this works as expected.

LiliDeng: during run reload module and down and up eth0 in the background, the pgrep may fail to execute.

PR #2396, opened 2022-11-15
Remove dhclient on bridge interfaces

dsrivastavv:
This should nested VM errors on Ubuntu 18.04 and Ubuntu 16.04


LiliDeng (dismissed):

LiliDeng:
@somil55 could you please check there is no regression for case validate_network_interface_reload_via_ip_link?

LiliDeng (approved):

PR #2398, opened 2022-11-15
Support attribute in doc generation

avylove:
This is a fix to support #2397 which broke doc generation. There is a lot more cleanup to do here and it looks like the generated values aren't even used, but this will get docs working again.


LiliDeng:
@avylove thanks! any idea for 6 build in `Expected — Waiting for status to be reported` status?

avylove:
> @avylove Avram Lubkin FTE thanks! any idea for 6 build in `Expected — Waiting for status to be reported` status?
Those are probably set as required checks in settings, but they don't exist anymore. I don't think I have access to check.

squirrelsc:
> CI Workflow

Removed from policy

squirrelsc (approved):

PR #2402, opened 2022-11-15
Create upgrade_installer transformer

kamalca:
This transformer will be used to update Ubuntu packages to test proposed repositories in a new pipeline.

### DIFF
--- a/lisa/transformers/upgrade_packages.py
+++ b/lisa/transformers/upgrade_packages.py
@@ -15,22 +15,26 @@ from lisa.util.logger import Logger, get_logger
 @dataclass_json()
 @dataclass
 class UpgradeInstallerSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
-    ...
+    repo_url: str = field(
+        default="http://azure.archive.ubuntu.com/ubuntu/",
+        metadata=field_metadata(required=True),
+    )
+
+    proposed: bool = field(
+        default=True,
+        metadata=field_metadata(required=True),
+    )
 
 
 @dataclass_json
 @dataclass
-class UpgradeInstallerTransformerSchema(schema.Transformer):
+class UpgradeTransformerSchema(schema.Transformer):
     # SSH connection information to the node
     connection: Optional[schema.RemoteNode] = field(
         default=None, metadata=field_metadata(required=True)
     )
-    repo_updater: Optional[UpgradeInstallerSchema] = field(
-        default=None, metadata=field_metadata(required=False)
-    )
-    # installer's parameters.
     installer: Optional[UpgradeInstallerSchema] = field(
-        default=None, metadata=field_metadata(required=False)
+        default=None, metadata=field_metadata(required=True)
     )
 
 
@@ -54,21 +58,21 @@ class UpgradeInstaller(subclasses.BaseClassWithRunbookMixin):
         raise NotImplementedError()
 
 
-class UpgradeInstallerTransformer(Transformer):
+class UpgradeTransformer(Transformer):
     @classmethod
     def type_name(cls) -> str:
-        return "upgrade_installer"
+        return "upgrade"
 
     @classmethod
     def type_schema(cls) -> Type[schema.TypedSchema]:
-        return UpgradeInstallerTransformerSchema
+        return UpgradeTransformerSchema
 
     @property
     def _output_names(self) -> List[str]:
         return []
 
     def _internal_run(self) -> Dict[str, Any]:
-        runbook: UpgradeInstallerTransformerSchema = self.runbook
+        runbook: UpgradeTransformerSchema = self.runbook
         assert runbook.connection, "connection must be defined."
         assert runbook.installer, "installer must be defined."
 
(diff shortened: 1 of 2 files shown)


lisa/transformers/upgrade_packages.py
```diff
@@ -0,0 +1,170 @@
...
+                    sudo=True,
+                    shell=True,
+                    expected_exit_code=0,
+                    expected_exit_code_failure_message=(
+                        "fail to make folder /var/cache/apt/archives/partial"
+                    ),
+                )
+            else:
+                node.os.install_packages(["debian-keyring", "debian-archive-keyring"])
+        node.execute(
```
squirrelsc, line 170: I think most methods are implemented in os and tools. The transformers are higher level classes, so they can reference tools and os as wanted.

lisa/transformers/upgrade_packages.py
```diff
@@ -0,0 +1,170 @@
...
+    @classmethod
+    def type_name(cls) -> str:
+        return "unattended_upgrade"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return BaseInstallerSchema
+
+    def validate(self) -> None:
+        assert isinstance(self._node.os, Debian), (
```
squirrelsc, line 116: How about make sure it's available for all distros? But so far, we just implement for Debian.

lisa/transformers/upgrade_packages.py
```diff
@@ -0,0 +1,170 @@
...
+from lisa.operating_system import Debian
+from lisa.tools import Uname
+from lisa.transformer import Transformer
+from lisa.util import field_metadata, subclasses
+from lisa.util.logger import Logger, get_logger
+
+
+@dataclass_json()
+@dataclass
+class BaseInstallerSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
```
squirrelsc: If nothing to abstract at this layer, it doesn't need to declare a baseinstaller layer. The kernel installer derives to package installer and source installer. For this repo installer, it doesn't need to have the base installer part. And the same "BaseInstaller" name is confusing.

kamalca: I changed some class names and schema names.

lisa/transformers/upgrade_packages.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+@dataclass_json
+@dataclass
+class UpgradeInstallerTransformerSchema(schema.Transformer):
+    # SSH connection information to the node
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    repo_updater: Optional[UpgradeInstallerSchema] = field(
```
squirrelsc: haven't seen the reason to split repo updater and upgrade into two parts.  so the two repo_updater and installer are not necessary.

lisa/transformers/upgrade_packages.py
```diff
@@ -0,0 +1,215 @@
...
+
+@dataclass_json()
+@dataclass
+class UpgradeInstallerSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
+    ...
+
+
+@dataclass_json
+@dataclass
+class UpgradeInstallerTransformerSchema(schema.Transformer):
```
squirrelsc: How about naming it like "UpgradeTransformerSchema"?

lisa/transformers/upgrade_packages.py
kamalca: The pipeline will have two stages, one where the transformer only changed the repo source than tests will be run (such as the dns servicing test) which updates packages and tests functionality without a reboot.

The second stage will update all packages then run a full LISA validation.

Since it is a common scenario for customers to have package updates without rebooting the VM, we want to test that functionality as well. Splitting up the repo update and the package upgrade will help us test these two different scenarios.

squirrelsc: The kernel installation splits to two stages, because 1) every time, it needs to run both, 2) each stage has different implementations.

For this case, the two stages can be two different transformers, like 1) add custom repo, 2) update packages. The rebooting can be a setting in update packages, so it can be skipped.

kamalca: I can split this into two transformers, I just thought that updating the repo list and upgrading packages was closely related. I imagine that both stages could have different implementations. You might want to remove repos, add a custom repo, add a ppa repo. You might want to upgrade specific packages, remove packages, etc.

Whether or not a reboot is done at the end of the transformer doesn't really matter, because the image is exported as a vhd and then rebooted. 
The only way to test the "upgrade and don't reboot" scenario is to have the upgrades be part of the test. As far as I know, we have no way to ensure a clean environment after every test, and to restrict the tests to ones that don't perform a reboot in the test.

kamalca:
@squirrelsc based on our discussions, I think I will simplify the transformer. It will be a package upgrade transformer with an option for upgrading from proposed. Then to achieve the different test scenarios, we will have to add functionality to LISA to be able to optionally run the transformer for every VM deployment.

Please let me know if you have any comments on this approach.

squirrelsc:
> @squirrelsc based on our discussions, I think I will simplify the transformer. It will be a package upgrade transformer with an option for upgrading from proposed. Then to achieve the different test scenarios, we will have to add functionality to LISA to be able to optionally run the transformer for every VM deployment.
> 
> Please let me know if you have any comments on this approach.

LGTM

squirrelsc (approved):

PR #2404, opened 2022-11-16
Initial Pylint configuration

avylove:
This adds a Nox session for Pylint and adds a test to the CI workflow. All failing checks have been disabled with the intent that they will be reenabled as the issues are resolved.


LiliDeng:
@squirrelsc could you please take a look?

squirrelsc (approved):

PR #2406, opened 2022-11-17
azure_deploy: make os disk size configurable through runbook

anirudhrb:
Move `osdisk_size_in_gb` up to `AzureNodeSchema` from `AzureNodeArmParameter` so that it can be configured from the runbook. If this property is set in the runbook, it will take precedence over the values computed from the VHD/marketplace image.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -118,6 +118,10 @@ class AzureNodeSchema:
     # It decides the real computer name. It cannot be too long.
     short_name: str = ""
     vm_size: str = ""
+    # Specifies the minimum OS disk size. The size of the disk that gets provisioned
+    # may be larger than this depending on other requirements set by VHD, marketplace
+    # image etc but it will never be smaller.
+    osdisk_size_in_gb: int = 30
     # Force to maximize capability of the vm size. It bypass requirements on
     # test cases, and uses to force run performance tests on any vm size.
     maximize_capability: bool = False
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1153,16 +1153,18 @@ def _create_node_arm_parameters(
         arm_parameters = AzureNodeArmParameter.from_node_runbook(runbook)
 
         os_disk_size = 30
-        if arm_parameters.vhd:
+        if arm_parameters.osdisk_size_in_gb is not None:
+            pass
+        elif arm_parameters.vhd:
```
pupacha: If both `osdisk_size_in_gb` and `vhd` are defined in runbook, this logic might fail to run `_get_deployable_vhd_path` which I believe is required for vhd deployments.

pupacha (dismissed):

lisa/sut_orchestrator/azure/common.py
```diff
@@ -118,6 +118,7 @@ class AzureNodeSchema:
     # It decides the real computer name. It cannot be too long.
     short_name: str = ""
     vm_size: str = ""
+    osdisk_size_in_gb: int = 30
```
squirrelsc, line 124: Add a comment above to explain it's the min size. If there are other requirements on osdisk size, it will be bigger.

anirudhrb: Done.

LiliDeng: to be frank, I am still confused from the comment.

anirudhrb: Which part are you confused about?

anirudhrb: @LiliDeng ^^^

LiliDeng (approved):

PR #2408, opened 2022-11-17
Fio l2 guest vm changes

smit-gardhariya:
This PR is raised to update RAM for L2 guest vm we spin up to test FIO testcases. We are using ubuntu image for L2 guest. Also, timeout is increased for apt-update command which run before installing FIO.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -786,7 +786,7 @@ class Debian(Linux):
     def _initialize_package_installation(self) -> None:
         # wait running system package process.
         self.wait_running_package_process()
-        result = self._node.execute("apt-get update", sudo=True)
+        result = self._node.execute("apt-get update", sudo=True, timeout=1800)
         if self._repo_not_exist_pattern.search(result.stdout):
             raise RepoNotExistException(self._node.os)
         result.assert_exit_code(message="\n".join(self.get_apt_error(result.stdout)))
--- a/microsoft/testsuites/performance/storageperf.py
+++ b/microsoft/testsuites/performance/storageperf.py
@@ -24,7 +24,7 @@ from lisa.features.network_interface import Sriov, Synthetic
 from lisa.messages import DiskSetupType, DiskType
 from lisa.node import RemoteNode
 from lisa.operating_system import SLES, Debian, Redhat
-from lisa.testsuite import TestResult
+from lisa.testsuite import TestResult, node_requirement
 from lisa.tools import FileSystem, Lscpu, Mkfs, Mount, NFSClient, NFSServer, Sysctl
 from lisa.util import SkippedException
 from microsoft.testsuites.performance.common import (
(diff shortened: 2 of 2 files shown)


lisa/operating_system.py
```diff
@@ -820,7 +820,7 @@ def _install_packages(
                 f"dpkg -i {' '.join(file_packages)}", sudo=True, timeout=timeout
             )
             # after install package, need update the repo
-            self._initialize_package_installation()
+            self._initialize_package_installation(timeout=1800)
```
LiliDeng: there is no parameter timeout for method _initialize_package_installation

smit-gardhariya: Yes i removed it new push.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -190,6 +190,11 @@ def perf_storage_over_nfs_synthetic_udp_4k(self, result: TestResult) -> None:
         """,
         priority=3,
         timeout=TIME_OUT,
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                memory_mb=search_space.IntRange(min=4 * 1024),
```
anirudhrb: How did you arrive at this number?

anirudhrb:
This PR sets the memory requirement for FIO irrespective of whether it is running in L2 or L1. So please update your commit message and title.

Also, since these are two unrelated changes, it's better to have two separate well defined commits in this PR.

microsoft/testsuites/performance/storageperf.py
smit-gardhariya: There is no system resource suggestion on [FIO Man page ](https://linux.die.net/man/1/fio) and FIO [readdocs](https://fio.readthedocs.io/en/latest/fio_doc.html). We have faces OOM with 512 MB memory. We deploy host azure VM with 64 GB so i kept 4GB requirement where FIO is running.

squirrelsc: > There is no system resource suggestion on [FIO Man page ](https://linux.die.net/man/1/fio) and FIO [readdocs](https://fio.readthedocs.io/en/latest/fio_doc.html). We have faces OOM with 512 MB memory. We deploy host azure VM with 64 GB so i kept 4GB requirement where FIO is running.

Good explanation, please add it to test case description.

smit-gardhariya: @anirudhrb  I will make RAM 2 GB as we discussed

smit-gardhariya: > There is no system resource suggestion on [FIO Man page ](https://linux.die.net/man/1/fio) and FIO [readdocs](https://fio.readthedocs.io/en/latest/fio_doc.html). We have faces OOM with 512 MB memory. We deploy host azure VM with 64 GB so i kept 4GB requirement where FIO is running.

Sure Chi

smit-gardhariya: @anirudhrb  @squirrelsc I have made changes as per our discussion

@LiliDeng I made changes for commit history

@anirudhrb Thnx for the help

LiliDeng (approved):

PR #2410, opened 2022-11-17
Unpin versions for CI tools

avylove:
In order to avoid stale tools and to catch issues early, CI tools should pull the latest version.

This removes version pinning for dev and CI tools and fixes typing errors raised in newer versions of mypy.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -3,7 +3,7 @@
 
 import re
 import sys
-from dataclasses import InitVar, dataclass, field
+from dataclasses import dataclass, field
 from datetime import datetime, timedelta
 from threading import Lock
 from time import sleep
@@ -145,11 +145,12 @@ class AzureNodeSchema:
     # image.
     is_linux: Optional[bool] = None
 
-    _marketplace: InitVar[Optional[AzureVmMarketplaceSchema]] = None
+    def __post_init__(self) -> None:
 
-    _shared_gallery: InitVar[Optional[SharedImageGallerySchema]] = None
+        # Caching
+        self._marketplace: Optional[AzureVmMarketplaceSchema] = None
+        self._shared_gallery: Optional[SharedImageGallerySchema] = None
 
-    def __post_init__(self, *args: Any, **kwargs: Any) -> None:
         # trim whitespace of values.
         strip_strs(
             self,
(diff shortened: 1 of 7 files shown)


lisa/util/__init__.py
```diff
@@ -466,18 +464,9 @@ def find_group_in_lines(
...
-        or isinstance(dest, bool)
-        or isinstance(dest, float)
-        or isinstance(dest, str)
-    ):
-        result = dest
-    else:
-        result = dest.copy()
 
-    if isinstance(result, dict):
+    if isinstance(dest, dict):
```
squirrelsc, line 468: This change may have regression. If the dest is int, bool, float or str, the changed logic will take src, instead of dest.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -171,80 +172,78 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
...
     @property
     def marketplace(self) -> Optional[AzureVmMarketplaceSchema]:
-        # this is a safe guard and prevent mypy error on typing
-        if not hasattr(self, "_marketplace"):
-            self._marketplace: Optional[AzureVmMarketplaceSchema] = None
-        marketplace: Optional[AzureVmMarketplaceSchema] = self._marketplace
-        if not marketplace:
-            if isinstance(self.marketplace_raw, dict):
+
+        if self._marketplace is not None:
```
squirrelsc, line 176: The marketplace object is used to make sure the `self._marketplace` is a final state, instead of an intermedia state. The value may be run at the same time. There is a race condition to get incomplete object. So it needs the local variable `marketplace` to get the whole object set correctly, then assign to `self._marketplace`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -145,11 +145,12 @@ class AzureNodeSchema:
     # image.
     is_linux: Optional[bool] = None
 
-    _marketplace: InitVar[Optional[AzureVmMarketplaceSchema]] = None
+    def __post_init__(self) -> None:
```
squirrelsc, line 148: The `__post_init__` may not be called, when creating the object in code. I'm not sure the logic, but there was some reason to declare them like InitVar, instead of put in `__post_init__`.

squirrelsc: Please try if it works with an Azure run.

lisa/util/__init__.py
avylove: The behavior hasn't changed. In the old logic, if `dest` was an `int`, `bool`, `float` or `str`, then `result` would not be a `dict`, causing `result` to fall back to `src`. This was all just useless logic.

lisa/sut_orchestrator/azure/common.py
avylove: `__post_init__()` is called by the `__init__()` method generating the data classes. This is where instance variables like this should go. `InitVar` types are passed to `__post_init__()`, but there was no logic for them, so that implementation was incomplete. It looks like they were just done that way to make mypy happy, but the logic in mypy changed and it was causing errors.

Kicked of an azure test local and it ran fine

squirrelsc (approved):

squirrelsc:
Let's merge it next Monday.

LiliDeng:
@squirrelsc is it right time to merge it? any testing need to do done before merging it?

squirrelsc:
> @squirrelsc is it right time to merge it? any testing need to do done before merging it?

It's ok to merge it today or next week.

squirrelsc (changes requested):
We can upgrade versions of tool, but not unpin them. The CI pipeline is being broken by flake8 upgrading.

avylove:
> We can upgrade versions of tool, but not unpin them. The CI pipeline is being broken by flake8 upgrading.

I don't see how the internal CI pipeline is related, the version used there hasn't changed. It's best to unpin so these things get caught early. Otherwise you have what we currently have with out-of-date validation tools.

squirrelsc:
> > We can upgrade versions of tool, but not unpin them. The CI pipeline is being broken by flake8 upgrading.
> 
> I don't see how the internal CI pipeline is related, the version used there hasn't changed. It's best to unpin so these things get caught early. Otherwise you have what we currently have with out-of-date validation tools.

I haven't said it happened on internal CI pipelines. But if we use the same way to unpin upstream packages in internal pipelines, that will happen someday. And we need to comeback to discuss and investigate in holidays again. The flake8 issue is more like a regression, instead of a new feature. It blocks me to merge some PRs. Even flake8 brings a new validation, it applies to existing code as a debt, which should be fixed by maintainers, instead of current PRs' contributors. I have to wait it's fixed. Out-of-date validation tools are cost, which can be accepted, but build breaks are not.

avylove:
> > > We can upgrade versions of tool, but not unpin them. The CI pipeline is being broken by flake8 upgrading.
> > 
> > 
> > I don't see how the internal CI pipeline is related, the version used there hasn't changed. It's best to unpin so these things get caught early. Otherwise you have what we currently have with out-of-date validation tools.
> 
> I haven't said it happened on internal CI pipelines. But if we use the same way to unpin upstream packages in internal pipelines, that will happen someday. And we need to comeback to discuss and investigate in holidays again. The flake8 issue is more like a regression, instead of a new feature. It blocks me to merge some PRs. Even flake8 brings a new validation, it applies to existing code as a debt, which should be fixed by maintainers, instead of current PRs' contributors. I have to wait it's fixed. Out-of-date validation tools are cost, which can be accepted, but build breaks are not.

It's all one in the same and unpinning has the lowest maintenance cost. What you're describing is signal, which is what we want. If someone submits a PR and gets and error unrelated to their PR, they should be able to identify it. We fix it quickly or pin temporarily. The alternative is we need to come up with a system to automatically generate PRs for the next version. It's too much work to do that way which is why the standard way is to have no upper bound pinning.

avylove:
Closing this since it's now handled by BumpDeps.
Typing changes for newer versions of MyPY merged into https://github.com/microsoft/lisa/pull/2545

PR #2412, opened 2022-11-18
Add features to ready branch

dsrivastavv:


lisa/sut_orchestrator/ready.py
```diff
@@ -18,7 +19,22 @@ def type_name(cls) -> str:
 
     @classmethod
     def supported_features(cls) -> List[Type[Feature]]:
-        return []
+        return [
+            features.Disk,
+            features.Gpu,
+            features.Nvme,
+            features.NestedVirtualization,
+            features.SerialConsole,
```
squirrelsc: Don't add all features, because some of them are not supported natively like Serial console, resize, security profile, start stop.

dsrivastavv: If we don't add Serial Console, then smoke test can't run

dsrivastavv: nvm. seems like it can't run it anyways

squirrelsc (approved):

PR #2417, opened 2022-11-21
[fix issues] wait for a while for waagent write device timeout value

LiliDeng:


microsoft/testsuites/core/storage.py
```diff
@@ -70,9 +70,19 @@ def verify_disks_device_timeout_setting(
...
+                device_timeout_from_distro = int(
+                    node.tools[Cat]
+                    .run(f"/sys/block/{disk}/device/timeout", force_run=True)
+                    .stdout
+                )
+                if root_device_timeout_from_waagent == device_timeout_from_distro:
+                    break
+                else:
+                    time.sleep(1)
+                    continue
```
squirrelsc: It doesn't need `continue` at the end of while loop.

LiliDeng: update

squirrelsc (approved):

PR #2418, opened 2022-11-21
Improve assert message for lscpu get_cpu_info

dsrivastavv:


lisa/tools/lscpu.py
```diff
@@ -230,11 +230,15 @@ def get_cpu_info(self) -> List[CPUInfo]:
         result = self.run("--extended=cpu,node,socket,cache").stdout
```
LiliDeng: should we assert the exit code?

LiliDeng:
@somil55 please have a check on this PR when you have time, thanks!

lisa/tools/lscpu.py
dsrivastavv: I gave it another thought, and I don't think we need assert exit code here. `lscpu` is a standard linux command and I haven't seen any failures on this command.

squirrelsc: > I gave it another thought, and I don't think we need assert exit code here. `lscpu` is a standard linux command and I haven't seen any failures on this command.

If it always succeeds, assert exit code should be fine. We cannot say something always right, so it doesn't need to be verified.

LiliDeng (approved):

PR #2419, opened 2022-11-21
DPDK: check vpp service before stop call

mcgov:
Add check if vpp service exists before calling stop_service, avoid possible assertion during cleanup.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -663,7 +663,10 @@ class Dpdk(TestSuite):
         for node in environment.nodes.list():
             modprobe = node.tools[Modprobe]
             if modprobe.module_exists("uio_hv_generic"):
-                node.tools[Service].stop_service("vpp")
+                # stop_service can assert if service doesn't exist
+                if node.tools[Service].check_service_exists("vpp"):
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -663,7 +663,10 @@ def after_case(self, log: Logger, **kwargs: Any) -> None:
         for node in environment.nodes.list():
             modprobe = node.tools[Modprobe]
             if modprobe.module_exists("uio_hv_generic"):
-                node.tools[Service].stop_service("vpp")
+                # stop_service can assert if service doesn't exist
+                if node.tools[Service].check_service_exists("vpp"):
+                    node.tools[Service].stop_service("vpp")
```
LiliDeng, line 668: in stop service method, it will check whether this service is running or not, just stop it when it is in running status.
Probably you need check the exception when restart vpp service. https://github.com/microsoft/lisa/blob/main/microsoft/testsuites/dpdk/dpdkvpp.py#L59

```failed. AssertionError: [ get unexpected exit code on cmd ['sudo', 'sh', '-c', 'systemctl restart vpp']] Expected <[0, 0]> to contain item <5>, but did not.```

PR #2420, opened 2022-11-21
DPDK VPP: bugfixes for new versions and SUSE

mcgov:
- SLES doesn't have vpp packages available for some versions, so check if there are none after adding the repository and refreshing. Skip if there is not a supported package available.

- Fix interface check to look for 'VSC' as well as 'VirtualFunction' since names can differ between versions.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkvpp.py
+++ b/microsoft/testsuites/dpdk/dpdkvpp.py
@@ -73,6 +73,7 @@ class DpdkVpp(Tool):
         vpp_detected_interface = (
             "GigabitEthernet" in vpp_interface_output
             or "VirtualFunctionEthernet" in vpp_interface_output
+            or "VSC" in vpp_interface_output
         )
         assert_that(vpp_detected_interface).described_as(
             "VPP did not detect the dpdk VF or Gigabit network interface"
@@ -104,7 +105,18 @@ class DpdkVpp(Tool):
                 "Could not install vpp with fdio provided installer"
             ),
         )
-        node.os.update_packages("")
+        node.os.get_repositories()
+        package_available = node.os.is_package_in_repo("vpp")
+        if not package_available:
+            raise SkippedException(
+                UnsupportedDistroException(
+                    node.os,
+                    "VPP package was not available in repository after adding "
+                    "VPP repo. This OS is likely not supported without a "
+                    "source build.",
+                )
+            )
+
         self._install_from_package_manager()
         return True
 


microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -104,7 +105,18 @@ def _install(self) -> bool:
                 "Could not install vpp with fdio provided installer"
             ),
         )
-        node.os.update_packages("")
+        node.os.get_repositories()
```
squirrelsc, line 108: What this line actually do? If it's to trigger some inner actions, the actions should be exposed to get logic clear.

mcgov: it calls `apt update` / `zypper refresh`  etc.

mcgov: The old way of calling update_packages with no argument was a hack, this other call is the interfaces proper way to update the current set of repositories.

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -104,7 +105,18 @@ def _install(self) -> bool:
...
+        package_available = node.os.is_package_in_repo("vpp")
+        if not package_available:
```
LiliDeng, line 110: I think it has already been covered in is_package_in_repo?

```
    def is_package_in_repo(self, package: Union[str, Tool, Type[Tool]]) -> bool:
        """
        Query if a package/tool exists in the repo
        Return Value - bool
        """
        if self._first_time_installation:
            self._initialize_package_installation()
            self._first_time_installation = False
        return self._is_package_in_repo(package_name)
```

microsoft/testsuites/dpdk/dpdkvpp.py
LiliDeng: I think it has already been covered in is_package_in_repo?

```
    def is_package_in_repo(self, package: Union[str, Tool, Type[Tool]]) -> bool:
        """
        Query if a package/tool exists in the repo
        Return Value - bool
        """
        if self._first_time_installation:
            self._initialize_package_installation()
            self._first_time_installation = False
        return self._is_package_in_repo(package_name)
```

microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -73,6 +73,7 @@ def run_test(self) -> None:
         vpp_detected_interface = (
             "GigabitEthernet" in vpp_interface_output
             or "VirtualFunctionEthernet" in vpp_interface_output
+            or "VSC" in vpp_interface_output
```
LiliDeng, line 76: Can you list the full outputs in the comment? I am not sure whether VSC_net is sriov type or not.

mcgov: The interface type when I tested last on SUSE was NetVSC0. I am pretty sure this corresponds to the net_vdev_netvsc PMD in dpdk, a newer PMD which makes use of the default hv_netvsc failover configuration. This would be the correct way to bind to that interface, older tests would have used net_failsafe since it was the only option but on DPDK >19.11 net_vdev_netvsc would be the way to go.

microsoft/testsuites/dpdk/dpdkvpp.py
mcgov: I don't think that is guaranteed to work if another package has been installed before adding the new repository info. Since we're adding a new repository with the install script, the instructions suggest running apt update/zypper renew etc. to make sure the repo info is added correctly.

In the event the test changes, a tool is installed which installs a package, or a dependency changes, I think it's better to explicitly ask for the repository cache update rather than rely on the side effect of the package_in_repo function.

microsoft/testsuites/dpdk/dpdkvpp.py
mcgov: (running some more tests to verify this)

LiliDeng:
@mcgov please update PR when you have time, thanks!

LiliDeng:
> @mcgov please update PR when you have time, thanks!

@mcgov could you please update this PR?

LiliDeng:
@mcgov feel free to create new one or reopen it when it is ready, thanks

PR #2421, opened 2022-11-22
Use compat mode editable install for dev environment

avylove:
This forces the .pth file created by the lisa editable install to be the legacy static format rather than the current dynamic format. This is to resolve a path issue for non-package directories in the root of lisa repot like microsoft. This is a workaround to restore the previous behavior provided by poetry. This only worked with poetry because of a quirk in the order of path evaluation and this behavior is not guaranteed by the Python API, so we'll need to prioritize moving these non-package directories.

run `nox -vs dev` after installing to update the dev virtual environment

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -180,6 +180,9 @@ def dev(session: nox.Session) -> None:
         "install",
         "--editable",
         f".[{extras}]",
+        # Workaround for non-package directories (microsoft, examples)
+        "--config-settings",
+        "editable_mode=compat",
         external=True,
     )
 


LiliDeng (approved):

LiliDeng (changes requested):
need more testing

LiliDeng (approved):

squirrelsc:
If it happens on old cloned code, can we provide steps in troubleshooting doc to fix it?

avylove:
> If it happens on old cloned code, can we provide steps in troubleshooting doc to fix it?

Are you asking for a way to fix it without updating to a release with this patch? Or for a way to fix it once you have this patch locally?

squirrelsc:
> > If it happens on old cloned code, can we provide steps in troubleshooting doc to fix it?
> 
> Are you asking for a way to fix it without updating to a release with this patch? Or for a way to fix it once you have this patch locally?

Either way is ok, so it doesn’t need to take this fix in code.

avylove:
> Either way is ok, so it doesn’t need to take this fix in code.

To fix it without this patch, just run `pip install -e .[azure,libvirt] --config-settings editable_mode=compat` with the virtual environment activated.

squirrelsc:
> > Either way is ok, so it doesn’t need to take this fix in code.
> 
> To fix it without this patch, just run `pip install -e .[azure,libvirt] --config-settings editable_mode=compat` with the virtual environment activated.

Cool, is it one time work? If so, can it be added in troubleshooting doc?

avylove:
> > > Either way is ok, so it doesn’t need to take this fix in code.
> > 
> > 
> > To fix it without this patch, just run `pip install -e .[azure,libvirt] --config-settings editable_mode=compat` with the virtual environment activated.
> 
> Cool, is it one time work? If so, can it be added in troubleshooting doc?

Yes, unless they update the dev virtual environment before having this patch. I don't think it needs to be in the troubleshooting guide because it's only a temporary workaround. What we want is for them to have the patch and run `tox -vs dev`

squirrelsc:
Just understood, it happens on new cloned code to, so get it in.

PR #2422, opened 2022-11-23
[improvement] download serial console every 5 mins during VM provision

LiliDeng:


mcgov:
👏👏👏👏
Excellent idea!

lisa/util/__init__.py
```diff
@@ -243,6 +243,14 @@ def __str__(self) -> str:
         return format_str
 
 
+class InternalTimeoutException(LisaException):
```
squirrelsc: `LisaTimeoutException`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -535,6 +540,42 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
             else:
                 log.debug("not wait deleting")
 
+    def save_console_log(
```
squirrelsc: It copies too many code. Try to run initialize_environment earlier and use serial console feature.

lisa/util/__init__.py
LiliDeng: fixed

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: feature was not initialized at that time; I moved the common login into method into common.py file. Let know any concern.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -535,6 +538,21 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
             else:
                 log.debug("not wait deleting")
 
+    def _save_console_log(
+        self, resource_group_name: str, environment: Environment, log: Logger
+    ) -> None:
+        compute_client = get_compute_client(self)
+        vms = compute_client.virtual_machines.list(resource_group_name)
+        for vm in vms:
+            saved_path = environment.log_path / get_datetime_path() / vm.name
```
squirrelsc: 1. don't call get_datetime_path multiple times in the loop, it will be different each time, it's not expected.
2. Don't make another level of folders of vm name. It doesn't need, because there are only one or two files for each vm.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1036,6 +1048,52 @@ def delete_file_share(
     share_service_client.delete_share(file_share_name)
 
 
+def get_diagnostic_data(
```
squirrelsc: It's not used outside of this module, make it private.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -535,6 +538,21 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
...
+            saved_path.mkdir(parents=True, exist_ok=True)
+            log_response_content = save_console_log(
+                resource_group_name, vm.name, self, log, saved_path
+            )
+            log_file_name = saved_path / "serial_console.log"
```
squirrelsc: put vm name in file name.

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -535,6 +538,26 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
             else:
                 log.debug("not wait deleting")
 
+    def _save_console_log(
+        self, resource_group_name: str, environment: Environment, log: Logger
+    ) -> None:
+        compute_client = get_compute_client(self)
+        vms = compute_client.virtual_machines.list(resource_group_name)
+        saved_path = environment.log_path / get_datetime_path()
```
squirrelsc: this is clearer to know, before entering the folder, `environment.log_path / f“{get_datetime_path()}_serial_log"`

lisa/util/__init__.py
squirrelsc: Not seeing it's fixed...

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1255,7 +1282,15 @@ def _deploy(
             deployment_operation = deployments.begin_create_or_update(
                 **deployment_parameters
             )
-            wait_operation(deployment_operation, failure_identity="deploy")
+            while True:
+                try:
+                    wait_operation(
+                        deployment_operation, time_out=480, failure_identity="deploy"
```
squirrelsc: Why not use 300 seconds, it's easier to understand...

lisa/util/__init__.py
LiliDeng: sorry, seems I missed this one, now it is fixed.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1036,6 +1048,40 @@ def delete_file_share(
...
+            f.write(screenshot_response.content)
+        try:
+            with Image.open(screenshot_raw_name) as image:
+                image.save(screenshot_name, "PNG", optimize=True)
+        except UnidentifiedImageError:
+            log.debug(
+                "The screenshot is not generated. "
+                "The reason may be the VM is not started."
+            )
+        unlink(screenshot_raw_name)
```
avylove: Simple changes to leverage pathlib Paths

```suggestion
    if saved_path:
        screenshot_raw_name = saved_path / f"{screenshot_file_name}.bmp"
        screenshot_response = requests.get(diagnostic_data.console_screenshot_blob_uri)
        screenshot_raw_name.write_bytes(screenshot_response.content)

        try:
            with Image.open(screenshot_raw_name) as image:
                image.save(saved_path / f"{screenshot_file_name}.png", "PNG", optimize=True)
        except UnidentifiedImageError:
            log.debug(
                "The screenshot is not generated. "
                "The reason may be the VM is not started."
            )
        screenshot_raw_name.unlink()
```

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1036,6 +1048,40 @@ def delete_file_share(
...
+    platform: "AzurePlatform",
+    log: Logger,
+    saved_path: Optional[Path],
+    screenshot_file_name: str = "serial_console",
+) -> bytes:
+    diagnostic_data = _get_diagnostic_data(platform, resource_group_name, vm_name)
+    if saved_path:
+        screenshot_raw_name = saved_path.joinpath(f"{screenshot_file_name}.bmp")
+        screenshot_name = saved_path.joinpath(f"{screenshot_file_name}.png")
+        screenshot_response = requests.get(diagnostic_data.console_screenshot_blob_uri)
```
avylove, line 1067: the response from `_get_diagnostic_data()` is manipulated twice before it's consumed and not used anywhere else, so why not move the call to requests and the content lookup below to `_get_diagnostic_data()` where it will be easy to add additional validation?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1036,6 +1048,40 @@ def delete_file_share(
...
+                image.save(screenshot_name, "PNG", optimize=True)
+        except UnidentifiedImageError:
+            log.debug(
+                "The screenshot is not generated. "
+                "The reason may be the VM is not started."
+            )
+        unlink(screenshot_raw_name)
+
+    log_response = requests.get(diagnostic_data.serial_console_log_blob_uri)
+    if 404 == log_response.status_code:
```
avylove: Convention is to have the value you are checking on the left
```suggestion
    if log_response.status_code == 404:
```

lisa/sut_orchestrator/azure/features.py
```diff
@@ -345,35 +343,15 @@ def _read(self) -> str:
...
-                    image.save(screenshot_name, "PNG", optimize=True)
-            except UnidentifiedImageError:
-                self._log.debug(
-                    "The screenshot is not generated. "
-                    "The reason may be the VM is not started."
-                )
-            unlink(screenshot_raw_name)
-
-        log_response = requests.get(diagnostic_data.serial_console_log_blob_uri)
+        log_response_content = save_console_log(
```
avylove: Why create a variable only to return it?

```suggestion
        return  save_console_log(
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -535,6 +538,26 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
...
+                resource_group_name,
+                vm.name,
+                self,
+                log,
+                saved_path,
+                screenshot_file_name=f"{vm.name}_serial_console",
+            )
+            log_file_name = saved_path / f"{vm.name}_serial_console.log"
+            with open(log_file_name, mode="wb") as f:
+                f.write(log_response_content)
```
avylove:
```suggestion
            log_file_name.write_bytes(log_response_content)
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1255,7 +1282,15 @@ def _deploy(
             deployment_operation = deployments.begin_create_or_update(
                 **deployment_parameters
             )
-            wait_operation(deployment_operation, failure_identity="deploy")
+            while True:
```
avylove, line 1284: Is there a maximum time on this loop?

lisa/util/__init__.py
```diff
@@ -243,6 +243,14 @@ def __str__(self) -> str:
         return format_str
 
 
+class LisaTimeoutException(LisaException):
+    """
+    This exception is used to indicate a timeout exception.
+    """
+
+    ...
```
avylove: You don't need an ellipsis here because you have a docstring.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: There is timeout in underlying code, and it raises another Exception.

lisa/sut_orchestrator/azure/common.py
squirrelsc: The `404 == ...` pattern is coming from C. It can avoid mistyped like `404 = ...`. It will cause a compiler error.

avylove: This not C. `if log_response = 404:` raises a syntax error in Python, so you don't need to do things like this. The Zen of Python says "Readability matter". I think this falls under that.

lisa/sut_orchestrator/azure/platform_.py
avylove: It should be obvious either through code or comments. This looks like it will keep running indefinitely until it fails or returns, there doesn't seem to be a maximum timeout on the loop.

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/features.py

lisa/util/__init__.py

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/common.py
LiliDeng: sorry, I didn't catch this, could you please explain it?

avylove: You created `_get_diagnostic_data()` to move logic out of this function, but then you have
`screenshot_response = requests.get(diagnostic_data.console_screenshot_blob_uri)` and `screenshot_response.content`. 
`_get_diagnostic_data()` hardly does anything, so why not rename it, `get_screenshot()` and have it call requests and only return the content? Then you can add additional error checking in case the requests call fails in that function.

avylove: Oh, sorry, I missed you were using it below. I would drop _get_diagnostic_data since it's only a couple lines of code and just merge it in here. You also need to add error checking for this requests call.

squirrelsc (approved):

PR #2424, opened 2022-11-24
Add sub-test-case support to JUnit notifier.

cwize1:
This change allows the JUnit notifier to output sub-test-case results. Unfortunately, the JUnit format doesn't sub-test-cases. So, the notifier adds the sub-test-cases as peers of the test-cases.

In addition, the JUnit notifier allows a test-case to start a sub-test- case without explictly completing it. Instead, if a new sub-test-case starts or the test-case completes, then the previous sub-test-case is closed out automatically. This avoids a test-case needing to add extensive try-catch blocks to ensure the sub-test-cases are closed out.

Other supporting changes:

1. Moved `stacktrace` and `is_completed()` from `TestResultMessage` to `TestResultMessageBase`. This allows sub-test-cases to report exception stacks.

2. Modify `create_test_result_message` so that it includes the time since the test-case started in every message. This is to allow the JUnit notifier to properly report how much time each test-case and sub-test-case took.

### DIFF
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -71,7 +71,12 @@ class TestResultMessageBase(MessageBase):
     name: str = ""
     status: TestStatus = TestStatus.QUEUED
     message: str = ""
+    stacktrace: Optional[str] = None
     information: Dict[str, str] = field(default_factory=dict)
+
+    @property
+    def is_completed(self) -> bool:
+        return _is_completed_status(self.status)
 
 
 @dataclass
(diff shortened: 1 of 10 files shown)


squirrelsc (dismissed):

squirrelsc:
Will merge later, once main branch break fixed.

squirrelsc:
> Will merge later, once main branch break fixed.

Build break on main branch fixed, please fix other failures.

cwize1:
> > Will merge later, once main branch break fixed.
> 
> Build break on main branch fixed, please fix other failures.

These have been fixed.

squirrelsc (approved):

PR #2425, opened 2022-11-24
Move get_resource_disk_mount_point to Features.Disks so testsuites can call it

lubaihua33:
By default, cloud-init will use /mnt as the mount point of the resource disk. In CentOS, cloud.cfg.d/91-azure_datasource.cfg customize mount point as /mnt/resource which can get from the Waagent config file.


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -237,12 +237,19 @@ def _check_supported(self, node: Node) -> None:
             ):
                 raise SkippedException("crashkernel=auto doesn't work for the distro.")
 
-    def _get_resource_disk_dump_path(self, node: Node) -> str:
-        if node.shell.exists(
-            PurePosixPath("/var/log/cloud-init.log")
-        ) and node.shell.exists(PurePosixPath("/var/lib/cloud/instance")):
+    def _get_resource_disk_dump_path(self, node: Node, log: Logger,) -> str:
+        # by default, cloudinit will use /mnt as mount point of resource disk
```
squirrelsc: I think this logic was handled somewhere, can it be reused?

lubaihua33: Done.

LiliDeng (dismissed):

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -21,7 +21,7 @@
 )
 from lisa.features import SerialConsole
 from lisa.operating_system import Redhat
-from lisa.sut_orchestrator.azure.tools import Waagent
+from lisa.sut_orchestrator.azure.common import get_resource_disk_mount_point
```
squirrelsc: The storage, kdump test cases are general test cases, they shouldn't directly depend on Azure. The resource disk method should be exposed by a feature like disk, so it can be used by normal test cases.

lubaihua33: > The storage, kdump test cases are general test cases, they shouldn't directly depend on Azure. The resource disk method should be exposed by a feature like disk, so it can be used by normal test cases.

Thank you @squirrelsc. Have moved to Features.Disks. Please review it again.

lisa/features/disks.py
```diff
@@ -2,12 +2,15 @@
...
 from functools import partial
+from pathlib import PurePosixPath
 from typing import Any, List, Optional, Type
 
 from assertpy.assertpy import assert_that
 
 from lisa import schema
 from lisa.feature import Feature
+from lisa.operating_system import CentOs
+from lisa.sut_orchestrator.azure.tools import Waagent
```
squirrelsc: It's not the right place. The `lisa/features` is neutral to platforms, so it shouldn't reference any platform. It can provide a method, and let platform to implement in subclasses.

lubaihua33: > It's not the right place. The `lisa/features` is neutral to platforms, so it shouldn't reference any platform. It can provide a method, and let platform to implement in subclasses.

Done

squirrelsc (approved):

PR #2426, opened 2022-11-25
Execute a command to check if the VM is boot up after kdump

lubaihua33:
Add _try_connect which wraps try_connect using func_set_timeout , using this way to avoid the stuck in try_connect function.

### DIFF
--- a/microsoft/testsuites/kdump/kdumpcrash.py
+++ b/microsoft/testsuites/kdump/kdumpcrash.py
@@ -3,7 +3,9 @@
 import time
 from pathlib import Path, PurePosixPath
 from random import randint
-from typing import cast
+from typing import Any, cast
+
+from func_timeout import FunctionTimedOut, func_set_timeout  # type: ignore
 
 from lisa import (
     LisaException,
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -328,6 +328,8 @@ def _kdump_test(self, node: Node, log_path: Path, log: Logger) -> None:
         except Exception as identifier:
             log.debug(f"ignorable ssh exception: {identifier}")
 
+        log.debug("Sleep 60s here to wait for the VM dump file and boot up")
+        time.sleep(60)
```
squirrelsc: Don't put a hard coded sleep. It makes the test case slow, and unstable. Wait some signals, and check regularly.

lubaihua33: While using try_connect or executing a command to check if the VM boots up, it might hit test case stuck then report "failed. TimeoutError: time out in 3600 seconds." in probability. Adding a sleep time to wait for the VM's crash kernel boots up and then executing the command can make this test more stable.

squirrelsc (dismissed):

LiliDeng:
@lubaihua33 please fix the flake8 error

lubaihua33:
> @lubaihua33 please fix the flake8 error

Done

LiliDeng (approved):

PR #2430, opened 2022-11-28
[fix issue] extend total wait time when encounter 'access denied by server' error during mount azure nfs share

LiliDeng:


lisa/tools/mount.py
```diff
@@ -68,7 +68,7 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
-    @retry(tries=10, delay=5)
+    @retry(tries=20, delay=6)
```
squirrelsc: It doesn't need to extend the delay. It's equivalent to tries=24 and delay=5. We need to make the test suite faster by trying faster.

LiliDeng: updated

squirrelsc (approved):

PR #2431, opened 2022-11-28
[fix issue] fix ping command fail to execute on freebsd

LiliDeng:


lisa/tools/ping.py
```diff
@@ -46,7 +46,18 @@ def ping_async(
     ) -> Process:
         if not target:
             target = INTERNET_PING_ADDRESS
-        args: str = f"{target} -c {count} -i {interval} -O"
+        args: str = ""
```
squirrelsc: refer to Windows in echo, 1) update create to supports FreeBSD like `if "FreeBSD" in node.os.name` in executable.py. 3) create a new class for FreeBSD.

LiliDeng: update, work like a charm

lisa/executable.py
```diff
@@ -164,6 +164,10 @@ def create(cls, node: Node, *args: Any, **kwargs: Any) -> Tool:
             windows_tool = cls._windows_tool()
             if windows_tool:
                 tool_cls = windows_tool
+        if "FreeBSD" in node.os.name:
```
squirrelsc: use `elif`, so the logic is clearer, and avoid strange problems.

LiliDeng: update

squirrelsc (dismissed):

PR #2433, opened 2022-11-29
kernel installer changes for CBLMariner-Dom0

smit-gardhariya:
These changes contain modification in kernel_installer transformer to add CBLMariner-Dom0 support. Also added sudo, cwd param to lisa tool Cp->copy() function.

We have modified the Git tool as well to accommodate auth_token so lisa can be use system token while we run it in pipeline where we need lisa to checkout the private azure repo

### DIFF
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -26,7 +26,12 @@ class BaseModifierSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
 @dataclass_json()
 @dataclass
 class BaseLocationSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
-    ...
+    kernel_config_file: str = field(
+        default="",
+        metadata=field_metadata(
+            required=False,
+        ),
+    )
 
 
 @dataclass_json()
@@ -38,13 +43,7 @@ class LocalLocationSchema(BaseLocationSchema):
             required=True,
         ),
     )
-    config_path: str = field(
-        default="",
-        metadata=field_metadata(
-            required=False,
-        ),
-    )
-
+    
 
 @dataclass_json()
 @dataclass
@@ -119,8 +118,8 @@ class SourceInstaller(BaseInstaller):
         # modify code
         self._modify_code(node=node, code_path=code_path)
 
-        config_path = source.get_config_path()
-        self._build_code(node=node, code_path=code_path, kconfig_path=config_path)
+        kconfig_file = source.get_kconfig_file()
+        self._build_code(node=node, code_path=code_path, kconfig_file=kconfig_file)
 
         self._install_build(node=node, code_path=code_path)
 
@@ -144,7 +143,7 @@ class SourceInstaller(BaseInstaller):
         )
         result.assert_exit_code()
 
-        if config_path:
+        if kconfig_file:
             # If it is dom0,
             # Name of the current kernel should be vmlinuz-<kernel version>
             uname = node.tools[Uname]
@@ -212,18 +211,18 @@ class SourceInstaller(BaseInstaller):
             self._log.debug(f"modifying code by {modifier.type_name()}")
             modifier.modify()
 
-    def _build_code(self, node: Node, code_path: PurePath, kconfig_path: str) -> None:
+    def _build_code(self, node: Node, code_path: PurePath, kconfig_file: str) -> None:
         self._log.info("building code...")
 
         uname = node.tools[Uname]
         kernel_information = uname.get_linux_information()
 
-        if kconfig_path:
-            kernel_config = code_path.joinpath(kconfig_path)
+        if kconfig_file:
+            kernel_config = code_path.joinpath(kconfig_file)
             err_msg = f"cannot find config path: {kernel_config}"
             assert node.shell.exists(kernel_config), err_msg
             result = node.execute(
-                f"cp {kconfig_path} .config",
+                f"cp {kconfig_file} .config",
                 cwd=code_path,
             )
             result.assert_exit_code()
@@ -362,7 +361,7 @@ class BaseLocation(subclasses.BaseClassWithRunbookMixin):
     def get_source_code(self) -> PurePath:
         raise NotImplementedError()
 
-    def get_config_path(self) -> str:
+    def get_kconfig_file(self) -> str:
         raise NotImplementedError()
 
 
@@ -409,9 +408,9 @@ class RepoLocation(BaseLocation):
 
         return code_path
 
-    def get_config_path(self) -> str:
+    def get_kconfig_file(self) -> str:
         runbook: RepoLocationSchema = self.runbook
-        return runbook.config_path
+        return runbook.kernel_config_file
 
 
 class LocalLocation(BaseLocation):
@@ -427,9 +426,9 @@ class LocalLocation(BaseLocation):
         runbook: LocalLocationSchema = self.runbook
         return self._node.get_pure_path(runbook.path)
 
-    def get_config_path(self) -> str:
+    def get_kconfig_file(self) -> str:
         runbook: LocalLocationSchema = self.runbook
-        return runbook.config_path
+        return runbook.kernel_config_file
 
 
 class BaseModifier(subclasses.BaseClassWithRunbookMixin):
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -43,7 +43,7 @@ class LocalLocationSchema(BaseLocationSchema):
             required=True,
         ),
     )
-    
+
 
 @dataclass_json()
 @dataclass
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -139,7 +139,7 @@ class SourceInstaller(BaseInstaller):
             destination_path = f"/boot/efi/{new_kernel_binary}"
             cp = node.tools[Cp]
             cp.copy(
-                src=PurePath(source_path),
+                src=source_path,
                 dest=PurePath(destination_path),
                 sudo=True
             )
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -50,6 +50,12 @@ class RepoLocationSchema(LocalLocationSchema):
     # fail the run if code exists
     fail_on_code_exists: bool = False
     cleanup_code: bool = False
+    auth_token: Optional[str] = field(
+        default=None,
+        metadata=field_metadata(
+            required=False,
+        ),
+    )
 
 
 @dataclass_json()
@@ -455,7 +461,10 @@ class RepoLocation(BaseLocation):
         self._log.info(f"cloning code from {runbook.repo} to {code_path}...")
         git = self._node.tools[Git]
         code_path = git.clone(
-            url=runbook.repo, cwd=code_path, fail_on_exists=runbook.fail_on_code_exists
+            url=runbook.repo,
+            cwd=code_path,
+            fail_on_exists=runbook.fail_on_code_exists,
+            auth_token=runbook.auth_token,
         )
 
         git.fetch(cwd=code_path)
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -218,7 +218,7 @@ class SourceInstaller(BaseInstaller):
         uname = node.tools[Uname]
         kernel_information = uname.get_linux_information()
 
-        if kconfig_path and len(kconfig_path) > 0:
+        if kconfig_path:
             kernel_config = code_path.joinpath(kconfig_path)
             err_msg = f"cannot find config path: {kernel_config}"
             assert node.shell.exists(kernel_config), err_msg
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -157,20 +157,20 @@ class SourceInstaller(BaseInstaller):
             result = node.execute(
                 "make kernelrelease 2>/dev/null",
                 cwd=code_path,
-                shell=True
+                shell=True,
             )
 
             kernel_version = result.stdout
             result.assert_exit_code(
                 0,
-                f"failed on get kernel version: {kernel_version}"
+                f"failed on get kernel version: {kernel_version}",
             )
 
             # copy current config back to system folder.
             result = node.execute(
                 f"cp .config /boot/config-{kernel_version}",
                 cwd=code_path,
-                sudo=True
+                sudo=True,
             )
             result.assert_exit_code()
 
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -144,7 +144,7 @@ class SourceInstaller(BaseInstaller):
         )
         result.assert_exit_code()
 
-        if config_path and len(config_path) > 0:
+        if config_path:
             # If it is dom0,
             # Name of the current kernel should be vmlinuz-<kernel version>
             uname = node.tools[Uname]
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -198,20 +198,23 @@ class SourceInstaller(BaseInstaller):
         uname = node.tools[Uname]
         kernel_information = uname.get_linux_information()
 
+        cp = node.tools[Cp]
         if kconfig_file:
             err_msg = f"cannot find config path: {kconfig_file}"
             assert node.shell.exists(node.get_pure_path(kconfig_file)), err_msg
-            result = node.execute(
-                f"cp {kconfig_file} .config",
+            cp.copy(
+                src=node.get_pure_path(kconfig_file),
+                dest=PurePath(".config"),
                 cwd=code_path,
             )
-            result.assert_exit_code()
         else:
-            result = node.execute(
-                f"cp /boot/config-{kernel_information.kernel_version_raw} .config",
+            cp.copy(
+                src=node.get_pure_path(
+                    f"/boot/config-{kernel_information.kernel_version_raw}"
+                ),
+                dest=PurePath(".config"),
                 cwd=code_path,
             )
-            result.assert_exit_code()
 
         config_path = code_path.joinpath(".config")
         sed = self._node.tools[Sed]
@@ -362,7 +365,7 @@ class Dom0Installer(SourceInstaller):
         current_kernel_binary = f"vmlinuz-{current_kernel}"
         new_kernel_binary = f"vmlinuz-{kernel_version}"
         source_path = code_path.joinpath("arch/x86/boot/bzImage")
-        destination_path = self._node.get_pure_path(f"/boot/efi/{new_kernel_binary}")
+        destination_path = node.get_pure_path(f"/boot/efi/{new_kernel_binary}")
         cp = node.tools[Cp]
         cp.copy(
             src=source_path,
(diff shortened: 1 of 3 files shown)


lisa/transformers/kernel_source_installer.py
```diff
@@ -114,22 +120,59 @@ def install(self) -> str:
...
+        config_type = source.get_config_type()
+        self._build_code(node=node, code_path=code_path, config_type=config_type)
 
         self._install_build(node=node, code_path=code_path)
 
-        result = node.execute(
-            "make kernelrelease 2>/dev/null", cwd=code_path, shell=True
-        )
+        kernel_version = ""
+        if self.__is_dom0(node) and config_type == "dom0":
```
LiliDeng: any other value for config_type, besides dom0?

lisa/transformers/kernel_source_installer.py
```diff
@@ -114,22 +120,59 @@ def install(self) -> str:
...
-        result.assert_exit_code(0, f"failed on get kernel version: {kernel_version}")
+            # Modify the linuxloader.conf to point new kernel binary
+            ll_conf_file = "/boot/efi/linuxloader.conf"
+            sed = self._node.tools[Sed]
+            sed.substitute(
+                regexp=current_kernel_binary,
+                replacement=new_kernel_binary,
+                file=ll_conf_file,
+                sudo=True,
+            )
```
LiliDeng: kernel_version will be empty if it is dom0?

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Hi Lili, as of now it is not. This is to identify which kernel config we should refer to. If nothing is passed then it will execute older flow. If we are building dom0 kernel, we will have to refer different file as .config

lisa/transformers/kernel_source_installer.py
smit-gardhariya: kernel_version is nothing but stdout of **make kernelrelease**. We dont have anything to do with **make kernelrelease** in case of dom0 as we dont process anything as part of it. We will have new kernel version in the log after rebooting the VM. So yes it will be empty string.

We are printing new kernel version in kernel_installer.py at line 162

lisa/transformers/kernel_source_installer.py
```diff
@@ -114,22 +120,59 @@ def install(self) -> str:
...
+        if self.__is_dom0(node) and config_type == "dom0":
+            # If it is dom0,
+            # Name of the current kernel should be vmlinuz-<kernel version>
+            uname = node.tools[Uname]
+            current_kernel = uname.get_linux_information().kernel_version_raw
+            current_kernel_binary = f"vmlinuz-{current_kernel}"
+
+            # Copy the binary to /boot/efi from : arch/x86/boot/bzImage
+            source_path = code_path.joinpath("arch/x86/boot/bzImage")
+            new_kernel_binary = "vmlinuz-%s" % str(uuid.uuid4())
```
anirudhrb: It should be `vmlinuz-<new kernel version>` and not `vmlinuz-<random guid>`.

lisa/transformers/kernel_source_installer.py
```diff
@@ -114,22 +120,59 @@ def install(self) -> str:
...
+                sudo=True,
+            )
 
-        kernel_version = result.stdout
-        result.assert_exit_code(0, f"failed on get kernel version: {kernel_version}")
+            # Modify the linuxloader.conf to point new kernel binary
+            ll_conf_file = "/boot/efi/linuxloader.conf"
+            sed = self._node.tools[Sed]
+            sed.substitute(
+                regexp=current_kernel_binary,
```
anirudhrb: Don't depend on the old value. Just find `VMLINUZ=` in the file and replace the value.

lisa/transformers/kernel_source_installer.py
```diff
@@ -171,17 +214,24 @@ def _modify_code(self, node: Node, code_path: PurePath) -> None:
...
         kernel_information = uname.get_linux_information()
 
-        result = node.execute(
-            f"cp /boot/config-{kernel_information.kernel_version_raw} .config",
-            cwd=code_path,
-        )
-        result.assert_exit_code()
+        if self.__is_dom0(node) and config_type == "dom0":
+            result = node.execute(
+                "cp arch/x86/configs/mshv_defconfig .config",
```
anirudhrb: The path to the config file should be configured from runbook. It will be an optional property and if it's not defined, we will fall back to using the config file from `/boot`.

smit-gardhariya: I will change config_type to config_path use it as you described.

lisa/transformers/kernel_source_installer.py
smit-gardhariya: There is no VMLINUZ= in the line where we need to change. It is **KERNEL_PATH=**

I will add that snipped in the string of regex and replaced one

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Sure i will make this change

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Added it

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Done

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Done

lisa/transformers/kernel_source_installer.py
smit-gardhariya: I have made changes to make code common for **make kernelrelease**

This will use kernel_version and create new binary with new version in name.

lisa/transformers/kernel_source_installer.py
```diff
@@ -114,23 +119,59 @@ def install(self) -> str:
...
         # copy current config back to system folder.
         result = node.execute(
-            f"cp .config /boot/config-{kernel_version}", cwd=code_path, sudo=True
+            f"cp .config /boot/config-{kernel_version}",
+            cwd=code_path,
+            sudo=True,
         )
         result.assert_exit_code()
 
+        if config_path and len(config_path) > 0:
```
squirrelsc: `if config_path` should be enough to check None or Empty string. But how the `config_path` relates to dom0? It's not straightforward on names and the logic.

lisa/transformers/kernel_source_installer.py
```diff
@@ -39,6 +38,12 @@ class LocalLocationSchema(BaseLocationSchema):
             required=True,
         ),
     )
+    config_path: str = field(
```
squirrelsc: The name `config_path` is not clear. It needs to be more specific.

lisa/transformers/kernel_source_installer.py
```diff
@@ -171,17 +212,27 @@ def _modify_code(self, node: Node, code_path: PurePath) -> None:
...
 
         uname = node.tools[Uname]
         kernel_information = uname.get_linux_information()
 
-        result = node.execute(
-            f"cp /boot/config-{kernel_information.kernel_version_raw} .config",
-            cwd=code_path,
-        )
-        result.assert_exit_code()
+        if kconfig_path and len(kconfig_path) > 0:
```
squirrelsc: It looks all logics are like `if` or `if/else` for dom0. So, it's better to implement a new Dom0Installer, which inherits from this installer.

lisa/transformers/kernel_source_installer.py
smit-gardhariya: I have made changes to user path instead of flag like dom0. Please refer the latest push.

lisa/transformers/kernel_source_installer.py
smit-gardhariya: This is to check if any config_path is getting passed from runbook. If it is , then we will refer it otherwise we will follow the default flow of referring the config from /boot.

config_path is the path where our config file is located. Ex. for dom0, it is arch/x86/configs/mshv_defconfig under the linux-dom0 repo.

I will remove the length condition.

smit-gardhariya: Removed the length condition

lisa/transformers/kernel_source_installer.py
smit-gardhariya: This is .config path which is needed while building the kernel so i name it like config_path

lisa/transformers/kernel_source_installer.py
smit-gardhariya: This is not specific to dom0. If we will be passing any config path from runbook , we should refer to it otherwise it will be default flow of referring config from /boot. We are making it generic so we have flexibility to pass config file dynamically from runbook. It can be any platform not just dom0

lisa/transformers/kernel_source_installer.py
squirrelsc: how about "kernel_config_file"?

smit-gardhariya: sure Chi, will make that change.

squirrelsc: Move it to BaseLocationSchema, so it doesn't need similar implementation on subclasses.

lisa/transformers/kernel_source_installer.py
```diff
@@ -114,23 +119,59 @@ def install(self) -> str:
...
 
+        if config_path:
+            # If it is dom0,
+            # Name of the current kernel should be vmlinuz-<kernel version>
+            uname = node.tools[Uname]
+            current_kernel = uname.get_linux_information().kernel_version_raw
+            current_kernel_binary = f"KERNEL_PATH=vmlinuz-{current_kernel}"
+
+            # Copy the binary to /boot/efi from : arch/x86/boot/bzImage
+            source_path = code_path.joinpath("arch/x86/boot/bzImage")
```
squirrelsc: Does this path apply to all kernels, like arm64, and what's the bzImage standard?

lisa/transformers/kernel_source_installer.py
smit-gardhariya: sure, Chi

lisa/transformers/kernel_source_installer.py
smit-gardhariya: bzImage is the image generated while we build the kernel. I think it would remain same but in case of normal flow it would be copied to /boot and make changes in grub and for dom0 we are copying kernel binary to /boot/efi and changing linuxloader.conf

lisa/transformers/kernel_source_installer.py
smit-gardhariya: done

lisa/transformers/kernel_source_installer.py
```diff
@@ -27,7 +26,12 @@ class BaseModifierSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
 @dataclass_json()
 @dataclass
 class BaseLocationSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
-    ...
+    kernel_config_file: str = field(
```
anirudhrb: This field needs to be documented properly. Please add a comment that addresses the following questions:
1. Is this a path on the local node (where lisa is running) or remote node?
2. Is it an absolute path or a path relative to the root of the kernel source tree?

lisa/transformers/kernel_source_installer.py
```diff
@@ -336,6 +408,10 @@ def get_source_code(self) -> PurePath:
 
         return code_path
 
+    def get_kconfig_file(self) -> str:
```
anirudhrb: This function can be implemented in base and inherited

lisa/transformers/kernel_source_installer.py
anirudhrb: Also, this field should probably be in SourceInstallerSchema since it's purpose is not for specifying the location of the kernel source code.

anirudhrb: @squirrelsc what do you think of this?

lisa/transformers/kernel_source_installer.py
squirrelsc: Can the arch be gotten from the system? We assume the current machine build the kernel with same arch.

lisa/transformers/kernel_source_installer.py
squirrelsc: The BaseLocationSchema is for location of source code. It has local (existing code), repo (download source code from remote repo). Both of them need to support kernel configuration. It's not about where the source code is from.

smit-gardhariya: Comment is added, Path is absolute now

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Removed this function in latest push as this param is pushed at installer level schema

lisa/transformers/kernel_source_installer.py
smit-gardhariya: I moved kernel config file to source installer schema. It can be absolute path of node.

lisa/transformers/kernel_source_installer.py
smit-gardhariya: As of now it is for x86/x86_64. There is no cross compilation planned as of now. We can get hw platform detail from uname.

Also we would need to install cross compiler dependencies for it.

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Pushed it to source installer schema. Implemented the Child class of it for dom0 specific actions.

lisa/transformers/kernel_source_installer.py
```diff
@@ -77,6 +76,20 @@ class SourceInstallerSchema(BaseInstallerSchema):
     # Steps to modify code by patches and others.
     modifier: List[BaseModifierSchema] = field(default_factory=list)
 
+    # This is absolute path where kernel config file is located
```
anirudhrb: If the config is in-tree, then users need to know where lisa is cloning the tree to be able to specify the absolute path. Users should not need to figure this out.

lisa/transformers/kernel_source_installer.py
```diff
@@ -171,17 +192,26 @@ def _modify_code(self, node: Node, code_path: PurePath) -> None:
...
-        result = node.execute(
-            f"cp /boot/config-{kernel_information.kernel_version_raw} .config",
-            cwd=code_path,
-        )
-        result.assert_exit_code()
+        if kconfig_file:
+            err_msg = f"cannot find config path: {kconfig_file}"
+            assert node.shell.exists(node.get_pure_path(kconfig_file)), err_msg
+            result = node.execute(
+                f"cp {kconfig_file} .config",
```
anirudhrb: Use Cp tool.

lisa/transformers/kernel_source_installer.py
```diff
@@ -269,13 +299,112 @@ def _install_build_tools(self, node: Node) -> None:
...
+        # If it is dom0,
+        # Name of the current kernel should be vmlinuz-<kernel version>
+        uname = node.tools[Uname]
+        current_kernel = uname.get_linux_information().kernel_version_raw
+
+        # Copy the kernel to /boot/efi from : arch/x86/boot/bzImage
+        current_kernel_binary = f"vmlinuz-{current_kernel}"
+        new_kernel_binary = f"vmlinuz-{kernel_version}"
+        source_path = code_path.joinpath("arch/x86/boot/bzImage")
+        destination_path = self._node.get_pure_path(f"/boot/efi/{new_kernel_binary}")
```
anirudhrb:
```suggestion
        destination_path = node.get_pure_path(f"/boot/efi/{new_kernel_binary}")
```

lisa/transformers/kernel_source_installer.py
jinankjain: Currently there is no support for arm64 for LinuxDom0 setup, so there is no point of extending LISA to cross compile for ARM64.

lisa/transformers/kernel_source_installer.py
smit-gardhariya: Previous implementation was using execute() way to run command so followed the same. This would require change in cp.copy() implementation to add **cwd** parameter, I will do it and use cp.copy().

smit-gardhariya: done

lisa/transformers/kernel_source_installer.py
smit-gardhariya: done

lisa/transformers/kernel_source_installer.py
smit-gardhariya: @anirudhrb Could you please give explain

anirudhrb: Let's say the config is inside the kernel source tree. Then, in order to specify an absolute path to it, the user of this transformer would need to know the location where lisa is cloning the repo.

lisa/transformers/kernel_source_installer.py
```diff
@@ -77,6 +76,20 @@ class SourceInstallerSchema(BaseInstallerSchema):
...
+        default="",
+        metadata=field_metadata(
+            required=False,
+        ),
+    )
+
+
+@dataclass_json()
+@dataclass
+class Dom0InstallerSchema(SourceInstallerSchema):
```
squirrelsc: Because there is no new thing from schema. it doesn't need to declare a schema for dom0.

smit-gardhariya: I declared it to maintain the schema and installer combination. As of now we dont have any parameter in dom0 schema but we might have in future.

lisa/transformers/kernel_source_installer.py
smit-gardhariya: I changed the source path for kernel binary. make install will execute /sbin/installkernel script and generate new kernel binary at /boot from there we can copy to /boo/efi. 

We dont need to follow path arch/x86/boot/bzImage.

I tried with arm64 as well and our source would remain same if we refer /boot. Please refer the latest commit

lisa/transformers/kernel_source_installer.py
smit-gardhariya: @anirudhrb  As we discussed, kernel config file will be relative path to the kernel source code path

smit-gardhariya: updated the comment

lisa/transformers/kernel_source_installer.py
smit-gardhariya: @anirudhrb As we discussed, kernel config file will be relative path to the kernel source code path

lisa/transformers/kernel_source_installer.py
smit-gardhariya: This is to maintain installer and its corresponding schema

squirrelsc: If extra parameter needed, it can be added in future. It doesn't need a new schema for each new type. The Installer can reference the source installer schema.

smit-gardhariya: ok chi i will remove dom0 schema and refer to source schema in dom0 installer

smit-gardhariya: Done

lisa/transformers/kernel_source_installer.py
```diff
@@ -51,6 +50,12 @@ class RepoLocationSchema(LocalLocationSchema):
     # fail the run if code exists
     fail_on_code_exists: bool = False
     cleanup_code: bool = False
+    auth_token: Optional[str] = field(
```
squirrelsc, line 53: It's ok to add it. Just FYI, internal repo can be used directly, if the running account is granted. It doesn't need auth token.

squirrelsc (approved):

squirrelsc:
Please fix the conflicts to get it merged.

vyadavmsft:
> Please fix the conflicts to get it merged.

I'm seeing no conflict.

squirrelsc:
> > Please fix the conflicts to get it merged.
> 
> I'm seeing no conflict.

It needs to rebase to latest main.
image

PR #2434, opened 2022-11-29
Provide an option to capture libvirt debug logs

anirudhrb:
Introduce a new option capture_libvirt_debug_logs in LibvirtPlatformSchema that tells LISA to capture libvirt debug logs during environment teardown.

Debug logs are enabled by using the log_outputs option in /etc/libvirt/libvirtd.conf. Here, we configure libvirt to write the debug logs to /var/log/libvirt/libvirtd.log.

We have been seeing multiple libvirt related errors in the pipeline and these logs can help in debugging those.

This is an opt-in feature since all users of libvirt platform might not be interested in debug level logs.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -12,7 +12,7 @@ import sys
 import tempfile
 import time
 import xml.etree.ElementTree as ET  # noqa: N817
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from threading import Lock, Timer
 from typing import Any, Dict, List, Optional, Tuple, Type, cast
 
@@ -26,9 +26,22 @@ from lisa.feature import Feature
 from lisa.node import Node, RemoteNode, local_node_connect
 from lisa.operating_system import CBLMariner
 from lisa.platform_ import Platform
-from lisa.tools import Chmod, Dmesg, Iptables, Journalctl, Ls, QemuImg, Uname
+from lisa.tools import (
+    Chmod,
+    Chown,
+    Cp,
+    Dmesg,
+    Iptables,
+    Journalctl,
+    Ls,
+    QemuImg,
+    Sed,
+    Service,
+    Uname,
+    Whoami,
+)
 from lisa.util import LisaException, constants, get_public_key_data
-from lisa.util.logger import Logger, filter_ansi_escape
+from lisa.util.logger import Logger, filter_ansi_escape, get_logger
 
 from . import libvirt_events_thread
 from .console_logger import QemuConsoleLogger
@@ -63,6 +76,13 @@ class _HostCapabilities:
 
 
 class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
+    LIBVIRTD_CONF_PATH = PurePosixPath("/etc/libvirt/libvirtd.conf")
+    LIBVIRT_DEBUG_LOG_PATH = PurePosixPath("/var/log/libvirt/libvirtd.log")
+    # A marker that identifies lines added by lisa in a config file. This can be
+    # appended as a comment and then used to identify the line to delete during
+    # cleanup.
+    CONFIG_FILE_MARKER = "lisa-libvirt-platform"
+
     _supported_features: List[Type[Feature]] = [
         SerialConsole,
         StartStop,
@@ -114,15 +134,8 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
             self.__platform_runbook_type(), type_name=type(self).type_name()
         )
 
-        self.__init_libvirt_conn_string()
-        self.libvirt_conn = libvirt.open(self.libvirt_conn_str)
-
-    def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
-        # Ensure environment log directory is created before connecting to any nodes.
-        _ = environment.log_path
-
         if len(self.platform_runbook.hosts) > 1:
-            log.warning(
+            self._log.warning(
                 "Multiple hosts are currently not supported. "
                 "Only the first host will be used."
             )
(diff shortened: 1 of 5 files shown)


squirrelsc:
@cwize1 please help on reviewing the change.

cwize1:
What was the problem with the current journalctl solution?

anirudhrb:
> What was the problem with the current journalctl solution?

No problem. That is still the default.

For debug logs libvirt recommends using a file as destination since journald might throttle the amount of logs.
https://libvirt.org/kbase/debuglogs.html

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -155,6 +167,11 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
                 parent_logger=log,
             )
 
+        if self.platform_runbook.capture_libvirt_debug_logs:
+            self._enable_libvirt_debug_log()
+
+        self.libvirt_conn = libvirt.open(self.libvirt_conn_str)
```
cwize1, line 173: Why was this line moved?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1187,3 +1199,29 @@ def _get_environment_information(self, environment: Environment) -> Dict[str, st
                     node.log.exception(f"error on get {key}.", exc_info=identifier)
 
         return information
+
+    def _enable_libvirt_debug_log(self) -> None:
```
cwize1, line 1216: Should this be disabled after all the tests have run?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1187,3 +1199,29 @@ def _get_environment_information(self, environment: Environment) -> Dict[str, st
...
+        )
+
+        self.host_node.tools[Service].restart_service("libvirtd")
+
+    def _capture_libvirt_logs(self) -> None:
+        libvirt_log_path = self.host_node.local_log_path / "libvirtd.log"
+
+        if self.platform_runbook.capture_libvirt_debug_logs:
+            user = self.host_node.tools[Whoami].get_username()
+            self.host_node.tools[Chown].change_owner(self.LIBVIRT_DEBUG_LOG_PATH, user)
```
cwize1: This seems like it might cause problems, especially if run on someone’s dev box. I recommend copying the file to `working_dir` first, changing the permissions on that file and then copying to the test runner machine.

lisa/sut_orchestrator/libvirt/platform.py
cwize1: If I recall correctly, I needed to place this line in `_initialize` to ensure that it is only opened once for all test suites/cases. Otherwise, the console logger would be unhappy when running multiple tests that include a reboot test.

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: I moved it because _enable_libvirt_debug_log() restarts libvirtd. So, the connection is closed. Looks like I need to move this to stuff to `_initialize`.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: What kind of problems do you expect this might cause?

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: Please check the latest iteration. I moved `host_node` also to `initialize` since it doesn't need to be done "per environment".

lisa/sut_orchestrator/libvirt/platform.py
cwize1: You shouldn’t change permissions in system directories to be that of a user.

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: Done.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: Looks to me like there isn't a good place to do this. `_delete_environment` being called doesn't necessarily mean that all tests have run right? If an environment goes down during test execution lisa will create a new one.
Do we need more lifecycle callbacks?

squirrelsc: Good point, please add a "cleanup" method on platform level, and call it in lisa_runner.close.

anirudhrb: Done. Please take a look.

anirudhrb:
@cwize1 @squirrelsc please do another round of review.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -168,12 +189,10 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
...
-        libvirt_log = self.host_node.tools[Journalctl].logs_for_unit(
-            "libvirtd", sudo=self.host_node.is_remote
-        )
-        libvirt_log_path = self.host_node.local_log_path / "libvirtd.log"
-        with open(str(libvirt_log_path), "w") as f:
-            f.write(libvirt_log)
+        self._capture_libvirt_logs()
+
+    def _cleanup(self) -> None:
+        self._disable_libvirt_debug_log()
```
cwize1: Only run this if `capture_libvirt_debug_logs` is true?

cwize1 (dismissed):

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: Done.

squirrelsc (approved):

PR #2437, opened 2022-12-02
This file causes internal issues and not needed on branch master-se-1

santoshx:


ShiningCreek:
Do you consider removing the case definition from the xml file?

PR #2438, opened 2022-12-02
[fix issue] set extended runbook after resizing cases

LiliDeng:
This will fix the image, location and vmsize are empty after running resize test cases. And vmsize will be filled as vmsize after resizing vm.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -1251,11 +1251,11 @@ class Resize(AzureFeatureMixin, features.Resize):
 
     def resize(
         self, resize_action: ResizeAction = ResizeAction.IncreaseCoreCount
-    ) -> schema.NodeSpace:
+    ) -> Tuple[schema.NodeSpace, str, str]:
         platform: AzurePlatform = self._platform  # type: ignore
         compute_client = get_compute_client(platform)
         node_context = get_node_context(self._node)
-        new_vm_size_info = self._select_vm_size(resize_action)
+        origin_vm_size, new_vm_size_info = self._select_vm_size(resize_action)
 
         # Creating parameter for VM Operations API call
         hardware_profile = HardwareProfile(vm_size=new_vm_size_info.vm_size)
@@ -1274,11 +1274,11 @@ class Resize(AzureFeatureMixin, features.Resize):
         self._node.close()
         new_capability = copy.deepcopy(new_vm_size_info.capability)
         self._node.capability = cast(schema.Capability, new_capability)
-        return new_capability
+        return new_capability, origin_vm_size, new_vm_size_info.vm_size
 
     def _select_vm_size(
         self, resize_action: ResizeAction = ResizeAction.IncreaseCoreCount
-    ) -> "AzureCapability":
+    ) -> Tuple[str, "AzureCapability"]:
         platform: AzurePlatform = self._platform  # type: ignore
         compute_client = get_compute_client(platform)
         node_context = get_node_context(self._node)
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1392,6 +1392,9 @@ def _select_vm_size(
         # Choose random size from the list to resize to
         index = randint(0, len(avail_eligible_intersect) - 1)
         resize_vm_size = avail_eligible_intersect[index]
+        node_runbook.vm_size = resize_vm_size.vm_size
```
squirrelsc: Can we set the property of the related test results, make sure both old/new vm sizes are logged? You can check if the environment can save this information, and bring to test results. It doesn't block this PR, just add a task to track it.

squirrelsc: How about to set the environment is dirty after this test case? So the resized env is not confusing following test cases.

LiliDeng: How about resize back to the original vm size?

squirrelsc: I think it takes longer time than deploying a new one. It may cause other problems. So mark it dirty is clearer.

LiliDeng:
this is information column
```{"resource_group_name": "lisa-xxxxxxxxxxx", "origin_vm_size": "Standard_DS3_v2", "final_vm_size": "Standard_E8_v3"}```

File: microsoft/testsuites/core/vm_resize.py [hunk 2]
```diff
@@ -143,6 +158,10 @@ def _verify_vm_resize(
         assert expected_vm_capability, "fail to find proper vm size"
         if not hot_resize:
             start_stop.start()
+
+        runbook = node.capability.get_extended_runbook(AzureNodeSchema)
+        test_result.information["final_vm_size"] = runbook.vm_size
```
squirrelsc: Make the resize method return resized new vm size, instead of read it form runbook. So the test case doesn't need to depend on Azure specific runbook.

LiliDeng: any other way to get origin vm size besides get from runbook?
```
        runbook = node.capability.get_extended_runbook(AzureNodeSchema)
        test_result.information["origin_vm_size"] = runbook.vm_size
```

squirrelsc: How about return a tuple with two vm sizes from `resize`, (origin, new)?

LiliDeng: updated

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1367,8 +1367,12 @@ def _select_vm_size(
         # Choose random size from the list to resize to
         index = randint(0, len(avail_eligible_intersect) - 1)
         resize_vm_size = avail_eligible_intersect[index]
+        origin_vm_size = node_runbook.vm_size
+        node_runbook.vm_size = resize_vm_size.vm_size
+        node_runbook.location = resize_vm_size.location
+        resize_vm_size.capability.set_extended_runbook(node_runbook)
         self._log.info(f"New vm size: {resize_vm_size.vm_size}")
-        return resize_vm_size
+        return resize_vm_size, origin_vm_size
```
squirrelsc: origin, then new size.

LiliDeng: this is used in https://github.com/microsoft/lisa/pull/2438/files#diff-05e1ad7a12a81a2eff6db4a7842324e2c04cbde8cf304116e8b072cc16bdd13bR1258

and here https://github.com/microsoft/lisa/pull/2438/files#diff-05e1ad7a12a81a2eff6db4a7842324e2c04cbde8cf304116e8b072cc16bdd13bR1277 used in test case

squirrelsc: I mean, swap their positions. The order old/new is more natural than new/old.

```suggestion
        return origin_vm_size, resize_vm_size
```

squirrelsc (dismissed):

PR #2439, opened 2022-12-02
Retry ssh process spawn if it times out

anirudhrb:
Many times, LISA fails with the following message:

"The remote node is timeout on execute [....<command>....]. It may be caused by paramiko/spur not support the shell of node"

This is an intermittent issue with no known stable repro. In the beginning of the run LISA can connect to the node and execute commands. So, it is clearly not due to paramiko/spur not supporting the shell of the node.

As a theoretical fix, add a mechanism to retry ssh process spawn if it fails due to timeout. Retry will make LISA resilient to any transient issues in the underlying layer.

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -192,6 +192,7 @@ class SshShell(InitializableMixin):
         self._inner_shell: Optional[spur.SshShell] = None
         self._jump_boxes: List[Any] = []
         self._jump_box_sock: Any = None
+        self._log = get_logger("SshShell")
 
         paramiko_logger = logging.getLogger("paramiko")
         paramiko_logger.setLevel(logging.WARN)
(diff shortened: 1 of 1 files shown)


lisa/util/shell.py
```diff
@@ -291,24 +292,43 @@ def spawn(
...
-                encoding=encoding,
-                use_pty=use_pty,
-                allow_error=allow_error,
-            )
-        except FunctionTimedOut:
-            raise LisaException(
-                f"The remote node is timeout on execute {command}. "
-                f"It may be caused by paramiko/spur not support the shell of node."
-            )
+        MAX_TRIES = 3
```
squirrelsc, line 295: I'm not preferring to try from this layer. When the command is time out, it's hard to know it's executed or not. The blinding retrying on the lower layer may make the command execute multiple times. So the retry should be write on higher layer, which the result is fully understood. Either it's ok to run multiple times, or the reason of timeout is clear.

squirrelsc:
> Many times, LISA fails with the following message:
> 
> "The remote node is timeout on execute [........]. It may be caused by paramiko/spur not support the shell of node"
> 
> This is an intermittent issue with no known stable repro. In the beginning of the run LISA can connect to the node and execute commands. So, it is clearly not due to paramiko/spur not supporting the shell of the node.
> 
> As a theoretical fix, add a mechanism to retry ssh process spawn if it fails due to timeout. Retry will make LISA resilient to any transient issues in the underlying layer.

We need to understand more on the timeout. If it's caused by connecting too earlier, we can retry on connecting phase. Some commands make the connection lost, we can reconnect after that. Anyway, blinding retry when executing commands, it makes other unexpected behavior.

anirudhrb:
Closing since issue is not too frequent. If needed, we will retry on higher level on a case by case basis.

PR #2444, opened 2022-12-02
Run LTP tests on WSL

dsrivastavv:


microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -57,6 +57,8 @@ def ltp_lite(
         # parse variables
         tests = variables.get("ltp_test", "")
         skip_tests = variables.get("ltp_skip_test", "")
+        use_block_drive = variables.get("use_block_drive", True)
+        use_temp_dir = variables.get("use_temp_dir", False)
```
squirrelsc: What's the reason need this? How about always use the /tmp folder when it's os disk?

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -57,6 +57,8 @@ def ltp_lite(
         # parse variables
         tests = variables.get("ltp_test", "")
         skip_tests = variables.get("ltp_skip_test", "")
+        use_block_drive = variables.get("use_block_drive", True)
```
squirrelsc: the name should be `use_os_drive` like in the code, and default value is "False".

microsoft/testsuites/ltp/ltp.py
```diff
@@ -346,7 +352,7 @@ def _install(self) -> bool:
         self.node.execute("autoreconf -f", cwd=ltp_path, sudo=True)
         make.make("autotools", cwd=ltp_path, sudo=True)
         self.node.execute("./configure --prefix=/opt/ltp", cwd=ltp_path, sudo=True)
-        make.make("all", cwd=ltp_path, sudo=True)
+        make.make("all", cwd=ltp_path, sudo=True, timeout=self.MAKE_TIMEOUT)
```
squirrelsc: When is it needed? what happened if it's timed out?

lisa/node.py
```diff
@@ -283,7 +283,9 @@ def capture_system_information(self, name: str = "") -> None:
         self.log.debug(f"capturing system information to {saved_path}.")
         self.os.capture_system_information(saved_path)
 
-    def find_partition_with_freespace(self, size_in_gb: int) -> str:
+    def find_partition_with_freespace(
+        self, size_in_gb: int, use_os_disk: bool = False
```
squirrelsc: `use_os_disk_only`

microsoft/testsuites/ltp/ltpsuite.py
dsrivastavv: This is not related to the use of os disk. It's a parameter in the LTP test. We don't need to enable it in our tests but the WSL testing is being done with this parameter

lisa/node.py
```diff
@@ -296,6 +298,28 @@ def find_partition_with_freespace(self, size_in_gb: int) -> str:
         lsblk = self.tools[Lsblk]
         disks = lsblk.get_disks()
 
+        if use_os_disk:
```
squirrelsc: The logic is most similar to next one. Try to reuse it.

microsoft/testsuites/ltp/ltpsuite.py
dsrivastavv: Our tests run with the default value of `True`. Few tests in LTP lite require this to be set and this increases the number of tests which are covered by LTP lite

microsoft/testsuites/ltp/ltp.py
dsrivastavv: I was seeing timeouts on WSL, so decided to increase the default value

lisa/node.py
dsrivastavv: We do not need to necessarily constraint to the OS disk. The logic here when `use_os_disk` is:
1. Try to check if os disk can be used and has enough available space
2. Otherwise check the data disks

microsoft/testsuites/ltp/ltpsuite.py
dsrivastavv: Also, OS disk is different from Block drive. Using block drive means LTP can use any block drive

lisa/node.py
squirrelsc: If so, I think it doesn't need to be checked. I don't see any case need to skip os disk. The data disk is attached, because in some cases the os disk is not enough.

dsrivastavv: I think your explanation makes sense. All we need to do in our test cases is to ensure that an additional data disk is available. I will make the change

microsoft/testsuites/ltp/ltpsuite.py
squirrelsc: Do you know why WSL need this parameter? Can we always put the results in tmp or home folder?

squirrelsc: The parameters will make test cases complex to configure and run. Even we provide the default values, the unneeded parameters will be confusing.

microsoft/testsuites/ltp/ltpsuite.py
squirrelsc: The os drive is a block device too. The variable name is confusing here.

microsoft/testsuites/ltp/ltp.py
squirrelsc: So, if it's timed out, what's error message here before? And after added this, what's expected?

microsoft/testsuites/ltp/ltpsuite.py
dsrivastavv: Fixed! Remove the temporary directory variable

microsoft/testsuites/ltp/ltpsuite.py
dsrivastavv: Fixed!

microsoft/testsuites/ltp/ltp.py
dsrivastavv: There is no error message. Just timeout because it is sometimes slow to make on WSL. So I have increased timeout

lisa/node.py
dsrivastavv: I went through the code again and we need this flag. For example, we need to pass a block device in LTP that it can modify and it can't be an OS disk

squirrelsc: 1. All disks are "block devices". It's confusing to use this variable to differentiate os disk.
2. Because the WSL is running in a managed environment, can it be changed to pass in a device name directly? If it's not possible, let's try a clearer name.

dsrivastavv: @squirrelsc Is there any action needed on this? I have updated variable name.

squirrelsc: The variable is still block disk. See my latest replies, considering passing in the disk name. It may be more flexible.
`        use_block_disk = variables.get("use_block_disk", True)`

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -58,6 +58,10 @@ def ltp_lite(
         tests = variables.get("ltp_test", "")
         skip_tests = variables.get("ltp_skip_test", "")
 
+        # block device is required for few ltp tests
+        # If not provided, we will find a disk with enough space
+        block_device = variables.get("block_device", None)
```
squirrelsc: `ltp_drive_name`

dsrivastavv: `lip_drive_name` seems related to installation drive. How about adding a test prefix like `test_block_device`?

squirrelsc: 1. It needs start with "ltp_" to get others know it applies to ltp test cases.
2. Disk is not the only block devices in Linux. I see you assigned it to an internal variable name "drive_name", so I set it to second part to make it consistent.
3. I just found the `device_name` is set to the file system type "-z" actually. Does it really work? It looks should be set to tmp folder "-d".

How about name it like "ltp_tmp_dir"?

dsrivastavv: There are two different things here:

```
if drive_name:
    parameters += f"-z {drive_name} "

parameters += f"-d {temp_dir} "
```

1. `drive_name`: This is set with `-z` option. This is used by some tests, probably to measure performance. 
2. `temp_dir`: This is set with `-d` option. This is used for storing temp files during ltp run. 

I will update the code to make it `ltp_block_device` and set `block_device` to `-z` option to be consistent.

dsrivastavv: From LTP manpage:

```
-z BIG_DEVICE   Some tests require a big unmounted block device
                    to run correctly.
 -d TMPDIR       Directory where temporary files will be created.
```

squirrelsc (dismissed):

dsrivastavv:
@squirrelsc We have verified this on the WSL. This PR should be good to go.

squirrelsc (approved):

PR #2445, opened 2022-12-05
Various fixes for Cloud Hypervisor tests

anirudhrb:
This PR contains fixes related to collecting dmesg logs and improved sub test reporting in Cloud Hypervisor
tests. Commits are well-defined. Here's a summary:

#### ch_tests_tool: save dmesg logs

  Save dmesg logs after text execution is finished. These logs are
  useful for investigating failures.

#### ch_tests: fix argument to Modprobe.remove()

  Modprobe tool's remove() function accepts module names as a
  `List[str]` but ch_test.py passes a `str`.
  
  Fix by passing a list as argument.

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -13,7 +13,7 @@ from lisa.executable import Tool
 from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
 from lisa.operating_system import CBLMariner
 from lisa.testsuite import TestResult
-from lisa.tools import Docker, Echo, Git, Whoami
+from lisa.tools import Dmesg, Docker, Echo, Git, Whoami
 
 
 @dataclass
@@ -48,6 +48,7 @@ class CloudHypervisorTests(Tool):
         environment: Environment,
         test_type: str,
         hypervisor: str,
+        log_path: Path,
         skip: Optional[List[str]] = None,
     ) -> None:
 
@@ -66,10 +67,10 @@ class CloudHypervisorTests(Tool):
             shell=True,
         )
 
+        # Report subtest results and collect logs before doing any
+        # assertions.
         results = self._extract_test_results(result.stdout)
         failures = [r.name for r in results if r.status == TestStatus.FAILED]
-        if not failures:
-            result.assert_exit_code()
 
         for r in results:
             self._send_subtest_msg(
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -275,3 +282,9 @@ def _process_perf_metric_test_result(self, output: str) -> str:
         if result:
             return result.group(0)
         return ""
+
+    def _save_dmesg_logs(self, log_path: Path) -> None:
+        dmesg_str = self.node.tools[Dmesg].get_output()
```
LiliDeng: use force_run, otherwise it always uses cache log

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -59,7 +63,7 @@ def run_tests(
         result = self.run(
             f"tests --hypervisor {hypervisor} --{test_type} -- -- {skip_args}"
             " -Z unstable-options --format json",
-            timeout=self.TIME_OUT,
+            timeout=self.CMD_TIME_OUT,
```
pupacha: I just realized that if its timed out, the stdout is set empty. https://github.com/microsoft/lisa/blob/main/lisa/util/process.py#L238
```
if is_timeout:
  process_result = spur.results.result(
      return_code=1, allow_error=True, output="", stderr_output=""
  )
````

File: microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py [hunk 3]
```diff
@@ -79,6 +81,9 @@ def run_tests(
                 r.status,
             )
 
+        self._save_dmesg_logs(log_path)
+
+        result.assert_exit_code()
         assert_that(failures, f"Unexpected failures: {failures}").is_empty()
```
pupacha, line 85: This would be unreachable statement as there is assert right above it and when there are failed subtests the exit code is not 0 from CH cmd.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
anirudhrb: Hmmm... I thought this was working. Looks like more work is needed. I'll drop this commit and take it up in a separate PR.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
anirudhrb: Fixed

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
anirudhrb: Done.

pupacha (approved):

LiliDeng (approved):

PR #2448, opened 2022-12-06
Add test to check IP of PF after StopStart

adityagesh:
verify_pf_ip creates traffic in VM using iperf3 and verifies IP is assigned to eth interface before and after StartStop operation


adityagesh:
Should it be priority 1 or 2 ?

LiliDeng:
@adityagesh let's discuss offline

PR #2449, opened 2022-12-06
doc: fix explanation on CustomScript

squirrelsc:


docs/write_test/extension.rst
```diff
@@ -106,10 +106,12 @@ Learn more about how to use the tool from `helloworld.py
 CustomScript
 ------------
 
-The ``CustomScript`` is like a lightweight tool. However, **please avoid
-using it** unless there are serious performance issues or other reasons,
-because it will return the original results to the test case. You can
-also package custom scripts as tools.
+The ``CustomScript`` is like a lightweight tool, which is composited by one or
```
dsrivastavv, line 109: [nit] **script** of LISA cannot be used to be compatible with different distros.

squirrelsc: It's talking about the tools of LISA cannot be used in the custom script. Is it clear to know, or any suggestions to be more specific?

dsrivastavv: I think we can separate it into two sentences. How about: 
For example, the script runs on nodes and the output may not be dumped into LISA log. Further, we cannot leverage distro-agnostic LISA tools in customscripts.

dsrivastavv (approved):

PR #2452, opened 2022-12-07
Fix /dev/mshv permission issue on HyperV root partition

pupacha:
`mshv` is the owning group of `/dev/mshv` in linux root partition. CH tests and mshv tests requires the user to have access to `/dev/msh`. Hence adding the user to `mshv` group.


pupacha:
@anirudhrb Please review.

anirudhrb (approved):

LiliDeng (approved):

PR #2455, opened 2022-12-08
[doc] add f-strings into Coding guidelines

LiliDeng:


docs/write_test/guidelines.rst
```diff
@@ -233,6 +234,27 @@ Learn more from `examples
...
+
+``print("Hello, %s. You are %s." % (name, age))``
+
+``print(("Hello, {0}. You are {1}.").format(name, age))``
+
+f-strings is brought since Python 3.6, f-strings are string literals that have an f at
+the beginning and curly braces containing expressions that will be replaced with their
+values. f-strings are a great new way to format strings. Not only are they
+more readable, more concise, and less prone to error than other ways of formatting, but
+they are also faster.
```
avylove: Suggestion for wording

```
f-strings were introduced in Python 3.6. f-strings are string literals which begin with an ``'f'``. Curly braces
containing expressions are used for substitution substitution. f-strings are easier to read, more concise,
less prone to error, and faster than other string interpolation methods.

It is recommended to use f-strings where possible when writing code. They are recommended in most
common cases except logging. When using Python's logging package, string interpolation is delayed until
it has been determined the message is of a level currently being logged. Using f-strings for logging overwrites
this delayed interpolation and should be avoided. %-formatting should be used in this case.
```

docs/write_test/guidelines.rst
```diff
@@ -233,6 +234,27 @@ Learn more from `examples
 <https://github.com/microsoft/lisa/tree/main/examples/testsuites>`__ and
 `assertpy document <https://github.com/assertpy/assertpy#readme>`__.
 
+Tips for writing code
+--------------------------
+
+Before Python 3.6, you had two main ways of embedding Python expressions inside string
+literals for formatting: %-formatting and str.format(). These ways have some
+limitations, so they are not recommended to use.
```
avylove: Suggestion for wording

```
Prior to Python 3.6, there were two methods for `string interpolation`_: %-formatting (sometimes referred to as
C-strings) and ``str.format()`` (along with the related ``format()`` and ``object.__format__()``. These methods
have some limitations, so their use should be limited when support for Python < 3.6 is not required.

.. _string interpolation:  https://en.wikipedia.org/wiki/String_interpolation
```

docs/write_test/guidelines.rst
squirrelsc: I had a quick look on f-string with logging. My understanding is, the f string will be parsed, even the log line won't output. So, there are perf impact. It won't be a concern in LISA, because LISA always generate debug level log in log files.
In another hand, I would just mention f-strings as a best practice in this doc, instead of putting too much content here. It's not bigger than other topics in this document. I hope it can save reader's burden on this document.

docs/write_test/guidelines.rst
squirrelsc: The whole thing is not worth to mention, because we're using >3.8. The key reasons to use f-string, because its readability is better.

docs/write_test/guidelines.rst
```diff
@@ -8,6 +8,7 @@ Coding guidelines
 -  `Logging <#logging>`__
 -  `Error message <#error-message>`__
 -  `Assertion <#assertion>`__
+-  `Tips for writing code <#tips-for-writing-code>`__
```
squirrelsc: Move it to lower priority, just before the "tips for non-native..."

docs/write_test/guidelines.rst
avylove: It is always a concern. F-strings are not used with the logging module, linters will flag it and it is bad practice. If the goal here is to educate people who are not proficient in Python, then it is relevant. If the assumption is the people already know Python well, then the entire section is unnecessary.

squirrelsc: ok, we can mention about this limitation, but not confusing with this topic. LISA is always recommending f-string.

avylove: I don't think you want to word is as always. Logging is a well-known case, but there are others where f-string either don't work or impose a performance penalty. Even if it's the same result, it's best to stick with accepted practices unless there is a good reason not to.

squirrelsc: I hope this doc can provide readers a clear guidance. Is there any (or potential) case with perf impact in LISA? If so, we can specify it here.

avylove: Just logging. The other cases where you don't use f-strings are usually around templating and shouldn't apply.

squirrelsc: My understanding is, the f string will be parsed, even the log line won't output. This is the performance concern on f-string. For example, in the `log.debug(f"{call(x)}")`, the call method will be executed, even debug level is not needed. But in LISA, Th debug level always output to log files. So there is no concern in LISA. To avoid confusing others' Python knowledge, we can mention that, but we should ask them to always use f-string in Python to improve readability.

avylove: Then that is a bug. Debug logs are not always outputted. It sounds like LISA is not using standard log levels or using log levels the way handlers are intended. Additionally, logging should always be configurable. So we'll add that to the list of non-standard things that need to be fixed. We are not going to teach people bad Python practices and we are not going to propagate them in our code.

squirrelsc: I haven't seen it's a widely practice, which considers debugging level shouldn't be dumped in log file by default. One of LISA's requirements is to tell a story from log. The Info level is used to tell the story. The debug level is to provide details. LISA is a test framework, and the test failure is normal situation. Python has only two normal levels, so there is no verbose level like some other languages.  If some line outputs thousands of lines in log file, it's considered a bug. Otherwise, it's by design.

avylove: Python has 5 logging levels by default and you can create any number of custom levels. These are derived from the standard syslog levels with the absence of alert and notice. Debug is intended for information about the program itself and intended for tracing issues, not regular output ( i.e. it shouldn't usually be needed). I believe what you're talking about is test output, which would typically be handled by a test-specific logger, independent from the main program logger. If that is not how lisa works, then that sounds like a bug.

docs/write_test/guidelines.rst
LiliDeng: updated

squirrelsc (approved):

PR #2458, opened 2022-12-08
Couple of fixes for MshvHostTestSuite

anirudhrb:
This PR has the following commits:

**mshv_root_tests: save dmesg at the correct place**

dmesg logs are not saved for some error paths which throw before
reaching the code that saves the logs.

Move this code to the correct place in verify_mshv_stress_vm_create.

**mshv_root_tests: use Azure temp disk for storing large files**

The testcase makes a copy of the disk image for each VM it
spawns. It is easy to run of space in OS disk especially
when the default value of 30GB is used.

Use the Azure temporary disk that is automatically mounted
at /mnt to store copies of the disk image.

### DIFF
--- a/microsoft/testsuites/mshv/mshv_root_tests.py
+++ b/microsoft/testsuites/mshv/mshv_root_tests.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 import time
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import Any, Dict
 
 from assertpy import assert_that
@@ -142,6 +142,7 @@ class MshvHostTestSuite(TestSuite):
         )
         hypervisor_fw_path = str(node.get_working_path() / self.HYPERVISOR_FW_NAME)
         disk_img_path = node.get_working_path() / self.DISK_IMG_NAME
+        disk_img_copy_path = self._get_disk_img_copy_path(node)
         cores = node.tools[Lscpu].get_core_count()
         vm_count = int(cores / cpus_per_vm)
         failures = 0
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -180,15 +183,19 @@ def _mshv_stress_vm_create(
...
-        dmesg_str = node.tools[Dmesg].get_output()
-        dmesg_path = log_path / f"dmesg_{times}_{cpus_per_vm}_{mem_per_vm_mb}"
-        with open(str(dmesg_path), "w") as f:
-            f.write(dmesg_str)
         assert_that(failures).is_equal_to(0)
 
+    def _get_disk_img_copy_path(self, node: Node) -> PurePath:
+        # Azure temporary disk is mounted at /mnt. It has more space then OS
+        # disk. Use it for storing copies of the disk image if it exists.
+        if node.tools[Ls].path_exists("/mnt"):
```
LiliDeng, line 194: in some case, it is /mnt/resource

anirudhrb: This test can only run on our dom0 image (because of the /dev/mshv check). There it is /mnt. So, I deliberately hard coded it.

pupacha (approved):

microsoft/testsuites/mshv/mshv_root_tests.py
squirrelsc: If it doesn't run within Azure, add a platform requirement to skip Azure.

anirudhrb: It runs within Azure if you deploy the dom0 VHD. Otherwise, it will skip in before_suite.

squirrelsc: Is there an easy way to check if it's dom0 testing scenario? It helps to skip this test on other test runs.

anirudhrb: We are already doing a check in `before_suite` and skipping this test if it is not dom0. It is a very simple check: if /dev/mshv exists continue with test, else skip.

LiliDeng (approved):

PR #2459, opened 2022-12-08
package: pump up paramiko fix warning in log

squirrelsc:


LiliDeng (approved):

LiliDeng:
I tested one case locally. When use pipeline to test, it always uses poetry to install older version based on internal code.

PR #2460, opened 2022-12-09
[fix issue] support detect vhd and gallery image vm generation.

LiliDeng:
We can get hyperv generation from API for marketplace image and gallery image.
For vhd format, we rely on customers' input, otherwise use default value gen1.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -119,6 +119,14 @@ class SharedImageGallerySchema:
     image_definition: str = ""
     image_version: str = ""
 
+    def __hash__(self) -> int:
+        return hash(
+            f"/subscriptions/{self.subscription_id}/resourceGroups/"
+            f"{self.resource_group_name}/providers/Microsoft.Compute/galleries/"
+            f"{self.image_gallery}/images/{self.image_definition}/versions/"
+            f"{self.image_version}"
+        )
+
 
 @dataclass_json()
 @dataclass
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1153,7 +1153,7 @@ class AzurePlatform(Platform):
         elif azure_node_runbook.shared_gallery:
             azure_node_runbook.marketplace = None
             azure_node_runbook.shared_gallery = self._parse_shared_gallery_image(
-                azure_node_runbook.location, azure_node_runbook.shared_gallery
+                azure_node_runbook.shared_gallery
             )
         elif not azure_node_runbook.marketplace:
             # set to default marketplace, if nothing specified
@@ -1719,7 +1719,7 @@ class AzurePlatform(Platform):
 
     @lru_cache(maxsize=10)  # noqa: B019
     def _parse_shared_gallery_image(
-        self, location: str, shared_image: SharedImageGallerySchema
+        self, shared_image: SharedImageGallerySchema
     ) -> SharedImageGallerySchema:
         new_shared_image = copy.copy(shared_image)
         compute_client = get_compute_client(self)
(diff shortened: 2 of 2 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -118,6 +118,7 @@ class SharedImageGallerySchema:
     image_gallery: str = ""
     image_definition: str = ""
     image_version: str = ""
+    hyperv_generation: str = "1"
```
squirrelsc: Refer to how it's implmented for marketplace image. It doesn't need a property on this level.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2457,6 +2472,10 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
 
             generation = _get_vhd_generation(image_info)
             node_space.features.add(features.VhdGenerationSettings(gen=generation))
+        if azure_runbook.vhd or azure_runbook.shared_gallery:
```
squirrelsc: use `elif` or `else`. The logic is related to runbook type. `elif` and `else` helps understand they are related.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1154,6 +1154,9 @@ def _create_node_runbook(
             azure_node_runbook.shared_gallery = self._parse_shared_gallery_image(
                 azure_node_runbook.location, azure_node_runbook.shared_gallery
             )
+            azure_node_runbook.hyperv_generation = int(
```
squirrelsc: Not add it here, refer to the marketplace image, the logic should be after L1183.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1747,9 +1750,21 @@ def _parse_shared_gallery_image(
                 )
                 if not time:
                     time = gallery_image.publishing_profile.published_date
+                    new_shared_image.image_version = image.name
```
squirrelsc, line 1761: The `_parse_shared_gallery_image` will be used multiple times like marketplace image. Add a cache decorator like market place image.

squirrelsc:
For vhd, please check with @kamalca if the PR of vhd object is ready. We need to support vhd object firstly. Then use runbook for vhd only to get the generation information.

lisa/sut_orchestrator/azure/common.py
LiliDeng: _parse_shared_gallery_image will return type SharedImageGallerySchema, so I add the property for hyperv_generation

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: updated

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py

LiliDeng:
@kamalca any plan for the PR of vhd object, when is it ready?

LiliDeng:
> For vhd, please check with @kamalca if the PR of vhd object is ready. We need to support vhd object firstly. Then use runbook for vhd only to get the generation information.

I don't think there is a dependency relationship between this PR with vhd object.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2151,6 +2163,14 @@ def _get_sig_info(
             expand="ReplicationStatus",
         )
 
+    def _get_detailed_sig(self, shared_image: SharedImageGallerySchema) -> GalleryImage:
```
squirrelsc, line 2167: Add cache on this method too.

LiliDeng: updated

squirrelsc (approved):

PR #2461, opened 2022-12-10
[regression] not throw exception when panic detected in lisa runner

LiliDeng:


lisa/runners/lisa_runner.py
```diff
@@ -354,8 +354,16 @@ def _run_test_task(
                 f"because after test case '{test_result.name}', "
                 f"node(s) cannot be accessible."
             )
-            # check panic when node(s) in bad status
-            environment.nodes.check_kernel_panics()
+            try:
+                # check panic when node(s) in bad status
+                environment.nodes.check_kernel_panics()
+            except Exception as identifier:
```
avylove: Can you be more specific about the exception class here?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -482,3 +482,10 @@ def _trigger_kdump_on_specified_cpu(
                 "The cpu count can't meet the test case's requirement. "
                 f"Expected more than {cpu_num} cpus, actual {cpu_count}"
             )
+
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
+        # kdump cases will trigger crash
+        # therefore we mark the node dirty to prevent future testing on this environment
+        # to aviod detecting the panic call trace wrongly
+        node = kwargs["node"]
+        node.mark_dirty()
```
avylove: As is, there is no reason to assign this to a variable.
```suggestion
        kwargs["node"].mark_dirty()
```

Is there any case there would be no node keyword?

LiliDeng: I think there is the node keyword for every case.

lisa/runners/lisa_runner.py
LiliDeng: updated, thank you.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -482,3 +482,9 @@ def _trigger_kdump_on_specified_cpu(
                 "The cpu count can't meet the test case's requirement. "
                 f"Expected more than {cpu_num} cpus, actual {cpu_count}"
             )
+
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
```
squirrelsc, line 486: This can expand the argument like test case. `    def after_case(self, log: Logger, node: Node) -> None:`

lisa/runners/lisa_runner.py
```diff
@@ -354,8 +354,16 @@ def _run_test_task(
...
             )
-            # check panic when node(s) in bad status
-            environment.nodes.check_kernel_panics()
+            try:
+                # check panic when node(s) in bad status
+                environment.nodes.check_kernel_panics()
+            except LisaException as identifier:
+                # not throw exception here, since it will cancel all tasks
+                # just print log here
+                self._log.debug(
```
squirrelsc, line 363: Besides log the failure, the environment should be marked as dirty, so it won't be reused.

lisa/runners/lisa_runner.py
```diff
@@ -354,8 +354,16 @@ def _run_test_task(
                 f"because after test case '{test_result.name}', "
                 f"node(s) cannot be accessible."
             )
-            # check panic when node(s) in bad status
-            environment.nodes.check_kernel_panics()
+            try:
+                # check panic when node(s) in bad status
+                environment.nodes.check_kernel_panics()
```
squirrelsc, line 359: if there is kernel_panic, and the test result is passed, it should be marked as failed.

PR #2464, opened 2022-12-13
Azure: support detection marketplace image Architecture

LiliDeng:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1905,3 +1905,77 @@ def can_disable(cls) -> bool:
...
     def enabled(self) -> bool:
         return True
+
+
+@dataclass_json()
+@dataclass()
+class ArchitectureSettings(schema.FeatureSettings):
+    type: str = "Architecture"
+    # Architecture in hyper-v
+    arch: str = "x64"
```
squirrelsc: restrict it to accept one of x64 and arm64.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1905,3 +1905,77 @@ def can_disable(cls) -> bool:
...
+
+    def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
+        assert isinstance(
+            capability, ArchitectureSettings
+        ), f"actual: {type(capability)}"
+
+        value = ArchitectureSettings()
+        if self.arch or capability.arch:
+            value.arch = getattr(search_space, f"{method_name}_setspace_by_priority")(
+                self.arch, capability.arch, ["x64", "Arm64"]
```
dsrivastavv, line 1963: Do we need to handle AMD as well?

LiliDeng: Thanks for the point, I think `x86-64 (also known as x64, x86_64, AMD64, and Intel 64) `, but will double check the return value of amd64 size.

dsrivastavv: I feel it would be worth separating intel and AMD. For example, I saw failures for nested cases in AMD but not intel. This would make it easy to see that certain failures are CPU vendor specific

squirrelsc: This field is for CPU arch only. The VHD with different arc maps to different VM sizes. The CPU info is a kind of information. We haven't collected it, but it's more related to vm sizes, not related to vhd.

lisa/sut_orchestrator/azure/features.py

lisa/sut_orchestrator/azure/features.py
LiliDeng: @somil55 only two values return from API level (image/vmsize), x64 and Arm64, for your case, we may think about how to achieve it in another way.

squirrelsc (approved):

lisa/sut_orchestrator/azure/features.py
LiliDeng: @somil55 agree?

PR #2466, opened 2022-12-13
Explicitly call out Ubuntu instead of just Debian

contre:
While Ubuntu is derived from Debian, explicitly call out the derived type instead

### DIFF
--- a/lisa/tools/ping.py
+++ b/lisa/tools/ping.py
@@ -4,7 +4,7 @@ import re
 from typing import Optional, Type
 
 from lisa.executable import Tool
-from lisa.operating_system import Debian
+from lisa.operating_system import Debian, Ubuntu
 from lisa.util import UnsupportedDistroException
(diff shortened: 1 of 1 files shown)


lisa/tools/ping.py
```diff
@@ -27,7 +27,7 @@ def can_install(self) -> bool:
         return True
 
     def install(self) -> bool:
-        if isinstance(self.node.os, Debian):
+        if isinstance(self.node.os, Ubuntu) or isinstance(self.node.os, Debian):
```
squirrelsc, line 30: If there is no difference on logic, it doesn't need to call out Ubuntu. The `isinstance` returns True, if it's Ubuntu, because the Ubuntu is a subclass of Debian.

contre: You're right. This issue ended up being a specific use of the ping command with node execute and not the Ping tool

PR #2468, opened 2022-12-14
Enable MOD_STRIP in kernel source installer

adityagesh:
Set INSTALL_MOD_STRIP=1 for kernel source installer.
The initrd file of new kernel is close to 512M (initrd of older kernel being ~25M) since the debug modules are enabled.
This causes few tests to fail eg: kdump and smoke test in small SKU etc.

### DIFF
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -137,7 +137,9 @@ class SourceInstaller(BaseInstaller):
         make = node.tools[Make]
         make.make(arguments="modules", cwd=code_path, sudo=True)
 
-        make.make(arguments="modules_install", cwd=code_path, sudo=True)
+        make.make(
+            arguments="INSTALL_MOD_STRIP=1 modules_install", cwd=code_path, sudo=True
+        )
 
         make.make(arguments="install", cwd=code_path, sudo=True)
 


lisa/transformers/kernel_source_installer.py
```diff
@@ -137,7 +137,9 @@ def _install_build(self, node: Node, code_path: PurePath) -> None:
         make = node.tools[Make]
         make.make(arguments="modules", cwd=code_path, sudo=True)
 
-        make.make(arguments="modules_install", cwd=code_path, sudo=True)
+        make.make(
```
squirrelsc, line 140: When is it needed? Does it affect other usage? Please explain and add into commit description.

adityagesh: The Linux Next and Linux Stable builds are using this transformer to install kernel from source. It results in very large initrd image.
Should this be changed to a parameter for the kernel source installer transformer instead of statically adding it to modules_install?

squirrelsc: If there is no side effect of this change, it can be static here.

adityagesh: Updated commit

squirrelsc (approved):

PR #2470, opened 2022-12-14
Use ping tool instead of node.execute

contre:


microsoft/testsuites/network/common.py
```diff
@@ -169,18 +169,17 @@ def sriov_vf_connection_test(
...
             )
-            if cmd_result.exit_code == 0:
+            if ping_result:
                 break
-        cmd_result.assert_exit_code(
-            message=f"fail to ping {dest_ip} from {source_node.name} to "
-            f"{dest_node.name} after retry {max_retry_times}"
-        )
+        assert ping_result, f"fail to ping {dest_ip} from {source_node.name} to "
+        f"{dest_node.name} after retry {max_retry_times}"
```
LiliDeng:
```
        assert ping_result, (
            f"fail to ping {dest_ip} from {source_node.name} to "
            f"{dest_node.name} after retry {max_retry_times} times"
        )
```

LiliDeng: without parentheses, only this ```f"fail to ping {dest_ip} from {source_node.name} to "``` can be printed out

contre: Done.

LiliDeng (approved):

PR #2472, opened 2022-12-15
update test result status and message when get call trace from serial console log

LiliDeng:
address comments from #2461


lisa/runners/lisa_runner.py
```diff
@@ -359,7 +359,10 @@ def _run_test_task(
                 environment.nodes.check_kernel_panics()
             except LisaException as identifier:
                 # not throw exception here, since it will cancel all tasks
-                # just print log here
+                # just print log here and set test result status as failed
+                if "found panic in serial log" in str(identifier):
```
squirrelsc: Define an exception type like `KernelPanicException`, instead of check by the string.

LiliDeng: added KernelPanicException

squirrelsc (approved):

PR #2473, opened 2022-12-17
Update transformer: override other repos

kamalca:
When a repo url is provided, override the other repos. This allows us to guarantee that the packages are being pulled from the repo we intend. We had to leave the main repo because tests may need to install packages that come from that repo, but there should not be any updates in that repo.

Other changes allow the transformer to be more general purpose. Examples:
Update packages with default repos: repo_url:""
Update from specific repo: repo_url:"$url"
Update from proposed repo: repo_url:"$url" proposed:True

### DIFF
--- a/lisa/transformers/upgrade_packages.py
+++ b/lisa/transformers/upgrade_packages.py
@@ -6,7 +6,7 @@ from dataclasses_json import dataclass_json
 from lisa import schema
 from lisa.node import Node, quick_connect
 from lisa.operating_system import Debian, Ubuntu
-from lisa.tools import Uname
+from lisa.tools import Sed, Uname
 from lisa.transformer import Transformer
 from lisa.util import field_metadata, subclasses
 from lisa.util.logger import Logger, get_logger
@@ -16,12 +16,12 @@ from lisa.util.logger import Logger, get_logger
 @dataclass
 class UpgradeInstallerSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
     repo_url: str = field(
-        default="http://azure.archive.ubuntu.com/ubuntu/",
+        default="",
         metadata=field_metadata(required=True),
     )
 
     proposed: bool = field(
-        default=True,
+        default=False,
         metadata=field_metadata(required=True),
     )
 
@@ -119,7 +119,9 @@ class UnattendedUpgradeInstaller(UpgradeInstaller):
         )
 
     def install(self) -> None:
-        self._update_repo()
+        if self.runbook.repo_url.strip():
+            self._update_repo()
+
         self._update_packages()
 
     def _update_repo(self) -> None:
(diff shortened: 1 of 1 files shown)


squirrelsc (dismissed):

lisa/transformers/upgrade_packages.py
```diff
@@ -133,6 +135,14 @@ def _update_repo(self) -> None:
             release
         ), f"cannot find codename from the os version: {node.os.information}"
 
+        node.execute(
+            f"sed -E '/{release} main/!s/(.*)/#\\1/' -i /etc/apt/sources.list",
+            sudo=True,
+            shell=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="fail to run unattended-upgrade",
```
johnsongeorge-w: Error message needs to be updated.

johnsongeorge-w (dismissed):

lisa/transformers/upgrade_packages.py
```diff
@@ -133,6 +135,14 @@ def _update_repo(self) -> None:
             release
         ), f"cannot find codename from the os version: {node.os.information}"
 
+        node.execute(
+            f"sed -E '/{release} main/!s/(.*)/#\\1/' -i /etc/apt/sources.list",
```
LiliDeng: use sed tool

kamalca: Resolved

lisa/transformers/upgrade_packages.py
```diff
@@ -119,7 +119,9 @@ def validate(self) -> None:
         )
 
     def install(self) -> None:
-        self._update_repo()
+        if self.runbook.repo_url.strip():
+            self._update_repo()
```
squirrelsc, line 117: how is the repo_url used?

lisa/transformers/upgrade_packages.py
```diff
@@ -16,12 +16,12 @@
 @dataclass
 class UpgradeInstallerSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
     repo_url: str = field(
-        default="http://azure.archive.ubuntu.com/ubuntu/",
+        default="",
```
squirrelsc: The repo url can be here for default behavior. It's not easy to use, for each required value. Unless this value is optional. But it looks not.

squirrelsc: If it's not a required value, please remove the below `metadata=field_metadata(required=True),`

lisa/transformers/upgrade_packages.py
kamalca: It specifies which repo to take the updates from. It might also be used in the future to specify a ppa repo. If no repo url is specified, then it will not change the repo list.

squirrelsc: The logic is confusing, if the repo_url is not used. Can it be changed to a meaningful flag?

lisa/transformers/upgrade_packages.py
kamalca: Resolved

lisa/transformers/upgrade_packages.py
kamalca: I feel like this logic makes sense: only update the repo list if a repo url is provided.

squirrelsc: If the repo url is used as a string, it makes sense. But it's not used, so it's to use a string empty or not to run logic or not. This should be done by a bool value.

kamalca: The repo url is used as a string in the `_update_repo` function

squirrelsc: I see. I missed it...

kamalca:
@squirrelsc This PR is still pending your approval

squirrelsc (dismissed):

kamalca:
Rebased the branch. No other changes made to the PR

squirrelsc (approved):

PR #2476, opened 2022-12-19
Add VMGS support

kamalca:
This makes changes to the ARM template to enable deploying from VHD and VMGS for CVMs.

It also changes the VHD field in the Azure Node schema to be an object instead of a string.

### DIFF
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -8,6 +8,12 @@
                 "description": "storage name for boot diagnosis"
             }
         },
+        "vhd_storage_name": {
+            "type": "string",
+            "metadata": {
+                "description": "storage name for copied vhds"
+            }
+        },
         "location": {
             "type": "string",
             "metadata": {
@@ -260,7 +266,7 @@
                 "name": "imageCopy",
                 "count": "[variables('node_count')]"
             },
-            "condition": "[not(empty(parameters('nodes')[copyIndex('imageCopy')]['vhd']))]",
+            "condition": "[and(not(empty(parameters('nodes')[copyIndex('imageCopy')]['vhd'])), empty(parameters('nodes')[copyIndex('imageCopy')]['vmgs']))]",
             "name": "[concat(parameters('nodes')[copyIndex('imageCopy')]['name'], '-image')]",
             "location": "[parameters('location')]",
             "properties": {
@@ -276,6 +282,33 @@
             }
         },
         {
+            "type": "Microsoft.Compute/disks",
+            "apiVersion": "2021-04-01",
+            "name": "[concat(parameters('nodes')[copyIndex('diskCopy')]['name'], '-disk')]",
+            "location": "[parameters('location')]",
+            "condition": "[not(empty(parameters('nodes')[copyIndex('diskCopy')]['vmgs']))]",
+            "copy": {
+                "name": "diskCopy",
+                "count": "[variables('node_count')]"
+            },
+            "sku": {
+                "name": "Standard_LRS"
+            },
+            "properties": {
+                "osType": "Linux",
+                "hyperVGeneration": "V2",
+                "securityProfile": {
+                    "securityType": "ConfidentialVM_VMGuestStateOnlyEncryptedWithPlatformKey"
+                },
+                "creationData": {
+                    "createOption": "ImportSecure",
+                    "storageAccountId": "[concat('/subscriptions/0cc2a67a-58b9-4e4f-98a8-bfa46a28e896/resourceGroups/',parameters('shared_resource_group_name'),'/providers/Microsoft.Storage/storageAccounts/', parameters('vhd_storage_name'))]",
+                    "securityDataUri": "[parameters('nodes')[copyIndex('diskCopy')]['vmgs']]",
+                    "sourceUri": "[parameters('nodes')[copyIndex('diskCopy')]['vhd']]"
+                }
+            }
+        },
+        {
             "apiVersion": "2022-03-01",
             "type": "Microsoft.Compute/virtualMachines",
             "copy": {
@@ -288,7 +321,7 @@
             "plan": "[parameters('nodes')[copyIndex('vmCopy')]['purchase_plan']]",
             "dependsOn": [
                 "[resourceId('Microsoft.Compute/availabilitySets', variables('availability_set_name'))]",
-                "[resourceId('Microsoft.Compute/images', concat(parameters('nodes')[copyIndex('vmCopy')]['name'], '-image'))]",
+                "[if( not(empty(parameters('nodes')[copyIndex('vmCopy')]['vmgs'])), concat(parameters('nodes')[copyIndex('vmCopy')]['name'], '-disk'), resourceId('Microsoft.Compute/images',concat(parameters('nodes')[copyIndex('vmCopy')]['name'],'-image')))]",
                 "[concat(parameters('nodes')[copyIndex('vmCopy')]['name'],'-nics')]"
             ],
             "properties": {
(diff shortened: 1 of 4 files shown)


kamalca:
@squirrelsc I think this is close to being done. I would like you to tell me if I am on the right track.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -132,8 +139,12 @@ class AzureNodeSchema:
     shared_gallery_raw: Optional[Union[Dict[Any, Any], str]] = field(
         default=None, metadata=field_metadata(data_key="shared_gallery")
     )
-    vhd: str = ""
-    vmgs: str = ""
+    vhd_raw: Optional[Union[Dict[Any, Any], str]] = field(
+        default=None, metadata=field_metadata(data_key="vhd")
+    )
+    vmgs_raw: Optional[str] = field(
```
squirrelsc: vmgs is new thing, so it doesn't need to be a string in Azure Node level. It should be in vhd object.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -297,6 +312,86 @@ def shared_gallery(self, value: Optional[SharedImageGallerySchema]) -> None:
         else:
             self.shared_gallery_raw = value.to_dict()  # type: ignore
 
+    @property
+    def vhd(self) -> Optional[str]:
```
squirrelsc: The return type should be VhdSchema

lisa/sut_orchestrator/azure/common.py
```diff
@@ -297,6 +312,86 @@ def shared_gallery(self, value: Optional[SharedImageGallerySchema]) -> None:
...
+                f"a dictionary with a vhd and vmgs."
+            )
+        self._vhd = vhd
+        if vhd:
+            return vhd
+        else:
+            return None
+
+    @vhd.setter
+    def vhd(self, value: Optional[str]) -> None:
```
squirrelsc: The setter type should be VhdSchema

lisa/sut_orchestrator/azure/common.py
```diff
@@ -111,6 +111,13 @@ class SharedImageGallerySchema:
     image_version: str = ""
 
 
+@dataclass_json()
```
squirrelsc, line 140: The VhdSchema commit should be a separated PR, so it can be gotten in firstly.

kamalca:
@squirrelsc
Changing vhd property to an object would change AzureNodeSchema.vhd from a string to an object and then you would have to access AzureNodeSchema.vhd.vhd and AzureNodeSchema.vhd.vmgs. This felt clunky to me. Is this the way that you want vhd and vmgs implemented?

How I designed it now, vhd and vmgs both act as strings (AzureNodeSchema.vhd and AzureNodeSchema.vmgs) but the underlying object is a VhdSchema with vhd and vmgs as fields

squirrelsc:
> @squirrelsc Changing vhd property to an object would change AzureNodeSchema.vhd from a string to an object and then you would have to access AzureNodeSchema.vhd.vhd and AzureNodeSchema.vhd.vmgs. This felt clunky to me. Is this the way that you want vhd and vmgs implemented?
> 
> How I designed it now, vhd and vmgs both act as strings (AzureNodeSchema.vhd and AzureNodeSchema.vmgs) but the underlying object is a VhdSchema with vhd and vmgs as fields

There are more fields coming to a vhd. The structure is used to same all of them in one object. So the runbook is clear. That's why, it should be like marketplace image and SIG, as an object.

you can specify its name like vhd.path and vhd.vmgs_path. The key thing is about the structure of those data. The name can be changed.

lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -422,8 +443,74 @@
...
+                    "parameters": [
+                        {
+                            "name": "diskName",
+                            "type": "string"
+                        }
+                    ],
+                    "output": {
+                        "type": "object",
+                        "value": {
+                            "createOption": "Attach",
```
squirrelsc, line 513: Does the attach method support to run in parallel? My concern is the attached vhd supports only one vm, and it's dirty after used once.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1514,12 +1514,12 @@ def _enable_secure_boot(cls, *args: Any, **kwargs: Any) -> None:
                 "DiskWithVMGuestState" if settings.encrypt_disk else "VMGuestStateOnly"
             )
 
-            virtual_machines["properties"]["storageProfile"]["osDisk"][
+            template["functions"][0]["members"]["getOSImage"]["output"]["value"][
                 "managedDisk"
             ] = (
-                "[if(not(equals(parameters('nodes')[copyIndex('vmCopy')]['disk_type'], "
+                "[if(not(equals(parameters('node')['disk_type'], "
```
squirrelsc, line 1570: Is it possible to move this string into arm template?

lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -275,6 +281,33 @@
...
+            },
+            "properties": {
+                "osType": "Linux",
+                "hyperVGeneration": "V2",
+                "securityProfile": {
+                    "securityType": "ConfidentialVM_VMGuestStateOnlyEncryptedWithPlatformKey"
+                },
+                "creationData": {
+                    "createOption": "ImportSecure",
+                    "storageAccountId": "[concat('/subscriptions/0cc2a67a-58b9-4e4f-98a8-bfa46a28e896/resourceGroups/',parameters('shared_resource_group_name'),'/providers/Microsoft.Storage/storageAccounts/', parameters('vhd_storage_name'))]",
```
squirrelsc: How does the storageAccountId use to? The subscription id shouldn't be hard coded. We can pass in the whole string, if it's needed.

lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -275,6 +281,33 @@
...
+            "copy": {
+                "name": "diskCopy",
+                "count": "[variables('node_count')]"
+            },
+            "sku": {
+                "name": "Standard_LRS"
+            },
+            "properties": {
+                "osType": "Linux",
+                "hyperVGeneration": "V2",
```
squirrelsc: This shouldn't be hard coded, even the cvm supports only v2.

lisa/sut_orchestrator/azure/arm_template.json
squirrelsc: If it's to create a copy of vhd, will it be deleted after vm destroyed?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1162,7 +1162,7 @@ def _create_node_arm_parameters(
         arm_parameters = AzureNodeArmParameter.from_node_runbook(runbook)
 
         os_disk_size = 30
-        if arm_parameters.vhd:
+        if arm_parameters.vhd and len(arm_parameters.vhd) > 0:
```
squirrelsc: It doesn't need to check the length of a string. `if str_value` works like False for both None and "".

lisa/sut_orchestrator/azure/arm_template.json
kamalca: It creates the disk from the storage account. The disk will be deleted with the RG

squirrelsc: > It creates the disk from the storage account. The disk will be deleted with the RG

If so, it shouldn't be in shared resource group, it should be in the same resource group as current VM. If so, I feel it doesn't need to specify subscription and resource group. Please test if it works with environment, which has two VMs. Make sure there is no conflict names.

lisa/sut_orchestrator/azure/platform_.py
kamalca: I was seeing really weird behavior during testing where arm_parameters.vhd == "" and it would still evaluate to true. I'm not sure if it had something to do with arm_parameters.vhd being a `@property` where maybe it evaluated as true while it was still a function handle.

squirrelsc: I had a quick test, the property works like expected.

```python
class A:
    @property
    def b(self) -> str:
        return "aaa"


x = A()
print(x.b)
if x.b:
    print("if ")
else:
    print("else")

```

kamalca: Resolved

lisa/sut_orchestrator/azure/arm_template.json
kamalca: Resolved hard-coded string

lisa/sut_orchestrator/azure/arm_template.json
kamalca: Resolved

lisa/sut_orchestrator/azure/features.py
kamalca: Is it possible for a feature to add arm template parameters? Would it be done in `on_before_deployment` or somewhere else? I would probably need three new parameters to control all the possible CVM configurations.

lisa/sut_orchestrator/azure/arm_template.json
kamalca: I confirmed that deployment works with two VMs (setting min_count to 2) works.
Is there another scenario that I should test?

kamalca:
Resolved merge conflicts

lisa/sut_orchestrator/azure/common.py
```diff
@@ -191,13 +203,15 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
                 "subscription_id",
                 "marketplace_raw",
                 "shared_gallery_raw",
-                "vhd",
+                "vhd_raw",
+                "vms_raw",
```
squirrelsc: Should it be removed?

kamalca: Resolved

lisa/sut_orchestrator/azure/features.py
squirrelsc: Yes, the arm template parameters are modified in this class too. It's good to control by parameters if it's possible.

lisa/sut_orchestrator/azure/arm_template.json
squirrelsc: The "shared_resource_group_name" is the shared resource group. My understanding is that this disk should be deleted, when test is done. So it shouldn't be in a shared place, it should be in the same resource group of VM.

lisa/sut_orchestrator/azure/common.py
squirrelsc: I mean the vms_raw should be removed. But vhd_raw is worth to filter out whitespaces. So it can be compatible with Azure pipelines.

kamalca: That is what I did. Please see new changes

lisa/sut_orchestrator/azure/arm_template.json
kamalca: The shared resource group is the location of the vhd. How vhd already worked before is that it is copied to the shared storage. The disk itself is created in the same RG as as the VM. This is no different than the previous way that vhd worked.

squirrelsc: ok, I see. Thank you let me know.

squirrelsc (approved):

lisa/sut_orchestrator/azure/features.py
kamalca: I will makes these changes in another PR.

PR #2479, opened 2022-12-21
[regression] fix ValidationError arch - Must be one of: x64, Arm64.

LiliDeng:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1913,10 +1912,13 @@ def enabled(self) -> bool:
 class ArchitectureSettings(schema.FeatureSettings):
     type: str = "Architecture"
     # Architecture in hyper-v
-    arch: str = field(
-        default="x64",
+    arch: str = field(  # type:ignore
```
squirrelsc: If so, the type should be consistent with the factory. Refer to `DiskOptionSettings/disk_type` in schema.py.

squirrelsc: BTW, what's the cause behind? I don't see how it got fixed by this change.

LiliDeng: I am still now sure, I just see when issue happened, the setting is `arch {'is_allow_set': False, 'items': ['x64']}`

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1913,10 +1912,13 @@ def enabled(self) -> bool:
 class ArchitectureSettings(schema.FeatureSettings):
     type: str = "Architecture"
     # Architecture in hyper-v
-    arch: str = field(
-        default="x64",
+    arch: Optional[Union[search_space.SetSpace[str], str]] = field(  # type:ignore
```
squirrelsc: Keep it as string, and assert req equals cap in `_call_requirement_method`. The reason is that the arch is exclusive each other on both image and vm size. It's impossible to support both in one of them.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1944,11 +1946,9 @@ def check(self, capability: Any) -> search_space.ResultReason:
...
         ), f"actual: {type(capability)}"
         result = super().check(capability)
-
-        result.merge(
-            search_space.check_setspace(self.arch, capability.arch),
-            "vhd Architecture",
-        )
+        if self.arch != capability.arch:
+            result.result = False
+            result.reasons.append("image arch should be consistent with vm size arch")
```
LiliDeng: are these changes good?

squirrelsc: this part is ok, please see my previous comments.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1913,12 +1912,7 @@ def enabled(self) -> bool:
 class ArchitectureSettings(schema.FeatureSettings):
     type: str = "Architecture"
     # Architecture in hyper-v
-    arch: str = field(
-        default="x64",
-        metadata=field_metadata(
-            validate=validate.OneOf(["x64", "Arm64"]),
-        ),
-    )
+    arch: str = "x64"
```
squirrelsc: Can we keep the validation? The value should be normalized before set to this field. So it wouldn’t bring other failures later.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1957,6 +1949,7 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
             capability, ArchitectureSettings
         ), f"actual: {type(capability)}"
 
+        assert self.arch == capability.arch, "req and capability should be the same"
```
squirrelsc: Use aseert_that to print original values

lisa/sut_orchestrator/azure/features.py
LiliDeng: I tried with validation, but still encounter this issue. Where should I add the normalized value?

lisa/sut_orchestrator/azure/features.py

lisa/sut_orchestrator/azure/features.py
squirrelsc: Is it possible there is an invalid value? What’s the value before assigned?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1944,11 +1944,9 @@ def check(self, capability: Any) -> search_space.ResultReason:
...
         ), f"actual: {type(capability)}"
         result = super().check(capability)
-
-        result.merge(
-            search_space.check_setspace(self.arch, capability.arch),
-            "vhd Architecture",
-        )
+        if self.arch != capability.arch:
+            result.result = False
+            result.reasons.append("image arch should be consistent with vm size arch")
```
squirrelsc: Log both image and env arch values.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1957,11 +1955,14 @@ def _call_requirement_method(self, method_name: str, capability: Any) -> Any:
...
+        assert_that(self.arch).described_as(
+            "req and capability should be the same"
+        ).is_equal_to(capability.arch)
+
         value = ArchitectureSettings()
-        if self.arch or capability.arch:
-            value.arch = getattr(search_space, f"{method_name}_setspace_by_priority")(
-                self.arch, capability.arch, ["x64", "Arm64"]
-            )
+        value.arch = next(
```
squirrelsc: It doesn’t so complicated, just take one of them.

lisa/sut_orchestrator/azure/features.py

lisa/sut_orchestrator/azure/features.py

squirrelsc (approved):

PR #2482, opened 2022-12-26
Add custom_blob_name in VhdTransformerSchema and try-except to catch the error of waagent in transformer

lubaihua33:
1. Add custom_blob_full_name in VhdTransformerSchema. So users can define their own blob name.
2. Change the level of log "vhd copied" into info. 
3. Ignore any error when running waagent deprovision in transformer. So the image which has an error when running waagent deprovison can still continue to export VHD .

### DIFF
--- a/lisa/sut_orchestrator/azure/transformers.py
+++ b/lisa/sut_orchestrator/azure/transformers.py
@@ -77,6 +77,9 @@ class VhdTransformerSchema(schema.Transformer):
     storage_account_name: str = ""
     container_name: str = DEFAULT_EXPORTED_VHD_CONTAINER_NAME
     file_name_part: str = ""
+    # Users may want to export VHD with the name they defined. For example,
+    # OpenLogic-CentOS-7.9-20221227.vhd, or OpenLogic/CentOS/7.9/20221227.vhd
+    custom_blob_name: str = ""
 
     # restore environment or not
     restore: bool = False
@@ -194,7 +197,10 @@ class VhdTransformer(Transformer):
             resource_group_name=runbook.shared_resource_group_name,
         )
 
-        path = _generate_vhd_path(container_client, runbook.file_name_part)
+        if runbook.custom_blob_name:
+            path = runbook.custom_blob_name
+        else:
+            path = _generate_vhd_path(container_client, runbook.file_name_part)
         vhd_path = f"{container_client.url}/{path}"
         blob_client = container_client.get_blob_client(path)
         blob_client.start_copy_from_url(sas_url, metadata=None, incremental_copy=False)
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -993,7 +993,7 @@ def wait_copy_blob(
     if timeout_timer.elapsed() >= timeout:
         raise LisaException(f"wait copying VHD timeout: {vhd_path}")
 
-    log.debug("vhd copied")
+    log.info("vhd copied")
 
 
 def get_share_service_client(
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -45,7 +45,12 @@ class Waagent(Tool):
         return False
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
-        self._command = "waagent"
+        if isinstance(self.node.os, CoreOs):
+            self._command = (
+                "/usr/share/oem/python/bin/python /usr/share/oem/bin/waagent"
+            )
+        else:
+            self._command = "waagent"
         self._distro_version: Optional[str] = None
 
     def get_version(self) -> str:


lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -194,7 +201,10 @@ def _export_vhd(self, platform: AzurePlatform, virtual_machine: Any) -> str:
             resource_group_name=runbook.shared_resource_group_name,
         )
 
-        path = _generate_vhd_path(container_client, runbook.file_name_part)
+        if not runbook.custom_blob_full_name:
+            path = _generate_vhd_path(container_client, runbook.file_name_part)
+        else:
+            path = runbook.custom_blob_full_name
```
LiliDeng: there is a possibility that this blob storage account is different with the one used to construct container_client, then copy will fail

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -145,10 +147,15 @@ def _prepare_virtual_machine(self, node: RemoteNode) -> None:
...
-        wa.deprovision()
+        try:
+            # prepare vm for exporting
+            wa = node.tools[Waagent]
+            node.execute("export HISTSIZE=0", shell=True)
+            wa.deprovision()
+        except Exception as identifier:
+            # It happens on some VMs. This error should be caught in test cases
+            # not here. So ignore any error here and collect information only.
+            node.log.debug(f"error on run waagent deprovision : {identifier}")
```
LiliDeng: give some detailed error + image info as examples in comments

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -77,6 +77,8 @@ class VhdTransformerSchema(schema.Transformer):
     storage_account_name: str = ""
     container_name: str = DEFAULT_EXPORTED_VHD_CONTAINER_NAME
     file_name_part: str = ""
+    # Users can define the blob full name
+    custom_blob_full_name: str = ""
```
squirrelsc: Give an example of the value. When is it needed? Can it use the same value as vhd?

lisa/sut_orchestrator/azure/transformers.py
lubaihua33: The custom_blob_full_name doesn't contain the storage account and container name. It's just the blob name. Let me change it to "custom_blob_name".

lisa/sut_orchestrator/azure/transformers.py
lubaihua33: For example the image "kinvolk flatcar-container-linux-corevm alpha 3446.0.0"
It has below error information when running transformer:

2022-12-24 00:09:15.117[1896][ERROR] lisa.RootRunner canceling runner due to exception
  File "D:\a\_work\1\s\lisa\lisa\runner.py", line 204, in start
    transformer.run(
  File "D:\a\_work\1\s\lisa\lisa\transformer.py", line 206, in run
    output_variables = _run_transformers(runbook_builder, phase=phase)
  File "D:\a\_work\1\s\lisa\lisa\transformer.py", line 188, in _run_transformers
    values = transformer.run()
  File "D:\a\_work\1\s\lisa\lisa\transformer.py", line 40, in run
    variables = self._internal_run()
  File "D:\a\_work\1\s\lisa\lisa\sut_orchestrator\azure\transformers.py", line 135, in _internal_run
    self._prepare_virtual_machine(node)
  File "D:\a\_work\1\s\lisa\lisa\sut_orchestrator\azure\transformers.py", line 151, in _prepare_virtual_machine
    wa = node.tools[Waagent]
  File "D:\a\_work\1\s\lisa\lisa\executable.py", line 543, in __getitem__
    return self.get(tool_type=tool_type)
  File "D:\a\_work\1\s\lisa\lisa\executable.py", line 611, in get
    raise LisaException(
lisa.util.LisaException: cannot find [waagent] on [lisa-None-xx-e0-n0], CoreOs, Remote(True) and 
installation of [waagent] isn't enabled in lisa.

In function _get_wala_version, it also has a similar try-except.

lisa/sut_orchestrator/azure/transformers.py
lubaihua33: Added some comments here.
When updating VHD snapshot of MarketPlace image, we want to define the VHD name as "publisher/offer/sku/version.vhd" format. I changed the filed name to "custom_blob_name", it is the blob name, not contain the storage account name and container name.

lisa/sut_orchestrator/azure/transformers.py
lubaihua33: Added a  fix in waagent tool, adding a command for CoreOs in the initialize phase.

squirrelsc (commented):
LGTM, please check with Lili on rest comments.

LiliDeng (dismissed):

LiliDeng (approved):

PR #2483, opened 2022-12-26
[new case] verify_disk_with_fio_verify_option

LiliDeng:


lisa/tools/fdisk.py
```diff
@@ -27,6 +27,8 @@ def make_partition(
         disk_name: str,
         file_system: FileSystem = FileSystem.ext4,
         format: bool = True,
+        first_sector: str = "",
```
squirrelsc: It looks they are not used. The two number is hard to use. When is it needed? If it's to create two partitions, handle it in this method internally should be easier.

microsoft/testsuites/storage/storagesuite.py
```diff
@@ -79,6 +100,75 @@ def verify_disk_with_nobarrier(self, node: Node, environment: Environment) -> No
...
+        disk_count = len(data_disks)
+        assert_that(disk_count).described_as(
+            "At least 2 data disk for testing."
+        ).is_greater_than(1)
+        partition_disks = _format_disk(node, data_disks, last_sector="+100G")
+        _stop_raid(node)
+        _make_mount_raid(
+            node,
+            partition_disks,
+            level="mirror",
```
squirrelsc, line 137: Is it the same as "0"? If so, it doesn't need to support both number and string.

LiliDeng: I thought they are the same, but actually only when set it as mirror, the issue can be repro.

lisa/tools/fdisk.py
LiliDeng: some scenario used default size without specifying the partition size, in this case, need specify the size as 100G

squirrelsc: If it needs 100GB, the argument can be like `disk_size`, instead of the sector. The sector can be detected in this method.

microsoft/testsuites/storage/storagesuite.py
squirrelsc: Add a comment to specify that, in case it's changed to "0" by someone in future.

lisa/tools/fdisk.py
LiliDeng: @squirrelsc update, please review, I will merge it in proper time.. Thanks.

squirrelsc (dismissed):

PR #45, opened 2019-03-20
Fix : HyperV VHD Copy issue

iamshital:
Fixed an issue where, LISAv2 was ignoring the VHD copy if it detected the existing disk identifier on remote host. But, if the VHD is updated, the disk identifier still  remains same, causing LISAv2 to not to use updated VHD.

PR #2485, opened 2022-12-28
Add EOL release check for Ubuntu image in dns upgrade test

lubaihua33:
Some images are end-of-life. There is an error message When upgrading the system. 
This commit adds an end-of-life list for the Ubuntu release. If the return result of "apt-get update" has the "no longer has a release file" message, then judge if the release is end-of-life, if so, raise  ReleaseEndOfLifeException.

In the DNS test case, if catch ReleaseEndOfLifeException, continue to do the next DNS check.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -38,6 +38,7 @@ from lisa.util import (
     BaseClassMixin,
     LisaException,
     MissingPackagesException,
+    ReleaseEndOfLifeException,
     RepoNotExistException,
     filter_ansi_escape,
     find_group_in_lines,
@@ -678,8 +679,15 @@ class Debian(Linux):
     _package_candidate_pattern = re.compile(
         r"([\w\W]*?)(Candidate: \(none\)|Unable to locate package.*)", re.M
     )
+    # E: The repository 'http://azure.archive.ubuntu.com/ubuntu impish-backports Release' does not have a Release file # noqa: E501
     # E: The repository 'http://azure.archive.ubuntu.com/ubuntu groovy Release' no longer has a Release file. # noqa: E501
-    _repo_not_exist_pattern = re.compile("does not have a Release file", re.M)
+    # E: The repository 'http://security.ubuntu.com/ubuntu zesty-security Release' does no longer have a Release file. # noqa: E501
+    _repo_not_exist_patterns: List[Pattern[str]] = [
+        re.compile("does not have a Release file", re.M),
+        re.compile("no longer has a Release file", re.M),
+        re.compile("does no longer have a Release file", re.M),
+    ]
+    end_of_life_releases: List[str] = []
 
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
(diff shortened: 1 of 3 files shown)


lisa/operating_system.py
```diff
@@ -797,13 +805,23 @@ def add_repository(
         if type(self._node.os) == Debian:
             self._node.execute("apt-get update", sudo=True)
 
+    def is_end_of_life_release(self) -> bool:
+        for release in self.end_of_life_releases:
+            if self.information.full_version == release:
+                return True
+        return False
```
LiliDeng: `return self.information.full_version in self.end_of_life_releases`
one line should be enough?

microsoft/testsuites/core/dns.py
```diff
@@ -40,6 +41,28 @@ def verify_dns_name_resolution(self, node: Node) -> None:
...
+    def _upgrade_system(self, node: Node) -> None:
+        if not isinstance(node.os, Posix):
+            raise UnsupportedDistroException(node.os)
+
+        try:
+            node.os.update_packages("")
+        except ReleaseEndOfLifeException as identifier:
+            # For the EOL release, collect the except information and skip upgrading
+            # system.
+            node.log.debug(f"{identifier}")
```
squirrelsc: Use PassedException here, because it's partial passed, not fully. It's like smoke test, when it cannot be rebooted.

microsoft/testsuites/core/dns.py
```diff
@@ -71,15 +94,3 @@ def verify_dns_name_resolution_after_upgrade(self, node: Node) -> None:
                 expected_exit_code_failure_message="fail to run unattended-upgrade",
                 timeout=2400,
             )
-        elif isinstance(node.os, Posix):
-            node.os.update_packages("")
-        else:
-            raise UnsupportedDistroException(node.os)
-        self._check_dns_name_resolution(node)
```
squirrelsc, line 78: Where is this logic?

lisa/operating_system.py

microsoft/testsuites/core/dns.py
lubaihua33: Done

LiliDeng: node.log.debug(identifier) is enough

microsoft/testsuites/core/dns.py
lubaihua33: The self._check_dns_name_resolution(node) check is moved to line #45.

microsoft/testsuites/core/dns.py
lubaihua33: Done

microsoft/testsuites/core/dns.py
```diff
@@ -40,6 +41,26 @@ def verify_dns_name_resolution(self, node: Node) -> None:
     )
     def verify_dns_name_resolution_after_upgrade(self, node: Node) -> None:
         self._check_dns_name_resolution(node)
+        self._upgrade_system(node)
+        self._check_dns_name_resolution(node)
+        node.reboot()
```
squirrelsc, line 54: Even the upgrade is failed, the test case should try to test ping after reboot. I know the original logic doesn't do that, but it can be added this time.

lubaihua33: Yes, this is the right way. Have updated, please review.

squirrelsc (approved):

PR #2486, opened 2022-12-28
[Fix issue] Remove vm size from candidate list if the arch/gen are different with current vm size

LiliDeng:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1345,9 +1345,30 @@ def _select_vm_size(
         assert_that(current_core_count).described_as(
             "Didn't return an integer to represent the current VM size core count."
         ).is_instance_of(int)
+        assert current_vm_size.capability.features
```
squirrelsc, line 1348: Does it need to handle generation too?

LiliDeng: updated

squirrelsc (approved):

PR #2488, opened 2022-12-29
Instead of using hardcoded x86_64 identifiers, use the detected arch

contre:
For several packages to be installed, we have hardcoded the arch to x86_64 which does not work on ARM64. This is addressed by using the detected arch instead of a hardcoded value.

### DIFF
--- a/microsoft/testsuites/xfstests/xfstests.py
+++ b/microsoft/testsuites/xfstests/xfstests.py
@@ -217,10 +217,18 @@ class Xfstests(Tool):
             if isinstance(self.node.os, Oracle):
                 posix_os.install_packages("oracle-softwarecollection-release-el7")
             else:
+                arch = self.node.os.get_kernel_information().hardware_platform
+                if arch != "x86_64":
+                    raise LisaException(
+                        f"This test case does not support {arch}."
+                        "This validation is only for x86_64."
+                    )
                 posix_os.install_packages(packages="centos-release-scl")
                 posix_os.install_packages(
-                    "http://mirror.centos.org/centos/7/os/x86_64/Packages/"
-                    "xfsprogs-devel-4.5.0-22.el7.x86_64.rpm"
+                    (
+                        "http://mirror.centos.org/centos/7/os/x86_64/Packages/"
+                        "xfsprogs-devel-4.5.0-22.el7.x86_64.rpm"
+                    )
                 )
             posix_os.install_packages(
                 packages="devtoolset-7-gcc*", extra_args=["--skip-broken"]
--- a/microsoft/testsuites/display/modetest.py
+++ b/microsoft/testsuites/display/modetest.py
@@ -36,17 +36,18 @@ class Modetest(Tool):
 
     def _install_dep_packages(self) -> None:
         if isinstance(self.node.os, Redhat):
+            arch = self.node.os.get_kernel_information().hardware_platform
             self.node.os.install_packages(
                 (
                     "git",
                     "make",
                     "autoconf",
                     "automake",
-                    "libpciaccess-devel.x86_64",
+                    f"libpciaccess-devel.{arch}",
                     "libtool",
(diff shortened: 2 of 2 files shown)


contre:
There are two additional packages which are not addressed in this PR. 

The first is numactl-devel.x86_64 which is defined in a different way than the others and is not immediately obvious to me what the correct solution should be. Feedback for this issue would be appreciated.

The second is for the cuda driver. I am not certain that the required ARM64 + Nvidia hardware combinations exist in Azure but if so, it should be a straightforward change.

microsoft/testsuites/xdp/xdptools.py
```diff
@@ -115,19 +115,25 @@ def _install(self) -> bool:
...
 
         elif isinstance(self.node.os, Fedora):
+            lscpu = self.node.tools[Lscpu]
+            arch = lscpu.get_architecture()
             if self.node.os.information.version >= "9.0.0":
                 self.node.os.install_packages(
-                    "http://mirror.stream.centos.org/9-stream/AppStream/"
-                    "x86_64/os/Packages/libpcap-devel-1.10.0-4.el9.x86_64.rpm"
+                    (
+                        f"http://mirror.stream.centos.org/9-stream/AppStream/"
```
LiliDeng: no need f for this line

microsoft/testsuites/xdp/xdptools.py
```diff
@@ -115,19 +115,25 @@ def _install(self) -> bool:
...
             else:
                 if isinstance(self.node.os, CentOs):
                     self.node.os.install_packages("iproute-tc")
                 else:
                     self.node.os.install_packages("tc")
                 self.node.os.install_packages(
-                    "https://vault.centos.org/centos/8/PowerTools/"
-                    "x86_64/os/Packages/libpcap-devel-1.9.1-5.el8.x86_64.rpm"
+                    (
+                        f"https://vault.centos.org/centos/8/PowerTools/"
```
LiliDeng: ditto

contre: Done.

microsoft/testsuites/xdp/xdptools.py
contre: Done.

LiliDeng:
@contre do you plan to update it?

contre:
Yes.

@LiliDeng do you have any feedback regarding the numactl driver?

LiliDeng:
> Yes.
> 
> @LiliDeng do you have any feedback regarding the numactl driver?

let's get this merged firstly. for numactl, we had offline discussion, please try that out, and see whether it can work without arch info

LiliDeng:
@contre could you please update this PR, thanks.

contre:
Updated.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -214,13 +214,17 @@ def _install_dep(self) -> None:
...
                 posix_os.install_packages("oracle-softwarecollection-release-el7")
             else:
+                arch = self.node.os.get_kernel_information().hardware_platform
                 posix_os.install_packages(packages="centos-release-scl")
                 posix_os.install_packages(
-                    "http://mirror.centos.org/centos/7/os/x86_64/Packages/"
-                    "xfsprogs-devel-4.5.0-22.el7.x86_64.rpm"
+                    (
+                        f"http://mirror.centos.org/centos/7/os/{arch}/Packages/"
+                        f"xfsprogs-devel-4.5.0-22.el7.{arch}.rpm"
```
LiliDeng: there is no arm64 folder under http://mirror.centos.org/centos/7/os/

contre: I thought it returned aarch64 which does exist

LiliDeng: image

contre: I did not realize that this was a different URL from the others. I've updated to handle this case now.

LiliDeng (approved):

PR #2490, opened 2022-12-29
Add tool to copy back folders from remote machine

dsrivastavv:
Sample test:
```
@TestCaseMetadata(
        description="""
            """,
        priority=2,
    )
    def check_remote_copy(
        self,
        node: Node,
        log: Logger,
        log_path: str,
    ) -> None:
        # setup a 3 level directory structure with files and folders
        mkdir = node.tools[Mkdir]
        mkdir.create_directory("test")
        mkdir.create_directory("test/helloworld1")
        mkdir.create_directory("test/helloworld1/helloworld2")
        mkdir.create_directory("test/helloworld1/helloworld2/helloworld3")

        # create a file in all the directories
        echo = node.tools[Echo]
        echo.write_to_file("hello world", PurePosixPath("test/helloworld1/file1.txt"))
        echo.write_to_file(
            "hello world", PurePosixPath("test/helloworld1/helloworld2/file2.txt")
        )
        echo.write_to_file(
            "hello world",
            PurePosixPath("test/helloworld1/helloworld2/helloworld3/file3.txt"),
        )
        remote_copy = node.tools[RemoteCopy]
        remote_copy.copy_to_local(
            PurePosixPath("test/helloworld1"),
            PurePath(log_path),
            recurse=True,
        )
        remote_copy.copy_to_local(
            PurePosixPath("/boot"),
            PurePath(log_path),
            recurse=False,
        )
        remote_copy.copy_to_local(
            PurePosixPath("/boot"),
            PurePath(log_path) / "boot_with_recurse",
            recurse=True,
        )
        remote_copy.copy_to_local(
            PurePosixPath("test/helloworld1/helloworld2/file2.txt"),
            PurePath(log_path) / "singleton_file",
        )
        remote_copy.copy_to_local(
            PurePosixPath("/boot/vmlinuz-5.4.0-1100-azure"),
            PurePath(log_path),
            recurse=True,
        )
 ```

### DIFF
--- a/lisa/tools/ls.py
+++ b/lisa/tools/ls.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from pathlib import PurePath
 from typing import List, Optional, Type
 
 from lisa.executable import Tool
@@ -26,7 +27,7 @@ class Ls(Tool):
 
     def list(self, path: str, sudo: bool = False) -> List[str]:
         cmd_result = self.run(
-            f"-d {path}/*",
+            f"-p -d {path}/*",
             force_run=True,
             sudo=sudo,
             shell=True,
@@ -49,6 +50,21 @@ class Ls(Tool):
             return cmd_result.stdout.split()
         else:
             return []
+
+    def is_file(self, path: PurePath, sudo: bool = False) -> bool:
+        # If `ls -al <path>` returns more than one line, it is a directory, else
+        # it is a file. This is because `ls -al <path>` returns info of the dir and
+        # parent dir.
+        cmd_result = self.run(
+            f"-al {path}",
+            force_run=True,
+            sudo=sudo,
+            shell=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=f"Failed to check if {path} is a file",
+        )
+
+        return len(cmd_result.stdout.splitlines()) == 1
 
     @classmethod
     def _windows_tool(cls) -> Optional[Type[Tool]]:
--- a/lisa/tools/remote_copy.py
+++ b/lisa/tools/remote_copy.py
@@ -1,0 +1,127 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import os
+from pathlib import PurePath
+
+from lisa.executable import Tool
+from lisa.tools.chown import Chown
+from lisa.tools.cp import Cp
+from lisa.tools.ls import Ls
+from lisa.tools.mkdir import Mkdir
+from lisa.tools.rm import Rm
+from lisa.tools.whoami import Whoami
+
+
+class RemoteCopy(Tool):
+    @property
+    def command(self) -> str:
+        return ""
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _prepare_tmp_copy(
+        self,
+        src: PurePath,
+        is_file: bool = False,
+        recurse: bool = False,
+    ) -> PurePath:
+        # copy file/folder to a temp location
+        tmp_location = PurePath("/tmp")
+
+        if is_file:
+            # we want to copy file at
+            # location : <tmp_location>/<src.parent.name>/<src.name>
+            tmp_dir = tmp_location / src.parent.name
+            tmp_location = tmp_dir / src.name
+            self.node.tools[Mkdir].create_directory(str(tmp_dir), sudo=True)
+        elif not recurse:
+            # we want to copy only the files in `src` folder at
+            # location : <tmp_location>/<src.name>
+            tmp_dir = tmp_location / src.name
+            tmp_location = tmp_dir
+            src = src / "*"
+            self.node.tools[Mkdir].create_directory(str(tmp_dir), sudo=True)
+        else:
+            # we want to copy the folder at
+            # location : <tmp_location>
+            tmp_dir = tmp_location
+            tmp_location = tmp_dir / src.name
+
+        # copy the required file/folder to the temp directory
+        self.node.tools[Cp].copy(src, tmp_dir, sudo=True, recur=recurse)
+
+        # change the owner of the temp directory
+        username = self.node.tools[Whoami].get_username()
+        self.node.tools[Chown].change_owner(
+            tmp_location, user=username, group=username, recurse=True
+        )
+
+        return tmp_location
+
+    def _copy(
+        self,
+        src: PurePath,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by dsrivastavv** [comment]
@darkwhite29 fyi

File: lisa/tools/copy_back.py [hunk 1]
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import os
+from pathlib import PurePath
+
+from lisa.executable import Tool
+from lisa.tools.ls import Ls
+
+
+class CopyBack(Tool):
```
squirrelsc: name it `RemoteCopy`, so it can cover bidirectional operations.

lisa/tools/copy_back.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def copy_dir(
+        self,
+        src: PurePath,
+        dest: PurePath,
+        sudo: bool = False,
```
squirrelsc: this sudo is not used, so this method doesn't support sudo.

lisa/tools/ls.py
```diff
@@ -50,6 +50,17 @@ def list_dir(self, path: str, sudo: bool = False) -> List[str]:
         else:
             return []
 
+    def list_files(self, path: str, sudo: bool = False) -> List[str]:
```
squirrelsc: How about use `ls` cmd to list file only? It doesn't need to run two commands to query files.

lisa/tools/copy_back.py
dsrivastavv: I forgot use it in `ls`. Will add it there

lisa/tools/ls.py
dsrivastavv: I couldn't find a straightforward way to do this. Let me see if I can find any behaviour specific to folders output

lisa/tools/copy_back.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # get list of files and folders in the source directory
+        _files = self.node.tools[Ls].list_files(str(src))
+        _dirs = self.node.tools[Ls].list_dir(str(src))
+        files = [PurePath(file) for file in _files]
+        dirs = [PurePath(dir) for dir in _dirs]
+
+        # copy files
+        for file in files:
+            self.node.shell.copy_back(file, destination_dir / file.name)
```
squirrelsc: We should find a way to support copy with sudo. Can you check if it's possible with the shell.copy_back? If not, we may have to implement it by cat (text only) or another tool. Maybe we can consider to zip the folder, and copy it back?

dsrivastavv: `spurshell.get` doesn't have sudo options. How about copying the folder into a temporary directory with `sudo cp`, changing ownership to current user, and then copying back the content? Using `zip` will introduce extra dependencies on both Remote and local machine

squirrelsc: ok, we can take it as a workaround. Make sure the temp files are deleted, after copied back. It takes longer time to copy twice. We should avoid copying too many files back.

lisa/tools/remote_copy.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # return if recursion is not required
+        if not recur:
+            return
+
+        # copy folders
+        for dir in dirs:
+            self._copy_dir(dir, destination_dir, recur=recur)
+
+    def copy_dir(
```
squirrelsc: The name should be `copy_to_local`. It can support both file and dir implicitly. The arguments can be like 'remote_source` and `local_dest`. So it's clearer.

lisa/tools/remote_copy.py
```diff
@@ -0,0 +1,85 @@
...
+        dest: PurePath,
+        recurse: bool = False,
+    ) -> None:
+        # if sudo is required, copy the files to a temp directory first
+        try:
+            # copy files to the destination directory
+            self._copy_dir(src, dest, recur=recurse)
+        except Exception as e:
+            self._log.debug(
+                f"Failed to copy files to {dest} with error: {e}, trying again"
```
squirrelsc: "trying again" -> "trying again with sudo permission"

lisa/tools/remote_copy.py
```diff
@@ -0,0 +1,85 @@
...
+            username = self.node.tools[Whoami].get_username()
+            self.node.tools[Chown].change_owner(
+                tmp_dir, user=username, group=username, recurse=True
+            )
+
+            # copy files
+            self._copy_dir(tmp_dir, dest, recur=recurse)
+
+            # remove the temp directory
+            self.node.tools[Rm].remove_directory(str(tmp_dir), sudo=True)
```
squirrelsc: this should be in a try/final, so it's always removed.

lisa/tools/remote_copy.py
```diff
@@ -0,0 +1,85 @@
...
+        # if sudo is required, copy the files to a temp directory first
+        try:
+            # copy files to the destination directory
+            self._copy_dir(src, dest, recur=recurse)
+        except Exception as e:
+            self._log.debug(
+                f"Failed to copy files to {dest} with error: {e}, trying again"
+            )
+
+            # create a temp directory and copy the files to it
```
squirrelsc: this logic can be in a separated method like `_prepare_tmp_copy`, so the main logic is easier to understand.

lisa/tools/remote_copy.py
```diff
@@ -0,0 +1,132 @@
...
+        self.node.tools[Cp].copy(src, tmp_dir, sudo=True)
+
+        # change the owner of the file
+        tmp_file = tmp_dir / src.name
+        username = self.node.tools[Whoami].get_username()
+        self.node.tools[Chown].change_owner(tmp_file, user=username, group=username)
+
+        return tmp_file
+
+    def _copy_dir(
```
squirrelsc: The logic of dir covers the logic of file. Merge them together can make the tool is easy to use.

dsrivastavv: My initial though was to do this as well but I think that `recurse` argument is confusing in context of file and the `_prepare_tmp_copy_*` behaves different for file and folder which we need to check, and that would be another SSH call.

squirrelsc: `recurse` is understandable, even if it's copying file. The preparing process should more rely on the behavior of cp. I think it can cover both file and dir scenarios.

squirrelsc (approved):

PR #2492, opened 2023-01-01
[fix issue] skip reload mlx module if it is built-in

LiliDeng:


microsoft/testsuites/network/sriov.py
```diff
@@ -596,24 +598,30 @@ def verify_sriov_ethtool_offload_setting(self, environment: Environment) -> None
...
-            device_vf_sg_settings = client_ethtool.get_device_sg_settings(
-                client_nic_info.lower, True
-            )
-            assert_that(
-                device_vf_sg_settings.sg_setting,
-                "sg setting is not sync into VF.",
-            ).is_equal_to(False)
+            if node.tools[KernelConfig].is_built_in(get_used_config(node)):
+                module_built_in = True
+                break
```
avylove:
```suggestion
module_built_in = any(node.tools[KernelConfig].is_built_in(get_used_config(node)) for environment.nodes.list())
```

microsoft/testsuites/network/sriov.py
```diff
@@ -596,24 +598,30 @@ def verify_sriov_ethtool_offload_setting(self, environment: Environment) -> None
...
+        if not module_built_in:
+            module_in_used = ""
+            for node in environment.nodes.list():
+                module_in_used = remove_module(node)
+            for node in environment.nodes.list():
+                load_module(node, module_in_used)
```
avylove: This looks like a bug. `module_in_used` will be the result of remove_module only for the last node in the list and will be the same for every call to `load_module()`.  Shouldn't it be something like this?

```suggestion
            for node in environment.nodes.list():
                load_module(node, remove_module(node))
```

microsoft/testsuites/network/sriov.py
```diff
@@ -596,24 +598,30 @@ def verify_sriov_ethtool_offload_setting(self, environment: Environment) -> None
...
+
+            # check VF still paired with synthetic nic
+            vm_nics = initialize_nic_info(environment)
+
+            # check VF's scatter-gather feature keep consistent with previous status
+            for _, client_nic_info in vm_nics[client_node.name].items():
```
avylove:
```suggestion
            for client_nic_info in vm_nics[client_node.name].values():
```

squirrelsc (commented):
Please take Avram's comments, others look good to me.

microsoft/testsuites/network/sriov.py
```diff
@@ -596,24 +596,27 @@ def verify_sriov_ethtool_offload_setting(self, environment: Environment) -> None
             ).is_equal_to(False)
 
         # reload sriov modules
-        module_in_used = ""
-        for node in environment.nodes.list():
-            module_in_used = remove_module(node)
-        for node in environment.nodes.list():
-            load_module(node, module_in_used)
+        module_built_in = False
```
avylove: You don't need this line since the value will be set to 'True' or 'False' in the next line.

LiliDeng: fixed

squirrelsc (approved):

PR #2493, opened 2023-01-01
[fix issue] install package which is available in repo

LiliDeng:
changes work well on below images
```
canonical ubuntuserver 18.04-lts latest
canonical 0001-com-ubuntu-server-focal 20_04-lts latest
canonical 0001-com-ubuntu-server-jammy 22_04-lts latest
```

ofed built fail against canonical 0001-com-ubuntu-server-kinetic 22_10 latest, not caused by this PR, will report it out.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -204,6 +204,8 @@ class Infiniband(Feature):
             "python-setuptools",
             "g++",
             "libc6-i386",
+            "cloud-init",
+            "walinuxagent",
         ]
         redhat_required_packages = [
             "git",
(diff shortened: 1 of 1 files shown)


lisa/features/infiniband.py
```diff
@@ -269,10 +271,14 @@ def setup_rdma(self) -> None:  # noqa: C901
                     )
                     node.os.install_packages("kernel-devel")
         elif isinstance(node.os, Ubuntu) and node.os.information.version >= "18.4.0":
-            if node.os.information.version >= "22.10.0":
-                ubuntu_required_packages.append("lib32gcc-9-dev python3-dev")
-            else:
-                ubuntu_required_packages.append("lib32gcc-8-dev python-dev")
+            for package in [
```
squirrelsc, line 274: Why change to this way? It may install more versions than needed. And the progress is slower than install them together.

LiliDeng: it will fail to install package, if any one of package list can't not be found in repo.

squirrelsc (approved):

PR #2494, opened 2023-01-03
[Fix issue] fix configure and package issue on arm64 images

LiliDeng:


microsoft/testsuites/xdp/xdptools.py
```diff
@@ -88,6 +88,7 @@ def _install(self) -> bool:
         # install dependencies
 
         config_envs: Dict[str, str] = {}
+        arch = self.node.tools[Lscpu].get_architecture()
```
squirrelsc: can it be gotten from node.os.version? If not, it should be there for sharing, instead of loading from lscpu every time.

microsoft/testsuites/xdp/xdptools.py
```diff
@@ -158,10 +165,14 @@ def _install(self) -> bool:
         # Errors happen if built with multi-threads. The program may not be
         # ready for concurrent build, but our make tool use multi-thread by
         # default. So set thread count to 1.
+        if arch == "aarch64":
+            update_envs = {"C_INCLUDE_PATH": "/usr/include/aarch64-linux-gnu/"}
```
squirrelsc, line 174: I thought it should be right, if the ARCH is right. Is it?

microsoft/testsuites/xdp/xdptools.py
LiliDeng: it can be got from self.node.os.get_kernel_information().hardware_platform

microsoft/testsuites/xdp/xdptools.py
LiliDeng: no, specify arch doesn't work

squirrelsc (approved):

PR #2499, opened 2023-01-05
DNS improvement for no repo enabled scenario

lubaihua33:
1.  Add check of no repo enabled for Redhat image. No need to check the exit code of the command "yum update -y --disablerepo='*' --enablerepo='*microsoft*'" in _initialize_package_installation. Some Redhat-based images don't set rhui repo.
2.  Raise passedException if RepoNotExistException is caught in DNS test.
3. Raise RepoNotExistException if there is no repo enabled in SUSE image

### DIFF
--- a/microsoft/testsuites/core/dns.py
+++ b/microsoft/testsuites/core/dns.py
@@ -13,7 +13,7 @@ from lisa import (
 )
 from lisa.operating_system import Debian, Posix
 from lisa.tools import Ping
-from lisa.util import PassedException, ReleaseEndOfLifeException
+from lisa.util import PassedException, ReleaseEndOfLifeException, RepoNotExistException
 
 
 @TestSuiteMetadata(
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/core/dns.py
```diff
@@ -44,7 +44,7 @@ def verify_dns_name_resolution_after_upgrade(self, node: Node) -> None:
         self._check_dns_name_resolution(node)
         try:
             self._upgrade_system(node)
-        except ReleaseEndOfLifeException as identifier:
+        except (ReleaseEndOfLifeException, RepoNotExistException) as identifier:
```
squirrelsc, line 49: What if a supported OS missed the repo? It should be an issue, but this change makes it pass.

lubaihua33: I was thinking that this case is to verify DNS. It should make the case only have one purpose. When the image is unable to upgrade, then the case can skip and continue the reboot test.
We should have another case to verify the image's repo and upgrade.

LiliDeng: We have other cases to validate the repo, there it won't skip the test.

LiliDeng:
@lubaihua33 please fix the conflict. thanks.

squirrelsc (dismissed):

LiliDeng (dismissed):

LiliDeng (approved):

PR #2502, opened 2023-01-05
Libvirt: Copy VM disk_img only once to host VM

pupacha:
Currently, the disk_img is being copied remote host for every new environament that is created. With this change, the copying happens only in the once for a single disk_img.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -100,6 +100,9 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         self._next_available_port: int
         self._port_forwarding_lock: Lock
 
+        # Lock used for scp-ing disk image to Remote host VM
+        self._disk_img_copy_lock: Lock
+
         self._host_environment_information_hooks = {
             KEY_HOST_DISTRO: self._get_host_distro,
             KEY_HOST_KERNEL: self._get_host_kernel_version,
@@ -129,6 +132,8 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         # 49512 is the first available private port
         self._next_available_port = 49152
         self._port_forwarding_lock = Lock()
+
+        self._disk_img_copy_lock = Lock()
 
         self.platform_runbook = self.runbook.get_extended_runbook(
             self.__platform_runbook_type(), type_name=type(self).type_name()
@@ -429,8 +434,9 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
 
         if self.host_node.is_remote:
             node_context.os_disk_source_file_path = node_runbook.disk_img
+            host = self.platform_runbook.hosts[0]
             node_context.os_disk_base_file_path = os.path.join(
-                self.vm_disks_dir, os.path.basename(node_runbook.disk_img)
+                host.lisa_working_dir, os.path.basename(node_runbook.disk_img)
             )
         else:
             node_context.os_disk_base_file_path = node_runbook.disk_img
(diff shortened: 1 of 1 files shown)


pupacha:
cc: @anirudhrb @cwize1

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -176,6 +176,10 @@ def _create_node_os_disk(
         self, environment: Environment, log: Logger, node: Node
     ) -> None:
         node_context = get_node_context(node)
+        # Give read permissions before accessing the file.
+        self.host_node.tools[Chmod].chmod(
+            node_context.os_disk_base_file_path, "a+r", sudo=True
```
cwize1: LISA can be run directly on a dev box. So, you can’t assume that passwordless sudo is available on the host node.

pupacha: Thanks for pointing this out. This is only needed when the host is remote. Will add a condition for it.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -95,6 +93,7 @@ def __init__(self, runbook: schema.Platform) -> None:
         self.platform_runbook: BaseLibvirtPlatformSchema
         self.host_node: Node
         self.vm_disks_dir: str
+        self.vm_disk_img: str
```
anirudhrb: Let's use `Path` objects?

pupacha: @anirudhrb I think its fine to have it as str. We dont do much Path operations here. The only usage of `vm_disk_img` is assigning appropriate value to `os_disk_base_file_path` which is str and is used as str in multiple places.

lisa/sut_orchestrator/libvirt/ch_platform.py
pupacha: @cwize1 Please review once again.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -166,6 +165,19 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
                 parent_logger=get_logger("libvirt-platform"),
             )
 
+        vm_disk_img = self.runbook.requirement[type(self).type_name()]["disk_img"]
```
cwize1: @squirrelsc What are your thoughts on this?

squirrelsc: I found there is right usage in this file, please refer to below logic to get the `disk_img` in extended runbook.

```python
            node_runbook: BaseLibvirtNodeSchema = node_space.get_extended_runbook(
                self.__node_runbook_type(), type_name=type(self).type_name()
            )
```

squirrelsc: BTW, the value is in runbook already, it doesn't need to save to platform.

pupacha: @squirrelsc  Yes, `get_extended_runbook` is the right way. But `node_space` (node requirement) is not yet available for platform's `_initialize` to use. Hence, directly parsing the value from runbook.

squirrelsc: It sounds the vhd image should be a platform level config, instead of node level, because it needs to be loaded when initializing. Considering, if it needs to support different images in the same LISA run, how this value is loaded in initializing?

pupacha: VHD ideally should be node requirement. But, to reduce copying vhd multiple times to remote host, I made changes to copy only once in platform `_initialize`.
> Considering, if it needs to support different images in the same LISA run, how this value is loaded in initializing?

In general, isn't the testing image always constant in a single LISA run? Currently, there is no such use case to test different images in same LISA run.

squirrelsc: VHD cache can be controlled by a shared dict and locks. If a copied VHD exists, the vhd path can be used directly. If it's copying, other deployments can wait it by locks.
LISA run can use different images to run debug/release, normal VM/CVM together.

pupacha: Thanks, Chi, for suggesting the approach. I'll make the changes. 
> LISA run can use different images to run debug/release, normal VM/CVM together.

Could you please point me to / provide an example of such way of running LISA?

squirrelsc: The vhd_img can be a variable, so it can work with combinator to test with more img. In the internal pipelines, there are examples, which uses the combinators.

pupacha: Oh ok. I was of the assumption that with combinators, every time a new platform object would be created. Thanks, I'll look into the examples.

squirrelsc: You are right, the combinator create a new lisa_runner and platform each time, so each platform is using the same img actually. But it doesn't need to rely on this assumption. We already find some limitations to use single image, in future, there may be hard requirements to use different images to deploy. In this PR, we can avoid setting img to platform.

pupacha: Made changes as suggested. @squirrelsc @cwize1 @anirudhrb  Please review. Thanks.

cwize1 (approved):

squirrelsc (approved):

PR #2505, opened 2023-01-06
Fix DNS test for some SUSE-based images

lubaihua33:
1. Modify _no_permission_pattern of ping.py to suit more scenarios
2. Add sudo=True when running the command of "pidof ping" for SUSE-based images


LiliDeng:
@lubaihua33 please rebase main branch, thanks.

LiliDeng (approved):

PR #2506, opened 2023-01-06
Add dom0_binaries installer to install kernel using image files

smit-gardhariya:
This PR will add new installer type of **dom0_binaries** which will allow kernel_installer transformer to install kernel using image files. Here kernel image is mandatory to pass and initrd image is optional. linuxloader.conf will get change according to the image getting passed.

dom0 related installer will be placed under new file.

### DIFF
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+import os
 from dataclasses import dataclass, field
 from pathlib import PurePath
 from typing import Any, List, Optional, Type, cast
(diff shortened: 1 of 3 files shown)


lisa/transformers/kernel_source_installer.py
```diff
@@ -91,6 +92,91 @@ class SourceInstallerSchema(BaseInstallerSchema):
...
+    def install(self) -> str:
+        node = self._node
+        runbook: BinaryInstallerSchema = self.runbook
+        kernel_image_path: str = runbook.kernel_image_path
+        initrd_image_path: str = runbook.initrd_image_path
+        is_initrd: bool = False
+
+        uname = node.tools[Uname]
+        current_kernel = uname.get_linux_information().kernel_version_raw
+        new_kernel = os.path.basename(kernel_image_path).split("-")[1].strip()
```
LiliDeng: give an example for kernel_image_path in comments

squirrelsc:
@adityagesh Can you help check if there is any regression on guest upstream pipelines?

smit-gardhariya:
> @adityagesh Can you help check if there is any regression on guest upstream pipelines?

@SRIKKANTH

squirrelsc (approved):

PR #2507, opened 2023-01-06
Pylint: Enable redefined-builtin

avylove:
Enables redefined-builtin check. Python allows redefining built-in methods and classes. This gives a lot of flexibility to the language, but can have unexpected consequences and introduce bugs when done unintentionally. This check fails when a built-in function or class is redefined, protecting against doing this unintentionally.

In some cases the variables could be removed
In others, I attempted to rename the variable
When a new name wasn't clear, I used the standard convention of adding a `_` to the end of the name.
VSCode rename symbol functionality was used to handle renaming and then manually reviewed.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -25,7 +25,7 @@ from tools import update_file, update_summary  # type: ignore # noqa: E402
 # -- Project information -----------------------------------------------------
 
 project = "Linux Integration Services Automation (LISA)"
-copyright = "Microsoft Corporation"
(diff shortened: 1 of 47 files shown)


squirrelsc:
@LiliDeng LGTM. Let's merge it next Monday.

LiliDeng:
> @LiliDeng LGTM. Let's merge it next Monday.

Ack.

LiliDeng (approved):

PR #2508, opened 2023-01-07
Fix timeout and sudo issues in copy var/log/azure

dsrivastavv:


lisa/tools/chmod.py
```diff
@@ -12,5 +12,7 @@ def command(self) -> str:
     def chmod(self, path: str, permission: str, sudo: bool = False) -> None:
         self.run(f"{permission} {path}", sudo=sudo, force_run=True)
 
-    def update_folder(self, path: str, permission: str, sudo: bool = False) -> None:
-        self.run(f"-R {permission} {path}", sudo=sudo, force_run=True)
+    def update_folder(
+        self, path: str, permission: str, sudo: bool = False, timeout: int = 600
```
squirrelsc: set default to 10. It doesn't make sense to call this method for 600 seconds by default. And because the default value is 10, it doesn't need to use it in above called.

dsrivastavv: 10 might be too less for large folders since this methods updates permissions recursively, How about 1 minute default?

squirrelsc: This argument is added in this change list, the used place set it to 10 seconds. We should use 10 as default value here, unless we found it's not enough. To reduce waiting time, when it hangs, we should try small value firstly.

squirrelsc (dismissed):

lisa/operating_system.py
```diff
@@ -385,14 +385,27 @@ def capture_system_information(self, saved_path: Path) -> None:
...
+        try:
+            file_list = find_tool.find_files(
+                self._node.get_pure_path("/var/log/azure/"),
+                file_type="f",
+                sudo=True,
+                ignore_not_exist=True,
+            )
+        except Exception:
+            # On systems like docker sudo is not present, so try without sudo
+            file_list = find_tool.find_files(
```
squirrelsc: I thought it again. we shouldn't handle it one by one for docker in the node.py

1. Change the logic of support_sudo to check if sudo supported for docker.
2. Update execute_async, not to raise exception, if support_sudo is not here.

dsrivastavv: Nice! This would be much cleaner

squirrelsc (approved):

PR #2509, opened 2023-01-07
[improvement] support resource_group_location in runbook

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -465,14 +467,20 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
         else:
             try:
                 if self._azure_runbook.deploy:
+                    if self._azure_runbook.resource_group_location:
```
squirrelsc: The env resource group can be the same as first node.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -246,6 +246,8 @@ class AzurePlatformSchema:
 
     shared_resource_group_name: str = AZURE_SHARED_RG_NAME
     resource_group_name: str = field(default="")
+    # specify resource group location
+    resource_group_location: Optional[str] = field(default=None)
```
squirrelsc: add one for shared_resource_group_location, and set its default value as `shared_resource_group_name`

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -464,6 +466,10 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
             log.info(f"dry_run: {self._azure_runbook.dry_run}")
         else:
             try:
+                location, deployment_parameters = self._create_deployment_parameters(
```
squirrelsc, line 469: Check if it affects when "deploy:false".

LiliDeng: I didn't see it affects when use existing rg.

squirrelsc (approved):

PR #2511, opened 2023-01-07
[improvement] exclude expired arm64 vm size

LiliDeng:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -2038,6 +2038,9 @@ def create_setting(
         cls, *args: Any, **kwargs: Any
     ) -> Optional[schema.FeatureSettings]:
         raw_capabilities: Any = kwargs.get("raw_capabilities")
+        resource_sku: Any = kwargs.get("resource_sku")
```
squirrelsc, line 2041: It's not a good practice, we need to add a flag in database to exclude vm sizes.

PR #2513, opened 2023-01-09
[improvement] allow to use users' resource group which not create from lisa

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1377,125 +1377,98 @@ def _parse_detail_errors(self, error: Any) -> List[str]:
 
     # the VM may not be queried after deployed. use retry to mitigate it.
     @retry(exceptions=LisaException, tries=150, delay=2)
-    def _load_vms(
-        self, environment: Environment, log: Logger
-    ) -> Dict[str, VirtualMachine]:
+    def load_vms_info(
```
squirrelsc: This change makes this method is called multiple times during initialization, it impacts the performance. So it's better to have a separated method to get public ip addresses and just share logics.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1377,125 +1377,98 @@ def _parse_detail_errors(self, error: Any) -> List[str]:
...
-        # load nics
         nics_map: Dict[str, NetworkInterface] = {}
-        network_interfaces = network_client.network_interfaces.list(
-            environment_context.resource_group_name
-        )
-        for nic in network_interfaces:
-            # nic name is like lisa-test-20220316-182126-985-e0-n0-nic-2, get vm
-            # name part for later pick only find primary nic, which is ended by
-            # -nic-0
-            node_name_from_nic = RESOURCE_ID_NIC_PATTERN.findall(nic.name)
```
squirrelsc, line 1428: Remove the patterns, if they are not used now.

LiliDeng: removed

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: updated, please review, thanks!

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1158,6 +1168,42 @@ def get_vm(platform: "AzurePlatform", node: Node) -> Any:
     return vm
 
 
+def get_nic_name(nic_resource_id: str) -> str:
+    nic_name_pattern = re.compile(r"Microsoft.Network/networkInterfaces/(.*)", re.M)
```
squirrelsc: always move re.compile out of method, so it doesn't need to be compiled every time, when calling the method.

LiliDeng:
@avylove can we use a fixed version of flake8-bugbear, and update the version regularly? the failures shown in my PR were caused by new warnings B028 added in latest version of flake8-bugbear. 

I have a quick fix here #2525

lisa/sut_orchestrator/azure/common.py

squirrelsc (approved):

avylove:
> @avylove can we use a fixed version of flake8-bugbear, and update the version regularly? the failures shown in my PR were caused by new warnings B028 added in latest version of flake8-bugbear.
> 
> I have a quick fix here #2525

https://github.com/microsoft/lisa/pull/2533

PR #2514, opened 2023-01-09
Pylint: Enable eval-used

avylove:
`eval()` allows arbitrary code execution and should be avoided. This adds a Pylint check and removed cases where is was used. Most of those were simple changes, but `ScriptTransformer` required restricting what user input can do. simpleeval was used as it's close to the implementation of a mini-language without effort of creating one.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -553,8 +553,7 @@ class Infiniband(AzureFeatureMixin, features.Infiniband):
     ) -> Optional[schema.FeatureSettings]:
         raw_capabilities: Any = kwargs.get("raw_capabilities")
 
-        value = raw_capabilities.get("RdmaEnabled", None)
-        if value and eval(value) is True:
+        if raw_capabilities.get("RdmaEnabled", None) == "True":
             return schema.FeatureSettings.create(cls.name())
 
         return None
@@ -1495,8 +1494,7 @@ class Hibernation(AzureFeatureMixin, features.Hibernation):
     ) -> Optional[schema.FeatureSettings]:
         raw_capabilities: Any = kwargs.get("raw_capabilities")
 
-        value = raw_capabilities.get("HibernationSupported", None)
-        if value and eval(value) is True:
+        if raw_capabilities.get("HibernationSupported", None) == "True":
             return schema.FeatureSettings.create(cls.name())
 
         return None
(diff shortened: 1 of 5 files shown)


pyproject.toml
```diff
@@ -23,6 +23,7 @@ dependencies = [
     "randmac ~= 0.1",
     "retry ~= 0.9.2",
     "semver ~= 2.13.0",
+    "simpleeval >= 0.9.12",
```
squirrelsc: pin it to 0.9.12

lisa/sut_orchestrator/azure/features.py
```diff
@@ -548,8 +548,7 @@ def create_setting(
     ) -> Optional[schema.FeatureSettings]:
         raw_capabilities: Any = kwargs.get("raw_capabilities")
 
-        value = raw_capabilities.get("RdmaEnabled", None)
-        if value and eval(value) is True:
+        if raw_capabilities.get("RdmaEnabled", None) == "True":
```
squirrelsc, line 556: I don't remember its value, may be "true"? Please have a run to make sure there is no regression.

avylove: The only value it could be with the existing code is `'True'`. `'true'` would raise a `NameError`. Typically this would all run through a data validation library, but that's out of scope for this PR.

LiliDeng: it is False or True

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1512,8 +1510,11 @@ def create_setting(
             "standardMSv2Family",
         ]:
             # https://learn.microsoft.com/en-us/azure/virtual-machines/trusted-launch#how-can-i-find-vm-sizes-that-support-trusted-launch # noqa: E501
-            tvm_disable_value = raw_capabilities.get("TrustedLaunchDisabled", "False")
-            if gen_value and ("V2" in str(gen_value)) and (not eval(tvm_disable_value)):
+            if (
+                gen_value
+                and ("V2" in str(gen_value))
+                and raw_capabilities.get("TrustedLaunchDisabled", "False") == "True"
```
LiliDeng: is it `raw_capabilities.get("TrustedLaunchDisabled", "False") == "False"`?

avylove: Yes, good catch!

pyproject.toml
avylove: Will do compatible pin to get this in, but we'll need to prioritize pin management

avylove:
@squirrelsc , @LiliDeng Any issues with this?

squirrelsc:
> @squirrelsc , @LiliDeng Any issues with this?

It touches the core part, we don't have a chance to verify it.

avylove:
Runbook I used to for testing., just changed `timeout_value` and `script` to test different cases.
`lisa -r test.yml --debug 2>&1 | grep lisa.transformer`

```yaml
name: local test
notifier:
  - type: html
environment:
  environments:
    - nodes:
        - type: local
platform:
  - type: ready

variable:
  - name: timeout_value
    value: 0

transformer:
  - type: script
    rename:
      script_should_run: should_run
    scripts:
      - name: should_run
        variables:
          - timeout_value
        script: int(timeout_value) <= 0
```

squirrelsc:
It needs to be tested with 1) gen1/gen2, 2) hibernate or rdma. Please let Lili know, if you don't know which test is suitable for this testing.

LiliDeng (approved):
tested TVM (gen2), sriov and RDMA, works well.

LiliDeng (changes requested):
will double check once back from holiday,  thanks!

squirrelsc (dismissed):

LiliDeng (approved):

PR #2515, opened 2023-01-10
Add test case to validate PMU is disbaled for ARM64

sharsonia:


microsoft/testsuites/core/timesync.py
```diff
@@ -351,3 +351,24 @@ def timesync_chrony(self, node: Node) -> None:
         chrony.check_sources_and_stats()
         # 5. Check chrony tracking.
         chrony.check_tracking()
+
+    @TestCaseMetadata(
+        description="""
+        This test is to check "initcall_blacklist=arm_pmu_acpi_init"
+        kernel parameter for ARM64 images.
+        """,
```
LiliDeng, line 364: can you put more context here, like why need blacklist this, is it a temp fix?

microsoft/testsuites/core/timesync.py
```diff
@@ -351,3 +351,24 @@ def timesync_chrony(self, node: Node) -> None:
...
+        lscpu = node.tools[Lscpu]
+        arch = lscpu.get_architecture()
+        if arch is not CpuArchitecture.ARM64:
+            raise SkippedException(
+                f"This test case does not support {arch}."
+                "This validation is only for ARM64."
+            )
+        # Check pmu is disabled in cmdline for arm64 images
+        cat = node.tools[Cat]
+        result = cat.run("/proc/cmdline", force_run=True)
```
LiliDeng, line 377: please double check whether all distro put blacklist in this file

sharsonia: /proc/cmdline will have boot params which is what we are verifying that the boot params have this string - "initcall_blacklist=arm_pmu_acpi_init"

/proc/cmdline exists in Ubuntu, Redhat (and CentOS). Let me verify for Alma and other such

sharsonia: > please double check whether all distro put blacklist in this file

@LiliDeng

Results from various runs
https://microsoft.visualstudio.com/LSG/_build/results?buildId=63133136&view=results
https://microsoft.visualstudio.com/LSG/_build/results?buildId=63114483&view=results
https://microsoft.visualstudio.com/LSG/_build/results?buildId=63146846&view=results

LiliDeng (approved):

PR #2516, opened 2023-01-10
Pylint: Enable modified-iterating-list

avylove:
Modifying an iterator while iterating over it can have unexpected behavior.
Enables modified-iterating-list check and fixes code.


avylove:
@squirrelsc , @LiliDeng Any issues with this?

squirrelsc:
> @squirrelsc , @LiliDeng Any issues with this?


It touches the core part, we don't have a chance to verify it.

avylove:
Rebased. This is now 3.5 months old.

squirrelsc:
Please either test with impacted scenarios or suppress this code to enable the rule. This place is safe, because the iterator is broken out after modified.

PR #2517, opened 2023-01-11
[fix issue] not run collect azure log when local machine is not test machine

LiliDeng:
I check only when node type is local and self._node.name is local, the test machine is the local machine.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -41,6 +41,7 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
         runbook: schema.Node,
         index: int,
         logger_name: str,
+        is_test_target: bool = True,
         base_part_path: Optional[Path] = None,
         parent_logger: Optional[Logger] = None,
     ) -> None:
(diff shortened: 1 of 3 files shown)


lisa/operating_system.py
```diff
@@ -381,28 +381,32 @@ def capture_system_information(self, saved_path: Path) -> None:
...
-            self._node.tools[Chmod].update_folder("/var/log/azure/", "a+rwX", sudo=True)
-        file_list.append("/etc/os-release")
-        file_list.append("/var/log/waagent.log")
-        for file in file_list:
-            try:
-                file_name = file.split("/")[-1]
-                self._node.shell.copy_back(
-                    self._node.get_pure_path(file),
-                    saved_path / f"{file_name}.txt",
+        if not (self._node.type_name() == "local" and self._node.name == "local"):
```
squirrelsc: we can use a flag, instead of by name.

squirrelsc: how about is_test_target? It can be a flag on node level, the remote node may have similar situation.

lisa/node.py
```diff
@@ -48,6 +48,7 @@ def __init__(
         self.is_default = runbook.is_default
         self.capability = runbook.capability
         self.name = runbook.name
+        self.is_test_target = True
```
squirrelsc: The default value should be False, and it can be a parameter of __init__ method. It's set to trye, when calling from local() and quick_connect().

squirrelsc (approved):

PR #2520, opened 2023-01-12
Add git functions to retrieve repo info

adityagesh:
Add git functions to get details such as current branch, repo url, details of latest commit

### DIFF
--- a/lisa/tools/git.py
+++ b/lisa/tools/git.py
@@ -3,7 +3,7 @@
 
 import pathlib
 import re
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 from assertpy import assert_that
 from semver import VersionInfo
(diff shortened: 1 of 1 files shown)


lisa/tools/git.py
```diff
@@ -276,3 +276,62 @@ def get_version(self) -> VersionInfo:
...
+        author_email = self.run(
+            "--no-pager log -n 1 --format='%ae'",
+            shell=True,
+            cwd=cwd,
+            force_run=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Failed to fetch author email.",
+        ).stdout
+
+        result = {
```
LiliDeng, line 331: is there a way to get latest_commit_id, commit_message_name and author_email in one command?

adityagesh: It is possible using a single command `git --no-pager log -n 1 --pretty=format:%H-%B-%ae`
However, we will have to split the output using the separating term '-' in the above example. If the data itself contains the separator (eg: commit message contains '-'), we might get the incorrect split values.

LiliDeng (approved):

PR #2521, opened 2023-01-12
fix: VM Resize with StopStart

adityagesh:
VM Resize testcase with StopStart leaves VM in Stop state if Resize raises an exception.
This PR is to Start VM before raising an exception

### DIFF
--- a/microsoft/testsuites/core/vm_resize.py
+++ b/microsoft/testsuites/core/vm_resize.py
@@ -154,9 +154,10 @@ class VmResize(TestSuite):
                 else:
                     raise identifier
                 time.sleep(1)
+            finally:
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/vm_resize.py
```diff
@@ -152,6 +154,8 @@ def _verify_vm_resize(
                 ):
                     retry = retry + 1
                 else:
+                    if not hot_resize:
+                        start_stop.start()
```
LiliDeng: put it in finally block?

adityagesh: done

LiliDeng (approved):

PR #2522, opened 2023-01-13
Bug fixes for libvirt/QEMU orchestrator.

cwize1:
1. Don't use `dmesg` if node is local machine, as some OSes (e.g. Fedora 37) will block its usage.

2. Fix OVMF firmware config search filtering to handle new versions of the OVMF firmware files.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -195,10 +195,11 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         if self.platform_runbook.capture_libvirt_debug_logs:
             self._disable_libvirt_debug_log()
 
-        dmesg_output = self.host_node.tools[Dmesg].get_output(force_run=True)
-        dmesg_path = self.host_node.local_log_path / "dmesg.txt"
-        with open(str(dmesg_path), "w") as f:
-            f.write(dmesg_output)
+        if self.host_node.is_remote:
+            dmesg_output = self.host_node.tools[Dmesg].get_output(force_run=True)
+            dmesg_path = self.host_node.local_log_path / "dmesg.txt"
+            with open(str(dmesg_path), "w") as f:
+                f.write(dmesg_output)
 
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -383,15 +383,20 @@ def capture_system_information(self, saved_path: Path) -> None:
...
-        find_tool = self._node.tools[Find]
-        file_list = find_tool.find_files(
-            self._node.get_pure_path("/var/log/azure/"),
-            file_type="f",
-            sudo=True,
-            ignore_not_exist=True,
-        )
-        if len(file_list) > 0:
-            self._node.tools[Chmod].update_folder("/var/log/azure/", "a+rwX", sudo=True)
+        file_list = []
```
squirrelsc: @LiliDeng is working on this, and mostly done.

https://github.com/microsoft/lisa/pull/2517

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -195,10 +195,11 @@ def _cleanup(self) -> None:
         if self.platform_runbook.capture_libvirt_debug_logs:
             self._disable_libvirt_debug_log()
 
-        dmesg_output = self.host_node.tools[Dmesg].get_output(force_run=True)
-        dmesg_path = self.host_node.local_log_path / "dmesg.txt"
-        with open(str(dmesg_path), "w") as f:
-            f.write(dmesg_output)
+        if self.host_node.is_remote:
```
squirrelsc, line 198: With Lili's change, we introduced a new flat called "is_test_target". Please consider if it's applicable here.

cwize1: `host_node` will never be the test target. So, I think `is_remote` is correct here.

anirudhrb (dismissed):

lisa/operating_system.py
squirrelsc: #2517 was merged. please rebase to the latest code.

cwize1:
@squirrelsc Friendly ping.

squirrelsc (approved):

PR #2524, opened 2023-01-16
Add handling for the blob copy pending scenario

lubaihua33:
There is a scenario that two processes are copying the same VHD. It needs to check the copy status if the blob exists. If the status is success, then the copied vhd can be directly used, if not, then need to wait for the copy to complete.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -10,7 +10,7 @@ import re
 import sys
 from copy import deepcopy
 from dataclasses import dataclass, field
-from datetime import datetime
+from datetime import datetime, timezone
 from difflib import SequenceMatcher
 from functools import lru_cache, partial
 from pathlib import Path
@@ -189,6 +189,10 @@ STORAGE_CONTAINER_BLOB_PATTERN = re.compile(
     r"/(?P<container>[^/]+)/?/(?P<blob>.*)",
     re.M,
 )
+
+# The timeout hours of the blob with copy pending status
+# If the blob is still copy pending status after the timeout hours, it can be deleted
+BLOB_COPY_PENDING_TIMEOUT_HOURS = 6
 _global_sas_vhd_copy_lock = Lock()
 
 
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -189,6 +189,10 @@
     r"/(?P<container>[^/]+)/?/(?P<blob>.*)",
     re.M,
 )
+
+# The timeout hours of the blob with copy pending status
+# If the blob is still copy pengding status after the timeout hours, it can be deleted
```
LiliDeng: typo `pengding`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1989,13 +1994,21 @@ def _get_deployable_vhd_path(
                     if blob.content_settings:
                         cached_key = blob.content_settings.get("content_md5", None)
                     if original_key == cached_key:
-                        # if it exists, return the link, not to copy again.
-                        log.debug("the sas url is copied already, use it directly.")
-                        return full_vhd_path
+                        # If it exists then check if the blob is available by checking
```
squirrelsc: It looks the logic just need to adjust, it doesn't need to have a new method. The key thing is even the key is the same, it needs to wait, until the blob is success. If the default timeout 1 hour is too short, we can extend it to 2 hours. But 6 hours to wait is too long.

If the vhd is pending more than 6 hours, it should be deleted, and copy again.

```python
            blobs = container_client.list_blobs(name_starts_with=vhd_path)
            vhd_exists = False
            for blob in blobs:
                if blob:
                    # check if hash key matched with original key.
                    if blob.content_settings:
                        cached_key = blob.content_settings.get("content_md5", None)
                    if _is_stuck_copying():
                        # Delete the stuck vhd.
                        blob_client.delete_blob(delete_snapshots="include")
                        break
                    elif original_key == cached_key:
                        # If it exists, return the link, not to copy again.
                        log.debug("the sas url is copied already, use it directly.")
                        vhd_exists = True
                    else:
                        log.debug("found cached vhd, but the hash key mismatched.")

            if not vhd_exists:
                blob_client = container_client.get_blob_client(vhd_path)
                blob_client.start_copy_from_url(
                    original_vhd_path, metadata=None, incremental_copy=False
                )

            wait_copy_blob(blob_client, vhd_path, log)
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1983,27 +1987,52 @@ def _get_deployable_vhd_path(
...
                 if blob:
                     # check if hash key matched with original key.
                     if blob.content_settings:
                         cached_key = blob.content_settings.get("content_md5", None)
-                    if original_key == cached_key:
-                        # if it exists, return the link, not to copy again.
+                    if self._is_stuck_copying(blob_client, log):
+                        # Delete the stuck vhd.
+                        blob_client.delete_blob(delete_snapshots="include")
+                        break
```
squirrelsc: The blob should have only one, so it doesn't need to break. The logic will be consistent.

squirrelsc (approved):

PR #2528, opened 2023-01-17
Get details from kernel source installer

adityagesh:
Add Interface get_details to export details from
kernel installers.
Expose these details as output of this transformer

### DIFF
--- a/lisa/transformers/kernel_installer.py
+++ b/lisa/transformers/kernel_installer.py
@@ -87,19 +87,20 @@ class BaseInstaller(subclasses.BaseClassWithRunbookMixin):
         self._node = node
         self._log = get_logger("kernel_installer", parent=parent_log)
 
+    @property
+    def information(self) -> Dict[str, Any]:
+        return dict()
+
     def validate(self) -> None:
         raise NotImplementedError()
 
     def install(self) -> str:
         raise NotImplementedError()
 
-    def get_details(self) -> Dict[str, Any]:
-        return dict()
-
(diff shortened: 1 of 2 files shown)


lisa/transformers/kernel_installer.py
```diff
@@ -93,8 +93,14 @@ def validate(self) -> None:
     def install(self) -> str:
         raise NotImplementedError()
 
+    def get_details(self) -> Dict[str, Any]:
```
squirrelsc: use a read-only property called `information`. the "details" is not fit here, because the installer doesn't provide "brief". So the information field can be used to provide all information.

lisa/transformers/kernel_source_installer.py
```diff
@@ -108,6 +113,9 @@ def validate(self) -> None:
         # nothing to validate before source installer started.
         ...
 
+    # def _internal_run(self) -> Dict[str, Any]:
```
squirrelsc: remove the commented parts.

lisa/transformers/kernel_source_installer.py
```diff
@@ -395,6 +413,26 @@ def get_source_code(self) -> PurePath:
 
         return code_path
 
+    def get_details(self) -> Dict[str, Any]:
```
squirrelsc: It looks this information is not only for repo location. It should be moved up to source installer.

squirrelsc (approved):

PR #2530, opened 2023-01-17
[fix issue] check rg created successfully before deployment

LiliDeng:
Fix below issue
`
  File "D:\a\_work\1\s\lisa\lisa\runners\lisa_runner.py", line 276, in _deploy_environment_task
    self.platform.deploy_environment(environment)
  File "D:\a\_work\1\s\lisa\lisa\platform_.py", line 167, in deploy_environment
    self._deploy_environment(environment, log)
  File "D:\a\_work\1\s\lisa\lisa\sut_orchestrator\azure\platform_.py", line 492, in _deploy_environment
    raise identifier
  File "D:\a\_work\1\s\lisa\lisa\sut_orchestrator\azure\platform_.py", line 485, in _deploy_environment
    self._validate_template(deployment_parameters, log)
  File "D:\a\_work\1\s\lisa\lisa\sut_orchestrator\azure\platform_.py", line 1284, in _validate_template
lisa.util.LisaException: ResourceGroupNotFound: Resource group 'lisa-Azure-Adhoc-20230117-024815-543-e145' could not be found.
`

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -58,6 +58,7 @@ from lisa.secret import PATTERN_HEADTAIL, PATTERN_URL, add_secret
 from lisa.util import (
     LisaException,
     LisaTimeoutException,
+    check_till_timeout,
     constants,
     field_metadata,
     get_matched_str,
@@ -1033,6 +1034,11 @@ def check_or_create_resource_group(
                 rm_client.resource_groups.create_or_update(
                     resource_group_name, {"location": location}
                 )
+            check_till_timeout(
+                lambda: rm_client.resource_groups.check_existence(resource_group_name)
+                is True,
+                timeout_message=f"wait for {resource_group_name} created",
+            )
 
 
 def wait_copy_blob(
@@ -1042,18 +1048,12 @@ def wait_copy_blob(
     timeout: int = 60 * 60,
 ) -> None:
     log.info(f"copying vhd: {vhd_path}")
-
-    timeout_timer = create_timer()
-    while timeout_timer.elapsed(False) < timeout:
-        props = blob_client.get_blob_properties()
-        if props.copy.status == "success":
-            break
-        # the copy is very slow, it may need several minutes. check it every
-        # 2 seconds.
-        sleep(2)
-    if timeout_timer.elapsed() >= timeout:
-        raise LisaException(f"wait copying VHD timeout: {vhd_path}")
-
+    check_till_timeout(
+        lambda: blob_client.get_blob_properties().copy.status == "success",
+        timeout_message=f"copying VHD: {vhd_path}",
+        timeout=timeout,
+        interval=2,
+    )
     log.info("vhd copied")
 
 
--- a/lisa/util/__init__.py
+++ b/lisa/util/__init__.py
@@ -7,6 +7,7 @@ import string
 import sys
 from datetime import datetime
 from pathlib import Path
+from time import sleep
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -29,6 +30,7 @@ from semver import VersionInfo
 
 from lisa import secret
 from lisa.util import constants
+from lisa.util.perf_timer import create_timer
 
 if TYPE_CHECKING:
     from lisa.operating_system import OperatingSystem
@@ -657,3 +659,18 @@ def strip_strs(obj: Any, fields: List[str]) -> Any:
             value = value.strip() if isinstance(value, str) else value
             setattr(obj, field, value)
     return obj
+
+
+def check_till_timeout(
+    func: Callable[..., Any],
+    timeout_message: str,
+    timeout: int = 60,
+    interval: int = 1,
+) -> None:
+    timer = create_timer()
+    while timer.elapsed(False) < timeout:
+        if func():
+            break
+        sleep(interval)
+    if timer.elapsed() >= timeout:
+        raise LisaException(f"timeout: {timeout_message}")


lisa/sut_orchestrator/azure/common.py
```diff
@@ -972,6 +972,16 @@ def check_or_create_resource_group(
                 rm_client.resource_groups.create_or_update(
                     resource_group_name, {"location": location}
                 )
+            timeout = 60
+            timeout_timer = create_timer()
+            while timeout_timer.elapsed(False) < timeout:
+                if rm_client.resource_groups.check_existence(resource_group_name):
+                    break
+                sleep(2)
```
squirrelsc: It's second time to use the timeout checking pattern, please create a method in utils/__init__.py.

lisa/util/__init__.py
```diff
@@ -657,3 +659,47 @@ def strip_strs(obj: Any, fields: List[str]) -> Any:
             value = value.strip() if isinstance(value, str) else value
             setattr(obj, field, value)
     return obj
+
+
+def check_get_attr(actual_object: Any, attr_name: str = "") -> Any:
```
squirrelsc: `getattr_with_check`

lisa/util/__init__.py
```diff
@@ -657,3 +659,47 @@ def strip_strs(obj: Any, fields: List[str]) -> Any:
...
+            current = actual_value
+    else:
+        actual_value = func(**test_kwargs)
+    if actual_value == expected:
+        return True
+    else:
+        return False
+
+
+def retry_till_timeout(func: Any, *func_args: Any, **kwargs: Any) -> None:
```
squirrelsc: Try this example. if the logic is complicated, which is not able to be handled by lambda in one line, use a function.

```python
from functools import partial
from time import sleep
from typing import Callable

from lisa import LisaException
from lisa.util.logger import get_logger
from lisa.util.perf_timer import create_timer

log = get_logger()


def check_till_timeout(
    func: Callable[..., bool], timeout_message: str, timeout: int = 20
) -> None:
    timer = create_timer()
    while timer.elapsed(False) < timeout:
        if func():
            break
        sleep(1)
    if timer.elapsed() >= timeout:
        raise LisaException(f"timeout: {timeout_message}")


def method_with_arguments(wait_for: int = 5) -> bool:
    log.info("method_with_arguments")
    if global_wait_timer.elapsed(False) < wait_for:
        return False
    else:
        return True


global_wait_timer = create_timer()
log.info("calling check_till_timeout example 1...")
check_till_timeout(
    partial(method_with_arguments, wait_for=2),
    timeout_message="check_till_timeout example 1 is timeout",
)

global_wait_timer = create_timer()
log.info("calling check_till_timeout example 2...")
try:
    check_till_timeout(
        partial(method_with_arguments),
        timeout_message="check_till_timeout example 2 is timeout",
        timeout=2,
    )
except LisaException as identifier:
    log.info(identifier)

global_wait_timer = create_timer()
log.info("calling lambda 1...")
check_till_timeout(
    lambda: False if global_wait_timer.elapsed(False) < 2 else True,
    timeout_message="lambda example is timeout",
)

global_wait_timer = create_timer()
log.info("calling lambda 2...")

try:
    check_till_timeout(
        lambda: False if global_wait_timer.elapsed(False) < 5 else True,
        timeout_message="lambda example is timeout",
        timeout=2,
    )
except LisaException as identifier:
    log.info(identifier)

```

lisa/util/__init__.py
```diff
@@ -657,3 +659,47 @@ def strip_strs(obj: Any, fields: List[str]) -> Any:
...
     return obj
+
+
+def check_get_attr(actual_object: Any, attr_name: str = "") -> Any:
+    if hasattr(actual_object, attr_name):
+        return getattr(actual_object, attr_name)
+    raise LisaException(f"attr {attr_name} not existing in object {actual_object}")
+
+
+def is_value_expected(
```
squirrelsc: This method looks over engineered, it doesn't need.

lisa/util/__init__.py
```diff
@@ -657,3 +659,38 @@ def strip_strs(obj: Any, fields: List[str]) -> Any:
...
+        raise LisaException(f"timeout: {timeout_message}")
+
+
+def getattr_with_check(actual_object: Any, attr_name: str = "") -> Any:
+    if hasattr(actual_object, attr_name):
+        return getattr(actual_object, attr_name)
+    raise LisaException(f"attr {attr_name} not existing in object {actual_object}")
+
+
+def method_with_arguments(
```
squirrelsc: method_with_arguments just an example, it doesn't need. Try to use the logic by lambda directly. If the logic is multiple lines, wrapper it to a function.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1042,18 +1050,17 @@ def wait_copy_blob(
...
-            break
-        # the copy is very slow, it may need several minutes. check it every
-        # 2 seconds.
-        sleep(2)
-    if timeout_timer.elapsed() >= timeout:
-        raise LisaException(f"wait copying VHD timeout: {vhd_path}")
-
+    check_till_timeout(
+        partial(
+            method_with_arguments,
```
squirrelsc: Try this.

```python
check_till_timeout(lambda: blob_client.get_blob_properties().copy.status == "success", timeout_message...
```

LiliDeng: thanks, this works.

lisa/util/__init__.py
```diff
@@ -657,3 +659,18 @@ def strip_strs(obj: Any, fields: List[str]) -> Any:
             value = value.strip() if isinstance(value, str) else value
             setattr(obj, field, value)
     return obj
+
+
+def check_till_timeout(
+    func: Callable[..., Any],
+    timeout_message: str,
+    timeout: int = 20,
+    waittime: int = 1,
```
squirrelsc: usually, it's called `interval`

lisa/util/__init__.py
```diff
@@ -657,3 +659,18 @@ def strip_strs(obj: Any, fields: List[str]) -> Any:
             value = value.strip() if isinstance(value, str) else value
             setattr(obj, field, value)
     return obj
+
+
+def check_till_timeout(
+    func: Callable[..., Any],
+    timeout_message: str,
+    timeout: int = 20,
```
squirrelsc: Give 60 as the default value, so at least one call doesn't need to specify it.

lisa/util/__init__.py

lisa/util/__init__.py

squirrelsc (approved):

PR #2532, opened 2023-01-18
Add some fix of error 'cannot find [kernelconfig]' for validate_netvsc_reload case

lubaihua33:
1. Add netvsc build-in check for FreeBSD and other special image
2. Modify the kernel config file path for CoreOs

### DIFF
--- a/microsoft/testsuites/network/netinterface.py
+++ b/microsoft/testsuites/network/netinterface.py
@@ -18,7 +18,8 @@ from lisa import (
 )
 from lisa.features import NetworkInterface, Synthetic
 from lisa.nic import Nics
-from lisa.tools import Dhclient, Ip, KernelConfig, Wget
+from lisa.operating_system import FreeBSD
+from lisa.tools import Dhclient, Ip, KernelConfig, Uname, Wget
 from lisa.util import perf_timer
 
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/network/netinterface.py
```diff
@@ -188,5 +189,30 @@ def validate_set_static_mac(self, node: Node, log: Logger) -> None:
...
+                )
+            except LisaException as identifier:
+                # Some image's kernel config is inconsistent with the kernel version.
+                # E.g. fatpipe-inc fatpipe-wanopt 10 0.0.3, then it has the exception.
+                # If so, check if netvsc is built-in using command "modinfo hv_netvsc".
+                node.log.debug(identifier)
+                is_built_in_module = (
+                    node.execute("modinfo hv_netvsc", sudo=True, shell=True).exit_code
+                    != 0
+                )
```
LiliDeng, line 222: I recall if the module is built in, modinfo also can show output, could you please double check?

lubaihua33: Changed to check '/lib/modules/$(uname -r)/modules.builtin'

LiliDeng (approved):

PR #2535, opened 2023-01-19
Modify the crashkernel command for ARM64 CentOS image

lubaihua33:


lisa/tools/kdump.py
```diff
@@ -427,14 +428,19 @@ def _get_crashkernel_update_cmd(self, crashkernel: str) -> str:
                 f' --args="crashkernel={crashkernel}"'
             )
         else:
-            if self.node.shell.exists(PurePosixPath("/sys/firmware/efi")):
+            lscpu = self.node.tools[Lscpu]
```
LiliDeng: use this instead of using tool lscpu
arch = self.node.os.get_kernel_information().hardware_platform  # type: ignore
arch == "x86_64":

squirrelsc: We need an enum for the arch strings.

lubaihua33: > use this instead of using tool lscpu arch = self.node.os.get_kernel_information().hardware_platform # type: ignore arch == "x86_64":

Done

LiliDeng (approved):

PR #2537, opened 2023-01-19
[fix issue] fix docker compose issue on arm64

LiliDeng:


squirrelsc (approved):

LiliDeng:
I still encounter issue with this fix, will raise new one, once case passed on arm64 distro

PR #2538, opened 2023-01-19
Pylint: Enable some import checks

avylove:
Enables consider-using-from-import and deprecated-module checks
consider-using-from-import check for things like import `lisa.util.constants as constants`
The deprecated-module check found one use of distutils which was deprecated in 3.10 and will be removed in 3.12. The only use is a simple function, so I added equivalent logic instead.

### DIFF
--- a/lisa/variable.py
+++ b/lisa/variable.py
@@ -4,7 +4,6 @@
 import os
 import re
 from dataclasses import dataclass
-from distutils.util import strtobool
 from typing import Any, Dict, List, Optional, Union
 
 import yaml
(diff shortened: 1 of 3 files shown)


lisa/variable.py
```diff
@@ -54,14 +53,13 @@ def _try_convert_type(original_value: Any, new_value: Any) -> Any:
...
-        target_type = original_type
-
+    target_type = new_type if new_type is not str else original_type
     try:
         if target_type is bool:
-            new_value = bool(strtobool(new_value))
+            if new_value.lower() in ("y", "yes", "t", "true", "on", "1"):
+                new_value = True
+            elif new_value.lower() in ("n", "no", "f", "false", "off", "0"):
+                new_value = False
```
squirrelsc, line 62: Others should raise ValueError.

avylove: If we were re-implementing the function independently, yes, but, in this case, an exception would just result in new_value's value not being changed.

squirrelsc: Other values should raise an error, so users know the value is incorrect. Otherwise, it's unexpected results.

avylove: They wouldn't know, the exception would be caught and discarded.

squirrelsc: ok, recalled the logic in mind.

squirrelsc:
@LiliDeng LGTM, please merge it in next Monday.

PR #2539, opened 2023-01-19
Pylint: enable missing-timeout

avylove:
Enables missing-timeout check to identify cases where requests is used without timeout values since this can hang waiting for a response.

I used 60 as the timeout, which is too much. I think 30 would be better, but wanted to be very conservative and see how others felt.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1134,7 +1134,9 @@ def save_console_log(
         )
     if saved_path:
         screenshot_raw_name = saved_path / f"{screenshot_file_name}.bmp"
-        screenshot_response = requests.get(diagnostic_data.console_screenshot_blob_uri)
+        screenshot_response = requests.get(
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1148,7 +1150,7 @@ def save_console_log(
             )
         screenshot_raw_name.unlink()
 
-    log_response = requests.get(diagnostic_data.serial_console_log_blob_uri)
+    log_response = requests.get(diagnostic_data.serial_console_log_blob_uri, timeout=60)
```
squirrelsc, line 1153: Is it the default value from the requests package?

avylove: The default is it will not timeout and hang until something else stops it. I picked 60 as a very conservative value, but open to other values.
https://requests.readthedocs.io/en/latest/user/advanced/#timeouts

squirrelsc (approved):

PR #2540, opened 2023-01-19
Test bpv2

sharsonia:


LiliDeng:
@sharsonia if it is only for testing purpose, no need to create PR.

sharsonia:
Tested

PR #2546, opened 2023-01-23
ch_tests: Support running specific subtests

pupacha:
* Add support to run specific set of subtests

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests.py
@@ -70,8 +70,19 @@ class CloudHypervisorTestSuite(TestSuite):
     ) -> None:
         hypervisor = self._get_hypervisor_param(node)
         ref = variables.get("cloudhypervisor_ref", "")
+        # below variable expects a comma separated list of full testnames
+        include_list, exclude_list = get_test_list(
+            variables, "ch_integration_tests_included", "ch_integration_tests_excluded"
+        )
         node.tools[CloudHypervisorTests].run_tests(
-            result, environment, "integration", hypervisor, log_path, ref
+            result,
+            environment,
+            "integration",
+            hypervisor,
+            log_path,
+            ref,
+            include_list,
+            exclude_list,
         )
 
     @TestCaseMetadata(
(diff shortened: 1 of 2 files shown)


pupacha:
@anirudhrb Please review

anirudhrb:
Please merge only after fixing the issue where tests in the exact list are run twice.

microsoft/testsuites/cloud_hypervisor/ch_tests.py
```diff
@@ -98,8 +110,20 @@ def verify_cloud_hypervisor_live_migration_tests(
     ) -> None:
         hypervisor = self._get_hypervisor_param(node)
         ref = variables.get("cloudhypervisor_ref", "")
+        # below variable expects a comma separated list of full testnames
```
squirrelsc, line 112: This logic should be shared.

microsoft/testsuites/cloud_hypervisor/ch_tests.py
```diff
@@ -70,8 +70,20 @@ def verify_cloud_hypervisor_integration_tests(
...
+        test_list_raw = variables.get("ch_integration_test_list", "")
```
squirrelsc: Make the name more consistent and simple, like `ch_tests_included`, `ch_tests_excluded`

pupacha: Changed.

microsoft/testsuites/cloud_hypervisor/ch_tests.py
pupacha: created a function to get the list and used it in relevant places.

squirrelsc: You can use one function to return two variables, like
```python
a, b = get_lists(....)

def get_lists(...):
  return ["a"], ["b"]
```

pupacha: why return two lists? Isn't it generic to do it with one variable?

squirrelsc: We should consider merging duplicated code as much as possible. 2 lines to 1 line, complex line to simple line. For example, there are two strings, it may be typo. In future, if the logic needs to be changed, it needs to change both places.

pupacha: Made changes to return two lists.

squirrelsc (approved):

anirudhrb (approved):

PR #2550, opened 2023-01-24
Add transformer to install tools

adityagesh:
Use ToolInstallerTransformer to install tools
to VM using LISA without running test cases

### DIFF
--- a/lisa/transformers/tool_installer.py
+++ b/lisa/transformers/tool_installer.py
@@ -1,0 +1,55 @@
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import schema, tools
+from lisa.executable import Tool
+from lisa.node import quick_connect
+from lisa.transformer import Transformer
+from lisa.util import LisaException, field_metadata
+
+
+@dataclass_json
+@dataclass
+class ToolInstallerTransformerSchema(schema.Transformer):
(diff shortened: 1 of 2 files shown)


lisa/transformers/tool_installer.py
```diff
@@ -0,0 +1,55 @@
...
+@dataclass
+class ToolInstallerTransformerSchema(schema.Transformer):
+    tool_names: List[str] = field(default_factory=List[str])
+    # the SSH connection information to the node
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+
+
+class ToolInstallerTransformer(Transformer):
```
squirrelsc, line 23: What is this used for? Can it be a test case in the internal repo?

adityagesh: Often, we verify VM behavior by manually executing the commands instead of running the automated test case. Mostly when trying to debug an issue found while executing a test or for performance related tests. This Transformer can be used in such cases to install the required tools in a repro VM, and then manually execute the commands.

I did not understand what you meant by "can it be a test case". I can ping you offline if more discussion is required.

squirrelsc: The similar logic can be in a test case, the test case is to install the tool. If it's not to setup something for every test case, it doesn't need to use transformer. The test case provides more rich functionality, like to run with combinators, set req and others. The env can be kept by keep_environment:always

PR #2553, opened 2023-01-24
[fix issue] skip check Hyper-v interrupts on arm64

LiliDeng:


microsoft/testsuites/core/cpu.py
```diff
@@ -187,8 +188,9 @@ def verify_vmbus_interrupts(self, node: Node, log: Logger) -> None:
 
             found_hyperv_interrupt = True
 
+        arch = node.os.get_kernel_information().hardware_platform  # type: ignore
         # Fail test execution if these hyper-v interrupts are not showing up
-        if not found_hyperv_interrupt:
+        if arch != CpuArchitecture.ARM64 and not found_hyperv_interrupt:
```
squirrelsc, line 193: ARM64 should skip this test case, instead of pass it.

sharsonia: No it shouldn't skip the test case. There is a way to check the hyper-v interrupts. I have the change, will push out PR by today EOD

LiliDeng: I have two bugs for this failure under my name. And I confirmed with Boqun that it is expected that these three types of interrupts not shown in arm64 VMs. Seems I miss some things. @sharsonia if you have PR to do such work, I'll close this one.

sharsonia: @LiliDeng oops my bad. In my mind, I was thinking about timesync_validate_ptp test. Sorry about that.
But now that we are talking about vmbus interrupts, let me verify for this as well since I remember there was a specific issue related to interrupts on ARM64 that affected perf. So let me check on that

sharsonia: So I checked the ARM64 interrupt issue that I mentioned. It seems there are some timer interrupts for ARM64, just not sure if we want to include those and check for as part of this test case?

LiliDeng: > So I checked the ARM64 interrupt issue that I mentioned. It seems there are some timer interrupts for ARM64, just not sure if we want to include those and check for as part of this test case?

You can include it in this case if you want to check interrupts for ARM64, please double confirm with Boqun.

LiliDeng: > ARM64 should skip this test case, instead of pass it.

this line still validates arm64 https://github.com/microsoft/lisa/blob/main/microsoft/testsuites/core/cpu.py#L148

LiliDeng: > > So I checked the ARM64 interrupt issue that I mentioned. It seems there are some timer interrupts for ARM64, just not sure if we want to include those and check for as part of this test case?
> 
> You can include it in this case if you want to check interrupts for ARM64, please double confirm with Boqun.

I'll finish this PR, you can create a new PR once know what interrupts can be validated on arm64

LiliDeng: > > > So I checked the ARM64 interrupt issue that I mentioned. It seems there are some timer interrupts for ARM64, just not sure if we want to include those and check for as part of this test case?
> > 
> > 
> > You can include it in this case if you want to check interrupts for ARM64, please double confirm with Boqun.
> 
> I'll finish this PR, you can create a new PR once know what interrupts can be validated on arm64

@squirrelsc Could you please review it again? this line still validates arm64 https://github.com/microsoft/lisa/blob/main/microsoft/testsuites/core/cpu.py#L148

squirrelsc (approved):

PR #2557, opened 2023-01-25
Support newer Python versions

avylove:
This PR adds testing for Python 3.11 and Python 3.12 pre-release. Since 3.12 is pre-release, the tests are set for non-blocking with `continue-on-error`.

A few changes were needed to support Python 3.11+
- The versions of PyGObject and Pillow needed to be upgraded for the `azure` optional dependencies
- Default values for dataclass fields must be hashable. To use an unhashable type, a default_factory must be specified instead.
- There was a bug in LISA that happened to work in Python <=3.10
  - `lisa.search_space.RequirementMethod` inherited from `str` and `enum.Enum` and the values were used like strings.
  - This happens to work in Python <=3.10 because `f'{RequirementMethod.generate_min_capability}'` results in `'generate_min_capability'`, but in Python >=3.11 it results in `'RequirementMethod.generate_min_capability'`
  - The method this was used in has the argument defined as `str` rather than `RequirementMethod`, so the issue was masked to Mypy

3.12 seems to work as expected on Linux and, for the example, on Windows, but the unit tests require the Azure dependencies and this causes excessive backtracking (reviewing the same dependencies multiple times) in Pip. The unit tests have been disabled for 3.12 on Windows for now and we can try again in a month or two.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -7,6 +7,7 @@ on:
       - main
 jobs:
   Check:
+    timeout-minutes: 40
     continue-on-error: ${{ matrix.optional || false }}
     runs-on: ${{ matrix.os }}
     name: >-
@@ -18,9 +19,20 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, windows-2019]
-        python-version: ['3.9', '3.10']
+        python-version: ['3.9', '3.10', '3.11', '3.12-dev']
         nox-session: [test, example]
+
+        exclude:
+          # Azure installion (Win, 3.12) currently hangs due to excessive pip backtracking
+          # Disable unit tests and retry in 2Q23
+          - python-version: '3.12-dev'
+            os: windows-2019
+            nox-session: test
+
         include:
+
+          - python-version: '3.12-dev'
+            optional: true
 
           - os: ubuntu-latest
             python-version: ''
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -23,16 +23,13 @@ jobs:
         nox-session: [test, example]
 
         exclude:
-          # Azure installion (Win, 3.12) currently hangs due to excessive pip backtracking
+          # Azure installation (Win, 3.12) currently hangs due to excessive pip backtracking
           # Disable unit tests and retry in 2Q23
           - python-version: '3.12-dev'
             os: windows-2019
             nox-session: test
 
         include:
-
-          - python-version: '3.12-dev'
-            optional: true
 
           - os: ubuntu-latest
             python-version: ''
--- a/docs/installation_linux.rst
+++ b/docs/installation_linux.rst
@@ -4,7 +4,7 @@ Install LISA on Linux
 Minimum System Requirements
 ---------------------------
 
-1. Your favorite Linux distribution supporting Python 3.8 - 3.10
+1. Your favorite Linux distribution supporting Python 3.8 or above
 2. Dual core processor
 3. 4 GB system memory
 
@@ -15,8 +15,8 @@ The following commands assume Ubuntu is being used.
 Install Python on Linux
 -----------------------
 
-LISA has been tested to work with `Python 3.8 - 3.10 64-bit <https://www.python.org/>`__.
-Python 3.10 is recommended. Support for 3.11+ is under development.
+LISA has been tested to work with `Python >=3.8 64-bit <https://www.python.org/>`__.
+Python 3.11 is recommended.
 If you find that LISA is not compatible with a supported version,
 `please file an issue <https://github.com/microsoft/lisa/issues/new>`__.
 
@@ -29,14 +29,12 @@ To check which version of Python is used on your system, run the following:
 If you need to install a different Python package, there are likely packaged versions for
 your distro.
 
-Here is an example to install Python 3.10 on Ubuntu 20.04
+Here is an example to install Python 3.11 on Ubuntu 22.04
 
 .. code:: bash
 
    sudo apt update
-   sudo apt install software-properties-common -y
-   sudo add-apt-repository ppa:deadsnakes/ppa -y
-   sudo apt install python3.10 python3.10-dev -y
+   sudo apt install python3.11 python3.11-dev -y
 
 
 Install system dependencies
--- a/docs/installation_windows.rst
+++ b/docs/installation_windows.rst
@@ -26,7 +26,7 @@ The full installer allows greater customization and doesn't have the security re
 of the Microsoft Store packages, so may be preferred in some situations.
 
 Navigate to `Python releases for Windows <https://www.python.org/downloads/windows/>`__.
-Download and install *Windows installer (64-bit)* for Python 3.8 - 3.10 64-bit.
+Download and install *Windows installer (64-bit)* for Python 3.8 64-bit or above.
 
 More information on the full installer, including installation without a GUI,
 can be found `here <https://docs.python.org/3/using/windows.html#the-full-installer>`_.
--- a/lisa/environment.py
+++ b/lisa/environment.py
@@ -79,7 +79,7 @@ def _get_environment_id() -> int:
 class EnvironmentMessage(MessageBase):
     type: str = "Environment"
     name: str = ""
-    runbook: schema.Environment = schema.Environment()
+    runbook: schema.Environment = field(default_factory=schema.Environment)
     status: EnvironmentStatus = EnvironmentStatus.New
 
 
--- a/lisa/features/nvme.py
+++ b/lisa/features/nvme.py
@@ -3,6 +3,7 @@
 
 import re
 from dataclasses import dataclass, field
+from functools import partial
 from typing import Any, List, Type
 
 from dataclasses_json import dataclass_json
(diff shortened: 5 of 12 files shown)


squirrelsc:
@avylove Can you remove 3.12-dev for now, so this change can bring 3.11 support in? And please update the versions in document too.

avylove:
I'll see if I can find time to update the docs, but why would we remove 3.12-dev? Even if it fails, it's non-blocking (`continue-on-error: true`)

squirrelsc:
> I'll see if I can find time to update the docs, but why would we remove 3.12-dev? Even if it fails, it's non-blocking (`continue-on-error: true`)

We can wait azure and 3.12 released, and revisit it. The pipelines should be pass or fail. optional failure is confusing.

docs/installation_linux.rst
```diff
@@ -29,14 +29,12 @@ To check which version of Python is used on your system, run the following:
 If you need to install a different Python package, there are likely packaged versions for
 your distro.
 
-Here is an example to install Python 3.10 on Ubuntu 20.04
+Here is an example to install Python 3.11 on Ubuntu 22.04
```
squirrelsc, line 32: is it tested under this combination?

docs/installation_linux.rst
```diff
@@ -29,14 +29,12 @@ To check which version of Python is used on your system, run the following:
...
 
 .. code:: bash
 
    sudo apt update
-   sudo apt install software-properties-common -y
-   sudo add-apt-repository ppa:deadsnakes/ppa -y
-   sudo apt install python3.10 python3.10-dev -y
+   sudo apt install python3.11 python3.11-dev -y
```
squirrelsc, line 37: What's the reason to add -dev?

avylove:
Pushed doc changes.
Optional pipelines are standard and intended for exactly use cases like this. They are not confusing for anyone familiar with any of the major CI platforms.
There's no good reason to defer 3.12. Waiting for GA is way too late. It already works and now we can avoid introducing changes that will break it.

docs/installation_linux.rst
avylove: I was being consistent with the existing example, but looking at the package, the -dev package seems to mainly have additional documentation. We can probably drop it.

docs/installation_linux.rst
avylove: Yes

squirrelsc:
> Optional pipelines are standard and intended for exactly use cases like this. They are not confusing for anyone familiar with any of the major CI platforms. There's no good reason to defer 3.12. Waiting for GA is way too late. It already works and now we can avoid introducing changes that will break it.

Contributors are not expected to be familiar with CI. It looks the only blocking comments to test and merge this PR further.

avylove:
Optional workflows will not block contributors, that's why they are optional. The alternative is to make them required, which is not what we want.

squirrelsc:
> Optional workflows will not block contributors, that's why they are optional. The alternative is to make them required, which is not what we want.

It doesn't block contributors. It's blocking me to bring this PR to be merged. Please remove 3.12 in the workflow, and I will ask Lili to do more testing, and merge this PR.

avylove:
Propose another way to ensure we don't introduce changes that will break on 3.12. I have implemented the standard way this is done. I'm open to other ideas, but you haven't provided one and evidence shows you have not addressed this adequately in the past. I suggest you take this opportunity to learn something new.

squirrelsc:
> Propose another way to ensure we don't introduce changes that will break on 3.12. I have implemented the standard way this is done. I'm open to other ideas, but you haven't provided one and evidence shows you have not addressed this adequately in the past. I suggest you take this opportunity to learn something new.

I don't want to set an expectation on contributors, they need to be familiar with CI system. I explained many times to new contributors to fix failures in checks, so we can merge PRs. I don't want to tell them like "you should fix failures in checks, but if it's optional, it can be ignored." I really appreciate your contribution on this, and I really want to bring the 3.11 supports into LISA. But if you are insisting on bringing 3.12-dev together. I have to create another PR, and cherry pick most changes from this PR to maximize your credits.

avylove:
You still haven't addressed the issue. 3.11 support is almost 5 months past GA when it should have been supported with 3.11 beta when it was released nearly 11 months ago.

You have not indicated how you intend to ensure 3.12 support is not broken like 3.11 support was. If you can not provide this, I don't see how you can hold up this PR.

So I will ask you very clearly. How you intend to ensure 3.12 support is not broken by future commits?

squirrelsc:
My point is to avoid optional checks. It's higher priority than others to me. If there is a perfect way to meet both, of course to take.

BTW, I see all checks passed now. Can it be required?

avylove:
We can make it required, I don't mind. The benefit of making it optional is we don't have to block if there is a 3.12-specific issue that might take more time to investigate. I don't expect that to happen. Once it works, it usually keeps working.

squirrelsc:
> We can make it required, I don't mind.

Good to know, can you remove the optional check? So we can continue to do more validation on this PR?

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -18,8 +19,16 @@ jobs:
...
       matrix:
         os: [ubuntu-20.04, windows-2019]
-        python-version: ['3.9', '3.10']
+        python-version: ['3.9', '3.10', '3.11', '3.12-dev']
         nox-session: [test, example]
+
+        exclude:
+          # Azure installation (Win, 3.12) currently hangs due to excessive pip backtracking
+          # Disable unit tests and retry in 2Q23
+          - python-version: '3.12-dev'
```
squirrelsc, line 28: Is it still an issue? Did you try LISA with Azure? It's our main scenario. If so, can you file a bug to Azure SDK? We should document it in our document. Ask users not to try 3.12 with LISA explicitly, that will save their time on troubleshooting it.

avylove: Yes, it's still an issue, but only on Windows. I did some digging and it seems there are two dependencies that have issues installing.
azure-identify requires portalocker which requires pywin32. pywin32 added support for 3.12 in November and are testing against it, but haven't done a release since then, so I filed a bug with them.
Pillow added 3.12 support to their repo in February and plan to role it out with Pillow 10.0 by July. At that point there should be a pre-built binary in PyPI that will work for us.
There may be other issues beyond these, but these are currently what is preventing the installation of azure dependencies on Windows.

squirrelsc: Thanks for confirming. Can we update document to say explicitly that LISA doesn't support 3.12. It doesn't need to include the details. Just avoid users to try LISA with Python 3.12, and waste their time.

avylove: Do we need that? Until it's GA then it's normal to assume it's not fully supported and if a downstream user finds an issue with 3.12, then we still want them to report it so we can track it.

squirrelsc:
@LiliDeng This PR looks good to me overall. Please merge it in next cycle. After merged it, please document that 3.12 is not supported so far with the issue, which mentioned by Avram.

LiliDeng:
tested tier 1 tests against gallery image, kvm stack test, work well. will merge it next week

LiliDeng:
@avylove please rebase latest code, I saw some errors when rebase latest code, and when run test against your branch, it has below error.

```
  File "D:\a\_work\1\s\lisa\lisa\runner.py", line 221, in start
    self._start_loop()
  File "D:\a\_work\1\s\lisa\lisa\runner.py", line 398, in _start_loop
    runner = next(runner_iterator)
  File "D:\a\_work\1\s\lisa\lisa\runner.py", line 278, in _fetch_runners
    for runner in runners:
  File "D:\a\_work\1\s\lisa\lisa\runner.py", line 325, in _generate_runners
    runner.initialize()
  File "D:\a\_work\1\s\lisa\lisa\util\__init__.py", line 334, in initialize
    raise identifier
  File "D:\a\_work\1\s\lisa\lisa\util\__init__.py", line 331, in initialize
    self._initialize(*args, **kwargs)
  File "D:\a\_work\1\s\lisa\lisa\runners\lisa_runner.py", line 69, in _initialize
    self._merge_test_requirements(
  File "D:\a\_work\1\s\lisa\lisa\runners\lisa_runner.py", line 765, in _merge_test_requirements
    node_requirement = original_node_requirement.intersect(
  File "D:\a\_work\1\s\lisa\lisa\search_space.py", line 67, in intersect
    return self._intersect(capability)
  File "D:\a\_work\1\s\lisa\lisa\search_space.py", line 81, in _intersect
    return self._call_requirement_method(
  File "D:\a\_work\1\s\lisa\lisa\schema.py", line 1002, in _call_requirement_method
    value.disk = getattr(search_space, method.value)(self.disk, capability.disk)
  File "D:\a\_work\1\s\lisa\lisa\search_space.py", line 674, in intersect
    return _call_requirement_method(
  File "D:\a\_work\1\s\lisa\lisa\search_space.py", line 654, in _call_requirement_method
    result = getattr(requirement, method.value)(capability)
  File "D:\a\_work\1\s\lisa\lisa\search_space.py", line 67, in intersect
    return self._intersect(capability)
  File "D:\a\_work\1\s\lisa\lisa\search_space.py", line 81, in _intersect
    return self._call_requirement_method(
  File "D:\a\_work\1\s\lisa\lisa\schema.py", line 604, in _call_requirement_method
    search_space, f"{method_name}_setspace_by_priority"
NameError: name 'method_name' is not defined
```

LiliDeng:
@avylove I raise a new PR #2689 which includes all your changes + fix for new merged code in RequirementMethod part, close this one, thank you.

.github/workflows/continuous-integration-workflow.yml
LiliDeng: @avylove `LISA with Python 3.12 on windows` doesn't work, but `LISA with Python 3.12 on Linux` works?

avylove: They both work now as far as the tests go. #2740

PR #2564, opened 2023-01-30
[improvement] collect provision and boot time

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -707,15 +710,27 @@ def _get_wala_distro_version(self, node: Node) -> str:
 
         return result
 
+    def _get_boot_time(self, node: Node) -> Dict[str, Any]:
+        boot_time: Dict[str, Any] = {}
+        try:
+            systemd_analyze_tool = node.tools[SystemdAnalyze]
+            boot_time = systemd_analyze_tool.get_boot_time()
+            boot_time["boot_type"] = node.execute(
```
squirrelsc: What's the output of boot_type? How about put it in system analyze or kernel tool?

lisa/testsuite.py
```diff
@@ -181,6 +186,7 @@ def set_status(
             if new_status == TestStatus.RUNNING:
                 self._timer = create_timer()
             self._send_result_message(self.stacktrace)
+            self._send_provision_boot_message()
```
squirrelsc: The boot message is about node, not test cases. So it should be sent from environment setup, not test case running.

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: I may change the column into boot_times, means first boot or second etc

lisa/messages.py
```diff
@@ -232,6 +232,19 @@ class NetworkUDPPerformanceMessage(PerfMessage):
     packet_size_kbytes: Decimal = Decimal(0)
 
 
+@dataclass
+class ProvisionBootTimeMessage(TestResultMessageBase):
+    type: str = "ProvisionBootTime"
+
+    boot_times: int = 0
```
squirrelsc, line 240: add comments above to explain it's how many times, the boot message is for.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1448,6 +1450,7 @@ def initialize_environment(self, environment: Environment, log: Logger) -> None:
                 password=node_context.password,
                 private_key_file=node_context.private_key_file,
             )
+            node.provision_time = environment_context.provision_time
```
LiliDeng, line 1455: @squirrelsc I add provision_time in node level to pass this value, is it good?

lisa/operating_system.py
```diff
@@ -382,6 +383,16 @@ def capture_system_information(self, saved_path: Path) -> None:
             saved_path / "modinfo-hv_netvsc.txt"
         )
 
+        if hasattr(self._node, "first_initialize") and self._node.first_initialize:
+            from lisa.tools import SystemdAnalyze
+
+            try:
+                systemd_analyze_tool = self._node.tools[SystemdAnalyze]
+                boot_time = systemd_analyze_tool.get_boot_time()
+                notifier.notify(boot_time)
```
LiliDeng: @squirrelsc I am not sure how to pass the testresults id into provision boot time message, there is testresultid column in provisionboottime table.

lisa/node.py
```diff
@@ -252,6 +252,8 @@ def cleanup(self) -> None:
         if hasattr(self, "_log_handler") and self._log_handler:
             remove_handler(self._log_handler, self.log)
             self._log_handler.close()
+        if hasattr(self, "first_initialize"):
```
squirrelsc: 1. The log_handler is to save two states (first time, initialized), so the getattr is used to check it. the first_initialize doesn't need. So just set it to false in `__init__`.
2. The name should be `_first_initialize`, because it's a private value.
3. The value should be managed in _initialize of node, not to call it out of node. The reason is to keep the states self-contained.

lisa/node.py
```diff
@@ -381,7 +383,11 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
             self._log_handler = create_file_handler(
                 self.local_log_path / "node.log", self.log
             )
-
+        if self.is_test_target:
```
squirrelsc: It doesn't matter it's test target or not to change the `_first_initialize`. If it needs to skip some real logic, just check `is_test_target` there.

lisa/operating_system.py
squirrelsc: The boot message is not related to test cases. So we may consider remove test result id from the message.

LiliDeng: then we need add platform, kernelversion, hostversion (if it has), image, vmsize, location info for provisionbootmessage.

lisa/node.py

lisa/node.py
LiliDeng: got it, updated

lisa/operating_system.py
squirrelsc: Yes, we need to add those informations to boot table.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -613,14 +616,45 @@ def _delete_boot_diagnostic_container(
...
+            information["vmsize"] = node_runbook.vm_size
+            information["image"] = node_runbook.get_image_name()
+        information["platform"] = self.type_name()
+        return information
 
-        node.log.debug("detecting vm generation...")
-        information[KEY_VM_GENERATION] = node.tools[VmGeneration].get_generation()
-        node.log.debug(f"vm generation: {information[KEY_VM_GENERATION]}")
+    @hookimpl
+    def get_node_information(self, node: Node) -> Dict[str, str]:
```
squirrelsc: It doesn't need to have another layer of wrapper.

squirrelsc: BTW, where is it registered? It may need to follow the pattern of env info, so it can avoid to be registered multiple times.

LiliDeng: I didn't add any register, but I saw it was invoked correctly.

squirrelsc: It may be invoked when call get_environment_information.

LiliDeng: I set breakpoint, it was invoked from 
```
                    informations: List[
                        Dict[str, str]
                    ] = plugin_manager.hook.get_node_information(node=self._node)
```

squirrelsc: It sounds not reliable. Did you try on ready or local run? Make sure Azure hook is not invoked.

LiliDeng: the register happened in platform class.
```
class Platform(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
        super().__init__(runbook)
        self._log = get_logger("", self.type_name())
        plugin_manager.register(self)
```

LiliDeng: > It sounds not reliable. Did you try on ready or local run? Make sure Azure hook is not invoked.

tried on ready, this hook was not invoked.

squirrelsc (approved):

LiliDeng:
testing

PR #2574, opened 2023-02-02
Add a fix of dns case for Alpine OS

lubaihua33:
1. Add operating system Alpine
2. Remove '-O' option of ping command for Alpine system

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -618,6 +618,12 @@ class CoreOs(Linux):
         return re.compile("^coreos|Flatcar|flatcar$")
 
 
+class Alpine(Linux):
+    @classmethod
+    def name_pattern(cls) -> Pattern[str]:
+        return re.compile("^Alpine")
+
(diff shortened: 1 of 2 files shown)


lisa/operating_system.py
```diff
@@ -618,6 +618,12 @@ def name_pattern(cls) -> Pattern[str]:
         return re.compile("^coreos|Flatcar|flatcar$")
 
 
+class Alpine(Linux):
+    @classmethod
+    def name_pattern(cls) -> Pattern[str]:
+        return re.compile("^Alpine")
```
LiliDeng, line 624: do you know it is centos or Debian type?

lubaihua33: It is not either type. It is a security-oriented, lightweight Linux distribution. It is designed for routers, firewalls, and virtual private networks. We have publishers using this system.

LiliDeng (approved):

PR #2575, opened 2023-02-03
Support concurrency in qemu/ch platforms

pupacha:
* Changed locks and port forwarding variables to static so that there is consistency when concurrency is used or with multiple lisa runners.
* Using lock to to perform certain host_node.<operation> atomically


pupacha:
@anirudhrb @cwize1 Please review

PR #2576, opened 2023-02-03
Add a fix for netvsc reload case

lubaihua33:
Sometimes the ssh session is inactive after reloading hv_nevsc multiple times, but the VM can be connected actually. In this PR, a try except is added when reload hv_netvsc module to process the false negative.

### DIFF
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -20,6 +20,8 @@ class Modprobe(Tool):
     # to leave the node without a network connection if things go wrong.
     def _reload_hv_netvsc(self) -> None:
         # These commands must be sent together, bundle them up as one line
+        # If the VM is disconnected after running below command, wait 60s is enough.
+        # Don't need to wait the default timeout 600s. So set timeout 60.
         self.node.execute(
             "modprobe -r hv_netvsc; modprobe hv_netvsc; "
             "ip link set eth0 down; ip link set eth0 up;"
@@ -27,6 +29,7 @@ class Modprobe(Tool):
             sudo=True,
             shell=True,
             nohup=True,
+            timeout=60,
         )
 
     def is_module_loaded(
--- a/microsoft/testsuites/network/netinterface.py
+++ b/microsoft/testsuites/network/netinterface.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 import time
+from pathlib import Path
 
 from assertpy import assert_that
 from randmac import RandMac  # type: ignore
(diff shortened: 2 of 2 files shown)


lisa/tools/modprobe.py
```diff
@@ -20,13 +20,16 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+        # If the VM is disconnected after running below command, wait 60s is enough.
+        # Don't need to wait the default timeout 600s. So set timeout 60.
         self.node.execute(
             "modprobe -r hv_netvsc; modprobe hv_netvsc; "
             "ip link set eth0 down; ip link set eth0 up;"
             "dhclient -r eth0; dhclient eth0",
             sudo=True,
             shell=True,
             nohup=True,
+            timeout=60,
```
LiliDeng, line 32: do you see timeout case? I so nohup is used, it should return immediately

microsoft/testsuites/network/netinterface.py
```diff
@@ -50,15 +54,33 @@ class NetInterface(TestSuite):
...
+                node.log.debug(identifier)
+                serial_console = node.features[SerialConsole]
+                serial_console.check_panic(
+                    saved_path=log_path, stage="after_reload_netvsc", force_run=True
+                )
+                node.close()
+                remote_node = cast(RemoteNode, node)
+                try_connect(remote_node._connection_info)
+                node.log.debug("The node can be connected.")
+                raise PassedException(
```
LiliDeng: it is in a test loop, throw exception here, it may reduce test times?

microsoft/testsuites/network/netinterface.py
```diff
@@ -50,15 +54,33 @@ class NetInterface(TestSuite):
...
             test_count += 1
             # Unload and load hv_netvsc
-            network_interface_feature.reload_module()
+            try:
+                network_interface_feature.reload_module()
+            except Exception as identifier:
+                # Sometimes the ssh session is inactive, call node.close() and retry
+                # can solve this problem. In this case, if there is no panic, then
+                # raise PassedException
+                node.log.debug(identifier)
```
squirrelsc: add more information in the debug message to explain the error is ignorable, and it's trying again.

microsoft/testsuites/network/netinterface.py
squirrelsc: Agree, it needs to add another loop to retry here. If it's passed in retry, it should try next test_count, instead of raise PassedException.

lisa/tools/modprobe.py
lubaihua33: I saw it returned after 3~5 mins several times when ssh session is becoming inactive or the VM couldn't be connected.

microsoft/testsuites/network/netinterface.py
lubaihua33: Done

microsoft/testsuites/network/netinterface.py
lubaihua33: Done

LiliDeng (approved):

squirrelsc (approved):

PR #2582, opened 2023-02-06
[nested] add check before launching L2 vm

LiliDeng:


lisa/tools/qemu.py
```diff
@@ -55,12 +57,14 @@ def create_vm(
 
         # store name of tap interfaces added to the vm
         added_taps: List[str] = []
-
+        cmd = ""
+        if timeout > 0:
+            cmd = f"timeout {timeout}  {self.command} "
```
dsrivastavv, line 62: Can we use timeout in `self.run` directly?

lisa/tools/qemu.py
```diff
@@ -95,21 +99,35 @@ def create_vm(
...
-        # start qemu
-        self.run(
-            cmd,
-            sudo=True,
-            shell=True,
-            expected_exit_code=0,
-            expected_exit_code_failure_message=f"Unable to start VM {guest_image_path}",
-        )
+        if timeout > 0:
+            cmd_result = self.node.execute(
```
dsrivastavv, line 112: Does this command exit after starting up qemu if `demonize` is not present?

microsoft/testsuites/nested/common.py
```diff
@@ -67,6 +68,21 @@ def qemu_connect_nested_vm(
...
+            nic_model=nic_model,
+            taps=taps,
+            bridge=bridge,
+            disks=disks,
+            cores=cores,
+            stop_existing_vm=stop_existing_vm,
+            run_in_background=False,
+            timeout=60,
+        )
+        qemu.interface_count = 0
```
dsrivastavv, line 85: Why is this needed?

LiliDeng:
after discussing with @somil55 , he will take care of this.

PR #2583, opened 2023-02-06
Update instructions to install in dev environment

dsrivastavv:


docs/installation_linux.rst
```diff
@@ -89,7 +89,13 @@ Clone code
...
 
 
-Directly install LISA (Option 1)
+Install in Dev environment (Option 1)
+-------------------------------------
+
+For making any code changes to LISA, you will need to setup a development environment. Please follow the instructions in :ref:`DevVirtEnv`.
+
+
+Directly install LISA (Option 2)
```
squirrelsc: It's confusing to provide options, instead of scenario based. The option 2 and option 3 is not helping to understand when to use them.

dsrivastavv: How about something like:
1. Development environment
2. Development environment without `nox`
3. Runtime environment

squirrelsc: > How about something like:
> 
> 1. Development environment
> 2. Development environment without `nox`
> 3. Runtime environment

It should be **scenario** based, instead of describing the underlying ways. User doesn't need to learn what's "nox", or when the "runtime" environment is needed. We can provide two options, development or run cases. Give an introduction.  explain what "development" means (like change lisa code).

dsrivastavv: Sure! Let me update

docs/installation_windows.rst
```diff
@@ -101,15 +105,6 @@ To install system-wide, run from and Administrator console.
 
 
 
-Install LISA in a virtual environment (Option 2)
```
dsrivastavv, line 104: @avylove Are there scenarios where user might have to use this installation?

squirrelsc (approved):

docs/installation_windows.rst
avylove: Given the large number of dependencies and the fact most are pinned, unless your environment is only used for LISA, I would expect installing in a virtual environment, either self-constructed or built using nox, is the only reasonable way to install LISA.

PR #2584, opened 2023-02-06
Fix model name regex in lscpu

dsrivastavv:


lisa/tools/lscpu.py
```diff
@@ -84,7 +84,7 @@ class Lscpu(Tool):
     )
     # Model name:          Intel(R) Xeon(R) Platinum 8168 CPU @ 2.70GHz
     # Model name:          AMD EPYC 7763 64-Core Processor
-    __cpu_model_name = re.compile(r"^Model name:\s+(?P<model_name>.*)\s*$", re.M)
+    __cpu_model_name = re.compile(r"^\s*Model name:\s+(?P<model_name>.*)\s*$", re.M)
```
squirrelsc, line 88: Add new examples above the regexp.

squirrelsc (approved):

PR #2585, opened 2023-02-08
fix release not correct issue

LiliDeng:


lisa/operating_system.py
```diff
@@ -1460,32 +1460,22 @@ def _is_package_in_repo(self, package: str) -> bool:
...
-            assert matches.group("vendor")
-            information = OsInformation(
-                version=self._parse_version(matches.group("version")),
-                vendor=matches.group("vendor"),
-                release=matches.group("version"),
-                codename=matches.group("codename"),
-                full_version=full_version,
-            )
-
+        cmd_result = self._node.execute(
```
squirrelsc: What's the reason to change it back to redhat-release? The redhat-release may be removed in future. We can use redhat-release as first try, but the os-release should be in except to prevent it fails in future.

LiliDeng: for some redhat and centos distro, the release got from os-release is 8. but actually it is 8.5, okay, I will change the code accordingly.

squirrelsc (approved):

PR #2586, opened 2023-02-08
add a new test function for IB naming check

darkwhite29:
checking the IB device naming/ordering to make sure IB cards named starting with "ib0".

### DIFF
--- a/microsoft/testsuites/hpc/infinibandsuite.py
+++ b/microsoft/testsuites/hpc/infinibandsuite.py
@@ -1,6 +1,5 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-
 from assertpy import assert_that
 
 from lisa import (
@@ -16,6 +15,7 @@ from lisa.features import Infiniband, Sriov
 from lisa.sut_orchestrator.azure.tools import Waagent
 from lisa.tools import Find, KernelConfig, Modprobe, Ssh
 from lisa.util import (
+    LisaException,
     SkippedException,
     UnsupportedDistroException,
     UnsupportedKernelException,
@@ -31,6 +31,26 @@ from lisa.util.parallel import run_in_parallel
     """,
 )
 class InfinibandSuite(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. List all available network interfaces
+        2. Check if InfiniBand cards are present
+        3. Ensure the first InfiniBand card is named starting with "ib0"
+        """,
+        priority=2,
+        requirement=simple_requirement(supported_features=[Infiniband]),
+    )
+    def verify_ib_naming(self, log: Logger, node: Node) -> None:
+        ib_interfaces = node.features[Infiniband].get_ib_interfaces()
+        ib_first_device_name = ib_interfaces[0].ib_device_name
+
+        if not ib_first_device_name:
+            raise LisaException("This node has no IB devices available.")
+        if "ib0" not in ib_first_device_name:
+            raise LisaException("The first IB device on this node is not named ib0.")
+        log.info("IB device naming/ordering has been verified successfully!")
+
     @TestCaseMetadata(
         description="""
         This test case will


microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -31,6 +32,33 @@
...
+        1. List all available network interfaces
+        2. Check if InfiniBand cards are present
+        3. Ensure InfiniBand cards are named starting with "ib0"
+        """,
+        priority=2,
+        requirement=simple_requirement(supported_features=[Infiniband]),
+    )
+    def verify_ib_naming(self, log: Logger, node: Node) -> None:
+        infiniband = node.features[Infiniband]
+        ib_interfaces = infiniband.get_ib_interfaces()
```
dsrivastavv: [nit] You can directly write `node.features[Infiniband].get_ib_interfaces()`. We do this only if the tool is invoked multiple times

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -31,6 +32,33 @@
...
+        ib_first_device_name = ib_interfaces[0].ib_device_name
+    
+        r = re.compile("ib[0-9]")
+        ib_names = list(filter(r.match, ib_first_device_name))
+        if len(ib_names) == 0:
+            raise UnsupportedDistroException(
```
dsrivastavv: Do you have a list of unsupported distros? If so, we can skip the test for those distros since those are not testable but this error indicates failure.

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -31,6 +32,33 @@
...
+                node.os, "This node has no IB devices available."
+            )
+        if 'ib0' not in ib_names:
+            raise UnsupportedDistroException(
```
dsrivastavv: raise LISAException. This is a test failure

microsoft/testsuites/hpc/infinibandsuite.py
darkwhite29: Resolved as suggested.

microsoft/testsuites/hpc/infinibandsuite.py
darkwhite29: Modified to be LISAException.

microsoft/testsuites/hpc/infinibandsuite.py
darkwhite29: Resolved as suggested.

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -31,6 +33,30 @@
...
+        1. List all available network interfaces
+        2. Check if InfiniBand cards are present
+        3. Ensure InfiniBand cards are named starting with "ib0"
+        """,
+        priority=2,
+        requirement=simple_requirement(supported_features=[Infiniband]),
+    )
+    def verify_ib_naming(self, log: Logger, node: Node) -> None:
+        ib_interfaces = node.features[Infiniband].get_ib_interfaces()
+        ib_first_device_name = ib_interfaces[0].ib_device_name
```
dsrivastavv, line 46: Your test descriptions says that the test checks all infiniband cards have name starting with "ib0" whereas in code you have only checked first device. Can you please clarify this?

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -31,6 +33,30 @@
...
+
+        r = re.compile("ib[0-9]")
```
dsrivastavv: [nit] Improve variable naming and provide an example in comments above this line

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -31,6 +33,30 @@
...
+        ib_names = list(filter(r.match, ib_first_device_name))
```
dsrivastavv: `ib_first_device_name` seems like a string. We can simply check using: `r.match(ib_first_device_name)` and it returns none if the regex match fails

microsoft/testsuites/hpc/infinibandsuite.py
darkwhite29: Modified the test description for clarification.

microsoft/testsuites/hpc/infinibandsuite.py
dsrivastavv: @darkwhite29 unresolving this. This is still not fixed. `ib_first_device_name` should be a list but is a string instead. You need to do something like this:
```
ib_first_device_name = [ib_interface.ib_device_name for ib_interface in ib_interfaces]
```

Also please update the above naming from `ib_first_device_name -> ib_first_device_names`

dsrivastavv (approved):
LGTM! @LiliDeng Can you please take a look?

LiliDeng:
@darkwhite29 if test case fail for naming reason, what is the next step?

darkwhite29:
> @darkwhite29 if test case fail for naming reason, what is the next step?

We will report the node ID to an IcM ticket for this naming violation, and pull out this node from future allocation.

LiliDeng:
> > @darkwhite29 if test case fail for naming reason, what is the next step?
> 
> We will report the node ID to an IcM ticket for this naming violation, and pull out this node from future allocation.

There is a known issue that ib device's name is not ib0, caused by the guest, if we add boot kernel parameter net.ifnames=0, then reboot, the ib name will be consistent. I am not sure if the issue you mentioned here is the same as the known issue or not, let's take this offline.

darkwhite29:
> > > @darkwhite29 if test case fail for naming reason, what is the next step?
> > 
> > 
> > We will report the node ID to an IcM ticket for this naming violation, and pull out this node from future allocation.
> 
> There is a known issue that ib device's name is not ib0, caused by the guest, if we add boot kernel parameter net.ifnames=0, then reboot, the ib name will be consistent. I am not sure if the issue you mentioned here is the same as the known issue or not, let's take this offline.

We have seen cases like IB device "ib0" being altered as "ib1" which causes additional issues. Probably it is what you described. This test is for detecting such cases. Do you think we can add your solution into this test? Thanks.

LiliDeng:
> > > > @darkwhite29 if test case fail for naming reason, what is the next step?
> > > 
> > > 
> > > We will report the node ID to an IcM ticket for this naming violation, and pull out this node from future allocation.
> > 
> > 
> > There is a known issue that ib device's name is not ib0, caused by the guest, if we add boot kernel parameter net.ifnames=0, then reboot, the ib name will be consistent. I am not sure if the issue you mentioned here is the same as the known issue or not, let's take this offline.
> 
> We have seen cases like IB device "ib0" being altered as "ib1" which causes additional issues. Probably it is what you described. This test is for detecting such cases. Do you think we can add your solution into this test? Thanks.

do you have detailed image and vm sku info which can repro the issue?

darkwhite29:
> > > > > @darkwhite29 if test case fail for naming reason, what is the next step?
> > > > 
> > > > 
> > > > We will report the node ID to an IcM ticket for this naming violation, and pull out this node from future allocation.
> > > 
> > > 
> > > There is a known issue that ib device's name is not ib0, caused by the guest, if we add boot kernel parameter net.ifnames=0, then reboot, the ib name will be consistent. I am not sure if the issue you mentioned here is the same as the known issue or not, let's take this offline.
> > 
> > 
> > We have seen cases like IB device "ib0" being altered as "ib1" which causes additional issues. Probably it is what you described. This test is for detecting such cases. Do you think we can add your solution into this test? Thanks.
> 
> do you have detailed image and vm sku info which can repro the issue?

The case occurred at a PBS cluster of nodes with SLES image. VM SKU info is not clear since no IcM tickets have been filed. Below is the conversation I got from my manager:

"Majid Arabgol 5/26/22 7:43 AM Amirreza , the one my name is in it, is my test cluster and AN is enabled, the screenshot I copied here is belong to cx I copied from ASC( Azure Support center) which AN is not enabled, IB nameing would be different if AN is enabled or disabled ?since both are implemented over the ib, when AN is enabled 2 ib devices appear in the VM, the naming would be random (i.e. on some the ethernet link is mlx5_0 on some it is mlx5_1) and this breaks the mpi which will try to send messages over same link all over the cluster, we need the ib name check script to check for this and make sure all InfiniBand devices are named the same over the cluster. The problem here is, we have a fix on CentOS for this, ie there is a script that makes the namings consistent across the cluster, however it looks to me that script has not been ran on SLES. We may have to share the script with the customer to run it over the cluster and make the names the same."

LiliDeng (approved):

PR #2587, opened 2023-02-09
Libvirt: Support concurrency in ch/qemu platforms

pupacha:
Introduced a new convention of using host_node.operations for atomicity as concurrnecy and combinators deal with multiple instances threads and instances of Platform.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -195,11 +195,11 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         if self.host_node.is_remote:
             self._stop_port_forwarding(environment, log)
 
-        self._capture_libvirt_logs()
-
     def _cleanup(self) -> None:
         if self.platform_runbook.capture_libvirt_debug_logs:
             self._disable_libvirt_debug_log()
+
+        self._capture_libvirt_logs()
 
         if self.host_node.is_remote:
             dmesg_output = self.host_node.tools[Dmesg].get_output(force_run=True)
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -89,6 +89,14 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         StartStop,
     ]
 
+    # Static variables to handle multiple task_runners
+    # Port var are used  for port forwarding in case of Remote Host
+    # 49512 is the first available private port
+    _next_available_port: int = 49152
+    _port_forwarding_lock: Lock = Lock()
+    # Lock used for scp-ing disk image to Remote host VM
+    _disk_img_copy_lock: Lock = Lock()
+
     def __init__(self, runbook: schema.Platform) -> None:
         super().__init__(runbook=runbook)
         self.libvirt_conn_str: str
@@ -96,13 +104,6 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         self.platform_runbook: BaseLibvirtPlatformSchema
         self.host_node: Node
         self.vm_disks_dir: str
-
-        # used for port forwarding in case of Remote Host
-        self._next_available_port: int
-        self._port_forwarding_lock: Lock
-
-        # Lock used for scp-ing disk image to Remote host VM
-        self._disk_img_copy_lock: Lock
 
         self._host_environment_information_hooks = {
             KEY_HOST_DISTRO: self._get_host_distro,
@@ -129,12 +130,6 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         libvirt_events_thread.init()
-
-        # 49512 is the first available private port
-        self._next_available_port = 49152
-        self._port_forwarding_lock = Lock()
-
-        self._disk_img_copy_lock = Lock()
 
         self.platform_runbook = self.runbook.get_extended_runbook(
             self.__platform_runbook_type(), type_name=type(self).type_name()
@@ -553,7 +548,7 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         if node_context.os_disk_source_file_path:
             # use lock to avoid multiple environments scp disk img to same
             # os_disk_base_file_path.
-            with self._disk_img_copy_lock:
+            with BaseLibvirtPlatform._disk_img_copy_lock:
                 source_exists = self.host_node.tools[Ls].path_exists(
                     path=node_context.os_disk_base_file_path, sudo=True
                 )
@@ -670,22 +665,23 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
 
             node_port = 22
             if self.host_node.is_remote:
-                with self._port_forwarding_lock:
+                with BaseLibvirtPlatform._port_forwarding_lock:
                     port_not_found = True
                     while port_not_found:
-                        if self._next_available_port > 65535:
+                        if BaseLibvirtPlatform._next_available_port > 65535:
                             raise LisaException(
                                 "No available ports on the host to forward"
                             )
 
                         # check if the port is already in use
                         output = self.host_node.execute(
-                            f"nc -vz 127.0.0.1 {self._next_available_port}"
+                            "nc -vz 127.0.0.1 "
+                            f"{BaseLibvirtPlatform._next_available_port}"
                         )
                         if output.exit_code == 1:  # port not in use
-                            node_port = self._next_available_port
+                            node_port = BaseLibvirtPlatform._next_available_port
                             port_not_found = False
-                        self._next_available_port += 1
+                        BaseLibvirtPlatform._next_available_port += 1
 
                 self.host_node.tools[Iptables].start_forwarding(
                     node_port, local_address, 22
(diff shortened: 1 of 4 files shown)


pupacha:
Thanks @anirudhrb for your inputs on this change.
@cwize1 Please review.

lisa/util/__init__.py
```diff
@@ -352,6 +356,32 @@ def enable(self) -> None:
...
+    For example,
+        obj = MutexObject(someObject)
+        with obj.get_object() as mutexobj:
+            mutexobj.method()
+
+    This is same as doing someObject.method() atomically.
+    """
+
+    __object: T
+    __lock: Lock
```
squirrelsc, line 373: Is the lock applied to specified subclass, or by the base class MutexObject? I'm not sure the behavior, can you have a test?

squirrelsc:
@cwize1 Please have a review.

lisa/util/__init__.py
```diff
@@ -352,6 +356,32 @@ def enable(self) -> None:
...
+
+    def __init__(self, obj: T) -> None:
+        self.__object = obj
+        self.__lock = Lock()
+
+    @contextmanager
+    def get_object(self) -> Generator[T, None, None]:
```
cwize1: IMHO `lock` would be a better name.

lisa/util/__init__.py
```diff
@@ -352,6 +356,32 @@ def enable(self) -> None:
...
+    with multiple threads.
+    For example,
+        obj = MutexObject(someObject)
+        with obj.get_object() as mutexobj:
+            mutexobj.method()
+
+    This is same as doing someObject.method() atomically.
+    """
+
+    __object: T
```
squirrelsc: It doesn't need to specify type in class level. It can be along with in `__init__`.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -89,21 +89,23 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         StartStop,
     ]
 
+    # Static variables to handle multiple task_runners
+    # Port var are used  for port forwarding in case of Remote Host
+    # 49512 is the first available private port
+    _next_available_port: int = 49152
```
cwize1, line 95: I believe only single instance of each platform class is created for the entire LISA process lifetime. So, do you need to make these variables static?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -147,30 +143,34 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+        host_node: Node
         if host.is_remote():
             assert host.address
             if not host.username:
                 raise LisaException("Username must be provided for remote host")
             if not host.private_key_file:
                 raise LisaException("Private key file must be provided for remote host")
 
-            self.host_node = RemoteNode(
+            self._is_host_remote = True
```
cwize1: nit: It would be good to set `self.host_node_object` and `self._is_host_remote` together.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -553,19 +555,20 @@ def _create_node(
...
-                )
-                if source_exists:
-                    self.host_node.tools[Chmod].chmod(
-                        node_context.os_disk_base_file_path, "a+r", sudo=True
-                    )
-                else:
-                    self.host_node.shell.copy(
-                        Path(node_context.os_disk_source_file_path),
-                        Path(node_context.os_disk_base_file_path),
+            with BaseLibvirtPlatform._disk_img_copy_lock:
```
cwize1, line 559: Considering that the `host_node` now has a lock, is the `_disk_img_copy_lock` now redundant?

cwize1: Would it also make sense to fold `_port_forwarding_lock` under the `host_node` lock as well?

cwize1 (changes requested):
.

lisa/util/__init__.py
pupacha: How about `get_locked_object`? Since it returns (yields) the actual object and not lock.

lisa/sut_orchestrator/libvirt/platform.py
pupacha: No, when combinator/concurrency is used, multiple task_runners are created. And each task_runner will create a new platform instance. Hence, it needs to be static.

lisa/sut_orchestrator/libvirt/platform.py
pupacha: `host_node` lock is specific to a platform instance where as `__disk_img_copy_lock` and `_port_forwarding_lock ` are static. These needs to be static as LISA might create multiple instances of platform.

lisa/util/__init__.py
pupacha: `lock` sounds better. change it.

anirudhrb (dismissed):
LGTM. Make sure to do a sanity run with our entire test suite.

pupacha:
@cwize1 Please review again.

cwize1 (approved):

squirrelsc (approved):

pupacha:
Hi @squirrelsc @cwize1 @anirudhrb 
I'm seeing some intermittent issues from libvirt and from LISA, when used with concurrency and combinators. These issues are not consistent though. _And when used without concurrency or combinators, I dont see it showing any issues._ So, I'm thinking to add a warning statement in the platform saying that "It is experimental" whenever concurrency is used. And I will take up and fix them accordingly, whenever there is some consistency in the issues seen. Only when there's some confidence on the working nature of this feature, we will remove the warning statement. What are your thoughts on this ?

squirrelsc:
> Hi @squirrelsc @cwize1 @anirudhrb I'm seeing some intermittent issues from libvirt and from LISA, when used with concurrency and combinators. These issues are not consistent though. _And when used without concurrency or combinators, I dont see it showing any issues._ So, I'm thinking to add a warning statement in the platform saying that "It is experimental" whenever concurrency is used. And I will take up and fix them accordingly, whenever there is some consistency in the issues seen. Only there's some confidence on the working nature of this feature, we will eventually remove the warning statement. What are your thoughts on this ?

What's the issue here? Is there any call trace?

pupacha:
> > Hi @squirrelsc @cwize1 @anirudhrb I'm seeing some intermittent issues from libvirt and from LISA, when used with concurrency and combinators. These issues are not consistent though. _And when used without concurrency or combinators, I dont see it showing any issues._ So, I'm thinking to add a warning statement in the platform saying that "It is experimental" whenever concurrency is used. And I will take up and fix them accordingly, whenever there is some consistency in the issues seen. Only there's some confidence on the working nature of this feature, we will eventually remove the warning statement. What are your thoughts on this ?
> 
> What's the issue here? Is there any call trace?

There's no call trace, but I see the following skip message when network performance tests are run.
`SKIPPED no available environment: ['no enough nodes, requirement: 2, capability: 1.']`
And sporadically libvirt fails deployment with messages like `libvirtError: internal error: failed to boot guest VM`

squirrelsc:
> > > Hi @squirrelsc @cwize1 @anirudhrb I'm seeing some intermittent issues from libvirt and from LISA, when used with concurrency and combinators. These issues are not consistent though. _And when used without concurrency or combinators, I dont see it showing any issues._ So, I'm thinking to add a warning statement in the platform saying that "It is experimental" whenever concurrency is used. And I will take up and fix them accordingly, whenever there is some consistency in the issues seen. Only there's some confidence on the working nature of this feature, we will eventually remove the warning statement. What are your thoughts on this ?
> > 
> > 
> > What's the issue here? Is there any call trace?
> 
> There's no call trace, but I see the following skip message when network performance tests are run. `SKIPPED no available environment: ['no enough nodes, requirement: 2, capability: 1.']` And sporadically libvirt fails deployment with messages like `libvirtError: internal error: failed to boot guest VM`

The first error looks like insufficient resources, for example, memory is not enough to create more VMs. It should be resolved by reduce the concurrency. I guess when the combinator and concurrency are used together, there is higher chance to deploy multiple envs with more nodes. LISA controls concurrency by environment number, not node number. So 4 envs of 1 node may be ok, but 4 envs of 2 nodes may not. In next step, you can raise ResourceAwaitableException when it's needed. It will let LISA runner to wait until timeout for new resources, instead of skipping test case.

For second one, please find where the log is from, and get it more stable. Make sure the libvirt log is collected with right timestamp, and compare with LISA log to evaluate where the error is from.

pupacha:
> The first error looks like insufficient resources, for example, memory is not enough to create more VMs. It should be resolved by reduce the concurrency. I guess when the combinator and concurrency are used together, there is higher chance to deploy multiple envs with more nodes. LISA controls concurrency by environment number, not node number. So 4 envs of 1 node may be ok, but 4 envs of 2 nodes may not. In next step, you can raise ResourceAwaitableException when it's needed. It will let LISA runner to wait until timeout for new resources, instead of skipping test case.

For my testing, I'm using just concurrency (without combinators) as 3, with Standard_D64ds_v5 Azure VM as Host machine and ran network performance tests. This SKU has 64 core and 256 GB RAM which is way good enough for the concurrent testing. Yet I see the `no available environment` mentioned above. Also, I couldn't see enough logs indicating as to which environment deployment has caused this.

> For second one, please find where the log is from, and get it more stable. Make sure the libvirt log is collected with right timestamp, and compare with LISA log to evaluate where the error is from.

This error is coming from the Libvirt daemon of Host machine. This needs more investigation to understand why guest boot is failing.

LiliDeng:
@pupacha any update for this PR?

pupacha:
> @pupacha any update for this PR?

No, I did not get a chance to investigate on the failures much. I'll probably investigate in April when I get some time.

LiliDeng:
feel free to create or reopen it when it is ready for review

PR #2591, opened 2023-02-10
set exit_code default value as -1

LiliDeng:


lisa/runner.py
```diff
@@ -229,16 +229,14 @@ async def start(self) -> None:
 
         if self._results_collector:
             results = [x for x in self._results_collector.results.values()]
+            self.exit_code = 0
```
squirrelsc: It should be set in else after L239. If there there is any failure above, the code is -1.

LiliDeng: updated

squirrelsc (approved):

PR #2593, opened 2023-02-10
use correct dummy command for firewall

LiliDeng:
- use echo ok instead of ls -lt

### DIFF
--- a/lisa/tools/firewall.py
+++ b/lisa/tools/firewall.py
@@ -8,7 +8,7 @@ from lisa.executable import Tool
 class Firewall(Tool):
     @property
     def command(self) -> str:
-        return "ls -lt"
+        return "echo"
 
     @property
     def can_install(self) -> bool:


microsoft/testsuites/network/common.py
```diff
@@ -128,7 +128,7 @@ def sriov_vf_connection_test(
 
     dest_ssh.enable_public_key(source_ssh.generate_key_pairs())
     # generate 200Mb file
-    source_node.execute("dd if=/dev/random of=large_file bs=1M count=200")
+    source_node.execute("dd if=/dev/zero of=large_file bs=1M count=200")
```
squirrelsc: I remembered it's changed to random last time, because of zeros are optimized by some layers, so it's hard to test right performance. What's the reason to change it back?

lisa/tools/firewall.py
```diff
@@ -8,7 +8,7 @@
 class Firewall(Tool):
     @property
     def command(self) -> str:
-        return "ls -lt"
+        return "ls"
```
squirrelsc: use a dummy command like "echo ok".

microsoft/testsuites/network/common.py
LiliDeng: in the ubuntu FIPS kernel, it hung on this command. 
I asked Wei for this, he said the random generated based on Entropy, and Entropy may depend on the interrupt, if the interrupts are not enough, then it is slow to generate the random.

squirrelsc: Is there a way to make it's not random, but not zero?

LiliDeng: I'll figure it out and raise another pr

lisa/tools/firewall.py
LiliDeng: updated

squirrelsc (dismissed):

microsoft/testsuites/network/common.py
squirrelsc: Can you try /dev/urandom?

LiliDeng: based on my search, both random and urandom rely on Entropy, but I will try

LiliDeng:
echo ok doesn't work on freebsd, retry ...
```
2023-02-14 05:32:22.019[188][DEBUG] lisa.env[generated_0].node[0].cmd[4751] cmd: ['sudo', 'sh', '-c', 'command -v echo ok'], cwd: None, shell: True, sudo: True, nohup: False, posix: True, remote: True
2023-02-14 05:32:22.716[3236][DEBUG] lisa.env[generated_0].node[0].cmd[4751].stdout command: wrong number of arguments
```

squirrelsc:
> echo ok doesn't work on freebsd, retry ...
> 
> ```
> 2023-02-14 05:32:22.019[188][DEBUG] lisa.env[generated_0].node[0].cmd[4751] cmd: ['sudo', 'sh', '-c', 'command -v echo ok'], cwd: None, shell: True, sudo: True, nohup: False, posix: True, remote: True
> 2023-02-14 05:32:22.716[3236][DEBUG] lisa.env[generated_0].node[0].cmd[4751].stdout command: wrong number of arguments
> ```

maybe just `echo`. The command doesn't execute actually.

microsoft/testsuites/network/common.py
squirrelsc: It doesn't matter to rely on entropy or not. The problem is, if the entropy is not enough, random will block to wait, and urandom never blocks. urandom reuse the entropy pool.

LiliDeng:
> > echo ok doesn't work on freebsd, retry ...
> > ```
> > 2023-02-14 05:32:22.019[188][DEBUG] lisa.env[generated_0].node[0].cmd[4751] cmd: ['sudo', 'sh', '-c', 'command -v echo ok'], cwd: None, shell: True, sudo: True, nohup: False, posix: True, remote: True
> > 2023-02-14 05:32:22.716[3236][DEBUG] lisa.env[generated_0].node[0].cmd[4751].stdout command: wrong number of arguments
> > ```
> 
> maybe just `echo`. The command doesn't execute actually.

yup. tested on freebsd, good now.

squirrelsc (approved):

PR #2594, opened 2023-02-10
need convert os info vendor for both situation

LiliDeng:


squirrelsc:
Not seeing changes rather than formatting? How the formatting error got merged into main?

PR #2600, opened 2023-02-13
get serial console setting from dmesg on mariner 2 and ubuntu 22.10

LiliDeng:


squirrelsc:
Remove the empty files.

squirrelsc (approved):

PR #2601, opened 2023-02-13
support pidof installation for FreeBSD distro

sharsonia:


lisa/tools/lsblk.py
```diff
@@ -120,6 +121,21 @@ class Lsblk(Tool):
...
 
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, FreeBSD):
+            self.node.os.install_packages("lsblk")
+        elif isinstance(self.node.os, Posix):
+            self.node.os.install_packages("util-linux")
```
LiliDeng, line 130: please move `isinstance(self.node.os, Posix)` into `if`, it will meet most situation.

LiliDeng (approved):

PR #2604, opened 2023-02-15
[improvement] support specify vm name

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1434,14 +1437,20 @@ def initialize_environment(self, environment: Environment, log: Logger) -> None:
         index = 0
         for node in environment.nodes.list():
             node_context = get_node_context(node)
-            vm_name = vms_name_list[index]
-            node_context.vm_name = vm_name
-            if not node.name:
+            if node_context.vm_name:
```
squirrelsc, line 1448: I haven't understood how the different logic works. Please explain in code comments. At the same time, try best to share more logic.

```python
            # what does it mean, if node_context.vm_name presents? What's the difference, if it presents?
            if node_context.vm_name:
                vm_name = node_context.vm_name
                vm = get_vm(self, node)
            else:
                vm_name = vms_name_list[index]
                node_context.vm_name = vm_name
                vm = vms_map[vm_name]
                index = index + 1
            node.name = vm_name
            public_address, private_address = get_primary_ip_addresses(
                self, resource_group_name, vm
            )
```

LiliDeng: updated

squirrelsc: Haven't see it's updated.

LiliDeng: I have updated the code. and node.name = vm_name can't be put after vm = get_vm(self, node), since in get_vm method, it uses node.name

LiliDeng: updated...

squirrelsc (approved):

PR #2608, opened 2023-02-16
Add resize feature for os disk image in CH Platform

smit-gardhariya:
This PR will make changes in CH Platform to add functionality to pass OS disk size from runbook and resize the OS image if so. This change is for guest image we pass to CH Platform.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/ch_platform.py
+++ b/lisa/sut_orchestrator/libvirt/ch_platform.py
@@ -176,6 +176,7 @@ class CloudHypervisorPlatform(BaseLibvirtPlatform):
         self, environment: Environment, log: Logger, node: Node
     ) -> None:
         node_context = get_node_context(node)
+
         if node_context.os_disk_base_file_fmt == DiskImageFormat.QCOW2:
             self.host_node.tools[QemuImg].convert(
                 "qcow2",
@@ -189,6 +190,12 @@ class CloudHypervisorPlatform(BaseLibvirtPlatform):
                 f" {node_context.os_disk_file_path}",
                 expected_exit_code=0,
                 expected_exit_code_failure_message="Failed to copy os disk image",
+            )
+
+        if node_context.os_disk_img_resize_gib:
+            self.host_node.tools[QemuImg].resize(
+                src_file=node_context.os_disk_file_path,
+                size_gib=node_context.os_disk_img_resize_gib,
             )
 
     def _get_vmm_version(self) -> str:
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -447,6 +447,9 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
             node_runbook.disk_img_format
         )
 
+        if node_runbook.disk_img_resize_gib:
+            node_context.os_disk_img_resize_gib = node_runbook.disk_img_resize_gib
+
         node_context.os_disk_file_path = os.path.join(
             self.vm_disks_dir, f"{node_context.vm_name}-os.qcow2"
         )
--- a/lisa/sut_orchestrator/libvirt/qemu_platform.py
+++ b/lisa/sut_orchestrator/libvirt/qemu_platform.py
@@ -42,6 +42,11 @@ class QemuPlatform(BaseLibvirtPlatform):
         self.host_node.tools[QemuImg].create_diff_qcow2(
             node_context.os_disk_file_path, node_context.os_disk_base_file_path
         )
+        if node_context.os_disk_img_resize_gib:
+            self.host_node.tools[QemuImg].resize(
+                src_file=node_context.os_disk_file_path,
+                size_gib=node_context.os_disk_img_resize_gib,
+            )
 
     def _get_vmm_version(self) -> str:
         result = "Unknown"
--- a/lisa/sut_orchestrator/libvirt/schema.py
+++ b/lisa/sut_orchestrator/libvirt/schema.py
@@ -64,6 +64,8 @@ class BaseLibvirtNodeSchema:
     disk_img: str = ""
     # Format of the disk image specified above.
     disk_img_format: str = ""
+    # Pass to resize the guest image os disk
+    disk_img_resize_gib: Optional[int] = None
     # Configuration options for cloud-init.
     cloud_init: Optional[CloudInitSchema] = None
     # Whether to use UEFI or BIOS firmware.
--- a/lisa/tools/qemu_img.py
+++ b/lisa/tools/qemu_img.py
@@ -55,3 +55,15 @@ class QemuImg(Tool):
             expected_exit_code=0,
             expected_exit_code_failure_message="Failed to convert disk image",
         )
+
+    def resize(
+        self,
+        src_file: str,
+        size_gib: int,
+    ) -> None:
+        self.run(
+            f"resize {src_file} {size_gib}G",
+            force_run=True,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Failed to resize disk image",
+        )
--- a/lisa/sut_orchestrator/libvirt/context.py
+++ b/lisa/sut_orchestrator/libvirt/context.py
@@ -37,6 +37,7 @@ class NodeContext:
     os_disk_base_file_path: str = ""
     os_disk_base_file_fmt: DiskImageFormat = DiskImageFormat.QCOW2
     os_disk_file_path: str = ""
+    os_disk_img_resize_gib: Optional[int] = None
     console_log_file_path: str = ""
     extra_cloud_init_user_data: List[Dict[str, Any]] = field(default_factory=list)
     use_bios_firmware: bool = False


lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -77,6 +79,13 @@ class BaseLibvirtNodeSchema:
     machine_type: str = ""
     # Whether to enable secure boot.
     enable_secure_boot: bool = False
+    # Pass disk_size to resize the guest image os disk
+    disk_size: int = field(
```
LiliDeng: disk_size_in_gb?

smit-gardhariya: sure Lili

smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -65,6 +65,9 @@ def _configure_node(
         else:
             node_context.firmware_path = node_runbook.firmware
 
+        if node_runbook.disk_size_in_gb:
+            node_context.os_disk_size_in_gb = node_runbook.disk_size_in_gb
+
```
pupacha: This has to be present in `_configure_node` of `BaseLibvirtPlatform` class. That way disk_size can be used with qemu platform as well.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -77,6 +79,13 @@ class BaseLibvirtNodeSchema:
     machine_type: str = ""
     # Whether to enable secure boot.
     enable_secure_boot: bool = False
+    # Pass disk_size_in_gb to resize the guest image os disk
+    disk_size_in_gb: int = field(
+        default=0,
+        metadata=field_metadata(
+            required=False,
+        ),
+    )
```
pupacha: How about this ?
`disk_size_gb: Optional[int] = None`

pupacha:
@anirudhrb @cwize1 Please review

pupacha (dismissed):

LiliDeng (dismissed):

squirrelsc:
1. Please always @cwize1 for changes on libvirt platforms.
2. Where is the new inherited feature change in libvirt?

squirrelsc (changes requested):
Please wait for @cwize1 review

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -451,6 +451,9 @@ def _configure_node(
             self.vm_disks_dir, f"{node_context.vm_name}-os.qcow2"
         )
 
+        if node_runbook.disk_size_gb:
+            node_context.os_disk_size_in_gb = node_runbook.disk_size_gb
```
squirrelsc: It doesn't need to make a copy in context, unless it needs to change the value. You can get runbook anytime from the node.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -64,6 +64,8 @@ class BaseLibvirtNodeSchema:
     disk_img: str = ""
     # Format of the disk image specified above.
     disk_img_format: str = ""
+    # Pass disk_size_gb to resize the guest image os disk
+    disk_size_gb: Optional[int] = None
```
cwize1: Perhaps `disk_img_resize_gib`?

`disk_img` to match the names of the existing properties. `resize` to make it clear that it is resizing the image. And `gib` to mean Gibibytes (1024^3) instead of gigabytes (1000^3).

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -567,6 +570,25 @@ def _create_node(
                         Path(node_context.os_disk_base_file_path),
                     )
 
+                    runbook: BaseLibvirtNodeSchema = environment.runbook
+                    os_disk_size_in_gb = runbook.disk_size_gb
+
+                    if os_disk_size_in_gb:
+                        qemu_tool = self.host_node.tools[QemuImg]
+                        src_file = node_context.os_disk_base_file_path
+                        qemu_tool.resize(
```
cwize1: It would probably be better to move this logic into `_create_node_os_disk` to keep the original OS disk file unchanged.

Also, I believe `qemu-img resize` works on qcow2 diff disks.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -451,6 +451,9 @@ def _configure_node(
             self.vm_disks_dir, f"{node_context.vm_name}-os.qcow2"
         )
 
+        # if node_runbook.disk_size_gb:
```
cwize1: What's with this commented out code?

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: This is removed now as corresponding code is platform specific as Chirs suggested to push it to _create_node_os_disk(). I used runbook as suggested to get the resize param.

lisa/sut_orchestrator/libvirt/schema.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/platform.py

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Removed. This code i kept as it is to get the resize param from runbook.

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -176,6 +176,22 @@ def _create_node_os_disk(
         self, environment: Environment, log: Logger, node: Node
     ) -> None:
         node_context = get_node_context(node)
+        if environment.runbook.nodes_requirement:
+            node_req: List[schema.NodeSpace] = environment.runbook.nodes_requirement
+            for _, node_space in enumerate(node_req):
+                node_runbook: BaseLibvirtNodeSchema = node_space.get_extended_runbook(
```
cwize1: I believe you can access the runbook via `self.runbook`.

lisa/tools/qemu_img.py
```diff
@@ -55,3 +55,15 @@ def convert(
             expected_exit_code=0,
             expected_exit_code_failure_message="Failed to convert disk image",
         )
+
+    def resize(
+        self,
+        src_file: str,
+        size_gb: int,
```
cwize1: nit: `size_gib`.

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -176,6 +176,22 @@ def _create_node_os_disk(
         self, environment: Environment, log: Logger, node: Node
     ) -> None:
         node_context = get_node_context(node)
+        if environment.runbook.nodes_requirement:
+            node_req: List[schema.NodeSpace] = environment.runbook.nodes_requirement
+            for _, node_space in enumerate(node_req):
```
cwize1: I don't think you should be running the resize operation multiple times.

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -176,6 +176,22 @@ def _create_node_os_disk(
...
+                node_runbook: BaseLibvirtNodeSchema = node_space.get_extended_runbook(
+                    self.node_runbook_type(),
+                    type_name=type(self).type_name(),
+                )
+
+                if node_runbook.disk_img_resize_gib:
+                    qemu_tool = self.host_node.tools[QemuImg]
+                    src_file = node_context.os_disk_base_file_path
+                    qemu_tool.resize(
```
cwize1: The resize operation should be done on the `os_disk_file_path` not the base path.

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -176,6 +176,29 @@ def _create_node_os_disk(
...
+        #                 src_file=src_file,
+        #                 size_gib=node_runbook.disk_img_resize_gib,
+        #             )
+        node_runbook: BaseLibvirtNodeSchema = node.runbook
+        if node_runbook.disk_img_resize_gib:
+            qemu_tool = self.host_node.tools[QemuImg]
+            qemu_tool.resize(
+                src_file=node_context.os_disk_file_path,
+                size_gib=node_runbook.disk_img_resize_gib,
+            )
```
pupacha: You should resize after the below cp or qemu-convert operation. Otherwise `os_disk_file_path` is non-existent path.

lisa/sut_orchestrator/libvirt/ch_platform.py
pupacha: self.runbook will have platform wide runbook. For node (requirement) specific runbook, this is the only way I guess https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/libvirt/platform.py#L381. Perhaps we can add disk size in `NodeContext` to get the disk size here.

lisa/sut_orchestrator/libvirt/platform.py
pupacha: @cwize1 Moving the logic to `_create_node_os_disk ` , for qemu case, will also moodify the original disk file since we use qcow diffs on `os_disk_base_file_path` itself.  https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/libvirt/qemu_platform.py#L42

Having it here right after scp and resizing on the original file will avoid multiple resize operations.

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -176,6 +176,29 @@ def _create_node_os_disk(
         self, environment: Environment, log: Logger, node: Node
     ) -> None:
         node_context = get_node_context(node)
+        # if environment.runbook.nodes_requirement:
```
anirudhrb: Remove commented code.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -64,6 +64,8 @@ class BaseLibvirtNodeSchema:
     disk_img: str = ""
     # Format of the disk image specified above.
     disk_img_format: str = ""
+    # Pass disk_size_gb to resize the guest image os disk
```
anirudhrb: Please update the comment. Looks like the property name has changed since.

anirudhrb (changes requested):

lisa/sut_orchestrator/libvirt/platform.py
cwize1: The code was very explicitly designed to avoid modifying the original file, to ensure tests are repeatable.

Also, for qcow2 files, resize operations are very quick. (I don't know about raw files. But I suspect it is also pretty quick.)

lisa/sut_orchestrator/libvirt/qemu_platform.py
```diff
@@ -39,6 +40,28 @@ def _create_node_os_disk(
...
+        #             qemu_tool = self.host_node.tools[QemuImg]
+        #             src_file = node_context.os_disk_base_file_path
+        #             qemu_tool.resize(
+        #                 src_file=src_file,
+        #                 size_gib=node_runbook.disk_img_resize_gib,
+        #             )
+        node_runbook: BaseLibvirtNodeSchema = node.runbook
+        if node_runbook.disk_img_resize_gib:
+            qemu_tool = self.host_node.tools[QemuImg]
+            qemu_tool.resize(
```
cwize1: This should be moved to be after the `create_diff_qcow2` call (as per my other comment).

cwize1: I would also do the same thing in `CloudHypervisorPlatform`.

cwize1 (changes requested):
.

lisa/sut_orchestrator/libvirt/ch_platform.py
cwize1: Yeah, if it is too difficult to access the runbook directly, then it makes sense to add it to `NodeContext`. (Doing this also makes it consistent with the other properties.)

squirrelsc: NodeContext is for dynamic information, and runbook is for configurations. self.runbook is for the platform, and the node_space is the runbook of node. node_space.get_extended_runbook is to get custom runbook parts for specfic platform. Hope it makes not difficult.

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: Here we will run only for given requirements. It will only run for once if it is the case in runbook.

lisa/tools/qemu_img.py
smit-gardhariya: Will do it

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: Here we are resizing first and then converting.

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: Yes Chi, that is what i am using to get the custom runbook under _create_node_os_disk()

lisa/sut_orchestrator/libvirt/qemu_platform.py
```diff
@@ -39,6 +40,20 @@ def _create_node_os_disk(
...
+                    self.node_runbook_type(),
+                    type_name=type(self).type_name(),
+                )
+                if node_runbook.disk_img_resize_gib:
+                    qemu_tool = self.host_node.tools[QemuImg]
+                    src_file = node_context.os_disk_base_file_path
+                    qemu_tool.resize(
+                        src_file=src_file,
+                        size_gib=node_runbook.disk_img_resize_gib,
+                    )
```
pupacha: We dont need to loop through all nodes in the environment here. `_create_node_os_disk` is specific to the node that is passed as argument. So, just need to get `node_context` and perform the resize operation for that node alone. Applicable for CH platform as well.

lisa/sut_orchestrator/libvirt/qemu_platform.py
```diff
@@ -39,6 +40,20 @@ def _create_node_os_disk(
...
+        if environment.runbook.nodes_requirement:
+            node_req: List[NodeSpace] = environment.runbook.nodes_requirement
+            for _, node_space in enumerate(node_req):
+                node_runbook: BaseLibvirtNodeSchema = node_space.get_extended_runbook(
+                    self.node_runbook_type(),
+                    type_name=type(self).type_name(),
+                )
+                if node_runbook.disk_img_resize_gib:
+                    qemu_tool = self.host_node.tools[QemuImg]
+                    src_file = node_context.os_disk_base_file_path
```
pupacha: Also, as @cwize1 mentioned earlier, let's not resize the `os_disk_base_file_path`. 
diffs do not support the resize, so we will need to copy the `os_disk_base_file_path` to `os_disk_file_path` (just how we do in ch platform) and resize.
@anirudhrb  any comments on this ?

pupacha (changes requested):

lisa/sut_orchestrator/libvirt/ch_platform.py
cwize1: I don't understand what you are saying. `NodeSpace` is only really used for provisioning the node. It shouldn't be used here. And it most definitely should not be used in a loop per node.

lisa/sut_orchestrator/libvirt/ch_platform.py
cwize1: While `NodeContext` is technically for dynamic information, please just use it anyway. This code is a mess. It is a lot simpler to just copy the relevant information into `NodeContext` during `_configure_node`.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Moved resize to os_disk_file_path.

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: Used NodeContext to get resize parameter

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: Using os_disk_file_path now.

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: using os_disk_file_path in _create_node_os_disk() after conver/copy under CH platform and after create_diff_qcow2() for QEMU platform

lisa/sut_orchestrator/libvirt/schema.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/qemu_platform.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/qemu_platform.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/qemu_platform.py
smit-gardhariya: resize is on os_disk_file_path.

smit-gardhariya: Resize worked on diff image as well. I am resolving this comment

lisa/sut_orchestrator/libvirt/ch_platform.py
smit-gardhariya: i used node_context  now as you suggested.

pupacha (approved):

cwize1 (approved):

squirrelsc (approved):

PR #2611, opened 2023-02-17
Revert to unattended-upgrade

kamalca:
There is a difference in behavior between unattended-upgrade and running apt upgrade. For this reason, I am switching this back to using unattended upgrade.
Since unattended-upgrade doesn't always upgrade all packages, I need to remove the packages that don't get upgraded from the return value.


kamalca:
@squirrelsc Ready for review

squirrelsc (approved):

PR #2612, opened 2023-02-18
[improvement] download udev rules to construct a set of symbolic link to data disks

LiliDeng:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1089,31 +1091,58 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._initialize_information(self._node)
 
     def get_raw_data_disks(self) -> List[str]:
-        pattern = re.compile(r"/dev/disk/azure/scsi[0-9]/lun[0-9][0-9]?", re.M)
+        pattern = re.compile(r"/dev/disk/azure/scsi[0-9]/lun[0-9][0-9]?$", re.M)
```
squirrelsc: add examples above, and it's better to put static patterns on class level.

squirrelsc (approved):

PR #2613, opened 2023-02-20
Add azcopy_path in AzurePlatformSchema and use azcopy to copy blob

lubaihua33:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1049,15 +1051,47 @@ def wait_copy_blob(
...
 
+def copy_blob_by_azcopy(
+    azcopy_path: str,
+    src: str,
+    dst: str,
+    log: Logger,
+    timeout: int = 60 * 60,
+) -> None:
+    try:
+        subprocess.check_output(
```
squirrelsc: use local node, so the log can be included in normal format.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -893,6 +894,7 @@ def generate_sas_token(
     account_name: str,
     resource_group_name: str,
     expired_hours: int = 2,
+    write: bool = False,
```
squirrelsc: `writable`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1921,19 +1934,51 @@ def _get_deployable_vhd_path(
...
+                            "Then no need to check the hash key"
+                        )
+                        vhd_exists = True
 
             if not vhd_exists:
-                blob_client.start_copy_from_url(
-                    original_vhd_path, metadata=None, incremental_copy=False
-                )
+                azcopy_path = self._azure_runbook.azcopy_path
+                log.debug(f"AzCopy path:{azcopy_path}")
```
squirrelsc: This line should be printed, only azcopy_path specifies.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1921,19 +1934,51 @@ def _get_deployable_vhd_path(
...
+                if azcopy_path and os.path.exists(azcopy_path):
```
squirrelsc: If the path specified, but not exists, it should raise an exception to make it obvious.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1921,19 +1934,51 @@ def _get_deployable_vhd_path(
...
+                        )
+                    except Exception as identifier:
+                        blob_client.delete_blob(delete_snapshots="include")
+                        raise LisaException(f"{identifier}")
+
+                    # Set metadata to mark the blob copied by AzCopy successfully
+                    metadata = {"AzCopyStatus": "Success"}
+                    blob_client.set_blob_metadata(metadata)
+                    log.info("vhd copied")
+                    return full_vhd_path
```
squirrelsc: Why return here, not at the end?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1921,19 +1934,51 @@ def _get_deployable_vhd_path(
...
+                else:
+                    blob_client.start_copy_from_url(
+                        src_vhd_sas_url, metadata=None, incremental_copy=False
+                    )
 
-            wait_copy_blob(blob_client, vhd_path, log)
+            wait_copy_blob(blob_client, dst_vhd_name, log)
```
squirrelsc, line 1985: If this line applies to Python SDK copy only, it should be in else part.

lubaihua33: This line also applies to AzCopy when there is another copy operation in parallel.

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: The copy operation with AzCopy is synchronous. We don't need to call wait_copy_blob. So return here.
wait_copy_blob needs to be called when there is another copy operation in parallel.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: Have changed it to log.info().
I didn't get the point of  "only azcopy_path specifies." this comment. Change to log.info(f"{azcopy_path}") ?

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/common.py
lubaihua33: Used local().execute() and deleted the new added function copy_blob_by_azcopy

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: This log line should be printed, only the azcopy_path variable has value. If it's empty, it doesn't need to print empty string.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: It still can reach the wait_copy_blob to make the logic consistent, and do an extra validation.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py

squirrelsc (approved):

squirrelsc:
Please merge it next Monday after released.

PR #2615, opened 2023-02-21
Calculate the crashkernel size for different distros

lubaihua33:
1. Calculate the crashkernel size for different distros. Use the default settings for Ubuntu.

### DIFF
--- a/lisa/tools/kdump.py
+++ b/lisa/tools/kdump.py
@@ -223,6 +223,34 @@ class KdumpBase(Tool):
                     "The kernel config {config} is not set. Kdump is not supported."
                 )
 
+    def calculate_crashkernel_size(self, total_memory: str) -> str:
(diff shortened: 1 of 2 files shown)


lisa/features/serial_console.py
```diff
@@ -183,6 +187,24 @@ def check_initramfs(
                 f"{initramfs_logs} {filesystem_exception_logs}"
             )
 
+    def check_ata_exception(
```
squirrelsc: Why will it be listed separately? Add code comments as well.

lubaihua33: When the image is ubuntu 22.04 and vm size is A1_V2, I found there are no 'initramfs' in the console log when having the same issue. So I wanted to add a new check. 

After consideration, adding ata exception check is not a good way because this is not a common check. It should not appear in the case. I have reverted it. Let me post another PR.

LiliDeng (approved):

PR #2619, opened 2023-02-22
Changes for KernelCI

adityagesh:
- Update make to return ExecutableResult
- Add function to generate sas token for blob instead of using container wide sas token
- Update ansi filter
    Example:
      Text: "\x1b[?1h\x1b=\rAdd linux-next specific files for 20230221\x1b[m\r\n\r\x1b[K\x1b[?1l\x1b>"
      Filter Result: '\x1b=\rAdd linux-next specific files for 20230221\r\n\r\x1b>'
      After change: '\rAdd linux-next specific files for 20230221\r\n\r'

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -39,6 +39,7 @@ from azure.mgmt.storage.models import (  # type: ignore
 )
 from azure.storage.blob import (
     AccountSasPermissions,
+    BlobSasPermissions,
     BlobServiceClient,
     ContainerClient,
     ResourceTypes,
@@ -885,6 +886,33 @@ def get_storage_credential(
         account_name=account_name, resource_group_name=resource_group_name
     ).keys[0]
     return {"account_name": account_name, "account_key": key.value}
+
+
+def generate_blob_sas_token(
+    credential: Any,
+    subscription_id: str,
+    account_name: str,
+    resource_group_name: str,
+    container_name: str,
+    file_name: str,
+    expired_hours: int = 2,
+) -> Any:
+    shared_key_credential = get_storage_credential(
+        credential=credential,
+        subscription_id=subscription_id,
+        account_name=account_name,
+        resource_group_name=resource_group_name,
+    )
+
+    sas_token = generate_blob_sas_token(
+        account_name=shared_key_credential["account_name"],
+        account_key=shared_key_credential["account_key"],
+        container_name=container_name,
+        blob_name=file_name,
+        permission=BlobSasPermissions(read=True),  # type: ignore
+        expiry=datetime.utcnow() + timedelta(hours=expired_hours),
+    )
+    return sas_token
 
 
 def generate_sas_token(
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -44,6 +44,7 @@ from azure.storage.blob import (
     ContainerClient,
     ResourceTypes,
     generate_account_sas,
+    generate_blob_sas,
 )
 from azure.storage.fileshare import ShareServiceClient  # type: ignore
 from dataclasses_json import dataclass_json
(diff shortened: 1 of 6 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -885,6 +887,33 @@ def get_storage_credential(
     return {"account_name": account_name, "account_key": key.value}
 
 
+def generate_blob_sas_token(
+    credential: Any,
+    subscription_id: str,
```
LiliDeng, line 894: I didn't see the method was used

adityagesh: I have created a kci_kernel_source_installer in internal repo. This method is used in that transformer

lisa/util/__init__.py
```diff
@@ -53,7 +53,7 @@
 
 
 # used to filter ansi escapes for better layout in log and other place
-__ansi_escape = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
+__ansi_escape = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~]|[=<>a-kzNM78])")
```
squirrelsc: What's the reason to change it? It may impact widely.

lisa/tools/make.py
```diff
@@ -84,7 +85,7 @@ def make(
             )
 
         # yes '' answers all questions with default value.
-        self.node.execute(
+        result = self.node.execute(
```
squirrelsc, line 88: When is the result needed? We should handle the common checks here.

lisa/util/__init__.py
adityagesh: It does not filter "\x1b=............\x1b"
Example:
Text: "\x1b[?1h\x1b=\rAdd linux-next specific files for 20230221\x1b[m\r\n\r\x1b[K\x1b[?1l\x1b>"
Filter Result: '\x1b=\rAdd linux-next specific files for 20230221\r\n\r\x1b>'
After change: '\rAdd linux-next specific files for 20230221\r\n\r'

Ref: https://stackoverflow.com/a/45448194

lisa/tools/make.py
adityagesh: Wanted the result to get the stdout and stderr of make command

lisa/util/__init__.py
squirrelsc: It can be like below. And add above example, so it's easy to prevent regression next time.

\x1b(?:[@-Z\\-_=<>a-kzNM78]|\[[0-?]*[ -/]*[@-~])

lisa/tools/make.py
squirrelsc: What's extra information of stdout and stderr is needed?

adityagesh: In my case, the caller (kernel_source_installer) reads the outputs (stdout and stderr)

lisa/util/__init__.py
adityagesh: Did you mean the below?
"\x1B(?:[@-Z\\-_=<>a-kzNM78]|[[0-?]*[ -/]*[@-~])"

The regex "\x1b(?:[@-Z\-_=<>a-kzNM78]|[[0-?][ -/][@-~])" is not filtering correctly.

adityagesh: oh, okay I understand, Github is removing double \ and * from the comments 😄

lisa/tools/make.py
squirrelsc: I understand some place to read it. But we need to think about why it's read, and what information is needed. Can it be parsed to structure information and then return.

adityagesh: At present It is being written to a file and uploaded as blob

squirrelsc: Is it a requirement to upload raw build output?

adityagesh: Yes, the build outputs are required for the community to debug build failures

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2620, opened 2023-02-22
check lagscope server port open or not

LiliDeng:


lisa/tools/lsof.py
```diff
@@ -18,6 +18,11 @@ def is_port_opened(self, port: int) -> bool:
         result = self.run(cmd, force_run=True, shell=True, sudo=True)
         return result.exit_code == 0
 
+    def is_port_opened_per_process_name(self, process_name: str) -> bool:
```
squirrelsc: I think it returns not only ports, also other file handlers. Can you confirm it?

LiliDeng: update.
```
lisa.env[generated_0].node[1].cmd[4596] cmd: ['sudo', 'sh', '-c', 'lsof -iTCP -P -n | grep -i lagscope'], cwd: None, shell: True, sudo: True, nohup: False, posix: True, remote: True
lisa.env[generated_0].node[1].cmd[4596].stdout lagscope  11443            root    3u  IPv4  41189      0t0  TCP 10.0.0.4:6001 (LISTEN)
```

squirrelsc (approved):

PR #2621, opened 2023-02-22
waagent.conf will be read once when check resource disk, add force_run to get the configuration

LiliDeng:
it happened after code change to use the resource disk for downloading the large size packages, and only for centos type os, it will read resource disk mount point from waagent.conf file.

### DIFF
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -73,14 +73,14 @@ class Waagent(Tool):
         # self.run("-deprovision+user --force", sudo=True)
         self.run("-deprovision --force", sudo=True, expected_exit_code=0)
 
-    def get_configuration(self) -> Dict[str, str]:
+    def get_configuration(self, force_run: bool = False) -> Dict[str, str]:
         if isinstance(self.node.os, CoreOs):
             waagent_conf_file = "/usr/share/oem/waagent.conf"
         else:
             waagent_conf_file = "/etc/waagent.conf"
 
         config = {}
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/tools.py
```diff
@@ -108,8 +108,8 @@ def is_swap_enabled(self) -> bool:
                 f"Unknown value for ResourceDisk.EnableSwap : {is_swap_enabled}"
             )
 
-    def is_rdma_enabled(self) -> bool:
-        waagent_configuration = self.get_configuration()
+    def is_rdma_enabled(self, force_run: bool = False) -> bool:
```
squirrelsc: Because all caller of `is_rdma_enabled` need `force_run=True`. It doesn't need to expose force_run here, just set it to True when calling get_configuration.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -73,31 +73,31 @@ def deprovision(self) -> None:
...
         for line in cfg.splitlines():
             matched = self._key_value_regex.fullmatch(line)
             if matched:
                 config[matched.group("key")] = matched.group("value")
 
         return config
 
-    def get_root_device_timeout(self) -> int:
-        waagent_configuration = self.get_configuration()
+    def get_root_device_timeout(self, force_run: bool = False) -> int:
```
squirrelsc: Those methods are not changed this time, let's keep them unchanged.

lisa/sut_orchestrator/azure/tools.py

lisa/sut_orchestrator/azure/tools.py

squirrelsc (approved):

PR #2623, opened 2023-02-22
Increase timeout for LTP tests

dsrivastavv:
Noticed tests timing out on Mariner and Debian distros


microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -28,7 +28,7 @@
     """,
 )
 class LtpTestsuite(TestSuite):
-    _TIME_OUT = 12000
+    _TIME_OUT = 18000
```
LiliDeng, line 31: after extend timeout, does it work?

LiliDeng (approved):

PR #2625, opened 2023-02-23
Set rust default toolchain to run cargo

smit-gardhariya:
These changes will set the default toolchain for rust setup and point stable toolchain for cargo binary to run the LISA tool. We will also run **cargo test** with verbose for better log.

This will resolve issue of **"rustup could not choose a version of cargo to run, because one wasn't specified explicitly, and no default is configured."**  This is for testcase **verify_rust_vmm_mshv_tests**

### DIFF
--- a/lisa/tools/cargo.py
+++ b/lisa/tools/cargo.py
@@ -1,7 +1,9 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+import os
+import re
 from pathlib import PurePath
-from typing import List, Optional, Type, cast
+from typing import Any, List, Optional, Type, cast
 
 from lisa.executable import Tool
 from lisa.operating_system import CBLMariner, Posix, Ubuntu
@@ -9,14 +11,28 @@ from lisa.tools.curl import Curl
 from lisa.tools.echo import Echo
 from lisa.tools.gcc import Gcc
 from lisa.tools.ln import Ln
-from lisa.util import UnsupportedDistroException
+from lisa.tools.rm import Rm
+from lisa.util import LisaException, UnsupportedDistroException
 from lisa.util.process import ExecutableResult
 
 
 class Cargo(Tool):
+    # cargo 1.67.1 (8ecd4f20a 2023-01-10)
+    _version_pattern = re.compile(
+        r"cargo (?P<major>\d+).(?P<minor>(\d+)).(?P<patch>(\d+)) ",
+        re.M,
+    )
+
+    # stable-x86_64-unknown-linux-gnu (default)
+    _rust_toolchain_pattern = re.compile(
+        r"\w+-\w+-\w+-\w+-\w+",
+        re.M,
+    )
+    toolchain: str = ""
+
     @property
     def command(self) -> str:
-        return "cargo"
+        return self._command
 
     @property
     def can_install(self) -> bool:
@@ -26,10 +42,13 @@ class Cargo(Tool):
     def dependencies(self) -> List[Type[Tool]]:
         return [Gcc]
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._command = "cargo"
+
     def _install(self) -> bool:
-        os = self.node.os
+        node_os = self.node.os
         cargo_source_url = "https://sh.rustup.rs"
-        if isinstance(os, CBLMariner) or isinstance(os, Ubuntu):
+        if isinstance(node_os, CBLMariner) or isinstance(node_os, Ubuntu):
             self.__install_dependencies()
 
             # install cargo/rust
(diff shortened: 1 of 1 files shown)


lisa/tools/cargo.py
```diff
@@ -9,14 +11,29 @@
...
+    _rust_toolchain_pattern = re.compile(
+        r"\w+-\w+-\w+-\w+-\w+",
+        re.M,
+    )
+    toolchain: str = ""
+
     @property
     def command(self) -> str:
-        return "cargo"
+        return self.cargo_command
```
LiliDeng: use return "cargo" here, in other place use self.command

lisa/tools/cargo.py
```diff
@@ -56,15 +73,42 @@ def _install(self) -> bool:
...
+            )
+
+        self.node.tools[Rm].remove_file(
+            "/usr/local/bin/cargo",
+            sudo=True,
+        )
+        self.node.tools[Rm].remove_file(
+            "/usr/local/bin/rustup",
+            sudo=True,
+        )
```
LiliDeng, line 101: why remove them?

smit-gardhariya: Once it is installed, we will point to PATH of toolchain so we dont need them. This was created while installing so we can pass **_check_exist()** and we can perform **rustup default stable** to set the rust default toolchain.

lisa/tools/cargo.py
smit-gardhariya: sure Lili , i will check on it.

smit-gardhariya: Implemented. Please review

lisa/tools/cargo.py
```diff
@@ -56,15 +77,42 @@ def _install(self) -> bool:
...
+        if is_installed:
+            self.node.execute("rustup default stable", shell=True)
+            self.toolchain = self.__get_rust_toolchain()
+            self._log.debug(f"Rust toolchain: {self.toolchain}")
+            self.command = "".join(
+                [
+                    f"{home_dir}/.rustup/toolchains/",
+                    f"{self.toolchain}/bin/cargo",
+                ]
+            )
```
anirudhrb: `join` is not needed. Python will concatenate the strings if you do this:
```suggestion
            self.command = (
                    f"{home_dir}/.rustup/toolchains/"
                    f"{self. Toolchain}/bin/cargo"
            )
```

lisa/tools/cargo.py
anirudhrb: What about the symlink for cargo? Do we even need to create it in the first place?

lisa/tools/cargo.py
```diff
@@ -110,23 +156,44 @@ def test(
...
-        home_path = echo.run(
-            "$HOME",
-            shell=True,
-            expected_exit_code=0,
-            expected_exit_code_failure_message="failure to grab $HOME path",
-        ).stdout
-        new_path = f"{home_path}/.cargo/bin:{original_path}"
+
+        if os.path.dirname(self.command) not in path:
+            path = f"{os.path.dirname(self.command)}:{path}"
```
anirudhrb: Why do we need this? `self.command` contains the absolute path to the binary right?

anirudhrb (changes requested):

lisa/tools/cargo.py
smit-gardhariya: Yes @anirudhrb it is needed as _check_exist() wont take environment variable like PATH as argument and **.cargo** and **.rustup** is getting install under home directory. If we will not create this symlink, _check_exist() will fail. This is what i have faced during the initial development of testcase.

lisa/tools/cargo.py
smit-gardhariya: @anirudhrb i have done it earlier that way only but flak8 validation is having issue with this way of string. It would say **BLK100 would make change** at this line. That is why i did in this way.

lisa/tools/cargo.py
smit-gardhariya: @anirudhrb  self.command will have absolute path but there are rustc binary that need to be there in PATH of cargo we are using for given toolchain otherwise it will throw error something like **not able to detect rustc version**. You can check for different pipeline run instance for that error.

lisa/tools/cargo.py
anirudhrb: What happens when you try to autocorrect that error?

smit-gardhariya: @anirudhrb  It wont show any error with flak8 validation in code editor but with pipeline it does. It wont mention anything in detail with pipeline instance, just **BLK100 would make change**

smit-gardhariya: @anirudhrb As we discussed, i ran the nox manually with python 3.8 locally and it is passing with flak8. It is failing with pipeline instance only when PR branch push. Please find below log.

nox > Session black was successful.
nox > Running session isort
nox > Creating virtual environment (virtualenv) using python.exe in .nox\isort
nox > python -m pip install 'isort ~= 5.12.0'
nox > isort .
Skipped 5 files
nox > Session isort was successful.
nox > Running session flake8
nox > Creating virtual environment (virtualenv) using python.exe in .nox\flake8
nox > python -m pip install 'black ~= 23.1.0' 'flake8 ~= 6.0.0' Flake8-pyproject 'flake8-black >= 0.3.2' 'flake8-bugbear ~= 23.2.13' 'flake8-isort >= 4.1.1' 'pep8-naming >= 0.12.1' 'isort ~= 5.12.0'
nox > flake8
**nox > Session flake8 was successful.**

anirudhrb: My guess is that black is complaining about some small formatting issue and not about the whole idea of using implicit f-string continuation. I believe there are other places in the code that do the same thing. You should try to get to the bottom of this.

This is not a blocker though.

smit-gardhariya: I removed join and used f-string. Earlier it was character length limit issue which leads to this. It is resolved now.

anirudhrb (dismissed):

lisa/tools/cargo.py
```diff
@@ -9,14 +11,33 @@
...
+        r"\w+-\w+-\w+-\w+-\w+",
+        re.M,
+    )
+    toolchain: str = ""
+    __command: str = "cargo"
+
     @property
     def command(self) -> str:
-        return "cargo"
+        return self.__command
```
LiliDeng: have you tried to use self.command?

smit-gardhariya: Hi Lili, yes i tried with setter of command property. Now i have used _initialize() to set self._command as we discussed. Please check and let me know.

LiliDeng (approved):

anirudhrb (approved):

PR #2628, opened 2023-02-23
Disk Controller Type

kamalca:
This enables NVMe OS disk (ASAP)

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -417,6 +417,17 @@ disk_type_priority: List[DiskType] = [
 ]
 
 
+class DiskControllerType(str, Enum):
+    SCSI = "SCSI"
+    NVME = "NVMe"
+
+
+disk_controller_type_priority: List[DiskControllerType] = [
+    DiskControllerType.SCSI,
+    DiskControllerType.NVME,
+]
+
+
 @dataclass_json()
 @dataclass()
 class DiskOptionSettings(FeatureSettings):
@@ -471,6 +482,19 @@ class DiskOptionSettings(FeatureSettings):
             allow_none=True, decoder=search_space.decode_count_space
         ),
     )
+    disk_controller_type: Optional[
+        Union[search_space.SetSpace[DiskControllerType], DiskControllerType]
+    ] = field(  # type:ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            items=[DiskControllerType.SCSI, DiskControllerType.NVME],
+        ),
+        metadata=field_metadata(
+            decoder=partial(
+                search_space.decode_set_space_by_type, base_type=DiskControllerType
+            )
+        ),
+    )
 
     def __eq__(self, o: object) -> bool:
         if not super().__eq__(o):
@@ -485,6 +509,7 @@ class DiskOptionSettings(FeatureSettings):
             and self.data_disk_iops == o.data_disk_iops
             and self.data_disk_size == o.data_disk_size
             and self.max_data_disk_count == o.max_data_disk_count
+            and self.disk_controller_type == o.disk_controller_type
         )
 
     def __repr__(self) -> str:
@@ -494,7 +519,8 @@ class DiskOptionSettings(FeatureSettings):
             f"caching: {self.data_disk_caching_type},"
             f"iops: {self.data_disk_iops},"
             f"size: {self.data_disk_size},"
-            f"max_data_disk_count: {self.max_data_disk_count}"
+            f"max_data_disk_count: {self.max_data_disk_count},"
+            f"disk_controller_type: {self.disk_controller_type}"
         )
 
     def __str__(self) -> str:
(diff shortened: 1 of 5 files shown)


lisa/schema.py
```diff
@@ -471,6 +482,21 @@ class DiskOptionSettings(FeatureSettings):
             allow_none=True, decoder=search_space.decode_count_space
         ),
     )
+    disk_controller_type: Optional[
+        Union[search_space.SetSpace[DiskControllerType], DiskControllerType]
+    ] = field(  # type:ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            items=[
+                DiskControllerType.SCSI,
```
squirrelsc: Put both in items, so it's by default enabled for all.

squirrelsc:
@LiliDeng This PR looks good to me, please double check and merge it after next release.

lisa/schema.py
kamalca: I was a bit confused about the use-case for this default. I thought it was the default for the runbook, in which case by default you want SCSI. For default of VM capability, I thought that was the purpose of the generate_maximum_capability function?

Any guidance?

kamalca: Fixed

squirrelsc: If the default value has only SCSI, it may not be able to deploy with NVMe by default. If put both of them here, as the priority list, SCSI will be default.

LiliDeng:
@kamalca it will break run against gen1 image.

```

deployment failed. LisaException: InvalidParameter: Disk Controller Type is only supported for VMs that are created from a Generation 2 Image or OS disk and VM Size that supports the same, refer to https://docs.microsoft.com/en-us/azure/virtual-machines/generation-2#creating-a-generation-2-vm for more details about creating Generation 2 VM.

```

kamalca:
I fixed the regression with gen1 images. But there is no check for if the runbook specifies `disk_controller_type: NVMe' and the hyperv_generation is V1. Instead the VM will be deployed with SCSI and no error will be given.
@squirrelsc Where should I add the check for hyperv_generation for the disk_controller_type?

squirrelsc:
> I fixed the regression with gen1 images. But there is no check for if the runbook specifies `disk_controller_type: NVMe' and the hyperv_generation is V1. Instead the VM will be deployed with SCSI and no error will be given. @squirrelsc Where should I add the check for hyperv_generation for the disk_controller_type?

In this case, what's set in arm template? Gen1 and NVMe together? Or LISA will switch to SCSI silently?

kamalca:
@squirrelsc Latest commit has changes to assert that NVMe is not set with gen 1

To answer your previous question, the behavior without the assert is that the ARM template will silently deploy all gen1 images with SCSI regardless of disk_controller_type in the ARM parameters. Now that I have added an assert, it will raise an error if NVMe is set on gen1 image.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1273,6 +1273,10 @@ def _create_node_arm_parameters(
         assert isinstance(
             capability.disk.disk_controller_type, schema.DiskControllerType
         )
+        assert (
```
squirrelsc, line 1283: Can it be check in Azure Disk feature?

kamalca: @squirrelsc None of the other functions made sense to put this check in. Do you have a suggestion?

squirrelsc: How about `check` in Azure Disk options? It won't bring an exception, but skip one or all test cases.

kamalca: The check function doesn't have access to the right information, right? Does it have access to the generation and the final disk_controller_type?

squirrelsc: I see, it looks cross feature validation have to be here. I'm ok now.

squirrelsc:
@LiliDeng LGTM. Let's try after next release.

LiliDeng:
@kamalca please try gen2 image + westus3 location, I got below issue.
```
deployment failed. LisaException: BadRequest: Disk Controller Type is not supported in location WestUS3.
```

kamalca:
@LiliDeng Can you please send logs for this run?
I do not see any SKUs in westus3 that have disk_controller_type = NVMe. When I ran smoke_test in westus3 with disk_controller_type: NVMe, the test skipped. What settings are you using that tried to force run with NVMe?

kamalca:
@LiliDeng @squirrelsc Fixed now. I changed the deployment template so that the DiskControllerType is only added if it is not equal to NVMe. Since no VM sizes in westus3 (or other unsupported regions) should have the DiskControllerType tag, this should not cause any more problems.

LiliDeng (approved):

PR #2629, opened 2023-02-24
fix repo not available issue for L2 vm

LiliDeng:
in our made qcow2 image, the repo is `sg.archive.ubuntu.com`, sometimes, it is unavailable. so, replace it into `archive.ubuntu.com`


LiliDeng:
the repo is back, I have added a failure type for this, close this PR, since it is an ugly fix.

PR #2637, opened 2023-03-02
Enable running kernel self tests automatically

niyelchu:
This PR enables running kernel self tests to help automation scenarios for kernel testing. Currently, only Mariner VMs are supported for running kselftests automatically using LISA.

The PR handles two scenarios:

1) When the user provides a tarball of kselftest binaries in a .yml file. For example, a new variable kselftest_file_path can be added in microsoft/runbook/azure.yml:

  _- name: kselftest_file_path
    value: <path_to_kselftests.tar.xz>
    is_case_visible: true_

Kselftest tool extracts the tar, executes the script run_kselftest.sh and redirects test results to a file kselftest-results.txt on both local and remote machine. 

2) When the user sets kselftest_file_path to "" (empty) in a .yml file or if kselftest_file_path variable is not present in a .yml file. 

Example:

  _- name: kselftest_file_path
    value: ""
    is_case_visible: true_

Kselftest tool clones mariner kernel, copies current config into .config, builds kselftests and generates a tarball of kselftest binaries. The tool further extracts the tar, executes the script run_kselftest.sh and redirects test results to a file kselftest-results.txt on both local and remote machine.

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest-suite.py
+++ b/microsoft/testsuites/kselftest/kselftest-suite.py
@@ -1,0 +1,59 @@
+from typing import Any, Dict
+
+from lisa import Environment, Node, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa.testsuite import TestResult
+from lisa.util import SkippedException, UnsupportedDistroException
+from microsoft.testsuites.kselftest.kselftest import Kselftest
+
+
+@TestSuiteMetadata(
+    area="kselftest",
+    category="community",
+    description="""
+    This test suite is used to run kselftests.
+    """,
+)
+class KselftestTestsuite(TestSuite):
+    # kselftests take about an hour to complete, timeout below
+    # is in seconds and set to 1.5x an hour i.e. 90mins
+    _TIME_OUT = 5400
+
+    @TestCaseMetadata(
+        description="""
+        This test case runs linux kernel self tests on Mariner VMs.
+        Cases:
+        1. When a tarball is specified in .yml file, extract the tar and run kselftests.
+        Example:
+        - name: kselftest_file_path
+          value: <path_to_kselftests.tar.xz>
+          is_case_visible: true
+        2. When a tarball is not specified in .yml file, clone Mariner kernel,
+        copy current config to .config, build kselftests and generate a tar.
+
+        For both cases, verify that the kselftest tool extracts the tar, runs the script
+        run_kselftest.sh and redirects test results to a file kselftest-results.txt.
+        """,
+        priority=3,
+        timeout=_TIME_OUT,
+    )
+    def verify_kselftest(
+        self,
+        node: Node,
+        environment: Environment,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        file_path = variables.get("kselftest_file_path", "")
+        try:
+            kselftest: Kselftest = node.tools.get(
+                Kselftest,
+                kselftest_file_path=file_path,
+            )
+            kselftest.run_all(
+                result,
+                environment,
+                log_path,
+            )
+        except UnsupportedDistroException as identifier:
+            raise SkippedException(identifier)
--- a/microsoft/testsuites/kselftest/kselftest.py
+++ b/microsoft/testsuites/kselftest/kselftest.py
@@ -1,0 +1,272 @@
+import os
+import re
+from dataclasses import dataclass
+from pathlib import PurePath, PurePosixPath
+from typing import Any, Dict, List
+
+from assertpy import assert_that
+
+from lisa import Environment, notifier
+from lisa.base_tools.uname import Uname
+from lisa.executable import Tool
+from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
+from lisa.node import Node
+from lisa.operating_system import CBLMariner
+from lisa.testsuite import TestResult
+from lisa.tools import Cp, Git, Ls, Make, RemoteCopy, Tar
+from lisa.tools.whoami import Whoami
+from lisa.util import LisaException, UnsupportedDistroException, find_groups_in_lines
+
+
+@dataclass
+class KselftestResult:
+    name: str = ""
+    status: TestStatus = TestStatus.QUEUED
+    exit_value: int = 0
+
+
+class Kselftest(Tool):
+    _MARINER_KERNEL_SRC_REPO = "https://github.com/microsoft/CBL-Mariner-Linux-Kernel"
+    _KERNEL_REPO_NAME = "CBL-Mariner-Linux-Kernel"
+    _KSELFTEST_TAR_PATH = (
+        "/build/kselftest/kselftest_install/kselftest-packages/kselftest.tar.gz"
+    )
+
+    # kselftest result log has "ok" and "not ok" prefixes, use regex to filter them
+    # example skip test log: "ok 6 selftests: cgroup: test_stress.sh # SKIP"
+    # example failed test log: "not ok 52 selftests: net: veth.sh # exit=1"
+    # example passed test log: "ok 2 selftests: breakpoints: breakpoint_test"
+    # example timeout test log: "not ok 8 selftests: netfilter: nft_concat_range.sh
+    # # TIMEOUT 45 seconds"
+    _RESULT_KSELFTEST_OK_REGEX = re.compile(
+        r"^(?P<status>(not ok|ok))\s+\d+\s+selftests:\s+\S+:\s+(?P<name>\S+)\s*(?:# (?:exit=)?(?P<reason>SKIP|TIMEOUT\d+).*)?(?:# exit=(?P<exit>\d+))?"  # noqa: E501
+    )
+
+    @property
+    def command(self) -> str:
+        return str(self._command)
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _check_exists(self) -> bool:
+        return self.node.tools[Ls].path_exists(str(self._remote_tar_path), sudo=True)
+
+    def __init__(
+        self, node: Node, kselftest_file_path: str, *args: Any, **kwargs: Any
+    ) -> None:
+        super().__init__(node, *args, **kwargs)
+
+        # tar file path specified in yml
+        self._tar_file_path = kselftest_file_path
+        if self._tar_file_path:
+            self._remote_tar_path = self.get_tool_path(
+                use_global=True
+            ) / os.path.basename(self._tar_file_path)
+        else:
+            self._remote_tar_path = (
+                self.get_tool_path(use_global=True)
+                / f"{self._KERNEL_REPO_NAME}{self._KSELFTEST_TAR_PATH}"
+            )
+
+        # command to run kselftests
+        if not self._tar_file_path:
+            self._command = self.get_tool_path(use_global=True) / "run_kselftest.sh"
+        else:
+            self._command = (
+                self.get_tool_path(use_global=True)
+                / "kselftest-packages/run_kselftest.sh"
+            )
+
+    # install common dependencies
+    def _install(self) -> bool:
+        if not isinstance(self.node.os, CBLMariner):
+            raise UnsupportedDistroException(
+                self.node.os, "kselftests are supported on Mariner VMs only."
+            )
+
+        if self._tar_file_path:
+            self.node.shell.copy(PurePath(self._tar_file_path), self._remote_tar_path)
+        else:
+            # clone kernel, build kernel, then build kselftests
+            self.node.os.install_packages(
+                [
+                    "bison",
+                    "flex",
+                    "build-essential",
+                    "openssl-devel",
+                    "bc",
+                    "dwarves",
+                    "rsync",
+                    "libcap-devel",
+                    "libcap-ng-devel",
+                    "fuse",
+                    "fuse-devel",
+                    "popt-devel",
+                    "numactl-devel",
+                    "libmnl-devel",
+                    "libinput",
+                    "mesa-libgbm-devel",
+                    "glibc-static",
+                    "clang",
+                ]
+            )
+
+            uname = self.node.tools[Uname]
+            uname_result = uname.get_linux_information(force_run=False)
+            version = uname_result.kernel_version
+            git = self.node.tools[Git]
+
+            # If version.patch is zero, clone major.minor kernel from
+            # upstream stable kernel. If version.patch is non-zero, clone
+            # major.minor.patch kernel corresponding to the distro in use
+            branch = "rolling-lts/mariner-2/"
+            branch += f"{version.major}.{version.minor}.{version.patch}.1"
+            branch_to_clone = f"{self._MARINER_KERNEL_SRC_REPO} -b {branch} --depth 1"
+            kernel_path = git.clone(
+                branch_to_clone,
+                self.get_tool_path(use_global=True),
+                fail_on_exists=False,
+            )
+            self.node.tools[Cp].copy(
+                src=self.node.get_pure_path(
+                    f"/boot/config-{uname_result.kernel_version_raw}"
+                ),
+                dest=PurePath(".config"),
+                cwd=kernel_path,
+                sudo=True,
+            )
+
+            # build kselftests
+            self.node.tools[Make].run(
+                "KBUILD_OUTPUT=build -C tools/testing/selftests gen_tar",
+                cwd=kernel_path,
+                sudo=True,
+                expected_exit_code=0,
+                expected_exit_code_failure_message="could not generate kselftest tar.",
+            ).assert_exit_code()
+
+        tool_path = self.get_tool_path(use_global=True)
+        self.node.tools[Tar].extract(
+            str(self._remote_tar_path), str(tool_path), sudo=True
+        )
+        self._log.debug(f"Extracted tar from path {self._remote_tar_path}!")
+
+        return self._check_exists()
+
+    def run_all(
+        self, test_result: TestResult, environment: Environment, log_path: str
+    ) -> List[KselftestResult]:
+        self.run(
+            " 2>&1 | tee kselftest-results.txt",
+            force_run=True,
+            sudo=True,
+            shell=True,
+            timeout=5000,
+        )
+
+        # get username
+        username = self.node.tools[Whoami].get_username()
+
+        # Allow read permissions for "others" to remote copy the file
+        # kselftest-results.txt
+        self.node.execute(
+            f"chmod 644 /home/{username}/kselftest-results.txt", sudo=True
+        )
+
+        # copy kselftest-results.txt from remote to local node for processing results
+        remote_copy = self.node.tools[RemoteCopy]
+        remote_copy.copy_to_local(
+            PurePosixPath(f"/home/{username}/kselftest-results.txt"), PurePath(log_path)
+        )
+
+        local_kselftest_results_path = PurePath(log_path) / "kselftest-results.txt"
+
+        # parse results from local_kselftest_results_path file
+        # read the file
+        with open(local_kselftest_results_path, encoding="utf8") as f:
+            result_output = f.read()
+            results = self._parse_results(result_output)
+
+        if not results:
+            raise LisaException("tests did not run, kselftest-results.txt is empty")
+
+        # assert that all tests passed
+        failed_tests = []
+        for result in results:
+            if result.status == TestStatus.FAILED:
+                failed_tests.append(result.name)
+
+            # create test result message
+            info: Dict[str, Any] = {}
+            info["information"] = {}
+            info["information"]["exit_value"] = result.exit_value
+            subtest_message = create_test_result_message(
+                SubTestMessage,
+                test_result,
+                environment,
+                result.name,
+                result.status,
+                other_fields=info,
+            )
+
+            # notify subtest result
+            notifier.notify(subtest_message)
+
+        # assert that none of the tests failed
+        assert_that(failed_tests).described_as("kselftests failed").is_empty()
+
+        return results
+
+    def _parse_results(self, result: str) -> List[KselftestResult]:
+        parsed_result: List[KselftestResult] = []
+
+        lines_list = result.splitlines()
+
+        for line in lines_list:
+            line_str = "".join(line)
+            matched = find_groups_in_lines(line_str, self._RESULT_KSELFTEST_OK_REGEX)
+
+            # get testcase data
+            if matched:
+                # example parsed_result: name='breakpoint_test_arm64',
+                # status=<TestStatus.FAILED: 4>, exit_value='1'
+                parsed_result.append(
+                    KselftestResult(
+                        name=matched[0]["name"],
+                        status=self._parse_status_to_test_status(matched),
+                        exit_value=self._parse_exit_val_to_test_status(matched),
+                    )
+                )
+
+        return parsed_result
+
+    def _parse_status_to_test_status(
+        self, result_list: List[Dict[str, str]]
+    ) -> TestStatus:
+        # example result_list for failing test
+        # [{'status': 'not ok', 'name': 'test_kmem', 'reason': None, 'exit': '126'}]
+        status = result_list[0]["status"]
+
+        # passed or skipped tests
+        if status == "ok":
+            if result_list[0]["reason"] == "SKIP":
+                return TestStatus.SKIPPED
+            else:
+                return TestStatus.PASSED
+
+        # failed or timed out tests
+        return TestStatus.FAILED
+
+    def _parse_exit_val_to_test_status(self, result_list: List[Dict[str, str]]) -> int:
+        # example result_list for failing test
+        # [{'status': 'not ok', 'name': 'test_kmem', 'reason': None, 'exit': '126'}]
+        exit_val = result_list[0]["exit"]
+
+        # tests that failed with exit code
+        if exit_val:
+            return int(exit_val)
+        # tests that passed, skipped or timed out
+        else:
+            return 0


microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,58 @@
...
+        # command to run kselftests
+        self._command = self.get_tool_path(use_global=True) / "kselftest-packages/run_kselftest.sh > kselftest-results.txt"
+        # tar file path specified in yml
+        self._tar_file_path = kwargs.pop("kselftest_file_path", "")
+
+        if not self._tar_file_path:
+            raise MissingPackagesException(["kselftest"])
+
+    def _install(self) -> bool:
+        # /home/lisatest/lisa_working/tool/kselftest
```
squirrelsc: Review the comments or make it more descriptive.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -0,0 +1,49 @@
...
+    description="""
+    This test suite is used to run kselftests.
+    """,
+)
+class KselftestTestsuite(TestSuite):
+    _TIME_OUT = 12000
+
+    @TestCaseMetadata(
+        description="""
+        This test case will run kselftests tests.
```
squirrelsc: Add more descriptions like how to pass in the parameters, so it can run well.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # command to run kselftests
+        self._command = self.get_tool_path(use_global=True) / "kselftest-packages/run_kselftest.sh > kselftest-results.txt"
+        # tar file path specified in yml
+        self._tar_file_path = kwargs.pop("kselftest_file_path", "")
+
+        if not self._tar_file_path:
+            raise MissingPackagesException(["kselftest"])
```
squirrelsc: Make the error message clearer to explain how to get this test case started, like what variable expected, how the package produced.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -0,0 +1,49 @@
...
+        """,
+        priority=3,
```
dsrivastavv, line 36: Can you please ensure that the priority is correct: https://mslisa.readthedocs.io/en/main/run_test/microsoft_tests.html#test-priority?

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -0,0 +1,49 @@
...
+        timeout=_TIME_OUT,
```
dsrivastavv, line 37: 12000 seconds might be a large timeout value for kselftest. What is the max test runtime that you have observed?

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -0,0 +1,49 @@
...
+        self,
+        node: Node,
+        environment: Environment,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        try:      
+            kselftest = node.tools.get(tool_type=Kselftest, kselftest_file_path=variables["kselftest_file_path"])
+        except MissingPackagesException as identifier:
```
dsrivastavv: Can you please provide comments around what this situation is? Also, if this is regarding missing packages on the distros it is better to throw an error. This is because your test could not run due to installation failures even though your intent was to run it.  Skipping is for the scenario where the test is definitely known not to run.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,58 @@
...
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # command to run kselftests
+        self._command = self.get_tool_path(use_global=True) / "kselftest-packages/run_kselftest.sh > kselftest-results.txt"
+        # tar file path specified in yml
+        self._tar_file_path = kwargs.pop("kselftest_file_path", "")
```
dsrivastavv: `kselftest_file_path` should be an argument in function rather than from kwarg from clarity

microsoft/testsuites/kselftest/kselftest.py
dsrivastavv: You can raise a LISAException directly

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,58 @@
...
+        # /home/lisatest/lisa_working/tool/kselftest
+        tool_path = self.get_tool_path(use_global=True)
+        # /home/lisatest/lisa_working/tool/kselftest/kselftests.tar.xz
+        file_name = tool_path/os.path.basename(self._tar_file_path)
+        # copies tar file from local machine to remote machine
+        self.node.shell.copy(self._tar_file_path, file_name)
+
+        tar = self.node.tools[Tar]
+        tar.extract(file_name, tool_path)
+        print(f"INSTALL from {self._tar_file_path}!")
```
dsrivastavv: We don't use print statements directly. Please use logger for any logging

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,58 @@
...
+        return str(self._command)
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # command to run kselftests
+        self._command = self.get_tool_path(use_global=True) / "kselftest-packages/run_kselftest.sh > kselftest-results.txt"
```
dsrivastavv: ` > kselftest-results.txt` can be moved to run all function for clarity and keep only uptil `.sh`

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def _install(self) -> bool:
+        # /home/lisatest/lisa_working/tool/kselftest
+        tool_path = self.get_tool_path(use_global=True)
+        # /home/lisatest/lisa_working/tool/kselftest/kselftests.tar.xz
+        file_name = tool_path/os.path.basename(self._tar_file_path)
+        # copies tar file from local machine to remote machine
+        self.node.shell.copy(self._tar_file_path, file_name)
+
+        tar = self.node.tools[Tar]
```
dsrivastavv: [nit] No need to define it in memory variable if it has to be used once

dsrivastavv:
@niyelchu Please also fix the pipeline errors. You can fix basic errors with `nox -vRt format`, and then fix remaining errors manually from the output of command `nox -vRt all`

microsoft/testsuites/kselftest/kselftest-suite.py
niyelchu: I have used the same priority as ltp. Should this be set to 4 instead?

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: It doesn't need to redirect the output here. You can get the whole output from the ExecutableResult. So the LISA log can also contains the raw output for tourbleshooting.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # command to run kselftests
+        self._command = self.get_tool_path(use_global=True) / "kselftest-packages/run_kselftest.sh > kselftest-results.txt"
+        # tar file path specified in yml
+        self._tar_file_path = kwargs.pop("kselftest_file_path", "")
+
+        if not self._tar_file_path:
+            raise LisaException("tar file not found")
```
squirrelsc: Give the tar file path together, so it's easier to figure out path issues.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -0,0 +1,50 @@
...
+    _TIME_OUT = 3600
+
+    @TestCaseMetadata(
+        description="""
+        This test case will run kselftests tests.
+        """,
+        priority=3,
+        timeout=_TIME_OUT,
+    )
+    def kselftest(
```
squirrelsc: Add prefix `verify_`, so we know it's a validation case, not a perf test case.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,57 @@
...
+        file_name = tool_path/os.path.basename(self._tar_file_path)
+        # copies tar file from local machine to remote machine
+        self.node.shell.copy(self._tar_file_path, file_name)
+
+        self.node.tools[Tar].extract(file_name, tool_path)
+        self._log.debug(f"INSTALL from {self._tar_file_path}!")
+
+        return self._check_exists()
+
+    def run_all(self, log_path: str,) -> bool:
```
squirrelsc: remove the `,`.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,57 @@
...
+        self.run("", force_run=True, sudo=True, shell=True)
+        # copy kselftest-results.txt from remote to local node for processing results
+        remote_copy = self.node.tools[RemoteCopy]
+        remote_copy.copy_to_local(
```
squirrelsc, line 180: In next step, the results should be parsed as sub test results. So it's can be collected in subtests reports.

microsoft/testsuites/kselftest/kselftest-suite.py
dsrivastavv: It's upto your use case. My comment was to ensure that it is set correctly

niyelchu:
> @niyelchu Please also fix the pipeline errors. You can fix basic errors with `nox -vRt format`, and then fix remaining errors manually from the output of command `nox -vRt all`

Fixed pipeline errors.

microsoft/testsuites/kselftest/kselftest.py
niyelchu: > ` > kselftest-results.txt` can be moved to run all function for clarity and keep only uptil `.sh`

Moved kselftest-results.txt to run_all function.

@squirrelsc, the logs from the run are over 47k lines, so I thought it would be better to redirect the output to a file. Do you still recommend using ExecutableResult?

squirrelsc: If one line is 100bytes, it looks like around 470 lines? It's ok to put in log. You can save the output to a file at the same time, if it needs to be parsed.

niyelchu: > If one line is 100bytes, it looks like around 470 lines? It's ok to put in log. You can save the output to a file at the same time, if it needs to be parsed.

@squirrelsc the logs are around 47,000 lines.

squirrelsc: Oh, sorry, I understand it incorrectly. If so, the log file is better. When you're passing the file, and generate the subtests, you can output meaningful information to log. If you copy back the result only, it needs to analyze the content manually. Besides copying it back, please also generate subtests results from it.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,160 @@
...
+        for result in results:
+            if result.status == TestStatus.FAILED:
+                failed_tests.append(result.name)
+
+            # create test result message
+            info: Dict[str, Any] = {}
+            info["information"] = {}
+            info["information"]["name"] = result.name
+            info["information"]["exit_value"] = result.exit_value
+            subtest_message = create_test_result_message(
```
dsrivastavv, line 205: This is incorrect. It should be something like this:
```
subtest_message = create_test_result_message(
                SubTestMessage,
                test_result,
                environment,
                result.name,
                result.status,
                other_fields=info,
            )
```

where test_result is of type `TestResult` and passed from the test suite itself. A complete example is given here: https://github.com/microsoft/lisa/blob/d50773629ee2f36deb15204b9136c3798e3e65d5/microsoft/testsuites/ltp/ltpsuite.py#L86-L93

niyelchu:
> @niyelchu Please also fix the pipeline errors. You can fix basic errors with `nox -vRt format`, and then fix remaining errors manually from the output of command `nox -vRt all`

Warnings fixed.

squirrelsc:
LGTM， @LiliDeng please help double check and sign off.

dsrivastavv:
@squirrelsc  @LiliDeng  I am planning to test this e2e with Nischala to ensure that the data is being added to the subtest dashboard

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,168 @@
...
+        return True
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # command to run kselftests
+        self._command = (
+            self.get_tool_path(use_global=True)
+            / "kselftest-packages/run_kselftest.sh -c breakpoints"
+        )
+        # tar file path specified in yml
```
dsrivastavv, line 61: [nit] Please add an empty line between comments and the code

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,170 @@
...
+
+@dataclass
+class KselftestResult:
+    name: str = ""
+    status: TestStatus = TestStatus.QUEUED
+    exit_value: int = 0
+
+
+class Kselftest(Tool):
+    _RESULT_KSELFTEST_OK_REGEX = re.compile(r"ok (.*)\s+")
```
LiliDeng: add an example as comments above this line

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,170 @@
...
+
+        self.node.tools[Tar].extract(file_name, tool_path)
+        self._log.debug(f"INSTALL from {self._tar_file_path}!")
+
+        return self._check_exists()
+
+    def run_all(
+        self, test_result: TestResult, environment: Environment, log_path: str
+    ) -> bool:
+        self.run(" > kselftest-results.txt", force_run=True, sudo=True, shell=True)
```
LiliDeng: the default time out for this run is 5 mins, is it enough?

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,170 @@
...
+        # copy kselftest-results.txt from remote to local node for processing results
+        remote_copy = self.node.tools[RemoteCopy]
+        remote_copy.copy_to_local(
+            PurePosixPath("/home/lisatest/kselftest-results.txt"),
```
LiliDeng: don't use hardcode username

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -0,0 +1,51 @@
...
+    def verify_kselftest(
+        self,
+        node: Node,
+        environment: Environment,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        kselftest = node.tools.get(
+            Kselftest,
```
LiliDeng: we can skip test, if no kselftest_file_path provided

microsoft/testsuites/kselftest/kselftest.py
niyelchu: > add an example as comments above this line

added comment

microsoft/testsuites/kselftest/kselftest.py
niyelchu: > the default time out for this run is 5 mins, is it enough?

the timeout has been set to an hour in the file kselftest-suite.py: 
_TIME_OUT = 3600

microsoft/testsuites/kselftest/kselftest.py
niyelchu: > don't use hardcode username

Is there a variable that tracks the path "/home/lisatest" in particular?

microsoft/testsuites/kselftest/kselftest-suite.py
niyelchu: > we can skip test, if no kselftest_file_path provided

There is a design change when kselftest_file_path is NOT specified. We decided to clone the upstream kernel, build the kernel, build kselftests to generate the tar. The latest version of PR has this change.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -0,0 +1,51 @@
...
+)
+class KselftestTestsuite(TestSuite):
+    # kselftests take under an hour to complete, timeout below is in seconds
+    _TIME_OUT = 3600
+
+    @TestCaseMetadata(
+        description="""
+        This test case will run linux kernel self tests.
+        Steps:
+        1. Provide a tarball of kselftest binaries kselftests.tar.xz in a .yml file.
```
dsrivastavv: Can you please update the new logic here?

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,241 @@
...
+
+@dataclass
+class KselftestResult:
+    name: str = ""
+    status: TestStatus = TestStatus.QUEUED
+    exit_value: int = 0
+
+
+class Kselftest(Tool):
+    repo = "git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git"
```
dsrivastavv: Global variables should be capitalized

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,241 @@
...
+    def command(self) -> str:
+        return str(self._command)
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def __init__(self, node: Node, kselftest_file_path: str, *args: Any, **kwargs: Any) -> None:
+        super().__init__(node, *args, **kwargs)
+        # command to run kselftests
```
dsrivastavv, line 73: [nit] Line break between comment and previous sentence if the not the start of a new block. Please apply this at the other places in code too

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,241 @@
...
+            # clone kernel, build kernel, then build kselftests
+            uname = self.node.tools[Uname]
+            uname_result = uname.get_linux_information(force_run=False)
+            version=uname_result.kernel_version
+            git = self.node.tools[Git]
+
+            if (version.patch == 0):
+                # clone major.minor kernel from upstream
+                branch = "linux-"
+                branch += str(version.major) + "." + str(version.minor) + ".y"
```
dsrivastavv: You can use f-string, which will be more cleaner and you won't have to use `str`

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,241 @@
...
+                branch_to_clone = self.repo + " -b " + branch + " --depth 1"
+                kernel_path = git.clone(branch_to_clone, self.get_tool_path(use_global=True), fail_on_exists=False)
+                self.node.execute(cmd="make defconfig", cwd=kernel_path)
```
somil55, line 68: Please use make tool. Also applicable in other places

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,241 @@
...
+            else:
+                if isinstance(self.node.os, CBLMariner):
+                    branch = "rolling-lts/mariner-2/"
+                    branch += str(version.major) + "." + str(version.minor) + "." + str(version.patch) + ".1"
+                    branch_to_clone = self.mariner_repo + " -b " + branch + " --depth 1"
+                    kernel_path = git.clone(branch_to_clone, self.get_tool_path(use_global=True), fail_on_exists=False)
+                    self.node.execute(cmd="wget 'https://raw.githubusercontent.com/microsoft/CBL-Mariner/2.0/SPECS/kernel/config' -O .config", cwd=kernel_path)
```
dsrivastavv: Use wget tool

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,241 @@
...
+                else:
+                    # clone major.minor.patch upstream stable kernel for distros other than Mariner
+                    branch += str(version.major) + "." + str(version.minor) + ".y"
+                    branch_to_clone = self.repo + " -b " + branch + " --depth 1"
+                    kernel_path = git.clone(branch_to_clone, self.get_tool_path(use_global=True), fail_on_exists=False)
+                    self.node.execute(cmd="make defconfig", cwd=kernel_path)
+            self.node.execute(cmd="make -j2", cwd=kernel_path)
```
dsrivastavv: You can speed this up by using the number of cores this VM is running on. Please use `cores = node.tools[Lscpu].get_core_count()`

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,241 @@
...
+            self.node.execute(cmd="make KBUILD_OUTPUT=build FORCE_TARGETS=1 -C tools/testing/selftests gen_tar", cwd=kernel_path, sudo=True)
+            self._tar_file_path = str(kernel_path) + "/build/kselftest/kselftest_install/kselftest-packages/kselftest.tar.gz"
```
dsrivastavv: [nit] Large string can be made a class level constant

dsrivastavv: Also please use f-strings

dsrivastavv:
The code still has nox static check errors. Can you please run `nox -vt all` and fix the errors?

microsoft/testsuites/kselftest/kselftest.py
niyelchu: > `kselftest_file_path` should be an argument in function rather than from kwarg from clarity

kselftest_file_path is now a function arg instead of kwarg, thanks for helping with this :)

microsoft/testsuites/kselftest/kselftest.py
niyelchu: Added empty line.

microsoft/testsuites/kselftest/kselftest-suite.py
niyelchu: Updated test description to include cloning and building kernel.

niyelchu:
> The code still has nox static check errors. Can you please run `nox -vt all` and fix the errors?

Fixed. Looks like mypy is failing for a different file in the repo.

microsoft/testsuites/kselftest/kselftest.py
niyelchu: > You can speed this up by using the number of cores this VM is running on. Please use `cores = node.tools[Lscpu].get_core_count()`

Updated code to use get_core_count().

niyelchu:
> The code still has nox static check errors. Can you please run `nox -vt all` and fix the errors?

Fixed all static check errors.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,277 @@
...
+from assertpy import assert_that
+
+from lisa import Environment, notifier
+from lisa.base_tools.uname import Uname
+from lisa.executable import Tool
+from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
+from lisa.node import Node
+from lisa.operating_system import CBLMariner, Ubuntu
+from lisa.testsuite import TestResult
+from lisa.tools.git import Git
```
dsrivastavv: Can you please merge tool imports together. You can do something like this: `from lisa.tools import Git, Lscpu, ...`

microsoft/testsuites/kselftest/kselftest.py
dsrivastavv: This should be fixed now!

dsrivastavv (dismissed):
LGTM!

dsrivastavv:
@LiliDeng Can you please take a look? We will also need to merge this as a single commit.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,263 @@
...
+                src=self.node.get_pure_path(
+                    f"/boot/config-{uname_result.kernel_version_raw}"
+                ),
+                dest=PurePath(".config"),
+                cwd=kernel_path,
+                sudo=True,
+            )
+
+            # build kselftests
+            self.node.tools[Make].run(
```
LiliDeng, line 142: please check whether it compile successfully or not

niyelchu: > please check whether it compile successfully or not

Added expected exit code, failure message and assert_exit_code().

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,276 @@
...
+                # example parsed_result: name='breakpoint_test_arm64',
+                # status=<TestStatus.FAILED: 4>, exit_value='1'
+                parsed_result.append(
+                    KselftestResult(
+                        name=matched[0]["name"],
+                        status=self._parse_status_to_test_status(matched),
+                        exit_value=self._parse_exit_val_to_test_status(matched),
+                    )
+                )
+                print("********")
```
dsrivastavv: Please use logging class for this rather than `print` function

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,276 @@
...
+        # example passed test log: "2 selftests: breakpoints: breakpoint_test"
+        # example timeout test log: not ok 8 selftests: netfilter: nft_concat_range.sh # TIMEOUT 45 seconds
+
+        status = result_list[0]["status"]
+        exit_code = result_list[0]["info"].split("#")
+
+        # passed or skipped tests
+        if status == 'ok':
+            if len(exit_code) == 2:
+                if exit_code[1] == "SKIP":
```
dsrivastavv: What should the test status be on the `else` block? It will return a `None` here.

dsrivastavv (dismissed):

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,274 @@
...
+    # install common dependencies
+    def _install(self) -> bool:
+        if not isinstance(self.node.os, CBLMariner):
+            raise SkippedException("kselftests are supported on Mariner VMs only.")
+
+        if self._tar_file_path:
+            self.node.shell.copy(PurePath(self._tar_file_path), self._remote_tar_path)
+        else:
+            # clone kernel, build kernel, then build kselftests
+            self.node.os.install_packages(
```
squirrelsc, line 87: How is the list compatible with different distros?

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,274 @@
...
+        else:
+            self._command = (
+                self.get_tool_path(use_global=True)
+                / "kselftest-packages/run_kselftest.sh"
+            )
+
+    # install common dependencies
+    def _install(self) -> bool:
+        if not isinstance(self.node.os, CBLMariner):
+            raise SkippedException("kselftests are supported on Mariner VMs only.")
```
squirrelsc: 1. `kselftests test case is supported Mariner`. We need to claim it's a limitation of LISA, instead of kselftests.
2. Here should raise unsupported exception, and convert it to skippped exception in test case level. The tools are not related to test purpose, so it shouldn't raise any skipped exception.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,274 @@
...
+        failed_tests = []
+        for result in results:
+            if result.status == TestStatus.FAILED:
+                failed_tests.append(result.name)
+
+            # create test result message
+            info: Dict[str, Any] = {}
+            info["information"] = {}
+            info["information"]["name"] = result.name
+            info["information"]["exit_value"] = result.exit_value
```
squirrelsc, line 204: It looks like the info is not used.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,274 @@
...
+            subtest_message = create_test_result_message(
+                SubTestMessage, test_result, environment, result.name, result.status
+            )
+
+            # notify subtest result
+            notifier.notify(subtest_message)
+
+        # assert that none of the tests failed
+        assert_that(
+            failed_tests, f"The following tests failed: {failed_tests}"
```
squirrelsc: The error message shows the non-empty data, so it doesn't need to print again.

niyelchu: > The error message shows the non-empty data, so it doesn't need to print again.

self.error('Expected <%s> to be empty, but was not.' % self.val)

Above is the error message I'm seeing in pipeline run. Please correct me if I'm wrong. I think printing failed tests is useful here since MarinerMOS team can just look at this list to identify new regressions when they sign-off on their releases.

squirrelsc: It looks like a line in call trace. The final error message should be different.

niyelchu: > It looks like a line in call trace. The final error message should be different.

Ah I see it now, expected to be empty message captures it, will fix, thank you!

microsoft/testsuites/kselftest/kselftest.py
niyelchu: This is a limitation of kselftests. Our focus for now is to run kselftests automatically on Mariner VMs. The main use case of this PR is for MarinerMOS team and LSG kernel team to validate kernels automatically in our pipelines which uses mariner build container. This feature can be extended for different distros in the future.

Initially we thought we'd just skip running tests if a non-mariner distro is used. Do you recommend changing this to UnsupportedDistroException instead?

niyelchu: > 1. `kselftests test case is supported Mariner`. We need to claim it's a limitation of LISA, instead of kselftests.
> 2. Here should raise unsupported exception, and convert it to skippped exception in test case level. The tools are not related to test purpose, so it shouldn't raise any skipped exception.

I have fixed all the comments, please re-review. Thanks!

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,285 @@
...
+    _MARINER_KERNEL_SRC_REPO = "https://github.com/microsoft/CBL-Mariner-Linux-Kernel"
+    _KERNEL_REPO_NAME = "CBL-Mariner-Linux-Kernel"
+    _KSELFTEST_TAR_PATH = (
+        "/build/kselftest/kselftest_install/kselftest-packages/kselftest.tar.gz"
+    )
+
+    # kselftest result log has "ok" and "not ok" prefixes, use regex to filter them
+    # [{'status': 'not ok', 'name': 'step_after_suspend_test', 'info': '# exit=126'}]
+    _RESULT_KSELFTEST_OK_REGEX = re.compile(
+        r"^(?P<status>(not ok|ok))\s+\d+\s+selftests:\s+\S+:\s+(?P<name>\S+)\s*(?P<info>.*)$"  # noqa: E501
```
squirrelsc: The regexp can be more accurate to just capture the values, which are needed, so it doesn't need to split string again in code.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,285 @@
...
+                result.name,
+                result.status,
+                other_fields=info,
+            )
+
+            # notify subtest result
+            notifier.notify(subtest_message)
+
+        # assert that none of the tests failed
+        assert_that(failed_tests).is_empty()
```
squirrelsc: Add `described_as`, so if it's failed here, the message can help find the right code line. I said not to print the array again, but the message is still helpful.

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: I mean not to use `SkippedException` in tool level. The tool is to execute, it shouldn't contain test purpose. The test case decide a case to fail or skip.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -0,0 +1,285 @@
...
+        )
+        self._log.debug(f"Extracted tar from path {self._remote_tar_path}!")
+
+        return self._check_exists()
+
+    def run_all(
+        self, test_result: TestResult, environment: Environment, log_path: str
+    ) -> List[KselftestResult]:
+        self.run(
+            # " > kselftest-results.txt",
```
LiliDeng: remove unused line

microsoft/testsuites/kselftest/kselftest.py
niyelchu: > The regexp can be more accurate to just capture the values, which are needed, so it doesn't need to split string again in code.

I tried to do this but we will still have to split the info string further as it can have 4 different possibilities:

not ok 1 selftests: vm: run_vmtests.sh # TIMEOUT 45 seconds
ok 2 selftests: cgroup: test_kmem # SKIP
not ok 4 selftests: clone3: clone3_cap_checkpoint_restore # exit=1
ok 3 selftests: clone3: clone3_set_tid

niyelchu: > > The regexp can be more accurate to just capture the values, which are needed, so it doesn't need to split string again in code.
> 
> I tried to do this but we will still have to split the info string further as it can have 4 different possibilities:
> 
> not ok 1 selftests: vm: run_vmtests.sh # TIMEOUT 45 seconds ok 2 selftests: cgroup: test_kmem # SKIP not ok 4 selftests: clone3: clone3_cap_checkpoint_restore # exit=1 ok 3 selftests: clone3: clone3_set_tid

Thanks for helping fix this, I have updated parsing logic.

LiliDeng (approved):

PR #2638, opened 2023-03-02
Add boot parameters transformer

contre:
Add a transformer to allow for adding additional boot parameters to the kernel commandline.

### DIFF
--- a/lisa/transformers/boot_parameters.py
+++ b/lisa/transformers/boot_parameters.py
@@ -1,0 +1,103 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from dataclasses import dataclass, field
+from pathlib import PurePosixPath
+from typing import Any, Dict, List, Optional, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import LisaException, schema
+from lisa.node import quick_connect
+from lisa.operating_system import Debian, Oracle, Redhat
+from lisa.tools import Cat, Sed, Uname
+from lisa.transformer import Transformer
+from lisa.util import field_metadata
+
+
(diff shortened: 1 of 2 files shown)


lisa/transformers/boot_parameters.py
```diff
@@ -0,0 +1,103 @@
...
+@dataclass
+class BootParametersTransformerSchema(schema.Transformer):
+    # the SSH connection information to the node
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    parameters: List[str] = field(default_factory=list)
+
+
+class BootParametersTransformer(Transformer):
```
squirrelsc, line 28: Which scenarios will the transformer be used to?

kamalca: Hello @squirrelsc,

Some background on this functionality:

CVM uses `swiotlb` / bounce buffer as part of its security functionalities. Recent testing of CVM revealed many bugs related to the bounce buffer that have been in the kernel for a long time. These bugs were not discovered earlier because regular VMs on Azure have the bounce buffer disabled. This transformer would allow the bounce buffer to be enabled on a regular VM and would allow us to run validation on new kernels before they are released for CVM.

The transformer should allow us to add the boot parameter `swiotlb=force` then run any / all lisa test cases on the modified image.

Hope this helps!

squirrelsc: The transformer need a connection, is it useful to ready environments only?

LiliDeng:
@johnsongeorge-w @kamalca is it needed? if yes, feel free to create or reopen it when it is ready for review

PR #2639, opened 2023-03-02
ch_tests_tool: don't leave sub-tests in RUNNING state

anirudhrb:
If subtests fail to finish within the stipulated time, explicitly mark them as FAILED instead of leaving them in RUNNING state.

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -301,6 +301,12 @@ class CloudHypervisorTests(Tool):
         for subtest in subtests:
             status = subtest_status[subtest]
             message = messages.get(status, "")
+
+            if status == TestStatus.RUNNING:
+                # Sub-test started running but didn't finish within the stipulated time.
+                # It should be treated as a failure.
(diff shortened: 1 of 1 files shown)


pupacha (approved):

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -301,6 +301,12 @@ def _extract_test_results(
         for subtest in subtests:
             status = subtest_status[subtest]
             message = messages.get(status, "")
+
+            if status == TestStatus.RUNNING:
+                # Sub-test started running but didn't finish within the stipulated time.
+                # It should be treated as a failure.
+                status = TestStatus.FAILED
```
LiliDeng, line 308: should we extend the running time?

anirudhrb: No, the running time is already very generous, and is not the root cause. There is some underlying issue in the kernel or Cloud Hypervisor that is causing this. The goal of this PR is to simply report the status correctly.

LiliDeng (approved):

PR #2640, opened 2023-03-02
support multi clouds

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -245,6 +251,12 @@ class AzurePlatformSchema:
             validate=validate.Regexp(constants.GUID_REGEXP),
         ),
     )
+    cloud: str = field(
+        default="public",
+        metadata=field_metadata(
+            validate=validate.OneOf(["mooncake", "public", "germany", "fairfax"]),
```
squirrelsc: Don't create other names, use names in the SDK like "AzureChinaCloud".

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -245,6 +251,12 @@ class AzurePlatformSchema:
             validate=validate.Regexp(constants.GUID_REGEXP),
         ),
     )
+    cloud: str = field(
```
squirrelsc: 1. Try if it's possible to inherit the CloudEndpoints, and add schema annotations on it, like,

```python
@dataclass_json()
@dataclass
class CloudEntry(CloudEndpoints):
    ...
```

2. Make it type like `Union[str, CloudEntry]`. If it's a string, parse it to CloudEntry. For other places, just pass the CloudEntry object in, so it can retrieve any url in it.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: I didn't change the location part yet, could you please check current change for this part? thanks!

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -224,6 +232,30 @@ class AzureLocation:
     capabilities: Dict[str, AzureCapability] = field(default_factory=dict)
 
 
+@dataclass_json()
+@dataclass
+class CloudEndpointsSchema:
```
squirrelsc, line 213: can it inherit from `msrestazure.azure_cloud` directly?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -511,6 +511,8 @@ def get_compute_client(
         credential=platform.credential,
         subscription_id=subscription_id,
         api_version=api_version,
+        base_url=platform.cloud.endpoints.resource_manager,
```
squirrelsc, line 555: For each part, it can pass in the whole cloud object. So it's flexible to take parts from it.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -326,12 +359,20 @@ class AzurePlatform(Platform):
...
+        "azureusgovernment": AZURE_US_GOV_CLOUD,
+    }
 
     def __init__(self, runbook: schema.Platform) -> None:
         super().__init__(runbook=runbook)
 
         # for type detection
         self.credential: DefaultAzureCredential
+        # by default, use public cloud
+        self.cloud: Cloud = AZURE_PUBLIC_CLOUD
```
squirrelsc: It should be in the same line to declaration.

```python
cloud: Optional[Union[str, Dict[str, Any]]] = field(default=AZURE_PUBLIC_CLOUD)
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -793,18 +837,65 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+        #       active_directory_resource_id: https://management.core.windows.net/
+        #       active_directory_graph_resource_id: https://graph.windows.net/
+        #       microsoft_graph_resource_id: https://graph.microsoft.com/
+        #     suffixes:
+        #       storage_endpoint: core.windows.net
+        #       keyvault_dns: .vault.azure.net
+        #       sql_server_hostname: .database.windows.net
+        #       azure_datalake_store_file_system_endpoint: azuredatalakestore.net
+        #       azure_datalake_analytics_catalog_and_job_endpoint: azuredatalakeanalytics.net  # noqa: E501
+        if isinstance(azure_runbook.cloud, dict):
```
squirrelsc: `elif`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -246,6 +278,7 @@ class AzurePlatformSchema:
             validate=validate.Regexp(constants.GUID_REGEXP),
         ),
     )
+    cloud: Optional[Union[str, Dict[str, Any]]] = field(default=None)
```
squirrelsc: It can follow marketplace pattern in AzureNodeSchema.

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: it should be cloud type to store the cloud type for current run?

lisa/sut_orchestrator/azure/common.py
LiliDeng: only resource manager is used in this method

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: I tried like below
`class CloudEndpointsSchema(CloudEndpoints):`
And see below pylance error
`Base class type is unknown, obscuring type of derived class`
I don't understand the benefit of inherit.

lisa/sut_orchestrator/azure/common.py
squirrelsc: I mean for every place, pass the object in. That will get all logic consistent, and simple to manage the interface.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Look at the marketplace, vhd and shared_image_gallery, use a raw object, and properties to manage it.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: ok, if it doesn't support, we can keep as is. Use the same object can reduce maintenance.

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: I have changed it like this way.

lisa/sut_orchestrator/azure/common.py
LiliDeng: I think passing platform would be enough.
```
def get_compute_client(
    platform: "AzurePlatform",
    api_version: Optional[str] = None,
    subscription_id: str = "",
) -> ComputeManagementClient:
```

squirrelsc: All other places should be fixed to pass cloud object, not only one or two used endpoints.

LiliDeng: updated

squirrelsc (dismissed):

LiliDeng:
hold on the merge, working on internal code

squirrelsc (dismissed):

squirrelsc (approved):

PR #2642, opened 2023-03-03
ch_tests_tool: Improve subtests result parsing logic

pupacha:
At times, we see multiple jsons present in a single line output because of which subtest result is not being parsed into json and is ignored. This will change will take care of extracting multiple JSON objects from a single line of string and none of the subtest results extraction is missed.
Ex line: {"version":"30.0.0"}{ "type": "test", "name": "common_parallel::test_boot_from_virtio_pmem", "event": "ok" }

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -250,49 +250,42 @@ class CloudHypervisorTests(Tool):
         # { "type": "test", "event": "ok", "name": "integration::test_vfio" }
         lines = output.split("\n")
         for line in lines:
-            matches = re.findall(r"{.*}", line)
-            if not matches:
-                continue
-            line = matches[0]
-            result = {}
             try:
-                result = json.loads(line)
+                json_results = [json.loads(line)]
             except json.decoder.JSONDecodeError:
-                continue
+                json_results = extract_jsons(line)
 
-            if type(result) is not dict:
-                continue
+            for result in json_results:
+                if type(result) is not dict:
+                    continue
+                if "type" not in result or result["type"] != "test":
+                    continue
+                if "event" not in result or result["event"] not in [
+                    "started",
+                    "ok",
+                    "failed",
+                    "ignored",
+                ]:
+                    continue
+                if result["event"] == "started":
+                    status = TestStatus.RUNNING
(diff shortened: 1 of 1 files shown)


squirrelsc (dismissed):

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -250,49 +250,39 @@ def _extract_test_results(
         # { "type": "test", "event": "ok", "name": "integration::test_vfio" }
         lines = output.split("\n")
         for line in lines:
-            matches = re.findall(r"{.*}", line)
-            if not matches:
+            if '"type": "test"' not in line:
```
anirudhrb: Why do we need this? This is not a reliable check. In the future, cargo might format the JSON without the space i.e. `"type":"test"`.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -250,49 +250,39 @@ def _extract_test_results(
...
-
-            if result["event"] == "started":
-                status = TestStatus.RUNNING
-            elif result["event"] == "ok":
-                status = TestStatus.PASSED
-            elif result["event"] == "failed":
-                status = TestStatus.FAILED
-            elif result["event"] == "ignored":
-                status = TestStatus.SKIPPED
+            for result in extract_jsons(line):
```
anirudhrb: We discussed that we would first try to parse the line as-is and if that fails, we will try `extract_jsons`. Why did you deviate from that?

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
pupacha: ok, I'll remove this check. This check isn't a necessity.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
pupacha: Changed it to parse the line as-is initially and try `extract_jsons` for failing case.

anirudhrb (approved):

pupacha:
@squirrelsc @LiliDeng Please review

LiliDeng (approved):

PR #2654, opened 2023-03-10
Doc remove Python install in MSStore &Run as Admin

paxue:
Remove Python install from Microsoft store and run as admin from Doc:

Lisa and pip libs Installation paths are different between the python
distributions installed from packages vs MicrosoftStore:
Using python 3.10 as an example,
(1). Installed from Microsoft Store, the libs installation path is:
"C:\Users\<alias>\AppData\Local\Packages\PythonSoftwareFoundation.
Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\Scripts"
(2). Installed from Python Package:
C:\Users\<alias>\AppData\Roaming\Python\Python310\Scripts

In (2) option, path could be included in Python Package installation
stage. There is a option to Add Path to Env Var. However, in the (1)
option, we don't have such kind of the config. User need to add it
manually. To facilitate users from missing config, remove the
instruction installing python from Microsoft Store.

At the same time, no need to run as admin during installation
as it will bring some other problems.

### DIFF
--- a/docs/installation_windows.rst
+++ b/docs/installation_windows.rst
@@ -22,20 +22,6 @@ LISA has been tested to work with `Python 3.8 64-bit <https://www.python.org/>`_
 The latest version of Python 3 is recommended. If you find that LISA is not compatible
 with higher version Python, `please file an issue <https://github.com/microsoft/lisa/issues/new>`__.
 
-Install from Microsoft Store (recommended)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-It is recommended to install Python from the Microsoft Store. Packages are regularly
-published by the Python Software Foundation and will set up paths as needed.
-
-To install from the Microsoft Store, search for Python in the store interface or,
-if no other Python version is installed, running `python3` from the command line
-will bring up the latest version.
-More details can be found `here<https://docs.python.org/3/using/windows.html#windows-store>`.
-
-Install using full installer (alternative)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
 The full installer allows greater customization and doesn't have the security restriction
 of the Microsoft Store packages, so may be preferred in some situations.
 
@@ -96,8 +82,7 @@ This installation method is used to run LISA if no change in source code is desi
 
    python3 -m pip install --upgrade pip
 
-The example below will install LISA directly for the invoking user.
-To install system-wide, run from and Administrator console.
+The example below will install LISA directly for the invoking user. No need to run as Administrator.
 
 .. code:: bash
 


LiliDeng:
@paxue thanks for your contribution, any reason we remove these content?

paxue:
> @paxue thanks for your contribution, any reason we remove these content?

Chi Song Suggested not to install python from Microsoft Store. The path setting has some problem. I have some problem when installing from Microsoft Store.
For the Lisa installation, no need to run as admin as it will bring some other problems.

docs/installation_windows.rst
```diff
@@ -22,18 +22,8 @@ LISA has been tested to work with `Python 3.8 64-bit <https://www.python.org/>`_
...
-It is recommended to install Python from the Microsoft Store. Packages are regularly
-published by the Python Software Foundation and will set up paths as needed.
-
-To install from the Microsoft Store, search for Python in the store interface or,
-if no other Python version is installed, running `python3` from the command line
-will bring up the latest version.
-More details can be found `here<https://docs.python.org/3/using/windows.html#windows-store>`.
 
-Install using full installer (alternative)
+Install using full installer
```
squirrelsc: This line can be removed too.

squirrelsc:
> > @paxue thanks for your contribution, any reason we remove these content?
> 
> Chi Song Suggested not to install python from Microsoft Store. The path setting has some problem. I have some problem when installing from Microsoft Store. For the Lisa installation, no need to run as admin as it will bring some other problems.

Good explanation, if they are in PR description or commit message, that would be clearer.

avylove:
This seems like the wrong approach. Shouldn't we identify the issues in LISA or the Microsoft store package and fix them? If there is a legitimate reason the Microsoft Store package won't work, that needs to be clearly explained. The danger of using another mechanism is the updates are not automatically maintained.

@zooba, any thoughts on this?

squirrelsc:
> This seems like the wrong approach. Shouldn't we identify the issues in LISA or the Microsoft store package and fix them? If there is a legitimate reason the Microsoft Store package won't work, that needs to be clearly explained. The danger of using another mechanism is the updates are not automatically maintained.
> 
> @zooba, any thoughts on this?

Feel free to explore solutions on store or LISA. Before that, we will remove this path from LISA, so LISA users won't be blocked.

avylove:
> > This seems like the wrong approach. Shouldn't we identify the issues in LISA or the Microsoft store package and fix them? If there is a legitimate reason the Microsoft Store package won't work, that needs to be clearly explained. The danger of using another mechanism is the updates are not automatically maintained.
> > @zooba, any thoughts on this?
> 
> Feel free to explore solutions on store or LISA. Before that, we will remove this path from LISA, so LISA users won't be blocked.

Please describe what the actual issue is and how this is blocking anyone. I have not seen any issues using Python from the Microsoft store with LISA and this PR shows no context. It doesn't even have a description.

squirrelsc:
> > > This seems like the wrong approach. Shouldn't we identify the issues in LISA or the Microsoft store package and fix them? If there is a legitimate reason the Microsoft Store package won't work, that needs to be clearly explained. The danger of using another mechanism is the updates are not automatically maintained.
> > > @zooba, any thoughts on this?
> > 
> > 
> > Feel free to explore solutions on store or LISA. Before that, we will remove this path from LISA, so LISA users won't be blocked.
> 
> Please describe what the actual issue is and how this is blocking anyone. I have not seen any issues using Python from the Microsoft store with LISA and this PR shows no context. It doesn't even have a description.

Did you try without admin privileges? I spent two afternoons last year to figure out it on clean environments. I don't want to spend more time, until it's confirmed thoroughly.

avylove:
> Did you try without admin privileges? I spent two afternoons last year to figure out it on clean environments. I don't want to spend more time, until it's confirmed thoroughly.

Normal user privileges. The only time I saw an issue was sometimes there is a permission error when a virtual environment already exists and was created with Python installed through another method, but you can just delete the .venv directory and use nox to recreate it. Not sure what issues others have had. If people are having issues, I'd expect to see an issue created so it's documented, but I didn't see one.

squirrelsc:
> > Did you try without admin privileges? I spent two afternoons last year to figure out it on clean environments. I don't want to spend more time, until it's confirmed thoroughly.
> 
> Normal user privileges. The only time I saw an issue was sometimes there is a permission error when a virtual environment already exists and was created with Python installed through another method, but you can just delete the .venv directory and use nox to recreate it. Not sure what issues others have had. If people are having issues, I'd expect to see an issue created so it's documented, but I didn't see one.

Can you try "pip install" as runtime environment? It's what paxue used.

avylove:
> Can you try "pip install" as runtime environment? It's what paxue used.

I don't see any issues except the scripts directory isn't in PATH, so you'll need to invoke with python3.X -m lisa or update PATH. I suggest @paxue open bug so we can investigate their specific issues. I also wouldn't install outside of a virtual environment unless LISA is the only thing you use Python for on that system.

paxue:
>

Let me explain what's going on here. Basically this is caused by the python lib PATH is not included correctly in Microsoft Store installation.
Lisa and pip libs Installation paths are different between the python distributions installed from packages & MicrosoftStore:
Using python 3.10 as an example,
(1). Installed from Microsoft Store, the libs installation path is: 
C:\Users\paxue\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0\LocalCache\local-packages\Python310\Scripts
(2). Installed from Python Package:              
C:\Users\paxue\AppData\Roaming\Python\Python310\Scripts

In the (2), path could be included in Python Package installation stage. There is a option to Add Path to Env Var. However, in the (1) stage, we don't have such kind of the config. 
We have 2 options here: If downloading from Microsoft Store, we ask user to adding this long path including "...PythonSoftwareFoundation.Python.3.10_qbz5n2kfra8p0..." into env var manually. Or we can remove this option, only suggest install from downloaded packages. From my perspective, I suggest to leaving the one which is under control by ourself. Since we can not easily control Microsoft Store behavior.

avylove:
@paxue I think this is an issue better fixed by talking with the team that maintains the Python package in the Microsoft Store. It's likely this is an issue that affects multiple people working with multiple projects. @zooba, do you know the best way to approach this?

For now, you can work around this in 3 ways:

1. Use `python -m lisa` instead of the `lisa` command
2. Use a virtual environment, either your own, or you can run `nox -vs dev` from the LISA repo to install the LISA development virtual environment
3. Automate adding the Scripts directory to the path. In powershell you can obtain the directory with `Join-Path -Path (Split-Path -Path (python -m site --user-site) -Parent) -ChildPath Scripts`

In the above examples I used `python` generically, but you may need to specify a specific Python if you have multiple installed.

The main benefit of using the Microsoft Store package is it will automatically be updated with bug and security fixes.

squirrelsc:
> @paxue I think this is an issue better fixed by talking with the team that maintains the Python package in the Microsoft Store. It's likely this is an issue that affects multiple people working with multiple projects. @zooba, do you know the best way to approach this?
> 
> For now, you can work around this in 3 ways:
> 
> 1. Use `python -m lisa` instead of the `lisa` command
> 2. Use a virtual environment, either your own, or you can run `nox -vs dev` from the LISA repo to install the LISA development virtual environment
> 3. Automate adding the Scripts directory to the path. In powershell you can obtain the directory with `Join-Path -Path (Split-Path -Path (python -m site --user-site) -Parent) -ChildPath Scripts`
> 
> In the above examples I used `python` generically, but you may need to specify a specific Python if you have multiple installed.
> 
> The main benefit of using the Microsoft Store package is it will automatically be updated with bug and security fixes.

@paxue is fixing the document to unblock others, who want to install LISA. To fix the Python in the store, it can be tracked in other places.

avylove:
> @paxue is fixing the document to unblock others, who want to install LISA. To fix the Python in the store, it can be tracked in other places.

It just creates another problem and a potential security issue. Probably it would be better to just create a non-dev nox session if this is a common use case, but I'm guessing in most cases outside of a pipeline, the dev environment is preferred.

squirrelsc:
> > @paxue is fixing the document to unblock others, who want to install LISA. To fix the Python in the store, it can be tracked in other places.
> 
> It just creates another problem and a potential security issue. Probably it would be better to just create a non-dev nox session if this is a common use case, but I'm guessing in most cases outside of a pipeline, the dev environment is preferred.

What's the problem created? Currently, there are two options on Python installation, two options on LISA installation. So 4 combinations here. That's why paxue meet the problem with an untested path. It needs to be simplified, and one-shot success on installation.

zooba:
I would strongly recommend you don't rely on PATH manipulations that have been done by other installers (such as Python's installer). They are off by default because they are unreliable.

Suggest that your users run with `python -m lisa` instead of just `lisa`, and then no problem will occur. You're already telling them to use `python -m pip`, so being consistent will ensure that everything works.

Right now, you're simply running a risk of obscure issues arising. Consistency is best, and `python -m` is the recommended place to be consistent.

paxue:
> 

avylove:
> What's the problem created? Currently, there are two options on Python installation, two options on LISA installation. So 4 combinations here. That's why paxue meet the problem with an untested path. It needs to be simplified, and one-shot success on installation.

The problem is the software doesn't get updated automatically unless you use a managed version like the Microsoft Store package. If you don't want to recommend any version, that's fine, but then you'd probably want to link to the [Python on Windows](https://docs.python.org/3/using/windows.html) page.

However it seems the actual problem is PATH, not the installer. Pip tells you the path and warns you when it's in PATH on install, so that could be sufficient. But, building on what @zooba said, it's probably a better idea to recommend `python -m lisa` or use a virtual environment.

squirrelsc (approved):

PR #2655, opened 2023-03-10
runner: add raise identifier

squirrelsc:
No sure why last time this line is removed. It looks like an incident. If not to raise the identifier, LISA printed the test results as normal, so some fatal errors are not easy to discover.

### DIFF
--- a/lisa/runner.py
+++ b/lisa/runner.py
@@ -224,6 +224,7 @@ class RootRunner(Action):
                 "canceling runner due to exception", exc_info=identifier
             )
             cancel()
+            raise identifier
         finally:
             self._cleanup()
 


LiliDeng (approved):

squirrelsc:
@LiliDeng I recalled the reason, it may need to wait other threads finish cleanup, so not to raise exception here. I will double confirm the behavior to prevent regressions.

squirrelsc (commented):
Need to double check what's impact on other threads, if raise exception here.

squirrelsc:
@LiliDeng Let's merge it to see if any impact. Current behavior is not straightforward for test run failures on notifier and other components.

PR #2659, opened 2023-03-14
Get the current user from node connection info in verify_no_pre_exist_users

lubaihua33:
For Bitnami images, the bitnami service will add changes for the current admin user after cloud-init adds this user. The output of the command "whoami" is not the admin user we specified in the ARM template. This will lead to reporting a false failure.

The bitnami.service changes the uid of the current user as 1000 which user 'bitnami' already has. From https://github.com/coreutils/coreutils/blob/master/src/whoami.c, whoami gets the name from EUID which is 1000. Then the output is 'bitnami' rather than the current user. 

"who -q" can get the login user, but For Debian images, they always have root in the output.

This PR modifies the code to get the username from the connection info of the node.

### DIFF
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -35,7 +35,7 @@ from lisa.operating_system import (
 )
 from lisa.sut_orchestrator import AZURE, READY
 from lisa.sut_orchestrator.azure.features import AzureDiskOptionSettings
-from lisa.tools import Cat, Dmesg, Ls, Lsblk, Lscpu, Pgrep, Ssh, Whoami
+from lisa.tools import Cat, Dmesg, Ls, Lsblk, Lscpu, Pgrep, Ssh
 from lisa.util import (
     LisaException,
     PassedException,
@@ -901,7 +901,13 @@ class AzureImageStandard(TestSuite):
             r" \\\"root\\\".\';echo;sleep .*\"",
             re.M,
         )
-        current_user = node.tools[Whoami].get_username()
+        # For Bitnami images, the bitnami.service changes the uid of current user as
+        # 1000 which user 'bitnami' already has.
+        # From https://github.com/coreutils/coreutils/blob/master/src/whoami.c, whoami
+        # gets name from EUID which is 1000. Then the output is 'bitnami' rather than
+        # the current user. So modified to get the admin user from node connection_info
+        remote_node = cast(RemoteNode, node)
+        current_user = str(remote_node.connection_info.get("username"))
         cat = node.tools[Cat]
         if isinstance(node.os, FreeBSD):
             shadow_file = "/etc/master.passwd"


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -901,7 +901,11 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
             r" \\\"root\\\".\';echo;sleep .*\"",
             re.M,
         )
-        current_user = node.tools[Whoami].get_username()
+        # For Bitnami images, the bitnami service will add changes for the current admin
+        # user. The output of the command "whoami" is not the logged user. So we get the
+        # username from node connetion info.
+        remote_node = cast(RemoteNode, node)
+        current_user = remote_node._connection_info.username
```
squirrelsc: Please double check the mechanism, make sure this is the right way.

lubaihua33: Yes, remote_node._connection_info.username is the admin user that is set in the arm template. I have run it on about 1000 images. That works well.

squirrelsc: I mean, if it's right to check by original name, instead of the name, which detects from the system. Which is the expected behavior.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -901,7 +901,13 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
             re.M,
         )
-        current_user = node.tools[Whoami].get_username()
+        # For Bitnami images, the bitnami.service changes the uid of current user as
+        # 1000 which user 'bitnami' already has.
+        # From https://github.com/coreutils/coreutils/blob/master/src/whoami.c, whoami
+        # gets name from EUID which is 1000. Then the output is 'bitnami' rather than
+        # the current user. So modified to use remote_node._connection_info.username
+        remote_node = cast(RemoteNode, node)
+        current_user = remote_node._connection_info.username
```
squirrelsc: Another thing is, it needs to read user name from `connection_info`, not `_connection_info`. The `_connection_info` is a private member.

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: For most images, the name detected by the system is the same as the original name. For now, only the name of bitnami image detected by the system is another name "bitnami" instead of "lisatest". Following the logic of this case, the "current_user" should be "lisatest" not "bitnami". So checking by the original name is expected.

This test case aims to detect if the authorized_keys files or passwords of previous users existed in the image. It needs to exclude the user ("lisatest" by default) which Lisa adds for the connection. (This user is the original admin user that sets in the arm template. After the VM boot up, this user and its authorized_keys are added by cloud-init.)

Before this fix, for bitnami images,  following the logic of this case,  "lisatest" is considered as the pre-existing user.  authorized_keys file is detected under /home/lisatest/.ssh/, then this case reports a false failed with the message "/home/lisatest/.ssh/authorized_keys is detected". 

PS:
Why the output of "whoami" is "bitnami" rather than "lisatest"? The bitnami.service changes the UID of "lisatest" as the UID 1000 which the name "bitnami" already has. "whoami" gets the name according to the UID, then the output is "bitnami".

squirrelsc (approved):

microsoft/testsuites/core/azure_image_standard.py
squirrelsc: Thanks for detail explanation!

PR #2660, opened 2023-03-15
Forcing 'use_new_environment=True' for all xfstests

SRIKKANTH:


LiliDeng (approved):

squirrelsc:
What's the reason to mark them use new environment? xfstest shouldn't related to environment. The use new environment should be set on cases like check pre existing users, bash history.

squirrelsc:
The PR description doesn't write reasons. The commit message is not meaningful even.

PR #2662, opened 2023-03-15
make before_suite/after_suite to deprecated

LiliDeng:


squirrelsc:
Please make sure it's not used in our code.

lisa/testsuite.py
```diff
@@ -539,10 +540,18 @@ def __init__(
...
-        ...
+        warnings.warn(
+            "before_suite is deprecated, use before_case instead",
+            DeprecationWarning,
+            stacklevel=2,
+        )
 
     def after_suite(self, log: Logger, **kwargs: Any) -> None:
-        ...
+        warnings.warn(
```
squirrelsc: deprecate is an annotation on methods, so it can raise error when writing code, instead of runtime.

LiliDeng:
> Please make sure it's not used in our code.

updated

lisa/testsuite.py
LiliDeng: I didn't figure out a way that can give warning or error when user is writing code without running the code yet.

squirrelsc: I just found out the "deprecated" is not a standard lib. We can continue to use `warnings.warn`.

squirrelsc (approved):

PR #2665, opened 2023-03-16
suppress error message when modinfo not find

LiliDeng:


lisa/operating_system.py
```diff
@@ -379,9 +379,12 @@ def capture_system_information(self, saved_path: Path) -> None:
             "uptime -s || last reboot -F | head -1 | awk '{print $9,$6,$7,$8}'",
             shell=True,
         ).save_stdout_to_file(saved_path / "uptime.txt")
-        self._node.execute("modinfo hv_netvsc").save_stdout_to_file(
-            saved_path / "modinfo-hv_netvsc.txt"
-        )
+        self._node.execute(
+            "modinfo hv_netvsc",
+            no_error_log=True,
+            no_debug_log=True,
```
squirrelsc: I think no error/info log should be enough. The debug log can keep.

LiliDeng: if this logic is correct, I have to make no_debug_log as True
```
        if no_debug_log:
            stdout_level = logging.NOTSET
        elif no_info_log:
            stdout_level = logging. DEBUG
```

squirrelsc (approved):

PR #2666, opened 2023-03-16
Improve documentation for keep environment

dsrivastavv:


docs/run_test/platform.rst
```diff
@@ -109,6 +109,39 @@ The advantage is that it can run all test cases of Azure. The shortage
 is that the VM name is fixed, and it should be node-0, so each resource
 group can put only one VM.
 
+To use existing deployment, follow the steps below:
+
+1. Start a run with the variable values set to following in the runbook:
```
squirrelsc, line 115: Please make it's supported in Azure.yml, and use command line to control those variables. That's easier to use and explain.

squirrelsc (approved):

LiliDeng:
there are some special cases here, if the VM provision time out, the keep option doesn't work, the resource group will be deleted automatically by the backend service unless we add the delete lock before deletion happens.

squirrelsc:
> there are some special cases here, if the VM provision time out, the keep option doesn't work, the resource group will be deleted automatically by the backend service unless we add the delete lock before deletion happens.

That's fine. It covers most scenarios.

PR #2667, opened 2023-03-17
Raise LisaException if exception is not "SSH session not active"

lubaihua33:
If the exception is "SSH session not active", then retry. If not, raise lisaexception to avoid useless retry which wastes time. For example,  if it is "cannot connect to TCP port", every retry will spend 300s until timeout.

When encountering the "cannot connect to TCP port" exception, it is not clear if the image has an issue. We can set the result as "Attempted" if the error message contains this exception after several times reloading the module.

### DIFF
--- a/microsoft/testsuites/network/netinterface.py
+++ b/microsoft/testsuites/network/netinterface.py
@@ -56,22 +56,43 @@ class NetInterface(TestSuite):
         network_interface_feature = node.features[NetworkInterface]
         # Test loading and unloading netvsc driver
         test_count = 0
-        while test_count < self.NETVSC_RELOAD_TEST_COUNT - 1:
+        while test_count < self.NETVSC_RELOAD_TEST_COUNT:
             test_count += 1
             # Unload and load hv_netvsc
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/network/netinterface.py
```diff
@@ -56,22 +56,43 @@ def validate_netvsc_reload(self, node: Node, log_path: Path) -> None:
...
                     saved_path=log_path, stage="after_reload_netvsc", force_run=True
                 )
-                node.close()
-        # Run the last time
-        network_interface_feature.reload_module()
+                if str(identifier) == "SSH session not active":
+                    node.log.debug(
+                        f"This exception '{identifier}' is ignorable. Try again"
+                    )
+                    node.close()
```
LiliDeng, line 84: where to check the panic?

lubaihua33: line 76-79 is checking the panic.

LiliDeng (approved):

PR #2669, opened 2023-03-20
Ignore missing marketplace info

kamalca:
This will allow private and preview marketplace images to still deploy.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -18,7 +18,7 @@ from threading import Lock
 from types import SimpleNamespace
 from typing import Any, Dict, List, Optional, Set, Tuple, Type, Union, cast
 
-from azure.core.exceptions import HttpResponseError
+from azure.core.exceptions import HttpResponseError, ResourceNotFoundError
 from azure.identity import DefaultAzureCredential
 from azure.mgmt.compute.models import (  # type: ignore
     GalleryImage,
@@ -1213,14 +1213,16 @@ class AzurePlatform(Platform):
                 azure_node_runbook.location, azure_node_runbook.marketplace
             )
             # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
-            azure_node_runbook.hyperv_generation = _get_vhd_generation(image_info)
+            if image_info:
+                azure_node_runbook.hyperv_generation = _get_vhd_generation(image_info)
 
             # retrieve the os type for arm template.
-            if azure_node_runbook.is_linux is None:
-                if image_info.os_disk_image.operating_system == "Windows":
-                    azure_node_runbook.is_linux = False
-                else:
-                    azure_node_runbook.is_linux = True
+            if (
+                azure_node_runbook.is_linux is None
+                and image_info
+                and image_info.os_disk_image.operating_system == "Windows"
+            ):
+                azure_node_runbook.is_linux = False
         elif azure_node_runbook.shared_gallery:
             azure_node_runbook.hyperv_generation = _get_gallery_image_generation(
                 self._get_detailed_sig(azure_node_runbook.shared_gallery)
@@ -1270,22 +1272,23 @@ class AzurePlatform(Platform):
             image_info = self._get_image_info(
                 arm_parameters.location, arm_parameters.marketplace
             )
-            arm_parameters.osdisk_size_in_gb = max(
-                arm_parameters.osdisk_size_in_gb,
-                image_info.os_disk_image.additional_properties.get("sizeInGb", 0),
-            )
-            if not arm_parameters.purchase_plan and image_info.plan:
-                # expand values for lru cache
-                plan_name = image_info.plan.name
-                plan_product = image_info.plan.product
-                plan_publisher = image_info.plan.publisher
-                # accept the default purchase plan automatically.
-                arm_parameters.purchase_plan = self._process_marketplace_image_plan(
-                    marketplace=arm_parameters.marketplace,
-                    plan_name=plan_name,
-                    plan_product=plan_product,
-                    plan_publisher=plan_publisher,
+            if image_info:
+                arm_parameters.osdisk_size_in_gb = max(
+                    arm_parameters.osdisk_size_in_gb,
+                    image_info.os_disk_image.additional_properties.get("sizeInGb", 0),
                 )
+                if not arm_parameters.purchase_plan and image_info.plan:
+                    # expand values for lru cache
+                    plan_name = image_info.plan.name
+                    plan_product = image_info.plan.product
+                    plan_publisher = image_info.plan.publisher
+                    # accept the default purchase plan automatically.
+                    arm_parameters.purchase_plan = self._process_marketplace_image_plan(
+                        marketplace=arm_parameters.marketplace,
+                        plan_name=plan_name,
+                        plan_product=plan_product,
+                        plan_publisher=plan_publisher,
+                    )
 
         # Set disk type
         assert capability.disk, "node space must have disk defined."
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1193,15 +1193,23 @@ def _create_node_runbook(
...
-            # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
-            azure_node_runbook.hyperv_generation = _get_vhd_generation(image_info)
+            image_info = None
+            try:
+                image_info = self._get_image_info(
+                    azure_node_runbook.location, azure_node_runbook.marketplace
+                )
+                # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
+                azure_node_runbook.hyperv_generation = _get_vhd_generation(image_info)
+            except Exception as e:
```
avylove: Having a broad exception like this is dangerous, it can mask all sorts of issues. Can you use the specific exceptions this will raise?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1193,15 +1193,23 @@ def _create_node_runbook(
...
-            )
-            # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
-            azure_node_runbook.hyperv_generation = _get_vhd_generation(image_info)
+            image_info = None
+            try:
+                image_info = self._get_image_info(
+                    azure_node_runbook.location, azure_node_runbook.marketplace
+                )
+                # HyperVGenerationTypes return "V1"/"V2", so we need to strip "V"
+                azure_node_runbook.hyperv_generation = _get_vhd_generation(image_info)
```
avylove, line 1217: Does this need to be in the try block? If self._get_image_info() doesn't raise an exception, will this?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1193,15 +1193,23 @@ def _create_node_runbook(
...
+            except Exception as e:
+                self._log.error(e)
+                self._log.error("Could not get marketplace information")
```
squirrelsc: Use one line to log exception and the message together.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1247,29 +1255,32 @@ def _create_node_arm_parameters(
...
+                    # expand values for lru cache
+                    plan_name = image_info.plan.name
+                    plan_product = image_info.plan.product
+                    plan_publisher = image_info.plan.publisher
+                    # accept the default purchase plan automatically.
+                    arm_parameters.purchase_plan = self._process_marketplace_image_plan(
+                        marketplace=arm_parameters.marketplace,
+                        plan_name=plan_name,
+                        plan_product=plan_product,
+                        plan_publisher=plan_publisher,
```
avylove, line 1290:
```suggestion
                    # accept the default purchase plan automatically.
                    arm_parameters.purchase_plan = self._process_marketplace_image_plan(
                        marketplace=arm_parameters.marketplace,
                        plan_name=image_info.plan.name,
                        plan_product=image_info.plan.product,
                        plan_publisher=image_info.plan.publisher,
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1247,29 +1255,32 @@ def _create_node_arm_parameters(
...
+                    )
+            except Exception as e:
```
avylove: Can you make this exception more specific too?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1644,19 +1655,28 @@ def _resolve_marketplace_image(
...
-            # any one should be the same to get purchase plan
-            new_marketplace.version = versioned_images[-1].name
+                try:
+                    versioned_images = compute_client.virtual_machine_images.list(
+                        location=location,
+                        publisher_name=marketplace.publisher,
+                        offer=marketplace.offer,
+                        skus=marketplace.sku,
+                    )
+                    if 0 == len(versioned_images):
```
avylove, line 1699:
```suggestion
                    if not versioned_images:
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1644,19 +1655,28 @@ def _resolve_marketplace_image(
...
+                        self._log.error(
+                            f"cannot find any version of image {marketplace.publisher} "
+                            f"{marketplace.offer} {marketplace.sku} in {location}"
+                        )
+                    else:
+                        # any one should be the same to get purchase plan
+                        new_marketplace.version = versioned_images[-1].name
+                except Exception as e:
```
avylove: Another bare exception

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2042,26 +2062,30 @@ def _generate_data_disks(
...
+                for default_data_disk in marketplace.data_disk_images:
+                    data_disks.append(
+                        DataDiskSchema(
+                            node.capability.disk.data_disk_caching_type,
+                            default_data_disk.additional_properties["sizeInGb"],
+                            azure_node_runbook.disk_type,
+                            DataDiskCreateOption.DATADISK_CREATE_OPTION_TYPE_FROM_IMAGE,
+                        )
+                    )
+            except Exception as e:
```
avylove: Another bare exception

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2483,17 +2507,20 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
                     image_info = self._get_image_info(
                         location, azure_runbook.marketplace
                     )
+                    generation = _get_vhd_generation(image_info)
+                    node_space.features.add(
```
squirrelsc: Make sure the test won't fail, if missing the two features.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Instead of add it every place the _get_image_info called. How about capturing the exception in this method? Other place just check if it's None or not. If it's None then skip some logic.

kamalca:
@squirrelsc Can I get some feedback on this?

lisa/sut_orchestrator/azure/platform_.py
kamalca: Private marketplace images will work like VHDs since there is no way to get info.

VHDs are already fine without defining architecture and hyper-v gen defaults to 1. If you want to use gen2 features on private marketplace image, you have to specify in the runbook that it is gen2 (similar to VHDs)

squirrelsc:
@LiliDeng Looks good to me.

LiliDeng:
@kamalca please fix check errors.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1213,14 +1213,16 @@ def _create_node_runbook(
...
-                if image_info.os_disk_image.operating_system == "Windows":
-                    azure_node_runbook.is_linux = False
-                else:
-                    azure_node_runbook.is_linux = True
+            if (
+                azure_node_runbook.is_linux is None
+                and image_info
+                and image_info.os_disk_image.operating_system == "Windows"
+            ):
+                azure_node_runbook.is_linux = False
```
squirrelsc, line 1225: Why remove the `azure_node_runbook.is_linux = False`?

kamalca: I removed the lines
``` Python
else:
    azure_node_runbook.is_linux = True
```
because Flake8 was complaining the logic was too complicated and the default value is already True.
See later in the function:
``` Python
if azure_node_runbook.is_linux is None:
    # fill it default value
    azure_node_runbook.is_linux = True
```

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2670, opened 2023-03-21
[new case] verify_services_state

LiliDeng:


lisa/base_tools/service.py
```diff
@@ -92,6 +98,9 @@ def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
 
 
 class Systemctl(Tool):
+    __STATE_PATTERN = re.compile(r"^(\s+)State:(\s+)(?P<state>.*)", re.M)
+    __NOT_FOUND = re.compile(r"not found", re.M)
```
avylove, line 102: These should be lower case, only globals are all caps.

microsoft/testsuites/network/sriov.py
```diff
@@ -67,6 +69,27 @@ def before_case(self, log: Logger, **kwargs: Any) -> None:
...
+        requirement=simple_requirement(
+            network_interface=features.Sriov(),
+        ),
+    )
+    def verify_services_state(self, environment: Environment) -> None:
+        try:
+            for node in environment.nodes.list():
+                assert_that(node.tools[Systemctl].state()).is_equal_to("running")
+        except UnsupportedDistroException as identifier:
+            raise SkippedException(identifier)
```
avylove: I'm not sure where `identifier` came from, it's a little confusing. The typical naming convention is `e` or sometimes `err`. You should also use exception chaining in cases like this.

```suggestion
        except UnsupportedDistroException as e:
            raise SkippedException(e) from e
```

lisa/base_tools/service.py
```diff
@@ -116,6 +125,23 @@ def enable_service(self, name: str) -> None:
...
+            "status --no-page",
+            shell=True,
+            sudo=True,
+            force_run=True,
+        )
+        if cmd_result.exit_code != 0 and self.__NOT_FOUND.findall(cmd_result.stdout):
+            raise UnsupportedDistroException(
+                os=self.node.os, message=f"command not find {self.command}"
+            )
+        group = find_patterns_groups_in_lines(
```
squirrelsc: if there is only one group, try `find_patterns_group_in_lines`. It doesn't need to use 2-dim resutls.

microsoft/testsuites/network/sriov.py
LiliDeng: thanks, updated

lisa/base_tools/service.py
LiliDeng: updated

squirrelsc (approved):

PR #2671, opened 2023-03-21
Raise skip on images which aren't receiving floppy fixes

mcgov:
Raise a SKIP on unmaintained images which aren't going to receive changes to remove the floppy driver.

### DIFF
--- a/microsoft/testsuites/core/floppy.py
+++ b/microsoft/testsuites/core/floppy.py
@@ -41,7 +41,7 @@ class Floppy(TestSuite):
         is_centos = isinstance(node.os, CentOs)
         if is_centos and os_version < "7.8.0":
             raise SkippedException(
-                "CentOS below 7.8 are not receiving fixes for this issue."
+                "CentOS <= 7.7 are not receiving fixes for block listing the floppy module."
             )
         modprobe = node.tools[Modprobe]
 
--- a/microsoft/testsuites/core/floppy.py
+++ b/microsoft/testsuites/core/floppy.py
@@ -3,7 +3,14 @@
 
 from assertpy import assert_that
 
-from lisa import RemoteNode, TestCaseMetadata, TestSuite, TestSuiteMetadata
+from lisa import (
+    RemoteNode,
+    SkippedException,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+)
+from lisa.operating_system import CentOs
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/floppy.py
```diff
@@ -30,6 +37,14 @@ class Floppy(TestSuite):
         priority=1,
     )
     def check_floppy_module(self, node: RemoteNode) -> None:
+        if (
+            isinstance(node.os, CentOs)
+            and node.os.information.version.major < 7
+            and node.os.information.version.major < 7
```
squirrelsc: Should it be minor? And it can be compared like `version < "7.7.0"`. BTW, it means the 7.7 is out of life, it should be `version <= "7.7.0"`.

microsoft/testsuites/core/floppy.py
```diff
@@ -30,6 +37,12 @@ class Floppy(TestSuite):
         priority=1,
     )
     def check_floppy_module(self, node: RemoteNode) -> None:
+        os_version = node.os.information.version
+        is_centos = isinstance(node.os, CentOs)
+        if is_centos and os_version <= "7.7.0":
+            raise SkippedException(
+                "CentOS < 7.7 are not receiving fixes for this issue."
```
squirrelsc: The comments should be "<=7.7" as well. But how about "7.7.1"? Does `< "7.8.0"` is more accurate?

microsoft/testsuites/core/floppy.py
```diff
@@ -30,6 +37,12 @@ class Floppy(TestSuite):
         priority=1,
     )
     def verify_floppy_module_is_blacklisted(self, node: RemoteNode) -> None:
+        os_version = node.os.information.version
+        is_centos = isinstance(node.os, CentOs)
+        if is_centos and os_version < "7.8.0":
+            raise SkippedException(
+                "CentOS below 7.8 are not receiving fixes for this issue."
```
squirrelsc: 1. Just double check, if 7.8 should skip or not? The `os_version < "7.8.0"` means 7.7.x are skipped, but 7.8 is not. So the message is clearer, if says "Cent OS 7.7.x and below ...".
2. "this issue" is not clear in the exception message. Please explain more.

squirrelsc (dismissed):

squirrelsc (approved):

PR #2673, opened 2023-03-22
Bugfix: add default size for property of os disks (not always present)

mcgov:
Tested in BVT this week.
Provide a default size of 0 for a property which some ARM images don't seem to have available.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1256,7 +1256,7 @@ class AzurePlatform(Platform):
             )
             arm_parameters.osdisk_size_in_gb = max(
                 arm_parameters.osdisk_size_in_gb,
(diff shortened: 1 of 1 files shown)


LiliDeng (approved):

LiliDeng (changes requested):
I am double checking why see pass and fail against the same image

mcgov:
@squirrelsc Do you know if this size field is supposed to be guaranteed to always be present? We hadn't seen this bug before and the last change in this section is pretty old. Could a change elsewhere be resulting in this size field not being populated?

LiliDeng (approved):

squirrelsc:
> @squirrelsc Do you know if this size field is supposed to be guaranteed to always be present? We hadn't seen this bug before and the last change in this section is pretty old. Could a change elsewhere be resulting in this size field not being populated?

We shouldn't assume it's always there. But if it's not, it should be misconfigured, and fixed.

PR #2674, opened 2023-03-24
print deprecationwarning if before_suite and after_suite are used

LiliDeng:


lisa/testsuite.py
```diff
@@ -584,15 +566,20 @@ def start(
...
-        suite_log = self.__log
-        (
-            is_suite_continue,
-            suite_error_message,
-            suite_error_stacktrace,
-        ) = self.__suite_method(
-            self.before_suite, test_kwargs=test_kwargs, log=suite_log
-        )
+        if hasattr(self, "before_suite"):
+            suite_log = self.__log
```
squirrelsc: the suite_log is used in other places, it should be initialized always.

LiliDeng: updated

squirrelsc: This line should be removed now.

LiliDeng: removed

squirrelsc (approved):

PR #2675, opened 2023-03-25
Fix serial console testing

dsrivastavv:


microsoft/testsuites/core/serial_console.py
```diff
@@ -33,9 +35,19 @@ class SerialConsoleSuite(TestSuite):
...
         command = "echo back"
         serial_console = node.features[SerialConsole]
-        _ = serial_console.read()
-        serial_console.write(command)
-        output = serial_console.read()
+
+        # retry to read serial console output, because it may not be ready
+        for _ in range(180):
+            _ = serial_console.read()
+            serial_console.write(command)
```
LiliDeng, line 42: do we need write in each loop?

LiliDeng (approved):

PR #2677, opened 2023-03-27
xdp: install xdpdump firstly to by pass some skip cases

LiliDeng:
We can invoke get_xdpdump firstly, since for some distro versions, they don't support XDP, but in current design, it also enters into install pktgen logic. This is the case for centos 79 arm64 image. @squirrelsc thoughts?

image

### DIFF
--- a/microsoft/testsuites/xdp/performance.py
+++ b/microsoft/testsuites/xdp/performance.py
@@ -362,15 +362,15 @@ class XdpPerformance(TestSuite):
 
         # install pktgen on sender, and xdpdump on receiver.
         try:
-            tools: List[Any] = run_in_parallel(
-                [partial(sender.tools.get, Pktgen), partial(get_xdpdump, receiver)]
-            )
+            tools: List[Any] = []
+            tools.append(get_xdpdump(receiver))
+            tools.append(sender.tools[Pktgen])
         except UnsupportedKernelException as identifier:
             raise SkippedException(identifier)
 
         # type annotations
-        pktgen: Pktgen = tools[0]
-        xdpdump: XdpDump = tools[1]
+        xdpdump: XdpDump = tools[0]
+        pktgen: Pktgen = tools[1]
 
         sender_nic = sender.nics.get_nic_by_index(1)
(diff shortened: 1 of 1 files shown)


squirrelsc:
> We can invoke get_xdpdump firstly, since for some distro versions, they don't support XDP, but in current design, it also enters into install pktgen logic. This is the case for centos 79 arm64 image. @squirrelsc thoughts?
> 
> image

Looks good to me.

microsoft/testsuites/xdp/pktgen.py
```diff
@@ -195,6 +198,12 @@ def _install_fedora(self) -> None:
                 "/all/Packages/k/kernel-modules-internal-"
                 f"{'.'.join(parts[0:3])}-{'.'.join(parts[3:-1])}.{parts[-1]}.rpm"
             ),
+            (
+                "http://20.151.28.143/openlogic/"
```
squirrelsc: Where is the IP from? Can we use a domain name? IP is not reliable.

LiliDeng: this is the only place I can find for kernel-modules-internal-5.4.155-200_ol008.el7.aarch64.rpm of centos 7.9 arm image, we can move all of them which we can find to our storage account

squirrelsc: Let's skip this kernel, if it's not an official place. It's not good to keep them in our storage account.

LiliDeng: update code to install tools in sequence, this image will be skip after installing xdump.
use image redhat rhel 9_1 9.1.2022112113 to test 
- install tools in parallel, total running time 3 mins 52 seconds
- install tools in sequence, total running time is 4 mins 30 seconds

squirrelsc: Is it another thing? I'm ok to install in parallel. I mean, if the package is got from an IP address, we cannot use it.

squirrelsc (approved):

PR #2678, opened 2023-03-28
CVM: Support for CMK through DES

kamalca:


lisa/features/security_profile.py
```diff
@@ -59,6 +59,12 @@ class SecurityProfileSettings(schema.FeatureSettings):
         ),
     )
     encrypt_disk: bool = field(default=False)
+    disk_encryption_set_id: str = field(
```
squirrelsc: If it's an Azure specified setting, it should be defined in Azure's feature settings.

kamalca: I'm having trouble moving this field. _call_requirement_method() needs access to the field but moving the method to Azure also doesn't seem to work. Please advise.

squirrelsc: Azure should inherit this method, and do extra logic after called the super one.

kamalca: Fixed

squirrelsc (approved):

PR #2679, opened 2023-03-29
Enable the use of existing azure vnet

bfjelds:
# Scenario 
Enable LISA to create Azure virtual machines that utilize an existing virtual network and subnet.

This will allow a test runner (that is attached to an existing virtual network) to use LISA to create and communicate with azure virtual machines (attached to the same virtual network).

# Changes
* Update the LISA ARM template to allow for the use of an existing vnet/subnet.
* Enable the use of existing vnet/subnet through a runbook

### DIFF
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -96,11 +96,15 @@
         }
     },
     "variables": {
-        "virtual_network_name": "lisa-virtualNetwork",
-        "vnet_id": "[resourceId('Microsoft.Network/virtualNetworks/', variables('virtual_network_name'))]",
+        
+        "virtual_network_name": "bfjelds-lisa-tests-vnet",
+        "virtual_network_resource_group": "bfjelds-lisa-tests",
+        
+        "vnet_id": "[resourceId(variables('virtual_network_resource_group'), 'Microsoft.Network/virtualNetworks/', variables('virtual_network_name'))]",
         "node_count": "[length(parameters('nodes'))]",
         "availability_set_name": "lisa-availabilitySet",
-        "subnet_prefix": "lisa-subnet-"
+        "subnet_name": "kva-functional-test-subnet",
+        "subnet_ref": "[resourceId(variables('virtual_network_resource_group'), 'Microsoft.Network/virtualNetworks/subnets', variables('virtual_network_name'), variables('subnet_name'))]"
     },
     "resources": [
         {
@@ -129,31 +133,6 @@
             }
         },
         {
-            "apiVersion": "2020-05-01",
-            "type": "Microsoft.Network/virtualNetworks",
-            "name": "[variables('virtual_network_name')]",
-            "location": "[parameters('location')]",
-            "properties": {
-                "addressSpace": {
-                    "addressPrefixes": [
-                        "10.0.0.0/16"
-                    ]
-                },
-                "copy": [
-                    {
-                        "name": "subnets",
-                        "count": "[parameters('subnet_count')]",
-                        "input": {
-                            "name": "[concat(variables('subnet_prefix'), copyIndex('subnets'))]",
-                            "properties": {
-                                "addressPrefix": "[concat('10.0.', copyIndex('subnets'), '.0/24')]"
-                            }
-                        }
-                    }
-                ]
-            }
-        },
-        {
             "type": "Microsoft.Resources/deployments",
             "apiVersion": "2019-10-01",
             "copy": {
@@ -162,8 +141,7 @@
             },
             "name": "[concat(parameters('nodes')[copyIndex('nicCopy')]['name'],'-nics')]",
             "dependsOn": [
-                "[resourceId('Microsoft.Network/publicIPAddresses/', concat(parameters('nodes')[copyIndex('nicCopy')]['name'],'-public-ip'))]",
-                "[variables('vnet_id')]"
+                "[resourceId('Microsoft.Network/publicIPAddresses/', concat(parameters('nodes')[copyIndex('nicCopy')]['name'],'-public-ip'))]"
             ],
             "properties": {
                 "expressionEvaluationOptions": {
@@ -179,11 +157,8 @@
                     "location": {
                         "value": "[parameters('location')]"
                     },
-                    "vnet_id": {
-                        "value": "[variables('vnet_id')]"
-                    },
-                    "subnet_prefix": {
-                        "value": "[variables('subnet_prefix')]"
+                    "subnet_ref": {
+                        "value": "[variables('subnet_ref')]"
                     },
                     "enable_sriov": {
                         "value": "[parameters('nodes')[copyIndex('nicCopy')]['enable_sriov']]"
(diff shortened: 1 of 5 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -256,6 +260,12 @@ class AzurePlatformSchema:
     vm_tags: Optional[Dict[str, Any]] = field(default=None)
     locations: Optional[Union[str, List[str]]] = field(default=None)
 
+    virtual_network_resource_group: str = field(default="")
+    virtual_network_name: str = field(default=AZURE_NEW_VIRTUAL_NETWORK)
+    subnet_prefix: str = field(default=AZURE_NEW_SUBNET_PREFIX)
+    use_existing_virtual_network: bool = field(default=False)
+    use_private_address_for_vm_connection: bool = field(default=False)
```
squirrelsc: The name is too long, how about `use_private_ip`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1241,9 +1251,16 @@ def load_environment(
         ) = get_primary_ip_addresses(
             platform, resource_group_name, vms_map[node_context.vm_name]
         )
+
+        connection_public_ip = node_context.public_ip_address
```
squirrelsc: Please submit a separated PR for private IP support.

lisa/sut_orchestrator/azure/platform_.py
bfjelds: updated in https://github.com/microsoft/lisa/pull/2682

lisa/sut_orchestrator/azure/common.py
bfjelds: created https://github.com/microsoft/lisa/pull/2682 for private IP support

lisa/sut_orchestrator/azure/common.py
```diff
@@ -74,6 +74,9 @@
     from .platform_ import AzurePlatform
 
 AZURE_SHARED_RG_NAME = "lisa_shared_resource"
+AZURE_NEW_VIRTUAL_NETWORK = "lisa-virtualNetwork"
```
squirrelsc: Name it like above `AZURE_VIRTUAL_NETWORK_NAME`.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -493,6 +496,12 @@ class AzureArmParameter:
     data_disks: List[DataDiskSchema] = field(default_factory=list)
     use_availability_sets: bool = False
     vm_tags: Dict[str, Any] = field(default_factory=dict)
+    
+    virtual_network_resource_group: str = ""
+    virtual_network_name: str = AZURE_NEW_VIRTUAL_NETWORK
+    subnet_prefix: str = AZURE_NEW_SUBNET_PREFIX
+    use_existing_virtual_network: bool = False
```
squirrelsc: Is the bool value needed?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -968,6 +980,14 @@ def _create_deployment_parameters(
         ]
         set_filtered_fields(self._azure_runbook, arm_parameters, copied_fields)
 
+        if self._azure_runbook.virtual_network_resource_group:
+            arm_parameters.virtual_network_resource_group = self._azure_runbook.virtual_network_resource_group
+        if self._azure_runbook.subnet_prefix:
+            arm_parameters.subnet_prefix = self._azure_runbook.subnet_prefix
```
squirrelsc: The subnet_prefix and virtual_network_name have default values in runbook, so it can be set always without checks.

lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -70,6 +70,34 @@
...
+        "use_existing_virtual_network": {
+            "type": "bool",
+            "defaultValue": false,
+            "metadata": {
+                "description": "flag to use an existing vnet"
+            }
+        },
+        "virtual_network_resource_group": {
+            "type": "string",
+            "defaultValue": "[resourceGroup().name]",
```
squirrelsc: Is the default value used?

lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -90,11 +118,10 @@
         }
     },
     "variables": {
-        "virtual_network_name": "lisa-virtualNetwork",
-        "vnet_id": "[resourceId('Microsoft.Network/virtualNetworks/', variables('virtual_network_name'))]",
+        "vnet_id": "[resourceId('Microsoft.Network/virtualNetworks/', parameters('virtual_network_name'))]",
         "node_count": "[length(parameters('nodes'))]",
         "availability_set_name": "lisa-availabilitySet",
-        "subnet_prefix": "lisa-subnet-"
+        "existing_subnet_ref": "[if(parameters('use_existing_virtual_network'), resourceId(parameters('virtual_network_resource_group'), 'Microsoft.Network/virtualNetworks/subnets', parameters('virtual_network_name'), parameters('subnet_prefix')), '')]"
```
squirrelsc, line 124: If `use_existing_virtual_network` is false, how the default virtual network and subnet prefix are applied?

lisa/sut_orchestrator/azure/arm_template.json
bfjelds: i had hoped it would be, but in practice it didn't.  i'll remove the default values from here.

lisa/sut_orchestrator/azure/common.py
bfjelds: i thought the value was needed.

this seems to work:
```
class WithValue:
    withValue: bool = False

    def Print(self):
        print(self.withValue)
    
WithValue().Print()
```

this complains that `AttributeError: 'WithoutValue' object has no attribute 'withoutValue'`
```
class WithoutValue:
    withoutValue: bool

    def Print(self):
        print(self.withoutValue)

WithoutValue().Print()
```

bfjelds: but am happy to be corrected :)

squirrelsc: I mean, the bool is used to check if the virtual_network_resource_group is set or not. But it can be checked by the virtual_network_resource_group is empty or not. So I think the additional bool value is not necessary.

lisa/sut_orchestrator/azure/common.py
bfjelds: done

lisa/sut_orchestrator/azure/platform_.py
bfjelds: thanks!

lisa/sut_orchestrator/azure/common.py
bfjelds: oh ... yeah, that can totally work!  

my thinking was that it would be better to be explicit rather than executing this codepath based on the side-effect of setting a resource group variable.

but i'm happy to do whatever you prefer!

squirrelsc: I want to make both runbook and arm template less parameters. They are many values already.

bfjelds: ok!

bfjelds: doen

squirrelsc:
@LiliDeng this PR looks good to me.

PR #2680, opened 2023-03-29
Configurable LTP version from LISA Runbook

adityagesh:
The below variable can be added to runbook to
override default LTP version used for LTP tests
  - name: ltp_tests_git_tag
    value: 20230127
    is_case_visible: true

### DIFF
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -49,7 +49,7 @@ class Ltp(Tool):
     _RESULT_LTP_ARCH_REGEX = re.compile(r"Machine Architecture: (.*)\s+")
 
     LTP_DIR_NAME = "ltp"
-    LTP_TESTS_GIT_TAG = "20200930"
+    LTP_TESTS_GIT_TAG = "20230127"
     LTP_GIT_URL = "https://github.com/linux-test-project/ltp.git"
     BUILD_REQUIRED_DISK_SIZE_IN_GB = 2
     LTP_RESULT_PATH = "/opt/ltp/ltp-results.log"
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -49,7 +49,7 @@ class Ltp(Tool):
     _RESULT_LTP_ARCH_REGEX = re.compile(r"Machine Architecture: (.*)\s+")
 
     LTP_DIR_NAME = "ltp"
-    LTP_TESTS_GIT_TAG = "20230127"
+    LTP_TESTS_GIT_TAG = "20200930"
     LTP_GIT_URL = "https://github.com/linux-test-project/ltp.git"
     BUILD_REQUIRED_DISK_SIZE_IN_GB = 2
     LTP_RESULT_PATH = "/opt/ltp/ltp-results.log"
@@ -68,6 +68,11 @@ class Ltp(Tool):
     @property
     def can_install(self) -> bool:
         return True
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        git_tag = kwargs.get("git_tag")
+        if git_tag:
+            self.LTP_TESTS_GIT_TAG = git_tag
 
     def run_test(
         self,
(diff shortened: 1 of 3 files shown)


LiliDeng:
@adityagesh make sure old distro works well with this tag, hope it doesn't have any compile issue

adityagesh:
@squirrelsc @LiliDeng can you please check if this is the correct approach

microsoft/testsuites/ltp/ltp.py
```diff
@@ -69,6 +69,11 @@ def dependencies(self) -> List[Type[Tool]]:
     def can_install(self) -> bool:
         return True
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
```
squirrelsc: Arguments can be passed in the construct method, instead of initializing.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -69,6 +69,11 @@ def dependencies(self) -> List[Type[Tool]]:
...
+        git_tag = kwargs.get("git_tag")
```
squirrelsc: Don't update the variables like `LTP_TESTS_GIT_TAG`, it's like consts. if it needs to use the value in other method, name it like `self._git_tag`.

```suggestion
        git_tag = kwargs.get("git_tag", self.LTP_TESTS_GIT_TAG)
```

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -57,6 +57,7 @@ def ltp_lite(
         # parse variables
         tests = variables.get("ltp_test", "")
         skip_tests = variables.get("ltp_skip_test", "")
+        ltp_tests_git_tag = variables.get("ltp_tests_git_tag")
```
squirrelsc: Give it a default value, or it will fail if not specified.

adityagesh: It would return None by default, right? Added "" similar to line 58,59

microsoft/testsuites/ltp/ltp.py
adityagesh: I did not understand, 
You're suggesting we initialize the value in __init__ ? It is recommended to not override init right?

microsoft/testsuites/ltp/ltp.py
adityagesh: updated

microsoft/testsuites/ltp/ltp.py
squirrelsc: It needs to overrite `__init__` and make sure the super methods is called. It's confusing to pass arguments to both `__init__` and `_initialize`.

microsoft/testsuites/ltp/ltpsuite.py
squirrelsc: It will fail if a value is not in dict. You can test it, after implemented.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -69,6 +69,10 @@ def dependencies(self) -> List[Type[Tool]]:
     def can_install(self) -> bool:
         return True
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        git_tag = kwargs.get("git_tag", "")
```
squirrelsc, line 73: The `git_tag` is only used to assign value, so one line is enough and simple.

```python
self._git_tag = kwargs.get("git_tag", "")
```

microsoft/testsuites/ltp/ltp.py
adityagesh: Updated

squirrelsc (approved):

microsoft/testsuites/ltp/ltp.py
```diff
@@ -49,7 +50,7 @@ class Ltp(Tool):
     _RESULT_LTP_ARCH_REGEX = re.compile(r"Machine Architecture: (.*)\s+")
 
     LTP_DIR_NAME = "ltp"
-    LTP_TESTS_GIT_TAG = "20200930"
+    DEFAULT_LTP_TESTS_GIT_TAG = "20200930"
```
squirrelsc, line 53: Even it's configurable, it needs to upgrade the default version too.

adityagesh: Can it be part of a different PR? I saw few timeouts and other failures while testing with few older images. Will need some time to test and triage with different images.

squirrelsc: If so, no problem.

PR #2681, opened 2023-03-29
Fix timeout in perf_nvme

adityagesh:
Creation of perf messages invokes multiple calls to environment.get_information() even though the environment details does not change. This leads to increased test execution time which might lead to timeout. 
This change reduces the number of calls to environment.get_information() from Fio tool.

The reduction in execution time will be observed for some duration and then will be changed for other tests/tools

### DIFF
--- a/lisa/environment.py
+++ b/lisa/environment.py
@@ -176,6 +176,7 @@ class Environment(ContextMixin, InitializableMixin):
         self.platform: Optional[Platform] = None
         self.log = get_logger("env", self.name)
         self.source_test_result: Optional[TestResult] = None
+        self._information_cache: Optional[Dict[str, str]] = None
 
         self._default_node: Optional[Node] = None
         self._is_dirty: bool = False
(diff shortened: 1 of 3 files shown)


squirrelsc:
Good point. We can add a flag called `force_run=False` on the get_information like tools.

adityagesh:
> Good point. We can add a flag called `force_run=False` on the get_information like tools.

Updated, set default value of force_run to be True. 
Also, cleared cache in _reset(). Is that sufficient?

lisa/environment.py
```diff
@@ -362,7 +365,7 @@ def _reset(self) -> None:
         self.runbook.reload_requirements()
 
         self.is_new = True
-
+        self._information_cache = {}
```
squirrelsc: Use `None`, so if nothing is found, it doesn't update every time.

adityagesh: Updated

lisa/environment.py
```diff
@@ -330,18 +331,20 @@ def create_node_from_requirement(
 
         return node
 
-    def get_information(self) -> Dict[str, str]:
-        final_information: Dict[str, str] = {}
+    def get_information(self, force_run: bool = True) -> Dict[str, str]:
+        if self._information_cache and not force_run:
```
squirrelsc: check like below, so if dict updated, but nothing got, it doesn't run again.

```python
        if self._information_cache is None and not force_run:
```

adityagesh: Yes, that makes sense. Updated.

adityagesh: Should I change the function return type to None|Dict[str,str]?

Some of the existing usages of this function expects Dict[str,str]

squirrelsc: No, the dict can keep the same returned type.

squirrelsc (dismissed):

squirrelsc (approved):

PR #2682, opened 2023-03-29
Enable the use of the private ip address

bfjelds:
# Scenario 
Enable LISA to create Azure virtual machines that utilize an existing virtual network and subnet.

This will allow a test runner (that is attached to an existing virtual network) to communicate over that virtual network with azure virtual machines (attached to the same virtual network).

# Changes
* Modify connection creation code to allow the use of the private ip
* Enable the use of private ip through a runbook


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1200,7 +1200,7 @@ def save_console_log(
 
 
 def load_environment(
-    platform: "AzurePlatform", resource_group_name: str, log: Logger
+    platform: "AzurePlatform", resource_group_name: str, use_private_ip: bool, log: Logger
```
squirrelsc, line 1203: Pass in the runbook directly, so it can be extended in future.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -255,6 +255,7 @@ class AzurePlatformSchema:
     availability_set_properties: Optional[Dict[str, Any]] = field(default=None)
     vm_tags: Optional[Dict[str, Any]] = field(default=None)
     locations: Optional[Union[str, List[str]]] = field(default=None)
+    use_private_ip: bool = field(default=False)
```
squirrelsc: It can be a flag on RemoteNode in `schema.py`, so it can be handled in set_connection_info of node.py.

squirrelsc:
Good feature, please also update related documents.

squirrelsc:
Please fix CI workflow failures.

lisa/sut_orchestrator/azure/platform_.py
bfjelds: cool!

would it make sense to make the flag use_public_address?  it seems like the alternative (use_address) isn't very helpful when you read it.

bfjelds: hmmm ... it seems i don't understand what all is needed to use a flag on RemoteNode instead of what i've done here.  can you provide more details?

bfjelds: would the RemoteNode flag be used instead of this AzurePlatformSchema flag?  or in addition to?

bfjelds: updated the PR with my best guess of what you had in mind :).  let me know what i got wrong ... once i have this correct, i'll rerun my validation to make sure it still works.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -99,6 +99,7 @@ class NodeContext:
     username: str = ""
     password: str = ""
     private_key_file: str = ""
+    use_public_ip_address: bool = True
```
squirrelsc: Make names consistent like `use_public_ip_address` on other places too.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Thank you, that most like what I thought!

squirrelsc:
@LiliDeng looks good to me.

lisa/sut_orchestrator/azure/common.py
bfjelds: ok ... was trying to match the collocated params (it is public_ip_address here, public_address in other places).

bfjelds: in some places there is a public_address property ... and others there is a public_ip_address property.

i was trying to match whatever was local (use_public_address when public_address is applicable, use_public_ip_address when public_ip_address is applicable).  if i'm not doing that, which do you prefer:  use_public_address or use_public_ip_address? (i'm good any way :))

squirrelsc: if so, `use_public_address` is shorter and better. :)

bfjelds: moved everything to use_public_address

bfjelds:
> Good feature, please also update related documents.

can you point me to where the documents are?

xordux:
This is great change. I have 1 doubt, ~won't we need to change the class variables(self.public_address and self.public_port) of [RemoteNode here](https://github.com/microsoft/lisa/blob/49bf0f8685a976acadff1ce79e4f1b20de281fa0/lisa/node.py#L544)? 
Otherwise some tests will still use the public address. (For example [smoke_test here](https://github.com/microsoft/lisa/blob/49bf0f8685a976acadff1ce79e4f1b20de281fa0/microsoft/testsuites/core/provisioning.py#L224))~

~A change like this could be done:~
```
self.public_address = public_address if use_public_address else address
self.public_port = public_port if use_public_address else port
```

Edit: On a second thought I think the smoke_test(and tests which are using public_address directly) should be modified to honor `use_public_address` setting.

LiliDeng:
@bfjelds please rebase and fix code conflict, thank you.

LiliDeng (approved):

PR #2683, opened 2023-03-30
Add perf tool tests

dsrivastavv:


lisa/messages.py
```diff
@@ -232,6 +232,20 @@ class NetworkUDPPerformanceMessage(PerfMessage):
     packet_size_kbytes: Decimal = Decimal(0)
 
 
+@dataclass
+class PerfToolMessagingMessage(PerfMessage):
+    average_time_sec: Decimal = Decimal(0)
```
squirrelsc, line 237: The messages should be metrics based, not tool based. If metrics are defined in other messages, it should be reused. If not, we can add it as a new metrics message. What's metrics are measured here?

dsrivastavv: We are measuring the latency in messaging module to pass messages

dsrivastavv: I think we can combine both these into a metric called AverageMetric

lisa/messages.py
```diff
@@ -232,6 +232,13 @@ class NetworkUDPPerformanceMessage(PerfMessage):
     packet_size_kbytes: Decimal = Decimal(0)
 
 
+@dataclass
+class AverageMetric(PerfMessage):
```
squirrelsc: It should be perf metrics, instead of general names. Like network, there are pps, throughput and so on.

dsrivastavv: I realised that we still need to two different metrics. But I have renamed it to make it specific to what we are testing rather than the tool we use it to test

squirrelsc (approved):

PR #2685, opened 2023-03-30
Correct regex for CH perf test filter

smit-gardhariya:
This change will modify regex used to filter the CH perf testcase names from stdout while listing it at runtime. Recently we have seen some other stdout lines which matched the older regex which lead to fetch garbage value in listing the testcase at runtime.


LiliDeng (approved):

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -350,10 +350,10 @@ def _list_perf_metrics_tests(self, hypervisor: str = "kvm") -> Set[str]:
 
         # Ex. String for below regex
         # "boot_time_ms" (test_timeout=2s,test_iterations=10)
```
squirrelsc, line 352: Please add new examples too. It helps preventing regression in future.

PR #2691, opened 2023-04-05
Add parent_test field in SubTestMessage 

pupacha:
To identify which lisa test the SubTestMessage is coming from, store LISA testname that is generating the SubTestMessage into 'parent_test' field

### DIFF
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -92,6 +92,7 @@ class TestResultMessage(TestResultMessageBase):
 class SubTestMessage(TestResultMessageBase):
     hardware_platform: str = ""
     type: str = "SubTestResult"
+    parent_test: str = ""
 
 
 class NetworkProtocol(str, Enum):
@@ -314,6 +315,10 @@ def create_test_result_message(
     message.status = test_status
     message.message = test_message
     message.elapsed = test_result.get_elapsed()
+    if message_type == SubTestMessage:
+        if not other_fields:
+            other_fields = {}
+        other_fields.update({"parent_test": test_result.runtime_data.name})
     if other_fields:
         dict_to_fields(other_fields, message)
     return message


pupacha:
@anirudhrb Please review

anirudhrb:
Is it useful to do this in a common place so that all substests automatically have this field populated?

pupacha:
> Is it useful to do this in a common place so that all substests automatically have this field populated?

How about following change in `create_test_result_message` 
https://github.com/microsoft/lisa/blob/bb1cacf165980bdc698af4071cd6cddf9b08aa71/lisa/messages.py ? @squirrelsc @LiliDeng @anirudhrb 

    if message_type == SubTestMessage and "parent_test" not in message.information:
        message.information["parent_test"] = test_result.runtime_data.name

lisa/messages.py
```diff
@@ -316,4 +316,6 @@ def create_test_result_message(
     message.elapsed = test_result.get_elapsed()
     if other_fields:
         dict_to_fields(other_fields, message)
+    if message_type == SubTestMessage and "parent_test" not in message.information:
```
squirrelsc: It should be a field on SubTestMessage.

pupacha: Made it as a field in SubTestMessage

anirudhrb (approved):

lisa/messages.py
```diff
@@ -314,6 +315,10 @@ def create_test_result_message(
     message.status = test_status
     message.message = test_message
     message.elapsed = test_result.get_elapsed()
+    if message_type == SubTestMessage:
+        if not other_fields:
+            other_fields = {}
+        other_fields.update({"parent_test": test_result.runtime_data.name})
```
LiliDeng, line 321: it should be filled in 
other_fields["information"]["parent_test"]

this is current data run from your changed code
```
2023-04-12 14:25:32.872[5320][DEBUG] lisa.notifier[Console] received message [SubTestResult]: SubTestMessage(type='SubTestResult', time=datetime.datetime(2023, 4, 12, 14, 25, 32, 794499), elapsed=2972.2109429, id_='lisa_0_0', name='binfmt_misc02', status=<TestStatus.PASSED: 5>, message='', stacktrace=None, information={'version': '20200930', 'exit_value': 0}, hardware_platform='x86_64', parent_test='ltp_lite')
```

pupacha: I initially used 'information' field to store 'parent_test' but Chi suggested to add a field in SubTestMessage itself instead. And I agree with that idea as well.

squirrelsc: It's a common requirement to know the parent test name, so I asked to add it to subtest message level.

LiliDeng (approved):

PR #2695, opened 2023-04-06
RemoteNode self.public_address should follow use_public_address

### LINKED ISSUE #2694: Private IP address is not used in test verify_stop_start_in_platform
## What
When `use_public_address` is set to `false` then test `verify_stop_start_in_platform` still uses public_address of Azure VMs (hence test fails in environment where only private_address are accessible)

## Why
When `use_public_address` is set to `false` then function `set_connection_info` in RemoteNode class only sets private_ip in `self._connection_info`  but variable self.public_address still points to public ip which is later used by test cases.

## Error log:
Log where is issue is reproduced: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=339536&view=logs&j=528a33a8-464e-565a-7855-df1244039b47&t=86425f9e-1e14-58f6-d288-a39ec4000eff&l=1845 

The above pipeline runs in an environment where only private_ip are accessible. The log says that connection timeout to ip `13.84.212.170:22` (which is a public ip)  while the private ip is `10.0.0.5:22`

xordux:
Fixes #2694

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -543,8 +543,8 @@ class RemoteNode(Node):
         )
         self._shell = SshShell(self._connection_info)
 
-        self.public_address = public_address
-        self.public_port = public_port
+        self.public_address = public_address if use_public_address else address
+        self.public_port = public_port if use_public_address else port
         self.internal_address = address
         self.internal_port = port
 


squirrelsc (dismissed):

squirrelsc:
It needs to fix the test case to get `connection_info`, and use address from there. The public_address won't be removed, because there may be other requirements to use it.

xordux:
> It needs to fix the test case to get `connection_info`, and use address from there. The public_address won't be removed, because there may be other requirements to use it.

Maybe I am not understanding the difference between class variable `public_address` and `_connection_info.public_address`, aren't they same?
Shouldn't those other requirements too use private_address when `use_public_address` is marked False?

I created a draft PR for fixing the test case instead here: https://github.com/microsoft/lisa/pull/2697

xordux:
Tested this change in the pipeline: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=338845&view=results

Other than this test_case, there are many tests which are using `node.public_address` and hence will be fixed by this change. For example, [verify_sriov_add_max_nics](https://github.com/microsoft/lisa/blob/f129987dd55f8f17e86eafc9129cbb3e21a35ca5/microsoft/testsuites/network/sriov.py#L326) and [verify_boot_with_debug_kernel](https://github.com/microsoft/lisa/blob/f129987dd55f8f17e86eafc9129cbb3e21a35ca5/microsoft/testsuites/core/boot.py#L66). Until now I've found 10 files where this change will fix the usage of private_ip address.

squirrelsc:
> > It needs to fix the test case to get `connection_info`, and use address from there. The public_address won't be removed, because there may be other requirements to use it.
> 
> Maybe I am not understanding the difference between class variable `public_address` and `_connection_info.public_address`, aren't they same? Shouldn't those other requirements too use private_address when `use_public_address` is marked False?
> 
> I created a draft PR for fixing the test case instead here: #2697

Thank you fixing this! public_address and private_address are information on the remote node. We assumed it's always connect from public address, so they are the same. Now, in some scenarios, it needs to be connected from private address, so the connection info is different now. The connection info doesn't save both public/private address, it's how LISA connects to the remote node.

I'm not sure if we need the public_address in future. So for now, we can always use connection info for new connection in LISA.

PR #2696, opened 2023-04-06
fix ready platform to run tests with requirement

squirrelsc:
If the Disk or NetworkAdapter is None, LISA fails on check req. With this fix, ready platform has the default empty value to work around it. And ready platform can add capability in runbook to run any test.

### DIFF
--- a/lisa/sut_orchestrator/ready.py
+++ b/lisa/sut_orchestrator/ready.py
@@ -7,6 +7,7 @@ from lisa import features
 from lisa.environment import Environment
 from lisa.feature import Feature
 from lisa.platform_ import Platform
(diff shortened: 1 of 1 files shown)


LiliDeng:
I have tried, if in test case, it invoked the features like disk feature, I still saw not implement error, does this change fix such kinds of issue?

squirrelsc:
> I have tried, if in test case, it invoked the features like disk feature, I still saw not implement error, does this change fix such kinds of issue?

It's ok, if some tests need special env, but it's failed. Currently, if a test case specify any environment requirements, a test case will fail.

LiliDeng (approved):

PR #2697, opened 2023-04-07
Use RemoteNode._connection_info for TCP connection

### LINKED ISSUE #2694: Private IP address is not used in test verify_stop_start_in_platform

xordux:
Fixes #2694 

Not tested yet.

### DIFF
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -4,6 +4,7 @@
 import io
 import logging
 import pathlib
+import re
 import shlex
 import signal
 import subprocess
@@ -19,7 +20,7 @@ from spur.errors import NoSuchCommandError  # type: ignore
 from lisa.util import LisaException, filter_ansi_escape
 from lisa.util.logger import Logger, LogWriter, add_handler, get_logger
 from lisa.util.perf_timer import create_timer
-from lisa.util.shell import Shell
+from lisa.util.shell import Shell, SshShell
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2169,13 +2169,13 @@ def _enable_ssh_on_windows(self, node: Node) -> None:
 
         log = node.log
         log.debug(
-            f"checking if SSH port {remote_node.public_port} is reachable "
+            f"checking if SSH port {remote_node._connection_info.port} is reachable "
```
squirrelsc: Thank you for making this fix! The _connection_info is private member, and shouldn't be used out of Node. Please use the `connection_info`. Because the wait_tcp_port_ready is always use both address and port of a Node, it's right time to change the interface to pass in the node only.

xordux: @squirrelsc this PR got closed automatically after I reverted my changes and force pushed the changes. The new PR is here: https://github.com/microsoft/lisa/pull/2701

PR #2698, opened 2023-04-07
Use minimal shell type when encounter CommandInitializationError

lubaihua33:
Some publishers images, such as azhpc-desktop, javlinltd and  vfunctiontechnologiesltd, there might have permission errors when scripts under /etc/profile.d directory are executed at startup of the bash shell for a non-root user. Then calling spawn to run any  Linux commands might raise CommandInitializationError.
The error messages are like this: 
Error while initializing command. The most likely cause is an unsupported shell. Try using a minimal shell type when calling 'spawn' or 'run'. 
(Failed to parse line 'b' /etc/profile.d/clover.sh:line 10:/opt/clover/bin/prepare-hostname.sh: Permission denied'' as integer)                
 
In this PR, except CommandInitializationError, then use minimal shell type.

When waiting to the result, if there is CommandInitializationError when calling spawn, the stdout has that error line before the output of every command. E.g. the stdout of command "uname -vrmo" is like:
 '/etc/profile.d/clover.sh: line 10: /opt/clover/bin/prepare-hostname.sh: Permission denied\r\n0\r\n3.10.0-1160.88.1.el7.x86_6 #1 SMP Tue Mar 7 15:41:52 UTC 2023 x86_64 GNU/Linux'

Other example:
'/etc/profile.d/vglrun.sh: line 3: lspci: command not found\r\nDescription:\tCentOS Linux release 7.9.2009 (Core)'

So in wait_result, filter this error line.

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -3,6 +3,7 @@
 
 import logging
 import os
+import re
 import shutil
 import socket
 import sys
@@ -25,6 +26,13 @@ from .logger import Logger, get_logger
 from .perf_timer import create_timer
 
 _get_jump_box_logger = partial(get_logger, name="jump_box")
+
+# (Failed to parse line 'b'/etc/profile.d/vglrun.sh: line 3: lspci: command not found'' as integer)  # noqa: E501
+# (Failed to parse line 'b"touch: cannot touch '/tmp/version-updated': Permission denied"' as integer)  # noqa: E501
+# (Failed to parse line 'b'/etc/profile.d/clover.sh: line 10: /opt/clover/bin/prepare-hostname.sh: Permission denied'' as integer)  # noqa: E501
+_spawn_initialization_error_pattern = re.compile(
+    r"(Failed to parse line \'b[\'\"](?P<linux_profile_error>.*?)[\'\"]\' as integer)"
+)
 
 
 def wait_tcp_port_ready(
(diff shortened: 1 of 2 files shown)


lisa/util/shell.py
```diff
@@ -290,25 +299,55 @@ def spawn(
...
+                    have_tried_minimal_type = True
+                    matched = _spawn_initialization_error_pattern.search(
+                        str(identifier)
+                    )
+                    if matched:
+                        self.spawn_initialization_error_string = matched.group(
+                            "linux_profile_error"
+                        )
+                else:
+                    raise LisaException(identifier)
```
squirrelsc: raise identifier is enough.

lisa/util/shell.py
```diff
@@ -290,25 +299,55 @@ def spawn(
...
-        except FunctionTimedOut:
-            raise LisaException(
-                f"The remote node is timeout on execute {command}. "
-                f"It may be caused by paramiko/spur not support the shell of node."
-            )
+        while True:
+            try:
+                if self._inner_shell._spur._shell_type == spur.ssh.ShellTypes.minimal:
+                    # minimal shell type doesn't support store_pid
+                    store_pid = False
```
squirrelsc, line 308: Because the pid is not saved, so the process cannot be killed. Any side effect here?

lubaihua33: From my understanding, only when the process timeout, process.kill() is called, then pid is needed. 
So far, in my local testing, I haven't seen any timeout yet. If there is timeout, the worst result might be one process in Linux couldn't be killed, it should not have an impact on the test.

lisa/util/shell.py
lubaihua33: Done

squirrelsc (approved):

PR #2702, opened 2023-04-10
support ignored_capability

LiliDeng:
usage:
```
    ignored_capability:
      - SerialConsole
```

### DIFF
--- a/lisa/runners/lisa_runner.py
+++ b/lisa/runners/lisa_runner.py
@@ -769,6 +769,23 @@ class LisaRunner(BaseRunner):
                             test_result.set_status(TestStatus.SKIPPED, str(identifier))
                             break
 
+                        if (
+                            node_requirement.features
+                            and hasattr(self, "platform")
+                            and self.platform.runbook.ignored_capability
+                        ):
+                            node_requirement.features.items = [
+                                x
+                                for x in node_requirement.features
+                                if str(x).lower()
+                                not in list(
+                                    map(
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -166,6 +166,7 @@ class AzureNodeSchema:
     # Force to maximize capability of the vm size. It bypass requirements on
     # test cases, and uses to force run performance tests on any vm size.
     maximize_capability: bool = False
+    maximize_single_feature: Optional[str] = None
```
squirrelsc: 1. Use a list, in case it needs to ignore multiple features.
2. It should be called `ignored_capability`. It's hard to maximize a feature actually, because the feature logic is complicated.
3. The field should be on platform level, and can be used by any platform. The logic should be in `_mage_test_requirements` in lisa runner to remove the feature requirement, when it's created.

lisa/runners/lisa_runner.py
```diff
@@ -702,6 +702,11 @@ def _merge_test_requirements(
             is_allow_set=True, items=[platform_type]
         )
 
+        ignored_capability: List[schema.FeatureSettings] = []
```
squirrelsc: It doesn't need to convert them to feature settings, I think it just needs to compare feature's name with the ignored strings.

squirrelsc (approved):

PR #2703, opened 2023-04-10
Save serial log after testcase completion 

adityagesh:
Save serial log of VMs after testcase completes.

Example log file path: 
1. runtime\log\20230410\20230410-095213-491\environments\20230410-095223-846-generated_0\lisa-remote1234-20230410-095213-491-e0-n0
2. runtime\log\20230410\20230410-095213-491\environments\20230410-095223-846-generated_0\lisa-remote1234-20230410-095213-491-e0-n1

### DIFF
--- a/lisa/platform_.py
+++ b/lisa/platform_.py
@@ -107,6 +107,14 @@ class Platform(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
         """
         pass
 
+    def _save_console_log(
+        self,
+        environment: Environment,
+        log: Logger,
+        dir_name: str,
+    ) -> None:
+        ...
+
     @hookimpl
     def get_environment_information(self, environment: Environment) -> Dict[str, str]:
         assert environment.platform
--- a/lisa/testsuite.py
+++ b/lisa/testsuite.py
@@ -640,6 +640,13 @@ class TestSuite:
                 log=case_log,
             )
 
+            if environment.platform:
+                environment.platform._save_console_log(
+                    environment=environment,
+                    log=self.__log,
+                    dir_name=f"{case_unique_name}",
+                )
+
             case_log.info(
                 f"result: {case_result.status.name}, " f"elapsed: {total_timer}"
             )
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -85,6 +85,7 @@ from .common import (
     AzureVmPurchasePlanSchema,
     DataDiskCreateOption,
     DataDiskSchema,
+    EnvironmentContext,
     SharedImageGallerySchema,
     check_or_create_resource_group,
     check_or_create_storage_account,
@@ -563,11 +564,19 @@ class AzurePlatform(Platform):
                 log.debug("not wait deleting")
 
     def _save_console_log(
-        self, resource_group_name: str, environment: Environment, log: Logger
+        self,
+        environment: Environment,
+        log: Logger,
+        dir_name: str = "",
     ) -> None:
         compute_client = get_compute_client(self)
+        resource_group_name = environment.get_context(
+            EnvironmentContext
+        ).resource_group_name
         vms = compute_client.virtual_machines.list(resource_group_name)
-        saved_path = environment.log_path / f"{get_datetime_path()}_serial_log"
+        saved_path = (
+            environment.log_path / f"{get_datetime_path()}_{dir_name}_serial_log"
+        )
         saved_path.mkdir(parents=True, exist_ok=True)
         for vm in vms:
             log_response_content = save_console_log(
@@ -1372,7 +1381,7 @@ class AzurePlatform(Platform):
                         deployment_operation, time_out=300, failure_identity="deploy"
                     )
                 except LisaTimeoutException:
-                    self._save_console_log(resource_group_name, environment, log)
+                    self._save_console_log(environment, log)
                     continue
                 break
         except HttpResponseError as identifier:


squirrelsc:
When is it needed? It costs time to capture on each test case. If some test cases need it, it can be called in the test cases.

LiliDeng:
> When is it needed? It costs time to capture on each test case. If some test cases need it, it can be called in the test cases.

in v2. We collect dmesg when testing starts and ends. now in v3, at the end of test, we don't collect such info since it uses the cache. sometimes we want to check there is any call trace or errors shown up in log (the case may pass or fail)

is it feasible that we add a switch to enable collect log at end of testing, by default, it is set as false?

squirrelsc:
> > When is it needed? It costs time to capture on each test case. If some test cases need it, it can be called in the test cases.
> 
> in v2. We collect dmesg when testing starts and ends. now in v3, at the end of test, we don't collect such info since it uses the cache. sometimes we want to check there is any call trace or errors shown up in log (the case may pass or fail)
> 
> is it feasible that we add a switch to enable collect log at end of testing, by default, it is set as false?

That can be collected in test cases by calling the feature. Make it configurable is not easy to use than add one line code.

adityagesh:
> > > When is it needed? It costs time to capture on each test case. If some test cases need it, it can be called in the test cases.
> > 
> > 
> > in v2. We collect dmesg when testing starts and ends. now in v3, at the end of test, we don't collect such info since it uses the cache. sometimes we want to check there is any call trace or errors shown up in log (the case may pass or fail)
> > is it feasible that we add a switch to enable collect log at end of testing, by default, it is set as false?
> 
> That can be collected in test cases by calling the feature. Make it configurable is not easy to use than add one line code.

I can think of two cases where we would like to have the serial logs
1. Perf tests where test passes but perf numbers are low
2. Failed cases - sometimes failures pass on rerun. It is good to have serial logs in these cases.

squirrelsc:
> > > > When is it needed? It costs time to capture on each test case. If some test cases need it, it can be called in the test cases.
> > > 
> > > 
> > > in v2. We collect dmesg when testing starts and ends. now in v3, at the end of test, we don't collect such info since it uses the cache. sometimes we want to check there is any call trace or errors shown up in log (the case may pass or fail)
> > > is it feasible that we add a switch to enable collect log at end of testing, by default, it is set as false?
> > 
> > 
> > That can be collected in test cases by calling the feature. Make it configurable is not easy to use than add one line code.
> 
> I can think of two cases where we would like to have the serial logs
> 
> 1. Perf tests where test passes but perf numbers are low
> 2. Failed cases - sometimes failures pass on rerun. It is good to have serial logs in these cases.

Can we collect log before the environments deleted? I think collect one time for each env should be enough. The serial log contains timestamp too. We can know what happened during each test cases from this log.

adityagesh:
> > > > > When is it needed? It costs time to capture on each test case. If some test cases need it, it can be called in the test cases.
> > > > 
> > > > 
> > > > in v2. We collect dmesg when testing starts and ends. now in v3, at the end of test, we don't collect such info since it uses the cache. sometimes we want to check there is any call trace or errors shown up in log (the case may pass or fail)
> > > > is it feasible that we add a switch to enable collect log at end of testing, by default, it is set as false?
> > > 
> > > 
> > > That can be collected in test cases by calling the feature. Make it configurable is not easy to use than add one line code.
> > 
> > 
> > I can think of two cases where we would like to have the serial logs
> > 
> > 1. Perf tests where test passes but perf numbers are low
> > 2. Failed cases - sometimes failures pass on rerun. It is good to have serial logs in these cases.
> 
> Can we collect log before the environments deleted? I think collect one time for each env should be enough. The serial log contains timestamp too. We can know what happened during each test cases from this log.

@johnsongeorge-w , Please share your thoughts on this suggestion

johnsongeorge-w:
> > > > > > When is it needed? It costs time to capture on each test case. If some test cases need it, it can be called in the test cases.
> > > > > 
> > > > > 
> > > > > in v2. We collect dmesg when testing starts and ends. now in v3, at the end of test, we don't collect such info since it uses the cache. sometimes we want to check there is any call trace or errors shown up in log (the case may pass or fail)
> > > > > is it feasible that we add a switch to enable collect log at end of testing, by default, it is set as false?
> > > > 
> > > > 
> > > > That can be collected in test cases by calling the feature. Make it configurable is not easy to use than add one line code.
> > > 
> > > 
> > > I can think of two cases where we would like to have the serial logs
> > > 
> > > 1. Perf tests where test passes but perf numbers are low
> > > 2. Failed cases - sometimes failures pass on rerun. It is good to have serial logs in these cases.
> > 
> > 
> > Can we collect log before the environments deleted? I think collect one time for each env should be enough. The serial log contains timestamp too. We can know what happened during each test cases from this log.
> 
> @johnsongeorge-w , Please share your thoughts on this suggestion

There are 2 major use case and I think few aspects are already discussed above.
1.	Collecting dmesg before and after each test will help to narrow down the potential problem (from kernel error message perspective) when analyzing any test failure or performance issues.
2.	Serial log is equally important to be collected after each failed test cases. In many situations, we are not able to conclude because lack of serial logs. Checking **last bootup (only for failed tests) ** will be more convenient rather searching the entire serial logs.  

Collecting these logs will considerably optimize (avoid re-runs) the time required to triage/analyze the test failures. This is critical when issue is intermittent and cannot be reproduced always.

lisa/testsuite.py
```diff
@@ -640,6 +640,13 @@ def start(
                 log=case_log,
             )
 
+            if environment.platform:
```
squirrelsc, line 643: 1. Never call private method out of the class. It breaks the design and fail other platform.
2. In test cases, use node.features.is_supported to check if SerialConsole is available, and then call it. Don't call platform directly.

lisa/platform_.py
```diff
@@ -107,6 +107,14 @@ def _cleanup(self) -> None:
         """
         pass
 
+    def _save_console_log(
```
squirrelsc, line 110: It's a feature of the node, not add it to platform.

PR #2706, opened 2023-04-12
Performance: Fix DPDK protocol type from default

mcgov:
DPDK testpmd generates Udp traffic, not TCP. Fixing default value being posted to DB results.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkperf.py
+++ b/microsoft/testsuites/dpdk/dpdkperf.py
@@ -11,7 +11,11 @@ from lisa import (
     simple_requirement,
 )
 from lisa.features import Gpu, Infiniband, IsolatedResource, Sriov
-from lisa.messages import NetworkPPSPerformanceMessage, create_perf_message
+from lisa.messages import (
+    NetworkPPSPerformanceMessage,
+    TransportProtocol,
+    create_perf_message,
+)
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkperf.py
```diff
@@ -385,6 +389,7 @@ def _create_pps_performance_results(
         for result_fields in [sender_fields, receiver_fields]:
             result_fields["tool"] = constants.NETWORK_PERFORMANCE_TOOL_DPDK_TESTPMD
             result_fields["test_type"] = "performance"
+            result_fields["protocol_type"] = TransportProtocol.Udp
```
LiliDeng, line 392: where we can get that it is udp type?

mcgov: Testpmd by default only generates UDP traffic, we have no Testpmd based tests currently that handle TCP traffic. I'll be opening a PR to add an L3 forwading test and will update this at that time. The forwader in that case will be forwarding TCP traffic.

mcgov: @squirrelsc or @LiliDeng  can we merge this today?

LiliDeng (approved):

PR #2707, opened 2023-04-13
feature: add IaaS feature

LiliDeng:
before fix, when test ran against the VM sizes which don't support IaaS, the result is like below
```
deployment failed. LisaException: BadRequest: Requested operation cannot be performed because VM size Standard_D5_v2_Internal does not support IaaS deployments.
```

after fix, the test case results are skipped.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -338,6 +338,7 @@ class AzurePlatform(Platform):
             features.NetworkInterface,
             features.Resize,
             features.StartStop,
+            features.IaaS,
             features.Infiniband,
             features.Hibernation,
             features.SecurityProfile,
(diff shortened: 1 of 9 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2286,7 +2287,11 @@ def _get_capabilities(
                 continue
 
             if vm_size in caps:
-                candidate_caps.append(caps[vm_size])
+                cap_features = caps[vm_size].capability.features
```
squirrelsc, line 2043: Add comments to explain how the feature affect deployment.

LiliDeng: Do you think it is a good place to put the logic? By default, the VM can only be provisioned on the VM sizes which have IaaS capability.

squirrelsc: Yes, we can hardcode like this. This feature is special, if it doesn't exist, it means not able to be tested.

squirrelsc: Can you add comments here to explain why this feature is needed in code?

squirrelsc (dismissed):

squirrelsc (approved):

PR #2708, opened 2023-04-13
Fix perftool install on debian backports and suse

dsrivastavv:


dsrivastavv:
@LiliDeng This PR should be good to go now

LiliDeng (approved):

PR #2709, opened 2023-04-14
Add capture_boot_time and capture_azure_information flag in schema

lubaihua33:
Add capture_boot_time and capture_azure_information flag in platform schema.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -385,7 +385,7 @@ class Posix(OperatingSystem, BaseClassMixin):
         ).save_stdout_to_file(saved_path / "modinfo-hv_netvsc.txt")
 
         if self._node.is_test_target:
-            if self._node._first_initialize:
+            if self._node.capture_boot_time and self._node._first_initialize:
                 from lisa.tools import SystemdAnalyze
 
                 try:
@@ -396,21 +396,24 @@ class Posix(OperatingSystem, BaseClassMixin):
                 except Exception as identifier:
                     self._node.log.debug(f"error on get boot time: {identifier}")
 
-            from lisa.tools import Chmod, Find
+            file_list = []
+            if self._node.capture_azure_information:
+                from lisa.tools import Chmod, Find
 
-            find_tool = self._node.tools[Find]
-            file_list = find_tool.find_files(
-                self._node.get_pure_path("/var/log/azure/"),
-                file_type="f",
-                sudo=True,
-                ignore_not_exist=True,
-            )
-            if len(file_list) > 0:
-                self._node.tools[Chmod].update_folder(
-                    "/var/log/azure/", "a+rwX", sudo=True
+                find_tool = self._node.tools[Find]
+                file_list = find_tool.find_files(
+                    self._node.get_pure_path("/var/log/azure/"),
+                    file_type="f",
+                    sudo=True,
+                    ignore_not_exist=True,
                 )
+                if len(file_list) > 0:
+                    self._node.tools[Chmod].update_folder(
+                        "/var/log/azure/", "a+rwX", sudo=True
+                    )
+                file_list.append("/var/log/waagent.log")
+
             file_list.append("/etc/os-release")
-            file_list.append("/var/log/waagent.log")
             for file in file_list:
                 try:
                     file_name = file.split("/")[-1]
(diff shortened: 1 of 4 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -290,6 +290,10 @@ class AzurePlatformSchema:
     wait_delete: bool = False
     # the AzCopy path can be specified if use this tool to copy blob
     azcopy_path: str = field(default="")
+    # capture azure log or not
+    capture_azure_information: bool = False
```
squirrelsc: They should be in general platform information like admin_password

lubaihua33: Thank you Chi for the comment.
I think you should mean "capture_boot_time" need in the general platform, and "capture_azure_information" is still in Azure platform, right?

For "capture_boot_time", I tried adding it in the general platform. But I didn't find a suitable location to pass the value to the node. Could you give me some hints if it doesn't take you much time?

squirrelsc: Both flag should be in general platform, because they are collected in Node.
The code can be in deploy_environment of general platform.

lisa/util/process.py
```diff
@@ -396,6 +402,15 @@ def _filter_profile_error(self, raw_input: str) -> str:
             )
         return raw_input
 
+    def _check_if_need_input_password(self, raw_input: str) -> None:
+        # Check if the stdout contains "[sudo] password for .*: " and
+        # "sudo: timed out reading password" strings. If so, raise exception
+        if re.search(REQUIRE_INPUT_PASSWORD_PATTERN, raw_input):
+            raise LisaException(
```
LiliDeng: is it RequireUserPasswordException?

lisa/operating_system.py
```diff
@@ -396,19 +396,21 @@ def capture_system_information(self, saved_path: Path) -> None:
...
+                    self._node.get_pure_path("/var/log/azure/"),
+                    file_type="f",
+                    sudo=True,
+                    ignore_not_exist=True,
                 )
+                if len(file_list) > 0:
+                    self._node.tools[Chmod].update_folder(
+                        "/var/log/azure/", "a+rwX", sudo=True
+                    )
             file_list.append("/etc/os-release")
```
LiliDeng, line 416: /etc/os-release can be collected, since it is not azure specific platform

lisa/util/process.py
lubaihua33: Yes, thank you Lili for reminding me. Have modifiled.

lisa/operating_system.py
lubaihua33: Yes, it is out of the capture_azure_information condition block.

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: Thank you Chi. Have modified.

lisa/util/process.py
```diff
@@ -17,11 +17,16 @@
 from assertpy.assertpy import AssertionBuilder, assert_that, fail
 from spur.errors import NoSuchCommandError  # type: ignore
 
-from lisa.util import LisaException, filter_ansi_escape
+from lisa.util import LisaException, RequireUserPasswordException, filter_ansi_escape
 from lisa.util.logger import Logger, LogWriter, add_handler, get_logger
 from lisa.util.perf_timer import create_timer
 from lisa.util.shell import Shell, SshShell
 
+# [sudo] password for lisatest: \r\nsudo: timed out reading password
```
squirrelsc: Move this change to a separated PR.

lisa/util/process.py
```diff
@@ -17,11 +17,16 @@
...
+REQUIRE_INPUT_PASSWORD_PATTERN = re.compile(
+    r"\[sudo\] password for.+\r\nsudo: timed out reading password"
```
squirrelsc: How long will it timeout? Can we detect it before timeout, and raise exception earlier?

lubaihua33: For primekey image, the timeout is 5 minutes, we couldn't detect it before the timeout because the process is still running at that time, and calling self._process.wait_for_result() is also needed to wait. So only when it timeouts, then the process is finished and can get the stdout.

lisa/util/process.py
lubaihua33: Done

lisa/operating_system.py
```diff
@@ -396,19 +396,21 @@ def capture_system_information(self, saved_path: Path) -> None:
...
             file_list.append("/var/log/waagent.log")
```
LiliDeng: this one belongs to azure specific file

LiliDeng (approved):

PR #2711, opened 2023-04-17
azure: generate ssh key pairs if no key file or password specified

LiliDeng:


lisa/schema.py
```diff
@@ -1257,7 +1257,7 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
         add_secret(self.admin_username, PATTERN_HEADTAIL)
         add_secret(self.admin_password)
 
-        if self.type != constants.PLATFORM_READY:
+        if self.type not in (constants.PLATFORM_READY, constants.PLATFORM_AZURE):
```
squirrelsc: We can remove this check.

lisa/util/__init__.py
```diff
@@ -369,7 +373,26 @@ def get_datetime_path(current: Optional[datetime] = None) -> str:
     return f"{date}-{time}"
 
 
-def get_public_key_data(private_key_file_path: str) -> str:
+def get_or_generate_key_pairs(key_length: int = 2048) -> Tuple[str, str]:
+    # azure platform key accepts only RSA key, so use RSA key.
+    # the key lentgh range is 2048-4096, 2048 is the minimum.
+    # for some older distro, it only supports 2048.
+    # so here use default 2048.
+    key = paramiko.RSAKey.generate(key_length)
```
squirrelsc: Please check if there is lower level package to implement the ssh key. That would be more transparent on used algorithms.

squirrelsc: Print log somewhere, so we know the ssh key is generated.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -960,8 +961,13 @@ def _create_deployment_parameters(
             arm_parameters.admin_key_data = get_public_key_data(
                 self.runbook.admin_private_key_file
             )
-        else:
+        elif self.runbook.admin_password:
             arm_parameters.admin_password = self.runbook.admin_password
+        else:
+            (
+                arm_parameters.admin_key_data,
```
squirrelsc: The key should be the same to every environment, so it's easier to use.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1938,7 +1944,10 @@ def _enable_ssh_on_windows(self, node: Node) -> None:
 
         log.debug("SSH port is not open, enabling ssh on Windows ...")
         # The SSH port is not opened, try to enable it.
-        public_key_data = get_public_key_data(self.runbook.admin_private_key_file)
+        (
```
squirrelsc: Should the Windows use the same way as Linux? Why does it need to generate key sepratedly?

squirrelsc: Maybe, you can double check how to support password only Windows as well. If it's not easy, create a task to track.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: How about `_initialize` of azure platform?

lisa/schema.py
LiliDeng: generating key pairs is not suitable for ready platform + when user specified the rg not from lisa run?

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: ok. in get_or_generate_key_pairs, if the key exists (means has been created) then reuse it

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: oh, I didn't realize it is windows

lisa/schema.py
squirrelsc: We can just remove this check for all platforms. The platform can take care the check, or use auto-generated keys like Azure. The Ready platform was skipped, because the local node doesn't need it. So now, it's time to remove it for all platforms.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: > ok. in get_or_generate_key_pairs, if the key exists (means has been created) then reuse it

Good point. The combinator may create multiple platforms. Please make sure it can be reused under that scenario as well.

lisa/schema.py

lisa/util/__init__.py
LiliDeng: please check current implement, thanks!

LiliDeng: added

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: removed

LiliDeng:
@squirrelsc could you please kindly review the latest change? Thanks.

lisa/util/__init__.py
```diff
@@ -385,7 +390,39 @@ def get_datetime_path(current: Optional[datetime] = None) -> str:
...
+def get_or_generate_key_pairs(
+    log: "Logger", key_length: int = 2048, algorthim: str = "RSA"
+) -> Tuple[str, str]:
+    # refer: https://learn.microsoft.com/en-us/azure/virtual-machines/linux/create-ssh-keys-detailed#supported-ssh-key-formats # noqa: E501
+    # azure platform key accepts only RSA key with minimum length of 2048 bits.
+    # for some older distro, it only supports 2048 bits.
+    key_class = algorthim_dict.get(algorthim.upper(), None)
+    assert key_class, f"unsupported key algorthim: {algorthim}"
+    public_key: str = ""
+    private_key: str = str(constants.RUN_LOCAL_LOG_PATH / "id_rsa")
```
squirrelsc: 1. The variable name should be `private_key_file`.
2. The public_key should be saved to disk as well.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -936,6 +937,16 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._rm_client = get_resource_management_client(
             self.credential, self.subscription_id, self.cloud
         )
+        if self.runbook.admin_private_key_file:
```
squirrelsc: It should be below the generating. So the generating doesn't need to return public key data together.

lisa/util/__init__.py
```diff
@@ -385,7 +390,39 @@ def get_datetime_path(current: Optional[datetime] = None) -> str:
...
+            key = key_class.from_private_key(f)
+        log.info(f"ssh key is loaded from {private_key}")
+    else:
+        key = key_class.generate(key_length)
+        with global_ssh_key_access_lock:
+            with open(private_key, "w") as f:
+                key.write_private_key(f)
+        log.info(f"ssh key is generated at {private_key}")
+    public_key = f"ssh-rsa {key.get_base64()}"
+    return public_key, f.name
```
squirrelsc: Don't use `f.name`, use `private_key`.

lisa/util/__init__.py
```diff
@@ -385,7 +390,39 @@ def get_datetime_path(current: Optional[datetime] = None) -> str:
...
+    # for some older distro, it only supports 2048 bits.
+    key_class = algorthim_dict.get(algorthim.upper(), None)
+    assert key_class, f"unsupported key algorthim: {algorthim}"
+    public_key: str = ""
+    private_key: str = str(constants.RUN_LOCAL_LOG_PATH / "id_rsa")
+
+    if Path(private_key).exists():
+        with open(private_key, "r") as f:
+            key = key_class.from_private_key(f)
+        log.info(f"ssh key is loaded from {private_key}")
```
squirrelsc: Don't print out, if it's loaded. In combinator, it will print a lot of times.

lisa/util/__init__.py
```diff
@@ -385,7 +389,39 @@ def get_datetime_path(current: Optional[datetime] = None) -> str:
...
+    # azure platform key accepts only RSA key with minimum length of 2048 bits.
+    # for some older distro, it only supports 2048 bits.
+    key_class = algorthim_dict.get(algorthim.upper(), None)
+    assert key_class, f"unsupported key algorthim: {algorthim}"
+    public_key_file: str = str(constants.RUN_LOCAL_LOG_PATH / "id_rsa.pub")
+    private_key_file: str = str(constants.RUN_LOCAL_LOG_PATH / "id_rsa")
+
+    if Path(private_key_file).exists():
+        with open(private_key_file, "r") as f:
+            key = key_class.from_private_key(f)
```
squirrelsc: It looks if the file exists, it doesn't need to load the key.

squirrelsc (approved):

LiliDeng:
during regression testing

PR #2713, opened 2023-04-17
Tools: Modprobe, Lsmod

mcgov:
Two commits fine-tuning behavior and capabilities on a few tools:

- Lsmod: 
Add a function to check for unloaded modules
If modules are compiled in, attempting to modprobe without checking if they're present can raise an error code. 
Add a function to minimize a list of drivers, removing any that are already loaded.

-Modprobe: 
add an option to use Lsmod to only load missing modules.

### DIFF
--- a/lisa/tools/lsmod.py
+++ b/lisa/tools/lsmod.py
@@ -2,7 +2,9 @@
 # Licensed under the MIT license.
 
 import re
-from typing import Any
+from typing import Any, List, Union
+
+from assertpy import assert_that
 
 from lisa.executable import Tool
 from lisa.util import find_patterns_in_lines
@@ -27,6 +29,26 @@ class Lsmod(Tool):
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._command = "lsmod"
+
+    def get_unloaded_modules(self, modules: Union[str, List[str]]) -> List[str]:
+        if isinstance(modules, str):
+            modules_list = modules.split(" ")
+        else:
+            modules_list = modules
+        assert_that(modules_list).described_as(
+            "Empty modules list passed to get_unloaded_modules in Lsmod tool"
+        ).is_not_empty()
+
+        result = self.run(sudo=True, force_run=True, expected_exit_code=0)
+        minimized_list = []
+        module_info = find_patterns_in_lines(result.stdout, [self.__output_pattern])
+
+        for module in modules_list:
+            # Don't add to minimized list if module is present in lsmod output
+            if any(module in info for sublist in module_info for info in sublist):
+                continue
+            minimized_list.append(module)
+        return minimized_list
 
     def module_exists(
         self,
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -3,6 +3,7 @@
 from typing import Any, List, Union
 
 from lisa.executable import Tool
+from lisa.tools import Lsmod
 
 
 class Modprobe(Tool):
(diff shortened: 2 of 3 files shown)


lisa/tools/lsmod.py
```diff
@@ -28,6 +30,22 @@ def _check_exists(self) -> bool:
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._command = "lsmod"
 
+    def get_unloaded_modules(self, modules: Union[str, List[str]]) -> List[str]:
```
squirrelsc, line 33: how about "check_modules_unloaded"? Because this method includes assertion.

lisa/tools/lsmod.py
```diff
@@ -28,6 +30,22 @@ def _check_exists(self) -> bool:
...
+        if isinstance(modules, str):
+            modules_list = modules.split(" ")
+        else:
+            modules_list = modules
+        assert_that(modules_list).described_as(
+            "Empty modules list passed to get_unloaded_modules in Lsmod tool"
```
squirrelsc, line 39: make it clearer like "Empty list is expected in check_modules_not_exists".

lisa/tools/lsmod.py
```diff
@@ -28,6 +30,22 @@ def _check_exists(self) -> bool:
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._command = "lsmod"
 
+    def get_unloaded_modules(self, modules: Union[str, List[str]]) -> List[str]:
+        if isinstance(modules, str):
+            modules_list = modules.split(" ")
+        else:
+            modules_list = modules
+        assert_that(modules_list).described_as(
```
squirrelsc, line 38: "modules shouldn't be empty, when calling get_unloaded_modules."

lisa/tools/lsmod.py
```diff
@@ -28,6 +30,22 @@ def _check_exists(self) -> bool:
...
+            "Empty modules list passed to get_unloaded_modules in Lsmod tool"
+        ).is_not_empty()
+
+        result = self.run(sudo=True, force_run=True, expected_exit_code=0)
+        minimized_list = []
+        for module in modules_list:
+            if module not in result.stdout:
```
squirrelsc: if it's possible, check the full string, instead of `in`. It makes some substring is recognized. For example, if "mlx5" in the string, the "mlx" is True.

lisa/tools/lsmod.py
mcgov: updated

lisa/tools/lsmod.py
mcgov: updated, will just leverage run and module_exists to check the list since the result will be cached

squirrelsc: It fixes, but much slower, if it needs to check multiple modules. Can we still get information from the one time output?

mcgov: Yep I can use a similar pattern to module_exists, commit incoming...

lisa/tools/modprobe.py
```diff
@@ -80,14 +81,22 @@ def load(
         self,
         modules: Union[str, List[str]],
         dry_run: bool = False,
+        skip_loaded: bool = False,
```
squirrelsc, line 84: this variable name is not clear. The method is `load`, but the variable is `skip_loaded`. Is it similar like `force_reload`?

mcgov: Hm, I had added this to handle a failure when trying to modprobe builtin drivers but maybe this is more of a testing issue on my end. I'm going to close and fix my test setup and we can re-design this in a smarter way if we hit the problem again.

PR #2714, opened 2023-04-17
[Tools] Tee

mcgov:
nic bind/unbind can fail when using echo and bash append.
This results in strange behavior in dpdk tests when nics are not set up correctly.

- Add Tee tool for using echo 'value' | sudo tee filename to ensure writes are not lost due to file permissions on special files
- Add option to not store NIC driver when calling get_nic_driver

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from assertpy import assert_that
 from retry import retry
 
-from lisa.tools import Echo, Ip, Lspci
+from lisa.tools import Ip, Lspci, Tee
 from lisa.util import InitializableMixin, LisaException, constants, find_groups_in_lines
 
 if TYPE_CHECKING:
@@ -166,7 +166,7 @@ class Nics(InitializableMixin):
 
     # update the current nic driver in the NicInfo instance
     # grabs the driver short name and the driver sysfs path
-    def get_nic_driver(self, nic_name: str) -> str:
+    def get_nic_driver(self, nic_name: str, store_driver: bool = True) -> str:
         # get the current driver for the nic from the node
         # sysfs provides a link to the driver entry at device/driver
         nic = self.get_nic(nic_name)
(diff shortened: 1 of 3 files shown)


lisa/tools/tee.py
```diff
@@ -0,0 +1,44 @@
...
+        **kwargs: Any,
+    ) -> None:
+        # tee breaks sudo ordering of tool.run()
+        # cwd, shell, etc will be passed through to execute
+        # sudo is the only one which needs special casing
+        cmd = "tee"
+        if append:
+            cmd = f"{cmd} -a"
+        if sudo:
+            cmd = f"sudo {cmd}"
```
squirrelsc, line 42: Can it depend on node.execute to add the "sudo"?

lisa/tools/tee.py
```diff
@@ -0,0 +1,44 @@
...
+        cmd = f"{cmd} {str(file)}"
+        self.node.execute(f"echo '{value}' | {cmd}", *args, **kwargs)
```
squirrelsc: Don't pass *args and **kwargs into execute, it looks unpredictable.

lisa/nic.py
```diff
@@ -216,22 +217,19 @@ def nic_info_is_present(self, nic_name: str) -> bool:
 
     def unbind(self, nic: NicInfo) -> None:
         # unbind nic from current driver and return the old sysfs path
-        echo = self._node.tools[Echo]
+        tee = self._node.tools[Tee]
```
squirrelsc, line 220: I'm not if tee is installed in most systems or not. If not, it reduces compatibility of test cases.

mcgov: True, I've never run into a system it's not on, but it's part of coreutils. I'll follow the procedure used for `stat` and install if not present.

lisa/tools/tee.py
mcgov: updated

lisa/nic.py
mcgov: update (see tee code)

squirrelsc:
@LiliDeng LGTM

squirrelsc (approved):

PR #2717, opened 2023-04-20
fix environment is empty when case skip for environment not match requirement

LiliDeng:


lisa/runners/lisa_runner.py
```diff
@@ -596,6 +596,7 @@ def _get_runnable_test_results(
                     ):
                         runnable_results.append(result)
                 except SkippedException as identifier:
+                    result.environment = environment
```
squirrelsc: I forget why the test case will skip here. Can you check and add more comments?

LiliDeng: I think the below comments have explained. should I move these lines above new added line
```
                    # when check the environment, the test result may be marked
                    # as skipped, due to the test result is assumed not to match
                    # any environment.
```

squirrelsc: That's I don't understand now. I don't know which situation it is. In this method, the test results check only one environment, how it can know no other environment can be matched, and skip now...

LiliDeng: can we merge this PR firstly, then make another PR to add comments? I have verified the change, and I didn't get a time to investigate why it designed like this.

squirrelsc: ok, but it should set the env, only result.environment is None. It looks more safe here.

squirrelsc (approved):

PR #2718, opened 2023-04-20
node: add excluded_vm_sizes to exclude vm sizes list

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -190,6 +190,7 @@ class AzureNodeSchema:
     # It decides the real computer name. It cannot be too long.
     short_name: str = ""
     vm_size: str = ""
+    excluded_vm_sizes: Optional[List[str]] = None
```
squirrelsc, line 193: When is it needed? Can it be used by some feature to exclude related vm sizes?

LiliDeng: in our testing, there are some internal sizes, they have quota, but we may see provision time out issue or they don't offer the capabilities which they should have.

PR #2723, opened 2023-04-24
Add timesync drift testcase

dsrivastavv:
Add timesync drift testcase

### DIFF
--- a/microsoft/testsuites/core/timesync.py
+++ b/microsoft/testsuites/core/timesync.py
@@ -1,3 +1,5 @@
+import time
+from copy import deepcopy
 from pathlib import PurePosixPath
 from time import sleep
 from typing import List, Optional, Union, cast
@@ -18,6 +20,7 @@ from lisa import (
 )
 from lisa.operating_system import CpuArchitecture, Redhat
 from lisa.tools import Cat, Chrony, Dmesg, Hwclock, Lscpu, Ntp, Ntpstat, Service
+from lisa.tools.date import Date
 from lisa.tools.lscpu import CpuType
 from lisa.util.shell import wait_tcp_port_ready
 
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/core/timesync.py
```diff
@@ -377,3 +380,55 @@ def verify_pmu_disabled_for_arm64(self, node: Node) -> None:
...
+                    and current_time.day == node_time.day
+                ):
+                    is_drift_corrected = True
+                    log.debug(
+                        f"Iteration {iteration}: Time drift is not corrected by "
+                        "chrony yet. Retrying..."
+                    )
+                    break
+
+            if not is_drift_corrected:
```
squirrelsc: Use `assert_that`

microsoft/testsuites/core/timesync.py
```diff
@@ -377,3 +380,55 @@ def verify_pmu_disabled_for_arm64(self, node: Node) -> None:
...
+            # Get current time
+            date = node.tools[Date]
+            node_time = date.current()
+            modified_time = deepcopy(node_time)
+            modified_time = modified_time.replace(year=node_time.year - 2)
+            date.set(modified_time)
+
+            # Poll every second and check if the time drift is corrected
+            # for 5 minutes
+            for _ in range(300):
```
squirrelsc: Can this timeout to be shorter by some settings?

microsoft/testsuites/core/timesync.py
```diff
@@ -377,3 +380,55 @@ def verify_pmu_disabled_for_arm64(self, node: Node) -> None:
...
+                    current_time.year == node_time.year
+                    and current_time.month == node_time.month
+                    and current_time.day == node_time.day
+                ):
+                    is_drift_corrected = True
+                    log.debug(
+                        f"Iteration {iteration}: Time drift is not corrected by "
+                        "chrony yet. Retrying..."
+                    )
+                    break
```
squirrelsc, line 420: The logic is weird. The debug message says it's not correct, but it breaks here.

microsoft/testsuites/core/timesync.py
```diff
@@ -377,3 +380,55 @@ def verify_pmu_disabled_for_arm64(self, node: Node) -> None:
...
+            modified_time = modified_time.replace(year=node_time.year - 2)
+            date.set(modified_time)
+
+            # Poll every second and check if the time drift is corrected
+            # for 5 minutes
+            for _ in range(300):
+                time.sleep(1)
+                current_time = date.current()
+                if (
+                    current_time.year == node_time.year
```
squirrelsc: There maybe a bug. if the test case running at last minute a day, after 5 minutes, the day maybe not the same anyway. The year and month have the same problem. Please find another stable way to check it.

microsoft/testsuites/core/timesync.py
dsrivastavv: Sure! The max I observed was around 60 sec for 3 iteration, so we can keep 120 seconds with some buffer and increase if we observe any errors

microsoft/testsuites/core/timesync.py
dsrivastavv: Yeah. I mistakenly put the debug log in the if condition. Should be fixed now

microsoft/testsuites/core/timesync.py
dsrivastavv: Fixed. Comparing datetime object directly now

microsoft/testsuites/core/timesync.py
```diff
@@ -377,3 +380,47 @@ def verify_pmu_disabled_for_arm64(self, node: Node) -> None:
...
+
+            # Poll every second and check if the time drift is corrected
+            # for 5 minutes
+            for _ in range(120):
+                time.sleep(1)
+                current_time = date.current()
+                if current_time > node_time:
+                    is_drift_corrected = True
+                    break
+                log.debug(
```
squirrelsc: It outputs too many lines, output it every 10 seconds, or only once.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2724, opened 2023-04-25
fix issue for storage perf case

LiliDeng:


lisa/tools/fio.py
```diff
@@ -65,6 +65,9 @@ def install(self) -> bool:
             posix_os.install_packages("fio")
         except Exception as e:
             self._log.debug(f"failed to install fio from package: {e}")
+            if isinstance(e, RepoNotExistException):
```
squirrelsc: The pattern is like below. But is it right to eat all exceptions, if fio is not installed?

```python
try:
  ...
except RepoNotExistException as e:
  raise e
except Exception as e:
  self._log.debug(f"failed to install fio from package: {e}")
```

squirrelsc (approved):

PR #2726, opened 2023-04-25
ASAP support for storage perf tests

SRIKKANTH:


squirrelsc:
@SRIKKANTH don't provide internal links in Github.

lisa/features/nvme.py
```diff
@@ -21,7 +21,7 @@ class Nvme(Feature):
     _device_pattern = re.compile(r".*(?P<device_name>/dev/nvme[0-9]$)", re.MULTILINE)
     # brw-rw---- 1 root disk 259, 0 Jun 21 03:08 /dev/nvme0n1
     _namespace_pattern = re.compile(
-        r".*(?P<namespace>/dev/nvme[0-9]n[0-9]$)", re.MULTILINE
+        r".*(?P<namespace>/dev/nvme[0-9]n[0-9]+$)", re.MULTILINE
```
squirrelsc, line 26: Add new example above.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -455,13 +464,34 @@ def _perf_premium_datadisks(
...
     ) -> None:
         disk = node.features[Disk]
-        data_disks = disk.get_raw_data_disks()
-        disk_count = len(data_disks)
-        assert_that(disk_count).described_as(
-            "At least 1 data disk for fio testing."
-        ).is_greater_than(0)
-        partition_disks = reset_partitions(node, data_disks)
-        filename = ":".join(partition_disks)
+        if disk.is_asap_enabled :
```
squirrelsc: Should it be added as another test case? This method called "datadisks". It's confused to test with os disk.

lisa/features/disks.py
```diff
@@ -64,6 +64,17 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
     def get_resource_disk_mount_point(self) -> str:
         raise NotImplementedError
 
+    # Check if OS disk is of NVMe type to confirm ASAP enablement.
+    # Incase of ASAP VMs, disk controller is NVMe.
+    # Even non NVMe disks are treated as NVMe.
+    @property
+    def is_asap_enabled(self) -> bool:
```
squirrelsc: Is the "asap" a azure specified term? If so, it should be changed. And the logic is not used widely, it can be written in test case level.

lisa/features/nvme.py
SRIKKANTH: This change is to support nvme namespaces more than 9.
Example added.

lisa/features/disks.py
SRIKKANTH: ASAP is azure specific name. Changed the name to more generic name 'disk_controller_type',

microsoft/testsuites/performance/storageperf.py
SRIKKANTH: We are not testing OS disk here. We are just checking OS info for detecting the disk controller type. The test is still about data disks only.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -455,13 +464,34 @@ def _perf_premium_datadisks(
...
     ) -> None:
         disk = node.features[Disk]
-        data_disks = disk.get_raw_data_disks()
-        disk_count = len(data_disks)
-        assert_that(disk_count).described_as(
-            "At least 1 data disk for fio testing."
-        ).is_greater_than(0)
-        partition_disks = reset_partitions(node, data_disks)
-        filename = ":".join(partition_disks)
+        if disk.disk_controller_type == 'nvme':
```
LiliDeng, line 467: how the nvme disk exist. do they exist for some vm sizes by default. no need to attach data disks?

SRIKKANTH: In VMs with disk controller type = nvme even non nvme disks are detected as nvme disks. They perform (IOPS, throughput) just like regular azure data disks but guest uses nvme driver to talk to them. This happens when ASAP is enabled. The current SKU which I am testing doesn't have NVMe disk support but as ASAP enabled, host presents all data disks to guest as PCI NVMe devices and NVMe driver is used to talk to all data disks including OS disk.

LiliDeng: this method will be used case which attaches many data disks, I don't think your scenario needs attach extra data disks? you need add a new case and add disk controller type = nvme as requirement.

microsoft/testsuites/performance/storageperf.py
```diff
@@ -455,13 +464,34 @@ def _perf_premium_datadisks(
...
+            nvme = node.features[Nvme]
+            nvme_namespaces = nvme.get_namespaces()
```
LiliDeng, line 469: these code are from perf_nvme, we need think about how to invoke common methods instead of copying them

lisa/features/nvme.py
```diff
@@ -20,8 +20,10 @@ class Nvme(Feature):
     # crw------- 1 root root 251, 0 Jun 21 03:08 /dev/nvme0
     _device_pattern = re.compile(r".*(?P<device_name>/dev/nvme[0-9]$)", re.MULTILINE)
     # brw-rw---- 1 root disk 259, 0 Jun 21 03:08 /dev/nvme0n1
+    # ...
```
squirrelsc, line 23: remove the `...`, it doesn't meaningful.

lisa/features/disks.py
```diff
@@ -64,6 +64,12 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
     def get_resource_disk_mount_point(self) -> str:
         raise NotImplementedError
 
+    # Check root partition's disk type to know the disk controller type
+    @property
+    def disk_controller_type(self) -> str:
```
squirrelsc, line 69: The disk_controller_type exists in disk settings already. How are they work together? If the asap is a capability of some vm size, asap should be a new disk_controller_type, instead of a new field on disk.

kamalca: @squirrelsc For clarification, disk_controller_type in the disk field is used to set disk_controller_type in the ARM template for SKUs with can be deployed with **either** SCSI or NVMe (ASAP) disk_controller_type.

For this test, Johnson suggested we check the OS disk type directly to determine the disk controller type. There are a few cases where this can be useful. For example, E112ibs_v5 only supports NVMe disk_controller_type. If disk field disk_dontroller_type is not set, we still need to detect that disk_controller_type is NVMe based on the OS disk type after deployment.

microsoft/testsuites/performance/storageperf.py
kamalca, line 467: This does not match formatting guidelines. Please ensure your development environment is properly set up to automatically apply flake8 changes.

https://mslisa.readthedocs.io/en/main/write_test/dev_setup.html

kamalca: @LiliDeng ASAP SKUs (Ebs_v5) deploy with only OS disk like most VM SKUs. In order to test disk performance, we need attach many disks. This is done the exact same way as most SKUs. The only difference is that the disks will show up as NVMe devices even when they are SCSI disks.

From the test perspective, the only difference is the disk names. This does not require different requirements or a different test.

kamalca (changes requested):

lisa/features/disks.py
squirrelsc: I had a discussion with Srikanth. The value is used to decide the patterns to detect disks under `/dev`. So it just needs to add more patterns to include nvme devices by default.

PR #2729, opened 2023-04-26
Add code to implement input of password (recommend)

lubaihua33:
This PR only solves the situation that VM is the password authenticate type and needs to input a password when running the sudo command.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -18,6 +18,7 @@ from lisa.util import (
     ContextMixin,
     InitializableMixin,
     LisaException,
+    RequireUserPasswordException,
     constants,
     fields_to_dict,
     get_datetime_path,
@@ -105,6 +106,7 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
             result = process.wait_result(10)
             if result.exit_code == 0:
                 self._support_sudo = True
+                self.check_sudo_password_required()
             else:
                 self._support_sudo = False
                 self.log.debug("node doesn't support sudo, may cause failure later.")
@@ -113,6 +115,49 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
             self._support_sudo = True
 
         return self._support_sudo
+
+    def check_sudo_password_required(self) -> None:
+        # check if password is required when running command with sudo
+        if self.is_remote and self.is_posix:
+            process = self._execute(
+                "echo LISA_TEST_FOR_PASSWORD", shell=True, sudo=True, no_info_log=True
+            )
+            result = process.wait_result(10)
+            if result.exit_code != 0 and "[sudo] password for" in result.stdout:
+                self.log.debug(
+                    "Running commands with sudo in this node needs input of password."
+                )
+                ssh_shell = cast(SshShell, self.shell)
+                ssh_shell.is_sudo_required_password = True
+                if not ssh_shell.connection_info.password:
+                    raise RequireUserPasswordException(
+                        "Running commands with sudo requires user's password,"
+                        " but no password is provided."
+                    )
+                # ssh_shell.is_sudo_required_password is true, so running sudo command
+                # will input password in process.wait_result. Check running sudo again
+                # and get password prompts. For most images, after inputting a password
+                # successfully, the prompt is changed when running sudo command again.
+                # So check twice to get two kinds of prompt
+                password_prompts = []
+                for i in range(1, 3):
+                    process = self._execute(
+                        "echo LISA_TEST_FOR_PASSWORD",
+                        shell=True,
+                        sudo=True,
+                        no_info_log=True,
+                    )
+                    result = process.wait_result(10)
+                    if result.exit_code != 0:
+                        raise RequireUserPasswordException(
+                            "The password might be invalid for running sudo command"
+                        )
+                    password_prompt = result.stdout.replace(
+                        "LISA_TEST_FOR_PASSWORD", ""
+                    )
+                    password_prompts.append(password_prompt)
+                    self.log.debug(f"password prompt {i}: {password_prompt}")
+                ssh_shell.password_prompts = password_prompts
 
     @property
     def is_connected(self) -> bool:
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -242,6 +242,9 @@ class Process:
         if self._result is None:
             assert self._process
             if is_timeout:
+                # LogWriter only flushes if "\n" is written, so we need to flush
+                # manually.
+                self._stdout_writer.flush()
                 process_result = spur.results.result(
                     return_code=1,
                     allow_error=True,
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -221,6 +221,21 @@ class Process:
 
         return split_command
 
+    def check_and_input_password(self) -> None:
+        if (
+            self._sudo
+            and isinstance(self._shell, SshShell)
+            and self._shell.is_sudo_required_password
+        ):
+            if not self._shell.connection_info.password:
+                raise RequireUserPasswordException(
+                    "Running commands with sudo requires user's password,"
+                    " but no password is provided."
+                )
+            assert self._process
+            self._process.stdin_write(f"{self._shell.connection_info.password}\n")
+            self._log.debug("The user's password is input")
+
     def wait_result(
         self,
         timeout: float = 600,
@@ -229,9 +244,13 @@ class Process:
     ) -> ExecutableResult:
         timer = create_timer()
         is_timeout = False
+        has_checked_password = False
 
         while self.is_running() and timeout >= timer.elapsed(False):
             time.sleep(0.01)
+            if timer.elapsed(False) > 0.5 and not has_checked_password:
+                self.check_and_input_password()
+                has_checked_password = True
 
         if timeout < timer.elapsed():
             if self._process is not None:
@@ -297,6 +316,10 @@ class Process:
         ):
             self._result.stdout = self._filter_profile_error(self._result.stdout)
         self._check_if_need_input_password(self._result.stdout)
+
+        self._result.stdout = self._filter_sudo_required_password_info(
+            self._result.stdout
+        )
 
         return self._result
 
@@ -410,9 +433,32 @@ class Process:
         # "sudo: timed out reading password" strings. If so, raise exception
         if re.search(REQUIRE_INPUT_PASSWORD_PATTERN, raw_input):
             raise RequireUserPasswordException(
-                "Running commands with sudo requires user's password,"
-                " which is not support in Lisa now"
+                "Running commands with sudo requires user's password"
             )
+
+    def _filter_sudo_required_password_info(self, raw_input: str) -> str:
+        # If system needs input of password when running commands with sudo, the output
+        # might have below lines:
+        # We trust you have received the usual lecture from the local System
+        # Administrator. It usually boils down to these three things:
+        #
+        #     #1) Respect the privacy of others.
+        #     #2) Think before you type.
+        #     #3) With great power comes great responsibility.
+        #
+        # [sudo] password for l****t:
+        # After inputting the right password, the output might have the following line
+        # when running commands with sudo next time.
+        # [sudo] password for l****t:
+        # Remove these lines
+        if (
+            self._sudo
+            and isinstance(self._shell, SshShell)
+            and self._shell.is_sudo_required_password
+        ):
+            for prompt in self._shell.password_prompts:
+                raw_input = raw_input.replace(prompt, "")
+        return raw_input
 
 
 def _create_exports(update_envs: Dict[str, str]) -> str:
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -200,6 +200,8 @@ class SshShell(InitializableMixin):
         self._inner_shell: Optional[spur.SshShell] = None
         self._jump_boxes: List[Any] = []
         self._jump_box_sock: Any = None
+        self.is_sudo_required_password: bool = False
+        self.password_prompts: List[str] = []
         self.spawn_initialization_error_string = ""
 
         paramiko_logger = logging.getLogger("paramiko")
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -196,7 +196,7 @@ class SshShell(InitializableMixin):
     def __init__(self, connection_info: schema.ConnectionInfo) -> None:
         super().__init__()
         self.is_remote = True
-        self._connection_info = connection_info
+        self.connection_info = connection_info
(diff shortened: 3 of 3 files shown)


lisa/node.py
```diff
@@ -114,6 +116,45 @@ def support_sudo(self) -> bool:
...
         return self._support_sudo
 
+    def check_sudo_password_required(self) -> None:
+        self.initialize()
+
+        # check if password is required when running command with sudo
+        if self.is_remote and self.is_posix:
+            process = self._execute("echo OK", shell=True, sudo=True, no_info_log=True)
+            result = process.wait_result(5)
+            if result.exit_code != 0 and "[sudo] password for" in result.stderr:
```
squirrelsc: stderr doesn't support in remote shell. So let's always use stdout. We need to copy stderr to stdout too.

lisa/node.py
```diff
@@ -114,6 +116,45 @@ def support_sudo(self) -> bool:
...
+                self.log.debug("Running command with sudo needs a password")
+                self.check_and_reset_password()
+
+    def check_and_reset_password(self) -> None:
+        if not self.is_remote:
```
squirrelsc: why check if it's remote?

lisa/node.py
```diff
@@ -114,6 +116,45 @@ def support_sudo(self) -> bool:
...
+            return
+        ssh_shell = cast(SshShell, self.shell)
+        if not self.is_password_authentication_type:
+            self.log.debug(
+                "The authentication type is not password. Try to reset password..."
+            )
+            username = ssh_shell._connection_info.username
```
squirrelsc: use `ssh_shell.connection_info`

lisa/node.py
lubaihua33: When it needs input of password, the process will be timed out, then the stdout of result comes from self._log_buffer. In this result, its stdout is "cmd: ['sudo', 'sh', '-c', 'echo OK'], cwd: None, shell: True...". So I added code below code in wait_result, under 'if is_timeout' branch:

                if not process_result.stderr_output:
                    # When the process is timedout, self._stdout_writer._buffer might
                    # have some information can tell the timeout reason
                    process_result.stderr_output = self._stdout_writer._buffer

lubaihua33: This part of code is in commit [Use process stdin_write to input password if required](https://github.com/microsoft/lisa/pull/2729/commits/8a350104fa9ee0987bb5461d4572063445c4bbba)

lisa/node.py
lubaihua33: Yes, it is duplicated.

lisa/node.py
lubaihua33: SshShell doesn't have connection_info, RemoteNode has.

lisa/util/process.py
```diff
@@ -248,6 +259,10 @@ def wait_result(
                     output=self._log_buffer.getvalue(),
                     stderr_output="",
                 )
+                if not process_result.stderr_output:
```
squirrelsc: Let's keep stderr as is, in case we can fix the bug in future. We need to use logging.LoggerAdapter() to combine stderr and stdout to stdout. So all outputs are in stdout.

squirrelsc: And move this change into a separated commit for easy tracking.

lisa/schema.py
```diff
@@ -1448,10 +1448,14 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
             # use password
             # spurplus doesn't process empty string correctly, use None
             self.private_key_file = None
+        elif not self.password:
+            self.password = None
```
squirrelsc: The logic should be changed to below to avoid missing check on private key file.

```python
if not self.password and not self.private_key_file:
    raise LisaException(
        "at least one of password or private_key_file need to be set when "
        "connecting"
    )
elif not self.private_key_file:
    # use password
    # spurplus doesn't process empty string correctly, use None
    self.private_key_file = None
elif not self.password:
    self.password = None


if self.private_key_file:
    if not Path(self.private_key_file).exists():
        raise FileNotFoundError(self.private_key_file)
```

lisa/node.py
```diff
@@ -102,9 +103,10 @@ def support_sudo(self) -> bool:
         # check if sudo supported
         if self.is_posix and self._support_sudo is None:
             process = self._execute("command -v sudo", shell=True, no_info_log=True)
-            result = process.wait_result(10)
+            result = process.wait_result(5)
```
squirrelsc: Let's keep 10 seconds here for slowing environments or connections.

lisa/node.py
```diff
@@ -114,6 +116,23 @@ def support_sudo(self) -> bool:
 
         return self._support_sudo
 
+    def check_sudo_password_required(self) -> None:
```
squirrelsc, line 119: It needs to save the whole password prompt. It can be used to detect password needed next time, and used to remove it from stdout for consistent.

lisa/node.py
```diff
@@ -114,6 +116,23 @@ def support_sudo(self) -> bool:
...
+        # check if password is required when running command with sudo
+        if self.is_remote and self.is_posix:
+            process = self._execute("echo OK", shell=True, sudo=True, no_info_log=True)
+            result = process.wait_result(5)
+            if result.exit_code != 0 and "[sudo] password for" in result.stderr:
+                self.log.debug(
+                    "Running commands with sudo in this node needs input of password."
+                )
+                ssh_shell = cast(SshShell, self.shell)
+                ssh_shell.set_sudo_required_password(True)
```
squirrelsc: The `set_sudo_required_password` is not in this commit. The commits should be organized by change purpose, not file names.

squirrelsc: try to use git add -p, git reset -p to split/merge changes in one file.

lisa/util/shell.py
```diff
@@ -285,6 +291,9 @@ def is_connected(self) -> bool:
             is_inner_shell_ready = True
         return is_inner_shell_ready
 
+    def set_sudo_required_password(self, value: Optional[bool]) -> None:
```
squirrelsc: It looks the value doesn't need to be None again, so the declaration is like, `def set_sudo_required_password(self, value bool) -> None:`

lisa/util/process.py
```diff
@@ -230,6 +230,17 @@ def wait_result(
         timer = create_timer()
         is_timeout = False
 
+        if (
+            self._sudo
+            and isinstance(self._shell, SshShell)
+            and self._shell._is_sudo_required_password
```
squirrelsc: `_is_sudo_required_password` is a private variable, so it shouldn't be used outside. Mentioned in another comment, make it public.

lisa/util/process.py
```diff
@@ -230,6 +230,17 @@ def wait_result(
...
+        ):
+            # Add a 0.5s delay to make sure the channel is ready to receive
+            time.sleep(0.5)
```
squirrelsc: It slowdown the system, it means every command need to wait for 0.5 seconds at least. Move the check into the next while loop. If it's running more than 0.5, execute the logic once.

lisa/util/process.py
```diff
@@ -230,6 +230,17 @@ def wait_result(
...
+            if self.is_running():
+                assert self._process
+                self._process.stdin_write(f"{self._shell._connection_info.password}\n")
```
squirrelsc: I think in some case, it doesn't need to input password for every commands. I see some system ask for password every 15 minutes. So it needs to check if the output buffer contains the password prompt, and then input it.

lisa/util/process.py
```diff
@@ -411,6 +435,30 @@ def _check_if_need_input_password(self, raw_input: str) -> None:
                 " which is not support in Lisa now"
             )
 
+    def _filter_sudo_required_password_info(self, raw_input: str) -> str:
```
squirrelsc, line 439: As mentioned in other comments, don't assume the content is the same. Detect it with "sudo echo OK".

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1145,10 +1145,11 @@ def _create_deployment_parameters(
             node_context.vm_name = azure_node_runbook.name
             # ssh related information will be filled back once vm is created. If
             # it's Windows, fill in the password always. If it's Linux, the
-            # private key has higher priority.
+            # private key has higher priority to connect a VM. The password is also
+            # filled in the node context if it is defined in runbook
             node_context.username = arm_parameters.admin_username
             if azure_node_runbook.is_linux:
-                node_context.password = arm_parameters.admin_password
+                node_context.password = self.runbook.admin_password
```
squirrelsc: Move this change to later PR.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1145,10 +1145,11 @@ def _create_deployment_parameters(
             node_context.vm_name = azure_node_runbook.name
             # ssh related information will be filled back once vm is created. If
             # it's Windows, fill in the password always. If it's Linux, the
-            # private key has higher priority.
+            # private key has higher priority to connect a VM. The password is also
+            # filled in the node context if it is defined in runbook
```
squirrelsc: This comment should be closer to the related line of code.

lisa/util/process.py
lubaihua33: Only the command is sudo and the node is required input of password, then will enter that branch. 
I have moved the check to the while loop.

lisa/util/process.py
lubaihua33: Yeah, put the check here is better. So we don't need the check in suport_sodo, right?

lisa/util/process.py
lubaihua33: Only the first check output in suport_sudo is "sudo echo OK". Why detect it with "sudo echo OK"?

lisa/schema.py
lubaihua33: Done. 
The password is coming from node_context, which is related to the below line change:
node_context.password = self.runbook.admin_password

Following another comment ”Move this change to later PR.“, so drop the whole commit.

lisa/node.py
lubaihua33: Done

lisa/node.py
lubaihua33: Moved these changes in one commit.

lisa/util/shell.py
lubaihua33: Done. I deleted this fun. is_sudo_required_password is changed to public.

lisa/util/process.py
lubaihua33: Done

lisa/util/process.py
lubaihua33: Done

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: Done. Drop this commit. I will create another PR when we need.

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: Drop this commit.

lisa/node.py
squirrelsc: ok, let's make the connection_info public.

lisa/schema.py
squirrelsc: This place is to support set both key and password in general schema. It shouldn't be related with node_context. But it's ok to move it into separated PR.

lisa/util/process.py
```diff
@@ -232,6 +246,8 @@ def wait_result(
 
         while self.is_running() and timeout >= timer.elapsed(False):
             time.sleep(0.01)
+            if timer.elapsed(False) > 0.5:
+                self.check_and_input_password()
```
squirrelsc, line 252: If the password is inputed, it shouldn't run again.

lisa/util/process.py
squirrelsc: > Yeah, put the check here is better. So we don't need the check in suport_sodo, right?

Not sure I fully understand. check password required or not happens only once for a shell or node, so it's hard to be done in process level, it's ok to put it in node level, but shell maybe ok as well. When the sudo password is required, it has two situations, 1) need it every time, 2) need it every x minutes. It needs to be detected in process level, so the password can be hacked into wait_result.

for the two situations, it needs to check if password is needed by the related prompt.

lisa/util/process.py
squirrelsc: I mean, when use "sudo echo OK". the password prompt should be saved, because the prompt can be changed, or localized.

lisa/node.py
lubaihua33: Done

lisa/util/process.py
squirrelsc: Not to put stdout to stderr. We need to use logging.LoggerAdapter() to combine stderr and stdout to stdout. So all outputs are in stdout. Please let me know, if it's unclear.

lubaihua33: I added the line:
self._stdout_writer.flush()
when process is timed out. The "[sudo] password for xx" string is not in the stderr, it is in the _stdout_writer._buffer. Because there is no "\n", so it doesn't flush into _log_buffer. We need to flush manually. Then we can get the "[sudo] password for xx" string from process_result.stdout.

lisa/util/process.py
```diff
@@ -221,6 +221,20 @@ def _process_command(
...
+            and isinstance(self._shell, SshShell)
+            and self._shell.is_sudo_required_password
+        ):
+            if not self._shell.connection_info.password:
+                raise RequireUserPasswordException(
+                    "Running commands with sudo requires user's password,"
+                    " but no password is provided."
+                )
+            assert self._process
+            self._process.stdin_write(f"{self._shell.connection_info.password}\n")
```
squirrelsc, line 236: Add a debug log to say the password is input. So we can know if the logic works right or not.

lisa/util/process.py
lubaihua33: Yeah, I have added a flag to check.

lisa/node.py
lubaihua33: Have added code to save the password prompt.

lisa/util/process.py
lubaihua33: Done

lisa/util/process.py
lubaihua33: Done

lisa/util/process.py
lubaihua33: Changed to check in suport_sudo. From the images I have seen, they all have this behavior that after inputting password successfully, it no needs to input password again if in the same session. If in a new session, it still needs to input password. In Lisav3, every command is a new session, so needs input password every time.

As we discussed, I changed to "echo OK" twice, then I can get all kinds of password prompts.

squirrelsc:
@LiliDeng LGTM. Please find an appropriate time to merge.

LiliDeng (approved):

PR #2730, opened 2023-04-27
Check if command needs input of password quickly

lubaihua33:


lisa/util/process.py
```diff
@@ -403,9 +399,9 @@ def _filter_profile_error(self, raw_input: str) -> str:
         return raw_input
 
     def _check_if_need_input_password(self, raw_input: str) -> None:
-        # Check if the stdout contains "[sudo] password for .*: " and
-        # "sudo: timed out reading password" strings. If so, raise exception
-        if re.search(REQUIRE_INPUT_PASSWORD_PATTERN, raw_input):
+        # Check if the stdout contains "[sudo] password for" string.
+        # If so, raise exception
+        if self._sudo and "[sudo] password for" in raw_input:
```
squirrelsc, line 404: Why not use the pattern?

lubaihua33: I found this way can also work well, so I deleted the pattern.

PR #2731, opened 2023-04-27
DPDK: Make test nic choice consistent

mcgov:
DPDK is sensitive to numa/core/hugepage/nic association settings. Since tests do not make use of more than one NIC at a time, make NIC choice consistent (choose second nic).

This should reduce weird test failures on larger sizes, on environments with more than two nics, etc.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -188,6 +188,14 @@ class Nics(InitializableMixin):
     def get_nic(self, nic_name: str) -> NicInfo:
         return self.nics[nic_name]
 
+    def get_primary_nic(self) -> NicInfo:
+        return self.get_nic_by_index(0)
+
+    def get_secondary_nic(self) -> NicInfo:
+        # get a nic which isn't servicing the SSH connection with lisa.
+        # will assert if none is present.
+        return self.get_nic_by_index(1)
+
     def get_nic_by_index(self, index: int = -1) -> NicInfo:
         # get nic by index, default is -1 to give a non-primary nic
         # when there are more than one nic on the system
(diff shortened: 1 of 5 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -138,7 +138,7 @@ def verify_dpdk_ovs(
 
         try:
             # run OVS tests, providing OVS with the NIC info needed for DPDK init
-            ovs.setup_ovs(node.nics.get_nic_by_index().pci_slot)
+            ovs.setup_ovs(node.nics.get_nic_by_index(1).pci_slot)
```
squirrelsc: does `1` mean first non-primary nic? A well named method can help it clearer.

mcgov: Good point! Added a commit to make some clearer getter functions, get_primary and get_secondary (see updates)

lisa/nic.py
```diff
@@ -188,6 +188,14 @@ def get_nic_driver(self, nic_name: str) -> str:
     def get_nic(self, nic_name: str) -> NicInfo:
         return self.nics[nic_name]
 
+    def get_primary_nic(self) -> NicInfo:
```
squirrelsc, line 191: It calls itself... Is it right?

mcgov: oof, thank you 🙂

squirrelsc:
@LiliDeng  LGTM

LiliDeng (approved):

LiliDeng (approved):

PR #2732, opened 2023-04-28
support to covert sizeInBytes for image os disk size

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1374,9 +1374,18 @@ def _create_node_arm_parameters(
                 arm_parameters.location, arm_parameters.marketplace
             )
             if image_info:
+                additional_properties = image_info.os_disk_image.additional_properties
+                if "sizeInGb" in additional_properties:
```
squirrelsc: It can be simpler like
```python
osdisk_size_in_gb = additional_properties.get("sizeInGb", 0)
if not osdisk_size_in_gb:
    osdisk_size_in_gb = round(additional_properties("sizeInBytes", 0) / 1024 / 1024 / 1024)

squirrelsc (dismissed):

PR #2739, opened 2023-05-01
fix validate_network_interface_reload_via_ip_link, put all commands in one command

LiliDeng:
after bring down and up the network interface.
if no ip address then run dhclient; if no default ip route then run ip route add

### DIFF
--- a/lisa/tools/ip.py
+++ b/lisa/tools/ip.py
@@ -81,10 +81,29 @@ class Ip(Tool):
                 f"ip addr add {ip} dev {nic_name}"
             )
 
-    def restart_device(self, nic_name: str, run_dhclient: bool = False) -> None:
-        cmd = f"ip link set dev {nic_name} down;ip link set dev {nic_name} up;"
+    def restart_device(
+        self,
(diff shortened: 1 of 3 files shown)


lisa/tools/ip.py
```diff
@@ -81,10 +81,21 @@ def add_ipv4_address(self, nic_name: str, ip: str, persist: bool = True) -> None
...
 
-    def restart_device(self, nic_name: str, run_dhclient: bool = False) -> None:
-        cmd = f"ip link set dev {nic_name} down;ip link set dev {nic_name} up;"
+    def restart_device(
+        self,
+        nic_name: str,
+        run_dhclient: bool = False,
+        default_route: str = "",
+    ) -> None:
+        cmd = f"ip link set dev {nic_name} down;ip link set dev {nic_name} up "
```
squirrelsc, line 90: Add comments to explain the logic.

LiliDeng: updated

squirrelsc (dismissed):

PR #2740, opened 2023-05-01
Remove CI exclusion for 3.12 tests on Windows

avylove:
Pywin32 released version 306 on March 25th adding tags for 3.12.
Pillow released version 9.5.0 on April 1st adding a pre-built binary for 3.12.

These were the only blockers previously identified for installing Azure dependencies so the unit tests would run with Python 3.12. This PR updates Pillow and removes the exclusion.


squirrelsc:
@LiliDeng  LGTM

LiliDeng (approved):

PR #2742, opened 2023-05-03
dhclient ethX to get ip address for old distro

LiliDeng:


microsoft/testsuites/network/common.py
```diff
@@ -43,6 +53,8 @@ def initialize_nic_info(
         node_nic_info = Nics(node)
         node_nic_info.initialize()
         for _, node_nic in node_nic_info.nics.items():
+            if not node_nic.ip_addr:
```
squirrelsc, line 57: Add code comments to explain the new logic.

LiliDeng: updated

squirrelsc (approved):

PR #2744, opened 2023-05-03
Skip docker tests on freebsd

dsrivastavv:


microsoft/testsuites/docker/dockersuite.py
```diff
@@ -211,6 +211,11 @@ def _skip_if_not_supported(self, node: Node) -> None:
             raise SkippedException(
                 f"Test not supported for RH/CentOS {node.os.information.release}"
             )
+        elif isinstance(node.os, BSD):
+            raise SkippedException(
+                f"Docker is not supported to run natively on {node.os.name}. "
+                "PLease check the supported distros here: https://docs.docker.com/engine/install"
```
squirrelsc: 1. “PLease" to "Please"
2. Use the unsupported OS exception, and convert it to skipped exception. It gives unified error messages.

dsrivastavv: Didn't get the second point. You mean raise the UnsupportedDistro exception, catch it and then raise a skipped exception?

squirrelsc: like this, https://github.com/microsoft/lisa/blob/10fa8dff95625723d0d7f6f8a06bb6a3a095b980/microsoft/testsuites/tvm/tvmsuite.py#L95

dsrivastavv: fixed. Also moved it to docker tool since that is the correct place, as docker as a tool is not supported

lisa/tools/docker.py
```diff
@@ -137,6 +142,15 @@ def _install(self) -> bool:
             self.node.os.install_packages(["moby-engine", "moby-cli"])
         elif isinstance(self.node.os, SLES):
             self.node.os.install_packages(["docker"])
+        elif isinstance(self.node.os, BSD):
+            raise SkippedException(
```
squirrelsc: Don't skip in tool level. Raise exception here, and skip in test case level.

dsrivastavv: Any specific reason? Shouldn't a testcase be skipped always if the distro is freebsd?

squirrelsc: The skip is a test case behavior. For tools, it just describes what happened.

squirrelsc (approved):

PR #2745, opened 2023-05-04
Added TDX case in verify_isolation_config, and fixed bit masking

Jose-Teuttli:
TDX was added as a valid_cvm_type. An if case was added for it as well.

Currently for TDX /w paravisor HV_SHARED_GPA_BOUNDARY_ACTIVE and HV_SHARED_GPA_BOUNDARY_BITS are not settled so that case is left blank for now.

Also fixed bit selection logic, now using bit masks instead.

### DIFF
--- a/microsoft/testsuites/acc/cvm.py
+++ b/microsoft/testsuites/acc/cvm.py
@@ -26,6 +26,21 @@ class CVMSuite(TestSuite):
         r"(?P<config_a>(0x[a-z,A-Z,0-9]+)), Group B.(?P<config_b>(0x[a-z,A-Z,0-9]+))"
     )
 
+    # HYPERV_CPUID_ISOLATION_CONFIG.EAX Structure
+    #     UINT32 ParavisorPresent : 1;
+    #     UINT32 Reserved0 : 31;
+    HV_PARAVISOR_PRESENT_MASK = 0b1
+
+    # HYPERV_CPUID_ISOLATION_CONFIG.EBX Structure
+    #     UINT32 IsolationType : 4;
+    #     UINT32 Reserved11 : 1;
+    #     UINT32 SharedGpaBoundaryActive : 1;
+    #     UINT32 SharedGpaBoundaryBits : 6;
+    #     UINT32 Reserved12 : 20;
+    HV_ISOLATION_TYPE_MASK = 0b1111
+    HV_SHARED_GPA_BOUNDARY_ACTIVE_MASK = 0b100000
+    HV_SHARED_GPA_BOUNDARY_BITS_MASK = 0b111111000000
+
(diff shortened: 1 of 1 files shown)


Jose-Teuttli:

microsoft/testsuites/acc/cvm.py
```diff
@@ -26,6 +26,21 @@ class CVMSuite(TestSuite):
         r"(?P<config_a>(0x[a-z,A-Z,0-9]+)), Group B.(?P<config_b>(0x[a-z,A-Z,0-9]+))"
     )
 
+    # HYPERV_CPUID_ISOLATION_CONFIG.EBX Structure
+    #     UINT32 ParavisorPresent : 1;
+    #     UINT32 Reserved0 : 31;
+    HV_PARAVISOR_PRESENT_MASK = 1
+
+    # HYPERV_CPUID_ISOLATION_CONFIG.EBX Structure
```
LiliDeng, line 34: the structure name is the same with line 29?

Jose-Teuttli: Whoops, the structure that holds the paravisor bit is actually HYPERV_CPUID_ISOLATION_CONFIG.EAX

Thanks for catching this.

microsoft/testsuites/acc/cvm.py
```diff
@@ -26,6 +26,21 @@ class CVMSuite(TestSuite):
...
+    #     UINT32 IsolationType : 4;
+    #     UINT32 Reserved11 : 1;
+    #     UINT32 SharedGpaBoundaryActive : 1;
+    #     UINT32 SharedGpaBoundaryBits : 6;
+    #     UINT32 Reserved12 : 20;
+    HV_ISOLATION_TYPE_MASK = 15
+    HV_SHARED_GPA_BOUNDARY_ACTIVE_MASK = 32
+    HV_SHARED_GPA_BOUNDARY_BITS_MASK = 4032
```
kamalca: I have a personal preference for bit masks to be expressed in binary, but others may disagree
``` python
HV_ISOLATION_TYPE_MASK = 0b1111
HV_SHARED_GPA_BOUNDARY_ACTIVE_MASK = 0b100000
HV_SHARED_GPA_BOUNDARY_BITS_MASK = 0b111111000000
```
Similar comment for the asserts below
``` python
 assert_that(
    isolation_config_b & self.HV_SHARED_GPA_BOUNDARY_BITS_MASK
).is_equal_to(0b101110000000)
```

Jose-Teuttli: I like the binary version more as well. Swapped over to it.

Thanks for the feedback.

LiliDeng (approved):

PR #2750, opened 2023-05-05
update message when vm in bad status caused by kernel panic

LiliDeng:
per previous code logic, when vm in bad state + kernel panic detected, we will set state as FAILED + panic message here https://github.com/microsoft/lisa/blob/main/lisa/runners/lisa_runner.py#L364
since the state is FAILED which is equal with the previous state of current test result, so the _send_result_message not invoked.
before this change
```
cannot connect to TCP port: [IP:PORT], error code: 10060
```
after this change
```
cannot connect to TCP port: [IP:PORT], error code: 10060 after_case found panic in serial log: ['[   18.054362][  T630] RIP: 0010:hv_irq_unmask+0xaf/0x390 [pci_hyperv]\r', '[   18.322878][  T630] RIP: 0010:hv_irq_unmask+0xaf/0x390 [pci_hyperv]\r']
```
not sure the impact of current change, maybe add one more parameter in set_status to indicate send result or not? @squirrelsc

### DIFF
--- a/lisa/testsuite.py
+++ b/lisa/testsuite.py
@@ -173,16 +173,20 @@ class TestResult:
     def set_status(
         self, new_status: TestStatus, message: Union[str, List[str]]
     ) -> None:
+        send_result = False
         if message:
             if isinstance(message, str):
                 message = [message]
             if self.message:
                 message.insert(0, self.message)
             self.message = "\n".join(message)
+            send_result = True
         if self.status != new_status:
             self.status = new_status
             if new_status == TestStatus.RUNNING:
                 self._timer = create_timer()
+            send_result = True
+        if send_result:
             self._send_result_message(self.stacktrace)
 
     def check_environment(


lisa/testsuite.py
```diff
@@ -178,6 +178,7 @@ def set_status(
             if self.message:
                 message.insert(0, self.message)
             self.message = "\n".join(message)
+            self._send_result_message(self.stacktrace)
```
squirrelsc: 1. It may cause two results sent out together. Update the logic to send only once. Either status changed, or new message coming in.
2. Please add more comments to describe the situation. The environment may found errors after test case completed. So it needs to update for message only.
3. If a test case is passed, and then found the kernel panic, how to handle it? How can we make sure the kernel panic happening in this test case, not previous one?

LiliDeng: I add `send` parameter for set_status. 
This situation only happens after test case run, the node can't be connected, then will check whether it has panic or not, at that time, the env is marked as bad, so it is impossible the panic caused by previous case.

squirrelsc: It's ok to send again, if new message to carry. We can see what happened. But the current logic may send duplicate message on same result. Try below

```python
        send_result = False
        if message:
            if isinstance(message, str):
                message = [message]
            if self.message:
                message.insert(0, self.message)
            self.message = "\n".join(message)
            send_result = True
        if self.status != new_status:
            self.status = new_status
                self._timer = create_timer()
            send_result = True
        if send_result:
            self._send_result_message(self.stacktrace)
```

squirrelsc (approved):

PR #2753, opened 2023-05-08
add case verify_enable_kprobe

LiliDeng:


microsoft/testsuites/core/debug.py
```diff
@@ -0,0 +1,115 @@
...
+
+@TestSuiteMetadata(
+    area="core",
+    category="functional",
+    description="""
+    This test suite covers kernel debug functionalities.
+    """,
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class Debug(TestSuite):
```
squirrelsc: `KernelDebug`

lisa/tools/mount.py
```diff
@@ -150,6 +153,19 @@ def get_mount_point_for_partition(self, partition_name: str) -> Optional[str]:
...
+        output: str = self.run(force_run=True).stdout
+        mount_points: List[str] = []
+        for line in output.splitlines():
+            matched = self._mount_info_regex.fullmatch(line)
+            if matched:
+                mount_points.append(
+                    matched.group("mount_point"),
+                )
+
+        self._log.debug(f"Found all mount points : {mount_points}")
```
squirrelsc: "Found mount points: "

lisa/tools/mount.py
```diff
@@ -150,6 +153,19 @@ def get_mount_point_for_partition(self, partition_name: str) -> Optional[str]:
 
         return partition.mount_point
 
+    def check_mount_point_exist(self, mount_point: str) -> bool:
+        output: str = self.run(force_run=True).stdout
+        mount_points: List[str] = []
+        for line in output.splitlines():
```
squirrelsc: Use `util.find_groups_in_lines`

microsoft/testsuites/core/debug.py
```diff
@@ -0,0 +1,118 @@
...
+                    )
+                ).is_equal_to("p:kprobes/my filp_close")
+                assert_that(
+                    cat.read(
+                        "/sys/kernel/debug/tracing/events/kprobes/my/enable",
+                        sudo=True,
+                        force_run=True,
+                    )
+                ).is_equal_to("1")
+            except Exception as ex:
```
squirrelsc: It do nothing, so can be totally removed. `try... finally...` is enough.

microsoft/testsuites/core/debug.py
```diff
@@ -0,0 +1,118 @@
...
+
+                cat = node.tools[Cat]
+                assert_that(
+                    cat.read(
+                        "/sys/kernel/debug/tracing/kprobe_events",
+                        sudo=True,
+                        force_run=True,
+                    )
+                ).is_equal_to("p:kprobes/my filp_close")
+                assert_that(
```
squirrelsc: add describe for each assertion.

microsoft/testsuites/core/debug.py
```diff
@@ -0,0 +1,118 @@
...
+                    ),
+                    sudo=True,
+                )
+                echo.write_to_file(
+                    origin_kprobe_events,
+                    node.get_pure_path("/sys/kernel/debug/tracing/kprobe_events"),
+                    sudo=True,
+                )
+        else:
+            raise SkippedException("CONFIG_KPROBE_TRACING is not enabled")
```
squirrelsc: Raise exception at the beginning, so it's not a long `if...else`. It helps to understand logic easier.

lisa/tools/mount.py
LiliDeng: updated

microsoft/testsuites/core/debug.py
LiliDeng: updated

microsoft/testsuites/core/debug.py
LiliDeng: updated

microsoft/testsuites/core/debug.py
LiliDeng: updated

microsoft/testsuites/core/debug.py
```diff
@@ -0,0 +1,128 @@
...
+         /sys/kernel/debug/tracing/events/kprobes/my/enable.
+
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+    )
+    def verify_enable_kprobe(self, node: Node) -> None:
+        if not node.tools[KernelConfig].is_enabled("CONFIG_KPROBE_EVENTS"):
+            raise SkippedException("CONFIG_KPROBE_TRACING is not enabled")
+        else:
```
squirrelsc: The `else` is not necessary. When the `if` block is true, and exception happens, the following code won't be executed. Removing the `else` will reduce code complex.

LiliDeng: updated

squirrelsc (approved):

PR #2754, opened 2023-05-08
Sleep & Timeout: add coreutils Sleep tool and first use

mcgov:
Add a tool for coreutils sleep.
- Add use in Timeout for adding a delay before start.
- Add use of both in dpdk for delaying sender to start after receiver

### DIFF
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -347,9 +347,14 @@ class DpdkTestpmd(Tool):
     def run_for_n_seconds(self, cmd: str, timeout: int) -> str:
         self._last_run_timeout = timeout
         self.node.log.info(f"{self.node.name} running: {cmd}")
-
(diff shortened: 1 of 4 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -347,9 +347,14 @@ def generate_testpmd_command(
     def run_for_n_seconds(self, cmd: str, timeout: int) -> str:
         self._last_run_timeout = timeout
         self.node.log.info(f"{self.node.name} running: {cmd}")
-
+        # add delays such that receiver will always start first and end last
```
squirrelsc, line 350: For start first, you can use wait_output to wait for a more reliable signal from its output. The end last can use wait_result to wait for other processes end firstly. It doesn't need to hard code a sleep.

mcgov: This is going to require a much larger refactor

PR #2755, opened 2023-05-08
DPDK: add ping check for other nodes in environment before send receive tests

mcgov:
Add a ping check for the other nodes in the environment before running DPDK send/receive tests.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -1,3 +1,4 @@
+import itertools
 import time
 from collections import deque
 from functools import partial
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -597,7 +597,7 @@ def verify_dpdk_send_receive_netvsc(
             min_nic_count=2,
             network_interface=Sriov(),
             unsupported_features=[Gpu, Infiniband],
-            supported_features=[IsolatedResource],
+            #supported_features=[IsolatedResource],
```
squirrelsc: Lili already add a feature to skip some features in runbook. Please avoid to operate like this, it's easy to bring unwanted changes into PRs.

mcgov: This was just for testing, going to pop that commit before I un-draft the PR

squirrelsc: Branch should be enough for testing. PR is for reviewing. I subscribed notifications on PRs. So if there is some changes on PR, I may take a look.

squirrelsc (approved):

PR #2756, opened 2023-05-09
Save serial console after test failure

adityagesh:
It maybe not be always possible to reproduce a failure. Save serial console after test failure to triage the failure from logs

### DIFF
--- a/lisa/testsuite.py
+++ b/lisa/testsuite.py
@@ -643,7 +643,7 @@ class TestSuite:
             )
 
             if case_result.status == TestStatus.FAILED:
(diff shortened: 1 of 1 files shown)


lisa/testsuite.py
```diff
@@ -663,6 +668,14 @@ def start(
     def stop(self) -> None:
         self._should_stop = True
 
+    def __write_serial_log_to_file(self, nodes: Nodes, log_path: Path) -> None:
```
squirrelsc: call it like `__save_serial_log`, and pass in the environment.

adityagesh: Updated

squirrelsc (approved):

PR #2757, opened 2023-05-09
Fix Azure Image Standard test for FreeBSD

dsrivastavv:
Tests fixed:
- `verify_resource_disk_file_system`: Fixed
- `verify_resource_disk_readme_file`: Fixed
- `verify_boot_error_fail_warnings` : Can't fix as the error seems legit. Error found: `error: Fssh_kex_exchange_identification: Connection closed by remote host`
- `verify_grub`: Skipped

### DIFF
--- a/lisa/tools/lsblk.py
+++ b/lisa/tools/lsblk.py
@@ -3,11 +3,22 @@
 
 import re
 from dataclasses import dataclass, field
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Type
 
 from lisa.executable import Tool
-from lisa.operating_system import FreeBSD, Posix
+from lisa.operating_system import BSD, Posix
 from lisa.util import LisaException, find_patterns_groups_in_lines
+
+
+def _get_size_in_bytes(size: int, size_unit: str) -> int:
+    if size_unit == "G":
+        return size * 1024 * 1024 * 1024
+    elif size_unit == "M":
+        return size * 1024 * 1024
+    elif size_unit == "K":
+        return size * 1024
+    else:
+        raise LisaException(f"Unknown size unit {size_unit}")
 
 
 @dataclass
@@ -125,11 +136,13 @@ class Lsblk(Tool):
     def can_install(self) -> bool:
         return True
 
+    @classmethod
+    def _freebsd_tool(cls) -> Optional[Type[Tool]]:
+        return BSDLsblk
+
     def _install(self) -> bool:
         if isinstance(self.node.os, Posix):
             self.node.os.install_packages("util-linux")
-        elif isinstance(self.node.os, FreeBSD):
-            self.node.os.install_packages("lsblk")
         else:
             raise LisaException(
                 f"tool {self.command} can't be installed in distro {self.node.os.name}."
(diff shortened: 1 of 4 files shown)


lisa/tools/docker.py
```diff
@@ -137,6 +137,13 @@ def _install(self) -> bool:
             self.node.os.install_packages(["moby-engine", "moby-cli"])
         elif isinstance(self.node.os, SLES):
             self.node.os.install_packages(["docker"])
+        elif isinstance(self.node.os, BSD):
```
squirrelsc: Should it be in a separated PR?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -992,9 +993,10 @@ def verify_resource_disk_readme_file(self, node: RemoteNode) -> None:
         node.features[Disk].get_partition_with_mount_point(resource_disk_mount_point)
 
         # verify lost+found folder exists
-        fold_path = f"{resource_disk_mount_point}/lost+found"
-        folder_exists = node.tools[Ls].path_exists(fold_path, sudo=True)
-        assert_that(folder_exists, f"{fold_path} should be present").is_true()
+        if not isinstance(node.os, BSD):
```
squirrelsc, line 1000: Add comments above to explain why FreeBSD skip this check.

lisa/tools/lsblk.py
```diff
@@ -65,6 +65,10 @@ class DiskInfo(object):
     type: str = ""
     partitions: List[PartitionInfo] = field(default_factory=list)
 
+    @classmethod
+    def _freebsd_tool(cls) -> Optional[Type[Tool]]:
```
squirrelsc: Why does the disk info need this method?

lisa/tools/lsblk.py
```diff
@@ -228,3 +234,109 @@ def find_mountpoint_by_volume_name(
...
+                    dev_type=entry["type"],
+                    partitions=disk_partition_map.get(entry["name"], []),
+                )
+            )
+
+        # sort disk with OS disk first
+        disks.sort(key=lambda disk: disk.is_os_disk, reverse=True)
+        return disks
+
+    @staticmethod
```
squirrelsc: This method doesn't need to be static.

dsrivastavv: We are not using any class information. We can also move it to common function if you feel so

lisa/tools/docker.py
dsrivastavv: Rebased to main

lisa/tools/lsblk.py
dsrivastavv: This was an error. Fixed

lisa/tools/lsblk.py
squirrelsc: You can move it to a function in the same file. It's not shared by other classes, so it doesn't need to move to common.

dsrivastavv:
@squirrelsc Not skipping `verify_repository_installed` testcase until Wei can confirm. But this doesn't need to block this PR

lisa/tools/lsblk.py
```diff
@@ -3,13 +3,24 @@
...
-from typing import Dict, List, Optional
+from typing import Dict, List, Optional, Type
 
 from lisa.executable import Tool
-from lisa.operating_system import FreeBSD, Posix
+from lisa.operating_system import BSD, Posix
 from lisa.util import LisaException, find_patterns_groups_in_lines
 
 
+def get_size_in_bytes(size: int, size_unit: str) -> int:
```
squirrelsc: Because it's not used out of this file, still make it preview starts with underscore.

dsrivastavv: fixed

squirrelsc (approved):

PR #2758, opened 2023-05-11
fix verify_gpu_cuda_with_pytorch test fail

paxue:
fix verify_gpu_cuda_with_pytorch test fail, it needs 5GB free space

### DIFF
--- a/lisa/tools/df.py
+++ b/lisa/tools/df.py
@@ -10,10 +10,11 @@ from lisa.util import find_patterns_groups_in_lines
 
 
 class Df(Tool):
-    # devtmpfs         4071644       0   4071644   0% /dev
+    # Filesystem                1K-blocks  Used Available Use% Mounted on
+    # /dev/mapper/rootvg-homelv   1038336 40472    997864   4% /home
     _DF_ENTRY_REGEX = re.compile(
-        r"(?P<name>\S+)\s+(?P<blocks>\d+)\s+(?P<used>\d+)\s+"
-        r"(?P<total>\d+)\s+(?P<percentage_use>\d+)%\s+(?P<mountpoint>\S+)"
+        r"(?P<name>\S+)\s+(?P<total>\d+)\s+(?P<used>\d+)\s+"
+        r"(?P<available>\d+)\s+(?P<percentage_use>\d+)%\s+(?P<mountpoint>\S+)"
     )
 
     @property
@@ -37,7 +38,7 @@ class Df(Tool):
                 PartitionInfo(
                     name=df_entry["name"],
                     mountpoint=df_entry["mountpoint"],
-                    available_blocks=int(df_entry["blocks"]),
+                    available_blocks=int(df_entry["available"]),
                     used_blocks=int(df_entry["used"]),
                     total_blocks=int(df_entry["total"]),
                     percentage_blocks_used=int(df_entry["percentage_use"]),
@@ -54,3 +55,22 @@ class Df(Tool):
             if partition.mountpoint == mountpoint:
                 return partition
         return None
+
+    def get_filesystem_available_space(
+        self, path: str = ".", force_run: bool = False
+    ) -> float:
+        # return the input 'path' available space in GB by running df
+        # If 'path' doesn't exist, return 0
+        # The df cmd output is in the following format:
+        # <Filesystem> <1K-blocks> <Used> <Available> <Use%> <Mounted on>
+        # Note : <Available> are in 1-k blocks, not in bytes
+        available_space = 0.0
+        out = self.run(parameters=path, sudo=True, force_run=force_run).stdout
+
+        df_entries = find_patterns_groups_in_lines(out, [self._DF_ENTRY_REGEX])[0]
+        if len(df_entries) == 0:  # path not exist
+            self._log.debug(f"Path: {path} not exist! return 0 available space.")
+            return 0.0
+
+        available_space = (float)(df_entries[0]["available"]) / 1024**2
+        return available_space
--- a/lisa/tools/python.py
+++ b/lisa/tools/python.py
@@ -7,6 +7,8 @@ from assertpy import assert_that
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
+from lisa.tools.chmod import Chmod
+from lisa.tools.mkdir import Mkdir
 from lisa.util import UnsupportedDistroException, get_matched_str
 
 
@@ -49,17 +51,33 @@ class Pip(Tool):
         self.node.os.install_packages(package_name)
         return self._check_exists()
 
-    def install_packages(self, packages_name: str) -> None:
-        cmd_result = self.run(
-            f"install -q {packages_name}",
-        )
+    def install_packages(self, packages_name: str, install_path: str = "") -> None:
+        node = self.node
+        cmd_line = f"install -q {packages_name}"
+
+        envs = {}
+        if install_path != "":
+            tagert_path = install_path + "/python_packages"
+            node.tools[Mkdir].create_directory(tagert_path, sudo=True)
+            cache_path = install_path + "/tmp"
+            node.tools[Mkdir].create_directory(cache_path, sudo=True)
+            # In some Ubuntu, find_partition_with_freespace() return root "/"
+            # for "install_path", we can't run chmod on "/", need to chmod 1 by 1
+            node.tools[Chmod].update_folder(tagert_path, "777", sudo=True)
+            node.tools[Chmod].update_folder(cache_path, "777", sudo=True)
+
+            cmd_line += f" -t {tagert_path} --cache-dir={cache_path} -b {cache_path}"
+            # Since Python 3.9, pip 21.2, -b for build path has been deprecated
+            # Using TMPDIR/TMP/TEMP Env Variable instead
+            envs = {"TMPDIR": cache_path}
+
+        cmd_result = self.run(cmd_line, update_envs=envs)
+
         if 0 != cmd_result.exit_code and get_matched_str(
             cmd_result.stdout, self._no_permission_pattern
         ):
-            cmd_result = self.run(
-                f"install -q {packages_name}",
-                sudo=True,
-            )
+            cmd_result = self.run(cmd_line, update_envs=envs, sudo=True)
+
         assert_that(
             cmd_result.exit_code, f"fail to install {packages_name}"
         ).is_equal_to(0)
--- a/microsoft/testsuites/gpu/gpusuite.py
+++ b/microsoft/testsuites/gpu/gpusuite.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import os
 import re
 from pathlib import Path
 from typing import Any, List
@@ -22,7 +23,19 @@ from lisa.features import Gpu, GpuEnabled, SerialConsole, StartStop
 from lisa.features.gpu import ComputeSDK
 from lisa.operating_system import AlmaLinux, Debian, Oracle, Suse, Ubuntu
 from lisa.sut_orchestrator.azure.features import AzureExtension
-from lisa.tools import Lspci, NvidiaSmi, Pip, Python, Reboot, Service, Tar, Wget
+from lisa.tools import (
+    Chmod,
+    Df,
+    Lspci,
+    Mkdir,
+    NvidiaSmi,
+    Pip,
+    Python,
+    Reboot,
+    Service,
+    Tar,
+    Wget,
+)
 from lisa.util import UnsupportedOperationException, get_matched_str
 
 _cudnn_location = (
@@ -247,21 +260,43 @@ class GpuTestSuite(TestSuite):
         _install_driver(node, log_path, log)
         _check_driver_installed(node, log)
 
-        _install_cudnn(node)
+        # Step 1, pytorch and CUDA needs 4GB space to download and install
+        torch_required_space = 5
 
-        gpu = node.features[Gpu]
+        work_path = str(node.working_path)
+        lisa_path_space = node.tools[Df].get_filesystem_available_space(work_path)
+        if lisa_path_space < torch_required_space:
+            work_path = node.find_partition_with_freespace(torch_required_space)
+
+        # Step 2, Install cudnn and pyTorch
+        _install_cudnn(node, log, work_path)
 
         pip = node.tools[Pip]
         if not pip.exists_package("torch"):
-            pip.install_packages("torch")
+            if lisa_path_space < torch_required_space:
+                # pip install to target work_path
+                pip.install_packages("torch", work_path)
+            else:
+                # using default parameter for pip install
+                pip.install_packages("torch")
 
+        # Step 3, verification
+        gpu = node.features[Gpu]
         gpu_script = "import torch;print(f'gpu count: {torch.cuda.device_count()}')"
         python = node.tools[Python]
         expected_count = gpu.get_gpu_count_with_lspci()
 
+        if lisa_path_space < torch_required_space:
+            python_path = os.environ.get("PYTHONPATH", "")
+            python_path += f":{work_path}/python_packages"
+            python_envs = {"PYTHONPATH": python_path}
+        else:
+            python_envs = {}
+
         script_result = python.run(
             f'-c "{gpu_script}"',
             force_run=True,
+            update_envs=python_envs,
         )
         gpu_count_str = get_matched_str(script_result.stdout, self._pytorch_pattern)
         script_result.assert_exit_code(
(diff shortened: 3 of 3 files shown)


LiliDeng:
I tried with `redhat rhel 9_1 9.1.2022112113`
it failed for 
```
failed on run gpu script: import torch;print(f'gpu count: {torch.cuda.device_count()}'), output: Traceback (most recent call last):

  File "<string>", line 1, in <module>

ModuleNotFoundError: No module named 'torch'
get unexpected exit code on cmd ['python3', '-c', "import torch;print(f'gpu count: {torch.cuda.device_count()}')"]] Expected <[0]> to contain item <1>, but did not
```

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +247,28 @@ def verify_gpu_cuda_with_pytorch(
...
+            sudo=True,
+        )
+        work_path = free_partition + "/lisawork"
+        log.debug(f"Current work dir is {work_path}")
 
-        gpu = node.features[Gpu]
+        _install_cudnn(node, log, work_path)
 
         pip = node.tools[Pip]
+        pip._install()
```
squirrelsc: 1. don't call private method out if its class.
2. install will be called automatically, it doesn't need to be called explicitly.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +247,28 @@ def verify_gpu_cuda_with_pytorch(
...
         if not pip.exists_package("torch"):
-            pip.install_packages("torch")
+            node.execute(
+                f"pip3 install torch -q --cache-dir={work_path} -b {work_path}",
```
squirrelsc: 1. add cache dir and -b in pip tool arguments. So others can reuse the logic.
2. refer to https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH, use PythonPath to add the extra path into python tool.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +247,28 @@ def verify_gpu_cuda_with_pytorch(
...
+                shell=True,
+                sudo=True,
```
squirrelsc: Don't overuse sudo.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +247,28 @@ def verify_gpu_cuda_with_pytorch(
         _install_driver(node, log_path, log)
         _check_driver_installed(node, log)
 
-        _install_cudnn(node)
+        # pytorch and CUDA needs 4GB space to download and install
+        free_partition = node.find_partition_with_freespace(5)
+        node.execute(
```
squirrelsc: use mkdir tool

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +247,28 @@ def verify_gpu_cuda_with_pytorch(
...
+            f"cd {free_partition}; mkdir lisawork; chmod -R 777 lisawork;",
+            shell=True,
+            sudo=True,
+        )
+        work_path = free_partition + "/lisawork"
```
squirrelsc: the folder name should be like python_packages.

microsoft/testsuites/gpu/gpusuite.py
paxue: removed

microsoft/testsuites/gpu/gpusuite.py
paxue: This command need to use sudo since it's writting the lib to "/usr/local/lib" path. I will remove sudo and use system pip3 install function as it will check permission and use sudo if needed

microsoft/testsuites/gpu/gpusuite.py
paxue: changed

squirrelsc:
Use `git rebase -i`, `git add -p`, `git reset -p` to make commits meaningful.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +246,33 @@ def verify_gpu_cuda_with_pytorch(
         _install_driver(node, log_path, log)
         _check_driver_installed(node, log)
 
-        _install_cudnn(node)
+        # pytorch and CUDA needs 4GB space to download and install
+        free_partition = node.find_partition_with_freespace(5)
+        work_path = free_partition + "/python_packages"
+        node.tools[Mkdir].create_directory(work_path, sudo=True)
```
squirrelsc: If sudo is not required, don't use it.

paxue:
> 



> 

Sure. This is not final version. There will be another fix for redhat 9.1 failure

lisa/tools/python.py
```diff
@@ -49,17 +49,30 @@ def _install(self) -> bool:
...
-        ):
+    def install_packages(self, packages_name: str, cache_dir="") -> None:
+        if (cache_dir == ""):
             cmd_result = self.run(
                 f"install -q {packages_name}",
-                sudo=True,
             )
+        else:
+            cmd_result = self.run(
+                f"install -q {packages_name} --cache-dir={cache_dir} -b {cache_dir}",
```
squirrelsc: Compose the command before run it. So it won't appear multiple times, like,
```python
cmd = f"install -q {packages_name}"
if cache_dir:
  cmd += f" --cache-dir={cache_dir} -b {cache_dir}"
```

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +246,33 @@ def verify_gpu_cuda_with_pytorch(
         _install_driver(node, log_path, log)
         _check_driver_installed(node, log)
 
-        _install_cudnn(node)
+        # pytorch and CUDA needs 4GB space to download and install
+        free_partition = node.find_partition_with_freespace(5)
+        work_path = free_partition + "/python_packages"
```
squirrelsc: The directory logic should be moved into pip tool. So just pass the installation path in, and then create directories inside.

lisa/tools/python.py
```diff
@@ -49,17 +49,30 @@ def _install(self) -> bool:
...
         return self._check_exists()
 
-    def install_packages(self, packages_name: str) -> None:
-        cmd_result = self.run(
-            f"install -q {packages_name}",
-        )
-        if 0 != cmd_result.exit_code and get_matched_str(
-            cmd_result.stdout, self._no_permission_pattern
-        ):
+    def install_packages(self, packages_name: str, cache_dir="") -> None:
```
squirrelsc: the cache_dir and -b are two different folders. It's not good practice to put them in the same folder, but it's ok, if it doesn't break.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +246,33 @@ def verify_gpu_cuda_with_pytorch(
...
+        node.tools[Mkdir].create_directory(work_path, sudo=True)
 
-        gpu = node.features[Gpu]
+        node.execute(
+            f"chmod -R 777 {work_path};",
```
squirrelsc: Use 755 by default.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +246,33 @@ def verify_gpu_cuda_with_pytorch(
...
+
+        _install_cudnn(node, log, work_path)
 
         pip = node.tools[Pip]
+        # pip.install_packages("torch")
+
         if not pip.exists_package("torch"):
-            pip.install_packages("torch")
+            pip.install_packages("torch", work_path)
+            # node.execute(
```
squirrelsc: Remove unused comments.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +246,33 @@ def verify_gpu_cuda_with_pytorch(
...
+        node.tools[Mkdir].create_directory(work_path, sudo=True)
 
-        gpu = node.features[Gpu]
+        node.execute(
+            f"chmod -R 777 {work_path};",
+            shell=True,
+            sudo=True,
+        )
+        # work_path = free_partition + "/lisawork"
+        log.debug(f"Current work dir is {work_path}")
```
squirrelsc: it doesn't need to print out again, I think it's already output in previous logs.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,14 +246,33 @@ def verify_gpu_cuda_with_pytorch(
...
-            pip.install_packages("torch")
+            pip.install_packages("torch", work_path)
+            # node.execute(
+            #     f"pip3 install torch -q --cache-dir={work_path} -b {work_path}",
+            #     shell=True,
+            #     sudo=True,
+            # )
 
+        gpu = node.features[Gpu]
         gpu_script = "import torch;print(f'gpu count: {torch.cuda.device_count()}')"
```
squirrelsc, line 285: I haven't seen the new package path is included in PythonPath environment variable, how to make sure it works?

microsoft/testsuites/gpu/gpusuite.py
paxue: This does needs Sudo. I tried no sudo it will fail

lisa/tools/python.py
paxue: changed

microsoft/testsuites/gpu/gpusuite.py
paxue: I tried 755, but not work, build folder needs to use 777

microsoft/testsuites/gpu/gpusuite.py
paxue: removed

microsoft/testsuites/gpu/gpusuite.py
paxue: Previously I don't add "--target" in the command, which means it will install to the default folder.
In the next change I will add PythonPath

lisa/tools/python.py
paxue: > I tried with `redhat rhel 9_1 9.1.2022112113` it failed for
> 
> ```
> failed on run gpu script: import torch;print(f'gpu count: {torch.cuda.device_count()}'), output: Traceback (most recent call last):
> 
>   File "<string>", line 1, in <module>
> 
> ModuleNotFoundError: No module named 'torch'
> get unexpected exit code on cmd ['python3', '-c', "import torch;print(f'gpu count: {torch.cuda.device_count()}')"]] Expected <[0]> to contain item <1>, but did not
> ```

Redhat 9.1 using a different python(3.9) and pip version, which made them does not treat "-b" parameter. Changed to update TMP folder.

microsoft/testsuites/gpu/gpusuite.py
paxue: changed

lisa/tools/python.py
paxue: > 

I added both build and cache parameters.

squirrelsc: It doesn't need to add both path, because the path is used to provide more space. You can call the path like "target_path", and create a path for cache, and another one for installed packages in the pip tool. So it's clear for external and internal of the pip tool both.

microsoft/testsuites/gpu/gpusuite.py
paxue: changed

lisa/tools/python.py
```diff
@@ -49,17 +49,40 @@ def _install(self) -> bool:
...
+    ) -> None:
+        node = self.node
+        cmd_line = f"install -q {packages_name}"
+
+        if install_path != "":
+            tagert_path = install_path + "/python_packages"
+            node.tools[Mkdir].create_directory(tagert_path, sudo=True)
+            tmp_path = install_path + "/tmp"
+            node.tools[Mkdir].create_directory(tmp_path, sudo=True)
+            node.execute(
```
squirrelsc: use chmod tool. Change the install_path once is enough.

lisa/tools/python.py
```diff
@@ -49,17 +49,40 @@ def _install(self) -> bool:
...
+        packages_name: str,
+        install_path=""
+    ) -> None:
+        node = self.node
+        cmd_line = f"install -q {packages_name}"
+
+        if install_path != "":
+            tagert_path = install_path + "/python_packages"
+            node.tools[Mkdir].create_directory(tagert_path, sudo=True)
+            tmp_path = install_path + "/tmp"
```
squirrelsc: call the variable name "cache_path" is more accurate. Avoid to use "temp", if the usage is clear.

lisa/tools/python.py
```diff
@@ -49,17 +49,40 @@ def _install(self) -> bool:
...
+            node.tools[Mkdir].create_directory(tmp_path, sudo=True)
+            node.execute(
+                f"chmod -R 777 {tagert_path}; chmod -R 777 {tmp_path}",
+                shell=True,
+                sudo=True,
+            )
+            cmd_line = cmd_line + f" --target {tagert_path} --cache-dir={tmp_path} -b {tmp_path}"
+            env_list = {"TMPDIR": tmp_path}
+        else:
+            env_list = {}
```
squirrelsc: 1. The variable name should be `envs`. Avoid to include type in name, an it's not a "list".
2. The envs can be initialized before the `if` block. So the `else` block is not necessary. The logic will be simpler.

lisa/tools/python.py
```diff
@@ -49,17 +49,40 @@ def _install(self) -> bool:
...
+            node.tools[Mkdir].create_directory(tagert_path, sudo=True)
+            tmp_path = install_path + "/tmp"
+            node.tools[Mkdir].create_directory(tmp_path, sudo=True)
+            node.execute(
+                f"chmod -R 777 {tagert_path}; chmod -R 777 {tmp_path}",
+                shell=True,
+                sudo=True,
+            )
+            cmd_line = cmd_line + f" --target {tagert_path} --cache-dir={tmp_path} -b {tmp_path}"
+            env_list = {"TMPDIR": tmp_path}
```
squirrelsc: Add comments above to explain the reason to set TMPDIR.

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -247,21 +247,26 @@ def verify_gpu_cuda_with_pytorch(
...
-            pip.install_packages("torch")
+            pip.install_packages("torch", free_partition)
 
+        gpu = node.features[Gpu]
         gpu_script = "import torch;print(f'gpu count: {torch.cuda.device_count()}')"
         python = node.tools[Python]
         expected_count = gpu.get_gpu_count_with_lspci()
+        python_path = os.environ.get('PYTHONPATH', '')
 
+        torch_install_path = free_partition + "/python_packages"
```
squirrelsc: The `torch_install_path` is used once, so the variable not necessarily. And the path composition logic should be put together, so it's easy to read. See below,

```python
        python_path = os.environ.get('PYTHONPATH', '')
        python_path += f":{free_partition}/python_packages"

        script_result = python.run(
            f'-c "{gpu_script}"',
            force_run=True,
            update_envs={"PYTHONPATH": python_path}
        )
```

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -311,25 +316,37 @@ def _check_driver_installed(node: Node, log: Logger) -> None:
...
+        shell=True,
+        sudo=True,
+    )
+
+    log.debug(f"CUDNN Extracted path is: {work_path}  ")
     download_path = wget.get(
-        url=_cudnn_location, filename=str(_cudnn_file_name), file_path=str(path)
+        url=_cudnn_location, filename=str(_cudnn_file_name), file_path=work_path
+    )
+    node.execute(
```
squirrelsc: use the tar tool

squirrelsc (dismissed):

squirrelsc (dismissed):

lisa/tools/python.py
```diff
@@ -49,17 +51,31 @@ def _install(self) -> bool:
...
+        node = self.node
+        cmd_line = f"install -q {packages_name}"
+
+        envs = {}
+        if install_path != "":
+            tagert_path = install_path + "/python_packages"
+            node.tools[Mkdir].create_directory(tagert_path, sudo=True)
+            cache_path = install_path + "/tmp"
+            node.tools[Mkdir].create_directory(cache_path, sudo=True)
+            node.tools[Chmod].update_folder(tagert_path, "777", sudo=True)
```
squirrelsc, line 66: why use chmod one by one? The command use recursive already, the install_path should be applied for both sub directories.

paxue: In some Ubuntu env, the find_partition_with_freespace() will return root "/" folder for "install_path". If we run "chmod install_path", it will change mode for everything in the root folder, which is not correct.

lisa/tools/df.py
```diff
@@ -10,10 +10,11 @@
 
 
 class Df(Tool):
-    # devtmpfs         4071644       0   4071644   0% /dev
+    # Filesystem                1K-blocks  Used Available Use% Mounted on
+    # /dev/mapper/rootvg-homelv   1038336 40472    997864   4% /home
     _DF_ENTRY_REGEX = re.compile(
-        r"(?P<name>\S+)\s+(?P<blocks>\d+)\s+(?P<used>\d+)\s+"
```
paxue, line 15: The old DF tools have bugs, the parsing part total and available are swapped.

lisa/tools/df.py
```diff
@@ -54,3 +55,19 @@ def get_partition_by_mountpoint(
             if partition.mountpoint == mountpoint:
                 return partition
         return None
+
+    def get_filesystem_available_space(
+        self, path: str = ".", force_run: bool = False
+    ) -> float:
+        # run df to get the available space and parse the output
```
squirrelsc: Explain the returned value is GB.

lisa/tools/df.py
```diff
@@ -54,3 +55,21 @@ def get_partition_by_mountpoint(
...
+        # return the input 'path' available space in GB by running df
+        # If 'path' doesn't exist, return 0
+        # The df cmd output is in the following format:
+        # <Filesystem> <1K-blocks> <Used> <Available> <Use%> <Mounted on>
+        # Note : <Available> are in 1-k blocks, not in bytes
+        available_space = 0.0
+        out = self.run(parameters=path, sudo=True, force_run=force_run).stdout
+
+        df_entries = find_patterns_groups_in_lines(out, [self._DF_ENTRY_REGEX])[0]
+        if len(df_entries) == 0:  # path not exist
```
squirrelsc, line 71: Add debug level log to explain the path is not found, returned 0 available space.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2761, opened 2023-05-12
Add ignore_error parameter for make.make

adityagesh:
Make.make currently raises exception for failures.
Default value of ignore_error is set to False to maintain current behavior as default.
Ignore_error can be set to True to not raise exception. The caller has to compare the result using result.assert_exit_code()

### DIFF
--- a/lisa/tools/make.py
+++ b/lisa/tools/make.py
@@ -66,7 +66,9 @@ class Make(Tool):
         timeout: int = 600,
         thread_count: int = 0,
         update_envs: Optional[Dict[str, str]] = None,
+        ignore_error: bool = False,
     ) -> ExecutableResult:
+        expected_exit_code: Optional[int] = 0
         if thread_count == 0:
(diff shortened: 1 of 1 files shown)


lisa/tools/make.py
```diff
@@ -84,15 +85,25 @@ def make(
...
-            f"yes '' | make -j{thread_count} {arguments}",
-            cwd=cwd,
-            timeout=timeout,
-            sudo=sudo,
-            shell=True,
-            update_envs=update_envs,
-            expected_exit_code=0,
-            expected_exit_code_failure_message="Failed to make",
-        )
+        if not ignore_error:
```
squirrelsc: It doesn't need to copy the method. Use default value for expected_exit_code and failure_message.

adityagesh: updated

LiliDeng (dismissed):

LiliDeng (approved):

PR #2763, opened 2023-05-15
add sizeInBytes for data disk

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2061,10 +2061,19 @@ def _generate_data_disks(
             #  storage profile.
             if marketplace:
                 for default_data_disk in marketplace.data_disk_images:
+                    additional_properties = default_data_disk.additional_properties
```
squirrelsc: Share the logic with another place to handle the addtional properties.

squirrelsc (approved):

PR #2768, opened 2023-05-17
node: add eligible_vm_sizes to random pick up vm size from given vm sizes list

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -195,6 +195,7 @@ class AzureNodeSchema:
     # It decides the real computer name. It cannot be too long.
     short_name: str = ""
     vm_size: str = ""
+    eligible_vm_sizes: Optional[List[str]] = None
```
squirrelsc, line 198: This shouldn't be a feature in LISA. It should be configured in pipeline.

LiliDeng: I didn't find a parameter which supports to pick up the size from the given list.

PR #2769, opened 2023-05-17
fix meson build issue

LiliDeng:


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -838,19 +839,22 @@ def _install_ninja_and_meson(self) -> None:
                 "Failed to update Meson to latest version with pip3"
             ),
         )
-        if node.shell.exists(node.get_pure_path("/usr/bin/meson")):
-            node.tools[Mv].move(
-                "/usr/bin/meson", "/usr/bin/meson.bak", overwrite=True, sudo=True
+
+        if node.shell.exists(node.get_pure_path("/usr/local/bin/meson")):
+            if node.shell.exists(node.get_pure_path("/usr/bin/meson")):
```
squirrelsc: Add comments to explain why move it can copy the local one here. If it doesn't need to recover, how about deleting it directly? If the test case runs multiple times, the file will be overwritten anyway.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -838,19 +839,22 @@ def _install_ninja_and_meson(self) -> None:
...
+                node.tools[Mv].move(
+                    "/usr/bin/meson", "/usr/bin/meson.bak", overwrite=True, sudo=True
+                )
+            node.execute(
```
squirrelsc, line 845: If the local file doesn't exist, what happened?

microsoft/testsuites/dpdk/dpdktestpmd.py
LiliDeng: added

squirrelsc (approved):

PR #2770, opened 2023-05-17
Add FreeBSD support for hv module testcases

dsrivastavv:


lisa/tools/kernel_config.py
```diff
@@ -22,20 +23,27 @@ def can_install(self) -> bool:
         return False
 
     def is_built_in(self, config_name: str) -> bool:
-        return (
-            self.node.execute(
-                f"grep ^{config_name}=y {self.config_path}", sudo=True, shell=True
-            ).exit_code
-            == 0
-        )
+        if isinstance(self.node.os, BSD):
```
squirrelsc: It looks very different on each method, it's worth to create a subclass for BSD.

dsrivastavv:
@whu2014 can you please take a look at this PR?

lisa/tools/kernel_config.py
```diff
@@ -53,3 +58,19 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
             self.config_path = f"/usr/boot/config-{kernel_ver}"
         else:
             self.config_path = f"/boot/config-{kernel_ver}"
+
+
+class KernelConfigFreeBSD(KernelConfig):
+    def is_built_in(self, config_name: str) -> bool:
+        return self.node.tools[KLDStat].module_statically_linked(config_name)
+
+    def is_built_as_module(self, config_name: str) -> bool:
```
squirrelsc, line 66: This method is exactly same as KernelConfig. So it doesn't need to be inherited.

lisa/tools/lsmod.py
```diff
@@ -48,3 +52,16 @@ def module_exists(
...
 
         return False
+
+
+class KLDStat(Lsmod):
+    @property
+    def command(self) -> str:
+        return "kldstat"
+
+    def module_statically_linked(
```
squirrelsc: Is it the same as `module_exists`? If so, the method should be called `module_exists`. At other places, whatever the distro is, it should always use `lsmod`, and let LISA decide which tool to use.

dsrivastavv: No this is different. `lsmod.module_exists` looks at the dynamically linked modules

lisa/tools/kernel_config.py
dsrivastavv: I made an error while fixing earlier comment. It should be fixed now

lisa/tools/lsmod.py
squirrelsc: If so, should it be implemented in kernel_config? The lsmod doesn't list static built modules.

dsrivastavv: This information in freebsd is available from kldstat command

dsrivastavv: Should I add kldstat tool separately and not linked to lsmod?

squirrelsc: LISA is designed for Linux-central on the tool abstract layer. So we don't need a tool for BSD independently. It breaks the design. As my understanding, the kldstat should be included in kernel_configs, because Linux kernel config is only place to get static module information.

squirrelsc (approved):

PR #2771, opened 2023-05-18
handle skip test results and long str for vm_size

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -880,7 +880,10 @@ def _get_environment_information(self, environment: Environment) -> Dict[str, st
             )
             if node_runbook:
                 information["location"] = node_runbook.location
-                information["vmsize"] = node_runbook.vm_size
+                if not node and len(node_runbook.vm_size.split(",")) > 0:
```
squirrelsc: The runbook may not be right place to get the vm sizes. Can you check if other place is better here?

LiliDeng: I updated code, I don't want to override the vmsize in all skip case, since it may caused by the picked up size doesn't have quota or specified vm size doesn't provide the capability for run case. we need keep the vmsize to know the details.

squirrelsc (approved):

PR #2772, opened 2023-05-18
Replace the blob name as the original vhd path in the error message

lubaihua33:
When the vhd has the wrong format or dynamic type, the error message prints the normalized vhd name but not the original one. The normalized vhd name will confuse users. This PR replaces the normalized vhd name as the original vhd name in the error message.

Before fix:
The error message like:
deployment failed. LisaException: InvalidParameter: The specified cookie value in VHD footer indicates that disk '**99990101/https---vcertsasouthcentralus-blob-core-windows-net-vhds-fortinet-fortinet-fortisiem-vm-fortinet-fortisiem-byol-blob-6-7-4-vhd-sv-2019-07-07-sr-b-sig-UYEGActDShEhn3ojE-2BCIIgrz8plT0-3D-st-2023-05-04T14-3A11-3A07Z-se-2023-05-19T14-3A11-3A07Z-sp-r.vhd**' with blob **https://lisatwestus36a28e896.blob.core.windows.net:8443/lisa-sas-copied/99990101/https---vcertsasouthcentralus-blob-core-windows-net-vhds-fortinet-fortinet-fortisiem-vm-fortinet-fortisiem-byol-blob-6-7-4-vhd-sv-2019-07-07-sr-b-sig-UYEG3ojE-2BCIIgrz8plT0-3D-st-2023-05-04T14-3A11-3A07Z-se-2023-05-19T14-3A11-3A07Z-sp-r.vhd** is not a supported VHD. Disk is expected to have cookie value 'conectix'.

After fix:
The error message like:
deployment failed. LisaException: InvalidParameter: The specified cookie value in VHD footer indicates that disk **fortinet-byol-blob-6-7-4.vhd?**** with blob **https://userstorageaccount.blob.core.windows.net/vhds/fortinet-byol-blob-6-7-4.vhd?**** is not a supported VHD. Disk is expected to have cookie value 'conectix'.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -60,7 +60,7 @@ from lisa import schema
 from lisa.environment import Environment, load_environments
 from lisa.feature import Features
 from lisa.node import Node, RemoteNode, local
-from lisa.secret import PATTERN_HEADTAIL, PATTERN_URL, add_secret
+from lisa.secret import PATTERN_HEADTAIL, PATTERN_URL, add_secret, replace
 from lisa.util import (
     LisaException,
     LisaTimeoutException,
@@ -1630,6 +1630,31 @@ def get_deployable_vhd_path(
     full_vhd_path = copy_vhd_to_storage(
         platform, storage_name, original_vhd_path, vhd_path, log
     )
+    log.debug(
+        f"deployable vhd is cached at: {full_vhd_path}. "
+        f"original vhd url: {original_vhd_path}"
+    )
+
+    # Add substitutes for these normalized vhd names. Replace them with the original
+    # vhd name without SAS when printing
+    # e.g.
+    # full_vhd_path: https://lisatwestus3xx.blob.core.windows.net/lisa-sas-copied/99990101/https---userstorageaccount-blob-core-windows-net-vhds-fortinet-blob-6-7-4-vhd-sv-2019-xx.vhd  # noqa: E501
+    # replace it with: https://userstorageaccount.blob.core.windows.net/vhds/fortinet/blob-6.7.4.vhd?***  # noqa: E501
+    # vhd_path: 99990101/https---userstorageaccount-blob-core-windows-net-vhds-fortinet-blob-6-7-4-vhd-sv-2019-xx.vhd  # noqa: E501
+    # replace it with: fortinet/blob-6.7.4.vhd
+    original_vhd_path_without_sas = replace(original_vhd_path, mask=PATTERN_URL)
+    add_secret(
+        vhd_path,
+        sub="/".join(original_vhd_path_without_sas.split("/")[4:]).rstrip("?*"),
+    )
+    add_secret(full_vhd_path, sub=original_vhd_path_without_sas)
+    # In the returned message from Azure when creating VM, the blob url might contain
+    # the default blob port 8443, so also add substitute for the vhd path with 8443 port
+    full_vhd_path_with_port = "/".join(
+        x + ":8443" if i == 2 else x for i, x in enumerate(full_vhd_path.split("/"))
+    )
+    add_secret(full_vhd_path_with_port, sub=original_vhd_path_without_sas)
+
     return full_vhd_path
 
 
--- a/lisa/util/__init__.py
+++ b/lisa/util/__init__.py
@@ -78,7 +78,9 @@ hookimpl = pluggy.HookimplMarker(_NAME_LISA)
 
 
 class LisaException(Exception):
-    ...
+    def __init__(self, *args: object) -> None:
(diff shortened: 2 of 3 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -216,6 +216,7 @@ class AzureNodeSchema:
     vhd_raw: Optional[Union[Dict[Any, Any], str]] = field(
         default=None, metadata=field_metadata(data_key="vhd")
     )
+    original_vhd_path: str = ""
```
squirrelsc: Don't add it in node schema, it will be exposed as a parameter in runbook, so it's confusing. Move it to AzureNodeContext, so it's only available runtime.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1516,9 +1535,33 @@ def _deploy(
...
+                        environment, error_message
+                    )
+
                 plugin_manager.hook.azure_deploy_failed(error_message=error_message)
                 raise LisaException(error_message)
 
+    def _replace_blob_name_as_original_vhd(
+        self, environment: Environment, error_message: str
+    ) -> str:
+        information = self._get_environment_information(environment)
```
squirrelsc: Because the original vhd path was saved already, it doesn't need to recover it from information. Just use the secret function to replace it.

lubaihua33: I didn't quite understand. You mean the original vhd path is saved in the node_context after I modified code according to your first comment, right? If so, in the _deploy function, I still need to get it from Environment. 

For the comment "Just use the secret function to replace it.", the original vhd path is saved in the secret list already, it will be masked when it is printed. But in the error message, it is still the original format that has the SAS token.  If I understand right, you mean that I can use function replace in secret.py right?

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: I added some changes.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1281,6 +1294,8 @@ def _create_node_runbook(
                 f"but {azure_node_runbook.hyperv_generation}",
             )
         if azure_node_runbook.vhd and azure_node_runbook.vhd.vhd_path:
+            node_context = get_node_context(node)
+            node_context.original_vhd_path = azure_node_runbook.vhd.vhd_path
```
squirrelsc: I think it just needs to call `secreat.add_secret` here. So original vhd path will be automatically replaced to original in log.

lubaihua33: In this PR, I want to fix the error message in the database. The original vhd path is already in the secret list. But the secret just masks the token in the log, it doesn't mask the error message updated in the database. So we need to truly replace it in the error message.

We might have some misunderstanding. Could you review the description of this PR at the top of this page, let me know if you have any concerns.

squirrelsc: Thank you for bring it up! Understood it more.

1. Make sure the masked vhd path still be in log at least once, so we know where the vhd is copied to. We need to log it with existing mask, and then replace the pattern of secret to print original vhd path in future.
2. Use secret lib to filter exceptions, so no secret leak into error message as well.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1630,6 +1630,20 @@ def get_deployable_vhd_path(
     full_vhd_path = copy_vhd_to_storage(
         platform, storage_name, original_vhd_path, vhd_path, log
     )
+    log.debug(f"deployable vhd path: {full_vhd_path}")
+
+    # Add substitutes for these normalized vhd names. Replace them with the original
+    # vhd name without SAS when printing
+    original_vhd_path_without_sas = replace(original_vhd_path, mask=PATTERN_URL)
+    add_secret(vhd_path, sub="/".join(original_vhd_path_without_sas.split("/")[4:]))
```
squirrelsc: Add examples above to explain.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1630,6 +1630,20 @@ def get_deployable_vhd_path(
     full_vhd_path = copy_vhd_to_storage(
         platform, storage_name, original_vhd_path, vhd_path, log
     )
+    log.debug(f"deployable vhd path: {full_vhd_path}")
```
squirrelsc: It should have the original vhd path in the same line, so it's easy to look for, like "deployable vhd is cached at: XXX, original vhd url: xxx."

squirrelsc:
@LiliDeng  LGTM

LiliDeng (approved):

PR #2773, opened 2023-05-18
Move CVM Logic to ARM template

kamalca:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1423,6 +1428,35 @@ def _create_node_arm_parameters(
         if capability.network_interface.data_path == schema.NetworkDataPath.Sriov:
             arm_parameters.enable_sriov = True
 
+        # Convert Security Profile Setting to Arm Parameter
+        _security_profile_mapping = {
```
squirrelsc: The logic is better in the feature than platform. The arm_parameter is passed into the before_deploy as well. It can be changed in feature too.

kamalca: Please correct me if I am wrong.
My consideration was that security profile should be 'per node' not 'per platform'.
If a runbook wanted to specify multiple nodes with different security profile feature settings, this function will run for every node but if I put the code in before_deployment it will only run once and only only security profile feature settings will be provided.

Is this a case I should be considering?

squirrelsc: Understood. The major consideration here is to organize code into features instead of a growing platform code base. So the features is a better place. The `on_before_deployment` runs once for all nodes together, so it can update for deployment level changes. But the environment is passed in, so it still can update for nodes. I haven't created another method for nodes level, because I don't want to bring more complexity in APIs. The flexibility here maybe confusing, but first consideration is to keep platform class handle the general flow, it decouples node level, feature level logics out.

kamalca: I have moved the logic to on_before_deployment.
I did not use the FeatureSettings, instead using the environment node capabilities to build the arm parameters.
Please make sure all my assumptions look safe, for example environment.nodes should always be equal length to arm_parameters.nodes.

squirrelsc: Yes, they are equal. The logic looks not organized, but provides max flexibility.

squirrelsc:
@LiliDeng  LGTM

LiliDeng:
@kamalca I encountered below issue when run tvm tests against image `canonical 0001-com-ubuntu-server-focal 20_04-lts-gen2 latest`

TvmTest.verify_secureboot_compatibility: FAILED   deployment failed. LisaException: BadRequest: The parameter 'managedDisk.securityProfile' can only be set for Confidential VMs.
TvmTest.verify_measuredboot_compatibility: FAILED   deployment failed. LisaException: BadRequest: The parameter 'managedDisk.securityProfile' can only be set for Confidential VMs.

kamalca:
> @kamalca I encountered below issue when run tvm tests against image `canonical 0001-com-ubuntu-server-focal 20_04-lts-gen2 latest`
> 
> TvmTest.verify_secureboot_compatibility: FAILED deployment failed. LisaException: BadRequest: The parameter 'managedDisk.securityProfile' can only be set for Confidential VMs. TvmTest.verify_measuredboot_compatibility: FAILED deployment failed. LisaException: BadRequest: The parameter 'managedDisk.securityProfile' can only be set for Confidential VMs.

Corrected the logic so that secure boot does not have security profile for the disk.

LiliDeng (approved):

PR #2774, opened 2023-05-19
Fix KVP testcase for FreeBSD

dsrivastavv:


microsoft/testsuites/core/kvp.py
```diff
@@ -65,31 +74,36 @@ def verify_kvp(self, node: Node, log: Logger) -> None:
...
-        )
-        hv_kvp = get_matched_str(result.stdout, self._hv_kvp_pattern)
-        assert_that(
-            hv_kvp,
-            f"Cannot find hv_kvp in '/proc/{kvp_daemon_pid}/fd'. "
-            "the kernel may not support hv_kvp.",
-        ).is_equal_to("hv_kvp")
+        if not isinstance(node.os, BSD):
+            result = node.execute(
+                f"ls -al /proc/{kvp_daemon_pid}/fd",
```
dsrivastavv, line 88: @whu2014 Can you please clarify if this check is valid for BSD?

dsrivastavv:
@squirrelsc Got a confirmation from Wei that this PR is good to go

lisa/tools/pgrep.py
```diff
@@ -42,3 +46,28 @@ def get_processes(self, process_identifier: str) -> List[ProcessInfo]:
...
+    @property
+    def command(self) -> str:
+        return "ps"
+
+    def get_processes(self, process_identifier: str) -> List[ProcessInfo]:
+        running_process: List[ProcessInfo] = []
+        output = self.run("-axceo user,pid,command", sudo=True, force_run=True).stdout
+        found_processes = find_patterns_in_lines(output, [self._process_map_regex])
+        running_process.extend(
+            ProcessInfo(name=item[2], pid=item[1]) for item in found_processes[0]
```
squirrelsc: Try to use names instead of like `item[2]`

dsrivastavv: I am guessing you mean something like `ProcessInfo(name=item.name, pid=item.pid)` but `item` here is a tuple. Can you please clarify?

squirrelsc: Use `find_groups_in_lines`, so it supports named items.

squirrelsc (approved):

PR #2776, opened 2023-05-22
NVSv4 - skip on linux since current it doesn't support on linux

LiliDeng:


lisa/features/gpu.py
```diff
@@ -137,6 +137,9 @@ def install_compute_sdk(self, version: str = "") -> None:
                     except Exception as e:
                         raise LisaException(f"Failed to install CUDA Driver {str(e)}")
                     self.gpu_vendor.add("nvidia")
+            elif driver == ComputeSDK.AMD:
+                # refer https://learn.microsoft.com/en-us/azure/virtual-machines/nvv4-series  # noqa: E501
+                raise SkippedException("AMD GPU is not supported in Linux yet.")
```
squirrelsc: I cannot find any information about AMD GPU doesn't support Linux in this page. If so, please clarify, it's not supported on Azure, or general not support Linux.

LiliDeng: should be this link https://learn.microsoft.com/en-us/azure/virtual-machines/sizes-gpu

image

LiliDeng: this page https://learn.microsoft.com/en-us/azure/virtual-machines/nvv4-series also refers to Windows only
image

squirrelsc: Good to know. I think it means the NVv4 series supports windows only. it doesn't mean the AMD GPU supports Windows only. Can we exclude the VM sizes in GPU feature, and add comment to explain there?

squirrelsc (approved):

PR #2777, opened 2023-05-22
node: add get_working_path_with_required_space

LiliDeng:


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -825,7 +827,33 @@ def _install_fedora_dependencies(self) -> None:
         if not self.use_package_manager_install():
             self._install_ninja_and_meson()
 
+    def _resize_home_folder(self) -> None:
```
squirrelsc: Can it be installed in other frees pace partition? Extend the home folder may not be appliable on all distros.

LiliDeng: now I just saw space is not enough on LVM images, here it only resizes when it is LVM partition. I will double check with finding a place with enough space.

squirrelsc: Panfeng fixed a similar disk space issue on gpu test cases. Can we have a common method to handle this?

lisa/tools/python.py
```diff
@@ -76,7 +72,7 @@ def install_packages(self, packages_name: str, install_path: str = "") -> None:
         if 0 != cmd_result.exit_code and get_matched_str(
             cmd_result.stdout, self._no_permission_pattern
         ):
-            cmd_result = self.run(cmd_line, update_envs=envs, sudo=True)
+            cmd_result = self.run(cmd_line, update_envs=envs)
```
squirrelsc: The sudo should be kept, because it's next try, if permission is required.

LiliDeng: I add chmod 777 in get_working_path_with_required_space and tested, all good

squirrelsc: If so, the Ln72 - 75 should be removed. But you can check on code history to know which distro need the sudo.

LiliDeng: oh oh, you are right, we can't assume this method always is invoked after get_working_path_with_required_space

squirrelsc (approved):

PR #2781, opened 2023-05-24
remove \n\t line when output contains 'sudo: unable to resolve host'

LiliDeng:


lisa/tools/lspci.py
```diff
@@ -122,9 +122,7 @@ def get_devices(self, force_run: bool = False) -> List[PciDevice]:
                 expected_exit_code=0,
                 sudo=True,
             )
-            for pci_raw in [
-                line for line in result.stdout.splitlines() if line.strip()
-            ]:
+            for pci_raw in result.stdout.splitlines():
```
squirrelsc, line 125: Why is it changed back?

LiliDeng: the change in _filter_sudo_result handle this blank output.

squirrelsc (approved):

PR #2783, opened 2023-05-26
Support for NVMe disk_controller_type

SRIKKANTH:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1134,61 +1134,80 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
-                self._log.debug(
-                    "download udev rules to construct a set of symbolic links "
-                    "under the /dev/disk/azure path"
-                )
-                if ls_tools.is_file(
-                    self._node.get_pure_path("/dev/disk/azure"), sudo=True
+        disk_array: List[str] = []
+
+        if (
+            self._node.capability.disk
```
LiliDeng, line 1138: can we have a new case, and put this logic in case level?

SRIKKANTH: We are still testing the same premium disks with disk_controller_type==NVMe. The only difference is that they are exposed to guest as NVMe devices. The reason is to go beyond the limits of SCSI protocol so we are using NVMe protocol/driver instead of SCSI.  So we can still use the same testcase.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1134,61 +1134,80 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+            # https://docs.microsoft.com/en-us/troubleshoot/azure/virtual-machines/troubleshoot-device-names-problems#identify-disk-luns  # noqa: E501
+            # /dev/disk/azure/scsi1/lun0
+            ls_tools = self._node.tools[Ls]
+            files = ls_tools.list("/dev/disk/azure/scsi1", sudo=True)
+
+            if len(files) == 0:
+                os = self._node.os
+                # there are known issues on ubuntu 16.04 and rhel 9.0
+                # try to workaround it
+                if (isinstance(os, Ubuntu) and os.information.release <= "16.04") or (  # noqa: E501,BLK100
```
squirrelsc, line 1158: Ubuntu 1604 is out of life cycle, we don't need to take care it.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1134,61 +1134,80 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+
+        if (
+            self._node.capability.disk
+            and self._node.capability.disk.disk_controller_type
+            == schema.DiskControllerType.NVME
+        ):
+            nvme = self._node.features[Nvme]
+            # Skip OS disk which is '[0]' in namespaces list
+            disk_array = nvme.get_namespaces()[1:]
+        else:
```
squirrelsc: It doesn't need to put other code in `else`.

LiliDeng (approved):

PR #2786, opened 2023-05-31
Remove the color code from stdout stream while getting vmbus version in dmesg tool

smit-gardhariya:
This change will remove the color code from stdout of dmesg while getting the vmbus version. Sometime i have faced the issue of VMBus version not getting filtered with regex due to presence of the color code.

### DIFF
--- a/lisa/tools/dmesg.py
+++ b/lisa/tools/dmesg.py
@@ -26,6 +26,8 @@ class Dmesg(Tool):
         r"\[\s+\d+.\d+\]\s+hv_vmbus:.*Vmbus version:(?P<major>\d+).(?P<minor>\d+)"
     )
 
+    __color_code_pattern = r"\x1b\[[0-9;]*m"
+
     @property
     def command(self) -> str:
         return "dmesg"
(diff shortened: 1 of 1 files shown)


lisa/tools/dmesg.py
```diff
@@ -69,7 +69,12 @@ def get_vmbus_version(self) -> VersionInfo:
         result.assert_exit_code(
             message=f"exit code should be zero, but actually {result.exit_code}"
         )
-        raw_vmbus_version = re.finditer(self.__vmbus_version_pattern, result.stdout)
+
+        # Remove the color code from stdout stream
+        color_code_pattern = r"\x1b\[[0-9;]*m"
```
LiliDeng: can you paste a sample lin in the comments?

smit-gardhariya: Hi Lili, i cannot paste the sample in comment as it is related to color code. dmesg stdout stream will have it. Here is the screenshot. VMBus version sample lines are already there in the code.

image

LiliDeng (approved):

PR #2787, opened 2023-05-31
Add testcase "verify_disk_controller_type"

SRIKKANTH:


lisa/features/disks.py
```diff
@@ -64,6 +64,15 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
     def get_resource_disk_mount_point(self) -> str:
         raise NotImplementedError
 
+    # Get disk controller type of the VM by checking the OS disk partition
+    @property
+    def controller_type(self) -> str:
+        os_disk = self.get_partition_with_mount_point("/")
+        if os_disk.disk == 'nvme':  # noqa BLK100
```
LiliDeng: remove  # noqa BLK100

microsoft/testsuites/core/storage.py
```diff
@@ -183,6 +183,36 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+
+        # Get disk controller type.
+        if node.capability.disk:
+            disk_controller_type = node.capability.disk.disk_controller_type
+
+        if (  # noqa E501
+            node.capability.disk
+            and os_disk_controller_type != disk_controller_type  # noqa E501
+        ):
+            raise LisaException(
```
LiliDeng, line 260: remove all  # noqa E501

lisa/features/disks.py
LiliDeng: it is not the case for redhat rhel 92-gen2 9.2.2023052501, it uses lvm partition.
```name: /dev/mapper/rootvg-rootlv, disk: mapper/rootvg-rootlv, mount_point: /, type: xfs, mount_point: /```

File: microsoft/testsuites/core/storage.py [hunk 3]
```diff
@@ -183,6 +183,36 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
             "content read from file should be equal to content written to file",
         ).is_equal_to(original_text)
 
+    @TestCaseMetadata(
+        description="""
+        This test will verify that disk controller type.
+
+        Steps:
+        1. Get the disk type of the os partition.
+        2. Compare it with diskcontrollertype passed through runbook.
```
LiliDeng: can we compare the value got from the os and value return from API level?

microsoft/testsuites/core/storage.py
```diff
@@ -183,6 +183,36 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
             "content read from file should be equal to content written to file",
         ).is_equal_to(original_text)
 
+    @TestCaseMetadata(
+        description="""
+        This test will verify that disk controller type.
```
squirrelsc: Add comments to explain more how the logic works, like a setting is required to test. If not, the logic doesn't verify anything.

microsoft/testsuites/core/storage.py
```diff
@@ -183,6 +183,36 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+        requirement=simple_requirement(
+            supported_platform_type=[AZURE],
+        ),
+    )
+    def verify_disk_controller_type(self, node: RemoteNode) -> None:
+        # Get OS disk type.
+        os_disk_controller_type = node.features[Disk].controller_type
+
+        # Get disk controller type.
+        if node.capability.disk:
```
squirrelsc: It should be a case visible variable, not from the disk config.

microsoft/testsuites/core/storage.py
```diff
@@ -183,6 +183,36 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+            disk_controller_type = node.capability.disk.disk_controller_type
+
+        if (  # noqa E501
```
squirrelsc: Use assert_that

lisa/features/disks.py
SRIKKANTH: Fixed this by changing the logic to verify disc controller type.

PR #2789, opened 2023-06-01
LsvmBus testcase on FreeBSD

dsrivastavv:


lisa/tools/lsvmbus.py
```diff
@@ -218,3 +229,40 @@ def get_device_channels(self, force_run: bool = False) -> List[VmBusDevice]:
                 self._vmbus_devices.append(vmbus_device)
 
         return self._vmbus_devices
+
+
+class VmBusDeviceBSD(VmBusDeviceResult):
```
squirrelsc: It doesn't need a new object for FreeBSD.

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -81,14 +92,18 @@ class LsVmBus(TestSuite):
...
         )
         lsvmbus_tool = node.tools[Lsvmbus]
         vmbus_devices_list = lsvmbus_tool.get_device_channels()
         actual_vmbus_device_names = [x.name for x in vmbus_devices_list]
         assert_that(actual_vmbus_device_names).is_not_none()
         assert_that(vmbus_devices.names).is_subset_of(actual_vmbus_device_names)
 
+        if isinstance(node.os, BSD):
+            # lsvmbus and associated information is not available on FreeBSD
+            return
```
squirrelsc: Let's split it to two cases, so it's clear on coverage.

lisa/tools/lsvmbus.py
```diff
@@ -45,7 +46,13 @@ def __repr__(self) -> str:
         return self.__str__()
 
 
-class VmBusDevice:
+class VmBusDeviceResult:
```
squirrelsc: Why need the new result class?

dsrivastavv: The `VmBusDevice` class is very tied to Linux. So I created a parent class with the basic variables we need to represent a Lsvmbus device

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -121,7 +136,7 @@ def lsvmbus_count_devices_channels(self, node: Node) -> None:
         """,
         priority=4,
         requirement=simple_requirement(
-            supported_platform_type=[AZURE],
+            supported_platform_type=[AZURE], unsupported_os=[BSD]
```
squirrelsc: It should add Windows, because the default value doesn't include Windows.

lisa/tools/lsvmbus.py
squirrelsc: Name the VmBusDevice as LinuxVmBusDeviceParser, and not inherit from VmBusDevice. So it would be clear, the VmBusDevice is the data structure, and the Parser is the logic.

microsoft/testsuites/core/lsvmbus.py
dsrivastavv: Is this applicable for future testcases we migrate too?

squirrelsc: yes

microsoft/testsuites/core/lsvmbus.py
```diff
@@ -75,19 +106,12 @@ class LsVmBus(TestSuite):
         """,
         priority=1,
         requirement=simple_requirement(
-            supported_platform_type=[AZURE],
+            supported_platform_type=[AZURE], unsupported_os=[BSD]
```
squirrelsc: It should add Windows.

dsrivastavv: Missed it. Add it back

squirrelsc (approved):

PR #2790, opened 2023-06-01
Stress-ng Fix for Mariner: Install packages for stress-ng Tool

adityagesh:
Install required packages and binaries which caused stress-ng installation to fail due to compilation issue on Mariner.

### DIFF
--- a/lisa/tools/stress_ng.py
+++ b/lisa/tools/stress_ng.py
@@ -4,7 +4,7 @@
 from typing import cast
 
 from lisa.executable import Tool
-from lisa.operating_system import Posix
+from lisa.operating_system import CBLMariner, Posix
 from lisa.util.process import Process
 
 from .gcc import Gcc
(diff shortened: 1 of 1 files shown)


LiliDeng (approved):

lisa/tools/stress_ng.py
```diff
@@ -76,12 +76,17 @@ def launch_class_async(
...
 
+    def _install_required_packages(self) -> None:
+        if isinstance(self.node.os, CBLMariner):
+            self.node.os.install_packages(["glibc-devel", "kernel-headers", "binutils"])
+
     def _install_from_src(self) -> bool:
         tool_path = self.get_tool_path()
         git = self.node.tools[Git]
         git.clone(self.repo, tool_path, ref=self.branch)
         self.node.tools.get(Gcc)  # Ensure gcc is installed
```
squirrelsc, line 87: gcc is installed with make.

adityagesh: Make does not have the function 'dependencies', but gcc is installed in _install


    def _install(self) -> bool:
        posix_os.install_packages([self, Gcc])


Shall I add dependencies to make and remove the get Gcc here?

squirrelsc: That's the same. Dependencies is installed separately, and the install_packages can install them in the same group.

PR #2791, opened 2023-06-02
Remove before/after suite

adityagesh:
before_suite and after_suite was deprecated in
PR #2662 with 3 month waiting period

### DIFF
--- a/lisa/testsuite.py
+++ b/lisa/testsuite.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass, field
 from functools import wraps
 from pathlib import Path
 from time import sleep
-from typing import Any, Callable, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Type, Union
 from warnings import warn
 
 from func_timeout import FunctionTimedOut, func_timeout  # type: ignore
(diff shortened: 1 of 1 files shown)


lisa/testsuite.py
```diff
@@ -584,18 +582,6 @@ def start(
...
-            (
-                is_suite_continue,
-                suite_error_message,
-                suite_error_stacktrace,
-            ) = self.__suite_method(
-                self.before_suite,  # type: ignore
-                test_kwargs=test_kwargs,
-                log=suite_log,
-            )
-            self.__print_warning("before_suite", "before_case")
```
squirrelsc, line 597: I think we should add one more step, to raise an error, so it's obvious to people to fix.

lisa/testsuite.py
```diff
@@ -583,19 +581,9 @@ def start(
         is_suite_continue = True
         suite_log = self.__log
         suite_error_stacktrace = None
+        if hasattr(self, "before_suite") or hasattr(self, "after_suite"):
+            raise LisaException("before_suite and after_suite is not supported")
```
squirrelsc: The error message is not friendly, because it's not actionable. Reuse existing error messages.

squirrelsc (approved):

LiliDeng:
@adityagesh @squirrelsc don't merge it, I saw other teams are using it. Let's do the communication before merging.

PR #2794, opened 2023-06-05
Fix verify_disks_device_timeout_setting in FreeBSD

dsrivastavv:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1217,8 +1217,17 @@ def get_raw_data_disks(self) -> List[str]:
     def get_all_disks(self) -> List[str]:
         # /sys/block/sda = > sda
```
squirrelsc: 1. Should the examples follow Linux patterns?
2. Move all patterns out of the method like other places.

microsoft/testsuites/core/storage.py
```diff
@@ -65,16 +66,41 @@ def verify_disks_device_timeout_setting(
...
+                    device_timeout_from_distro_unformatted = (
+                        node.tools[Sysctl]
+                        .run(
+                            f"kern.cam.{device_type}.default_timeout",
+                            force_run=True,
+                            shell=True,
+                        )
+                        .stdout
+                    )
+                    device_timeout_from_distro = int(
```
squirrelsc, line 100: use regexp as much as possible.

lisa/sut_orchestrator/azure/features.py
dsrivastavv: I have fixed first comment. For second, I feel moving regex out will make the testcase more complicated since we will add more lines to deal with different regex variable name.

squirrelsc: Moving them out of the method will save some time on compiling regex. It's not high in most time. And move them on top of class, that helps to review and fix regressions around them.

squirrelsc (approved):

PR #2785, opened 2023-05-30
unify test case names

LiliDeng:

PR #2797, opened 2023-06-06
Fix storage testcases

dsrivastavv:
Cases fixed:

- Storage.verify_os_partition_identifier
- Storage.verify_resource_disk_io
- Storage.verify_resource_disk_mtab_entry
- Storage.verify_swap

### DIFF
--- a/microsoft/testsuites/core/storage.py
+++ b/microsoft/testsuites/core/storage.py
@@ -26,7 +26,7 @@ from lisa.schema import DiskType
 from lisa.sut_orchestrator import AZURE
 from lisa.sut_orchestrator.azure.features import AzureDiskOptionSettings
 from lisa.sut_orchestrator.azure.tools import Waagent
-from lisa.tools import Blkid, Cat, Dmesg, Echo, Lsblk, NFSClient, Swap, Sysctl
+from lisa.tools import Blkid, Cat, Dmesg, Echo, Lsblk, Mount, NFSClient, Swap, Sysctl
 from lisa.util import BadEnvironmentStateException, LisaException, get_matched_str
 from lisa.util.perf_timer import create_timer
 
@@ -146,10 +146,18 @@ class Storage(TestSuite):
             .get_partition_with_mount_point(self.os_disk_mount_point)
             .disk
         )
-        mtab = node.tools[Cat].run("/etc/mtab").stdout
-        resource_disk_from_mtab = get_matched_str(
-            mtab, self._get_mtab_mount_point_regex(resource_disk_mount_point)
-        )
+        if isinstance(node.os, BSD):
+            partition_info = node.tools[Mount].get_partition_info()
+            resource_disk_from_mtab = [
+                entry
+                for entry in partition_info
+                if entry.mount_point == resource_disk_mount_point
+            ][0].mount_point
+        else:
+            mtab = node.tools[Cat].run("/etc/mtab").stdout
+            resource_disk_from_mtab = get_matched_str(
+                mtab, self._get_mtab_mount_point_regex(resource_disk_mount_point)
+            )
         assert (
             resource_disk_from_mtab
         ), f"resource disk mountpoint not found {resource_disk_mount_point}"
--- a/microsoft/testsuites/core/storage.py
+++ b/microsoft/testsuites/core/storage.py
@@ -21,7 +21,7 @@ from lisa.features.disks import (
     DiskStandardSSDLRS,
 )
 from lisa.node import Node
-from lisa.operating_system import BSD
+from lisa.operating_system import BSD, Windows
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/core/storage.py
```diff
@@ -112,10 +114,17 @@ def verify_resource_disk_mtab_entry(self, node: RemoteNode) -> None:
             .get_partition_with_mount_point(self.os_disk_mount_point)
             .disk
         )
-        mtab = node.tools[Cat].run("/etc/mtab").stdout
-        resource_disk_from_mtab = get_matched_str(
-            mtab, self._get_mtab_mount_point_regex(resource_disk_mount_point)
-        )
+        if isinstance(node.os, BSD):
+            mtab = node.tools[Mount].run().stdout
+            mtab_mount_point_regex = re.compile(
```
squirrelsc: This kind of pattern should be in tools. I think we already have some logic to parse mount points, is it right? And the regs should always be with examples.

microsoft/testsuites/core/storage.py
```diff
@@ -146,10 +146,18 @@ def verify_resource_disk_mtab_entry(self, node: RemoteNode) -> None:
             .get_partition_with_mount_point(self.os_disk_mount_point)
             .disk
         )
-        mtab = node.tools[Cat].run("/etc/mtab").stdout
-        resource_disk_from_mtab = get_matched_str(
-            mtab, self._get_mtab_mount_point_regex(resource_disk_mount_point)
-        )
+        if isinstance(node.os, BSD):
```
squirrelsc, line 149: Can Linux use the same logic?

squirrelsc, line 149: The test case is to check mtab. If FreeBSD doesn't have it, maybe it can just skip the FreeBSD.

squirrelsc: See test case metadata for details.

dsrivastavv: On linux the output of /etc/mtab and mount command is very similar, and freebsd has mount command. So I feel that this extension makes sense since we are doing similar comparison in both Linux and BSD

squirrelsc: The test case is defined for validating /etc/mtab. Please double check what the test case is for, and if BSD need the similar validation.

dsrivastavv: Sure! I will clarify it from v2

microsoft/testsuites/core/storage.py
dsrivastavv: This is fixed!

microsoft/testsuites/core/storage.py
squirrelsc: What does "v2" mean? I have a look on the test case description. It looks the mtab is the way to verify where the resource disk is mounted. So, BSD is still valid to run this test case, but the test description should include the test methodology of BSD.

dsrivastavv: From v2 documentation:

.. _verify_resource_disk_mtab_entry:

    .. method:: verify_resource_disk_mtab_entry
        :noindex:

        :description: |         This test will check that the resource disk is present in the list of mounted
                      |         devices. Most VMs contain a resource disk, which is not a managed disk and
                      |         provides short-term storage for applications and processes. It is intended to
                      |         only store data such as page or swap files.
                      |         Steps:
                      |         1. Get the mount point for the resource disk. If `/var/log/cloud-init.log`
                      |         file is present, mount location is `/mnt`, otherwise it is obtained from
                      |         `ResourceDisk.MountPoint` entry in `waagent.conf` configuration file.
                      |         2. Verify that "/dev/<disk> <mount_point>` entry is present in
                      |         `/etc/mtab` file and the disk should not be equal to os disk.


I think the intent here is to verify that resource disk is present by getting information from the list of mounted devices. Though I agree that the test case name gives a wrong indication that we instead checking file `/etc/mtab`. How about we add a separate testcase for FreeBSD that checks that resource disk is mounted and we can name it something like: `verify_resource_disk_mounted`

squirrelsc: Please work with Lili to rename the test case. It may need to evaluate impact on changing test case name.

LiliDeng: There is no resource disk handled in freebsd, you can double check. @somil55

LiliDeng: update test case name in PR #2785

dsrivastavv: Thanks @LiliDeng . I can rebase once your PR is merged

squirrelsc (approved):

PR #2798, opened 2023-06-07
Added a test for vmsnapshot_extension

arisettisanjana:
Adding a test to validate vmsnapshot_extension
The test involves creating a restore point and if that gets succeeded, the extension flow is verified.
Changes made:
added a file named vmsnapshot_extension.py

### DIFF
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -5,6 +5,8 @@ import time
 import uuid
 from datetime import datetime
 
+from assertpy.assertpy import assert_that
+
 from lisa import (
     Environment,
     Logger,
@@ -14,7 +16,13 @@ from lisa import (
     TestSuiteMetadata,
     simple_requirement,
 )
-from lisa.sut_orchestrator.azure.common import get_compute_client
+from lisa.sut_orchestrator import AZURE
+from lisa.sut_orchestrator.azure.common import (
+    AzureNodeSchema,
+    get_compute_client,
+    get_node_context,
+    wait_operation,
+)
 from lisa.sut_orchestrator.azure.features import AzureExtension
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
 
@@ -38,20 +46,20 @@ class BVTExtension(TestSuite):
         self, log: Logger, node: Node, environment: Environment
     ) -> None:
         unique_name = str(uuid.uuid4())
-        information = environment.get_information()
-        resource_group_name = information["resource_group_name"]
-        location = information["location"]
-        vm_name = node.name
-        log.info(f"information {information}")
+        node_context = get_node_context(node)
+        resource_group_name = node_context.resource_group_name
+        vm_name = node_context.vm_name
+        node_location = node.capability.get_extended_runbook(AzureNodeSchema, AZURE)
+        location = node_location.location
         restore_point_collection = "rpc_" + unique_name
         assert environment.platform
         platform: AzurePlatform = environment.platform  # type: ignore
         assert isinstance(platform, AzurePlatform)
         sub_id = platform.subscription_id
         # creating restore point collection
-        client = get_compute_client(environment.platform)
+        client = get_compute_client(platform)
         response = client.restore_point_collections.create_or_update(
-            resource_group_name=information["resource_group_name"],
+            resource_group_name=resource_group_name,
             restore_point_collection_name=restore_point_collection,
             parameters={
                 "location": location,
@@ -67,32 +75,32 @@ class BVTExtension(TestSuite):
                 },
             },
         )
-        log.info("rpc created")
+        log.info("restore point collection created")
         log.info(f"response {response}")
         count = 0
 
         while count < 10:
-            vm = client.virtual_machines.get(resource_group_name, vm_name)
-            # check the state of the VM
-            if vm.provisioning_state == "Succeeded":
-                try:
-                    # create a restore point for the VM
-                    response = client.restore_points.begin_create(
-                        resource_group_name=information["resource_group_name"],
-                        restore_point_collection_name=restore_point_collection,
-                        restore_point_name="rp_"
-                        + datetime.now().strftime("%Y-%m-%d-%H-%M-%S"),
-                        parameters={},
-                    )
-                    response.wait(3600)
-                    log.info("restore point created")
-                    break
-                except Exception as e:
-                    if "Changes were made to the Virtual Machine" in str(e):
-                        pass
-                    else:
-                        log.info(f"error {e}")
-                        raise AssertionError("Test failed: Unexpected error occurred")
+            try:
+                # create a restore point for the VM
+                restore_point = "rp_" + datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
+                response = client.restore_points.begin_create(
+                    resource_group_name=resource_group_name,
+                    restore_point_collection_name=restore_point_collection,
+                    restore_point_name=restore_point,
+                    parameters={},
+                )
+                wait_operation(response)
+                log.info(f"restore point {restore_point} created")
+                break
+            except Exception as e:
+                # This message is sometimes seen while rp creation
+                # so we will be retrying it after some time
+                if "Changes were made to the Virtual Machine" in str(e):
+                    pass
+                else:
+                    raise e
             time.sleep(1)
             count = count + 1
-        assert count < 10, "Failed in Creating Restore Point"
+        assert_that(
+            count, "Restore point creation failed."
+        ).is_less_than(10)
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -53,7 +53,6 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
         vm_name = node_context.vm_name
         node_capability = node.capability.get_extended_runbook(AzureNodeSchema, AZURE)
         location = node_capability.location
-        # restore_point_collection = "rpc_" + vm_name + "_" + unique_name
         restore_point_collection = "rpc_" + unique_name
         assert environment.platform
         platform: AzurePlatform = environment.platform  # type: ignore
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -45,7 +45,7 @@ class BVTExtension(TestSuite):
         log.info(f"information {information}")
         restore_point_collection = "rpc_" + unique_name
         assert environment.platform
-        platform = environment.platform
+        platform: AzurePlatform = environment.platform  # type: ignore
         assert isinstance(platform, AzurePlatform)
         sub_id = platform.subscription_id
         # creating restore point collection
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -1,18 +1,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-import requests
-import http.server
-import json
-import threading
-import socketserver
 from assertpy import assert_that
+
 import uuid
-import os
+
 from datetime import datetime
+
 import time
 
-from azure.identity import DefaultAzureCredential
 from azure.mgmt.compute import ComputeManagementClient
 
 from lisa import (
@@ -28,6 +24,7 @@ from lisa.sut_orchestrator.azure.features import AzureExtension
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
 from lisa.sut_orchestrator.azure.common import get_compute_client
 
+
 @TestSuiteMetadata(
     area="vm_extension",
     category="functional",
@@ -35,15 +32,16 @@ from lisa.sut_orchestrator.azure.common import get_compute_client
     requirement=simple_requirement(unsupported_os=[]),
 )
 class BVTExtension(TestSuite):
- 
     @TestCaseMetadata(
         description="""
-        creates a restore point collection then a restore point which helps in validating the VMSnapshot extension.
+        creates a restore point collection then a restore point which helps in 
+        validating the VMSnapshot extension.
         """,
         priority=1,
         requirement=simple_requirement(supported_features=[AzureExtension]),
     )
-    def verify_vmsnapshot_extension(self, log: Logger, node: Node, environment: Environment) -> None :
+    def verify_vmsnapshot_extension(self, log: Logger, node: Node,
+                                   environment: Environment) -> None :
 
         unique_name = str(uuid.uuid4())
         information = environment.get_information()
@@ -51,22 +49,21 @@ class BVTExtension(TestSuite):
         location = information["location"]
         vm_name = node.name
         log.info(f"information {information}")
-        restore_point_collection = "rpc_"+unique_name
-        extension = node.features[AzureExtension]
+        restore_point_collection = "rpc_" + unique_name
         platform = environment.platform
         assert isinstance(platform, AzurePlatform)
-        sub_id = platform.subscription_id
-        
-        #creating restore point collection
+        sub_id = platform.subscription_id   
+        # creating restore point collection
         client = get_compute_client(environment.platform)
         response = client.restore_point_collections.create_or_update(
             resource_group_name=information["resource_group_name"],
             restore_point_collection_name=restore_point_collection,
             parameters={
-                "location": information["location"],
+                "location": location,
                 "properties": {
                     "source": {
-                        "id": "/subscriptions/"+sub_id+"/resourceGroups/"+resource_group_name+"/providers/Microsoft.Compute/virtualMachines/"+vm_name
+                        "id": "/subscriptions/" + sub_id + "/resourceGroups/" + 
+                        resource_group_name + "/providers/Microsoft.Compute/virtualMachines/" + vm_name
                     }
                 },
             },
@@ -75,20 +72,21 @@ class BVTExtension(TestSuite):
         log.info(f"response {response}")
         count = 0
 
-        while(count < 10):
+        while (count < 10):
             vm = client.virtual_machines.get(
                 resource_group_name,
                 vm_name
             )
             # check the state of the VM
-            if(vm.provisioning_state == "Succeeded"):
+            if (vm.provisioning_state == "Succeeded"):
                 try:
                     # create a restore point for the VM
                     response = client.restore_points.begin_create(
-                        resource_group_name=information["resource_group_name"],
-                        restore_point_collection_name=restore_point_collection,
-                        restore_point_name="rp_"+datetime.now().strftime("%Y-%m-%d-%H-%M-%S"),
-                        parameters={},
+                        resource_group_name = information["resource_group_name"],
+                        restore_point_collection_name = restore_point_collection,
+                        restore_point_name = "rp_" + 
+                        datetime.now().strftime("%Y-%m-%d-%H-%M-%S"),
+                        parameters = {},
                     )
                     response.wait(3600)
                     log.info("restore point created")
@@ -98,8 +96,7 @@ class BVTExtension(TestSuite):
                         pass
                     else:
                         log.info(f"error {e}")
-                        assert False, "Test failed: Unexpected error occurred"
+                        raise AssertionError("Test failed: Unexpected error occurred")
             time.sleep(1)
-            count = count + 1
-        
+            count = count + 1    
         assert count < 10, "Failed in Creating Restore Point"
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+            if vm.provisioning_state == "Succeeded":
+                try:
+                    # create a restore point for the VM
+                    response = client.restore_points.begin_create(
+                        resource_group_name=information["resource_group_name"],
+                        restore_point_collection_name=restore_point_collection,
+                        restore_point_name="rp_" + datetime.now().strftime("%Y-%m-%d-%H-%M-%S"),
+                        parameters={},
+                    )
+                    response.wait(3600)
```
squirrelsc: What's the type of response? In LISA, you can use wait_operation to wait with timeout in lisa.sut_orchestrator.azure.common. It handles different situations.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+        creates a restore point collection then a restore point which helps in
+        validating the VMSnapshot extension.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_features=[AzureExtension]),
+    )
+    def verify_vmsnapshot_extension(self, log: Logger, node: Node,
+                                    environment: Environment) -> None :
+        unique_name = str(uuid.uuid4())
+        information = environment.get_information()
```
squirrelsc: use get_node_context in azure.common to get the resource_group_name

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+        resource_group_name = information["resource_group_name"]
+        location = information["location"]
```
squirrelsc: refer to run_command.py to get location by below script. We will update LISA to get all of them from context.
`node.capability.get_extended_runbook(AzureNodeSchema, AZURE)`

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+        vm_name = node.name
```
squirrelsc: Get vm_name from NodeContext, it's more reliable.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+        log.info(f"information {information}")
```
squirrelsc: Check the actual log, before print more in log. The information is in log already. It doesn't need to print again.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+            parameters={
+                "location": location,
+                "properties": {
+                    "source": {
+                        "id": "/subscriptions/" + sub_id + "/resourceGroups/" + resource_group_name + "/providers/Microsoft.Compute/virtualMachines/" + vm_name
+                    }
+                },
+            },
+        )
+        log.info("rpc created")
```
squirrelsc: Current log should show the extension is called already.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+        # creating restore point collection
+        client = get_compute_client(environment.platform)
+        response = client.restore_point_collections.create_or_update(
+            resource_group_name=information["resource_group_name"],
+            restore_point_collection_name=restore_point_collection,
+            parameters={
+                "location": location,
+                "properties": {
+                    "source": {
+                        "id": "/subscriptions/" + sub_id + "/resourceGroups/" + resource_group_name + "/providers/Microsoft.Compute/virtualMachines/" + vm_name
```
squirrelsc: Use f string is easier to compose a string like, f"/subscriptions/{sub_id}/resourceGroups/{resource_group_name}/providers/Microsoft.Compute/virtualMachines/{vm_name}"

squirrelsc: If it's too long, it can be written like,
```python
f"/subscriptions/{sub_id}/resourceGroups/{resource_group_name}"
f"/providers/Microsoft.Compute/virtualMachines/{vm_name}"
```

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+        log.info(f"response {response}")
+        count = 0
+
+        while count < 10:
+            vm = client.virtual_machines.get(
+                resource_group_name,
+                vm_name
+            )
+            # check the state of the VM
+            if vm.provisioning_state == "Succeeded":
```
squirrelsc, line 72: It doesn't need to wait the VM is ready. By default, LISA enters test case, when the VM is ready.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+                try:
+                    # create a restore point for the VM
+                    response = client.restore_points.begin_create(
+                        resource_group_name=information["resource_group_name"],
+                        restore_point_collection_name=restore_point_collection,
+                        restore_point_name="rp_" + datetime.now().strftime("%Y-%m-%d-%H-%M-%S"),
+                        parameters={},
+                    )
+                    response.wait(3600)
+                    log.info("restore point created")
```
squirrelsc: Log the restore_point_name, it may be easier for troubleshooting in future.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
squirrelsc: It doesn't need to wait for provisioning. By default, when LISA enters a test case, the VM is ready.

squirrelsc: So it doesn't need to try 10 times.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+                        "id": "/subscriptions/" + sub_id + "/resourceGroups/" + resource_group_name + "/providers/Microsoft.Compute/virtualMachines/" + vm_name
+                    }
+                },
+            },
+        )
+        log.info("rpc created")
+        log.info(f"response {response}")
+        count = 0
+
+        while count < 10:
```
squirrelsc: How long does it create 10 snapshots? We should keep the P1 test case as fast as possible.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+                        resource_group_name=information["resource_group_name"],
+                        restore_point_collection_name=restore_point_collection,
+                        restore_point_name="rp_" + datetime.now().strftime("%Y-%m-%d-%H-%M-%S"),
+                        parameters={},
+                    )
+                    response.wait(3600)
+                    log.info("restore point created")
+                    break
+                except Exception as e:
+                    if "Changes were made to the Virtual Machine" in str(e):
```
squirrelsc: Why is it considered to be passed? Please add comments above to explain.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+                        pass
+                    else:
+                        log.info(f"error {e}")
```
squirrelsc: It can raise error directly here. It brings the original error for troubleshooting in error message. The logic is like,
```python
                except Exception as e:
                    if "Changes were made to the Virtual Machine" not in str(e):
                        raise e

```

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,92 @@
...
+                        raise AssertionError("Test failed: Unexpected error occurred")
+            time.sleep(1)
+            count = count + 1
+        assert count < 10, "Failed in Creating Restore Point"
```
squirrelsc: 1. use assertpy, it shows original values. Refer to examples in code about how to assert an integer.
2. It looks not necessary. If count is smaller than 10, either in the loop, or raised exception.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: Till this point extension isn't called, extension gets triggered when restore point is called

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: It will be creating only a single snapshot but have kept 10 retries.
In case a restorepoint/snapshot is successful it would take nearly 2 to 3 mins, but the timeout of this operation is 3hrs.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: Some times when a restore point is created we see this kind of error to retry after few minutes.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: This was used to fail the test case when the count >= 10 which means it has retried 10 times but still was not successful in creating a restore point. (This can happen when it do not get into the if block if vm.provisioning_state == "Succeeded": i.e VM_state is not ready and keeps waiting for it.)
As you have mentioned the VM would always be in ready state will remove this check.

arisettisanjana: Will continue to have the check as if say 10 times "Changes were made to the Virtual Machine" error occurred then it would just simply pass and when exceeded 10 retries it just comes out of the loop passing the test, so to avoid it and trying to fail the test

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: ok, will remove the check.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: done

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+    wait_operation,
+)
+from lisa.sut_orchestrator.azure.features import AzureExtension
+from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
+
+
+@TestSuiteMetadata(
+    area="vm_extension",
+    category="functional",
+    description="Test for VMSnapshot extension ",
```
iamashwani-who: nit: extra space at the end.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class BVTExtension(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        creates a restore point collection then a restore point which helps in
+        validating the VMSnapshot extension.
```
iamashwani-who: Can be more descriptive e.g. 
1. Create a restore point collection for the virtual machine.
2. Validate the virtual machine is in running state.
3. Create application consistent restore point on the restore point collection. Validate response for each restore point for validity. Attempt it a few items to rule out cases when VM is under changes.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
+
+
+@TestSuiteMetadata(
+    area="vm_extension",
+    category="functional",
+    description="Test for VMSnapshot extension ",
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class BVTExtension(TestSuite):
```
iamashwani-who: Is the class name expected to be fixed? can we name it VmSnapsotLinuxBVTExtension?

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+    def verify_vmsnapshot_extension(
+        self, log: Logger, node: Node, environment: Environment
+    ) -> None:
+        unique_name = str(uuid.uuid4())
+        node_context = get_node_context(node)
+        resource_group_name = node_context.resource_group_name
+        vm_name = node_context.vm_name
+        node_location = node.capability.get_extended_runbook(AzureNodeSchema, AZURE)
+        location = node_location.location
+        restore_point_collection = "rpc_" + unique_name
```
iamashwani-who, line 56: better to append VM name in the RPC name. rpc_vmName_unique_name

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+        assert environment.platform
+        platform: AzurePlatform = environment.platform  # type: ignore
+        assert isinstance(platform, AzurePlatform)
+        sub_id = platform.subscription_id
+        # creating restore point collection
+        client = get_compute_client(platform)
+        response = client.restore_point_collections.create_or_update(
```
iamashwani-who, line 63: is there option to create restore point collection via async operation like we are doing for restore point?

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+            try:
+                # create a restore point for the VM
+                restore_point = "rp_" + datetime.now().strftime("%Y-%m-%d-%H-%M-%S")
+                response = client.restore_points.begin_create(
+                    resource_group_name=resource_group_name,
+                    restore_point_collection_name=restore_point_collection,
+                    restore_point_name=restore_point,
+                    parameters={},
+                )
+                wait_operation(response)
```
iamashwani-who: check if async/await pattern can be utilized.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+                # This message is sometimes seen while rp creation
+                # so we will be retrying it after some time
+                if "Changes were made to the Virtual Machine" in str(e):
+                    pass
+                else:
+                    raise e
+            time.sleep(1)
+            count = count + 1
+        assert_that(
+            count, "Restore point creation failed."
```
iamashwani-who: can we format it into single line for better readability.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+                        + "/providers/Microsoft.Compute/virtualMachines/"
+                        + vm_name
+                    }
+                },
+            },
+        )
+        log.info("restore point collection created")
+        log.info(f"response {response}")
+        count = 0
+
```
iamashwani-who: Is there a way to validate if VM is in running state and if not start the VM? That will be the precondition for creating RPs.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
LiliDeng: change it into one single line will fix the CI check error

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+        requirement=simple_requirement(supported_features=[AzureExtension]),
+    )
+    def verify_vmsnapshot_extension(
+        self, log: Logger, node: Node, environment: Environment
+    ) -> None:
+        unique_name = str(uuid.uuid4())
+        node_context = get_node_context(node)
+        resource_group_name = node_context.resource_group_name
+        vm_name = node_context.vm_name
+        node_location = node.capability.get_extended_runbook(AzureNodeSchema, AZURE)
```
LiliDeng: node_capability =?

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+                        + sub_id
+                        + "/resourceGroups/"
+                        + resource_group_name
+                        + "/providers/Microsoft.Compute/virtualMachines/"
+                        + vm_name
+                    }
+                },
+            },
+        )
+        log.info("restore point collection created")
```
LiliDeng, line 77: use log.debug

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+        log.info(f"response {response}")
```
LiliDeng: do we need check the response to make sure the change is made successfully?

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+                    restore_point_name=restore_point,
+                    parameters={},
+                )
+                wait_operation(response)
+                log.info(f"restore point {restore_point} created")
+                break
+            except Exception as e:
+                # This message is sometimes seen while rp creation
+                # so we will be retrying it after some time
+                if "Changes were made to the Virtual Machine" in str(e):
```
LiliDeng, line 98: can you paste the whole exception in the comments?

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
squirrelsc: It's ok to log it on info level, if it's important information, and tell a story in log.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
squirrelsc: It's not fixed.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
squirrelsc: 1. use `for count in range(10):` is simpler.
2. If it takes 2-3 minutes normally, can we set timeout to 10 minutes? The timeout looks too long. In case it's failed in some reason, the test case takes long time. The default LISA test case timeout is 10 or 60 minutes, but it's changeable on test case metadata.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
squirrelsc: It doesn't support in LISA methods.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: Sure will set it to 10 minutes.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: As @squirrelsc mentioned by default, when LISA enters a test case, the VM is ready.
Removed the check.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: I did not find one.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: Tried it ashwani, by appending the vm_name it's not meeting the rpc naming criteria
'restorePointCollectionName' is invalid according to its validation rule: ^[^_\W][\w-._]{0,79}(?<![-.])$.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -0,0 +1,106 @@
...
+    def verify_vmsnapshot_extension(
+        self, log: Logger, node: Node, environment: Environment
+    ) -> None:
+        unique_name = str(uuid.uuid4())
+        node_context = get_node_context(node)
+        resource_group_name = node_context.resource_group_name
+        vm_name = node_context.vm_name
+        node_capability = node.capability.get_extended_runbook(AzureNodeSchema, AZURE)
+        location = node_capability.location
+        # restore_point_collection = "rpc_" + vm_name + "_" + unique_name
```
LiliDeng: remove this line

LiliDeng (approved):

PR #2799, opened 2023-06-07
Add tests for CSE

mayankdaruka-msft:
- Custom Script Extension tests
- RCv2 fixes (skip over BSD, arm64 images)
- Fix goal state agent error with multiconfig support

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -48,6 +48,7 @@ from lisa.features.security_profile import (
 from lisa.node import Node, RemoteNode
 from lisa.operating_system import CentOs, Redhat, Suse, Ubuntu
 from lisa.search_space import RequirementMethod
+from lisa.secret import add_secret
 from lisa.tools import Curl, Dmesg, Ls, Lspci, Modprobe, Rm
 from lisa.util import (
     LisaException,
@@ -1996,6 +1997,12 @@ class AzureExtension(AzureFeatureMixin, Feature):
             protected_settings=protected_settings,
             suppress_failures=suppress_failures,
         )
+
+        if protected_settings:
+            add_secret(
+                str(extension_parameters.as_dict()["protected_settings"]),
+                sub="***REDACTED***",
+            )
 
         self._log.debug(f"extension_parameters: {extension_parameters.as_dict()}")
 
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -55,6 +55,7 @@ class Waagent(Tool):
         else:
             self._command = "waagent"
         self._distro_version: Optional[str] = None
+        self._waagent_conf_path: Optional[str] = None
 
     def get_version(self) -> str:
         result = self.run("-version")
@@ -77,12 +78,7 @@ class Waagent(Tool):
         self.run("-deprovision --force", sudo=True, expected_exit_code=0)
 
     def get_configuration(self, force_run: bool = False) -> Dict[str, str]:
-        if isinstance(self.node.os, CoreOs):
-            waagent_conf_file = "/usr/share/oem/waagent.conf"
-        elif isinstance(self.node.os, BSD):
-            waagent_conf_file = "/usr/local/etc/waagent.conf"
-        else:
-            waagent_conf_file = "/etc/waagent.conf"
+        waagent_conf_file = self.get_waagent_conf_path()
 
         config = {}
         cfg = self.node.tools[Cat].run(waagent_conf_file, force_run=force_run).stdout
@@ -100,6 +96,13 @@ class Waagent(Tool):
     def get_resource_disk_mount_point(self) -> str:
         waagent_configuration = self.get_configuration()
         return waagent_configuration["ResourceDisk.MountPoint"]
+
+    def is_autoupdate_enabled(self) -> bool:
+        waagent_configuration = self.get_configuration()
+        if waagent_configuration.get("AutoUpdate.Enabled") == "n":
+            return False
+        else:
+            return True  # if set or not present, defaults to "y"
 
     def is_swap_enabled(self) -> bool:
         waagent_configuration = self.get_configuration()
@@ -122,6 +125,38 @@ class Waagent(Tool):
             return False
         else:
             raise LisaException(f"Unknown value for OS.EnableRDMA : {is_rdma_enabled}")
+
+    def get_waagent_conf_path(self) -> str:
+        if self._waagent_conf_path is not None:
+            return self._waagent_conf_path
+
+        for python_cmd in self._python_candidates:
+            python_exists, use_sudo = self.command_exists(command=python_cmd)
+            self._log.debug(
+                f"{python_cmd} exists: {python_exists}, use sudo: {use_sudo}"
+            )
+            if python_exists:
+                break
+
+        # Try to use waagent code to detect
+        result = self.node.execute(
+            f'{python_cmd} -c "from azurelinuxagent.common.osutil import get_osutil;'
+            'print(get_osutil().agent_conf_file_path)"',
+            sudo=use_sudo,
+        )
+        if result.exit_code == 0:
+            waagent_path = result.stdout
+        else:
+            if isinstance(self.node.os, CoreOs):
+                waagent_path = "/usr/share/oem/waagent.conf"
+            elif isinstance(self.node.os, BSD):
+                waagent_path = "/usr/local/etc/waagent.conf"
+            else:
+                waagent_path = "/etc/waagent.conf"
+
+        self._waagent_conf_path = waagent_path
+
+        return self._waagent_conf_path
 
     def get_distro_version(self) -> str:
         """
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -79,8 +79,8 @@ class Waagent(Tool):
         # self.run("-deprovision+user --force", sudo=True)
         self.run("-deprovision --force", sudo=True, expected_exit_code=0)
 
-    def get_configuration(self, force_run: bool = False) -> Dict[str, str]:
-        waagent_conf_file = self.get_waagent_conf_path()
+    def _get_configuration(self, force_run: bool = False) -> Dict[str, str]:
+        waagent_conf_file = self._get_waagent_conf_path()
 
         config = {}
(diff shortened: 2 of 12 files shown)


lisa/sut_orchestrator/azure/tools.py
```diff
@@ -123,6 +126,38 @@ def is_rdma_enabled(self) -> bool:
         else:
             raise LisaException(f"Unknown value for OS.EnableRDMA : {is_rdma_enabled}")
 
+    def get_waagent_conf_path(self) -> str:
+        if self._waagent_conf_path is not None:
+            return self._waagent_conf_path
+
+        for python_cmd in self._python_candidates:
```
squirrelsc, line 136: Can you move this part to be shared like `_get_python_cmd`? The path can be cached for next use.

microsoft/testsuites/vm_extensions/custom_script.py
```diff
@@ -0,0 +1,533 @@
...
+            node=node, environment=environment
+        )
+
+        settings = {"fileUris": [blob_url], "commandToExecute": f"sh {blob_name}"}
+
+        protected_settings = {
+            "storageAccountName": credentials["account_name"],
+            "storageAccountKey": credentials["account_key"],
+        }
+
```
squirrelsc: Make sure the key is not leaked in the log. If it is, add it to secret to mask it in log.

microsoft/testsuites/vm_extensions/common.py
```diff
@@ -0,0 +1,147 @@
+# Copyright (c) Microsoft Corporation.
```
squirrelsc: The file name may be confused, because we asked all extensions test cases to be here. Can you put all 3 files into a subfolder?

microsoft/testsuites/vm_extensions/common.py
```diff
@@ -0,0 +1,147 @@
...
+    get_storage_account_name,
+    get_storage_credential,
+)
+from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
+from lisa.sut_orchestrator.azure.tools import Waagent
+from lisa.util import SkippedException
+from semver import VersionInfo
+
+
+class CommandInfo(object):
```
squirrelsc: It looks like wrapping as a method is better to understand. It means given file name and expected_exit_code, it check and raise error.

microsoft/testsuites/vm_extensions/common.py
```diff
@@ -0,0 +1,147 @@
...
+            self.failure_message = (
+                f"File {file_name} was not created on the test machine"
+            )
+        else:
+            self.failure_message = (
+                f"File {file_name} downloaded on test machine though it should not have"
+            )
+
+
+def verify_waagent_version_supported(node: Node, environment: Environment) -> None:
```
squirrelsc: Avoid to start name with "verify_". The prefix is used for test case names. It won't be any problem, but confusing people.

microsoft/testsuites/vm_extensions/common.py
```diff
@@ -0,0 +1,147 @@
...
+    platform = environment.platform
+    assert isinstance(platform, AzurePlatform)
+
+    env_information = platform.get_environment_information(environment)
```
squirrelsc: I think you can just use WaAgent tool to get the version. It doesn't need to get it from environment information.

microsoft/testsuites/vm_extensions/common.py
```diff
@@ -0,0 +1,147 @@
...
+    if "wala_version" in env_information:
+        wala_version = env_information["wala_version"]
+        result = VersionInfo.parse(wala_version).compare("2.4.0")
+        if result < 0:
+            auto_update_enabled_from_waagent = node.tools[
+                Waagent
+            ].is_autoupdate_enabled()
+            if not auto_update_enabled_from_waagent:
+                raise SkippedException(
+                    f"Node with Goal State Agent version {wala_version}"
```
squirrelsc: Is it goal state version? I feel it's the original version, because auto updateis not enabled.

microsoft/testsuites/vm_extensions/custom_script.py
```diff
@@ -0,0 +1,533 @@
...
+        5. Both base64 script and command in public settings (should fail)
+        6. File uri and base64 script in public settings
+        7. File uri and gzip'ed base64 script in public settings
+        8. File uri and command in protected settings
+        9. Private file uri without sas token or credentials (should fail)
+        10. Private file uri with storage account credentials
+        11. Private sas file uri and command in public settings
+        12. File uri (pointing to python script) and command in public settings
+    """,
+    requirement=simple_requirement(unsupported_os=[BSD]),
```
squirrelsc: The requirement cannot be specified in both test suite and test case level. The test suite level is a default value, if it's not specified at test case level. The value [BSD] won't be kept, when test case level rewritten it. But if you have multiple test cases share the same requirements, it's useful to reduce repeat defining on each test case. If some test cases are different with most, you can define test case level for them.

microsoft/testsuites/vm_extensions/common.py
narrieta: Agree that the 3 files should be in a subdirectory. Some of the code in this file seems specific to RunCommand (e,g, the check for architecture, because RunCommand is not published on ARM64 yet).

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -139,6 +140,47 @@ def get_distro_version(self) -> str:
...
+
+    def get_waagent_conf_path(self) -> str:
+        if self._waagent_conf_path is not None:
+            return self._waagent_conf_path
+
+        python_cmd, use_sudo = self.get_python_cmd()
+
+        # Try to use waagent code to detect
+        result = self.node.execute(
+            f'{python_cmd} -c "from azurelinuxagent.common.osutil import get_osutil;'
```
narrieta, line 157: Not sure if this is an issue, just FYI. For the import to work, python_cmd needs to be exactly the same Python used to run the agent (otherwise the azurelinuxagent module won't be installed). Some distros, RedHat was one of them, started using a "platform python" that was used for the agent and some other utilities and included a limited set of modules. Then users could install other Python versions on their own. I believe at this point they dropped the "platform python" and use a full python, so just FYI.

mayankdaruka-msft: @squirrelsc Just wanted to make sure you've acknowledged this, since we're also using a similar approach to retrieve the waagent version.

narrieta: also just FYI: to workaround those issues, our automation gets the python that is running the agent from the /proc filesystem. That works if the Agent is currently running; otherwise we have a fallback logic that tries to discover the correct Python. The full logic is here: https://github.com/Azure/WALinuxAgent/blob/develop/tests_e2e/orchestrator/scripts/get-agent-python

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -123,13 +128,9 @@ def is_rdma_enabled(self) -> bool:
...
             raise LisaException(f"Unknown value for OS.EnableRDMA : {is_rdma_enabled}")
 
-    def get_distro_version(self) -> str:
-        """
-        This method is to get the same distro version string like WaAgent. It
-        tries best to handle different python version and locations.
-        """
-        if self._distro_version is not None:
-            return self._distro_version
+    def get_python_cmd(self) -> Tuple[str, bool]:
```
squirrelsc: For private method in a class, start the name with `_`.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1997,6 +1998,12 @@ def create_or_update(
             suppress_failures=suppress_failures,
         )
 
+        if protected_settings:
+            add_secret(
+                str(extension_parameters.as_dict()["protected_settings"]),
+                sub="***REDACTED***",
+            )
+
         self._log.debug(f"extension_parameters: {extension_parameters.as_dict()}")
```
squirrelsc, line 2007: The log line should not be useful now, if it's not printed out.

microsoft/testsuites/vm_extensions/runtime_extensions/custom_script.py
```diff
@@ -0,0 +1,518 @@
...
+            assert_exception=HttpResponseError,
+        )
+
+    @TestCaseMetadata(
+        description="""
+        Runs the Custom Script VM extension with a base64 script
+        and command with no file uris.
+        """,
+        priority=3,
+        requirement=simple_requirement(
```
squirrelsc: Should it be removed? It's no difference with test suite level.

microsoft/testsuites/vm_extensions/runtime_extensions/run_command.py
```diff
@@ -219,7 +164,9 @@ def verify_script_run_with_unnamed_parameter(self, log: Logger, node: Node) -> N
         """,
         priority=3,
         requirement=simple_requirement(
-            supported_features=[AzureExtension], supported_platform_type=[AZURE]
+            supported_features=[AzureExtension],
```
squirrelsc: Should it be removed?

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -101,6 +97,13 @@ def get_resource_disk_mount_point(self) -> str:
         waagent_configuration = self.get_configuration()
         return waagent_configuration["ResourceDisk.MountPoint"]
 
+    def is_autoupdate_enabled(self) -> bool:
+        waagent_configuration = self.get_configuration()
+        if waagent_configuration.get("AutoUpdate.Enabled") == "n":
+            return False
+        else:
+            return True  # if set or not present, defaults to "y"
```
squirrelsc: Good comment to help understanding the default! Don't put comment behide the code. Put it on top of code.

lisa/sut_orchestrator/azure/tools.py
mayankdaruka-msft: In that case, we would do the same for get_configuration and get_waagent_conf_path right?
Update: went ahead and made changes for those methods as well

lisa/sut_orchestrator/azure/features.py
mayankdaruka-msft: All the parameters are still being logged as before, except for the protected settings which contains sensitive data.

lisa/sut_orchestrator/azure/tools.py
squirrelsc: yes

squirrelsc:
@LiliDeng  LGTM

LiliDeng (approved):

PR #2800, opened 2023-06-08
fix verify_docker_java_app error on ubuntu build

paxue:
alpine image build need to specify “--network host”


lisa/tools/docker.py
```diff
@@ -15,9 +15,15 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
-    def build_image(self, image_name: str, dockerfile: str) -> None:
+    def build_image(
+        self, image_name: str, dockerfile: str, use_host_network: bool = False
```
squirrelsc: The method is used only once. and it's set to True. If there is no regression on all images, we can add the command parameter by default without an argument.

squirrelsc (approved):

PR #2801, opened 2023-06-08
DPDK: Send receive refactoring

mcgov:
Cleaning up some logic which has been tripping send/receive testing up.
Timeout refactor:
- Allow async execution to enable DPDK send/recv refactor

Refactor send/receive
- Ensure receiver starts first
- Simplify multiple queue test logic to allow use of shared logic
- Simplify queue and core selection to fix issues across variable threads/core/numa configurations.
- Rename variables to make logical core vs physical core model in DPDK EAL explicit.


mcgov:
yet another edit (black) on first commit, will re-open

PR #2804, opened 2023-06-13
code change for pci driver not enabled

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1825,6 +1826,40 @@ def check_blob_exist(
     return blob_client.exists()
 
 
+def is_mana_device(node: Node) -> bool:
```
squirrelsc: is_mana_device_discovered

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1825,6 +1826,40 @@ def check_blob_exist(
...
+
+
+def is_mlx_device(node: Node) -> bool:
+    pci_devices = node.tools[Lspci].get_devices_by_type(constants.DEVICE_TYPE_SRIOV)
+    if pci_devices and "Mellanox Technologies" in pci_devices[0].device_info:
+        return True
+    return False
+
+
+def is_mana_driver_enabled(node: Node) -> bool:
```
squirrelsc: It check both device and driver, how about call it `is_mana_enabled`?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1825,6 +1826,40 @@ def check_blob_exist(
...
+    if is_mana_device(node) and node.tools[KernelConfig].is_enabled(
+        "CONFIG_MICROSOFT_MANA"
+    ):
+        return True
+    return False
+
+
+def is_vf_exist(node: Node) -> bool:
+    if is_mlx_device(node):
```
squirrelsc: use `or` in same line is the same behavior.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1825,6 +1826,40 @@ def check_blob_exist(
...
+
+def is_mana_driver_enabled(node: Node) -> bool:
+    if is_mana_device(node) and node.tools[KernelConfig].is_enabled(
+        "CONFIG_MICROSOFT_MANA"
+    ):
+        return True
+    return False
+
+
+def is_vf_exist(node: Node) -> bool:
```
squirrelsc: `exists_vf`

microsoft/testsuites/network/common.py
```diff
@@ -8,6 +8,7 @@
 from lisa import Environment, Node, RemoteNode, constants
 from lisa.features import NetworkInterface
 from lisa.nic import NicInfo, Nics
+from lisa.sut_orchestrator.azure.common import is_mana_device, is_vf_exist
```
squirrelsc: The test case shouldn't refer to Azure package, unless it's a Azure only test case.

squirrelsc: With the new changes, we should consider below,
1. Wrap all logic in the network feature. so Azure specified things in Azure, and abstract from standard interface. Like `is_sriov_enabled` check if the sriov enabled e2e, `get_vf_driver` to get the mlx or mana.
2. To be clearer on different concepts, what's difference among them, which are azure only, which are general concepts, like SRIOV, VF, Synthetic...

microsoft/testsuites/network/common.py
```diff
@@ -73,9 +74,12 @@ def get_used_module(node: Node) -> str:
...
     )
-    # there will not be multiple Mellanox types in one VM
-    # get the used module using any one of sriov device
-    return lspci.get_used_module(devices_slots[0])
+    if is_vf_exist(node):
+        # there will not be multiple Mellanox types in one VM
+        # get the used module using any one of sriov device
+        return lspci.get_used_module(devices_slots[0])
+    else:
+        return "mana"
```
squirrelsc: If it needs to be hard coded, it should be in Azure logic.

mcgov:
I think some of this code should be in the generic NIC class instead of making this Azure-specific. I have some changes which could help, Lili and I have a meeting later so I'll share it and discuss further.

microsoft/testsuites/network/common.py
LiliDeng: NetworkInterface is for azure vm network interface, out of the VM, is it suitable to put the logic of VM there? I will try to put logic in nic level.

squirrelsc: Using the feature is one of the ways to wrap platform specific logic. But the nic can do similar things.

lisa/nic.py
```diff
@@ -132,6 +132,23 @@ class Nics(InitializableMixin):
 
     _file_not_exist = re.compile(r"No such file or directory", re.MULTILINE)
 
+    # ConnectX-3 uses mlx4_core
+    # mlx4_en and mlx4_ib depends on mlx4_core
+    #  need remove mlx4_en and mlx4_ib firstly
+    #  otherwise will see modules is in used issue
+    # ConnectX-4/ConnectX-5 uses mlx5_core
+    # mlx5_ib depends on mlx5_core, need remove mlx5_ib firstly
+    _reload_modules_dict: Dict[str, List[str]] = {
```
squirrelsc: Merge the two dict into one, so it's easier to manage data on same concept. You can create a class to hold multiple data on the same name.

lisa/nic.py
```diff
@@ -456,3 +483,81 @@ def _get_default_nic(self) -> None:
...
+
+    def is_mana_enabled(self) -> bool:
+        if self.is_mana_device_discovered() and self._node.tools[
+            KernelConfig
+        ].is_enabled("CONFIG_MICROSOFT_MANA"):
+            return True
+        return False
+
+    def exists_vf(self) -> bool:
+        if self.is_mlx_device() or self.is_mana_enabled():
```
squirrelsc: Should the mana check device as well? So it keeps consistent.

squirrelsc: If it's needed, make two methods, `is_vf_enabled`, and `exists_vf_device`.

lisa/nic.py
```diff
@@ -430,6 +447,16 @@ def _get_node_nic_info(self) -> None:
             nic_info = NicInfo(nic_name)
             self.append(nic_info)
 
+        if not self.exists_vf():
```
squirrelsc: The logic is not clear to me. It looks like to looking for pci devices, if vf doesn't exist. I don't know why it search vf is not here...

lisa/nic.py
```diff
@@ -456,3 +483,81 @@ def _get_default_nic(self) -> None:
...
+        )
+        if (
+            pci_devices
+            and "Device 00ba" in pci_devices[0].device_info
+            and pci_devices[0].vendor == "Microsoft Corporation"
+        ):
+            return True
+        return False
+
+    def is_mlx_device(self) -> bool:
```
squirrelsc: This name should be similar as above, because their logics are similar.

lisa/nic.py
```diff
@@ -456,3 +483,81 @@ def _get_default_nic(self) -> None:
...
+        ].is_enabled("CONFIG_MICROSOFT_MANA"):
+            return True
+        return False
+
+    def exists_vf(self) -> bool:
+        if self.is_mlx_device() or self.is_mana_enabled():
+            return True
+        return False
+
+    def expected_vf_count(self) -> int:
```
squirrelsc: We should be clear on different concepts. 
1. Here looks like should be "secondary", but I don't know why if no vf exists, it should return 1.
2. Why the lower nics represent the "vf"? When mana is functional inside of synthetic. Should it be a VF as well?

lisa/nic.py
LiliDeng: this method is to identify when network device is mana type, but mana driver is not enabled in distro. in this situation the synthetic nic behavior like sriov nic.

lisa/nic.py
LiliDeng: for mlx, if there are 6 sriov nics, then lspci will return 6 pci devices. so previously we count expected value by count length of total sriov nic (paired nic of synthetic nic)
for mana, it is always 1.

lisa/nic.py
LiliDeng: sorry, I didn't get your point.

squirrelsc: Put the module names and config values into a class, so the two dicts can be merged into one.
```python
@dataclass
class ModuleInformation:
    drivers: List[str]
    configs: List[str]

_device_module_map = {
  "mlx5_core": ModuleInformation(["mlx5_ib"], ["CONFIG_MLX5_CORE"],
}
```

lisa/nic.py
```diff
@@ -147,26 +175,29 @@ def __len__(self) -> int:
...
-
-    def get_upper_nics(self) -> List[str]:
-        return list(self.nics.keys())
+        return [x.name for x in self.nics.values() if not x.lower]
 
     def get_lower_nics(self) -> List[str]:
-        return [x.lower for x in self.nics.values() if x.lower]
+        return [x.name for x in self.nics.values() if x.upper]
+
+    def get_upper_nics_info(self) -> Dict[str, NicInfo]:
```
squirrelsc: It can return a list of NicInfo. And we can switch from `get_upper_nics` to this method step by step.

lisa/nic.py
```diff
@@ -284,6 +310,15 @@ def reload(self) -> None:
         self.nics.clear()
         self._initialize()
 
+    def check_sriov_enabled(self, sriov_enabled: bool) -> None:
```
squirrelsc: Is this method needed? As we discussed, the sriov is a technology. It shouldn't be used in tests.

lisa/nic.py
```diff
@@ -381,7 +414,7 @@ def _get_nic_numa_node(self, name: str) -> int:
             numa = 0
         return numa
 
-    def _get_node_nic_info(self) -> None:
+    def _get_nic_info(self) -> None:
```
squirrelsc: Call it `_load_nics`

lisa/nic.py
```diff
@@ -416,25 +456,44 @@ def _get_node_nic_info(self) -> None:
...
+        for nic_name in [x for x in self.nic_names if x not in self.nics.keys()]:
+            nic_info = NicInfo(name=nic_name)
             self.append(nic_info)
 
         assert_that(len(self)).described_as(
             "During Lisa nic info initialization, Nics class could not "
             f"find any nics attached to {self._node.name}."
         ).is_greater_than(0)
 
+    def _get_vf_info(self) -> None:
```
squirrelsc: call it `_load_pci_nics`, but why not merge with above `_get_nic_info`?

lisa/nic.py
LiliDeng: I'll check the method name, it is needed when there is VF nic paired with synthetic nic, VF nic need time to be initialized and paired with synthetic nic

lisa/nic.py
```diff
@@ -29,23 +29,34 @@ class NicInfo:
...
     def __init__(
         self,
-        upper: str,
+        name: str,
+        upper: str = "",
         lower: str = "",
         pci_slot: str = "",
+        is_upper: bool = True,
+        driver_sysfs_path: Optional[PurePosixPath] = None,
+        expected_no_vf: Optional[bool] = None,
```
squirrelsc: Consider both PF and VF. what's the purpose here? Is it to identify only VF, or it's to identify pci nics? We should use "pci", instead of sriov, vf, or pf.

squirrelsc: The name should be like `is_vf` or `is_synthetic`. Avoid to use `no`, `not` in variable name.

lisa/nic.py
```diff
@@ -456,3 +515,57 @@ def _get_default_nic(self) -> None:
...
+        modprobe = self._node.tools[Modprobe]
+        modprobe.load(module_name)
+
+    def get_packets(self, nic_name: str, name: str = "tx_packets") -> int:
+        cat = self._node.tools[Cat]
+        return int(
+            cat.read(f"/sys/class/net/{nic_name}/statistics/{name}", force_run=True)
+        )
+
+    def expected_no_vf(self) -> bool:
```
squirrelsc: Should it check only primary or all nics?

lisa/nic.py
squirrelsc: The expected value shouldn't be saved in nics or features. The expected values should be managed in test case level.

LiliDeng: I used it as flag to skip some verification with VF nics when there is no VF, not sure how to handle the logic well.

lisa/nic.py
```diff
@@ -284,6 +310,15 @@ def reload(self) -> None:
         self.nics.clear()
         self._initialize()
 
+    def check_sriov_enabled(self, sriov_enabled: bool) -> None:
+        if sriov_enabled and not self.get_lower_nics():
+            raise LisaException("SRIOV is enabled, but VF is not found.")
+        elif not sriov_enabled and self.get_lower_nics():
+            raise LisaException("SRIOV is disabled, but VF exists.")
+        else:
```
squirrelsc: Refine the logic as below, it's clearer to understand they are independenc logic.

```python
assert_that(sriov_enabled).described_as("AN enablement and pci device are inconsistent").equals(self.get_lower_nics())
```

lisa/nic.py
squirrelsc: We should avoid to use sriov, vf. It should use `pci`.

lisa/nic.py
```diff
@@ -284,6 +310,15 @@ def reload(self) -> None:
...
+            # the enabled flag is consistent with VF presents.
+            ...
+
     @retry(tries=15, delay=3, backoff=1.15)
     def wait_for_sriov_enabled(self) -> None:
```
squirrelsc: This method is similar to `check_sriov_enabled`. They can be merged on wait or check pci enabled or not.

lisa/nic.py
squirrelsc: I think it's misused in two purposes. One is to set expected value on nic in `wait_for_sriov_enabled`, it should be pass in from the method. The other is to check if it's synthetic or pci, so it should be renamed to `is_pci`.

lisa/nic.py
```diff
@@ -416,25 +456,44 @@ def _get_node_nic_info(self) -> None:
...
         ).is_greater_than(0)
 
+    def _get_vf_info(self) -> None:
+        lspci = self._node.tools[Lspci]
+        pci_devices = lspci.get_devices_by_type(
+            constants.DEVICE_TYPE_SRIOV, force_run=True
+        )
+        if len(pci_devices) > 0:
+            if (
+                "Device 00ba" in pci_devices[0].device_info
```
squirrelsc: Does it assume only one nic here? it maybe incorrect.

lisa/nic.py
```diff
@@ -456,3 +515,57 @@ def _get_default_nic(self) -> None:
...
+    def get_used_module(self) -> str:
+        lspci = self._node.tools[Lspci]
+        pci_devices = lspci.get_devices_by_type(
+            constants.DEVICE_TYPE_SRIOV, force_run=True
+        )
+        # there will not be multiple Mellanox types in one VM
+        # get the used module using any one of sriov device
+        used_module = lspci.get_used_module(pci_devices[0].slot)
+
+        if not used_module:
```
squirrelsc: The all logic should be proceeded in some initialization or load method. other places just check the loaded class info. It can make the logic clearer, and easy to maintain.

lisa/nic.py
squirrelsc: The whole logic should be like,
1. Load all nics from net
2. Load all nic pcis from lspci
3. Compose and match them to generate `nics` dict.

Whenever it needs to refresh the nic information, run above logic again.

lisa/nic.py
```diff
@@ -456,3 +515,57 @@ def _get_default_nic(self) -> None:
         ).is_true()
         self.default_nic: str = default_interface_name
         self.default_nic_route = dev_match.group()
+
+    def get_used_module(self) -> str:
+        lspci = self._node.tools[Lspci]
+        pci_devices = lspci.get_devices_by_type(
+            constants.DEVICE_TYPE_SRIOV, force_run=True
+        )
+        # there will not be multiple Mellanox types in one VM
```
squirrelsc: What if the first nic is synthetic only, and the second attaches a pci nic? We know some vm sizes allow limited pci devices. It may happen. So it's better to go through all information to find used modules.

lisa/nic.py
squirrelsc: The "mana" should be saved in `module_name` of NicInfo.

lisa/nic.py
```diff
@@ -456,3 +515,57 @@ def _get_default_nic(self) -> None:
...
+                raise LisaException("not find used module")
+        return used_module
+
+    def get_used_config(self) -> str:
+        return self._modules_config_dict[self.get_used_module()]
+
+    def is_module_reloadable(self) -> bool:
+        return self._node.tools[KernelConfig].is_built_as_module(self.get_used_config())
+
+    def remove_module(self) -> str:
```
squirrelsc: Should it be called `unload_module`, so be consistent with next `load_module`?

lisa/nic.py
LiliDeng: you mean in `wait_for_sriov_enabled`. I should not use expected_no_vf, instead, I need a parameter in the method to tell whether we need check the VF nic or not?

lisa/nic.py
```diff
@@ -285,37 +325,38 @@ def reload(self) -> None:
...
-            self.reload()
-        if not self.get_lower_nics():
-            assert_that(self.get_lower_nics()).described_as(
-                "Did not detect any upper/lower sriov paired nics!: "
-                f"upper: {self.get_upper_nics()} "
-                f"lower: {self.get_lower_nics()} "
-                f"unpaired: {self.get_unpaired_devices()}"
-                f"vfs: {','.join([str(pci) for pci in vfs])}"
-            ).is_not_empty()
+        pci_devices = lspci.get_devices_by_type(
```
squirrelsc: The method should be like below,
1. call `_load_nics`.
2. Check the information from the class only, not execute commands. use has_device, has_module should be enough.

lisa/nic.py
```diff
@@ -147,26 +176,41 @@ def __len__(self) -> int:
...
 
     def is_empty(self) -> bool:
         return len(self.nics) == 0
 
     def get_unpaired_devices(self) -> List[str]:
-        return [x.upper for x in self.nics.values() if not x.lower]
+        return [x.name for x in self.nics.values() if not x.pci_device_name]
+
+    def get_pci_nics(self) -> List[str]:
+        return [x.pci_device_name for x in self.nics.values() if x.pci_device_name]
```
squirrelsc: It should check has_module, instead of check the name.

lisa/nic.py
```diff
@@ -147,26 +176,41 @@ def __len__(self) -> int:
...
     def append(self, next_node: NicInfo) -> None:
-        self.nics[next_node.upper] = next_node
+        self.nics[next_node.name] = next_node
 
     def is_empty(self) -> bool:
         return len(self.nics) == 0
 
     def get_unpaired_devices(self) -> List[str]:
-        return [x.upper for x in self.nics.values() if not x.lower]
+        return [x.name for x in self.nics.values() if not x.pci_device_name]
```
squirrelsc: "unpaired" means it has no upper and no lower.

lisa/nic.py
```diff
@@ -200,39 +243,36 @@ def get_secondary_nic(self) -> NicInfo:
     def get_nic_by_index(self, index: int = -1) -> NicInfo:
         # get nic by index, default is -1 to give a non-primary nic
         # when there are more than one nic on the system
-        number_of_nics = len(self.get_upper_nics())
+        number_of_nics = len(self.get_nics())
```
squirrelsc: `get_nics` are used multiple times in this method, it should be saved in a variable.

lisa/nic.py
```diff
@@ -381,14 +417,15 @@ def _get_nic_numa_node(self, name: str) -> int:
...
+    def _load_nics(self) -> None:
         # Identify which nics are slaved to master devices.
         # This should be really simple with /usr/bin/ip but experience shows
         # the tool isn't super consistent across distros in this regard
 
-        # use sysfs to gather upper/lower nic pairings and pci slot info
+        # use sysfs to gather synthetic/pci nic pairings and pci slot info
         nic_info_fetch_cmd = "ls -la /sys/class/net/*/lower*/device"
         self._node.log.debug(f"Gathering NIC information on {self._node.name}.")
+        lspci = self._node.tools[Lspci]
```
squirrelsc, line 367: How about use `lspci.get_devices_by_type` to get all nic pcis firstly? The current logic may not be right to load mlx and other unknown nics.
And then query from net devices to map them.

lisa/nic.py
```diff
@@ -403,38 +440,91 @@ def _get_node_nic_info(self) -> None:
...
+                constants.DEVICE_TYPE_SRIOV, force_run=True
+            )
+            for pci_device in pci_devices:
+                for nic in self.get_unpaired_devices():
+                    self.nics[nic].pci_slot = pci_device.slot
+                    self.nics[nic].has_pci_module = False
+                    self.nics[nic].has_pci_device = True
+                break
+
+    def is_mana_device_discovered(self) -> bool:
```
squirrelsc: This should be a private method.

microsoft/testsuites/xdp/common.py
```diff
@@ -151,7 +151,7 @@ def _aggregate_count(
     patterns: List[Pattern[str]],
 ) -> int:
     ethtool = node.tools[Ethtool]
-    nic_names = [nic.upper, nic.lower]
+    nic_names = [nic.name, nic.pci_device_name]
```
squirrelsc: This is to get all nics' names. Is it included the lower nic name?

microsoft/testsuites/network/sriov.py
```diff
@@ -693,6 +693,8 @@ def verify_sriov_interrupts_change(self, environment: Environment) -> None:
         server_iperf3.run_as_server_async()
         client_interrupt_inspector = client_node.tools[InterruptInspector]
         for _, client_nic_info in vm_nics[client_node.name].items():
+            if client_nic_info.has_pci_device and not client_nic_info.has_pci_module:
```
squirrelsc: should it check pci_module enough?

lisa/tools/ip.py
```diff
@@ -20,8 +20,8 @@ class Ip(Tool):
     )
     __ip_addr_show_regex = re.compile(
```
squirrelsc, line 55: Add examples above.

LiliDeng: added

lisa/nic.py
LiliDeng: updated

lisa/nic.py
LiliDeng: can you elaborate this a little bit, thanks

lisa/nic.py
LiliDeng: updated

lisa/nic.py
LiliDeng: updated

lisa/nic.py
```diff
@@ -285,37 +325,33 @@ def reload(self) -> None:
...
+            ).is_not_zero()
+            if self._is_mana_device_discovered() and not self._node.tools[
+                KernelConfig
+            ].is_enabled("CONFIG_MICROSOFT_MANA"):
+                return
+            assert_that(pci_enabled).described_as(
+                "AN enablement and pci device are inconsistent"
+            ).is_equal_to(any(self.get_pci_nics()))
+        else:
+            assert_that(len(self.get_device_slots())).described_as(
```
squirrelsc: Change it like original code, assert the array `is_not_empty`. So, if it's not empty, it will print out values. It helps troubleshooting.

lisa/nic.py
```diff
@@ -285,37 +325,33 @@ def reload(self) -> None:
...
-            ).is_not_empty()
+    def check_pci_enabled(self, pci_enabled: bool) -> None:
+        self.reload()
+        if pci_enabled:
+            assert_that(len(self.get_device_slots())).described_as(
+                "Could not identify any pci devices on the test node."
+            ).is_not_zero()
+            if self._is_mana_device_discovered() and not self._node.tools[
+                KernelConfig
+            ].is_enabled("CONFIG_MICROSOFT_MANA"):
```
squirrelsc: Please double check with devs, if it's need to check the kernel configs here.

lisa/nic.py
squirrelsc: This method looks complicated than I imagined. Anyway, it can be refined next time.

microsoft/testsuites/xdp/common.py
LiliDeng: I didn't change its logic

microsoft/testsuites/network/sriov.py
LiliDeng: I suppose so

lisa/nic.py
LiliDeng: current logic is to try to find the pairs firstly, then remain ones will be initialized as upper nics, so there is no cases for it has lower but no upper.

lisa/nic.py
LiliDeng: updated

lisa/nic.py
LiliDeng: updated

lisa/nic.py
squirrelsc: how about below? It looks clearer.
`[x.name for x in self.nics.values() if not x.lower]`

lisa/nic.py
```diff
@@ -285,37 +266,33 @@ def reload(self) -> None:
...
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._node.log.debug("loading nic information...")
-        self.nic_names = self._get_nic_names()
-        self._get_node_nic_info()
-        self._get_default_nic()
-        self.load_interface_info()
+        self._nic_names = self._get_nic_names()
+        self._load_nics()
+        self.nics_list: List[str] = list(self.nics.keys())
```
squirrelsc: Is it the same as `_nic_names`? If so, should the `_get_nic_names` be removed? This class needs a method called `get_nic_names` to return all names to external, instead of a variable. Because it already provided get_nic, get_nic_by_index. It doesnt' make sense to expose the list directly.

lisa/nic.py
LiliDeng: there is no `lower` now, I use pci_device_name instead.

microsoft/testsuites/xdp/common.py
squirrelsc: It should be like `nic_names = [nic.name, nic.lower]`. It's easier to understand. My understanding is the `pci_device_name` should be presented in pci nic in normal case, instead of the upper synthetic nic.

microsoft/testsuites/network/sriov.py
squirrelsc: if so, will you change it?

lisa/nic.py
squirrelsc: The lower should be kept to support all scenarios. For example, how to differentiate single functional pci device (like bare metal), synthetic device with slave pci? What's difference between systhetic/pci pair?

lisa/nic.py
```diff
@@ -402,57 +348,128 @@ def _get_node_nic_info(self) -> None:
...
+                        self.append(
+                            NicInfo(
+                                name=nic_name,
+                                pci_device_name=pci_device_name,
+                                has_pci_device=True,
+                                has_pci_module=has_pci_module,
+                                pci_slot=pci_slot,
+                                pci_module_name=used_module,
+                            )
+                        )
```
LiliDeng, line 411: Currently self.nics will store synthetic nic as key, if there is a pci nic paired then fill the slot, pci module info etc for the same nic key.

Should I initialize the self.nics to store all the nics? then use different method or variables to get the upper/lower nics etc?

squirrelsc: Let's consider it from two ways.

1. Can it meet all current and future known scenarios? If yes, we keep the current design.
2. From the names, we should use lower to save like "eth1" instead of pci_device_name. It looks clearer than "pci_device_name". Because in no synthetic scenario, the name and pci_device_name should be the same. But it's hard to use.
3. How about add a method called like `get_pci_device_names`. It can returns all `name` without synthetic AND `lower` of synthetic nics.

microsoft/testsuites/xdp/common.py
LiliDeng: updated

microsoft/testsuites/network/sriov.py
LiliDeng: now I use pci_slot + lower to judge whether continue following testing

lisa/nic.py
LiliDeng: I change back to us get_nic_names method

microsoft/testsuites/network/sriov.py
squirrelsc: we defined `is_pci` when designing. Is it should be used here? Either use `is_pci` to track the upper, or track both upper and lower. The pci_slot is not a good part to track for this purpose. It's not clear enough. The `module_name` doesn't specify if it's a PCI device. So we need a variable or property to represent the information.

lisa/nic.py
LiliDeng: now I use is_pci here

lisa/nic.py
```diff
@@ -5,116 +5,76 @@
...
-            f"lower: {self.lower}\n"
+            f"name: {self.name}\n"
             f"pci_slot: {self.pci_slot}\n"
             f"ip_addr: {self.ip_addr}\n"
             f"mac_addr: {self.mac_addr}\n"
         )
 
+    @property
+    def is_pci(self) -> bool:
+        return len(self.lower) == 0 and len(self.pci_slot) > 0
```
squirrelsc: What does mean pci here? My understanding is the synthetic nic with VF is also a PCI. If not, how to present the lower pci device?

LiliDeng: I want to use it when there is no VF, but synthetic nic has AN capabilities here. the lower will have name if it is VF (lower nic)

microsoft/testsuites/network/common.py
```diff
@@ -161,24 +115,30 @@ def sriov_vf_connection_test(
...
 
-        if remove_module or turn_off_vf:
+        if remove_module or turn_off_pci_nic:
             source_nic = source_synthetic_nic
             dest_nic = dest_synthetic_nic
-        if turn_off_vf:
-            source_node.execute(f"ip link set dev {source_vf_nic} down", sudo=True)
-            dest_node.execute(f"ip link set dev {dest_vf_nic} down", sudo=True)
+
+        if turn_off_pci_nic:
```
squirrelsc: The original logic to be true for all pci_devices. With the condition `source_nic_info.is_pci_device and not source_nic_info.is_pci_module_enabled`, most pci devices skip the logic.

microsoft/testsuites/network/sriov.py
```diff
@@ -501,21 +502,24 @@ def verify_sriov_reload_modules(self, environment: Environment) -> None:
             ),
         ),
     )
-    def verify_sriov_ethtool_offload_setting(self, environment: Environment) -> None:
+    def verify_sriov_ethtool_offload_setting(  # noqa: C901
```
squirrelsc: Split some functions, instead of add the noqa.

microsoft/testsuites/network/sriov.py
```diff
@@ -693,12 +697,17 @@ def verify_sriov_interrupts_change(self, environment: Environment) -> None:
         server_iperf3.run_as_server_async()
         client_interrupt_inspector = client_node.tools[InterruptInspector]
         for _, client_nic_info in vm_nics[client_node.name].items():
+            if (
```
squirrelsc: Because the method is for sriov (PCI) only. Can it just run with `is_pci_module_enabled == true`? So it doesn't need a special logic here.

lisa/nic.py
```diff
@@ -182,13 +210,15 @@ def get_nic_driver(self, nic_name: str, store_driver: bool = True) -> str:
...
         ).is_not_equal_to("")
-        if store_driver:
-            nic.bound_driver = driver_name
+        nic.module_name = driver_name
         return driver_name
 
     def get_nic(self, nic_name: str) -> NicInfo:
         return self.nics[nic_name]
 
+    def get_nic_names(self) -> List[str]:
```
squirrelsc, line 219: 1. It's not used out of the nics class, so it should be a private method.
2. Does it need to include lower as well?

lisa/nic.py
```diff
@@ -403,56 +379,116 @@ def _get_node_nic_info(self) -> None:
...
             )
-        assert_that(dev_match.groups()).is_length(1)
-        default_interface_name = dev_match.group(1)
-        assert_that(default_interface_name in self.nic_names).described_as(
+            for pci_device in pci_devices:
+                for nic in self.get_unpaired_devices():
+                    self.nics[nic].pci_slot = pci_device.slot
+                break
+
+    def _is_mana_device_discovered(self) -> bool:
```
squirrelsc, line 440: It should check the name of module only.

microsoft/testsuites/network/common.py
LiliDeng: when no lower nic, we can't set down the upper nic

lisa/nic.py
LiliDeng: It is used in other places besides nics. 
No need to include lower nics.

lisa/nic.py
LiliDeng: if there is no mana driver, the module name is empty when check the device
we can just judge whether there is a module for the pci device to indicate is_pci_module_enabled, but we may miss the issue for mlx pci devices but the mlx driver is not enabled.

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/sriov.py

microsoft/testsuites/network/common.py
squirrelsc: the variable name is confusing, how about `turn_off_lower`? And the logic can check `if source_nic_info.lower` only.

lisa/nic.py
squirrelsc: How about the pci_slot? Does it have enough information?

lisa/nic.py
```diff
@@ -403,56 +379,116 @@ def _get_node_nic_info(self) -> None:
...
+        all_mana_devices = False
+        for pci_device in pci_devices:
+            if (
+                "Device 00ba" in pci_device.device_info
+                and pci_device.vendor == "Microsoft Corporation"
+            ):
+                all_mana_devices = True
+            else:
+                all_mana_devices = False
+                break
```
squirrelsc: It looks should break, when found any `True`, instead of `False`.

lisa/nic.py
LiliDeng: lspci is to get the pci info, there is no much more info from pci_slot itself.

microsoft/testsuites/network/common.py
LiliDeng: updated

lisa/nic.py
LiliDeng: updated

squirrelsc (approved):

PR #2805, opened 2023-06-13
Add Tests for VMAccess

mayankdaruka-msft:
- VMAccess Linux extension tests
- 2 new tests for RCv2 Linux (dependency on VMAccess) - tests runAsUser and runAsPassword functionality
- CSE bug fix for python command on RedHat


microsoft/testsuites/vm_extensions/runtime_extensions/common.py
```diff
@@ -15,9 +15,31 @@
...
     get_storage_account_name,
 )
+from lisa.sut_orchestrator.azure.features import AzureExtension
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
 from lisa.sut_orchestrator.azure.tools import Waagent
-from lisa.util import SkippedException
+from lisa.util import SkippedException, parse_version
+
+
+def create_and_verify_vmaccess_extension_run(
```
squirrelsc, line 24: If this method is for vmaccess only, it should be in vmaccess.py.

microsoft/testsuites/vm_extensions/runtime_extensions/run_command.py
```diff
@@ -294,3 +299,74 @@ def verify_script_run_with_timeout_failed(self, log: Logger, node: Node) -> None
...
+
+    @TestCaseMetadata(
+        description="""
+        Runs the Run Command v2 VM extension with a different valid user on the VM.
+        """,
+        priority=3,
+    )
+    def verify_script_run_with_valid_user(self, log: Logger, node: Node) -> None:
+        username = "vmaccessuser-valid"
+        password = "vmaccesspassword"
```
squirrelsc: Use a random password to keep the created VMs are secured.

microsoft/testsuites/vm_extensions/runtime_extensions/vmaccess.py
```diff
@@ -0,0 +1,283 @@
...
+        8. Username, OpenSSH format public key, and valid expiration date
+    """,
+    requirement=simple_requirement(
+        supported_features=[AzureExtension],
+        supported_platform_type=[AZURE],
+        unsupported_os=[BSD],
+    ),
+)
+class VMAccessTests(TestSuite):
+    _OPENSSH_KEY = (
```
squirrelsc: @LiliDeng please give the sample code to generate ssh keys, instead of hard code them here. It won't pass security checks.

microsoft/testsuites/vm_extensions/runtime_extensions/common.py
mayankdaruka-msft: It is also used by the new RCv2 extension tests I added, which is why it is in common.py

microsoft/testsuites/vm_extensions/runtime_extensions/vmaccess.py
mayankdaruka-msft: I will need to have a way to retrieve 3 different types of keys then - OpenSSH format, SSH2 format, and a certificate containing the key in pem format. I can look into this as well.

LiliDeng: try this
```
import paramiko

# Create an instance of the RSAKey class
key = paramiko.RSAKey.generate(2048)

public_key = key.get_base64()
```

mayankdaruka-msft: I don't think this will work for any keys except for OpenSSH keys. Also, the keys I added in vmaccess.py are only public keys, and I don't need private keys at all for this logic. So would that still be a security risk?

LiliDeng: I updated it to get the public key. Another way we can generate keys in the VM, and get the public key, then pass it into extension setting.

squirrelsc: Generating the keys from VM is a way. You can find a command line tool to do that, and add the tool into LISA.

microsoft/testsuites/vm_extensions/runtime_extensions/run_command.py
squirrelsc: There are a few passwords in code.

squirrelsc:
@LiliDeng  LGTM, please continue to review and sign off.

LiliDeng:
@mayankdaruka-msft  please check vmaccess cases on openlogic centos 7.6 7.6.20190808.

```
2023-06-15 01:55:31.030[5380][DEBUG] lisa.env[generated_32].node[0].cmd[5775] cmd: ['sh', '-c', 'ssh-keygen -f /tmp/22852d6d-16e1-40a1-b263-df85bde37b50 -N .'], cwd: None, shell: True, sudo: False, nohup: False, posix: True, remote: True
2023-06-15 01:55:31.061[4676][DEBUG] lisa.env[generated_32].node[0].cmd[5775].stdout Generating public/private rsa key pair.
2023-06-15 01:55:31.092[4676][DEBUG] lisa.env[generated_32].node[0].cmd[5775].stdout Saving key "/tmp/22852d6d-16e1-40a1-b263-df85bde37b50" failed: passphrase is too short (minimum five characters)
2023-06-15 01:55:31.108[5380][DEBUG] lisa.env[generated_32].node[0].cmd[5775] execution time: 0.071 sec, exit code: 1
```

jscalev (approved):

LiliDeng (approved):

PR #2810, opened 2023-06-17
ignored capability: fix issue, add log and doc

LiliDeng:
enhancement for #2702

### DIFF
--- a/docs/run_test/platform.rst
+++ b/docs/run_test/platform.rst
@@ -142,6 +142,9 @@ deployment.
       use_public_address: "<true or false>"
       requirement:
          ...
+         ignored_capability:
+            - SerialConsole
+            - Isolated_Resource
          azure:
             ...
             location: "<one or multiple locations, split by comma>"
@@ -170,6 +173,17 @@ deployment.
   public IP addresses.  False means to connect with the private IP addresses.
   If not provided, the connections will default to using the public IP
   addresses.
+* **ignored_capability**. Specify feature names which will be ignored in 
+  test requirement. You can find the feature name from its name method in source code.
+  For example, IsolatedResource feature's name defined in ``lisa/features/isolated_resource.py`` as below:
+
+   .. code:: python
+
+             @classmethod
+             def name(cls) -> str:
+               return FEATURE_NAME_ISOLATED_RESOURCE
+
+  Then, you can add ``isolated_resource`` to ``ignored_capability``.
 * **location**. Specify which locations is used to deploy VMs. It can be one or
   multiple locations. For example, westus3 or westus3,eastus. If multiple
   locations are specified, it means each environment deploys VMs in one of
--- a/lisa/runners/lisa_runner.py
+++ b/lisa/runners/lisa_runner.py
@@ -3,7 +3,7 @@
 
 import copy
 from functools import partial
-from typing import Any, Callable, Dict, List, Optional, cast
+from typing import Any, Callable, Dict, List, Optional, Set, cast
 
 from lisa import (
     ResourceAwaitableException,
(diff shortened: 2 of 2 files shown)


lisa/runners/lisa_runner.py
```diff
@@ -776,6 +776,13 @@ def _merge_test_requirements(
                             and hasattr(self, "platform")
                             and self.platform.runbook.ignored_capability
                         ):
+                            self._log.debug(
```
squirrelsc: It may produce many lines. Use a variable to log the lines only once.

squirrelsc: It doesn't need to log the information, because it can be found in other lines. We need to log line, when a feature is ignored. So it means the ignored feature is effective.

docs/run_test/platform.rst
```diff
@@ -169,6 +172,11 @@ deployment.
   public IP addresses.  False means to connect with the private IP addresses.
   If not provided, the connections will default to using the public IP
   addresses.
+* **ignored_capability**. Specify feature names which will be ignored in 
+  test requirement. It is useful to ignore some features which are not supported 
+  by the specified VM size. For example, if the VM size is not isolated type, 
+  but you still want to run the test with this size, then you can add 
+  `Isolated_Resource` into this list.
```
squirrelsc: The key thing is to explain where the feature name is from, so people can know how to get it.

docs/run_test/platform.rst
```diff
@@ -169,6 +172,15 @@ deployment.
   public IP addresses.  False means to connect with the private IP addresses.
   If not provided, the connections will default to using the public IP
   addresses.
+* **ignored_capability**. Specify feature names which will be ignored in 
+  test requirement. You can find the feature name in files under ``lisa/features``.
+  For example, ``lisa/features/isolated_resource.py`` defined feature name as below:
+
+   .. code:: python
+
+         FEATURE_NAME_ISOLATED_RESOURCE = "isolated_resource"
```
squirrelsc: It needs to tell users the source to look for value. The value is from the name method of a feature.

```
    @classmethod
    def name(cls) -> str:
        return FEATURE_NAME_ISOLATED_RESOURCE
```

lisa/runners/lisa_runner.py
```diff
@@ -776,17 +776,21 @@ def _merge_test_requirements(
...
+                                if str(feature).lower() in list(
                                     map(
                                         str.lower,
                                         self.platform.runbook.ignored_capability,
                                     )
+                                ):
+                                    node_requirement.features.items.remove(feature)
+                                    ignored_features.append(feature)
+                            if ignored_features:
+                                self._log.debug(
```
squirrelsc: Make sure it's logged only once in the whole log, because it's used to know if a feature string is effective or not. It can be like below, but you can keep to improve the string.

"the ignored feature 'aaa' is used.."

docs/run_test/platform.rst
LiliDeng: updated

lisa/runners/lisa_runner.py
LiliDeng: the ignore feature is based on test case requirement, it will be different results for different test cases, once for the whole log is insufficient.

I feel "the ignored feature 'aaa' is used.." makes users confused

squirrelsc: It depends on how the log helps. If it's to log which requirements are ignored for each test case, it's ok to print like this. But it's not needed and unclear for each test cases by below reasons. Please help refine the message to make it clear.
1. The log is unclear. It cannot differentiate which test case requirements are affected from this line. It doesn't print the test case name.
2. It's duplicate information for each test case. It ignores features by name, so its behavior is consistent for all requirements. Log it once is enough to prove the string is effective.

LiliDeng:
@squirrelsc could you please help review this PR, thanks! I will merge it in proper time.

lisa/runners/lisa_runner.py
```diff
@@ -706,6 +706,25 @@ def _skip_test_results(
 
             test_result.set_status(TestStatus.SKIPPED, reasons)
 
+    def _ignore_features(self, nodes: List[schema.NodeSpace]) -> List[str]:
```
squirrelsc: `_get_ignored_features`

lisa/runners/lisa_runner.py
```diff
@@ -706,6 +706,25 @@ def _skip_test_results(
...
+        ignored_features: List[str] = []
```
squirrelsc: Use a set here, instead of convert at the end.

lisa/runners/lisa_runner.py
LiliDeng: update

lisa/runners/lisa_runner.py
LiliDeng: update

squirrelsc (approved):

PR #2815, opened 2023-06-21
Add AzurePerformanceDiagnostics VM Extension test suite

ollegge:
**Description**

These changes add a Test Suite for the Azure Performance Diagnostics VM Extension. 

The initial test case `verify_azure_performance_diagnostics` does the following: 

-  Creates a storage account whose name and credentials are later passed in the public and protected settings used to run the VM Extension
- Install and run the Azure Performance Diagnostics VM Extension
- Delete the VM Extension
- Verify installation of VM Extension succeeded
- Verify a report was created and uploaded to the storage account
- Verify the VM Extension no longer exists on the VM after deletion


ollegge:

PR #2816, opened 2023-06-21
Test suite and test cases for Azure Performance Diagnostics VM Extension

ollegge:
**Description**

These changes add a Test Suite for the Azure Performance Diagnostics VM Extension.

The initial test case `verify_azure_performance_diagnostics` does the following:
-  Creates a storage account whose name and credentials are later passed in the public and protected settings used to run the VM Extension
- Install and run the Azure Performance Diagnostics VM Extension
- Delete the VM Extension
- Verify installation of VM Extension succeeded
- Verify a report was created and uploaded to the storage account
- Verify the VM Extension no longer exists on the VM after deletion

### DIFF
--- a/microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
+++ b/microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
@@ -15,6 +15,7 @@ from lisa import (
 from lisa.operating_system import Redhat
 from lisa.sut_orchestrator.azure.common import (
     check_or_create_storage_account,
+    delete_storage_account,
     get_or_create_storage_container,
     get_storage_credential,
 )
@@ -145,6 +146,15 @@ class AzurePerformanceDiagnostics(TestSuite):
         ).described_as(
             "Found the VM Extension still unexpectedly exists on the VM after deletion"
         ).is_false()
+
+        delete_storage_account(
+            credential=platform.credential,
+            subscription_id=platform.subscription_id,
+            cloud=platform.cloud,
+            account_name=storage_account_name,
+            resource_group_name=resource_group_name,
+            log=log,
+        )
 
     def _is_supported_linux_distro(self, node: Node) -> bool:
         supported_major_versions = {
--- a/microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
+++ b/microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
@@ -12,10 +12,19 @@ from lisa import (
     TestSuiteMetadata,
     simple_requirement,
 )
-from lisa.operating_system import Redhat
+from lisa.operating_system import (
+    AlmaLinux,
+    CBLMariner,
+    CentOs,
+    Debian,
+    Oracle,
+    Redhat,
+    Suse,
+    Ubuntu,
+)
 from lisa.sut_orchestrator.azure.common import (
     check_or_create_storage_account,
-    delete_storage_account,
+    get_node_context,
     get_or_create_storage_container,
     get_storage_credential,
 )
@@ -61,8 +70,9 @@ class AzurePerformanceDiagnostics(TestSuite):
         random_str = generate_random_chars(string.ascii_lowercase + string.digits, 10)
         storage_account_name = f"lisasc{random_str}"
         information = environment.get_information()
-        resource_group_name = information["resource_group_name"]
         location = information["location"]
+        node_context = get_node_context(node)
+        resource_group_name = node_context.resource_group_name
 
         check_or_create_storage_account(
             credential=platform.credential,
@@ -147,50 +157,26 @@ class AzurePerformanceDiagnostics(TestSuite):
             "Found the VM Extension still unexpectedly exists on the VM after deletion"
         ).is_false()
 
-        delete_storage_account(
-            credential=platform.credential,
-            subscription_id=platform.subscription_id,
-            cloud=platform.cloud,
-            account_name=storage_account_name,
-            resource_group_name=resource_group_name,
-            log=log,
-        )
-
     def _is_supported_linux_distro(self, node: Node) -> bool:
-        supported_major_versions = {
-            "redhat": [7, 8],
-            "red hat": [7, 8],
-            "centos": [6, 7],
-            "oracle": [6, 7],
-            "debian": [8, 9, 10, 11],
-            "ubuntu": [14, 16, 18, 20],
-            "suse": [12, 15],
-            "sles": [12, 15],
-            "almalinux": [8],
-            "mariner": [2],
-        }
-
-        version_list = supported_major_versions.get(
-            (node.os.information.vendor).lower()
-        )
-
-        if (
-            version_list is not None
-            and node.os.information.version.major in version_list
-            and not self._is_unsupported_version(node)
-        ):
-            return True
-        else:
-            return False
-
-    def _is_unsupported_version(self, node: Node) -> bool:
-        """
-        These are specific Linux distro versions that are
-        not supported by Azure Performance Diagnostics,
-        even though the major version is generally supported
-        """
         version = node.os.information.version
-        if isinstance(node.os, Redhat) and (version == "8.0.0" or version == "8.0.1"):
-            return True
-        else:
-            return False
+        major_version = version.major
+        if type(node.os) == CentOs:
+            return major_version in [6, 7]
+        elif type(node.os) == Oracle:
+            return major_version in [6, 7]  
+        elif type(node.os) == Debian:
+            return major_version in [8, 9, 10, 11]
+        elif type(node.os) == Ubuntu:
+            return major_version in [14, 16, 18, 20]
+        elif type(node.os) == AlmaLinux:
+            return major_version in [8]
+        elif type(node.os) == CBLMariner:
+            return major_version in [2]
+        elif isinstance(node.os, Suse):
+            return major_version in [12, 15]
+        elif type(node.os) == Redhat:
+            if major_version in [7, 8]:
+                if version != "8.0.0" and version != "8.0.1":
+                    return True
+
+        return False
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,192 @@
...
+        else:
+            return False
+
+    def _is_unsupported_version(self, node: Node) -> bool:
+        """
+        These are specific Linux distro versions that are
+        not supported by Azure Performance Diagnostics,
+        even though the major version is generally supported
+        """
+        unsupported_versions = {
```
squirrelsc: The distro type can be compared as `isinstance(node.os, RedHat)`. It can address derived OS as well.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,192 @@
...
+            "redhat": [{"major": 8, "minor": 0}],
+            "red hat": [{"major": 8, "minor": 0}],
+        }
+
+        unsupported_versions_list = unsupported_versions.get(
+            (node.os.information.vendor).lower()
+        )
+        if unsupported_versions_list:
+            for version in unsupported_versions_list:
```
squirrelsc: Try to compare by version string directly, like `node.os.information.version == "8.0.0"`

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,192 @@
...
+                f"{str(node.os.information.full_version)} is not supported."
+            )
+
+    @TestCaseMetadata(
+        description="""
+        Installs and runs the Azure Performance Diagnostics VM Extension.
+        Verifies a report was created and uploaded to the storage account.
+        Deletes the VM Extension.
+        """,
+        priority=3,
```
squirrelsc: How long is this case taken? You can have a P1 test case to be covered in most test scenarios.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,192 @@
...
+        report_exists = False
+        for _ in blob_iter:
+            report_exists = True
+
+        assert_that(report_exists).is_true()
+
+        # Delete VM Extension
+        extension.delete("AzurePerformanceDiagnosticsLinux")
+
+        assert_that(
```
squirrelsc, line 157: add `described_as` to provide an error message to help know what happened.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,186 @@
...
+        storage_account_name = f"lisasc{random_str}"
+        information = environment.get_information()
+        resource_group_name = information["resource_group_name"]
+        location = information["location"]
+
+        check_or_create_storage_account(
+            credential=platform.credential,
+            subscription_id=platform.subscription_id,
+            cloud=platform.cloud,
+            account_name=storage_account_name,
```
LiliDeng, line 84: should we delete the storage account after testing if it is new create one?

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,186 @@
...
+        extension.delete("AzurePerformanceDiagnosticsLinux")
+
+        assert_that(
+            extension.check_exist("AzurePerformanceDiagnosticsLinux")
+        ).described_as(
+            "Found the VM Extension still unexpectedly exists on the VM after deletion"
+        ).is_false()
+
+    def _is_supported_linux_distro(self, node: Node) -> bool:
+        supported_major_versions = {
```
squirrelsc, line 164: The versions should use LISA class to handle derived versions, like redhat can cover centos, and there are rocky, alma linux are not included as well.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,186 @@
...
+            subscription_id=platform.subscription_id,
+            cloud=platform.cloud,
+            account_name=storage_account_name,
+            resource_group_name=resource_group_name,
+        )
+
+        # Run VM Extension
+        extension = node.features[AzureExtension]
+        settings = {
+            "performanceScenario": "quick",
```
LiliDeng, line 101: any other scenario besides `quick` we can cover in new cases?

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,186 @@
...
+            credential=platform.credential,
+            subscription_id=platform.subscription_id,
+            cloud=platform.cloud,
+            account_name=storage_account_name,
+            container_name="azdiagextnresults",
+            resource_group_name=resource_group_name,
+        )
+
+        blob_iter = self._container_client.list_blobs(
+            name_starts_with="PerformanceDiagnostics"
```
LiliDeng, line 143: how we know this blob is generated for the test run if there are many blobs with name prefix is `PerformanceDiagnostics`

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
ollegge: Thanks for checking. Right now, the 'quick' scenario is the only one supported by our VM Extension for Linux

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
ollegge: Hi Chi,  What exactly do you think would be the best way to do this? I noticed that `isinstance(node.os, Redhat)` returns true for CentOs and others because of the inheritance structure. Since we support for different Redhat versions than we do CentOs, Alma Linux, etc., the logic doesn't work out the same.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
ollegge: The storage account is created specifically for this test, so there shouldn't be other blobs in it, especially now that I've added a delete storage account method at your suggestion.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,196 @@
...
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
+        platform = environment.platform
+        assert isinstance(platform, AzurePlatform)
+
+        # Create storage account and get credentials
+        random_str = generate_random_chars(string.ascii_lowercase + string.digits, 10)
+        storage_account_name = f"lisasc{random_str}"
+        information = environment.get_information()
+        resource_group_name = information["resource_group_name"]
```
squirrelsc: Don't get resource_group_name from information. it should be gotten from AzureContext of a node. Check other extension cases.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
squirrelsc: As the logic, it should be deleted with the resource group together.

LiliDeng: oh, yes,

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
squirrelsc: It depends on the test case scenarios. If it needs to support derivatives, use `isinstance(node.os, Redhat)`. If it needs to support the exact distro only, use `type(node.os) == Redhat`. But most tools support the derived distros together. In this case, the distro versions are considered as the same. Anyway, it doesn't need to check like "redhat/red hat" or "suse/sles".

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
ollegge: I'll remove the extra call to delete the storage account then. Thanks!

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,187 @@
...
+            extension.check_exist("AzurePerformanceDiagnosticsLinux")
+        ).described_as(
+            "Found the VM Extension still unexpectedly exists on the VM after deletion"
+        ).is_false()
+
+    def _is_supported_linux_distro(self, node: Node) -> bool:
+        is_supported = False
+        version = node.os.information.version
+        major_version = version.major
+        if type(node.os) == CentOs:
```
squirrelsc: If it's a big matrix, this part still can be data driven from an array like before.

ollegge: Oops. Good point. Thank you. Fixed.

microsoft/testsuites/vm_extensions/azureperformancediagnostics.py
```diff
@@ -0,0 +1,197 @@
...
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
+        platform = environment.platform
+        assert isinstance(platform, AzurePlatform)
+
+        # Create storage account and get credentials
+        random_str = generate_random_chars(string.ascii_lowercase + string.digits, 10)
+        storage_account_name = f"lisasc{random_str}"
+        information = environment.get_information()
+        location = information["location"]
```
squirrelsc: The location is in node_context as well.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2817, opened 2023-06-22
Network testcase fix for FreeBSD

dsrivastavv:
Note: `verify_device_gro_lro_settings_change` is skipped since there is no gro support in freebsd

### DIFF
--- a/lisa/tools/ethtool.py
+++ b/lisa/tools/ethtool.py
@@ -1,12 +1,18 @@
 import re
 from dataclasses import dataclass
-from typing import Any, Dict, List, Optional, Set, cast
+from typing import Any, Dict, List, Optional, Set, Type, cast
 
 from lisa.executable import Tool
 from lisa.operating_system import Posix
-from lisa.util import LisaException, UnsupportedOperationException, find_groups_in_lines
+from lisa.util import (
+    LisaException,
+    UnsupportedOperationException,
+    find_group_in_lines,
+    find_groups_in_lines,
+)
 
 from .find import Find
+from .ip import Ip
 from .lscpu import Lscpu
 
 # Few ethtool device settings follow similar pattern like -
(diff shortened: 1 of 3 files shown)


lisa/tools/ethtool.py
```diff
@@ -131,25 +137,32 @@ class DeviceFeatures:
...
-        self._parse_feature_info(interface, device_feature_raw)
-
-    def _parse_feature_info(self, interface: str, raw_str: str) -> None:
-        matched_features_info = self._feature_info_pattern.search(raw_str)
+    def __init__(self, interface: str, enabled_features: List[str]) -> None:
+        self.device_name = interface
+        self.enabled_features = enabled_features
+
+    @classmethod
+    def from_raw_feature(
```
squirrelsc: It looks work for Linux only. So it should be moved into Linux tool.

lisa/tools/ifconfig.py
```diff
@@ -0,0 +1,37 @@
...
+class Ifconfig(Tool):
+    @property
+    def command(self) -> str:
+        return "ifconfig"
+
+    @classmethod
+    def _freebsd_tool(cls) -> Optional[Type[Tool]]:
+        return IfconfigFreebsd
+
+    def get_interface_list(self) -> list[str]:
```
squirrelsc: What should be used for Linux? BSD should support the interface, which Linux uses.

lisa/tools/ethtool.py
dsrivastavv: Yes. It is specific to Linux. I will move this logic into ethtool class for linux

lisa/tools/ifconfig.py
dsrivastavv: Ifconfig is deprecated in Linux. `ip` tool is the recommended method for Linux but this is not present in BSD.

squirrelsc: As our design principle, all tools should be along with Linux tools. So the FreeBSD should use ifconfig commands to simulate ip in Linux. So it needs a BSD version of ip tool.

dsrivastavv: Sure! I can move this logic to the ip tool

lisa/tools/ip.py
```diff
@@ -230,3 +235,21 @@ def setup_tap(self, name: str, bridge: str) -> None:
 
         # start interface
         self.up(name)
+
+    def get_interface_list(self) -> list[str]:
+        raise NotImplementedError()
+
+
+class IpFreebsd(Ip):
```
squirrelsc, line 243: Because we don't provide the `ip` tool for Linux. It should be in ethtool, or directly execute the command as "ip -l"

dsrivastavv: Sorry didn't this comment. `ip` is the default  tools in Linux to get network interfaces information and we do provide `ip` tool for Linux. We can move this into ethtool but I think this is a better place since it will allow reuse of code in future.

squirrelsc: Oh, I see. I thought ip is deprecating.

squirrelsc (approved):

PR #2820, opened 2023-06-26
Fix broken links in transformers.rst

adityagesh:


adityagesh:
https://github.com/microsoft/lisa/blob/aditya/fix_transformer_doc/docs/run_test/transformers.rst
https://github.com/microsoft/lisa/blob/aditya/fix_transformer_doc/docs/run_test/runbook.rst#type

docs/run_test/runbook.rst
```diff
@@ -86,7 +86,7 @@ In simple terms:
    you from lengthy command-line commands and makes it easy to adjust
    configurations.`
 
-See :ref:`write_test/concepts:runbook` for further knowledge.
+See `Documentation for Runbook <https://github.com/microsoft/lisa/blob/aditya/fix_transformer_doc/docs/write_test/concepts.rst#runbook>`__ for further knowledge.
```
squirrelsc, line 89: It should be a relative link to readdocs, instead of back to github page.

adityagesh: The existing links were broken
image

Let me try to fix the relative linking.

PR #2828, opened 2023-06-29
ch_tests_tool: Increase timeout for Reboot

pupacha:
These tests are usually run on server grade machines and it might take longer time for the machine to reboot. Hence, increase the reboot timeout

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -222,7 +222,7 @@ class CloudHypervisorTests(Tool):
         if username not in res.stdout:  # if current user is not in docker group
             self.node.execute(f"usermod -a -G docker {username}", sudo=True)
             # reboot for group membership change to take effect
-            self.node.reboot()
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -222,7 +222,7 @@ def _install(self) -> bool:
         if username not in res.stdout:  # if current user is not in docker group
             self.node.execute(f"usermod -a -G docker {username}", sudo=True)
             # reboot for group membership change to take effect
-            self.node.reboot()
+            self.node.reboot(time_out=900)
```
LiliDeng, line 225: is it expected value? does it indicate any regression?

pupacha: This is not a regression in current systems. But going forward we are planning to run these tests on baremetal servers. And the reboot time is more than the default timeout defined in lisa (i.e., 300). Hence, this is a precautionary change. It doesn't effect any of the current running scenarios.

LiliDeng (approved):

PR #2831, opened 2023-07-03
Move the part of getting marketplace image locations into a function

lubaihua33:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2477,9 +2477,7 @@ def _resolve_marketplace_image_version(
                     node_runbook.location, node_runbook.marketplace
                 )
 
-    def _add_image_features(self, node_space: schema.NodeSpace) -> None:
-        # Load image information, and add to requirements.
-
+    def _get_marketplace_image_locations(self) -> List[str]:
```
squirrelsc: `_find_marketplace_image_location` is better to describe the logic.

squirrelsc (approved):

PR #2832, opened 2023-07-03
Add is_accelerated_network_supported in VhdSchema

lubaihua33:
Some VHDs don't support accelerated network. Add is_accelerated_network_supported in VhdSchema, so users can define the accelerated network capability for the VHD.

Add _check_image_capability to check if the image can support AN. If not, and test cases require an SRIOV data path, then raise NotMeetRequirementException.

If the VHD doesn't support AN, then set the network interface data path as Synthetic, Lisa will deploy VM without AN enabled.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -6,7 +6,7 @@ import re
 import sys
 from dataclasses import InitVar, dataclass, field
 from datetime import datetime, timedelta, timezone
-from functools import lru_cache
+from functools import lru_cache, partial
 from pathlib import Path
 from threading import Lock
 from time import sleep
@@ -68,7 +68,7 @@ from msrestazure.azure_cloud import Cloud  # type: ignore
 from PIL import Image, UnidentifiedImageError
 from retry import retry
 
-from lisa import schema
+from lisa import schema, search_space
 from lisa.environment import Environment, load_environments
 from lisa.feature import Features
 from lisa.node import Node, RemoteNode, local
@@ -135,6 +135,15 @@ global_credential_access_lock = Lock()
 # add a lock to prevent it happens.
 _global_storage_account_check_create_lock = Lock()
 
+MARKETPLACE_IMAGE_KEYS = ["publisher", "offer", "sku", "version"]
+SIG_IMAGE_KEYS = [
+    "subscription_id",
+    "resource_group_name",
+    "image_gallery",
+    "image_definition",
+    "image_version",
+]
+
 
 @dataclass
 class EnvironmentContext:
@@ -165,9 +174,30 @@ class AzureVmPurchasePlanSchema:
     publisher: str
 
 
+@dataclass_json
+@dataclass
+class AzureImageSchema:
+    network_data_path: Optional[
+        Union[search_space.SetSpace[schema.NetworkDataPath], schema.NetworkDataPath]
+    ] = field(  # type: ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            items=[
+                schema.NetworkDataPath.Synthetic,
+                schema.NetworkDataPath.Sriov,
+            ],
+        ),
+        metadata=field_metadata(
+            decoder=partial(
+                search_space.decode_set_space_by_type, base_type=schema.NetworkDataPath
+            )
+        ),
+    )
+
+
 @dataclass_json()
 @dataclass
-class AzureVmMarketplaceSchema:
+class AzureVmMarketplaceSchema(AzureImageSchema):
     publisher: str = "Canonical"
     offer: str = "0001-com-ubuntu-server-jammy"
     sku: str = "22_04-lts"
@@ -179,7 +209,7 @@ class AzureVmMarketplaceSchema:
 
 @dataclass_json()
 @dataclass
-class SharedImageGallerySchema:
+class SharedImageGallerySchema(AzureImageSchema):
     subscription_id: str = ""
     resource_group_name: Optional[str] = None
     image_gallery: str = ""
@@ -197,7 +227,7 @@ class SharedImageGallerySchema:
 
 @dataclass_json()
 @dataclass
-class VhdSchema:
+class VhdSchema(AzureImageSchema):
     vhd_path: str = ""
     vmgs_path: Optional[str] = None
 
@@ -284,7 +314,10 @@ class AzureNodeSchema:
                 # The lower() normalizes the image names,
                 #  it has no impact on deployment.
                 self.marketplace_raw = dict(
-                    (k, v.lower()) for k, v in self.marketplace_raw.items()
+                    (k, v.lower())
+                    if isinstance(v, str) and k in MARKETPLACE_IMAGE_KEYS
+                    else (k, v)
+                    for k, v in self.marketplace_raw.items()
                 )
                 marketplace = schema.load_by_type(
                     AzureVmMarketplaceSchema, self.marketplace_raw
@@ -309,7 +342,12 @@ class AzureNodeSchema:
                     )
 
                     if len(marketplace_strings) == 4:
-                        marketplace = AzureVmMarketplaceSchema(*marketplace_strings)
+                        marketplace = AzureVmMarketplaceSchema(
+                            publisher=marketplace_strings[0],
+                            offer=marketplace_strings[1],
+                            sku=marketplace_strings[2],
+                            version=marketplace_strings[3],
+                        )
                         # marketplace_raw is used
                         self.marketplace_raw = marketplace.to_dict()  # type: ignore
                     else:
@@ -345,7 +383,8 @@ class AzureNodeSchema:
             # The lower() normalizes the image names,
             #  it has no impact on deployment.
             self.shared_gallery_raw = dict(
-                (k, v.lower()) for k, v in self.shared_gallery_raw.items()
+                (k, v.lower()) if isinstance(v, str) and k in SIG_IMAGE_KEYS else (k, v)
+                for k, v in self.shared_gallery_raw.items()
             )
             shared_gallery = schema.load_by_type(
                 SharedImageGallerySchema, self.shared_gallery_raw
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -186,6 +186,7 @@ def __hash__(self) -> int:
 class VhdSchema:
     vhd_path: str = ""
     vmgs_path: Optional[str] = None
+    is_accelerated_network_supported: Optional[bool] = None
```
squirrelsc: Add a base class for marketplace, sig, and vhd. So this field can be used on all of them.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -488,6 +488,7 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
         # covert to azure node space, so the azure extensions can be loaded.
         for req in nodes_requirement:
             self._load_image_features(req)
+            self._check_image_capability(req)
```
squirrelsc: It should be called in `_check_image_capability`, because image features are updated there. Rename `_load_image_features` like `_set_image_features`. So it's not only to load new features.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2547,6 +2548,26 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
 
+    def _check_image_capability(self, node_space: schema.NodeSpace) -> None:
+        is_an_supported: Optional[bool] = None
+        azure_runbook = node_space.get_extended_runbook(AzureNodeSchema, AZURE)
+        if azure_runbook.vhd:
+            is_an_supported = azure_runbook.vhd.is_accelerated_network_supported
+
+        # Network interface feature is handled by node_space directly.
+        # Don't need to add in featuers
+        if isinstance(is_an_supported, bool) and not is_an_supported:
```
squirrelsc: It doesn't need to check twice. Just check like, `if azure_runbook.vhd.is_accelerated_network_supported == False`. Add comments to explain the logic.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2547,6 +2548,26 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
+            data_path = schema.NetworkDataPath.Synthetic
+            # Check if requirement is supported in image capability
+            if node_space.network_interface:
```
squirrelsc, line 2598: I think it doesn't need to call `check_xxx`, just set it to Synthetic.

lisa/sut_orchestrator/azure/common.py
squirrelsc: From below logic, it doesn't need to be optional. So, set default value to `True`.

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: Some test cases require SRIOV data_path, such as verify_sriov_basic, then node_space.network_interface.data_path is SRIOV. For these cases, they should raise NotMeetRequirementException if VHD doesn't support AN before deploying the VM.

If just set node_space.network_interface.data_path to Synthetic, the logic can still find a matched VM size and deploy the VM. Until running the case, it just skips the case.

So it still needs to cal check_xxx, if not match, then raise NotMeetRequirementException before copying VHD and deploying.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/common.py

lubaihua33: > Add a base class for marketplace, sig, and vhd. So this field can be used on all of them.

For marketplace image and SIG, users don't need to set this field. The AN support information can be got from image_info.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/common.py
squirrelsc: There are some cases, the image info is not right, or lack of image info. So for SIG and marketplace images, it loads image info firstly. If there are user set information, then override image info.

lubaihua33: Have added.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -352,12 +362,12 @@ def shared_gallery(self) -> Optional[SharedImageGallerySchema]:
                 r"[/]+", self.shared_gallery_raw.strip().lower()
             )
             if len(shared_gallery_strings) == 5:
-                shared_gallery = SharedImageGallerySchema(*shared_gallery_strings)
+                shared_gallery = SharedImageGallerySchema(True, *shared_gallery_strings)
```
squirrelsc: It's an optional field, doesn't need to provide value. Create a method like `create` to initialize a schema, so it would be easier.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Does the image work with AN VM sizes, but AN is disabled? Or it needs the AN unavailable vm sizes?

lubaihua33: Yes, the image works with AN VM sizes but AN is disabled.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -151,9 +151,15 @@ class AzureVmPurchasePlanSchema:
     publisher: str
 
 
+@dataclass_json
+@dataclass
+class AzureImageSchema:
+    is_accelerated_network_supported: bool = True
```
squirrelsc: Let's give it more generic name, and handle more scenario. Make it the same type as `NetworkInterfaceOptionSettings.data_path`

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2547,12 +2547,28 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
 
-    def _load_image_features(self, node_space: schema.NodeSpace) -> None:
+    def _check_image_capability(self, node_space: schema.NodeSpace) -> None:
+        azure_runbook = node_space.get_extended_runbook(AzureNodeSchema, AZURE)
+
+        # If VHD doesn't enable AN, check if the requirement is supported
+        if azure_runbook.vhd and not azure_runbook.vhd.is_accelerated_network_supported:
+            synthetic_data_path = schema.NetworkDataPath.Synthetic
+            if node_space.network_interface:
+                matched_result = search_space.check_setspace(
```
squirrelsc: Just call `intersect`, and catch and rewrite the type of the Exception.

lisa/sut_orchestrator/azure/common.py
lubaihua33: If so, the runbook should be defined as below if I understand right. Seems like this makes the runbook a little complex to use.
  requirement:
    azure:
       vhd: 
          network_data_path: 
            - Synthetic
            - Sriov

For user, they might not know directly if their image's network is synthetic or sriov. They can know directly if their images enable AN or support AN. 

Take marketplace image and VHD as examples:
If it is a marketplace image, we can get "**IsAcceleratedNetworkSupported**" field from image features. If it is a VHD format, in the Azure Certification scenario, the parameter passed from the upper layer is "**AcceleratedNetworking** **IsEnabled**".  So keeping the current type and name can connect with these scenarios and better understand.

squirrelsc: 1. It has the default value to provide both. Only if the data_path is not fully supported, this field should be set.
2. In future, there may be a scenario, which support only AN. Above setting can fit more scenarios.

lisa/sut_orchestrator/azure/common.py
lubaihua33: Do you mean to add a "create" function in the parent class?
I have modified it as the parameters are set to value separately. Does that look fine?

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: Has called intersect_setspace_by_priority.

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: About the comment "and catch and rewrite the type of the Exception", do you mean to write a new exception?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2547,12 +2547,24 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
 
-    def _load_image_features(self, node_space: schema.NodeSpace) -> None:
+    def _check_image_capability(self, node_space: schema.NodeSpace) -> None:
+        azure_runbook = node_space.get_extended_runbook(AzureNodeSchema, AZURE)
+        if azure_runbook.vhd:
+            if node_space.network_interface:
+                data_path = search_space.intersect_setspace_by_priority(  # type: ignore
+                    node_space.network_interface.data_path,
+                    azure_runbook.vhd.network_data_path,
+                    [],
```
squirrelsc, line 2602: Use the priority `_network_data_path_priority`, which defined in schema.py. Make it public.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: It doesn't need a new exception, I just forget the behavior of intersect.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -330,7 +357,8 @@ def shared_gallery(self) -> Optional[SharedImageGallerySchema]:
             # The lower() normalizes the image names,
             #  it has no impact on deployment.
             self.shared_gallery_raw = dict(
-                (k, v.lower()) for k, v in self.shared_gallery_raw.items()
+                (k, v.lower()) if k != "network_data_path" else (k, v)
```
squirrelsc: If it's needed, it should give names of shared_gallery_image, instead of `network_data_path`. We may add more fields in future, so it shouldn't a place, which we have to update every time.

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: In the function intersect_setspace_by_priority, the parameter priority_list isn't used. In this function, it has the below comment:
intersect doesn't need to take care about priority.

squirrelsc:
@LiliDeng  LGTM

lisa/sut_orchestrator/azure/common.py
lubaihua33: Done.
Added "image_name_keys" in AzureVmMarketplaceSchema and SharedImageGallerySchema.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -151,26 +151,55 @@ class AzureVmPurchasePlanSchema:
...
 @dataclass_json()
 @dataclass
-class SharedImageGallerySchema:
+class SharedImageGallerySchema(AzureImageSchema):
     subscription_id: str = ""
     resource_group_name: Optional[str] = None
     image_gallery: str = ""
     image_definition: str = ""
     image_version: str = ""
+    image_name_keys = [
```
squirrelsc: move it out of the class, because it's a data class, all class fields will be regarded as object fields.

lubaihua33: Yeah, I didn't aware of that. Have modified.

LiliDeng (approved):

PR #2834, opened 2023-07-03
ip addr show pattern change

LiliDeng:


lisa/tools/ip.py
```diff
@@ -55,9 +58,9 @@ class Ip(Tool):
     """
     __ip_addr_show_regex = re.compile(
         (
-            r"\d+: (?P<name>\w+): \<.+\> .+\n\s+link\/(?:ether|infiniband) "
-            r"(?P<mac>[0-9a-z:]+) .+\n(?:(?:.+\n\s+|.*)altname \w+)?"
-            r"(?:\s+inet (?P<ip_addr>[\d.]+)\/.*\n)?"
+            r"\d+: (?P<name>\w+): \<.*\>.*link\/(?:ether|infiniband|loopback) "
```
squirrelsc: Does it need to update examples?

lisa/tools/ip.py
```diff
@@ -224,16 +231,15 @@ def get_info(self, nic_name: Optional[str] = None) -> List[IpInfo]:
...
             ),
         )
-        entries = find_groups_in_lines(
-            result.stdout, self.__ip_addr_show_regex, single_line=False
-        )
+        raw_list = re.finditer(self.__ip_details_pattern, result.stdout)
         found_nics: List[IpInfo] = []
-        for entry in entries:
+        for ip_raw in raw_list:
+            matched = self._get_matched_dict("".join(ip_raw.group().splitlines()))
```
squirrelsc: It doesn't need to split and join. Set the `single_line=False`.

lisa/tools/ip.py
LiliDeng: updated

lisa/tools/ip.py
LiliDeng: it doesn't work when string contains `\r\n`

squirrelsc: Do you mean, the `single_line==False` works for `\n` only? Can you change it to support both? Like to replace `\r\n` to `\n`.

LiliDeng: I'll try.

LiliDeng: updated

squirrelsc (approved):

lisa/tools/ip.py
squirrelsc: It's not using the utils with `single_line==False`. What's the reason?

LiliDeng: when I use find_groups_in_lines single_line = False / True with current pattern, I can't get the value what I want, probably ip_addr, mac and name are not in the same group. previous pattern works well except for below output

```
  3: eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000  # noqa: E501
        link/ether 00:15:5d:33:ff:0b brd ff:ff:ff:ff:ff:ff
```

PR #2835, opened 2023-07-03
Correct env variable name in mariner-mshv.cfg as per new vhd

smit-gardhariya:
Change env variable name in mariner-mshv.cfg as per new vhd configuration for kernel and initrd image.

### DIFF
--- a/lisa/transformers/dom0_kernel_installer.py
+++ b/lisa/transformers/dom0_kernel_installer.py
@@ -178,22 +178,22 @@ def _update_mariner_config(
     current_kernel: str,
     new_kernel: str,
 ) -> None:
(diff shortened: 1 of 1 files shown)


lisa/transformers/dom0_kernel_installer.py
```diff
@@ -178,22 +178,22 @@ def _update_mariner_config(
     current_kernel: str,
     new_kernel: str,
 ) -> None:
-    ll_conf_file: str = "/boot/mariner-mshv.cfg"
+    mariner_config: str = "/boot/mariner-mshv.cfg"
```
LiliDeng, line 181: can you confirm with dev, is there a way to detect/read the expected names, instead of changing them every time when the file name changes

LiliDeng (approved):

PR #2836, opened 2023-07-04
skip installing driver if module has been loaded

LiliDeng:


lisa/features/gpu.py
```diff
@@ -177,7 +172,7 @@ def get_supported_driver(self) -> List[ComputeSDK]:
         raise NotImplementedError()
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
-        self.gpu_vendor: Set[str] = set()
+        self.gpu_vendor: List[str] = ["nvidia"]
```
squirrelsc: What happened, if the vendor is not nvidia?

LiliDeng: from current logic, I found all of them use nvidia and assigned in the code, maybe I can move it to azure feature.

squirrelsc: Can it be checked from code?

LiliDeng: updated, now use whether there is any module used for nvidia pci devices, if no driver installed, the used module is empty.

squirrelsc (approved):

PR #2837, opened 2023-07-06
Move add_azure_core_repo function into operating_system.py

lubaihua33:
Move add_azure_core_repo function into operating_system.py, so this method can be used by other test suites.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -443,6 +443,9 @@ class Posix(OperatingSystem, BaseClassMixin):
     def get_repositories(self) -> List[RepositoryInfo]:
         raise NotImplementedError("get_repositories is not implemented")
 
+    def add_azure_core_repo(self) -> None:
+        raise NotImplementedError("add_azure_core_repo is not implemented")
+
     def _process_extra_package_args(self, extra_args: Optional[List[str]]) -> str:
         if extra_args:
             add_args = " ".join(extra_args)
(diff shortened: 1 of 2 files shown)


lisa/operating_system.py
```diff
@@ -763,6 +766,20 @@ def _get_package_information(self, package_name: str) -> VersionInfo:
         )
         return self._cache_and_return_version_info(package_name, version_info)
 
+    def add_azure_core_repo(self) -> None:
+        self.install_packages(["gnupg", "software-properties-common"])
+        # no azure-compatscanner package in azurecore-debian
+        # use azurecore instead
+        codename = "bionic"
```
LiliDeng, line 780: please add comments for why hardcode codename here :)

lisa/operating_system.py
```diff
@@ -1119,6 +1136,22 @@ def _get_information(self) -> OsInformation:
 
         return information
 
+    def add_azure_core_repo(self) -> None:
+        # there is no available repo for distro which higher than ubuntu 18.04,
+        # use bionic for temp solution
+        if self.information.version >= "18.4.0":
```
squirrelsc: The code was written in 3 years ago. We should check other versions, and add support for 20.04, and >=22.04 as default.

lisa/operating_system.py
lubaihua33: Done

lisa/operating_system.py

squirrelsc:
LGTM

LiliDeng (approved):

PR #2838, opened 2023-07-06
libvirt: add ignition support for Flatcar

tormath1:
Hello,

In this PR, we add the Ignition support to LISA. It allows to provide Ignition configuration instead of cloud-init for initial provisioning on the QEMU platform.

There is base config that is merged with an extra user-data (like for Mariner) and set as fw_cfg on the domain definition. 

```
$ lisa run -r ./microsoft/runbook/qemu/flatcar.yml -v "qcow2:/flatcar/stable/flatcar_production_qemu_image.img" -v "admin_private_key_file:/id_rsa"
 ----------------------------------information----------------------------------- 
area: provisioning
category: functional
tags: []
description: 
        This case verifies whether a node is operating normally.

        Steps,
        1. Connect to TCP port 22. If it's not connectable, failed and check whether
            there is kernel panic.
        2. Connect to SSH port 22, and reboot the node. If there is an error and kernel
            panic, fail the case. If it's not connectable, also fail the case.
        3. If there is another error, but not kernel panic or tcp connection, pass with
            warning.
        4. Otherwise, fully passed.
        
priority: 0
owner: Microsoft
platform: qemu
host_distro: Gentoo Linux
host_kernel_version: 6.1.31-gentoo-dist
libvirt_version: libvirtd (libvirt) 9.3.0
vmm_version: 7.2.0
environment: generated_0
hardware_platform: x86_64
distro_version: Flatcar Container Linux by Kinvolk 3510.2.2 (Oklo)
kernel_version: 5.15.111-flatcar
```

### DIFF
--- a/lisa/sut_orchestrator/libvirt/context.py
+++ b/lisa/sut_orchestrator/libvirt/context.py
@@ -28,11 +28,18 @@ class EnvironmentContext:
 
 
 @dataclass
+class InitSystem:
+    CLOUD_INIT: str = "cloud-init"
+    IGNITION: str = "ignition"
+
+
+@dataclass
 class NodeContext:
     vm_name: str = ""
     firmware_source_path: str = ""
     firmware_path: str = ""
     cloud_init_file_path: str = ""
+    ignition_file_path: str = ""
     os_disk_source_file_path: Optional[str] = None
     os_disk_base_file_path: str = ""
     os_disk_base_file_fmt: DiskImageFormat = DiskImageFormat.QCOW2
@@ -45,6 +52,7 @@ class NodeContext:
     next_disk_index: int = 0
     machine_type: Optional[str] = None
     enable_secure_boot: bool = False
+    init_system: str = InitSystem.CLOUD_INIT
 
     console_logger: Optional[QemuConsoleLogger] = None
     domain: Optional[libvirt.virDomain] = None
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -48,6 +48,7 @@ from . import libvirt_events_thread
 from .console_logger import QemuConsoleLogger
 from .context import (
     DataDiskContext,
+    InitSystem,
     NodeContext,
     get_environment_context,
     get_node_context,
@@ -405,6 +406,9 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
     ) -> None:
         node_context = get_node_context(node)
 
+        if node_runbook.ignition:
+            node_context.init_system = InitSystem.IGNITION
+
         if (
             not node_runbook.firmware_type
             or node_runbook.firmware_type == FIRMWARE_TYPE_UEFI
@@ -430,9 +434,14 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         if not node.name:
             node.name = node_context.vm_name
 
-        node_context.cloud_init_file_path = os.path.join(
-            self.vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
-        )
+        if node_context.init_system == InitSystem.CLOUD_INIT:
+            node_context.cloud_init_file_path = os.path.join(
+                self.vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
+            )
+        else:
+            node_context.ignition_file_path = os.path.join(
+                self.vm_disks_dir, f"{node_context.vm_name}-ignition.json"
+            )
 
         if self.host_node.is_remote:
             node_context.os_disk_source_file_path = node_runbook.disk_img
@@ -570,8 +579,12 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
                         Path(node_context.os_disk_base_file_path),
                     )
 
-        # Create cloud-init ISO file.
-        self._create_node_cloud_init_iso(environment, log, node)
+        if node_context.init_system == InitSystem.CLOUD_INIT:
+            # Create cloud-init ISO file.
+            self._create_node_cloud_init_iso(environment, log, node)
+        else:
+            # Prepate Ignition injection.
+            self._create_node_ignition(environment, log, node)
 
         # Create OS disk from the provided image.
         self._create_node_os_disk(environment, log, node)
@@ -709,13 +722,48 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
                 private_key_file=self.runbook.admin_private_key_file,
             )
 
-            # Ensure cloud-init completes its setup.
-            node.execute(
-                "cloud-init status --wait",
-                sudo=True,
-                expected_exit_code=0,
-                expected_exit_code_failure_message="waiting on cloud-init",
+            node_context = get_node_context(node)
+            if node_context.init_system == InitSystem.CLOUD_INIT:
+                # Ensure cloud-init completes its setup.
+                node.execute(
+                    "cloud-init status --wait",
+                    sudo=True,
+                    expected_exit_code=0,
+                    expected_exit_code_failure_message="waiting on cloud-init",
+                )
+
+    # Setup Ignition for a VM.
+    def _create_node_ignition(
+        self, environment: Environment, log: Logger, node: Node
+    ) -> None:
+        environment_context = get_environment_context(environment)
+        node_context = get_node_context(node)
+
+        user_data = {
+            "ignition": {
+                "version": "3.3.0",
+            },
+            "passwd": {
+                "users": [
+                    {
+                        "name": self.runbook.admin_username,
+                        "sshAuthorizedKeys": [environment_context.ssh_public_key],
+                    },
+                ],
+            },
+        }
+
+        tmp_dir = tempfile.TemporaryDirectory()
+        try:
+            ignition_path = os.path.join(tmp_dir.name, "ignition.json")
+            with open(ignition_path, "w") as f:
+                json.dump(user_data, f)
+
+            self.host_node.shell.copy(
+                Path(ignition_path), Path(node_context.ignition_file_path)
             )
+        finally:
+            tmp_dir.cleanup()
 
     # Create a cloud-init ISO for a VM.
     def _create_node_cloud_init_iso(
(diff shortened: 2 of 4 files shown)


squirrelsc:
@cwize1 Please help reviewing the PR from flatcar. Thank you.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -68,6 +74,9 @@ class BaseLibvirtNodeSchema:
     disk_img_resize_gib: Optional[int] = None
     # Configuration options for cloud-init.
     cloud_init: Optional[CloudInitSchema] = None
+
+    # Configuration options for ignition.
+    ignition: Optional[IgnitionSchema] = None
```
squirrelsc: Add `__post_init__` to validate the schema, make sure cloud_init and ignition not set together.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -405,6 +405,8 @@ def _configure_node(
     ) -> None:
         node_context = get_node_context(node)
 
+        node_context.use_ignition = node_runbook.ignition
```
squirrelsc: Make sure the type is right.
`node_context.use_ignition = node_runbook.ignition is not None`

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -430,9 +432,14 @@ def _configure_node(
         if not node.name:
             node.name = node_context.vm_name
 
-        node_context.cloud_init_file_path = os.path.join(
-            self.vm_disks_dir, f"{node_context.vm_name}-cloud-init.iso"
-        )
+        if not node_context.use_ignition:
```
squirrelsc: It looks more straightforward, if check cloud-init

`if node_context.cloud_init:`

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -709,13 +722,74 @@ def _fill_nodes_metadata(self, environment: Environment, log: Logger) -> None:
...
             )
 
-            # Ensure cloud-init completes its setup.
-            node.execute(
-                "cloud-init status --wait",
-                sudo=True,
-                expected_exit_code=0,
-                expected_exit_code_failure_message="waiting on cloud-init",
+            node_context = get_node_context(node)
+            if not node_context.use_ignition:
```
squirrelsc: `if node_context.cloud_init:`

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -929,14 +1003,22 @@ def _create_node_domain_xml(
...
 
-        self._add_disk_xml(
-            node_context,
-            devices,
-            node_context.cloud_init_file_path,
-            "cdrom",
-            "raw",
-            "sata",
-        )
+        if not node_context.use_ignition:
```
squirrelsc: `if node_context.cloud_init:`

lisa/util/shell.py
```diff
@@ -137,12 +137,13 @@ def try_connect(
     timeout_start = time.time()
     while time.time() < timeout_start + ssh_timeout:
         try:
+            pkey = paramiko.RSAKey.from_private_key_file(connection_info.private_key_file)
```
squirrelsc: What's the reason to change it? It looks the logic works for RSA only.

lisa/sut_orchestrator/libvirt/schema.py
tormath1: I get your point. Can you elaborate on this `__post__init__` ? I tried to simply raise an exception from it but this one is never raised. 🤔

lisa/util/shell.py
tormath1: Weird - I had an issue with paramiko last week. It now works, I upgraded OpenSSL to version 3.0.9 maybe it solved the issue:
```
q must be exactly 160, 224, or 256 bits long
```

lisa/sut_orchestrator/libvirt/context.py
```diff
@@ -45,7 +45,7 @@ class NodeContext:
     next_disk_index: int = 0
     machine_type: Optional[str] = None
     enable_secure_boot: bool = False
-    use_ignition: bool = False
+    use_cloud_init: bool = False
```
squirrelsc: It sets default value to False, make sure no regression for existing usage.

lisa/util/shell.py
squirrelsc: Maybe it relates to OpenSSL version. It happens on some old CentOS 6.x distros as well, because paramiko doesn't fully support some old SSH behaviors.

lisa/sut_orchestrator/libvirt/context.py
tormath1: Correct. Set to True by default to keep current behavior as the default one.

lisa/sut_orchestrator/libvirt/schema.py
squirrelsc: Refer to `lisa/schema.py`. The class `Variable`, `Platform` and `ConnectionInfo` raise exception in `__post_init__`. But this is an extension runbook, maybe the behavior is different. Please help checking if log has the exception dumpped, but not raised out. I will take a look if it is.

lisa/sut_orchestrator/libvirt/context.py
```diff
@@ -45,6 +45,7 @@ class NodeContext:
     next_disk_index: int = 0
     machine_type: Optional[str] = None
     enable_secure_boot: bool = False
+    use_cloud_init: bool = True
```
cwize1: Would it make sense to change this to a string enum named something like "init_system" with values "cloud-init" and "ignition". That way, it is easier to add another system in the future.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -709,13 +722,74 @@ def _fill_nodes_metadata(self, environment: Environment, log: Logger) -> None:
...
+                   {
+                       "name": self.runbook.admin_username,
+                       "sshAuthorizedKeys": [environment_context.ssh_public_key],
+                   },
+               ],
+           },
+        }
+
+        # Iterate through all the top-level properties.
+        for extra_user_data in node_context.extra_cloud_init_user_data:
```
cwize1: The cloud-init data structure is fairly flat. So, this simple merge scheme works well for most things. But it looks like Ignition schema is a little more nested. For example, the `users` list sits under the `passwd` property. So, the merge scheme might not be particularly helpful.

Do you actually have a use case for merging user provided values? If not, it might be worth removing this for now.

cwize1: Please avoid duplicating code. (Though this might be a moot comment based on the outcome of the other comment.)

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -709,13 +722,74 @@ def _fill_nodes_metadata(self, environment: Environment, log: Logger) -> None:
...
+                    # Just replace the existing property.
+                    user_data[key] = value
+
+        user_data_string = json.dumps(user_data)
+
+        tmp_dir = tempfile.TemporaryDirectory()
+        try:
+            ignition_path = os.path.join(tmp_dir.name, "ignition.json")
+            with open(ignition_path, "w") as f:
+                f.write(user_data_string)
```
cwize1: nit: `json.dump(user_data, f)`

The cloud-init logic uses `json.dumps` because it needs to add some extra stuff to the JSON file.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -929,14 +1003,22 @@ def _create_node_domain_xml(
...
+                devices,
+                node_context.cloud_init_file_path,
+                "cdrom",
+                "raw",
+                "sata",
+            )
+        else:
+            sysinfo_tag = ET.SubElement(domain, "sysinfo")
+            sysinfo_tag.attrib["type"] = "fwcfg"
+            entry_tag = ET.SubElement(sysinfo_tag, "entry")
```
cwize1, line 993: nit: Add new line before this line

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -15,6 +15,12 @@ class CloudInitSchema:
     # Additional values to apply to the cloud-init user-data file.
     extra_user_data: Optional[Union[str, List[str]]] = None
 
+# Configuration options for Ignition provisioning for the VM.
+@dataclass_json()
+@dataclass
+class IgnitionSchema:
+    # Additional values to apply to the Ignition user-data file.
+    extra_user_data: Optional[Union[str, List[str]]] = None
```
cwize1: In cloud-init, the file is literally named `user-data`. Are the equivalent files also called `user-data` for Ignition?

cwize1 (changes requested):
.

LiliDeng:
@tormath1 let know, do you plan to update this PR, thank you.

tormath1:
@LiliDeng thanks for the heads-up.

squirrelsc:
@cwize1 Can you have a look?

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -56,6 +56,12 @@ class DiskImageFormat(Enum):
     RAW = "raw"
 
 
+# Possible provisioning tools
+class InitSystem(Enum):
```
cwize1: This type isn't a part of the runbook schema. So, it should to `context.py` instead.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -405,6 +406,9 @@ def _configure_node(
     ) -> None:
         node_context = get_node_context(node)
 
+        if node_runbook.ignition:
```
cwize1, line 409: It would probably be goodness to ensure that `ignition` and `cloud_init` are not specified at the same time.

cwize1 (changes requested):

lisa/sut_orchestrator/libvirt/platform.py
tormath1: Good idea, done.

cwize1 (approved):

squirrelsc (approved):

PR #2840, opened 2023-07-09
only use production vm sizes

LiliDeng:


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1486,14 +1486,15 @@ def _select_vm_size(
         # current location and that are available for the current vm size to resize to
         for size in available_sizes:
             vm_size_name = size.as_dict()["name"]
-            # Getting eligible vm sizes and their capability data
-            new_vm_size = next(
-                (x for x in sorted_sizes if x.vm_size == vm_size_name), None
-            )
-            if not new_vm_size:
-                continue
+            if vm_size_name.startswith(("Standard_", "Basic_")):
```
squirrelsc, line 1489: Does it need to hardcode rules here? It makes other vm sizes are not testable.

LiliDeng: @johnsongeorge-w do you know any public doc saying that azure public vm sizes's names just start with Standard_ and Basic_?

squirrelsc: It's not about if the prefix correctly or not. My concern is if it blocks tests on other names.

LiliDeng: this fix is for `Resize fails with non-Standard SKUs`, we don't want VM to be resized into an internal VM size.

squirrelsc: I see, I haven't seen it's for resize only.

squirrelsc (approved):

PR #2844, opened 2023-07-11
skip install ofed driver when the image is hpc type

LiliDeng:


lisa/features/infiniband.py
```diff
@@ -171,7 +171,34 @@ def get_pkey(self) -> str:
...
+        sed = node.tools[Sed]
+        sed.substitute(
+            regexp="SELINUX=enforcing",
+            replacement="SELINUX=disabled",
+            file="/etc/selinux/config",
+            sudo=True,
+        )
+
+        if isinstance(node.os, Ubuntu) and node.os.information.version > "18.4.0":
+            node.tools[Sed].substitute(
```
squirrelsc, line 199: Add comments above to explain the purpose.

LiliDeng: updated

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -433,20 +433,42 @@ def verify_ibm_mpi(self, environment: Environment, log: Logger) -> None:
         server_ssh.add_known_host(client_ip)
         client_ssh.add_known_host(server_ip)
 
+        if server_node.tools[Ls].path_exists("/usr/share/modules/init/bash"):
```
squirrelsc: why need to check this bash? what's different if it has or not? If it's really useful, add a comment above.

LiliDeng: it indicates a tool module has been installed, now I add is_hpc_image field.

squirrelsc (approved):

PR #2846, opened 2023-07-13
fix pytorch case found in hpc images

LiliDeng:


lisa/node.py
```diff
@@ -394,6 +395,9 @@ def find_partition_with_freespace(
                     else:
                         mountpoint = partition.mountpoint
 
+                    if df.get_filesystem_available_space(mountpoint, True) < size_in_gb:
```
squirrelsc: There is a check `if not partition.size_in_gb >= size_in_gb:` already, why need to check here again?

lisa/node.py
```diff
@@ -412,6 +416,9 @@ def find_partition_with_freespace(
                 else:
                     mountpoint = disk.mountpoint
 
+                if df.get_filesystem_available_space(mountpoint, True) < size_in_gb:
```
squirrelsc: What's the reason to check it again?

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -45,6 +45,10 @@ class GpuTestSuite(TestSuite):
     TIMEOUT = 2000
 
     _pytorch_pattern = re.compile(r"^gpu count: (?P<count>\d+)", re.M)
+    _pytorch_error_pattern = re.compile(
```
squirrelsc: Name it like `numpy_error`

lisa/node.py
LiliDeng: this is to compare total size, not free size,  I will remove this, just check free size

lisa/node.py

microsoft/testsuites/gpu/gpusuite.py
LiliDeng: updated

lisa/node.py
squirrelsc: Move the logic to be shared. The return part happens on both `if/else`, so it can be moved out from `if/else` too.

squirrelsc (approved):

PR #2848, opened 2023-07-14
Add MDM test case

lubaihua33:
1. Add parameter repo_name in add_azure_core_repo. Then it can choose to use which repo for different scenarios.
2. Add _uninstall_package for Debian and RPMDistro. 
3. Add _package_exists for CBLMariner. When checking if a package exists,  it needs _initialize_package_installation to get _dnf_tool_name for the Mariner image.
4. Add MDM test case.


lisa/operating_system.py
```diff
@@ -443,7 +464,7 @@ def get_package_information(
     def get_repositories(self) -> List[RepositoryInfo]:
         raise NotImplementedError("get_repositories is not implemented")
 
-    def add_azure_core_repo(self) -> None:
+    def add_azure_core_repo(self, repo_name: Optional[AzureCoreRepo] = None) -> None:
```
squirrelsc, line 467: Should the repo name is set by distro class? It doesn't need to be passed in from caller.

lubaihua33: This **repo_name** is primarily used for Debian images. In the past, only tvm test cases required the addition of the **azurecore** repo. The test case **verify_measuredboot_compatibility** needs the **azure-compatscanner** package, which is absent in the **azurecore-debian** repo, so the **azurecore** repo from Ubuntu Bionic is used instead.

However, if more test cases require the **azurecore** repo, using Ubuntu's repo for Debian images would result in the **azurecore-debian** repo not being covered in tests.


To address this, this change has been made. By default, if the **repo_name** parameter is None, Debian Buster images will use the **azurecore-debian** repo. In other scenarios, such as when the "**azure-compatscanner**" package is required, the caller can define the use of the **azurecore** repo from Ubuntu.

squirrelsc: Ok. If it's for increasing coverage, it looks good to me. If it has to be specified for different distro, it should be handled inside of the os classes.

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,97 @@
...
+        mdm_option = variables.get("mdm_option", "")
+        if mdm_option.strip():
+            echo = node.tools[Echo]
+            echo.write_to_file(
+                value=f"{mdm_option}",
+                file=PurePath("/etc/default/mdm"),
+                sudo=True,
+            )
+            log.info("Setting MDM_OPTIONS")
+            log.info(node.execute("cat /etc/default/mdm", sudo=True).stdout)
```
squirrelsc: The output is already in debug level, it doesn't need to log it again. The info level is to tell a story. Please review all other info log, make sure it's really necessary.

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,97 @@
...
+            posix_os.add_azure_core_repo()
+
+        # Install metricsext2
+        is_installed = posix_os.package_exists(package)
+        if not is_installed:
+            if posix_os.is_package_in_repo(package):
+                posix_os.install_packages(package)
+                log.info("MetricsExtension is installed successfully")
+            else:
+                raise LisaException(
```
squirrelsc: Raise the unsupported OS with extra message, so the OS information can be in the error message.

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,97 @@
...
+                    f"The supported distro doesn't have {package} in its repo"
+                )
+        else:
+            log.info("MetricsExtension is already installed")
+
+        # Set MDM options
+        mdm_option = variables.get("mdm_option", "")
+        if mdm_option.strip():
```
squirrelsc: move `.strip()` to above line, so the logic is easier to read.

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,97 @@
...
+                value=f"{mdm_option}",
+                file=PurePath("/etc/default/mdm"),
+                sudo=True,
+            )
+            log.info("Setting MDM_OPTIONS")
+            log.info(node.execute("cat /etc/default/mdm", sudo=True).stdout)
+            log.info("MetricsExtension is installed successfully")
+
+            # Start the service and check the status
+            service = node.tools[Service]
```
squirrelsc: Should the service starting out of installation? Maybe the service is installed, but not started.

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,97 @@
...
+            service.restart_service("mdm")
+            log.info("MetricsExtension is started successfully")
```
squirrelsc: This log line is duplicate with below line, so it should be removed.

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,97 @@
...
+    This test is a BVT for MDM MetricsExtension
+    """,
+)
+class MetricsExtension(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            Verify whether MetricsExtension is installed, running,
+            and uninstalled successfully
+        """,
+        priority=1,
```
squirrelsc, line 32: Limit this test case for Azure, Ready only.

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Done

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Done

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Done

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Yeah, the above code checks if metricsext2 package is already installed. If installed, then restart the service directly.

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Done

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Done

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,106 @@
...
+        self._is_supported(node)
+
+        # Add repo
+        if isinstance(posix_os, CBLMariner):
+            release = posix_os.information.release
+            curl = node.tools[Curl]
+            curl.fetch(
+                arg="-o /etc/yum.repos.d/mariner-extras.repo",
+                execute_arg="",
+                url=f"https://raw.githubusercontent.com/microsoft/CBL-Mariner/{release}/SPECS/mariner-repos/mariner-extras.repo",  # noqa: E501
```
squirrelsc: Split it to two strings like
```python
url=f"https://raw.githubusercontent.com/microsoft/CBL-Mariner/{release}"
         "/SPECS/mariner-repos/mariner-extras.repo",

Comment (lubaihua33) [reply microsoft/testsuites/vm_extensions/mdm.py]: Done

File: microsoft/testsuites/vm_extensions/mdm.py [hunk 9]
```diff
@@ -0,0 +1,107 @@
...
+        assert_that(service.check_service_status("mdm")).is_equal_to(True)
+        log.info("MetricsExtension is running successfully")
+
+        # Uninstall metricsext2 if it doesn't exist originally
+        if not is_installed:
+            posix_os.uninstall_packages(package)
+            log.info("MetricsExtension is removed successfully")
+
+    def _is_supported(self, node: Node) -> None:
+        # MDM is only supportedi
```
GreyCat:
```suggestion
        # MetricsExtension only supports these distributions
```

microsoft/testsuites/vm_extensions/mdm.py
```diff
@@ -0,0 +1,107 @@
...
+        supported_versions: Dict[Any, List[str]] = {
+            CBLMariner: ["1.0", "2.0"],
+            Ubuntu: ["18.04", "20.04"],
+            Debian: [f"10.{i}" for i in range(0, 14)],
+        }
+        release = node.os.information.release
+        if release not in supported_versions.get(type(node.os), []):
+            raise SkippedException(
+                UnsupportedDistroException(node.os, "MDM doesn't support this version.")
```
GreyCat:
```suggestion
                UnsupportedDistroException(node.os, "MetricsExtension doesn't support this version.")
```

squirrelsc:
@LiliDeng  LGTM

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Done

microsoft/testsuites/vm_extensions/mdm.py
lubaihua33: Done

LiliDeng (approved):

PR #2850, opened 2023-07-16
Fix SRIOV testcases for FreeBSD

dsrivastavv:


microsoft/testsuites/network/common.py
```diff
@@ -26,7 +27,10 @@ def initialize_nic_info(
             assert_that(sriov_count).described_as(
                 f"there is no sriov nic attached to VM {node.name}"
             ).is_greater_than(0)
-        nics_info = Nics(node)
+        if not isinstance(node.os, BSD):
```
squirrelsc: This should be wrapped into nic.py. Use a method like `create_nics`

lisa/nic.py
```diff
@@ -492,3 +492,85 @@ def get_packets(self, nic_name: str, name: str = "tx_packets") -> int:
...
+            nic_names = [nic_name]
+        else:
+            nic_names = self.get_nic_names()
+        nics = [self.nics[nic_name] for nic_name in nic_names]
+        for nic in nics:
+            nic.ip_addr = ip_tool.get_ip_address(nic.name)
+            nic.mac_addr = ip_tool.get_mac(nic.name)
+
+    def _get_nic_uuids(self) -> None:
+        pass
```
squirrelsc, line 582: Will it be implemented later? If not, add comments to explain why it's empty.

dsrivastavv: No. These are not needed yet but could be for later testcases. I will add a comment mentioning this.

microsoft/testsuites/network/common.py
dsrivastavv: Sure! `Nics` class is also used in bunch of other places and I am testing if replacing those won't have any regression in existing testcases

lisa/nic.py
squirrelsc: If so, it looks like to skip Linux logic. Please add comments to explain it.

microsoft/testsuites/network/common.py
```diff
@@ -26,7 +27,7 @@ def initialize_nic_info(
             assert_that(sriov_count).described_as(
                 f"there is no sriov nic attached to VM {node.name}"
             ).is_greater_than(0)
-        nics_info = Nics(node)
+        nics_info = create_nics(node)
```
squirrelsc: It looks `node.nics` should work now. It doesn't need to use like `Nics(node)`

microsoft/testsuites/network/netinterface.py
```diff
@@ -123,7 +123,7 @@ def verify_network_interface_reload_via_ip_link(
         # Ensure netvsc module is loaded
         network_interface_feature.reload_module()
 
-        node_nic_info = Nics(node)
+        node_nic_info = create_nics(node)
```
squirrelsc, line 123: replace to node.nics.

squirrelsc:
@LiliDeng Overall, LGTM. Please start to look at it, so it can be merged early.

lisa/tools/dhclient.py
```diff
@@ -72,3 +76,20 @@ def renew(self, interface: Optional[str] = None) -> None:
...
             0, f"dhclient renew return non-zero exit code: {result.stdout}"
         )
+
+
+class DhclientFreeBSD(Dhclient):
+    @property
+    def command(self) -> str:
+        return "dhclient"
+
+    def renew(self, interface: str | None = None) -> None:
```
squirrelsc: Write like below, so it doesn't need to deal with None
```suggestion
    def renew(self, interface: str = “”) -> None:
```

LiliDeng (dismissed):

microsoft/testsuites/network/netinterface.py
dsrivastavv: I will update this in my next PR dealing with network testcases. This would give me a chance to verify the change properly.

dsrivastavv: @squirrelsc Is this okay?

LiliDeng (approved):

microsoft/testsuites/network/netinterface.py
squirrelsc: Can all `create_nics` change to node.nics?

dsrivastavv: If any of the test case changes the nic information like ip or disable mlx driver, we would need to call `create_nic` again since the previous information is outdated

squirrelsc: In this case, it should call nics.reload to make it meaningful.

dsrivastavv: Discussed offline. I will address this change soon in a separate PR

PR #2853, opened 2023-07-18
DPDK: add is_mana, handle driver load, use pkg-config for version check

mcgov:
Two commits of a larger set refactoring DPDK.

MANA: add is_mana and handle driver loading & testpmd arugments.

pkg-config:
Parsing git tags and tar file names was a dumb choice on my part, moving to pkg-config to get libdpdk information. This has the benefit of allowing for shared gallery images with pre-build versions of DPDK to speed up testing.

- Add Pkg-Config tool, allows coercing of nonstandard version strings for semver
- Use pkg-config tool to get DPDK version info instead of parsing tags and filenames.

### DIFF
--- a/lisa/tools/pkgconfig.py
+++ b/lisa/tools/pkgconfig.py
@@ -50,10 +50,15 @@ class Pkgconfig(Tool):
         coerced_version = ".".join([str(int(x)) for x in [major, minor, patch]])
         return VersionInfo.parse(coerced_version)
 
-    def _get_package_info(self, package_name: str, update_cached: bool = False) -> str:
+    def get_package_info(
+        self, package_name: str, check_exists: bool = False, update_cached: bool = False
+    ) -> str:
         package_info_result = self.run(
             f"--modversion {package_name}", force_run=update_cached
         )
+        if check_exists and package_info_result.exit_code != 0:
+            return ""
+
         assert_that(package_info_result.exit_code).described_as(
             (
                 f"pkg-config information was not available for {package_name}. "
@@ -66,5 +71,5 @@ class Pkgconfig(Tool):
     def get_package_version(
         self, package_name: str, update_cached: bool = False
     ) -> VersionInfo:
-        version_info = self._get_package_info(package_name, update_cached=update_cached)
+        version_info = self.get_package_info(package_name, update_cached=update_cached)
         return self._coerce_version_info_string(version_info)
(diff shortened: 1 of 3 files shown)


lisa/tools/pkgconfig.py
```diff
@@ -0,0 +1,67 @@
...
+        ).is_not_none()
+
+        coerced_version = ".".join([str(int(x)) for x in [major, minor, patch]])
+        return VersionInfo.parse(coerced_version)
+
+    def package_info_exists(self, package_name: str) -> bool:
+        return self.run(f"--modversion {package_name}").exit_code == 0
+
+    def get_package_version(self, package_name: str) -> VersionInfo:
+        assert_that(self.package_info_exists(package_name)).described_as(
```
squirrelsc: Did you meet this error before? If there is any error on detection or installation, it should be fix there.

mcgov: Nope, just added it for defensive programming.

squirrelsc: This step slow down the progress. So it should be removed.

mcgov: Wouldn't it be cached after the first call?

mcgov: I can re-work to make this a little cleaner.

mcgov: re-worked! Made it clearer that there should be one (cacheable) call to run()

mcgov: Oh wait, no that's not right. Don't merge, need more time to fix this.

mcgov: Apologies, I started to second guess what was there. Currently package information *should* always be available for libdpdk when this tool is called.

There is an edge case where you check pkg-config --modversion, see DPDK is not a high enough version, update the package, then check again. In that case you would want to force_run to update the cached package information after the update. This could occur when environments get re-used depending on the test case, specifically in Ubuntu 18.04.

For now I would like to leave that assertion in place; this is new code and we support (and use) many methods for installing DPDK.

mcgov: Since I moved some things around, I'm running additional testing now. I'll update when it's ready for another review.

mcgov: Should good to go, I have additional changes blocked on this so we can go ahead and merge.

mcgov:
breaking up and will open new PRs

PR #2858, opened 2023-07-24
Enable hot add disk testcases on FreeBSD

dsrivastavv:


squirrelsc:
Add code comment to explain why it needs force_rerun.

dsrivastavv:
> Add code comment to explain why it needs force_rerun.

Updated the PR. The function is parameterized to take it as an argument

squirrelsc (approved):

PR #2861, opened 2023-07-25
Ultra Disk Support

kamalca:
- Separates OS disk type from the data disk type
- Adds VM SKU feature check for Ultra disk
- Adds Ultra disk type
- Makes changes to ARM template to allow for Ultra disk to be deployed
- Set Ultra Disk IOPS and throughput in the ARM template. I implemented this by always setting to the max IOPS and throughput for a given disk size. Restricting IOPS / throughput on a larger disk size is not a current use case. 

NOT implemented (yet)
- Availability zones: many regions require ultra disks to be deployed in an availability zone
- Resolving availability set conflict gracefully: Ultra disks cannot be deployed in an availability set. This will necessarily conflict with any tip session parameters set in a runbook.
- Caching: caching type must be set to none

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -409,6 +409,7 @@ class DiskType(str, Enum):
     Ephemeral = "Ephemeral"
     StandardHDDLRS = "StandardHDDLRS"
     StandardSSDLRS = "StandardSSDLRS"
+    UltraSSDLRS = "UltraSSDLRS"
 
 
 # disk types are ordered by commonly and cost. The earlier is lower cost.
@@ -417,6 +418,7 @@ disk_type_priority: List[DiskType] = [
     DiskType.StandardSSDLRS,
     DiskType.Ephemeral,
     DiskType.PremiumSSDLRS,
+    DiskType.UltraSSDLRS,
 ]
 
 
@@ -435,7 +437,7 @@ disk_controller_type_priority: List[DiskControllerType] = [
 @dataclass()
 class DiskOptionSettings(FeatureSettings):
     type: str = constants.FEATURE_DISK
-    disk_type: Optional[
+    os_disk_type: Optional[
         Union[search_space.SetSpace[DiskType], DiskType]
     ] = field(  # type:ignore
         default_factory=partial(
@@ -445,6 +447,22 @@ class DiskOptionSettings(FeatureSettings):
                 DiskType.StandardSSDLRS,
                 DiskType.Ephemeral,
                 DiskType.PremiumSSDLRS,
+            ],
+        ),
+        metadata=field_metadata(
+            decoder=partial(search_space.decode_set_space_by_type, base_type=DiskType)
+        ),
+    )
+    disk_type: Optional[
+        Union[search_space.SetSpace[DiskType], DiskType]
+    ] = field(  # type:ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            items=[
+                DiskType.StandardHDDLRS,
+                DiskType.StandardSSDLRS,
+                DiskType.PremiumSSDLRS,
+                DiskType.UltraSSDLRS,
             ],
         ),
         metadata=field_metadata(
@@ -512,6 +530,7 @@ class DiskOptionSettings(FeatureSettings):
         assert isinstance(o, DiskOptionSettings), f"actual: {type(o)}"
         return (
             self.type == o.type
+            and self.os_disk_type == o.os_disk_type
             and self.disk_type == o.disk_type
             and self.data_disk_count == o.data_disk_count
             and self.data_disk_caching_type == o.data_disk_caching_type
@@ -523,6 +542,7 @@ class DiskOptionSettings(FeatureSettings):
 
     def __repr__(self) -> str:
         return (
+            f"os_disk_type: {self.os_disk_type},"
             f"disk_type: {self.disk_type},"
             f"count: {self.data_disk_count},"
             f"caching: {self.data_disk_caching_type},"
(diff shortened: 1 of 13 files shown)


kamalca:
Hey @squirrelsc,

I want eyes on these changes early on. Supporting ultra disks requires changing a lot about how disks currently work because I have to separate the OS disk type from the data disk type. Disk type is used in many places throughout the code and test cases. Each use has different intentions for affecting os disk type or data disk type.

If you have advice for splitting up these fields without breaking many things, please let me know.

lisa/schema.py
```diff
@@ -450,6 +452,22 @@ class DiskOptionSettings(FeatureSettings):
             decoder=partial(search_space.decode_set_space_by_type, base_type=DiskType)
         ),
     )
+    disk_type: Optional[
```
squirrelsc: name it like `data_disk_type`?

lisa/schema.py
```diff
@@ -434,7 +436,7 @@ class DiskControllerType(str, Enum):
 @dataclass()
 class DiskOptionSettings(FeatureSettings):
     type: str = constants.FEATURE_DISK
-    disk_type: Optional[
+    os_disk_type: Optional[
```
squirrelsc, line 440: Does os disk support all types?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1643,6 +1650,13 @@ def _resource_sku_to_capability(  # noqa: C901
 
         # fill supported features
         azure_raw_capabilities: Dict[str, str] = {}
+        for location_info in resource_sku.location_info:
```
squirrelsc, line 1656: What is this for? Is it related? if not, move it to a separated PR.

squirrelsc:
> Hey @squirrelsc,
> 
> I want eyes on these changes early on. Supporting ultra disks requires changing a lot about how disks currently work because I have to separate the OS disk type from the data disk type. Disk type is used in many places throughout the code and test cases. Each use has different intentions for affecting os disk type or data disk type.
> 
> If you have advice for splitting up these fields without breaking many things, please let me know.

Thank you, Kameron. Overall, it looks straightforward. For current test cases, we assume it tests both os and data disks together. So let the test requirements to be copied. The test cases' behavior should be the same as before.

lisa/schema.py
kamalca: No, OS disk does not support Ultra Disk.

lisa/sut_orchestrator/azure/platform_.py
kamalca: The SKU capability for Ultra Disk is stored under location info because sometimes it is only available in certain availability zones.

``` yaml
"locationInfo": [
{
  "location": "WestUS3",
  "zoneDetails": [
    {
      "Name": [
        "1",
        "3",
        "2"
      ],
      "capabilities": [
        {
          "name": "UltraSSDAvailable",
          "value": "True"
        }
      ],
      "name": null
    }
  ],
  "zones": [
    "1",
    "3",
    "2"
    ]
  }
],
```

lisa/schema.py
kamalca: Changed the name

kamalca:
@squirrelsc 
I followed your advice on copying the test requirements to keep the same behavior.

I am afraid that changing the name from `disk_type` to `data_disk_type` has broken AWS deployment. I have never done AWS deployment. I'm not sure how to test that functionality. My original plan was to not change AWS, but I don't think that is possible with the name change. Please advise on the subject of AWS.

kamalca:
@LiliDeng, Chi is OOF until the 12th. Can you give feedback until then?

@rainierside It looks like you implemented a lot of the AWS functionality. Can you provide input on whether these changes are breaking? I'm also curious if you have a workstream for testing AWS functionality of LISA.

rainierside:
When I was working on the AWS functionality, it supports some basic functionality, like create EC2 instance, set up network, attach some disk and log into the EC2 to run test case.
As the disk type of Lisa test cases is combined closely with Azure, not all disk types of AWS are mapped in the Lisa.

Please let me know if you have any more questions.

Thanks,
Ye

LiliDeng:
have you triggered some testing? it failed for me with below combination 

```
      disk:
        data_disk_type: StandardHDDLRS
        os_disk_type: StandardHDDLRS
        data_disk_count:
          min: 2
        data_disk_caching_type: None
        data_disk_size:
          min: 10
```

LiliDeng:
is it below one enough to run a deployment with UltraSSDLRS, I got skip result

```
      disk:
        os_disk_type: UltraSSDLRS
```

kamalca:
> have you triggered some testing? it failed for me with below combination
> 
> ```
>       disk:
>         data_disk_type: StandardHDDLRS
>         os_disk_type: StandardHDDLRS
>         data_disk_count:
>           min: 2
>         data_disk_caching_type: None
>         data_disk_size:
>           min: 10
> ```

@LiliDeng 
Apparently I did not do enough testing... caught a small typo and should be fixed now.


---


> is it below one enough to run a deployment with UltraSSDLRS, I got skip result
> 
> ```
>       disk:
>         os_disk_type: UltraSSDLRS
> ```

This should be `data_disk_type: UltraSSDLRS`. Ultra disk is not supported as OS disk.

Make sure you are not setting `availability_set_tags` or `availability_set_properties`. Also, currently only the following regions support without availability zone:

- Australia Central
- Brazil South
- Central India
- East Asia
- Germany West Central
- Korea Central
- Korea South
- North Central US, South Central US, West US
- US Gov Arizona, US Gov Texas, US Gov Virginia

If you have any other issues, please send me the full runbook.

lisa/sut_orchestrator/azure/platform_.py
kamalca: I may wait to add this until availability zones are supported by LISA. If we set data disk to Ultra Disk and leave location blank, it would be better to select location where availability zones are not necessary / supported.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

LiliDeng:
There are some items we need to think about. @kamalca 

- os disk caching is not customized?
- test should be skip instead of fail when use available set 
- test should be skip instead of fail when the current location needs availability zone

PR #2862, opened 2023-07-27
Git Bisect Combinator

adityagesh:
- Git bisect combinator clones code into a VM and loops the extended_phase in runbook
- Add KernelBuildMessage, which can be subscribed by notifiers
- Add "bisect" command in git tool

Sample Runbook:

`transformer:
  - type: azure_deploy
    name: bisect_vm
    requirement:
      azure:
        location: $(location)
      core_count: 2
    enabled: true
  - type: azure_deploy
    phase: expanded
    name: build_vm
    requirement:
      azure:
        location: $(location)
      core_count: 16
    enabled: true
  - type: kernel_installer
    phase: expanded
    connection:
      address: $(build_vm_address)
    installer:
      type: source
      location:
        type: repo
        path: /mnt/code
        ref: $(ref)
        repo: "git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git"
    raise_exception: False
  // Do not create vhd when build fails
  - type: azure_vhd
    enabled: $(kernel_installer_boot_success)
    phase: expanded
    resource_group_name: $(build_vm_resource_group_name)
    rename:
      azure_vhd_url: vhd
  // Do not run test when build fails
  - type: script
    phase: expanded
    scripts:
      - name: test_case_name
        variables:
          - test_case_name
          - kernel_installer_build_success
          - kernel_installer_boot_success
        script: "test_case_name if kernel_installer_build_success and kernel_installer_boot_success else ''"
  - type: azure_delete
    resource_group_name: $(build_vm_resource_group_name)
    phase: expanded_cleanup
  - type: azure_delete
    resource_group_name: $(bisect_vm_resource_group_name)
    phase: cleanup

combinator:
  - type: git_bisect
    repo: "git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git"
    bad_commit: "HEAD"
    good_commit: "fe1de55167963a1c0ebe1579e37a8a41495f0a81"
    connection:
       address: $(bisect_vm_address)

platform:
  - type: azure
    azure:
      subscription_id: $(subscription_id)
    requirement:
      azure:
        location: $(location)
        vhd: $(vhd)

testcase:
  - criteria:
      name: $(script_test_case_name)`

### DIFF
--- a/lisa/combinators/git_bisect_combinator.py
+++ b/lisa/combinators/git_bisect_combinator.py
@@ -1,0 +1,191 @@
+import pathlib
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import messages, notifier, schema
+from lisa.combinator import Combinator
+from lisa.messages import KernelBuildMessage, TestResultMessage, TestStatus
+from lisa.node import Node, quick_connect
+from lisa.tools.git import Git, GitBisect
+from lisa.util import LisaException, constants, field_metadata
+
+STOP_PATTERNS = ["first bad commit", "This means the bug has been fixed between"]
+
+
+# Combinator requires a node to clone the source code.
+@dataclass_json()
+@dataclass
+class GitBisectCombinatorSchema(schema.Combinator):
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    repo: str = field(
+        default="",
+        metadata=field_metadata(
+            required=True,
+        ),
+    )
+    good_commit: str = field(
+        default="",
+        metadata=field_metadata(
+            required=True,
+        ),
+    )
+    bad_commit: str = field(
+        default="",
+        metadata=field_metadata(
+            required=True,
+        ),
+    )
+
+
+# GitBisect Combinator is a loop that runs "expanded" phase
+# of runbook until the bisect is complete.
+# There can be any number of expanded phases, but the
+# GitBisectTestResult notifier should have on boolean/None output per
+# phase.
+
+
+class GitBisectCombinator(Combinator):
+    def __init__(
+        self,
+        runbook: GitBisectCombinatorSchema,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(runbook)
+        self._iteration = 0
+        self._result_notifier = GitBisectResult(schema.Notifier())
+        notifier.register_notifier(self._result_notifier)
+        self._source_path: pathlib.PurePath
+        self._node: Optional[Node] = None
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._clone_source()
+        if self._source_path:
+            self._start_bisect()
+        else:
+            raise LisaException(
+                "Source path is not set. Please check the source clone."
+            )
+
+    @classmethod
+    def type_name(cls) -> str:
+        return constants.COMBINATOR_GITBISECT
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return GitBisectCombinatorSchema
+
+    def _next(self) -> Optional[Dict[str, Any]]:
+        _next: Optional[Dict[str, Any]] = None
+        self._process_result()
+        if not self._check_bisect_complete():
+            _next = {}
+            _next["ref"] = self._get_current_commit_hash()
+        else:
+            self._log.info("Bisect Complete")
+        self._result_notifier.result = None
+        self._iteration += 1
+        return _next
+
+    def _process_result(self) -> None:
+        if self._iteration == 0:
+            return
+        if self._result_notifier.result is not None:
+            results = self._result_notifier.result
+            if results:
+                self._bisect_good()
+            else:
+                self._bisect_bad()
+        else:
+            raise LisaException(
+                "Bisect combinator does not get result for next iteration. Please check"
+                " GitBisectResult notifier."
+            )
+
+    def _get_remote_node(self) -> Node:
+        if not self._node or not self._node.is_connected:
+            self._node = quick_connect(self.runbook.connection, "source_node")
+        return self._node
+
+    def _clone_source(self) -> None:
+        node = self._get_remote_node()
+        git = node.tools[Git]
+        self._source_path = git.clone(
+            url=self.runbook.repo, cwd=node.working_path, timeout=1200
+        )
+        node.close()
+
+    def _start_bisect(self) -> None:
+        node = self._get_remote_node()
+        git_bisect = node.tools[GitBisect]
+        git_bisect.start(cwd=self._source_path)
+        git_bisect.good(cwd=self._source_path, ref=self.runbook.good_commit)
+        git_bisect.bad(cwd=self._source_path, ref=self.runbook.bad_commit)
+        node.close()
+
+    def _bisect_bad(self) -> None:
+        node = self._get_remote_node()
+        git_bisect = node.tools[GitBisect]
+        git_bisect.bad(cwd=self._source_path)
+        node.close()
+
+    def _bisect_good(self) -> None:
+        node = self._get_remote_node()
+        git_bisect = node.tools[GitBisect]
+        git_bisect.good(cwd=self._source_path)
+        node.close()
+
+    def _check_bisect_complete(self) -> bool:
+        node = self._get_remote_node()
+        git_bisect = node.tools[GitBisect]
+        result = git_bisect.check_bisect_complete(cwd=self._source_path)
+        node.close()
+        return result
+
+    def _get_current_commit_hash(self) -> str:
+        node = self._get_remote_node()
+        git = node.tools[Git]
+        result = git.get_current_commit_hash(cwd=self._source_path)
+        node.close()
+        return result
+
+
+class GitBisectResult(notifier.Notifier):
+    @classmethod
+    def type_name(cls) -> str:
+        return "git_bisect_result"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return schema.Notifier
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.result: Optional[bool] = None
+
+    def _received_message(self, message: messages.MessageBase) -> None:
+        if isinstance(message, messages.TestResultMessage):
+            self._update_test_result(message)
+        elif isinstance(message, messages.KernelBuildMessage):
+            self._update_result(message.is_success)
+        else:
+            raise LisaException(f"Received unsubscribed message type: {type(message)}")
+
+    def _update_test_result(self, message: messages.TestResultMessage) -> None:
+        if message.is_completed:
+            if message.status == TestStatus.FAILED:
+                self._update_result(False)
+            elif message.status == TestStatus.PASSED:
+                self._update_result(True)
+
+    def _update_result(self, result: bool) -> None:
+        current_result = self.result
+        if current_result is not None:
+            self.result = current_result and result
+        else:
+            self.result = result
+
+    def _subscribed_message_type(self) -> List[Type[messages.MessageBase]]:
+        return [TestResultMessage, KernelBuildMessage]
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -263,6 +263,15 @@ class ProvisionBootTimeMessage(MessageBase):
     information: Dict[str, str] = field(default_factory=dict)
 
 
+@dataclass
+class KernelBuildMessage(MessageBase):
+    type: str = "KernelBuild"
+    old_kernel_version: str = ""
+    new_kernel_version: str = ""
+    is_success: bool = False
+    error_message: str = ""
+
+
 def _is_completed_status(status: TestStatus) -> bool:
     return status in [
         TestStatus.FAILED,
--- a/lisa/transformers/kernel_installer.py
+++ b/lisa/transformers/kernel_installer.py
@@ -6,10 +6,11 @@ import re
 from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional, Type, cast
 
-from assertpy import assert_that
+from assertpy.assertpy import assert_that
 from dataclasses_json import dataclass_json
 
-from lisa import schema
+from lisa import notifier, schema
+from lisa.messages import KernelBuildMessage
 from lisa.node import Node, quick_connect
 from lisa.operating_system import Posix, Ubuntu
 from lisa.secret import PATTERN_HEADTAIL, add_secret
@@ -73,6 +74,7 @@ class KernelInstallerTransformerSchema(schema.Transformer):
     installer: Optional[BaseInstallerSchema] = field(
         default=None, metadata=field_metadata(required=True)
     )
+    raise_exception: Optional[bool] = True
 
 
 class BaseInstaller(subclasses.BaseClassWithRunbookMixin):
@@ -101,6 +103,8 @@ class BaseInstaller(subclasses.BaseClassWithRunbookMixin):
 
 class KernelInstallerTransformer(Transformer):
     _information_output_name = "information"
+    _is_success_output_name = "is_success"
+
     _information: Dict[str, Any] = dict()
 
     @classmethod
@@ -120,6 +124,10 @@ class KernelInstallerTransformer(Transformer):
         assert runbook.connection, "connection must be defined."
         assert runbook.installer, "installer must be defined."
 
+        message = KernelBuildMessage()
+        build_sucess: bool = False
+        boot_success: bool = False
+
         node = quick_connect(runbook.connection, "installer_node")
 
         uname = node.tools[Uname]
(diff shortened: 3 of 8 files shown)


adityagesh:
Raised the PR to get early review comments

I will add a sample runbook with "kernel_installer"

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,202 @@
...
+
+# GitBisect Combinator is a loop that runs "expanded" phase
+# of runbook until the bisect is complete.
+# There can be any number of expanded phases, but the
+# GitBisectTestResult notifier should have on boolean output per
+# iteration.
+
+
+def with_remote_node(func: Callable[..., Any]) -> Callable[..., Any]:
+    def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
```
adityagesh: Type of self should be "GitBisectCombinator", however there is a cyclic dependency. How can this be solved?

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,121 @@
...
+    # def _validate_node(self) -> None:
+    #     node = self._get_remote_node()
+    #     assert node.test_connection(), "connection to remote node failed."
+    #     node.close()
+    #     node.tools[Git]
+
+    def _clone_source(self) -> None:
+        node = self._get_remote_node()
+        node.execute(
+            cmd=f"mkdir -p {SOURCE_PATH}", shell=True, sudo=True, expected_exit_code=0
```
LiliDeng: can we use mkdir tool here?

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,121 @@
...
+        )
+        node.execute(cmd=f"chmod 777 {SOURCE_PATH}", sudo=True, expected_exit_code=0)
```
LiliDeng: we can also use tool chmod

LiliDeng:
please rebase code from main branch, it will fix the check error.

lisa/tools/git.py
```diff
@@ -172,6 +173,20 @@ def apply(
         )
         result.assert_exit_code(message=f"failed on applying patches. {result.stdout}")
 
+    def bisect(
```
squirrelsc: Let's make it wrapper better to use like bisect_start, bisect_stop, bisect_good, bisect_bad.

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -303,7 +303,6 @@ def _internal_run(self) -> Dict[str, Any]:
         platform.deploy_environment(environment)
 
         resource_group_name = get_environment_context(environment).resource_group_name
-
```
squirrelsc: Remove the unchanged file.

squirrelsc:
This change is big, and contains multiple features. Please split to small PR and make every commit meaningful. For example, you can put git changes in one PR or commit. Runner changes to support expanded cleanup should be another PR.

lisa/combinators/git_bisect_combinator.py
squirrelsc: git.clone already includes the mkdir. it doesn't need to added here.

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,202 @@
...
+                    self._bisect_bad()
+            else:
+                raise LisaException(
+                    f"Result missing for interation {iteration}, {results}"
+                )
+
+    @with_remote_node
+    def _clone_source(self, node: Node) -> None:
+        # node = self._get_remote_node()
+        node.execute(
```
squirrelsc: Don't use the hard code path, use a folder under `node.working_path`

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,202 @@
...
+from lisa.messages import KernelBuildMessage, TestResultMessage, TestStatus
+
+SOURCE_PATH = Path("/mnt/code")
+STOP_PATTERNS = ["first bad commit", "This means the bug has been fixed between"]
+
+
+# Combinator requires a node to clone the source code.
+@dataclass_json()
+@dataclass
+class GitBisectCombinatorSchema(schema.Combinator):
```
squirrelsc, line 24: Does it need to input `gitref`?

lisa/transformers/kernel_installer.py
```diff
@@ -100,6 +102,9 @@ def install(self) -> str:
 
 class KernelInstallerTransformer(Transformer):
     _information_output_name = "information"
+    _build_success_output_name = "build_success"
```
squirrelsc: `build_status`

squirrelsc:
Can you give an example runbook in description, so it's easier to understand how it works? And please also update document to introduce the new combinator.

lisa/messages.py
```diff
@@ -262,6 +262,16 @@ class ProvisionBootTimeMessage(MessageBase):
     information: Dict[str, str] = field(default_factory=dict)
 
 
+@dataclass
+class KernelBuildMessage(MessageBase):
+    type: str = "KernelBuild"
+    old_kernel_version: str = ""
+    new_kernel_version: str = ""
+    build_sucess: bool = False
```
squirrelsc: It looks build/boot not used separatedly. Merge them into one result.

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,202 @@
...
+    @classmethod
+    def type_name(cls) -> str:
+        return ""
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return schema.Notifier
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._iteration = 0
```
squirrelsc: The iteration looks not useful much. You can use a result field with triple status like `result: Optional[bool]`. If it's None, it means there is no result receive. True/False present the test result.

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,202 @@
...
+        return self._iteration
+
+    def _received_message(self, message: messages.MessageBase) -> None:
+        if isinstance(message, messages.TestResultMessage):
+            self._update_test_result(message)
+        elif isinstance(message, messages.KernelBuildMessage):
+            self._update_result(message.build_sucess)
+            self._update_result(message.boot_sucess)
+        else:
+            self._log.error("Received unsubscribed message type")
```
squirrelsc: Raise exception.

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,202 @@
...
+                else:
+                    self._bisect_bad()
+            else:
+                raise LisaException(
+                    f"Result missing for interation {iteration}, {results}"
+                )
+
+    @with_remote_node
+    def _clone_source(self, node: Node) -> None:
+        # node = self._get_remote_node()
```
squirrelsc: Remove the commented out code.

lisa/messages.py
adityagesh: I separated them out since it may be useful to identify stage of failure in general.
Can you suggest a good name for the status variable? Should it simply be "status"?

lisa/transformers/kernel_installer.py
adityagesh: The reason behind using "build_succcess" is because the value is True/False.
I felt it's easier to understand build_succcess -> False means build failed

Does "build_status" -> False convey the same meaning?

lisa/combinators/git_bisect_combinator.py
adityagesh: The reason for creating a directory (with sudo) and changing permission before clone was to ensure the clone can happen without sudo.
But I guess, changing the clone dir to node.working_path should fix this issue

lisa/combinators/git_bisect_combinator.py
adityagesh: -- update
This code will probably be removed due to https://github.com/microsoft/lisa/pull/2862#discussion_r1276542134

lisa/combinators/git_bisect_combinator.py
adityagesh: @squirrelsc Can you help here

lisa/tools/git.py
adityagesh: Bisect has too many options like 'bisect log', 'bisect reset' etc 
We will need to create too many functions here, Do you still recommend having one function each for 'start', 'reset', 'good', 'bad', 'log', 'skip' etc

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,183 @@
...
+        next: Optional[Dict[str, Any]] = None
+        self._process_result()
+        if not self._check_bisect_complete():
+            next = {}
+            next["ref"] = self._get_current_commit_hash()
+        self._result_notifier.result = None
+        return next
+
+    def _process_result(self) -> None:
+        if self._result_notifier.result is not None:
```
squirrelsc, line 96: raise exception, if it's `None`.

adityagesh: Updated. Add "iteration" in the combinator to skip result check at start of first phase

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,189 @@
...
+        if self._iteration == 0:
+            return
+        if self._result_notifier.result is not None:
+            results = self._result_notifier.result
+            if results:
+                self._bisect_good()
+            else:
+                self._bisect_bad()
+        else:
+            raise LisaException("Notifier result should not be None")
```
squirrelsc: The message should be actionable, and more meaningful. Imaging after 3 months, the error is raised, how can you know happend before looking at the code. For example, "bisect combinator doesn't get results before starting next bisect check."

adityagesh: Updated

lisa/tools/git.py
squirrelsc: The methods make it easy to use. The logic is here anyway. To avoid create too many methods in Git tool. You can create a bisect class, and put all methods inside.

squirrelsc:
> Can you give an example runbook in description, so it's easier to understand how it works? And please also update document to introduce the new combinator.

I mean to provide an example runbook in PR description, not a separated file. Maybe put it in document is better.

lisa/messages.py
squirrelsc: The values are used together. And if there is any error, it should be reflected from the error message. So, they can be merged like `is_success`.

adityagesh: Updated

microsoft/runbook/examples/git_bisect.yml
```diff
@@ -0,0 +1,118 @@
...
+combinator:
+  type: git_bisect
+  repo: $(repo_url)
+  bad_commit: $(bad_commit)
+  good_commit: $(good_commit)
+  connection:
+    address: $(bisect_vm_address)
+    private_key_file: $(admin_private_key_file)
+
+platform:
```
squirrelsc: Reference the azure yaml, instead of copy the settings.

microsoft/runbook/examples/git_bisect.yml
```diff
@@ -0,0 +1,118 @@
...
+  - type: azure
+    admin_private_key_file: $(admin_private_key_file)
+    azure:
+      subscription_id: $(subscription_id)
+    requirement:
+      azure:
+        location: $(location)
+        vhd: $(vhd)
+
+testcase:
```
squirrelsc: Refer to tier yamls.

microsoft/runbook/examples/git_bisect.yml
```diff
@@ -0,0 +1,118 @@
...
+    raise_exception: False
+  # Do not create vhd when build fails
+  - type: azure_vhd
+    enabled: $(kernel_installer_is_success)
+    phase: expanded
+    resource_group_name: $(build_vm_resource_group_name)
+    rename:
+      azure_vhd_url: vhd
+  # Do not run test when build fails
+  - type: script
```
squirrelsc: It can reuse the $(kernel_installer_is_success) variable to set `enabled`. It doesn't need to use script to clean the test case name.

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,196 @@
...
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.result: Optional[bool] = None
+
+    def _received_message(self, message: messages.MessageBase) -> None:
+        if isinstance(message, messages.TestResultMessage):
+            self._update_test_result(message)
+        elif isinstance(message, messages.KernelBuildMessage):
+            self._update_result(message.is_success)
+        else:
+            raise LisaException("Received unsubscribed message type")
```
squirrelsc: Print out type together.

lisa/combinators/git_bisect_combinator.py
squirrelsc: use `"GitBisectCombinator"`

adityagesh: image

adityagesh: The definition of with_remote_node is before GitBisectCombinator
and methods in GitBisectCombinator is using with_remote_node

adityagesh: Should I create another interface with just "_get_remote_node" and make that as the type for self?

lisa/tools/git.py
adityagesh: https://github.com/microsoft/lisa/pull/2984

microsoft/runbook/examples/git_bisect.yml
adityagesh: I did not understand.
Are you suggesting to keep default value for test_case_name as "" and set it to correct value by using `enabled: kernel_installer_is_success` instead of if else condition in script? `script: "test_case_name if kernel_installer_is_success else ''"`

lisa/combinators/git_bisect_combinator.py
squirrelsc: It's a decorator, so it runs very early. In this case, use `Any` is ok.

microsoft/runbook/examples/git_bisect.yml
squirrelsc: You can control which test cases to run by `enabled` flag in `TestCase`. So it doesn't need to control test case names by the script transformer.

squirrelsc (commented):
Forgot to publish comments.

squirrelsc:
I think we need to update document to explain how to use this transformer. PR may not be easy to find.

microsoft/runbook/examples/git_bisect.yml
adityagesh: Currently I am including tiers.yml
Should I add the enabled property to t0.yml, t1.yml etc?
Or is it better to define the test case block within the sample runbook?

```yml
testcase:
  - criteria:
      name: $(test_case_name)
enabled: $(kernel_installer_is_success)   
```

lisa/combinators/git_bisect_combinator.py
adityagesh: There is a parameter warning for Every function that uses the decorator 
image

adityagesh: Should I ignore E1120?

squirrelsc: I suggest not to be complex, just save a node object in the combinator and use it in every method.

microsoft/runbook/examples/git_bisect.yml
squirrelsc: The tier files doesn't have enabvled flag, you can add to all them. The name can be like `enable_tests`. Here you can direct output the same name.

adityagesh: https://github.com/microsoft/lisa/pull/3014/files

lisa/combinators/git_bisect_combinator.py
adityagesh: Removed decorator

squirrelsc:
@LiliDeng LGTM

adityagesh:
I will squash all commits to a single commit if no more changes are required.

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,191 @@
...
+        if self._result_notifier.result is not None:
+            results = self._result_notifier.result
+            if results:
+                self._bisect_good()
+            else:
+                self._bisect_bad()
+        else:
+            raise LisaException(
+                "Bisect combinator does not get result for next iteration. Please check"
+                "GitBisectResult notifier."
```
LiliDeng: " GitBisectResult notifier." leave a space between `check` and `GitBisectResult `

lisa/messages.py
```diff
@@ -263,6 +263,15 @@ class ProvisionBootTimeMessage(MessageBase):
     information: Dict[str, str] = field(default_factory=dict)
 
 
+@dataclass
+class KernelBuildMessage(MessageBase):
+    type: str = "KernelBuild"
+    old_kernel_version: str = ""
+    new_kernel_version: str = ""
+    is_success: bool = False
+    error_message: str = ""
```
LiliDeng, line 272: should we keep the first bad commit?

adityagesh: This message is used by Kernel Installer, The bad commit is detected by the Combinator.

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,191 @@
...
+    def _get_current_commit_hash(self) -> str:
+        node = self._get_remote_node()
+        git = node.tools[Git]
+        result = git.get_current_commit_hash(cwd=self._source_path)
+        node.close()
+        return result
+
+
+class GitBisectResult(notifier.Notifier):
+    @classmethod
```
LiliDeng, line 157: @squirrelsc I encountered `2023-10-26 07:01:34.979[904][ERROR] lisa. registered [] subclass again. It should happen in UT only. new: [<class 'lisa.runner.RunnerResult'>], exist: [<class 'lisa.combinators.git_bisect_combinator.GitBisectResult'>]` in today's testing, is it the previous PR wrong?

lisa/combinators/git_bisect_combinator.py
```diff
@@ -0,0 +1,191 @@
...
+    def type_name(cls) -> str:
+        return ""
```
LiliDeng: return "git_bisect_result"

LiliDeng (approved):

lisa/combinators/git_bisect_combinator.py
squirrelsc: @LiliDeng, please add a work item to set type name for RunnerResult, and assert empty when loading subclasses.

LiliDeng: > @LiliDeng, please add a work item to set type name for RunnerResult, and assert empty when loading subclasses.

Added.

PR #2863, opened 2023-07-27
Fix nvme basic testcase for FreeBSD

dsrivastavv:


lisa/features/nvme.py
```diff
@@ -43,32 +30,13 @@ def enabled(self) -> bool:
...
 
     def get_devices(self) -> List[str]:
-        devices_list = []
-        self._get_device_from_ls()
-        for row in self._ls_devices.splitlines():
-            matched_result = self._device_pattern.match(row)
-            if matched_result:
-                devices_list.append(matched_result.group("device_name"))
-        return devices_list
+        return self._node.tools[Ls].get_nvme_devices()
```
squirrelsc: `Ls` should focus it's own logic on "list files". The nvme logic shouldn't be put in ls.

lisa/tools/lsblk.py
```diff
@@ -10,13 +10,15 @@
 from lisa.util import LisaException, find_patterns_groups_in_lines
 
 
-def _get_size_in_bytes(size: int, size_unit: str) -> int:
+def _get_size_in_bytes(size: float, size_unit: str) -> int:
+    if size_unit == "T":
+        return int(size * 1024 * 1024 * 1024 * 1024)
```
squirrelsc, line 15: Why convert it to int?

lisa/tools/lsblk.py
```diff
@@ -246,7 +248,7 @@ def find_mountpoint_by_volume_name(
 class BSDLsblk(Lsblk):
     # da1p1          0:103  12G freebsd-ufs                                       - /mnt/resource   # noqa: E501
     _ENTRY_REGEX = re.compile(
-        r"\s*(?P<name>\S+)\s+\d+:\d+\s+(?P<size>\d+)(?P<size_unit>\w+)\s+"
+        r"\s*(?P<name>\S+)\s+\d+:\d+\s+(?P<size>\d+|\d+.\d+)(?P<size_unit>\w+)\s+"
```
squirrelsc, line 252: Add examples above.

lisa/tools/nvmecli.py
```diff
@@ -87,3 +95,39 @@ def support_device_format(self, device_name: str) -> bool:
...
+        namespaces_cli = []
+        nvme_list = self.run("list", shell=True, sudo=True)
+        for row in nvme_list.stdout.splitlines():
+            matched_result = self._namespace_cli_pattern.match(row)
+            if matched_result:
+                namespaces_cli.append(matched_result.group("namespace"))
+        return namespaces_cli
+
+
+class BSDNvmecli(Nvmecli):
```
squirrelsc, line 109: Does it use the same installation as Linux?

dsrivastavv: It's preinstalled in BSD

squirrelsc: You can set can_install to false to prevent installation logic reused.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2865, opened 2023-07-28
verify_serial_console_is_enabled: Verify tty0 in journalctl log

adityagesh:
Check for journalctl log before failing the test if log files are not found in /var/log

### DIFF
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -36,7 +36,7 @@ from lisa.operating_system import (
 )
 from lisa.sut_orchestrator import AZURE, READY
 from lisa.sut_orchestrator.azure.features import AzureDiskOptionSettings
-from lisa.tools import Cat, Dmesg, Ls, Lsblk, Lscpu, Pgrep, Ssh
+from lisa.tools import Cat, Dmesg, Journalctl, Ls, Lsblk, Lscpu, Pgrep, Ssh
 from lisa.util import (
     LisaException,
     PassedException,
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -763,17 +763,32 @@ def verify_serial_console_is_enabled(self, node: Node) -> None:
...
                 log_file = "/var/log/messages"
+                log_output = cat.read(log_file, force_run=True, sudo=True)
             elif node.shell.exists(node.get_pure_path("/var/log/syslog")):
                 log_file = "/var/log/syslog"
+                log_output = cat.read(log_file, force_run=True, sudo=True)
             else:
+                log_file = "journalctl"
+                journalctl = node.tools[Journalctl]
+                journalctl_out = journalctl.run(
+                    "-b | head -n 1000",
```
LiliDeng: how we can make sure the str can be found in first 1000 lines

adityagesh: I just tried for few distro's and it passed. Not sure if there is any way to guarantee. 
Without the head I thought the output would be too large. Do you recommend removing head -n 1000?

LiliDeng: if the vm was reused by many cases, probably the serial console config output is not within 1000 lines
if it is too large, can we try to grep it from the file? if it doesn't have a file. we can redirect the output into the file.

adityagesh: @LiliDeng , but with -b option in journalctl, we will get logs starting from previous boot.
So, running multiple cases should not matter, right?

LiliDeng: yes, journalctl -b means the recent reboot.
we can move this into journalctl tool and make the line number as a parameter, then get the output from one method in journalctl

LiliDeng (approved):

PR #2866, opened 2023-07-28
Sockperf: add sockperf network tests

mcgov:
- Supports most linux and BSD
- required changes to gcc tool to handle g++ install on some platforms
- required change to Make tool to handle make/gmake naming on BSD


lisa/tools/gcc.py
```diff
@@ -50,5 +50,26 @@ def get_version(self) -> VersionInfo:
 
     def _install(self) -> bool:
         posix_os: Posix = cast(Posix, self.node.os)
-        posix_os.install_packages("gcc")
+        if isinstance(posix_os, BSD):
```
squirrelsc: Do you need to proceed BSD now, or leaving it later?

lisa/tools/gcc.py
```diff
@@ -50,5 +50,26 @@ def get_version(self) -> VersionInfo:
...
+            node_os.install_packages("g++")
+        elif (
+            isinstance(node_os, Fedora)
+            or isinstance(node_os, Suse)
+            or isinstance(node_os, CBLMariner)
+        ):
+            node_os.install_packages("gcc-c++")
+        elif isinstance(node_os, BSD):
+            # BSD installs g++ with lang/gcc
+            return
```
squirrelsc: Leave a log line, so we know what happened. If log works, it doesn't need code comment.

lisa/tools/gcc.py
```diff
@@ -50,5 +50,26 @@ def get_version(self) -> VersionInfo:
...
+        else:
+            raise NotImplementedError(
```
squirrelsc: Wrap an OS unsupported excpetion.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -278,6 +280,130 @@ def perf_udp_iperf_sriov(self, result: TestResult) -> None:
             udp_mode=True,
         )
 
+    @TestCaseMetadata(
+        description="""
+        This test case uses sockperf to test sriov network latency.
+        """,
+        priority=2,
```
squirrelsc: Make them p3 by default.

lisa/tools/gcc.py
mcgov: going to remove it since gcc build of sockperf doesn't work, a c/c++ compiler clang is shipped by default for freebsd at least so that code isn't needed here.

lisa/tools/gcc.py
mcgov: removing, see other comment

lisa/tools/sockperf.py
```diff
@@ -0,0 +1,162 @@
...
+sockperf: ---> percentile 99\.000 =\s+(?P<latency99_percentile_us>[0-9]+\.[0-9]+)\s+\
+sockperf: ---> percentile 90\.000 =\s+(?P<latency_us_90>[0-9]+\.[0-9]+)\s+\
+sockperf: ---> percentile 75\.000 =\s+(?P<latency_us_75>[0-9]+\.[0-9]+)\s+\
+sockperf: ---> percentile 50\.000 =\s+(?P<latency_us_50>[0-9]+\.[0-9]+)\s+\
+sockperf: ---> percentile 25\.000 =\s+(?P<latency_us_25>[0-9]+\.[0-9]+)\s+\
+sockperf: ---> <MIN> observation =\s+(?P<min_latency_us>[0-9]+\.[0-9]+)",
+    )
+
+    def _get_protocol_flag(self, mode: str) -> str:
+        assert (
```
squirrelsc: use assert_that like `in`, so if it's unexpected value, it will be printed out in the log line.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -278,6 +280,130 @@ def perf_udp_iperf_sriov(self, result: TestResult) -> None:
...
+        ),
+    )
+    def perf_sockperf_latency_udp_synthetic(self, result: TestResult) -> None:
+        perf_sockperf(result, SOCKPERF_UDP, "perf_sockperf_latency_udp_synthetic")
+
+    @TestCaseMetadata(
+        description="""
+        This test case uses lagscope to test synthetic network latency.
+        """,
+        priority=2,
```
squirrelsc: make priority to 3

mcgov:
edit: resolved, tests passing :)

mcgov:
> one lingering bug in the mariner run, debugging...

should be resolved now. Testing full run of 10 images.

mcgov:
@squirrelsc  @LiliDeng should be good to review/merge now

microsoft/testsuites/performance/networkperf.py
squirrelsc: Still P2

microsoft/testsuites/performance/networkperf.py
```diff
@@ -278,6 +280,130 @@ def perf_udp_iperf_sriov(self, result: TestResult) -> None:
...
+            min_count=2,
+            network_interface=Synthetic(),
+        ),
+    )
+    def perf_sockperf_latency_tcp_synthetic_busy_poll(self, result: TestResult) -> None:
+        perf_sockperf(
+            result,
+            SOCKPERF_TCP,
+            "perf_sockperf_latency_tcp_synthetic_busy_poll",
+            set_busy_poll=True,
```
LiliDeng, line 407: Do we need check set_busy_poll before case, and restore them into origin values after testing?

microsoft/testsuites/performance/common.py
```diff
@@ -494,6 +497,40 @@ def perf_iperf(
...
+
+    if set_busy_poll:
+        for node in [client, server]:
+            node.tools[Sysctl].enable_busy_polling("50")
+
+    run_in_parallel([lambda: client.tools[Sockperf], lambda: server.tools[Sockperf]])
+
+    # disable any firewalls running which might mess with the test
+    for firewall in [client.tools[Firewall], server.tools[Firewall]]:
+        firewall.stop()
```
LiliDeng: can we move this to sockperf  tool, if it is needed when use this tool

microsoft/testsuites/performance/networkperf.py
mcgov: We sure do! Though I wonder how many of these other network perf tests should be running with it enabled  🤔

mcgov:
introduced a bug, fixing, don't merge

mcgov:
Fixed bug, running longer test with sockperf ntttcp and lagscope suites to check for regressions after the cleanup fixes.

microsoft/testsuites/performance/common.py
mcgov: Works for me :) fixed

microsoft/testsuites/performance/networkperf.py
mcgov: Should be fixed, I added a reset function to Sysctl. There are a lot of places this tool is used where users are doing manual tracking and cleanup, I didn't switch those all over to this function yet.

LiliDeng (approved):

PR #2873, opened 2023-07-31
Added testcase "verify_disk_controller_type"

SRIKKANTH:
Added testcase "verify_disk_controller_type"

### DIFF
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -80,7 +80,6 @@ class Disk(Feature):
 
     # Get disk controller type from the VM by checking the boot partition
     def get_os_disk_controller_type(self) -> schema.DiskControllerType:
-        os_disk_controller_type: schema.DiskControllerType = ""
         boot_partition = self.get_os_boot_partition()
         assert boot_partition, "'boot_partition' must not be 'None'"
 
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -80,7 +80,7 @@ class Disk(Feature):
     # Get disk controller type from the VM by checking the boot partition
     def get_os_disk_controller_type(self) -> schema.DiskControllerType:
         boot_partition = self.get_os_boot_partition()
-        assert boot_partition, "'boot_partition_disk' must not be 'None'"
+        assert boot_partition, "'boot_partition' must not be 'None'"
         if boot_partition.disk == "nvme":
             return schema.DiskControllerType.NVME
         else:
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -10,6 +10,7 @@ from lisa import schema
 from lisa.feature import Feature
 from lisa.tools import Mount
 from lisa.tools.mount import PartitionInfo
+from lisa.util import LisaException
 
 
 class Disk(Feature):
@@ -79,12 +80,17 @@ class Disk(Feature):
 
     # Get disk controller type from the VM by checking the boot partition
     def get_os_disk_controller_type(self) -> schema.DiskControllerType:
+        os_disk_controller_type: schema.DiskControllerType = ""
         boot_partition = self.get_os_boot_partition()
         assert boot_partition, "'boot_partition' must not be 'None'"
-        if boot_partition.disk == "nvme":
-            return schema.DiskControllerType.NVME
+
+        if boot_partition.disk.startswith("nvme"):
+            os_disk_controller_type = schema.DiskControllerType.NVME
+        elif boot_partition.disk.startswith("sd"):
+            os_disk_controller_type = schema.DiskControllerType.SCSI
         else:
-            return schema.DiskControllerType.SCSI
+            raise LisaException(f"Unknown OS boot disk type {boot_partition.disk}")
+        return os_disk_controller_type
 
 
 DiskEphemeral = partial(
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -79,9 +79,9 @@ class Disk(Feature):
 
     # Get disk controller type from the VM by checking the boot partition
     def get_os_disk_controller_type(self) -> schema.DiskControllerType:
-        boot_partition_disk = self.get_os_boot_partition()
-        assert boot_partition_disk, "'boot_partition_disk' must not be 'None'"
-        if boot_partition_disk.disk == "nvme":
+        boot_partition = self.get_os_boot_partition()
+        assert boot_partition, "'boot_partition_disk' must not be 'None'"
+        if boot_partition.disk == "nvme":
             return schema.DiskControllerType.NVME
         else:
             return schema.DiskControllerType.SCSI
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -67,7 +67,7 @@ class Disk(Feature):
     def get_resource_disk_mount_point(self) -> str:
         raise NotImplementedError
 
-    # Get disk controller type from the VM by checking the boot partition
+    # Get boot partition of VM by looking for "/boot" and "/boot/efi"
     def get_os_boot_partition(self) -> Optional[PartitionInfo]:
         partition_info = self._node.tools[Mount].get_partition_info()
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/core/storage.py
```diff
@@ -225,6 +230,39 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+        # Get VM's 'disk controller type' with azure api
+        vm_disk_controller_type = node.features[Disk].get_disk_controller_type()
+
+        # Get 'disk controller type' from within VM.
+        os_disk_controller_type = node.features[Disk].os_disk_controller_type()
+
+        # With certain SKUs & gen1 images 'disk_controller_type' will be 'None'
+        if not vm_disk_controller_type:
+            raise SkippedException(
+                f"VM disk_controller_type is '{vm_disk_controller_type}'"
```
adityagesh: Can this message be improved?
It would display "VM disk_controller_type is None" as reason for being skipped

microsoft/testsuites/core/storage.py
```diff
@@ -225,6 +230,39 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+            )
+
+        if os_disk_controller_type != vm_disk_controller_type:
```
adityagesh: We can use assert_that(  ).is_equal_to( ) here

lisa/features/disks.py
```diff
@@ -64,6 +67,21 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+    # Get disk controller type from the VM by checking the boot partition
+    def os_disk_controller_type(self) -> schema.DiskControllerType:
+        partition_info = self._node.tools[Mount].get_partition_info()
+
+        # On certain gen2 image only "/boot/efi" exists
+        for partition in partition_info:
+            if partition.mount_point in ("/boot", "/boot/efi"):
+                boot_partition_disk = partition.disk
+                break
+
```
adityagesh: Can this be split into two functions?
get_boot_partition() and get_os_disk_controller_type()

get_os_disk_controller_type() calls get_boot_partition() to get the boot partition details

adityagesh: or get_boot_partition can be a method of the Mount tool.

Thoughts on this @LiliDeng ?

adityagesh (changes requested):

lisa/features/disks.py
SRIKKANTH: I can do that if its really a useful method. @LiliDeng please share your thoughts on this.

LiliDeng: yes, you can add it by looking for .startswith("/boot")

adityagesh (changes requested):
@SRIKKANTH 
Can you modify https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/features.py#L1271-L1273 to get disks based on the controller type

adityagesh (changes requested):
Similarly https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/features.py#L1212-L1247
is using scsi pattern

adityagesh:
The above two comments can be part of different PR if required.

microsoft/testsuites/core/storage.py
```diff
@@ -225,6 +230,40 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+
+        # Get VM's 'disk controller type' with azure api
+        vm_disk_controller_type = node_disk.get_disk_controller_type()
+
+        # Get 'disk controller type' from within VM.
+        os_disk_controller_type = node_disk.get_os_disk_controller_type()
+
+        # With certain SKUs & gen1 images 'disk_controller_type' will be 'None'
+        if not vm_disk_controller_type:
+            raise SkippedException(
```
squirrelsc, line 259: This description use the `vm_disk_controller_type`, but it's None in this case. So use "None" in the error message will be clearer. And "vm_disk_controller_type" is the variable name, not the configuration on VM. So it should give clearer message, like,

"The VM size doesn't have disk controller type information."

microsoft/testsuites/core/storage.py
```diff
@@ -225,6 +230,40 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+                f"VM disk_controller_type is '{vm_disk_controller_type}'."
+                f"Few SKUs& gen1 images doesn't support 'disk_controller_type'"
+            )
+
+        assert_that(
+            os_disk_controller_type,
+            "VM and OS disk_controller_types are different"
```
squirrelsc: It should describe as the expected results, or wrong results. The "are different" doesn't say it's expected or not. update it like, "The disk controller types of VM and OS should be the same."

lisa/features/disks.py
```diff
@@ -47,6 +47,9 @@ def get_raw_data_disks(self) -> List[str]:
     def get_all_disks(self) -> List[str]:
         raise NotImplementedError
 
+    def get_disk_controller_type(self) -> schema.DiskControllerType:
```
squirrelsc: Differentiate the name more like "get_hardware_disk_controller_type".

SRIKKANTH: Method changed to "get_hardware_disk_controller_type"

microsoft/testsuites/core/storage.py
SRIKKANTH: Updated the failure message to "The disk controller types of VM and OS should be the same."

microsoft/testsuites/core/storage.py
SRIKKANTH: Updated the skip message to "The VM size doesn't have disk controller type information."

lisa/features/disks.py
SRIKKANTH: Added a method "get_os_boot_partition()"

microsoft/testsuites/core/storage.py
```diff
@@ -227,6 +232,40 @@ def verify_resource_disk_io(self, node: RemoteNode) -> None:
...
+        # With certain SKUs & gen1 images 'disk_controller_type' will be 'None'
+        if not vm_disk_controller_type:
+            raise SkippedException(
+                "The VM size doesn't have disk controller type information."
+                "Few VM Sizes and gen1 images doesn't support 'disk_controller_type'"
+            )
+
+        assert_that(
+            os_disk_controller_type,
+            "The disk controller types of VM and OS should be the same.",
```
squirrelsc: vm and os is confusing. Maybe call it like Configured/Runtime or Hardware/OS?

SRIKKANTH: How about hardware and OS ?

squirrelsc: Yes, it's ok to me.

lisa/features/disks.py
```diff
@@ -64,6 +67,25 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+                break
+        return boot_partition
+
+    # Get disk controller type from the VM by checking the boot partition
+    def get_os_disk_controller_type(self) -> schema.DiskControllerType:
+        boot_partition = self.get_os_boot_partition()
+        assert boot_partition, "'boot_partition' must not be 'None'"
+        if boot_partition.disk == "nvme":
+            return schema.DiskControllerType.NVME
+        else:
```
squirrelsc, line 90: Only set to scsi, if it is. For other cases, raise exception unless we know how it happens.

microsoft/testsuites/core/storage.py
SRIKKANTH: Updated the message with "hardware and OS"

lisa/features/disks.py
SRIKKANTH: Added condition check for SCSI and exception for others.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
I'll check and merge in proper time.

PR #2876, opened 2023-08-02
added an additional validation

arisettisanjana:
Added an additional validation for vmsnapshot_extension
To ensure that the snapshot taken was not Crash consistent, and also validating the status of restore point created.

### DIFF
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -150,8 +150,7 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
             else:
                 log.info(f"rp status is {response.provisioning_state}")
                 attempts += 1
-                if attempts < max_attempts:
-                    time.sleep(2)
+                time.sleep(2)
         raise ValueError(
             "Restore point provisioning status not Succeeded "
             "after multiple attempts."
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -108,7 +108,7 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
                     pass
                 else:
                     raise e
-            time.sleep(5)
+            time.sleep(1)
             count = count + 1
         assert_that(count, "Restore point creation failed.").is_less_than(10)
 
@@ -125,7 +125,7 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
         assert isinstance(platform, AzurePlatform)
         client = get_compute_client(platform)
         attempts = 0
-        max_attempts = 15
+        max_attempts = 450
         while attempts < max_attempts:
             response = client.restore_points.get(
                 resource_group_name=resource_group_name,
@@ -151,7 +151,7 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
                 log.info(f"rp status is {response.provisioning_state}")
                 attempts += 1
                 if attempts < max_attempts:
-                    time.sleep(60)
+                    time.sleep(2)
         raise ValueError(
             "Restore point provisioning status not Succeeded "
             "after multiple attempts."
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -37,7 +37,8 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
     @TestCaseMetadata(
         description="""
         Create a restore point collection for the virtual machine.
-        Create application consistent restore point on the restore point collection.
+        Create application consistent restore point on the restore point
+        collection.
         Validate response of the restore point for validity.
         Attempt it a few items to rule out cases when VM is under changes.
         """,
@@ -89,7 +90,14 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
                     parameters={},
                 )
                 wait_operation(response, time_out=600)
-                log.info(f"restore point {restore_point} created")
+                # check the status of rp and validate the result.
+                self.get_restore_point(
+                    log,
+                    environment,
+                    resource_group_name,
+                    restore_point_collection,
+                    restore_point,
+                )
                 break
             except Exception as e:
                 # Changes were made to the Virtual Machine, while the operation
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -81,16 +82,34 @@ def verify_vmsnapshot_extension(
...
+                    resource_group_name=resource_group_name,
+                    restore_point_collection_name=restore_point_collection,
+                    restore_point_name=restore_point,
+                    expand=None)
+                if response.provisioning_state == "Succeeded":
+                    log.info(f"restore point {restore_point} created")
+                    consistency_mode = response.consistency_mode
+                    log.info(
+                        f"consistency mode is {consistency_mode}")
+                    if "File" in consistency_mode or "App" in consistency_mode:
```
iamashwani-who: I think we can use exact value match.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -81,16 +82,34 @@ def verify_vmsnapshot_extension(
...
+                        break
+                    else:
+                        raise ValueError(
+                            "Consistency mode is not File or App")
```
iamashwani-who: FileSystemConsistent or ApplicationConsistent.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -81,16 +82,34 @@ def verify_vmsnapshot_extension(
...
+                else:
+                    log.info(f"rp status is {response.provisioning_state}")
```
iamashwani-who: if provisioning is not succeeded, check if we need to wait for some more time to check the completion status. The reason is that if it is in transitioning state, then subsequent create RP calls will fail too. We can fail the test with message that VM is in transitioning state and test can not proceed?

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -100,6 +129,34 @@ def verify_vmsnapshot_extension(
...
+        client = get_compute_client(platform)
+        response = client.restore_points.get(
+            resource_group_name=resource_group_name,
+            restore_point_collection_name=restore_point_collection,
+            restore_point_name=restore_point,
+            expand=None)
+        if response.provisioning_state == "Succeeded":
+            log.info(f"restore point {restore_point} created")
+            consistency_mode = response.consistency_mode
+            log.info(
```
iamashwani-who: nit: line breaks appear inconsistent. line 161 has all logged content in single line but this line has line breaks. Simi,ar observation for L120, 117 and others.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -81,16 +83,43 @@ def verify_vmsnapshot_extension(
...
+                        break
+                    else:
+                        if return_value == 1:
+                            err_msg = (
+                                "Consistency mode is not FileSystemConsistent \
+                                    or ApplicationConsistent")
+                        else:
+                            err_msg = (
+                                "Provisioning status is not yet Succeeded")
+                        raise ValueError(err_msg)
```
iamashwani-who: repeated code. can be moved into a helper function.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: this was done to ensure that the length of the line is not exceeding certain characters, ig 79

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -100,6 +119,44 @@ def verify_vmsnapshot_extension(
...
+                if (
+                    "FileSystemConsistent" in consistency_mode
+                    or "ApplicationConsistent" in consistency_mode
+                ):
+                    return 0
+                else:
+                    return 1
+            else:
+                log.info(f"rp status is {response.provisioning_state}")
+                time.sleep(900)
```
LiliDeng: we can check status in a loop, exit the loop till meet some condition, instead of hard code wait time

arisettisanjana: So in this case we wanted to check the status of restore point creation after 15 minutes. and still if it did not succeed then we would be returning 2 to fail the test.

LiliDeng: we should check the status in a loop, set max wait time as 15 mins, when status is expected, then exit the loop.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -100,6 +118,45 @@ def verify_vmsnapshot_extension(
...
+                    "FileSystemConsistent" in consistency_mode
+                    or "ApplicationConsistent" in consistency_mode
+                ):
+                    return 0
+                else:
+                    return 1
+            else:
+                log.info(f"rp status is {response.provisioning_state}")
+                attempts += 1
+                if attempts < max_attempts:
```
LiliDeng: no need this step?

arisettisanjana: just to skip sleep for 1 min at last.
i.e during 14th attempt, if provisioning state is not succeeded then attempts would be incremented to 15, and sleep for a min and then check the exit criteria(while condition) and then simply exits.
So sleeping for a minute is like of no use at the end as we are not checking the status again, so I have put that check.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -100,6 +118,45 @@ def verify_vmsnapshot_extension(
                     pass
                 else:
                     raise e
-            time.sleep(1)
+            time.sleep(60)
```
squirrelsc: I prefer to try more times instead of sleep 1 min and try less. Even it needs 10 minutes, it should check around every 5 seconds.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -89,8 +90,25 @@ def verify_vmsnapshot_extension(
...
-                break
+                # check the status of rp and validate the result.
+                return_value = self.get_restore_point(
+                    log,
+                    environment,
+                    resource_group_name,
+                    restore_point_collection,
+                    restore_point,
+                )
+                if return_value == 0:
```
squirrelsc: The return value should be more meaningful, instead of a number. In this case, it looks like the method can raise different exceptions directly, if it needs to return 1 or 2. So it doesn't need to return values, if everything works fine.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -100,6 +118,45 @@ def verify_vmsnapshot_extension(
...
+        resource_group_name: str,
+        restore_point_collection: str,
+        restore_point: str,
+    ) -> int:
+        assert environment.platform
+        platform: AzurePlatform = environment.platform  # type: ignore
+        assert isinstance(platform, AzurePlatform)
+        client = get_compute_client(platform)
+        attempts = 0
+        max_attempts = 15
```
squirrelsc: use `retry` package, which has a lot of examples in LISA code base.

arisettisanjana: Modified the function itself to raise errors instead of returning the value.
and upon checking the retry package as far as i understood, we are retrying the execution of a function that resulted in an error.
As i am raising 2 different errors, and for one i don't want to retry and for the other i wish to retry so avoiding it here.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
squirrelsc: Cannot get the point here. Sleep 60 seconds is too long in any cases. It should check some conditions continually in a loop, once it meets, get out of the loop. The interval in the loop should be less than 5 seconds. So once the condition meets, it waits 5 seconds at most, instead of 60 seconds.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -100,6 +108,51 @@ def verify_vmsnapshot_extension(
                     pass
                 else:
                     raise e
-            time.sleep(1)
+            time.sleep(5)
```
squirrelsc: keep 1 second is better. If it needs to wait some condition, use a loop to check it.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
squirrelsc: You can specify an error/exception to retry, other exceptions will be raised at first time.

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
arisettisanjana: reduced the interval to 2 and increased the iteration number.
The above point I was mentioning is : why i have an extra if condition check. That is to skip sleeping for the specified amount of time during the last attempt.

LiliDeng (approved):

PR #2877, opened 2023-08-02
Consolidated MANA testing changes:

mcgov:
Consolidated MANA testing changes:
- Adds better MANA pmd arguments handling
- Adds option to force use of net_failsafe on newer platforms for failure mode testing
- Swaps from exclusion to inclusion of test nics for handling scenarios where all dpdk interfaces share a pci bus address
- Adds handling of MANA idiosyncracy of setting devices down before testing.
- Adds get_mac_address function for nic class to handle setting shared MAC devices down. 

Acknowledging this last point is a hack, but the addition of that function will lead to a larger fix soon. The get_mac_address function covers an issue in the NIC class where slave devices can go undetected, refactoring is needed to fix this. This get_mac_address function will be used in the eventual refactoring, so this backup check will eventually be moved into the parent class.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -437,6 +437,16 @@ class Nics(InitializableMixin):
                     self.nics[nic].pci_slot = pci_device.slot
                 break
 
+    def get_mac_address(self, nic_name: str) -> str:
+        return self._node.execute(
+            f"cat /sys/class/net/{nic_name}/address",
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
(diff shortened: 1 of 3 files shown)


lisa/nic.py
```diff
@@ -437,6 +437,16 @@ def _load_nics(self) -> None:
                     self.nics[nic].pci_slot = pci_device.slot
                 break
 
+    def get_mac_address(self, nic_name: str) -> str:
```
LiliDeng, line 440: where it is used?

PR #2883, opened 2023-08-03
Azure NodeContext: Add location and subscription_id

adityagesh:
Multiple extensions needs subscription_id and location. These are fetched from different sources. Add it to NodeContext to easily access it.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -13,6 +13,7 @@ from time import sleep
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 import requests
+from assertpy import assert_that
 from azure.mgmt.compute import ComputeManagementClient  # type: ignore
 from azure.mgmt.compute.models import VirtualMachine  # type: ignore
 from azure.mgmt.marketplaceordering import MarketplaceOrderingAgreements  # type: ignore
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1451,6 +1453,9 @@ def load_environment(
         node_context.username = platform_runbook.admin_username
         node_context.password = platform_runbook.admin_password
         node_context.private_key_file = platform_runbook.admin_private_key_file
+        node_context.location = get_vm(platform, node).location
```
LiliDeng: can we use vms_map?

adityagesh: updated

adityagesh: Changed cast to assert as well.

LiliDeng (approved):

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1444,13 +1447,19 @@ def load_environment(
     for node in environment.nodes.list():
         assert isinstance(node, RemoteNode)
 
+        vm = vms_map.get(node.name)
+        assert_that(vm).is_not_none()
```
squirrelsc, line 1451: Add error message, so it's easy to know what's wrong with the assertion failure.

PR #2884, opened 2023-08-04
ACC_BVT: various fixes, add logging in case of failure

mcgov:
fixing an old bug, bitwise & operator used for boolean comparison

### DIFF
--- a/microsoft/testsuites/acc/bvt.py
+++ b/microsoft/testsuites/acc/bvt.py
@@ -9,7 +9,7 @@ from lisa import (
     simple_requirement,
 )
 from lisa.features import acc
-from lisa.operating_system import Debian
+from lisa.operating_system import Ubuntu
 from lisa.tools import Dmesg, Make, Wget, Whoami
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/acc/bvt.py
```diff
@@ -183,7 +181,14 @@ def verify_sgx(self, log: Logger, node: Node) -> None:
...
+        if result.stderr:
+            log.info(f"sgx tool stderr had data: {result.stderr}")
+        assert_that(result.stdout).described_as(
+            "Decrypted data did not match the enclave internal secret data"
+        ).contains(trace_msg)
+        validation_count = result.stdout.count(trace_msg)
+        assert_that(validation_count).described_as(
+            "Both enclaves in test validate two messages. "
+            f"Only found {validation_count} success messages."
+        ).is_equal_to(4)
```
LiliDeng, line 194: why it must equal to 4?

LiliDeng (approved):

PR #2885, opened 2023-08-04
Git: clean up handling branch/tag/head checkout

mcgov:
Fixing an annoyance, passing branches and tags can result in a checkout failure. Rather than force everyone to think about this in-depth, just get the commit id for whatever users pass in.

If the branch/tag/head/commit is invalid, it will still fail in the usual way.

show-ref is present in git releases above at least 1.4, having a hard time finding older releases other than 0.8 which is not in use.


LiliDeng:
what failures when we use branch to checkout?

PR #2886, opened 2023-08-07
Azure Key Vault Linux Extension VM test case validation

t-ecastillo:
The following test case validates the Azure Key Vault Linux Extension while creating the following resources:
-A resource group
-A VM
-A Key Vault
-Two certificates in the Key Vault
-Retrieval of the certificate's secrets through SecretClient class from the Azure SDK.
-Installation of the Azure Key Vault Linux Extension on the VM.
-Rotation of the certificates (After KVVM Extension has been installed)
-Asserts after the creation/installation of all of these resources.
All of the resources have been created by using the Azure SDK Python.


microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -41,7 +41,7 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
         Validate response of the restore point for validity.
         Attempt it a few items to rule out cases when VM is under changes.
         """,
-        priority=1,
+        priority=6,
```
LiliDeng: why change this?

t-ecastillo: I was testing locally and wanted my test to go first, we can ignore this change.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -39,6 +41,10 @@
     Sku,
     StorageAccountCreateParameters,
 )
+from azure.identity import DefaultAzureCredential
+from azure.mgmt.compute import ComputeManagementClient
+from azure.keyvault.secrets import SecretClient
+
```
LiliDeng, line 47: if it is a new package, please add it here refer https://github.com/microsoft/lisa/blob/main/pyproject.toml#L42

t-ecastillo: Done! :)

PR #2891, opened 2023-08-07
Sockperf: fixes for centos 7.4 and 16.04

mcgov:
- Modify sockperf tool and test to account for old centos idiosyncrasies.
- Add skip for EOL Ubuntu, no packaged version and build is broken.

### DIFF
--- a/microsoft/testsuites/performance/common.py
+++ b/microsoft/testsuites/performance/common.py
@@ -8,7 +8,14 @@ from typing import Any, Dict, List, Optional, Union, cast
 from assertpy import assert_that
 from retry import retry
 
-from lisa import Environment, Node, RemoteNode, notifier, run_in_parallel
+from lisa import (
+    Environment,
+    Node,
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/performance/common.py
```diff
@@ -510,6 +518,8 @@ def perf_sockperf(
     client = cast(RemoteNode, environment.nodes[0])
     server = cast(RemoteNode, environment.nodes[1])
     sysctls: List[Sysctl] = []
+    if isinstance(client.os, Ubuntu) and (client.os.information.version < "18.4.0"):
+        raise SkippedException
```
LiliDeng, line 522: please add detailed skip reason

LiliDeng: @mcgov please address this in a new PR, I will merge it since there are some failures caused by this case every day.

LiliDeng (approved):

PR #2892, opened 2023-08-07
Test case Azure Key Vault Linux Extension VM

t-ecastillo:
The following test case validates the Azure Key Vault Linux Extension while creating the following resources:
-A resource group
-A VM
-A Key Vault
-Two certificates in the Key Vault
-Retrieval of the certificate's secrets through SecretClient class from the Azure SDK.
-Installation of the Azure Key Vault Linux Extension on the VM.
-Rotation of the certificates (After KVVM Extension has been installed)
-Asserts after the creation/installation of all of these resources.
All of the resources have been created by using the Azure SDK Python.

### DIFF
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -338,11 +338,7 @@
             "copy": {
                 "name": "vmCopy",
                 "count": "[variables('node_count')]"
-            },
-            "identity": {
-                "type": "SystemAssigned"
-            },
-                        
+            },         
             "name": "[parameters('nodes')[copyIndex('vmCopy')]['name']]",
             "location": "[parameters('nodes')[copyIndex('vmCopy')]['location']]",
             "tags": "[parameters('vm_tags')]",
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -332,6 +332,7 @@
                 }
             }
         },
+        
         {
             "apiVersion": "2022-08-01",
             "type": "Microsoft.Compute/virtualMachines",
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -339,6 +339,10 @@
                 "name": "vmCopy",
                 "count": "[variables('node_count')]"
             },
+            "identity": {
+                "type": "SystemAssigned"
+            },
+                        
             "name": "[parameters('nodes')[copyIndex('vmCopy')]['name']]",
             "location": "[parameters('nodes')[copyIndex('vmCopy')]['location']]",
             "tags": "[parameters('vm_tags')]",
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -338,7 +338,7 @@
             "copy": {
                 "name": "vmCopy",
                 "count": "[variables('node_count')]"
-            },         
+            },
             "name": "[parameters('nodes')[copyIndex('vmCopy')]['name']]",
             "location": "[parameters('nodes')[copyIndex('vmCopy')]['location']]",
             "tags": "[parameters('vm_tags')]",
--- a/lisa/sut_orchestrator/azure/arm_template.json
+++ b/lisa/sut_orchestrator/azure/arm_template.json
@@ -332,7 +332,6 @@
                 }
             }
         },
-        
         {
             "apiVersion": "2022-08-01",
             "type": "Microsoft.Compute/virtualMachines",
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
 import os
 import re
 import sys
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -10,12 +10,13 @@ from functools import lru_cache
 from pathlib import Path
 from threading import Lock
 from time import sleep
-import time
 from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
-from msrestazure import AzureConfiguration
 
 import requests
 from assertpy import assert_that
+from azure.identity import DefaultAzureCredential
+from azure.keyvault.certificates import CertificateClient, CertificatePolicy
+from azure.keyvault.secrets import SecretClient
 from azure.mgmt.compute import ComputeManagementClient  # type: ignore
 from azure.mgmt.compute.models import VirtualMachine  # type: ignore
 from azure.mgmt.marketplaceordering import MarketplaceOrderingAgreements  # type: ignore
@@ -42,10 +43,6 @@ from azure.mgmt.storage.models import (  # type: ignore
     Sku,
     StorageAccountCreateParameters,
 )
-from azure.identity import DefaultAzureCredential
-from azure.mgmt.compute import ComputeManagementClient
-from azure.keyvault.secrets import SecretClient
-
 from azure.storage.blob import (
     AccountSasPermissions,
     BlobClient,
@@ -80,10 +77,6 @@ from lisa.util import (
 from lisa.util.logger import Logger
 from lisa.util.parallel import check_cancelled
 from lisa.util.perf_timer import create_timer
-
-
-from azure.keyvault.certificates import CertificateClient, CertificatePolicy
-from azure.identity import DefaultAzureCredential
 
 if TYPE_CHECKING:
     from .platform_ import AzurePlatform
@@ -1917,8 +1910,13 @@ class DataDisk:
             raise LisaException(f"Data disk type {disk_type} is unsupported.")
 
 
-def create_certificates(vault_url: str, credential: DefaultAzureCredential, log: Logger, retries=5, delay=2):
-    
+def create_certificates(
+    vault_url: str,
+    credential: DefaultAzureCredential,
+    log: Logger,
+    retries: int = 5,
+    delay: int = 2,
+) -> List[str]:
     certificate_client = CertificateClient(vault_url=vault_url, credential=credential)
     secret_client = SecretClient(vault_url=vault_url, credential=credential)
 
@@ -1931,40 +1929,63 @@ def create_certificates(vault_url: str, credential: DefaultAzureCredential, log:
         for attempt in range(retries):
             try:
                 # Create certificate
-                create_certificate_result = certificate_client.begin_create_certificate(cert_name, policy=cert_policy)
-                certificate_client.update_certificate_properties(certificate_name=cert_name, enabled=True)
+                create_certificate_result = certificate_client.begin_create_certificate(
+                    cert_name, policy=cert_policy
+                )
+                log.info(
+                    f"Certificate has been created: "
+                    f"Result: {create_certificate_result}"
+                )
+                certificate_client.update_certificate_properties(
+                    certificate_name=cert_name, enabled=True
+                )
                 break
             except Exception as e:
                 if attempt < retries - 1:
                     sleep(delay)
                 else:
-                    log.error(f"Failed to create certificate named '{cert_name}' after {retries} retries.")
+                    log.error(
+                        f"Failed to create certificate named"
+                        f"{cert_name}' after {retries} retries."
+                        f"Error info {e}"
+                    )
                     raise
-            except Exception as e:
-                log.error(f"Unexpected error occurred while creating certificate named '{cert_name}': {e}")
-                raise
 
         try:
-            secret_id = secret_client.get_secret(name=cert_name).id
-            secret_url_without_version = secret_id.rsplit('/', 1)[0]
-            secret_urls.append(secret_url_without_version)
+            secret_id: Optional[str] = secret_client.get_secret(name=cert_name).id
+            if secret_id:
+                secret_url_without_version = secret_id.rsplit("/", 1)[0]
+                secret_urls.append(secret_url_without_version)
+            else:
+                log.error(
+                    f"Failed to retrieve secret ID for certificate named '{cert_name}'."
+                )
         except Exception as e:
-            log.error(f"Failed to retrieve secret ID for certificate named '{cert_name}': {e}")
+            log.error(
+                f"Failed to retrieve secret ID for certificate named '{cert_name}': {e}"
+            )
             raise
 
     return secret_urls
 
 
-def rotate_certificates(self, log: Logger, vault_url: str, credential: DefaultAzureCredential, cert_name_to_rotate: str) -> None:
-    
+def rotate_certificates(
+    self,
+    log: Logger,
+    vault_url: str,
+    credential: DefaultAzureCredential,
+    cert_name_to_rotate: str,
+) -> None:
     certificate_client = CertificateClient(vault_url=vault_url, credential=credential)
-    
+
     try:
         # Retrieve the old version of the certificate
-        old_certificate = certificate_client.get_certificate(cert_name_to_rotate)
-        log.info(f"Old version of certificate '{cert_name_to_rotate}': {old_certificate.properties.version}")
+        certificate_client.get_certificate(cert_name_to_rotate)
     except Exception as e:
-        log.error(f"Failed to retrieve old version of certificate named '{cert_name_to_rotate}': {e}")
+        log.error(
+            f"Failed to retrieve old version of certificate:"
+            f"{cert_name_to_rotate}': {e}"
+        )
         raise
 
     cert_policy = CertificatePolicy.get_default()
@@ -1973,36 +1994,57 @@ def rotate_certificates(self, log: Logger, vault_url: str, credential: DefaultAz
     for attempt in range(2):
         try:
             # Create certificate
-            create_certificate_poller = certificate_client.begin_create_certificate(cert_name_to_rotate, policy=cert_policy)
-            create_certificate_result = create_certificate_poller.result()  # Wait for completion
-            certificate_client.update_certificate_properties(certificate_name=cert_name_to_rotate, enabled=True)
+            create_certificate_poller = certificate_client.begin_create_certificate(
+                cert_name_to_rotate, policy=cert_policy
+            )
+            create_certificate_result = create_certificate_poller.result()
+
+            # Handle possible None value
+            if (
+                create_certificate_result
+                and hasattr(create_certificate_result, "properties")
+                and create_certificate_result.properties
+            ):
+                new_certificate_version = create_certificate_result.properties.version
+                log.info(
+                    f"New version of certificate '{cert_name_to_rotate}': "
+                    f"{new_certificate_version}"
+                )
+                log.info("Certificate rotated")
+            else:
+                log.error(
+                    "Failed to retrieve properties from create certificate result."
+                )
+
+            certificate_client.update_certificate_properties(
+                certificate_name=cert_name_to_rotate, enabled=True
+            )
             break
         except Exception as e:
             if attempt < 1:
                 sleep(1)
             else:
-                log.error(f"Failed to rotate certificate named '{cert_name_to_rotate}' after 2 attempts.")
+                log.error(
+                    f"Failed to rotate certificate:"
+                    f"{cert_name_to_rotate}' after 2 attempts."
+                    f"More exception info: {e}"
+                )
                 raise
-        except Exception as e:
-            log.error(f"Unexpected error occurred while rotating certificate named '{cert_name_to_rotate}': {e}")
-            raise
-
-    # Retrieve the new version of the certificate from the creation result
-    new_certificate_version = create_certificate_result.properties.version
-    log.info(f"New version of certificate '{cert_name_to_rotate}': {new_certificate_version}")
-    log.info("Certificate rotated")
 
 
 def check_system_status(node: Node, log: Logger) -> None:
     # Check the status of the akvvm_service service
-    akvvm_service_result = node.execute("systemctl status akvvm_service.service", sudo=True, timeout=10)
+    akvvm_service_result = node.execute(
+        "systemctl status akvvm_service.service", sudo=True, timeout=10
+    )
     log.info(f"akvvm_service status: {akvvm_service_result.stdout}")
-    # List the contents of the directory /var/lib/waagent/Microsoft.Azure.KeyVault
-    ls_result = node.execute("sudo ls /var/lib/waagent/Microsoft.Azure.KeyVault -la", sudo=True)
+
+    # List the contents directory /var/lib/waagent/Microsoft.Azure.KeyVault
+    ls_result = node.execute(
+        "sudo ls /var/lib/waagent/Microsoft.Azure.KeyVault -la", sudo=True
+    )
     log.info(f"Directory contents: {ls_result.stdout}")
+
     # Get the OS release information
     os_release_result = node.execute("cat /etc/os-release", sudo=False)
     log.info(f"OS release: {os_release_result.stdout}")
-
-
-
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -14,13 +14,13 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 import requests
 from assertpy import assert_that
-from azure.identity import DefaultAzureCredential
-from azure.keyvault.certificates import (
+from azure.identity import DefaultAzureCredential  # type: ignore
+from azure.keyvault.certificates import (   # type: ignore
     CertificateClient,
     CertificatePolicy,
     KeyVaultCertificate,
 )
-from azure.keyvault.secrets import SecretClient
+from azure.keyvault.secrets import SecretClient  # type: ignore
 from azure.mgmt.compute import ComputeManagementClient  # type: ignore
 from azure.mgmt.compute.models import VirtualMachine  # type: ignore
 from azure.mgmt.marketplaceordering import MarketplaceOrderingAgreements  # type: ignore
@@ -47,7 +47,7 @@ from azure.mgmt.storage.models import (  # type: ignore
     Sku,
     StorageAccountCreateParameters,
 )
-from azure.storage.blob import (
+from azure.storage.blob import (  # type: ignore
     AccountSasPermissions,
     BlobClient,
     BlobSasPermissions,
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -68,12 +68,10 @@ from PIL import Image, UnidentifiedImageError
 from retry import retry
 
 from lisa import schema
-from lisa.base_tools.service import Service
 from lisa.environment import Environment, load_environments
 from lisa.feature import Features
 from lisa.node import Node, RemoteNode, local
 from lisa.secret import PATTERN_HEADTAIL, PATTERN_URL, add_secret, replace
-from lisa.tools.ls import Ls
 from lisa.util import (
     LisaException,
     LisaTimeoutException,
@@ -1931,24 +1929,19 @@ def add_system_assign_identity(
     params_identity = {"type": "SystemAssigned"}
     params_create = {"location": location, "identity": params_identity}
 
-    try:
-        vm_poller = compute_client.virtual_machines.begin_create_or_update(
-            resource_group_name,
-            vm_name,
-            params_create,
-        )
-        vm_result = vm_poller.result()
-        object_id_vm = vm_result.identity.principal_id
-        log.info(object_id_vm)
+    vm_poller = compute_client.virtual_machines.begin_create_or_update(
+        resource_group_name,
+        vm_name,
+        params_create,
+    )
+    vm_result = vm_poller.result()
+    object_id_vm = vm_result.identity.principal_id
+    log.info(object_id_vm)
 
-        if not object_id_vm:
-            raise ValueError("object_id_vm is not set.")
+    if not object_id_vm:
+        raise ValueError("object_id_vm is not set.")
 
-        return object_id_vm
-
-    except Exception as e:
-        log.error(f"Error creating VM: {e}")
-        raise LisaException(str(e))
+    return object_id_vm
 
 
 def get_key_vault_management_client(
@@ -1962,7 +1955,6 @@ def create_keyvault(
     resource_group_name: str,
     tenant_id: str,
     object_id: str,
-    object_id_vm: str,
     location: str,
     vault_name: str,
 ) -> Any:
@@ -1979,13 +1971,6 @@ def create_keyvault(
                     keys=["all"], secrets=["all"], certificates=["all"]
                 ),
             ),
-            AccessPolicyEntry(
-                tenant_id=tenant_id,
-                object_id=object_id_vm,
-                permissions=Permissions(
-                    keys=["all"], secrets=["all"], certificates=["all"]
-                ),
-            ),
         ],
     )
 
@@ -1994,6 +1979,34 @@ def create_keyvault(
     )
     keyvault_poller = keyvault_client.vaults.begin_create_or_update(
         resource_group_name, vault_name, parameters
+    )
+
+    return keyvault_poller.result()
+
+
+def assign_access_policy_to_vm(
+    platform: "AzurePlatform",
+    resource_group_name: str,
+    tenant_id: str,
+    object_id_vm: str,
+    vault_name: str,
+) -> Any:
+    keyvault_client = get_key_vault_management_client(platform)
+
+    # Fetch the current policies and add the VM's policy
+    vault = keyvault_client.vaults.get(resource_group_name, vault_name)
+    current_policies = vault.properties.access_policies
+    new_policy = AccessPolicyEntry(
+        tenant_id=tenant_id,
+        object_id=object_id_vm,
+        permissions=Permissions(keys=["all"], secrets=["all"], certificates=["all"]),
+    )
+    current_policies.append(new_policy)
+
+    # Update the vault with the new policies
+    vault.properties.access_policies = current_policies
+    keyvault_poller = keyvault_client.vaults.begin_create_or_update(
+        resource_group_name, vault_name, vault
     )
 
     return keyvault_poller.result()
@@ -2168,20 +2181,3 @@ def rotate_certificates(
                     f"More exception info: {e}"
                 )
                 raise LisaException(str(e))
-
-
-def check_system_status(node: Node, log: Logger) -> None:
-    # Check the status of the akvvm_service service using the Service tool
-    service = node.tools[Service]
-    if service.is_service_running("akvvm_service.service"):
-        log.info("akvvm_service is running")
-    else:
-        log.error("akvvm_service is not running")
-        raise LisaException("akvvm_service is not running. Test case failed.")
-
-    # List the contents of the directory
-    ls = node.tools[Ls]
-    directory_contents = ls.run(
-        "/var/lib/waagent/Microsoft.Azure.KeyVault -la", sudo=True
-    ).stdout
-    log.info(f"Directory contents: {directory_contents}")
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -15,7 +15,11 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 import requests
 from assertpy import assert_that
 from azure.identity import DefaultAzureCredential
-from azure.keyvault.certificates import CertificateClient, CertificatePolicy
+from azure.keyvault.certificates import (
+    CertificateClient,
+    CertificatePolicy,
+    KeyVaultCertificate,
+)
 from azure.keyvault.secrets import SecretClient
 from azure.mgmt.compute import ComputeManagementClient  # type: ignore
 from azure.mgmt.compute.models import VirtualMachine  # type: ignore
@@ -1970,7 +1974,6 @@ def create_certificates(
 
 
 def rotate_certificates(
-    self,
     log: Logger,
     vault_url: str,
     credential: DefaultAzureCredential,
@@ -1998,10 +2001,9 @@ def rotate_certificates(
                 cert_name_to_rotate, policy=cert_policy
             )
             create_certificate_result = create_certificate_poller.result()
-
             # Handle possible None value
             if (
-                create_certificate_result
+                isinstance(create_certificate_result, KeyVaultCertificate)
                 and hasattr(create_certificate_result, "properties")
                 and create_certificate_result.properties
             ):
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1944,8 +1944,7 @@ def get_tenant_id(credential: Any) -> Any:
 
 def get_identity_id() -> Any:
     # Define constants
-    graph_api_url = "https://graph.microsoft.com/.default"
-    request_url = "https://graph.microsoft.com/v1.0/me"
+    graph_api_url = "https://graph.microsoft.com/v1.0/serviceprincipals"
 
     # Get a token for the Microsoft Graph API
     token_credential = DefaultAzureCredential()
@@ -1958,7 +1957,7 @@ def get_identity_id() -> Any:
     }
 
     # Set a timeout of 10 seconds for the request
-    response = requests.get(request_url, headers=headers, timeout=10)
+    response = requests.get(graph_api_url, headers=headers, timeout=10)
 
     if response.status_code != 200:
         raise LisaException(
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -2163,7 +2163,7 @@ def rotate_certificate(
         certificate_name=cert_name, enabled=True
     )
 
- 
+
 @retry(tries=10, delay=1)
 def delete_certificate(
     platform: "AzurePlatform",
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1976,7 +1976,7 @@ def add_system_assign_identity(
     )
     vm_result = vm_poller.result()
     object_id_vm = vm_result.identity.principal_id
-    log.info(object_id_vm)
+    log.info(f"VM object ID assigned: {object_id_vm}")
 
     if not object_id_vm:
         raise ValueError(
@@ -2060,8 +2060,7 @@ def delete_keyvault(
     keyvault_client = get_key_vault_management_client(platform)
 
     keyvault_poller = keyvault_client.vaults.delete(resource_group_name, vault_name)
-    log.info(f"{keyvault_poller}")
-    log.info(f"Key Vault {vault_name} deleted successfully.")
+    log.debug(f"{keyvault_poller} - Key Vault {vault_name} deleted successfully.")
 
 
 def get_certificate_client(
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -2087,9 +2087,14 @@ def create_certificate(
 
     secret_id: Optional[str] = secret_client.get_secret(name=cert_name).id
     if secret_id:
-        match = re.match(r"(https://.+?/secrets/.+?)(?:/[^/]+)?$", secret_id)
+        # Example: "https://example.vault.azure.net/secrets/Cert-123/SomeVersion"
+        # Expected match for 'cert_url':
+        # "https://example.vault.azure.net/secrets/Cert-123"
+        match = re.match(
+            r"(?P<cert_url>https://.+?/secrets/.+?)(?:/[^/]+)?$", secret_id
+        )
         if match:
-            secret_url_without_version = match.group(1)
+            secret_url_without_version = match.group("cert_url")
             return secret_url_without_version
         else:
             raise LisaException(
@@ -2152,9 +2157,9 @@ def rotate_certificate(
     ):
         new_certificate_version = create_certificate_result.properties.version
         log.debug(
-            f"New version of certificate '{cert_name}': " f"{new_certificate_version}"
+            f"New version of certificate '{cert_name}': {new_certificate_version}. "
+            "Certificate rotated."
         )
-        log.debug("Certificate rotated")
     else:
         error_message = "Failed to retrieve properties from create certificate result."
         raise LisaException(error_message)
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -15,6 +15,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 from msrestazure import AzureConfiguration
 
 import requests
+from assertpy import assert_that
 from azure.mgmt.compute import ComputeManagementClient  # type: ignore
 from azure.mgmt.compute.models import VirtualMachine  # type: ignore
 from azure.mgmt.marketplaceordering import MarketplaceOrderingAgreements  # type: ignore
@@ -150,6 +151,8 @@ class NodeContext:
     use_public_address: bool = True
     public_ip_address: str = ""
     private_ip_address: str = ""
+    location: str = ""
+    subscription_id: str = ""
 
 
 @dataclass_json()
@@ -1454,6 +1457,9 @@ def load_environment(
     for node in environment.nodes.list():
         assert isinstance(node, RemoteNode)
 
+        vm = vms_map.get(node.name)
+        assert_that(vm).is_not_none()
+
         node_context = get_node_context(node)
         node_context.vm_name = node.name
         node_context.resource_group_name = resource_group_name
@@ -1461,6 +1467,9 @@ def load_environment(
         node_context.username = platform_runbook.admin_username
         node_context.password = platform_runbook.admin_password
         node_context.private_key_file = platform_runbook.admin_private_key_file
+        node_context.location = vm.location
+        node_context.subscription_id = platform.subscription_id
+
         (
             node_context.public_ip_address,
             node_context.private_ip_address,
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -14,7 +14,6 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 
 import requests
 from assertpy import assert_that
-from azure.identity import DefaultAzureCredential
 from azure.keyvault.certificates import (
     CertificateClient,
     CertificatePolicy,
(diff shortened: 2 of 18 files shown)


LiliDeng:
put the change in https://github.com/microsoft/lisa/pull/2889 into this PR

t-ecastillo:
> put the change in #2889 into this PR

Done!

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        # Identity assignment
+        compute_client = ComputeManagementClient(credential, platform.subscription_id)
+        vm = compute_client.virtual_machines.get(
+            resource_group_name, node_context.vm_name
+        )
+        object_id_vm = vm.identity.principal_id
+        if object_id_vm is None:
+            raise ValueError("object_id_vm is not set.")
+
+        # Key Vault properties
```
bamuir: We could create wrappers in common.py for KeyVaultManagmentClient. For example, storage accounts are done similarly:

https://github.com/microsoft/lisa/blob/573570bd22f696b01f85bf3aabe98f838ddac1ef/lisa/sut_orchestrator/azure/common.py#L885

I have a WIP in this format, we can sync on it. Not blocking for your PR, I can make that improvement with my ADE PR

t-ecastillo: Yep, there is a lot of room for improvement in the code, it was my first-time using LISA so definitely lots of part of the code that can be refactored.

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        # Section for environment setup
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
+        platform = environment.platform
+        assert isinstance(platform, AzurePlatform)
+        # VM attributes
+        node_context = get_node_context(node)
+        resource_group_name = node_context.resource_group_name
+        node_capability = node.capability.get_extended_runbook(AzureNodeSchema, AZURE)
+        location = node_capability.location
```
LiliDeng: we can use node_context.location

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        A Key Vault
+        Two certificates in the Key Vault
+        Retrieval of the certificate's secrets
+        through SecretClient class from Azure SDK.
+        Installation of the Azure Key Vault Linux Extension on the VM.
+        Rotation of the certificates (After KVVM Extension has been installed)
+        All of the resources have been created by using the Azure SDK Python.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_features=[AzureExtension]),
```
LiliDeng: Add unsupported_os=[BSD], then remove https://github.com/microsoft/lisa/pull/2892/files#diff-e6524431f41c0a29341f49d0d135923d404fddfde2c675e3206fe0dad3b5656aR74

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1912,141 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+        except Exception as e:
+            if attempt < 1:
+                sleep(1)
+            else:
+                log.error(
+                    f"Failed to rotate certificate:"
+                    f"{cert_name_to_rotate}' after 2 attempts."
+                    f"More exception info: {e}"
+                )
+                raise
```
LiliDeng: raise LisaException(str(e))

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1912,141 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+
+
+def check_system_status(node: Node, log: Logger) -> None:
+    # Check the status of the akvvm_service service
+    akvvm_service_result = node.execute(
+        "systemctl status akvvm_service.service", sudo=True, timeout=10
+    )
```
LiliDeng, line 2150: use node.tools[Service]

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1912,141 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    log.info(f"akvvm_service status: {akvvm_service_result.stdout}")
+
+    # List the contents directory /var/lib/waagent/Microsoft.Azure.KeyVault
+    ls_result = node.execute(
+        "sudo ls /var/lib/waagent/Microsoft.Azure.KeyVault -la", sudo=True
+    )
+    log.info(f"Directory contents: {ls_result.stdout}")
```
LiliDeng, line 2048: use node.tools[Ls]

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1912,141 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+def check_system_status(node: Node, log: Logger) -> None:
+    # Check the status of the akvvm_service service
+    akvvm_service_result = node.execute(
+        "systemctl status akvvm_service.service", sudo=True, timeout=10
+    )
+    log.info(f"akvvm_service status: {akvvm_service_result.stdout}")
+
+    # List the contents directory /var/lib/waagent/Microsoft.Azure.KeyVault
+    ls_result = node.execute(
+        "sudo ls /var/lib/waagent/Microsoft.Azure.KeyVault -la", sudo=True
```
LiliDeng: use node.tools[Service]

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1912,141 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    )
+    log.info(f"Directory contents: {ls_result.stdout}")
+
+    # Get the OS release information
+    os_release_result = node.execute("cat /etc/os-release", sudo=False)
```
LiliDeng: it is not used, we can remove this?

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        ).is_equal_to(vault_name_a)
+        assert_that(keyvault_result.location).described_as(
+            "Expected the Key Vault location to match the given value"
+        ).is_equal_to(location)
+
+        # Certificates
+        log.info(
+            "About to call create_certificates with vault_url: %s",
+            keyvault_result.properties.vault_uri,
+        )
```
LiliDeng: f"About to call create_certificates with vault_url: {keyvault_result.properties.vault_uri}"

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        certificate1_secret_id, certificate2_secret_id = create_certificates(
+            vault_url=keyvault_result.properties.vault_uri,
+            credential=credential,
+            log=log,
+        )
+        log.info(
+            "Certificates created. Cert1 ID: %s, Cert2 ID: %s",
+            certificate1_secret_id,
+            certificate2_secret_id,
+        )
```
LiliDeng: f"Certificates created. Cert1 ID: {certificate1_secret_id}, Cert2 ID: {certificate2_secret_id}"

lisa/sut_orchestrator/azure/common.py
t-ecastillo: My manager wants to see the OS of the VM, we use it :)

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1912,141 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+        except Exception as e:
+            log.error(
+                f"Failed to retrieve secret ID for certificate named '{cert_name}': {e}"
+            )
+            raise
+
+    return secret_urls
+
+
+def rotate_certificates(
```
LiliDeng, line 1976: move this to azure/common.py

pyproject.toml
```diff
@@ -49,6 +50,9 @@ azure = [
     "azure-mgmt-storage ~= 20.0.0",
     "azure-storage-blob ~= 12.11.0",
     "azure-storage-file-share ~= 12.4.0",
+    "azure-keyvault-secrets ~= 4.7.0", 
+    "azure-keyvault-certificates ~= 4.7.0",
+    "azure-mgmt-authorization ~= 4.0.0",
```
LiliDeng: is it used?

t-ecastillo: yes, authorization is not, I will delete it

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        requirement=simple_requirement(supported_features=[AzureExtension]),
+    )
+    def verify_key_vault_extension(
+        self, log: Logger, node: Node, result: TestResult
+    ) -> None:
+        # Section for vault name and supported OS check
+        vault_name = os.getenv(
+            "vault_name", f"python-keyvault-{random.randint(1, 100000):05}"
+        )
+        vault_name_a = vault_name + "prp"
```
LiliDeng: vault_name = f"python-keyvault-{random.randint(1, 100000):05}prp"

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        compute_client = ComputeManagementClient(credential, platform.subscription_id)
+        vm = compute_client.virtual_machines.get(
+            resource_group_name, node_context.vm_name
+        )
+        object_id_vm = vm.identity.principal_id
+        if object_id_vm is None:
+            raise ValueError("object_id_vm is not set.")
+
+        # Key Vault properties
+        keyvault_client = KeyVaultManagementClient(credential, platform.subscription_id)
```
LiliDeng: yes, move KeyVaultManagementClient into get_keyvault_client in azure/common.py

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+                    ),
+                ),
+            ],
+        )
+
+        # Create Key Vault
+        parameters = VaultCreateOrUpdateParameters(
+            location=location, properties=vault_properties
+        )
+        keyvault_poller = keyvault_client.vaults.begin_create_or_update(
```
LiliDeng: create create_update_keyvault in azure/common.py

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        ).is_equal_to("Succeeded")
+
+        # Rotate certificates
+        rotate_certificates(
+            log,
+            vault_url=keyvault_result.properties.vault_uri,
+            credential=credential,
+            cert_name_to_rotate="Cert1",
+        )
+        assert True, "Cert1 has been rotated."
```
LiliDeng: no need this?

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        check_system_status(node, log)
```
LiliDeng: here is to check what things need to be checked after certificates rotated

LiliDeng: uninstall extensions,  and clean up for all created resources in the end

lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -339,6 +339,10 @@
                 "name": "vmCopy",
                 "count": "[variables('node_count')]"
             },
+            "identity": {
```
LiliDeng: remove this to reduce the potential impact

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1436,6 +1443,7 @@ def load_environment(
         node_schema = schema.RemoteNode(name=vm.name)
         environment_runbook.nodes_raw.append(node_schema)
         vms_map[vm.name] = vm
+        vm.identity = {"type": "SystemAssigned"}
```
LiliDeng: remove this

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,211 @@
...
+        # Section for vault name and supported OS check
+        vault_name = os.getenv(
+            "vault_name", f"python-keyvault-{random.randint(1, 100000):05}"
+        )
+        vault_name_a = vault_name + "prp"
+        if isinstance(node.os, FreeBSD):
+            raise SkippedException(f"unsupported distro type: {type(node.os)}")
+
+        # Section for environment setup
+        environment = result.environment
```
LiliDeng: add a method in azure/common.py to enable SYSTEM_ASSIGNED_IDENTITY for the vm.
refer https://learn.microsoft.com/en-us/samples/azure-samples/compute-python-msi-vm/compute-python-msi-vm/

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1918,310 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+def get_key_vault_access_control_client(
+    vault_url: str, credential: DefaultAzureCredential
+) -> KeyVaultAccessControlClient:
+    return KeyVaultAccessControlClient(vault_url=vault_url, credential=credential)
+
+
+# If RBAC policy is needed/wanted for Key Vault
+# Remember then to set up enable_rbac_role = True
+# In Vault properties
+def assign_role_to_principal(
```
LiliDeng: seems it is not used

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1918,310 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+) -> None:
+    # Use the helper function to get certificate_client
+    certificate_client = get_certificate_client(vault_url, credential)
+
+    try:
+        # Retrieve the old version of the certificate
+        certificate_client.get_certificate(cert_name_to_rotate)
+    except Exception as e:
+        log.error(
+            f"Failed to retrieve old version of certificate:"
```
LiliDeng: remove the beginning `f`

lisa/sut_orchestrator/azure/common.py
LiliDeng: os version always shown for each deployment, so no need here

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1918,310 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    service = node.tools[Service]
+    if service.is_service_running("akvvm_service.service"):
+        log.info("akvvm_service is running")
+    else:
+        log.info("akvvm_service is not running")
+
+    # List the contents of the directory
+    ls = node.tools[Ls]
+    directory_contents = ls.run("/var/lib/waagent/Microsoft.Azure.KeyVault -la").stdout
+    log.info(f"Directory contents: {directory_contents}")
```
LiliDeng: should we fail the case if the status is unexpected? seems it only prints some info after rotate cert

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,222 @@
...
+        priority=0,
+        requirement=simple_requirement(
+            supported_features=[AzureExtension], unsupported_os=[BSD]
+        ),
+    )
+    def verify_key_vault_extension(
+        self, log: Logger, node: Node, result: TestResult
+    ) -> None:
+        # Section for vault name and supported OS check
+        vault_name_a = f"python-keyvault-{random.randint(1, 100000):05}prp"
```
LiliDeng: change all vault_name_a => vault_name

lisa/sut_orchestrator/azure/common.py
t-ecastillo: it adds flexibility for users to decide in which type of access policies they want to implement on their key vault, I can delete but chance of someone adding this in the future are very likely

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,222 @@
...
+            raise ValueError("Environment variable 'tenant_id' is not set.")
+        user_object_id = os.environ.get("user_object_id")
+        if user_tenant_id is None:
+            raise ValueError("Environment variable 'user_object_id' is not set.")
+        assert user_object_id is not None
+        credential = DefaultAzureCredential(additionally_allowed_tenants=["*"])
+
+        # Object ID System assignment
+        object_id_vm = add_system_assign_identity(
+            credential=credential,
```
LiliDeng: we can use platform.credential?

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,220 @@
...
+        settings = {
+            "secretsManagementSettings": {
+                "autoUpgradeMinorVersion": True,
+                "enableAutomaticUpgrade": True,
+                "pollingIntervalInS": "360",
+                "certificateStoreLocation": "/var/lib/waagent/Microsoft.Azure.KeyVault",
+                "observedCertificates": [
+                    certificate1_secret_id,
+                    certificate2_secret_id,
+                ],
```
LiliDeng: certificates_secret_id: List[str]
    for cert_name in ["Cert1", "Cert2"]:
        certificate_secret_id = create_certificate(
            vault_url=keyvault_result.properties.vault_uri,
            credential=platform.credential,
            log=log,
            cert_name=cert_name,
        )
        log.info(
            f"Certificates created. Cert ID: {certificate_secret_id}, "
        )
        assert_that(certificate_secret_id).described_as(
            "First certificate created successfully"
        ).is_not_none()
        certificates_secret_id.append(certificate_secret_id)

    # Extension
    extension_name = "KeyVaultForLinux"
    extension_publisher = "Microsoft.Azure.KeyVault"
    extension_version = "2.0"
    settings = {
        "secretsManagementSettings": {
            "autoUpgradeMinorVersion": True,
            "enableAutomaticUpgrade": True,
            "pollingIntervalInS": "360",
            "certificateStoreLocation": "/var/lib/waagent/Microsoft.Azure.KeyVault",
            "observedCertificates": [
                certificates_secret_id[0],
                certificates_secret_id[1],
            ],
        }
    }

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,220 @@
...
+                cert_name="Cert1",
+                credential=platform.credential,
+                log=log,
+            )
+        ).is_false
+
+        delete_certificate(
+            vault_url=keyvault_result.properties.vault_uri,
+            credential=platform.credential,
+            cert_name="Cert2",
```
LiliDeng: use loop when delete certs

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,220 @@
...
+        node_context = get_node_context(node)
+
+        # User's attributes
+        user_tenant_id = os.environ.get("tenant_id")
+        if user_tenant_id is None:
+            raise ValueError("Environment variable 'tenant_id' is not set.")
+        user_object_id = os.environ.get("user_object_id")
+        if user_tenant_id is None:
+            raise ValueError("Environment variable 'user_object_id' is not set.")
+        assert user_object_id is not None
```
LiliDeng: when tenant id and app id specified in runbook, we can use below os environment.

tenant_id = os.environ["AZURE_TENANT_ID"]
object_id = os.environ["AZURE_CLIENT_ID"]

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1917,269 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+
+def get_key_vault_management_client(
+    credential: DefaultAzureCredential, subscription_id: str
+) -> KeyVaultManagementClient:
+    return KeyVaultManagementClient(credential, subscription_id)
+
+
+def create_keyvault(
+    credential: DefaultAzureCredential,
+    subscription_id: str,
```
LiliDeng: use `platform: "AzurePlatform"` then use platform.credential, platform.subscription_id in this method.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1917,269 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    keyvault_poller = keyvault_client.vaults.begin_create_or_update(
+        resource_group_name, vault_name, parameters
+    )
+
+    return keyvault_poller.result()
+
+
+def delete_keyvault(
+    credential: Any,
+    subscription_id: str,
```
LiliDeng: use platform: "AzurePlatform"

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1917,269 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
             return iops_dict[min_iops]
         else:
             raise LisaException(f"Data disk type {disk_type} is unsupported.")
+
+
+def add_system_assign_identity(
+    credential: DefaultAzureCredential,
+    subscription_id: str,
+    resource_group_name: str,
```
LiliDeng, line 1976: ues `platform: "AzurePlatform",`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1917,269 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+
+        return object_id_vm
+
+    except Exception as e:
+        log.error(f"Error creating VM: {e}")
+        raise LisaException(str(e))
+
+
+def get_key_vault_management_client(
+    credential: DefaultAzureCredential, subscription_id: str
```
LiliDeng: use `platform: "AzurePlatform"`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1917,269 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    return CertificateClient(vault_url=vault_url, credential=credential)
+
+
+def get_secret_client(vault_url: str, credential: Any) -> SecretClient:
+    return SecretClient(vault_url=vault_url, credential=credential)
+
+
+def create_certificate(
+    vault_url: str,
+    credential: Any,
```
LiliDeng: use `platform: "AzurePlatform"`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,272 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+        vm_result = vm_poller.result()
+        object_id_vm = vm_result.identity.principal_id
+        log.info(object_id_vm)
+
+        if not object_id_vm:
+            raise ValueError("object_id_vm is not set.")
+
+        return object_id_vm
+
+    except Exception as e:
```
squirrelsc: The try/except doesn't perform special logic. The log will be printed out automatically. So remove the try/except is similar behavior.

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,201 @@
...
+        priority=1,
+        requirement=simple_requirement(
+            supported_features=[AzureExtension], unsupported_os=[BSD]
+        ),
+    )
+    def verify_key_vault_extension(
+        self, log: Logger, node: Node, result: TestResult
+    ) -> None:
+        # Section for vault name and supported OS check
+        vault_name = f"python-keyvault-{random.randint(1, 100000):05}prp"
```
squirrelsc: Make the kv name associated with the test case name like `kv-extension-...`

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,201 @@
...
+            object_id=object_id,
+            object_id_vm=object_id_vm,
+            location=node_context.location,
+            vault_name=vault_name,
+        )
+
+        log.info(f"Created Key Vault {keyvault_result.properties.vault_uri}")
+
+        # Certificates
+        log.info(
```
squirrelsc: Either print log before actions, or after actions, instead of mixing them. The above line printed after key vault created, but this line printed before the action. And this line doesn't provide extra information than above line. This line should be removed.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,272 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+            else:
+                log.error(
+                    f"Failed to rotate certificate:"
+                    f"{cert_name_to_rotate}' after 2 attempts."
+                    f"More exception info: {e}"
+                )
+                raise LisaException(str(e))
+
+
+def check_system_status(node: Node, log: Logger) -> None:
```
squirrelsc: This method is used by akv test case only, so move it to the test case, and as a private method (start with `_`).

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,201 @@
+# Copyright (c) Microsoft Corporation.
```
squirrelsc: Make the file name be consistent with the extension name.

microsoft/testsuites/vm_extensions/kvvm_extension.py
```diff
@@ -0,0 +1,201 @@
...
+    category="functional",
+    description="BVT for Azure Key Vault Extension",
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class AzureKeyVaultExtensionBvt(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        The following test case validates the Azure Key Vault Linux
+        Extension while creating the following resources:
+        A resource group
```
squirrelsc: use a `*` as bullet list.

microsoft/testsuites/vm_extensions/kvvm_extension.py
squirrelsc: The test case may be run more than 10k times in a month, it's easy to be conflict. Please use a shared key vault to avoid conflict.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,272 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+        sku=KeyVaultSku(name="standard"),
+        access_policies=[
+            AccessPolicyEntry(
+                tenant_id=tenant_id,
+                object_id=object_id,
+                permissions=Permissions(
+                    keys=["all"], secrets=["all"], certificates=["all"]
+                ),
+            ),
+            AccessPolicyEntry(
```
squirrelsc: The access policy of the machine should be in a separated method. In this method, it creates the keyvault, so it shouldn't do extra work, like assign policy to a vm. So this method should only assign policy to the current account for convenient.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1914,270 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    vault_name: str,
+    log: Logger,
+) -> None:
+    keyvault_client = get_key_vault_management_client(platform)
+
+    try:
+        keyvault_poller = keyvault_client.vaults.delete(resource_group_name, vault_name)
+        log.info(f"{keyvault_poller}")
+        log.info(f"Key Vault {vault_name} deleted successfully.")
+    except Exception as e:
```
squirrelsc: This try/except isn't useful as well. Please check all similar usage.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,224 @@
...
+
+@TestSuiteMetadata(
+    area="vm_extension",
+    category="functional",
+    description="BVT for Azure Key Vault Extension",
+    requirement=simple_requirement(unsupported_os=[]),
+)
+class AzureKeyVaultExtensionBvt(TestSuite):
+    # Private method for checking system status
+    def _check_system_status(self, node: Node, log: Logger) -> None:
```
squirrelsc: Move private methods behind the public methods. For this method, it doesn't use any method of the test class, so it can be a function, instead of a method in the class.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1914,270 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+                log.error(
+                    "Failed to retrieve properties from create certificate result."
+                )
+
+            certificate_client.update_certificate_properties(
+                certificate_name=cert_name_to_rotate, enabled=True
+            )
+            break
+        except Exception as e:
+            if attempt < 1:
```
squirrelsc: The pattern looks complex. Refer to other LISA code, use the `retry` package for attempts.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1914,270 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    location: str,
+    log: Logger,
+) -> Any:
+    compute_client = ComputeManagementClient(
+        platform.credential, platform.subscription_id
+    )
+    params_identity = {"type": "SystemAssigned"}
+    params_create = {"location": location, "identity": params_identity}
+
+    vm_poller = compute_client.virtual_machines.begin_create_or_update(
```
squirrelsc: Call it to use update only, avoid adding a VM incidentally.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1914,270 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+        resource_group_name,
+        vm_name,
+        params_create,
+    )
+    vm_result = vm_poller.result()
+    object_id_vm = vm_result.identity.principal_id
+    log.info(object_id_vm)
+
+    if not object_id_vm:
+        raise ValueError("object_id_vm is not set.")
```
squirrelsc: make it more descriptive, like "cannot retrieve managed identity after set system assigned identity on vm."

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,224 @@
...
+        # Section for environment setup
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
+        platform = environment.platform
+        assert isinstance(platform, AzurePlatform)
+        # VM attributes
+        node_context = get_node_context(node)
+
+        # User's attributes
+        tenant_id = os.environ["AZURE_TENANT_ID"]
```
squirrelsc: Add a method in common.py to call Azure API to get current tenant id, instead of reading from the environment variable. The variable may not exist.

squirrelsc: Do similar with the client id as well.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1914,270 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+) -> bool:
+    certificate_client = CertificateClient(
+        vault_url=vault_url, credential=platform.credential
+    )
+    try:
+        # Try to get the certificate. If this call succeeds, the certificate exists.
+        certificate_client.get_certificate(cert_name)
+        return True
+    except Exception as e:
+        log.error(f"Failed check existance of cert:'{cert_name}': {e}")
```
squirrelsc: Don't log error, the error log is hard to check. Raise exception directly. If it doesn't need to fail a case, catch it in test case level.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # Section for environment setup
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
+        platform = environment.platform
+        assert isinstance(platform, AzurePlatform)
+        runbook = platform.runbook.get_extended_runbook(AzurePlatformSchema)
+        # Use the shared_resource_group_name if the resource_group_name is not provided
+        resource_group_name = (
+            runbook.resource_group_name or runbook.shared_resource_group_name
```
LiliDeng: we can use node_context.resource_group_name

t-ecastillo: @squirrelsc recommended me to create the key vault in the shared resource group.

squirrelsc: @LiliDeng The Key Vault name is unique globally, it's easy to be conflict. And a key vault name won't be reused in 90 days, after it's deleted. So, it needs to avoid creating a lot of key vaults.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,285 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    return keyvault_poller.result()
+
+
+def delete_keyvault(
+    platform: "AzurePlatform", vault_name: str, log: Logger, resource_group_name: str
+) -> None:
+    keyvault_client = get_key_vault_management_client(platform)
+
+    keyvault_poller = keyvault_client.vaults.delete(resource_group_name, vault_name)
+    log.info(f"{keyvault_poller}")
```
squirrelsc: Merge two lines into one line log. And check all `info`, if debug is enough.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,231 @@
...
+        priority=1,
+        requirement=simple_requirement(
+            supported_features=[AzureExtension], unsupported_os=[BSD]
+        ),
+    )
+    def verify_key_vault_extension(
+        self, log: Logger, node: Node, result: TestResult
+    ) -> None:
+        # Section for vault name and supported OS check
+        vault_name = f"kve-{time.strftime('%y%m%d%H%M%S')}-{random.randint(1, 1000):03}"
```
squirrelsc: The vault name should be consistent for a subscription, so the key vault will be reused in next run. Only the certificate name should be a random one to avoid conflict on different test cases. You can refer to other place, use the last chars of subscription id to get a consistent key vault name.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Section for environment setup
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
+        platform = environment.platform
+        assert isinstance(platform, AzurePlatform)
+        runbook = platform.runbook.get_extended_runbook(AzurePlatformSchema)
+        # Use the shared_resource_group_name if the resource_group_name is not provided
+        resource_group_name = (
```
squirrelsc: Always use the shared resource_group_name to create the keyvault.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,231 @@
...
+        for cert_name in ["Cert2", "Cert1"]:
+            delete_certificate(
+                platform=platform,
+                vault_url=keyvault_result.properties.vault_uri,
+                cert_name=cert_name,
+                log=log,
+            )
+
+            retries = 2
+            for _ in range(retries):
```
squirrelsc: search for `@retry(` in LISA to use the package to deal with retry logic.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,231 @@
...
+                certificate_exists = check_certificate_existence(
+                    vault_url=keyvault_result.properties.vault_uri,
+                    cert_name=cert_name,
+                    log=log,
+                    platform=platform,
+                )
+                if not certificate_exists:
+                    break
+                time.sleep(5)  # wait for 5 seconds before retrying
```
squirrelsc: 5 seconds is too long to check. Try 10 times every 1 second. It still allow 10 seconds before timeout, but won't waste time more than 1 second.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,231 @@
...
+
+            assert_that(certificate_exists).is_false()
+        # Deleting key vault
+        delete_keyvault(
```
squirrelsc: Because the keyvault is shared, so it doesn't need to delete it in test case level.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,231 @@
...
+                    vault_url=keyvault_result.properties.vault_uri,
+                    cert_name=cert_name,
+                    log=log,
+                    platform=platform,
+                )
+                if not certificate_exists:
+                    break
+                time.sleep(5)  # wait for 5 seconds before retrying
+
+            assert_that(certificate_exists).is_false()
```
squirrelsc: Use `described_as` to add error message. So once it's failed, people know what happened and reason or solution.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,285 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    params_create = {"location": location, "identity": params_identity}
+
+    vm_poller = compute_client.virtual_machines.begin_create_or_update(
+        resource_group_name,
+        vm_name,
+        params_create,
+    )
+    vm_result = vm_poller.result()
+    object_id_vm = vm_result.identity.principal_id
+    log.info(object_id_vm)
```
squirrelsc: Don't print out value only, it should let people know what's the value.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+
+def get_tenant_id(credential: Any) -> Any:
+    # Initialize the Subscription client
+    subscription_client = SubscriptionClient(credential)
+    # Get the subscription
+    subscription = next(subscription_client.subscriptions.list())
+    return subscription.tenant_id
+
+
+def get_object_id() -> Any:
```
squirrelsc: how about `get_identity_id`?

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
t-ecastillo: should I delete the method I created that deletes a key vault in common.py or just leave it?

squirrelsc: It can be deleted.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    log: Logger,
+) -> bool:
+    certificate_client = get_certificate_client(vault_url, platform)
+
+    try:
+        certificate_client.begin_delete_certificate(cert_name)
+        log.info(f"Certificate {cert_name} deleted successfully.")
+        return True
+    except Exception as e:
+        log.error(f"Error deleting certificate {cert_name}: {e}")
```
squirrelsc: As previous comments, don't use log.error. If it's an error, raise an exception. If it's not, print a debug level log. Please update similar places.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    # Set up the API call headers
+    headers = {
+        "Authorization": f"Bearer {token.token}",
+        "Content-Type": "application/json",
+    }
+
+    # Set a timeout of 10 seconds for the request
+    response = requests.get(request_url, headers=headers, timeout=10)
+
+    if response.status_code == 200:
```
squirrelsc: The logic can be simpler like

```suggestion
    if response.status_code != 200:
      raise LisaException(
            f"Failed to retrieve user object ID. "
            f"Status code: {response.status_code}. "
            f"Response: {response.text}"
      )

  return response.json().get("id")
```

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+def delete_keyvault(
+    platform: "AzurePlatform", vault_name: str, log: Logger, resource_group_name: str
+) -> None:
+    keyvault_client = get_key_vault_management_client(platform)
+
+    keyvault_poller = keyvault_client.vaults.delete(resource_group_name, vault_name)
+    log.debug(f"{keyvault_poller} - Key Vault {vault_name} deleted successfully.")
+
+
+def get_certificate_client(
```
squirrelsc: group get_xxx_client method together.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+            cert_name, policy=cert_policy
+        )
+        log.info(
+            f"Certificate '{cert_name}' has been created. "
+            f"Result: {create_certificate_result}"
+        )
+        certificate_client.update_certificate_properties(
+            certificate_name=cert_name, enabled=True
+        )
+        break
```
squirrelsc: It looks the break makes it try only once.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+
+    secret_id: Optional[str] = secret_client.get_secret(name=cert_name).id
+    if secret_id:
+        secret_url_without_version = secret_id.rsplit("/", 1)[0]
```
squirrelsc: use regexp, it's easy to fix and handle different patterns.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+def check_certificate_existence(
+    vault_url: str, cert_name: str, log: Logger, platform: "AzurePlatform"
+) -> bool:
+    certificate_client = CertificateClient(
+        vault_url=vault_url, credential=platform.credential
+    )
+
+    try:
+        certificate = certificate_client.get_certificate(cert_name)
+        log.info(f"Cert found '{certificate.name}'")
```
squirrelsc: Move it to debug level log. It doesn't need to be info level.

squirrelsc: There are other similar places. info level log is to tell a story, debug level is to provide details. In most cases, only test case level needs a few info level logs.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    except Exception as e:
+        log.error(f"Error deleting certificate {cert_name}: {e}")
+        return False
+
+
+def rotate_certificates(
+    log: Logger,
+    vault_url: str,
+    platform: "AzurePlatform",
+    cert_name_to_rotate: str,
```
squirrelsc: `cert_name` is enough. And make parameters order consistent with other methods. For example, put log as last one, and the platform as first one.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,284 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    try:
+        certificate_client.begin_delete_certificate(cert_name)
+        log.info(f"Certificate {cert_name} deleted successfully.")
+        return True
+    except Exception as e:
+        log.error(f"Error deleting certificate {cert_name}: {e}")
+        return False
+
+
+def rotate_certificates(
```
squirrelsc: It rotates only one cert, so the method should be called `rotate_certificate`.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -0,0 +1,231 @@
...
+from lisa.util import LisaException
+
+
+def _check_system_status(node: Node, log: Logger) -> None:
+    # Check the status of the akvvm_service service using the Service tool
+    service = node.tools[Service]
+    if service.is_service_running("akvvm_service.service"):
+        log.info("akvvm_service is running")
+    else:
+        log.error("akvvm_service is not running")
```
squirrelsc: It doesn't need to log error message before an exception raised.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1708,7 +1720,7 @@ def check_or_create_gallery(
                 "location": gallery_location,
                 "description": gallery_description,
             }
-            operation = compute_client.galleries.begin_create_or_update(
+            operation = compute_client.galleries.begin_update(
```
LiliDeng: no need update this

t-ecastillo: ah my bad, good catch

lisa/sut_orchestrator/azure/common.py
squirrelsc: Make this update only.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -75,32 +74,27 @@ class AzureKeyVaultExtensionBvt(TestSuite):
         * Printing the cert after rotation from the VM
         * Deletion of the resources
         """,
-        priority=1,
+        priority=0,
```
squirrelsc: keep it p1

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2137,70 +2121,61 @@ def check_certificate_existence(
...
-                f"{new_certificate_version}"
-            )
-            log.info("Certificate rotated")
-        else:
-            error_message = (
-                "Failed to retrieve properties from create certificate result."
-            )
-            log.error(error_message)
-            raise LisaException(error_message)
+        log.debug("Certificate rotated")
```
squirrelsc: Merge this line with above one. Avoid printing two lines of logs together. Each line of log should place on different code paths, unless there are too much information.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2061,56 +2061,40 @@ def assign_access_policy_to_vm(
...
+    )
+    certificate_client.update_certificate_properties(
+        certificate_name=cert_name, enabled=True
+    )
 
     secret_id: Optional[str] = secret_client.get_secret(name=cert_name).id
     if secret_id:
-        secret_url_without_version = secret_id.rsplit("/", 1)[0]
-        return secret_url_without_version
+        match = re.match(r"(https://.+?/secrets/.+?)(?:/[^/]+)?$", secret_id)
```
squirrelsc: 1. provide examples above the regexp patterns, so it's easy to fix, if there is any regression, or new patterns are found.
2. If it's a static pattern, put it in class level to avoid it's compiled multiple times. It doesn't improve performance a lot, but it's we used in LISA.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2061,56 +2061,40 @@ def assign_access_policy_to_vm(
...
+        if match:
+            secret_url_without_version = match.group(1)
```
squirrelsc: Use named group make the logic clearer.

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
squirrelsc: Is it updated?

squirrelsc:
@LiliDeng LGTM. Eli has to complete it this week. Let's prioritize this PR.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1916,271 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    token = token_credential.get_token(graph_api_url)
+
+    # Set up the API call headers
+    headers = {
+        "Authorization": f"Bearer {token.token}",
+        "Content-Type": "application/json",
+    }
+
+    # Set a timeout of 10 seconds for the request
+    response = requests.get(request_url, headers=headers, timeout=10)
```
LiliDeng: in our pipeline we use app token not user token, so we will encounter below error, we should handle both situation.
```
Failed to retrieve user object ID. Status code: 400. Response: {"error":{"code":"BadRequest","message":"/me request is only valid with delegated authentication flow.","innerError":{"date":"2023-08-17T02:25:00","request-id":"baede3c6-8aa1-46da-9c64-d56a4a751d8a","client-request-id":"baede3c6-8aa1-46da-9c64-d56a4a751d8a"
```

t-ecastillo: Ok I will start some research.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1915,278 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    if not application_id:
+        base_url = "https://graph.microsoft.com/"
+        api_version = "v1.0"
+        endpoint = "me"
+        graph_api_url = f"{base_url}{api_version}/{endpoint}"
+    else:
+        base_url = "https://graph.microsoft.com/"
+        api_version = "v1.0"
+        endpoint = f"servicePrincipals(appId='{application_id}')"
+        graph_api_url = f"{base_url}{api_version}/{endpoint}"
```
LiliDeng: move the common code above if

    base_url = "https://graph.microsoft.com/"
    api_version = "v1.0"
    if application_id:
        endpoint = f"servicePrincipals(appId='{application_id}')"
    else:
        endpoint = "me"
    graph_api_url = f"{base_url}{api_version}/{endpoint}"

t-ecastillo: this makes better sense thanks

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1915,278 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+def add_system_assign_identity(
+    platform: "AzurePlatform",
+    resource_group_name: str,
+    vm_name: str,
+    location: str,
+    log: Logger,
+) -> Any:
+    compute_client = ComputeManagementClient(
+        platform.credential, platform.subscription_id
+    )
```
LiliDeng: compute_client = get_compute_client(platform)

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1904,3 +1915,278 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    log: Logger,
+) -> bool:
+    certificate_client = get_certificate_client(vault_url, platform)
+
+    try:
+        certificate_client.begin_delete_certificate(cert_name)
+        log.debug(f"Certificate {cert_name} deleted successfully.")
+        return True
+    except Exception:
+        raise LisaException
```
LiliDeng: please give the detailed exception message

LiliDeng (approved):

PR #2894, opened 2023-08-08
verify_os_partition_identifier replace dmesg with journalctl

adityagesh:
Since environments are reused, this test has chance to fail because dmesg logs gets overwritten.
Fetch log from journalctl from last boot to verify


adityagesh:
Received review to check dmesg initially and fallback to journalctl. Will make that change.

LiliDeng (approved):

PR #2896, opened 2023-08-08
MANA and Pkg-Config

mcgov:
consolidated pkg-config and mana changes
too annoying to test seperately.

- adds pkg-config tool
- adds support for mana, failsafe and netvsc with old and new dpdks
- adds support for running dpdk with prebuilt shared gallery images (pkg-config)
- handles mana idiosyncrasy for setting lower/upper down
^ hack until NIC refactor/cleanup is ready

### DIFF
--- a/lisa/tools/pkgconfig.py
+++ b/lisa/tools/pkgconfig.py
@@ -1,0 +1,49 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from assertpy import assert_that
+from semver import VersionInfo
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.util import parse_version
+
+
+class Pkgconfig(Tool):
+    @property
+    def command(self) -> str:
+        return "pkg-config"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        assert isinstance(self.node.os, Posix)
+        self.node.os.install_packages("pkg-config")
+        return True
+
+    def package_info_exists(self, package_name: str) -> bool:
+        package_info_result = self.run(f"--modversion {package_name}", force_run=True)
+        return package_info_result.exit_code != 0
+
+    def get_package_info(
+        self,
+        package_name: str,
+        update_cached: bool = False,
+    ) -> str:
+        info_exists = self.package_info_exists(package_name=package_name)
+        assert_that(info_exists).described_as(
+            (
+                f"pkg-config information was not available for {package_name}. "
+                "This indicates an installation or package detection bug. "
+                f"ensure .pc file is available for {package_name} on this OS."
+            )
+        ).is_true()
+        return self.run(f"--modversion {package_name}").stdout
+
+    def get_package_version(
+        self, package_name: str, update_cached: bool = False
+    ) -> VersionInfo:
+        version_info = self.get_package_info(package_name, update_cached=update_cached)
+        return parse_version(version_info)
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -3,7 +3,7 @@
 
 import re
 from pathlib import PurePosixPath
-from typing import Any, List, Pattern, Tuple, Type, Union
+from typing import Any, List, Tuple, Type, Union
 
 from assertpy import assert_that, fail
 from semver import VersionInfo
@@ -20,6 +20,7 @@ from lisa.tools import (
     Lspci,
     Modprobe,
     Pidof,
+    Pkgconfig,
     Rm,
     Service,
     Tar,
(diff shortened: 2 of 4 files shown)


lisa/nic.py
```diff
@@ -437,6 +437,16 @@ def _load_nics(self) -> None:
                     self.nics[nic].pci_slot = pci_device.slot
                 break
 
+    def get_mac_address(self, nic_name: str) -> str:
```
squirrelsc: The mac information is in NicInfo already. Can it be reused?

lisa/tools/pkgconfig.py
```diff
@@ -0,0 +1,75 @@
...
+# Licensed under the MIT license.
+import re
+
+from assertpy import assert_that
+from semver import VersionInfo
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+
+coerce_version_info_regex = re.compile(
```
squirrelsc: Provide examples above.

lisa/tools/pkgconfig.py
```diff
@@ -0,0 +1,75 @@
...
+    def install(self) -> bool:
+        assert isinstance(self.node.os, Posix)
+        self.node.os.install_packages("pkg-config")
+        return True
+
+    def _coerce_version_info_string(self, version_string: str) -> VersionInfo:
+        if VersionInfo.isvalid(version_string):
+            return VersionInfo.parse(version_string)
+        # else, attempt to coerce
+        matches = coerce_version_info_regex.search(version_string)
```
squirrelsc: Does the `utils.parse_version` implement similar functionality?

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -148,18 +149,21 @@ def dependencies(self) -> List[Type[Tool]]:
...
-        if not dpdk_version:
+        if not self.has_dpdk_version():
             fail(
                 "Test suite bug: dpdk version was not set prior "
                 "to querying the version information."
             )
 
         # black doesn't like to direct return VersionInfo comparison
-        return bool(dpdk_version >= "19.11.0")  # appease the type checker
+        return bool(self.get_dpdk_version() >= "19.11.0")  # appease the type checker
```
squirrelsc, line 166: It doesn't need to convert the result to bool. It should be already.

squirrelsc: avoid writing comments in the same line. It's hard to read.

lisa/tools/pkgconfig.py
mcgov: it DOES
moved to using that, this also saves me from writing a regex example 🙂

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: fixed

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -435,14 +452,20 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             self.dpdk_path = self.node.get_pure_path(work_path).joinpath(
                 self._dpdk_repo_path_name
             )
-        self.find_testpmd_binary(assert_on_fail=False)
+        self._determine_network_hardware()
+        # if dpdk is already installed, find the binary and check the version
+        if self.find_testpmd_binary(assert_on_fail=False):
+            pkgconfig = self.node.tools[Pkgconfig]
+            if pkgconfig.get_package_info("libdpdk", check_exists=True):
```
squirrelsc: This method is confusing. If it's to check a package exists or not, it needs a method like `exists_package`

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -247,12 +252,16 @@ def generate_testpmd_command(
         #   --eth-peer=<port id>,<receiver peer MAC address> \
         #   --stats-period <display interval in seconds>
 
+        # txq and rxq will always be equal in our testing
         if multiple_queues:
-            txq = 4
-            rxq = 4
+            if self.is_mana:
+                queues = 8
```
squirrelsc, line 266: Use the previous name `rxq` may be clearer.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -247,12 +252,16 @@ def generate_testpmd_command(
         #   --eth-peer=<port id>,<receiver peer MAC address> \
         #   --stats-period <display interval in seconds>
 
+        # txq and rxq will always be equal in our testing
```
squirrelsc: It looks not true from below logic.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: I think queues is better as long as the value is shared for both the txq and rxq flags. I'll switch it back if we move to support some other setup.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: clarified comments

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: added package_info_exists

squirrelsc (approved):

PR #2899, opened 2023-08-14
Mana and NVMe enabled in the info field

kamalca:
These changes were cherry picked from `checkpoint_v3_1`.

Adds information about MANA and NVMe capabilities to the info field of LISA test results.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -426,9 +426,7 @@ class Nics(InitializableMixin):
         ).is_greater_than(0)
 
         # handle situation when there is no mana driver, but have mana pci devices
-        if self._is_mana_device_discovered() and not self._node.tools[
-            KernelConfig
-        ].is_enabled("CONFIG_MICROSOFT_MANA"):
+        if self.is_mana_device_present() and not self.is_mana_driver_enabled():
             pci_devices = lspci.get_devices_by_type(
                 constants.DEVICE_TYPE_SRIOV, force_run=True
             )
@@ -437,10 +435,7 @@ class Nics(InitializableMixin):
                     self.nics[nic].pci_slot = pci_device.slot
                 break
 
-    def is_mana_present(self) -> bool:
-        return self._is_mana_device_discovered()
-
-    def _is_mana_device_discovered(self) -> bool:
+    def is_mana_device_present(self) -> bool:
         lspci = self._node.tools[Lspci]
         pci_devices = lspci.get_devices_by_type(
             constants.DEVICE_TYPE_SRIOV, force_run=True
@@ -456,6 +451,9 @@ class Nics(InitializableMixin):
             else:
                 all_mana_devices = False
         return all_mana_devices
+
+    def is_mana_driver_enabled(self) -> bool:
+        return self._node.tools[KernelConfig].is_enabled("CONFIG_MICROSOFT_MANA")
 
     def _get_default_nic(self) -> None:
         self.default_nic: str = ""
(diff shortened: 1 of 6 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -732,6 +735,26 @@ def _get_node_information(self, node: Node) -> Dict[str, str]:
...
             node.log.debug(f"vm generation: {information[KEY_VM_GENERATION]}")
+            node.log.debug("detecting mana driver enabled...")
+            information[KEY_MANA_ENABLED] = node.tools[KernelConfig].is_enabled(
+                "CONFIG_MICROSOFT_MANA"
+            )
+            node.log.debug(f"mana enabled: {information[KEY_MANA_ENABLED]}")
+
+            node.log.debug("detecting nvme driver enabled...")
+
+            _has_nvme_core = node.tools[KernelConfig].is_built_in(
```
squirrelsc, line 745: Should they be test cases, instead of collecting information from all cases?

kamalca: I believe we concluded to continue with this approach?

squirrelsc: yes

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -732,6 +735,26 @@ def _get_node_information(self, node: Node) -> Dict[str, str]:
             node.log.debug("detecting vm generation...")
             information[KEY_VM_GENERATION] = node.tools[VmGeneration].get_generation()
             node.log.debug(f"vm generation: {information[KEY_VM_GENERATION]}")
+            node.log.debug("detecting mana driver enabled...")
+            information[KEY_MANA_ENABLED] = node.tools[KernelConfig].is_enabled(
```
squirrelsc: There should be a method in nic already, please use it.

kamalca: Even inside of nic.py, it looks like this is the way mana driver is checked.
``` python
# handle situation when there is no mana driver, but have mana pci devices
if self._is_mana_device_discovered() and not self._node.tools[
    KernelConfig
].is_enabled("CONFIG_MICROSOFT_MANA"):
```

squirrelsc: I mean it should reuse the method in nic.py. In nic.py, the _is_mana_device_discovered means the mana presents as a pci device. Here the code checks only if mana is in build config. Please double check the logic and try to reuse the nic.py.

kamalca: The goal is to know whether the mana driver is enabled in distros, not to know whether the mana device is actually present. I don't think there is any code in nic.py to reuse unless I created a function just for checking the mana kernel config.

squirrelsc: So, it needs to be added into nic, and make names clear, including `is_mana_driver_enabled`, `is_mana_device_present`.

lisa/nic.py
```diff
@@ -437,7 +435,7 @@ def _load_nics(self) -> None:
                     self.nics[nic].pci_slot = pci_device.slot
                 break
 
-    def is_mana_present(self) -> bool:
+    def is_mana_device_present(self) -> bool:
```
squirrelsc, line 438: We can change the original `_is_mana_device_discovered` to the new name, and remove this public method. It just call private method, no other logic.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -169,6 +170,8 @@
 KEY_WALA_VERSION = "wala_version"
 KEY_WALA_DISTRO_VERSION = "wala_distro"
 KEY_HARDWARE_PLATFORM = "hardware_platform"
+KEY_MANA_ENABLED = "mana_enabled"
```
squirrelsc: To be consistent, it should be `mana_driver_enabled`.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2902, opened 2023-08-15
Azure Orchestrator: load_environment: add error message

adityagesh:
Add error message if fetching vm details from vms_map fails

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1448,7 +1448,10 @@ def load_environment(
         assert isinstance(node, RemoteNode)
 
         vm = vms_map.get(node.name)
-        assert_that(vm).is_not_none()
+        assert_that(vm).described_as(
+            f"Cannot find vm with name {node.name}. Make sure the VM exists in "
+            f"resource group {resource_group_name}"
+        ).is_not_none()
 
         node_context = get_node_context(node)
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1448,7 +1448,9 @@ def load_environment(
         assert isinstance(node, RemoteNode)
 
         vm = vms_map.get(node.name)
-        assert_that(vm).is_not_none()
+        assert_that(vm).described_as(
+            f"VM details for {node.name} is {vm}. This is an unexpected failure"
```
squirrelsc: This message just prints values, no reason or solution. It should be like, "Cannot find vm with name {node.name}. Make sure the VM exists in resource group {resource_group_name}"

adityagesh: @squirrelsc I initially thought of something like this, but 
vms = compute_client.virtual_machines.list(resource_group_name) 
this api dynamically fetches the vm from the resource group, so it is not possible to get a vm that is not available in rg, right?

squirrelsc: Yes, so if the vm is not found. The error message indicates the expected vm in a resource group.

adityagesh: Okay, Updated

squirrelsc (approved):

PR #2906, opened 2023-08-16
Nic: cleanup node nic info gathering

mcgov:
Move to using ethtool, ip, and lspci tools (now that they exist) to gather nic info. This avoids some issues when gathering nic info on systems like MANA where there are many interfaces with the same bus info.
- directly call `ip` to get master/slave pairs rather than inferring from sysfs. Each master interface is identified by having a child interface, rather than inferring it the other way around.
- use ethtool to fetch bus-info (aka 'pci address' in LISA).

This cleans up a lot of the logic, rather than making one large call and regex parsing the output of `ip`. These calls are run in parallel where possible.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -6,14 +6,17 @@ import os
 import re
 from collections import OrderedDict
 from dataclasses import dataclass
+from functools import partial
 from pathlib import PurePosixPath
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
 from assertpy import assert_that
 from retry import retry
 
-from lisa.tools import Cat, Ip, KernelConfig, Ls, Lspci, Modprobe, Tee
+from lisa.tools import Cat, Ethtool, Ip, KernelConfig, Ls, Lspci, Modprobe, Tee
+from lisa.tools.ip import IpInfo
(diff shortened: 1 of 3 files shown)


lisa/nic.py
```diff
@@ -358,67 +363,55 @@ def _get_nic_uuids(self) -> None:
 
     def _load_nics(self) -> None:
         # Identify which nics are slaved to master devices.
-        # This should be really simple with /usr/bin/ip but experience shows
-        # the tool isn't super consistent across distros in this regard
 
-        # use sysfs to gather synthetic/pci nic pairings and pci slot info
-        nic_info_fetch_cmd = "ls -la /sys/class/net/*/lower*/device"
         self._node.log.debug(f"Gathering NIC information on {self._node.name}.")
+        nic_names = self._get_nic_names()
```
squirrelsc, line 368: @LiliDeng please check if there is any regression.

mcgov: @squirrelsc  @LiliDeng 
If there hasn't been any regression could we merge and give this a shot this week? We can always back it out before including in the release candidate.

mcgov:
@squirrelsc @LiliDeng Any update on regression testing?

PR #2907, opened 2023-08-16
Fix fio test for freebsd

dsrivastavv:


lisa/tools/lsblk.py
```diff
@@ -100,6 +100,16 @@ def is_mounted(self) -> bool:
 
         return any(partition.mountpoint for partition in self.partitions)
 
+    @property
+    def is_resource_disk(self) -> bool:
```
squirrelsc: It's Azure specific, it should be in Azure feature.

lisa/tools/mkfs.py
```diff
@@ -102,3 +120,60 @@ def _install(self) -> bool:
             package = "btrfsprogs"
         posix_os.install_packages(package)
         return self._check_exists()
+
+
+class BSDMkfs(Mkfs):
+    _PARTITION_ID_REGEX = re.compile(r"^(?P<disk>/dev/(da|nvd)\d+)p(?P<index>\d+)$")
```
squirrelsc, line 128: Add examples above.

lisa/tools/mkfs.py
```diff
@@ -2,17 +2,31 @@
...
+    [
+        "xfs",
+        "ext2",
+        "ext3",
+        "ext4",
+        "btrfs",
+        "hugetlbfs",
+        "nfs",
+        "tracefs",
+        "ufs",
```
LiliDeng, line 22: I recall for freebsd, it also has zfs file system type?

dsrivastavv: Fio tests work with the most common disk format for the distro, which for bsd is `ufs`. There are other file systems as well, but they are currently not required so I skipped adding them.

lisa/tools/mkfs.py
dsrivastavv: @squirrelsc I have fixed this comment. Can you please check?

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2908, opened 2023-08-17
A couple fixes for SSH access onto 'minimal' nodes

glima:
In one LISA derivative we have, the nodes at the other side of the SSH channel will be of the 'minimal' type, for real.
We mean non-POSIX, very barebones. These are fixes to make that setup work, while not harming the general case.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -55,6 +55,7 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
         is_test_target: bool = True,
         base_part_path: Optional[Path] = None,
         parent_logger: Optional[Logger] = None,
+        encoding: str = "utf-8",
     ) -> None:
         super().__init__(runbook=runbook)
         self.is_default = runbook.is_default
@@ -65,6 +66,7 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
         self.provision_time: float
         self._shell: Optional[Shell] = None
         self._first_initialize: bool = False
+        self._encoding = encoding
 
         # will be initialized by platform
         self.features: Features
@@ -499,6 +501,7 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
             no_error_log=no_error_log,
             no_info_log=no_info_log,
             no_debug_log=no_debug_log,
+            encoding=self._encoding,
             cwd=cwd,
             update_envs=update_envs,
         )
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -429,10 +429,11 @@ class Process:
             isinstance(self._shell, SshShell)
             and self._shell.spawn_initialization_error_string
         ):
-            raw_input = re.sub(
-                re.compile(rf"{self._shell.spawn_initialization_error_string}\r\n"),
-                "",
-                raw_input,
+            raw_input = raw_input.replace(
+                rf"{self._shell.spawn_initialization_error_string}\n", ""
+            )
+            raw_input = raw_input.replace(
+                rf"{self._shell.spawn_initialization_error_string}\r\n", ""
             )
             self._log.debug(
                 "filter the profile error string: "
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -327,6 +327,13 @@ class Process:
         return self._result
 
     def kill(self) -> None:
+        if (
+            isinstance(self._shell, SshShell)
+            and self._shell._inner_shell
+            and self._shell._inner_shell._spur._shell_type
+            == spur.ssh.ShellTypes.minimal
+        ):
+            return
         if self._process:
             self._log.debug(f"Killing process : {self._id_}")
             try:
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -108,6 +108,7 @@ class Process:
         no_error_log: bool = False,
         no_info_log: bool = False,
         no_debug_log: bool = False,
+        encoding: str = "utf-8",
     ) -> None:
         """
         command include all parameters also.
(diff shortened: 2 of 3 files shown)


lisa/util/shell.py
```diff
@@ -340,6 +350,12 @@ def spawn(
                 # Except CommandInitializationError then use minimal shell type.
                 if not have_tried_minimal_type:
                     self._inner_shell._spur._shell_type = spur.ssh.ShellTypes.minimal
+
+                    funcType = type(spur.ssh.ShellTypes.minimal.generate_run_command)
```
squirrelsc: Can you add code comments above to explain purpose of added lines?

lisa/util/process.py
```diff
@@ -421,10 +429,8 @@ def _filter_profile_error(self, raw_input: str) -> str:
             isinstance(self._shell, SshShell)
             and self._shell.spawn_initialization_error_string
         ):
-            raw_input = re.sub(
-                re.compile(rf"{self._shell.spawn_initialization_error_string}\r\n"),
-                "",
-                raw_input,
+            raw_input = raw_input.replace(
+                rf"{self._shell.spawn_initialization_error_string}\n", ""
```
squirrelsc, line 433: Please handle both `\r\n` and `\n`. It may be related LISA running from Windows or Linux.

lisa/node.py
```diff
@@ -267,6 +267,7 @@ def execute(
         no_debug_log: bool = False,
         cwd: Optional[PurePath] = None,
         timeout: int = 600,
+        encoding: str = "utf-8",
```
squirrelsc: Because the encoding of node won't be changed, so it can be set when node is created.

lisa/util/shell.py
```diff
@@ -340,6 +355,20 @@ def spawn(
...
                 if not have_tried_minimal_type:
                     self._inner_shell._spur._shell_type = spur.ssh.ShellTypes.minimal
+
+                    # Dynamically override that object's method. Here,
+                    # we don't enclose every shell token under single
+                    # quotes anymore. That's an assumption from spur
+                    # that minimal shells will still be POSIX
+                    # compliant--not true for some cases for LISA
+                    # users.
+                    func_type = type(spur.ssh.ShellTypes.minimal.generate_run_command)
```
squirrelsc, line 365: @lubaihua33 Please check if there is regression on distro, which you tested with minimal shell type.

lisa/node.py
glima: OK, I can store this state in the node.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@lubaihua33 I have tested the common gallery image, it didn't have any regression, could you please check on the image which only supports minimal shell type, thanks!

squirrelsc:
> @lubaihua33 I have tested the common gallery image, it didn't have any regression, could you please check on the image which only supports minimal shell type, thanks!

Let's merge today, and test later. The impact won't be big, if it fails on other minimal shells.

LiliDeng (approved):

lubaihua33:
> > @lubaihua33 I have tested the common gallery image, it didn't have any regression, could you please check on the image which only supports minimal shell type, thanks!
> 
> Let's merge today, and test later. The impact won't be big, if it fails on other minimal shells.

I tested one image "javlinltd1579185328273 cloverdx-server-vm-image 6-1-0 1.0.0". This PR has regression on this image. 
1. Below modifies should be added:
raw_input = raw_input.replace(
                **rf**"{self._shell.spawn_initialization_error_string}\r\n", ""
            )
should be modified as:
raw_input = raw_input.replace(
                **f**"{self._shell.spawn_initialization_error_string}\r\n", ""
            )

2. The change of the below line will lead to a imeout for some commands like "'sudo' 'sh' '-c' 'cat /etc/shadow'"
	def minimal_escape_sh(value: str) -> str:
	    return value.replace("'", "'\\''")
	
	For example:
	command_args is like ['sudo', 'sh', '-c', 'cat /etc/shadow | gr..."lisatest"']
	
	The return value of the original fun 'generate_run_command' is
        '\'sudo\' \'sh\' \'-c\' \'cat /etc/shadow | grep -v "lisatest"\''.
       The value of the override one minimal_generate_run_command is
       'sudo sh -c cat /etc/shadow | grep -v "lisatest"',
       which will lead to a timeout. Because the system recognizes this command as  'sudo sh -c cat'.

glima:
> 
> I tested one image "javlinltd1579185328273 cloverdx-server-vm-image 6-1-0 1.0.0". This PR has regression on this image.
> 
> 1. Below modifies should be added:
>    raw_input = raw_input.replace(
>    **rf**"{self._shell.spawn_initialization_error_string}\r\n", ""
>    )
>    should be modified as:
>    raw_input = raw_input.replace(
>    **f**"{self._shell.spawn_initialization_error_string}\r\n", ""
>    )

Ok, minimal but we can fix, ty.

> 2. The change of the below line will lead to a imeout for some commands like "'sudo' 'sh' '-c' 'cat /etc/shadow'"
>    def minimal_escape_sh(value: str) -> str:
>    return value.replace("'", "'\''")
>    For example:
>    command_args is like ['sudo', 'sh', '-c', 'cat /etc/shadow | gr..."lisatest"']
>    The return value of the original fun 'generate_run_command' is
>    ''sudo' 'sh' '-c' 'cat /etc/shadow | grep -v "lisatest"''.
>    The value of the override one minimal_generate_run_command is
>    'sudo sh -c cat /etc/shadow | grep -v "lisatest"',
>    which will lead to a timeout. Because the system recognizes this command as  'sudo sh -c cat'.

Is it problematic for pipes only? The problem is tokenizing like that for non-POSIX shells will make them also unusable. My use case has no parser for `'command'`, it expects plain `command` instead. We need to find a middle ground and support for non-POSIX shells too.

lubaihua33:
@glima 
It is problematic not only for pipes. It is for the scenario that 'shell=true' that "'sh' '-c'" is added in the command.
For example:
if the command_args is like ['sudo', 'sh', '-c', 'cat /etc/shadow'], then the return value of the override fun minimal_generate_run_command is "sudo sh -c cat /etc/shadow". There is no apostrophe for the command "cat /etc/shadow". Then the Linux system (the tested VM) recognizes this command as 'sudo sh -c cat ', which leads to a timeout.

PR #2909, opened 2023-08-18
HPC Suite: Execute in Azure only for specialized images

adityagesh:
HPC is a specialized SKU in Azure, hence it has to run only on Images specialized for HPC.
"/opt/azurehpc/component_versions.txt" is a file that is present in Azure HPC images, even though it is not publicly documented.
At present there is no other way to identify if the image is an Azure HPC image.


adityagesh:
@hcoona FYI and to monitor the discussions

LiliDeng:
there is a condition check during installing ofed driver based on this file, we need know whether there are customers using RDMA with non-hpc images or not, then decide to skip on non-hpc images or not.

LiliDeng:
as we discussed, we don't want to miss the test coverage, so close it @adityagesh

PR #2910, opened 2023-08-18
nested cvm azure host precheck and remove color code from dmesg output

smit-gardhariya:
This PR contains 2 different commit for below things

Remove color code from dmesg output
Add testcase to check SEV-SNP enablement on azure vm (host vm for nested case)


microsoft/testsuites/cvm/cvm_azure_host.py
```diff
@@ -0,0 +1,58 @@
...
+    def verify_azure_vm_snp_enablement(
+        self,
+        log: Logger,
+        node: Node,
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        sev_enabled_pattern = r"mshv: SEV-SNP is supported"
```
LiliDeng: the normal azure vm will fail in this case or not?

smit-gardhariya: yes, i guess, it will if we dont find that message. This supposed to run only for nested CVM supported SKU and for Dom0 image only.

smit-gardhariya: I can not skip the testcase because it has to be failed if it does not find the message. What we can do is, if OS is Mariner then only we can execute it. Let me know

LiliDeng: ues sev_partition_pattern =re.compile(r"mshv: Maximum supported SEV-SNP partitions are: (\d+)") in class level, then use get_matched_str some existing methods

smit-gardhariya: sure Lili, will make that change

smit-gardhariya: @LiliDeng Added before case for Mariner so it wont affect other distro

smit-gardhariya: @LiliDeng Done

microsoft/testsuites/cvm/cvm_azure_host.py
```diff
@@ -0,0 +1,81 @@
...
+    __sev_partition_pattern = re.compile(
+        r"mshv: Maximum supported SEV-SNP partitions are: (\d+)"
+    )
+
+    def before_case(self, log: Logger, **kwargs: Any) -> None:
+        node = kwargs["node"]
+        if not isinstance(node.os, (CBLMariner)):
+            raise SkippedException(
+                f"CVMAzureHostTestSuite is not implemented for {node.os.name}"
+            )
```
LiliDeng, line 43: @squirrelsc overall the code is good, per @smit-gardhariya the case will pass on mariner + DCas_cc_v5 vm sizes.
So the only concern is the test case will fail when it runs against mariner image + non DCas_cc_v5 vm size. if we can't create a feature based on vm size DCas_cc_v5, we have to judge whether the size belongs to this vm sku after vm deployment. then decide skip or continue the testing. thoughts?

smit-gardhariya: Dom0 flavor of mariner to be precise here

smit-gardhariya: Other way can be, I can pass SKU from our testing infra as variable, and we can mark this test SKIPPED if that is None or not as per expectation (some pre-defined listed SKUs)

lisa/tools/dmesg.py
```diff
@@ -37,7 +37,8 @@ def _check_exists(self) -> bool:
 
     def get_output(self, force_run: bool = False) -> str:
         command_output = self._run(force_run=force_run)
-        return command_output.stdout
+        stdout = re.sub(self.__color_code_pattern, "", command_output.stdout)
```
squirrelsc: Please change it to use `util.filter_ansi_escape`.

smit-gardhariya: @squirrelsc Done in latest push. Please check.

microsoft/testsuites/cvm/cvm_azure_host.py
smit-gardhariya: @squirrelsc Could you please share your thought here.

lisa/tools/dmesg.py
squirrelsc: Please help refine existing code, to remove duplicate pattern.

microsoft/testsuites/cvm/cvm_azure_host.py
squirrelsc: You can add an azure only feature, and enable this test case for specified VM sizes only.

smit-gardhariya: @squirrelsc @LiliDeng I modified before_case to add validation on VM SKU. This SKU is not publicly available yet, so i am checking it in before case instead of adding it in azure feature where we can check for SKU family.

lisa/tools/dmesg.py
smit-gardhariya: @squirrelsc Done in latest push. Please check

microsoft/testsuites/cvm/cvm_azure_host.py
squirrelsc: Please use a feature, and make vm sizes are configurable. So it can be rewritten by runbook.

smit-gardhariya: @squirrelsc Thnx for the review here. Could you please review the latest push. I have made the changes for creating feature CVMNestedVirtulization

LiliDeng (approved):

PR #2912, opened 2023-08-21
add baremetal platform

LiliDeng:


microsoft/runbook/bare_metal.yml
```diff
@@ -0,0 +1,70 @@
+name: $(test_name)
```
squirrelsc: bare metal is highly customizable, so we cannot provide a sample runbook. Please move samples to document. We need to provide current supported components in documents.

lisa/util/shell.py
```diff
@@ -241,9 +282,15 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
 
         if stdout_content and "Windows" in stdout_content:
             self.is_posix = False
+            self.is_bmc = False
```
squirrelsc: Try if minimal shell type is useful.

lisa/operating_system.py
```diff
@@ -129,10 +129,11 @@ class OperatingSystem:
 
     __posix_factory: Optional[Factory[Any]] = None
 
-    def __init__(self, node: "Node", is_posix: bool) -> None:
+    def __init__(self, node: "Node", is_posix: bool, is_bmc: bool = False) -> None:
         super().__init__()
         self._node: Node = node
         self._is_posix = is_posix
+        self.is_bmc = is_bmc
```
squirrelsc: Does it conflict with posix or not? If it doesn't conflict, it means there are 4 combinations with posix. So Does all of them supported?

lisa/operating_system.py
```diff
@@ -153,30 +154,33 @@ def create(cls, node: "Node") -> Any:
...
-                if os_info_item:
-                    os_infos.append(os_info_item)
-                    for sub_type in posix_factory.values():
-                        posix_type: Type[Posix] = sub_type
-                        pattern = posix_type.name_pattern()
-                        if pattern.findall(os_info_item):
-                            detected_info = os_info_item
-                            result = posix_type(node)
-                            matched = True
+            if node.shell.is_bmc:
```
squirrelsc: Can it use detect string flow to get the version?

lisa/operating_system.py
```diff
@@ -663,6 +667,11 @@ class BSD(Posix):
     ...
 
 
+class BMC(Posix):
+    def capture_system_information(self, saved_path: Path) -> None:
```
squirrelsc: The MacOS are not bypass this method, why BMC need it?

lisa/sut_orchestrator/__init__.py
```diff
@@ -7,3 +7,4 @@
 CLOUD_HYPERVISOR = "cloud-hypervisor"
 AWS = "aws"
 READY = PLATFORM_READY
+BAREMETAL = "baremetal"
```
squirrelsc: Use the constant string in constants.py like PLATFOMR_READY.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+from lisa.util.shell import try_connect
+from lisa.util.subclasses import Factory
+
+from .. import BAREMETAL
+from . import features
+
+
+@dataclass_json()
+@dataclass
+class SourceBuildSchema:
```
squirrelsc: Should it be removed?

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+    ) -> None:
+        super().__init__(runbook=runbook)
+
+    @classmethod
+    def type_name(cls) -> str:
+        return BAREMETAL
+
+    @classmethod
+    def supported_features(cls) -> List[Type[feature.Feature]]:
+        return [features.StartStop]
```
squirrelsc: Will you included SerialConsole later?

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,74 @@
...
+    username: Optional[str] = field(default="")
+    password: Optional[str] = field(default="")
+    private_key_file: Optional[str] = field(default="")
+    port: Optional[int] = field(default=22)
+    iso_http_url: Optional[str] = field(default="")
+
+
+@dataclass_json()
+@dataclass
+class ClusterSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
```
squirrelsc: Move the schema into common.py, so it avoids circle reference.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+
+@dataclass_json()
+@dataclass
+class SourceBuildSchema:
+    source: Optional[SourceSchema] = None
+
+
+@dataclass_json()
+@dataclass
+class BareMetalPlatformSchema:
```
squirrelsc: Move schemas into common.py or create a file called schema.py under baremetal.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+        assert baremetal_runbook, "platform runbook cannot be empty"
+        self._baremetal_runbook = baremetal_runbook
+
+    def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        return True
+
+    def _deploy_environment(self, environment: Environment, log: Logger) -> None:
+        # copy build (shared, check if it's copied)
+        source_path: str = ""
+        if self._baremetal_runbook.source and self._baremetal_runbook.source:
```
squirrelsc: Dulplicated conditions.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) Microsoft Corporation.
```
squirrelsc, line 1: The cluster folder should put only "cluster" classes. So the ip_getter, ready_checker should be in another folder.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+        # ready checker cleanup
+        if cluster.ready_checker:
+            ready_checker_runbook = ready_checker_factory.create_by_runbook(
+                cluster.ready_checker
+            )
+            ready_checker_runbook.clean_up()
+
+        # copy build if source exists
+        cluster_runbook = cluster_factory.create_by_runbook(cluster)
+        cluster_runbook.initialize(source_path=source_path)
```
squirrelsc: The "copy" step shouldn't be in `_initialize`. It's not easy to understand the logic. So it needs a method called `copy`.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+        cluster = self._baremetal_runbook.cluster
+        # ready checker cleanup
+        if cluster.ready_checker:
+            ready_checker_runbook = ready_checker_factory.create_by_runbook(
+                cluster.ready_checker
+            )
+            ready_checker_runbook.clean_up()
+
+        # copy build if source exists
+        cluster_runbook = cluster_factory.create_by_runbook(cluster)
```
squirrelsc: The created instance is the cluster itself, not a "runbook". The variable name should be `cluster` without "runbook".

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+        if self._baremetal_runbook.source and self._baremetal_runbook.source:
+            source_factory = Factory[Source](Source)
+            source = source_factory.create_by_runbook(self._baremetal_runbook.source)
+            self._log.debug(f"found build '{source.type_name()}', to expand runbook.")
+            source.initialize()
+            source_path = source.source_path
+
+        ready_checker_factory = Factory[ReadyChecker](ReadyChecker)
+        cluster_factory = Factory[Cluster](Cluster)
+        ip_getter_factory = Factory[IpGetterChecker](IpGetterChecker)
```
squirrelsc: Move factories to be an object in baremetal, so it doesn't need to be created everytime, when an instance is needed.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+        if cluster.ready_checker:
+            ready_checker_runbook = ready_checker_factory.create_by_runbook(
+                cluster.ready_checker
+            )
+            ready_checker_runbook.clean_up()
+
+        # copy build if source exists
+        cluster_runbook = cluster_factory.create_by_runbook(cluster)
+        cluster_runbook.initialize(source_path=source_path)
+
```
squirrelsc, line 66: When a build was copied, it should be saved, and avoid to copy again when deploy next environment.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+                lambda: ready_checker_runbook.is_ready() is True,
+                timeout_message="wait for ready check ready",
+            )
+
+        # get ip address
+        if cluster.ip_getter:
+            ip_getter_runbook = ip_getter_factory.create_by_runbook(cluster.ip_getter)
+            ip_address = ip_getter_runbook.get_ip()
+
+        if cluster_runbook.runbook.client.ip_address:
```
squirrelsc: If the ip_address is set, it should be used before trying to use ip_getter. If none of them set, there should be an exception. So it's like,

```python
if cluster_runbook.runbook.client.ip_address:
    ...
elif cluster.ip_getter:
    ...
else:
    raise LISAException(...)
```

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+
+        assert environment.runbook.nodes_requirement
+        for node_space in environment.runbook.nodes_requirement:
+            assert isinstance(
+                node_space, schema.NodeSpace
+            ), f"actual: {type(node_space)}"
+            environment.create_node_from_requirement(node_space)
+
+        index = 0
+        for node in environment.nodes.list():
```
squirrelsc: The loop should include the ready checker and ip getter. Considering there maybe multiple clients in an environment.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+                public_port=cluster_runbook.runbook.client.port,
+                **fields_to_dict(
+                    connection_info,
+                    ["address", "port", "username", "password", "private_key_file"],
+                ),
+            )
+            try_connect(connection_info)
+            index = index + 1
+
+        self._log.debug("deploy environment successfully")
```
squirrelsc: Print the environment name together.

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
```diff
@@ -0,0 +1,64 @@
...
+    @classmethod
+    def type_name(cls) -> str:
+        return "rackmanager"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return RackManagerSchema
+
+    def deploy(self) -> Any:
+        self._log.debug("deploy client through rackmanager")
```
squirrelsc: The cluster type should be printed in platform level, so it doesn't need to be replicated in every cluster.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,138 @@
...
+@dataclass
+class SourceBuildSchema:
+    source: Optional[SourceSchema] = None
+
+
+@dataclass_json()
+@dataclass
+class BareMetalPlatformSchema:
+    source: Optional[SourceSchema] = field(default=None)
+    cluster: ClusterSchema = field(default_factory=ClusterSchema)
```
squirrelsc: As we discussed, the cluster should be an array. We can deal with first one so far, but the schema should be extendable.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,74 @@
...
+            if self.cluster_runbook.build:
+                build = build_factory.create_by_runbook(self.cluster_runbook.build)
+                build.initialize(
+                    source_path=source_path,
+                    files=self.cluster_runbook.build.files,
+                )
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "cluster"
```
squirrelsc: It doesn't need to be set in base class, since it's must have in subclasses.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,74 @@
...
+class Cluster(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
+    def __init__(
+        self,
+        runbook: ClusterSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.cluster_runbook: ClusterSchema = self.runbook
+        self._log = get_logger("cluster", self.__class__.__name__)
+
+    def initialize(self, *args: Any, **kwargs: Any) -> None:
```
squirrelsc: Don't overwrite the `initialize` anytime, overwrite `_initialize` instead.

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
```diff
@@ -0,0 +1,64 @@
...
+        rack_manager.set_connection_info(
+            public_port=self.rm_runbook.port,
+            **fields_to_dict(
+                connection_info, ["address", "port", "username", "password"]
+            ),
+        )
+        try_connect(connection_info)
+        assert self.rm_runbook.client
+        rack_manager.execute(
+            f"set system reset -i {self.rm_runbook.client.management_port}"
```
squirrelsc: the client should be passed in, instead of hardcode here. There maybe multiple clients in a rack manager cluster.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "cluster"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return ClusterSchema
+
+    def deploy(self) -> Any:
```
squirrelsc: The `deploy` method should accept an environment parameter. Each node may bring information like management port of rack manager.

lisa/sut_orchestrator/baremetal/cluster/readychecker.py
```diff
@@ -0,0 +1,78 @@
...
+        return FileSingleSchema
+
+    def clean_up(self) -> None:
+        if os.path.exists(self.file_single_runbook.file):
+            os.remove(self.file_single_runbook.file)
+            self._log.debug(
+                f"The file {self.file_single_runbook.file} has been removed"
+            )
+        else:
+            self._log.debug(f"The file {self.file_single_runbook.file} does not exist")
```
squirrelsc: `"The file {self.file_single_runbook.file} does not exist, so it doesn't need to be cleaned up."`

lisa/sut_orchestrator/baremetal/cluster/ip_getter.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@dataclass_json()
+@dataclass
+class FileSingleSchema(IpGetterSchema):
+    file: str = ""
+
+
+class FileSingleChecker(IpGetterChecker):
+    __ip_addr_regex = re.compile(r"(?P<ip_addr>[\d.]+)", re.M)
```
squirrelsc: Add examples above.

lisa/sut_orchestrator/baremetal/cluster/ip_getter.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(
+        self,
+        runbook: IpGetterSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.ip_getter_runbook: IpGetterSchema = self.runbook
+        self._log = get_logger("ip_getter", self.__class__.__name__)
+
+    @classmethod
+    def type_name(cls) -> str:
```
squirrelsc: It doesn't need to be specified at base class level.

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        self._log.debug(f"set system reset -i {self.rm_runbook.client.management_port}")
```
squirrelsc: Log some meaningful information, if it's needed.

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def deploy(self) -> Any:
+        self._log.debug("deploy client through rackmanager")
+        connection_info = schema.ConnectionInfo(
+            address=self.rm_runbook.address,
+            port=self.rm_runbook.port,
+            username=self.rm_runbook.username,
+            password=self.rm_runbook.password,
+        )
+        rack_manager = RemoteNode(
```
squirrelsc: use `quick_connect` in node.py.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -0,0 +1,134 @@
...
+            password=self.idrac_runbook.password,
+        )
+        self.login()
+        self.change_boot_order_once("VCD-DVD")
+        assert self.idrac_runbook.client
+        assert self.idrac_runbook.client.iso_http_url
+        self.insert_virtual_media(self.idrac_runbook.client.iso_http_url)
+        self.reboot()
+
+    def login(self) -> None:
```
squirrelsc: Make it private, if it doesn't need to be called out of the class.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -0,0 +1,134 @@
...
+    @classmethod
+    def type_name(cls) -> str:
+        return "idrac"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return IdracSchema
+
+    def deploy(self) -> Any:
+        self.redfish_instance = redfish.redfish_client(
```
squirrelsc: If it's onetime used, don't save it in class.

lisa/sut_orchestrator/baremetal/common.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@dataclass
+class EnvironmentContext:
+    ...
+
+
+@dataclass
+class NodeContext:
+    ip_address: str = ""
```
squirrelsc: Use connecitonInfo, instead of create another similar schema.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -0,0 +1,134 @@
...
+        self.reboot()
+
+    def login(self) -> None:
+        self.redfish_instance.login(auth="session")
+        self._log.debug(f"Login to {self.redfish_instance.get_base_url()} successful.")
+
+    def exit(self) -> None:
+        self.redfish_instance.logout()
+
+    def wait_for_completion(self, response: Any) -> None:
```
squirrelsc: Add a timeout parameter to prevent wait forever.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
squirrelsc: It still needs the cluster folder to put idrac and rack manager.

lisa/sut_orchestrator/baremetal/ip_getter.py
```diff
@@ -0,0 +1,69 @@
...
+    @classmethod
+    def type_name(cls) -> str:
+        return "file_single"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return FileSingleSchema
+
+    def get_ip(self) -> str:
+        self._log.debug(self.file_single_runbook.file)
```
squirrelsc: don't print value only, add meaningful description for the value.

lisa/sut_orchestrator/baremetal/source.py
```diff
@@ -0,0 +1,79 @@
...
+        with open(
+            artifact_path,
+            "wb",
+        ) as download_file:
+            response = requests.get(
+                download_url, auth=("", personal_access_token), timeout=600
+            )
+            download_file.write(response.content)
+
+        if file_extension == "zip":
```
squirrelsc: Don't handle the unzip implicitly. Maybe some system doesn't want to extract files. The actual copy/extract actions should be in build class, not in the source class. The source class should describe the source files only.

pupacha (commented):
Firstly, thanks @LiliDeng for bringing up BareMetal platform support in LISA this fast.

I know this PR is still in Draft mode and not yet ready for review. But I have a comment. Looking high level at schema and some of the code, I feel `client` term is slightly confusing to be used to represent a BareMetal server node.  Any specific reason for using the terminology? Is there a possibility to use a different terminology say `node` ( `node` in a `cluster` terminology )?

squirrelsc:
> Firstly, thanks @LiliDeng for bringing up BareMetal platform support in LISA this fast.
> 
> I know this PR is still in Draft mode and not yet ready for review. But I have a comment. Looking high level at schema and some of the code, I feel `client` term is slightly confusing to be used to represent a BareMetal server node. Any specific reason for using the terminology? Is there a possibility to use a different terminology say `node` ( `node` in a `cluster` terminology )?

The bare metal schema is trying to cover different scenarios. In iDRAC, there is only one machine with a management port. So, the server, client, node are the same one. In other scenarios, like the Rack Manager, one rack manager may manage several servers. In some deployment, there is another side server to provide boot images, private network, DHCP and so on.

The client here is to describe the assignable resources. The "node" may be confusing on server node or client node. So it's called client as an array.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
squirrelsc: The cluster base class shouldn't be in this folder. Because it may be referenced by other files, it causes cycle imports.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -0,0 +1,92 @@
...
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return SMBBuildSchema
+
+    def copy(self, source_path: str, files_map: List[FileSchema]) -> None:
+        username = self.smb_runbook.username
+        password = self.smb_runbook.password
+        server_name = self.smb_runbook.server_name
+        share_name = self.smb_runbook.share
+
+        conn = SMBConnection(
```
squirrelsc: Check if the SMBConnection support context, so it can be used by `with`. So the connection will be closed, if there is any exception.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,48 @@
...
+class Cluster(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
+    def __init__(
+        self,
+        runbook: ClusterSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.cluster_runbook: ClusterSchema = self.runbook
+        self._log = get_logger("cluster", self.__class__.__name__)
+
+    def copy(self, source_path: str) -> None:
```
squirrelsc: The copy method should be the same for all clusters. And as the logic, it should be an action of platform, not the cluster.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,153 @@
...
+        if cluster_instance.ready_checker:
+            ready_checker = self.ready_checker_factory.create_by_runbook(
+                cluster_instance.ready_checker
+            )
+            ready_checker.clean_up()
+
+        # copy build if source exists
+        if not self.copy_build:
+            cluster.copy(source_path=self.source_path)
+            self.copy_build = True
```
squirrelsc: This flag shouldn't be in platform level, it should mark a build is copied to a cluster. if it's in platform and there are two clusters, the build is not able to be copied on second cluster.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,153 @@
...
+
+        index = 0
+        for node in environment.nodes.list():
+            # ready checker
+            if ready_checker:
+                ready_checker.is_ready(node)
+
+            # get ip address
+            if cluster_instance.ip_getter:
+                ip_getter_runbook = self.ip_getter_factory.create_by_runbook(
```
squirrelsc: The created object is not a `runbook`, so it should be called like `ip_getter`.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,153 @@
...
+        # deploy cluster
+        cluster.deploy(environment)
+
+        if cluster_instance.ready_checker:
+            ready_checker = self.ready_checker_factory.create_by_runbook(
+                cluster_instance.ready_checker
+            )
+
+        index = 0
+        for node in environment.nodes.list():
```
squirrelsc: use it like
`for index, node in enumerate(environment.nodes.list()):`

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,153 @@
...
+                    index
+                ].connection.address = ip_getter_runbook.get_ip()
+
+            assert cluster.runbook.client[
+                index
+            ].connection, "client connection is empty"
+            assert cluster.runbook.client[
+                index
+            ].connection.address, "ip address is empty"
+            assert isinstance(node, RemoteNode)
```
squirrelsc: Add error message for assert. If it's to assert the type, print the actual type as well. There are many examples in LISA, search for `assert isinstance`

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,153 @@
...
+        for node_space in environment.runbook.nodes_requirement:
+            assert isinstance(
+                node_space, schema.NodeSpace
+            ), f"actual: {type(node_space)}"
+            environment.create_node_from_requirement(node_space)
+
+        index = 0
+        for node in environment.nodes.list():
+            node_context = get_node_context(node)
+            node_context.management_port = cluster.runbook.client[index].management_port
```
squirrelsc: The management port should be used by rack manager only, so it shouldn't be assigned in platform level. And the management_port should be managed by platform as the resource. The platform needs to track if a resource is allocated or not. If no more client is available, the platform should raise NoResource exception.

lisa/sut_orchestrator/baremetal/source.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "ado"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return ADOSourceSchema
+
+    def download(self) -> str:
```
squirrelsc, line 48: The different source may need different implementation, and to be called in build class.

lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -0,0 +1,107 @@
...
+    name: str = ""
+
+
+@dataclass_json()
+@dataclass
+class ADOSourceSchema(SourceSchema):
+    organization_url: str = ""
+    project: str = ""
+    build_id: int = 0
+    pat: str = ""
```
squirrelsc: Mark the pat as a secret. Same to other fields like password.

lisa/sut_orchestrator/baremetal/readychecker.py
```diff
@@ -0,0 +1,127 @@
...
+
+        connection_info = schema.ConnectionInfo(
+            address=context.address,
+            port=context.port,
+            username=context.username,
+            private_key_file=context.private_key_file,
+            password=context.password,
+        )
+        remote_node.set_connection_info(
+            public_port=context.port,
```
squirrelsc: Why handle the public_port separately? The port will be used, if public_port is not set.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,48 @@
...
+            self._log.debug("no build to copy")
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return ClusterSchema
+
+    def deploy(self, environment: Environment) -> Any:
+        raise NotImplementedError()
+
+    def has_serial_console(self) -> Any:
```
squirrelsc: the return value should be bool.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,48 @@
...
+        pass
+
+    def get_serial_console(self) -> Any:
```
squirrelsc: Return the serial console feature?

lisa/sut_orchestrator/baremetal/context.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+@dataclass
+class EnvironmentContext:
+    ssh_public_key: str = ""
+
+
+@dataclass
+class NodeContext:
+    management_port: int = -1
```
squirrelsc: It should be a field for rackmanager only. The Context should be inherited for different cluster.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -0,0 +1,129 @@
...
+            base_url="https://" + self.idrac_runbook.address,
+            username=self.idrac_runbook.username,
+            password=self.idrac_runbook.password,
+        )
+        self._login()
+        self._change_boot_order_once("VCD-DVD")
+        for client in self.idrac_runbook.client:
+            assert client.iso_http_url, "iso_http_url is required for idrac client"
+            self._insert_virtual_media(client.iso_http_url)
+            self._reboot()
```
pupacha: Few comments:
1. I think it's better to change the boot order after virtual media insertion; just in case virtual media insertion fails, we don't want the server to have its boot order changed.
2. calling _eject_virtual_media() right before _insert_virtual_media() will also be good, as it will remove any stale inserted media.
3. After _reboot(), it's good to do sleep(<few seconds>) as _reboot does graceful shutdown and it might take sometime.
4.  I dont know if _exit() is implicitly called. Otherwise we should call _logout()

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def deploy(self, environment: Environment) -> Any:
+        self.redfish_instance = redfish.redfish_client(
+            base_url="https://" + self.idrac_runbook.address,
+            username=self.idrac_runbook.username,
+            password=self.idrac_runbook.password,
+        )
+        self._login()
+        self._change_boot_order_once("VCD-DVD")
+        for client in self.idrac_runbook.client:
```
pupacha: Usually, iDRAC is dedicated to a server which means there would be only one client per idrac_runbook. So may be we can enfore only one entry for `clients` section in the schema and remove the loop here.

lisa/sut_orchestrator/baremetal/platform_.py
LiliDeng: updated, not test yet

lisa/sut_orchestrator/baremetal/cluster/cluster.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/build.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/platform_.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/platform_.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/platform_.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/schema.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/readychecker.py
LiliDeng: removed

lisa/sut_orchestrator/baremetal/cluster/cluster.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/cluster/cluster.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/cluster/idrac.py
squirrelsc: The client is defined in base schema, and it helps in general logic. In this loop, we can add an assertion to make sure only one client defined.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
squirrelsc: For 1, Do you mean change it back to boot from disk? As a test target, I think it doesn't help. The old distro may be confusing, people may ignore the deployment failure, thought it's deployed correctly. When deploying a good image next time, the system will come back.
For 3, it can call turn off directly. Because it's fast and reliable, and we don't need to care about existing system.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,159 @@
...
+
+        # ready checker cleanup
+        if cluster_instance.ready_checker:
+            ready_checker = self.ready_checker_factory.create_by_runbook(
+                cluster_instance.ready_checker
+            )
+            ready_checker.clean_up()
+
+        # copy build if source exists
+        if cluster.runbook.build and cluster.runbook.build.is_copied:
```
squirrelsc: It should print debug log, if build is not copied.

lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@dataclass_json()
+@dataclass
+class BuildSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
+    type: str = field(default="smb", metadata=field_metadata(required=True))
+    name: str = ""
+    share: str = ""
+    files: List[FileSchema] = field(default_factory=list)
+    is_copied: bool = False
```
squirrelsc: This value shouldn't be put in schema, because the value in schema may be changed in runbook. It cause unexpected behavior.
You can use the context pattern to maintain a context on each build object.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -0,0 +1,159 @@
...
+
+            assert cluster.runbook.client[
+                index
+            ].connection, "client connection is empty"
+            assert cluster.runbook.client[
+                index
+            ].connection.address, "ip address is empty"
+            assert isinstance(
+                node, RemoteNode
+            ), "The client node must be remote node with connection information"
```
squirrelsc: Refer to other pattern, it's better to print the actual type, so the error message is helpful.

lisa/sut_orchestrator/baremetal/context.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@dataclass
+class EnvironmentContext:
+    ssh_public_key: str = ""
+
+
+@dataclass
+class NodeContext:
+    address: str = ""
```
squirrelsc: Use `ConnectionInfo` in schema.py

lisa/sut_orchestrator/baremetal/cluster/idrac.py
pupacha: For 1, I meant to do `self._change_boot_order_once("VCD-DVD")` after `_insert_virtual_media`. This is because we should be changing boot order only if the virtual media insertion is successful.

For 3 - Yes, we dont need to care about existing system, but server might not turn off immediately, I think it does graceful shutdown, and I've noticed cases where it takes some time for it to turn off. 
Since there is `ready_checker (ssh)` immediately after `deploy`, spur/paramiko might connect to the older existing system and LISA would think deploying is done. I have noticed this behavior when I was testing this PR for our usecases. Hence the suggestion to `sleep()` so that this is avoided. Another we could do here instead of sleep() is, check for the `reboot` job status through iDRAC api. 
I'm fine with using any of the above 2 methods.

squirrelsc: @LiliDeng Can we change the process to shutdown or turn off firstly, when it's done, then start it up? It should be more reliable.

lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+@dataclass_json()
+@dataclass
+class BuildSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
+    type: str = field(default="smb", metadata=field_metadata(required=True))
+    name: str = ""
+    share: str = ""
+    files: List[FileSchema] = field(default_factory=list)
+    unzip: bool = True
```
squirrelsc: It should be in FileSchema, specify which file to be unzip. If there is any other data needed, inherit another schema from FileSchema.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -0,0 +1,43 @@
...
+    def deploy(self, environment: Environment) -> Any:
+        raise NotImplementedError()
+
+    def has_serial_console(self) -> bool:
+        raise NotImplementedError()
+
+    def get_serial_console(self) -> features.SerialConsole:
+        raise NotImplementedError()
+
+    def start(self) -> None:
```
squirrelsc: How about define StartStop feature for each cluster like SerialConsole?

lisa/sut_orchestrator/baremetal/cluster/idrac.py
pupacha: @LiliDeng Looks like point 1 is still not taken care.

For 3, I see you are doing  `self.stop()` and ` self.start()` . I think stop  is asynchronous operation. So should we do `_wait_for_completion` inside `stop` ?

lisa/sut_orchestrator/baremetal/cluster/cluster.py
LiliDeng: you mean define like below way? I didn't implement get_serial_console yet. not sure how to implement in the method. should it return features.StartStop or the baremetal features.StartStop ?
```
    def get_start_stop(self) -> features.StartStop:
        raise NotImplementedError()
```

lisa/sut_orchestrator/baremetal/cluster/idrac.py
LiliDeng: @pupacha I have changed for point 1.
in _reset, it has _wait_for_completion, stop/start/reboot invoke _reset

pupacha: ok thanks.

lisa/sut_orchestrator/baremetal/schema.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/cluster/idrac.py
LiliDeng: updated

lisa/sut_orchestrator/baremetal/cluster/cluster.py
LiliDeng: Sorry, I didn't figure out the actual implement for this. 
Can you check the features.py, I think current stratstop in features.py works, if we define `def get_start_stop(self) -> features.StartStop:`, do we still need features.py? when `get_start_stop` will be used? I didn't figure it out.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -0,0 +1,97 @@
...
+                    file_path = rf"{source_path}\{file_map.source}".replace(
+                        "\\", "\\\\"
+                    )
+                    pattern = re.compile(
+                        file_path,
+                        re.I | re.M,
+                    )
+                    for file in all_files:
+                        if pattern.match(file):
+                            with open(file, "rb") as f:
```
squirrelsc: use a function to reduce logic levels.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        self._login()
+        self._eject_virtual_media()
+        self._change_boot_order_once("VCD-DVD")
+        assert client.iso_http_url, "iso_http_url is required for idrac client"
+        self.stop()
+        self._insert_virtual_media(client.iso_http_url)
+        self.start()
+
+    def restart(self) -> None:
```
squirrelsc: Move the holw operation related parts to StartStop.

lisa/sut_orchestrator/baremetal/features.py
```diff
@@ -0,0 +1,45 @@
...
+if TYPE_CHECKING:
+    from .platform_ import BareMetalPlatform
+
+
+class StartStop(Feature):
+    def __getattr__(self, key: str) -> Any:
+        assert self._inner, "inner is not set"  # type: ignore
+        return getattr(self._inner, key)  # type: ignore
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
```
squirrelsc, line 19: Create a base class for all features, so the logic won't be repeated on each feature.

squirrelsc: You can add an inherit3ed method to get _feature_type.

```python
class ClusterFeature(Feature):
    def __getattr__(self, key: str) -> Any:
        assert self._inner, "inner is not set"  # type: ignore
        return getattr(self._inner, key)  # type: ignore

        _feature_type = self._get_inner_type()
        self._inner = _feature_type(
            schema.FeatureSettings.create(_feature_type.name()),
            self._node,
            self._platform,
            *args,
            **kwargs,
        )

class StartStop(ClusterFeature):
    def _get_inner_type(self) -> Feature:
        platform: BareMetalPlatform = self._platform  # type: ignore
       return self.platform.get_start_stop()
```

squirrelsc (approved):

PR #2914, opened 2023-08-21
TRANSFORMER_PHASE_ENVIRONMENT_CONNECTED

kamalca:
Adds a transformer phase that runs every time a VM is deployed.


kamalca:
@squirrelsc Previous transformers are passed a node connection in the runbook. However, for transformers in this phase I want to access the node that was just deployed. Do you have thoughts on the best way to implement that?

squirrelsc:
> @squirrelsc Previous transformers are passed a node connection in the runbook. However, for transformers in this phase I want to access the node that was just deployed. Do you have thoughts on the best way to implement that?

Deployed phase is different than other phases. You can create a new base class for deployed phase transformers, which can accept an environment argument on `__init__`. This argument can be optional, so if a transformer doesn't support this argument, it still can be used at deployed phase. Versus, a deployed transformer shouldn't raise an exception, if it can be used by other phases.

kamalca:
@squirrelsc It looks like my changes changed the behavior of LisaRunner so that CI workflows are failing. Can you help me understand what is going wrong?

kamalca:
So the changed behavior is due to
``` python
transformer.run(
    self._runbook_builder, phase=constants.TRANSFORMER_PHASE_DEPLOYED
)
```
But I would assume that since there are no transformers in that phase, it should not affect behavior. I will have to investigate further how this is making the CI tests fail.

lisa/util/subclasses.py
```diff
@@ -47,7 +47,10 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         # initialize types from subclasses.
         # each type should be unique in code, or there is warning message.
         for subclass_type in self._get_subclasses(self._base_type):
-            subclass_type_name = subclass_type.type_name()
+            try:
+                subclass_type_name = subclass_type.type_name()
```
squirrelsc: When will it happen? If it needs to instant for middle classes, maybe we have to add them. If capture the not implemented error here, it's not useful also.

kamalca: Should I implement type_name() in DeploymentTransformer? We should never have an instance of DeploymentTransformer so I didn't know how to handle this situation.

kamalca: This case occurs when creating a Transformer factory and it tries to get the name of DeploymentTransformer since it is a subclass.

squirrelsc: Yes, it's confusing. But it looks like the type_name for an abstract class is needed to bypass this error.

kamalca: Okay, I went forward with just adding a class name to the DeploymentTransformer

examples/runner/mock_runner.py
```diff
@@ -41,10 +42,14 @@ class MockRunner(BaseRunner):
     _is_done = False
 
     def __init__(
-        self, runbook: schema.Runbook, index: int, case_variables: Dict[str, Any]
+        self,
+        runbook_builder: RunbookBuilder,
+        runbook: schema.Runbook,
+        index: int,
```
squirrelsc, line 48: Remove the runbook, it can be generated from the runbook builder.

lisa/runner.py
```diff
@@ -94,11 +94,13 @@ class BaseRunner(BaseClassMixin, InitializableMixin):
 
     def __init__(
         self,
+        runbook_builder: RunbookBuilder,
         runbook: schema.Runbook,
```
squirrelsc, line 98: Remove the runbook, it can be generated from the runbook builder.

lisa/schema.py
```diff
@@ -1223,6 +1224,15 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
         add_secret(self.private_key_file)
 
 
+@dataclass_json
+@dataclass
+class DeploymentTransformer(Transformer):
```
squirrelsc: It can be put into transformers folder like other transformers.

lisa/transformer.py
```diff
@@ -71,6 +73,30 @@ def _internal_run(self) -> Dict[str, Any]:
         raise NotImplementedError()
 
 
+class DeploymentTransformer(Transformer):
```
squirrelsc: Try to define it like a Mixins pattern, and move it into transformers folder.

lisa/transformer.py
```diff
@@ -174,7 +205,10 @@ def _run_transformers(
     for value in runbook_builder.variables.values():
         copied_variables[value.name] = value.copy()
 
-    factory = subclasses.Factory[Transformer](Transformer)
+    if node:
+        factory = subclasses.Factory[Transformer](DeploymentTransformer)
```
squirrelsc: I feel it doesn't need a different factory. If the `__init__` accepts `**kwargs`. The unused `node` parameter doesn't hurt other transformers.

lisa/transformer.py
```diff
@@ -198,16 +232,25 @@ def _run_transformers(
...
-    if constants.TRANSFORMER not in root_runbook_data:
+    if not root_runbook_data or constants.TRANSFORMER not in root_runbook_data:
         log.debug("no transformer found, skipped")
         return
 
     # real run
     log.debug(f"detecting or running transformers of phase '{phase}'...")
-    output_variables = _run_transformers(runbook_builder, phase=phase)
-    merge_variables(runbook_builder.variables, output_variables)
+    if environment and environment.nodes:
```
squirrelsc, line 228: Add code comments to explain why the environment is given sometimes.

lisa/transformer.py
kamalca: Without a different factory, it won't run the `__init__` logic in DeploymentTransformer, right?

lisa/runner.py
kamalca: If I only pass the runbook builder, the runner filters will have to be reapplied. Either I have to copy a lot of code or find a better location for the code.

kamalca: See `RootRunner._generate_runners` for the code that would need to be run again on the runbook.

lisa/transformer.py
kamalca: Done.

lisa/transformer.py
squirrelsc: Did you have a chance to test it? The `create_by_runbook` uses kwargs, and it passes through all named arguments to `__init__`, so it should be the same. If a transformer doesn't use node, but it's passed in, there is no error. When a transformer requires the node, but it's not passed in, we can raise an excpetion.

lisa/runner.py
squirrelsc: I see. It needs to rename the `runbook` to `transformer_runbook`.

kamalca: Will that be confusing? We also use `runbook` in the transformer to mean `schema.Transformer`. By calling it transformer_runbook, it sounds to me like it is the `schema.Transformer` that initilizes the Transformer, but that's not the case here.

lisa/transformer.py
squirrelsc: It's the only one comment.

kamalca: I moved into transformers folder. Can you clarify on this comment?

squirrelsc: Try to write it as a Mixin pattern like InitializableMixin. So, when there are more inheriting on transformers, it could be more flexible.

kamalca: Since I am overriding `__init__` to take in an extra parameter, I'm not sure how well it fits as a Mixin class. I don't want the implementation to require a specific order for multi-class inheritance.
I agree with you that this could be made more flexible, but I'm not sure the best way to implement it.

squirrelsc: ok, we can go with this.

squirrelsc (approved):

pupacha:
This was much needed. Thanks @kamalca for adding this feature.

PR #2916, opened 2023-08-22
MANA: handle ubuntu backport updates

mcgov:
Canonical has backported 6.2+ mana changes to Ubuntu 5.15, releasing in 1045.
Handle installing the correct kernel versions on Ubuntu < 23.04 (ships with 6.2).

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -267,6 +267,40 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
         )
 
 
+UBUNTU_DPDK_MINIMUM_KERNEL = "5.15.0-1045"
+
+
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -267,13 +267,52 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
...
+
+
+def _set_force_linux_next_for_mana(node: Node, variables: Dict[str, Any]) -> None:
+    distro = node.os
+    # Handle mana minimum ubuntu kernel,
+    #  accounting for backports to 5.15 (5.4 not available)
+    is_mana = node.nics.is_mana_present()
+    if is_mana and isinstance(distro, Ubuntu) and distro.information.version < "23.4.0":
+        if distro.information.version < "20.4.0":
+            raise SkippedException(
```
squirrelsc, line 280: Should it be skipped? My understanding is the mana device presents is not related to mana driver. Please sync up with Lili offline.

PR #2917, opened 2023-08-22
Azure Disk Encryption Tests V1

bamuir:
Draft PR for internal team discussions and ensuring this POC meets our specs

Implemented 2 of the core ADE Extension tests;

1. verify_azure_disk_encryption_provisioned (P1)

- A P0 test, designed to run in under 5 minutes, which will only check if the ADE extension provisioned successfully on the VM

2. verify_azure_disk_encryption_enabled (P3)

- A P1 test, designed to run in under 2 hours, which will ensure the ADE extension successfully encrypted the VM OS disk

- Modified creation of keyvault to allow passing VaultProperties
- Check if a kv exists using a regex, for avoiding duplicate creations

Example test usage with debug:

lisa -r ./microsoft/runbook/debug.yml -v origin:./azure.yml -v "admin_private_key_file:$key_path" -v case:verify_azure_disk_encryption_enabled -v subscription_id:7$subID -v marketplace_image:MicrosoftCBLMariner:cbl-mariner:cbl-mariner-2-gen2:latest -v vm_size:Standard_D2S_V3 -v location:centralus -d


microsoft/runbook/tiers/t0.yml
```diff
@@ -1,3 +1,3 @@
 testcase:
   - criteria:
-      priority: 0
+      name: $(test_name)
```
squirrelsc: Please revert changes to the default tier files.

microsoft/runbook/azure.yml
```diff
@@ -2,6 +2,8 @@ name: azure default
 include:
   - path: ./tiers/tier.yml
 variable:
+  - name : test_name
```
squirrelsc: Remove test name. The platform runbook is for platform setup only.

pyproject.toml
```diff
@@ -185,4 +188,4 @@ packages = ["lisa"]
 
 [tool.setuptools_scm]
 version_scheme = "post-release"
-local_scheme = "no-local-version"
+local_scheme = "no-local-version"
```
squirrelsc: Add back the last empty line.

microsoft/testsuites/vm_extensions/azurediskencryption.py
```diff
@@ -0,0 +1,186 @@
...
+
+        assert_that(os_status).described_as(
+            "Expected the OS status to be 'Encrypted'"
+        ).is_equal_to("Encrypted")
+
+    @TestCaseMetadata(
+        description="""
+        Runs the ADE extension and verifies the extension provisioned successfully on the remote machine.
+        """,
+        priority=0,
```
squirrelsc: Move it to P1. We have only one P0 test case, which can run on all images.

microsoft/testsuites/vm_extensions/azurediskencryption.py
```diff
@@ -0,0 +1,186 @@
...
+        supported_platform_type=[AZURE],
+        supported_os=[Ubuntu, CBLMariner, CentOs, Oracle, Redhat, SLES, Suse],
+    ),
+)
+class AzureDiskEncryption(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        Runs the ADE extension and verifies it enabled and fully encrypted the remote machine successfully.
+        """,
+        priority=1,
```
squirrelsc: Move it to P3, because it takes 2 hours.

microsoft/runbook/tiers/t0.yml
bamuir: Yes will revert all the dev setup I have. Still just WIP to document some of these things and share with other devs from our team who may also work on LISA tests. Would be good if we could document some of these tricks/available parameters for runbooks. (or maybe I missed it?)

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1895,3 +1916,292 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    # Set up the API call headers
+    headers = {
+        "Authorization": f"Bearer {token.token}",
+        "Content-Type": "application/json",
+    }
+
+    # Set a timeout of 10 seconds for the request
+    response = requests.get(request_url, headers=headers, timeout=10)
+
+    if response.status_code != 200:
```
squirrelsc, line 1963: Does it support both service principal and user account?

microsoft/testsuites/vm_extensions/azurediskencryption.py
```diff
@@ -0,0 +1,186 @@
...
+def _enable_ade_extension(node: Node, log: Logger, result: TestResult) -> any:
+    environment = result.environment
+    assert environment, "fail to get environment from testresult"
+    platform = environment.platform
+    assert isinstance(platform, AzurePlatform)
+    runbook = platform.runbook.get_extended_runbook(AzurePlatformSchema)
+    tenant_id = get_tenant_id(platform.credential)
+    if tenant_id is None:
+        raise ValueError("Environment variable 'tenant_id' is not set.")
+    object_id = get_sp_object_id(runbook.service_principal_client_id)
```
squirrelsc: The runbook.service_principal_client_id should be the same as object_id. And the runbook.service_principal_client_id is optional, what happened if it's not set?

microsoft/testsuites/vm_extensions/azurediskencryption.py
```diff
@@ -0,0 +1,186 @@
...
+    log.debug(f"Object ID: {object_id}")
+    log.debug(f"Runbook: {runbook}")
```
squirrelsc: You don't need to print out the runbook, it's printed in log already.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1895,3 +1916,292 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+    log: Logger,
+) -> bool:
+    certificate_client = get_certificate_client(vault_url, platform)
+
+    try:
+        certificate_client.begin_delete_certificate(cert_name)
+        log.debug(f"Certificate {cert_name} deleted successfully.")
+        return True
+    except Exception:
+        raise LisaException
```
squirrelsc: except and raise is not necessary. And it deletes details of an exception.

LiliDeng:
@bamuir thanks for your contribution! Please rebase the code from the main branch, #2892 has been merged, you can just reflect your own changes in this PR.

microsoft/runbook/azure.yml
```diff
@@ -47,4 +47,4 @@ platform:
         marketplace: $(marketplace_image)
         vhd: $(vhd)
         location: $(location)
-        vm_size: $(vm_size)
+        vm_size: $(vm_size)
```
bamuir: @squirrelsc do you know why github is complaining about this line here?

LiliDeng: @bamuir there is a new line in the end, don't remove it

microsoft/testsuites/vm_extensions/azurediskencryption.py
```diff
@@ -0,0 +1,200 @@
...
+    ),
+)
+class AzureDiskEncryption(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        Runs the ADE extension and verifies it
+        fully encrypted the remote machine successfully.
+        """,
+        priority=3,
+        requirement=simple_requirement(supported_features=[AzureExtension]),
```
LiliDeng: I encounter this error
 ```raise Exception("Not enough memory for enabling encryption on OS volume. 8 GB memory is recommended.")```, 

please add 8 Gb memory as case requirement
``` 
       requirement=node_requirement(
            supported_features=[AzureExtension],
            node=schema.NodeSpace(
                core_count=1, memory_mb=search_space.IntRange(min=8 * 1024)
            ),
        ),
```

microsoft/testsuites/vm_extensions/azurediskencryption.py
```diff
@@ -0,0 +1,200 @@
+# Copyright (c) Microsoft Corporation.
```
LiliDeng: file name as azure_disk_encryption.py is more straight

microsoft/testsuites/vm_extensions/azurediskencryption.py
```diff
@@ -0,0 +1,200 @@
...
+        supported_os=[Ubuntu, CBLMariner, CentOs, Oracle, Redhat, SLES, Suse],
+    ),
+)
+class AzureDiskEncryption(TestSuite):
+    @TestCaseMetadata(
+        description="""
+        Runs the ADE extension and verifies it
+        fully encrypted the remote machine successfully.
+        """,
+        priority=3,
```
LiliDeng: add timeout=3600 * 2 here, default timeout value is 3600 seconds.

microsoft/testsuites/vm_extensions/azurediskencryption.py
bamuir: I added the node_requirement in TestSuiteMetadata rather than TestCase.

squirrelsc: The test case level requirement overwrites the test suite level requirement.

bamuir: @squirrelsc I was having an issue adding supported_features to node_requirement. It seemed to complain in mypy, and only work well with simple_requirement. Whats the best way to do this?

squirrelsc: Can you share the error met?

bamuir: @TestCaseMetadata(
        description="""
        Runs the ADE extension and verifies it
        fully encrypted the remote machine successfully.
        """,
        priority=3,
        timeout=3600*2,
        requirement=node_requirement(
            supported_features=[AzureExtension],
            node=schema.NodeSpace(
                core_count=1, memory_mb=search_space.IntRange(min=8 * 1024)
            ),
        ),
    )

mypy ./microsoft/testsuites/vm_extensions/azure_disk_encryption.py 
microsoft/testsuites/vm_extensions/azure_disk_encryption.py:134:21: error: Unexpected keyword argument "supported_features" for "node_requirement"
Found 1 error in 1 file (checked 1 source file)

bamuir: Also, if test case level overwrites, seems I should only use testcase level to define the metadata, since each test has a different priority.

squirrelsc: priority is not part of the requirements, so it doesn't affect the requirements by overwriting. The type should be node_requirement, instead of simple_requirement. Because the simple_requirement doesn't have the `supported_features` argument.

bamuir: I pasted the code I am using when seeing the error. The pushed code is not seeing the error. 

I see the error using node_requirement. simple_requirement does have the supported_features argument. I could remove the argument in general for node_requirement?

squirrelsc: Sorry, I mixed them incorrectly. It needs to use `simple_requirement` or use the `node_requirement` within a node space. BTW, I just found it set the core_count to 1, it means use a VM size with only 1 core. It would be very slow, make sure it's the test purpose.

```python
node_requirement(
  node=schema.NodeSpace(
    supported_features=[AzureExtension],
    memory_mb=search_space.IntRange(min=8 * 1024)
  )
)
```

bamuir: We would want more cores required if possible, actually. Encryption can take some time, so having a good vm size is key for reliable testing. 

Could we require at least 2, perhaps 4?

squirrelsc: Yes, you can define core_count like memory_mb to set a min value.

bamuir: `
        priority=3,
        timeout=3600*2,
        requirement=node_requirement(
            node=schema.NodeSpace(
                supported_features=[AzureExtension],
                memory_mb=search_space.IntRange(min=8 * 1024),
                core_count = 2
            ),
            supported_platform_type=[AZURE],
            supported_os=[Ubuntu, CBLMariner, CentOs, Oracle, Redhat, SLES, Suse],
        )
    )
`


and 

`
        priority=1,
        requirement=node_requirement(
            node=schema.NodeSpace(
                memory_mb=search_space.IntRange(min=8 * 1024),
            ),
            supported_features=[AzureExtension],
            supported_platform_type=[AZURE],
            supported_os=[Ubuntu, CBLMariner, CentOs, Oracle, Redhat, SLES, Suse],
        )
    )
`

are both still throwing :

microsoft/testsuites/vm_extensions/azure_disk_encryption.py:130:18: error: Unexpected keyword argument "supported_features" for "NodeSpace"
microsoft/testsuites/vm_extensions/azure_disk_encryption.py:197:21: error: Unexpected keyword argument "supported_features" for "node_requirement"

squirrelsc: Sorry, please try `features`, instead of `supported_features`. I found it's not used in test cases, so it may need to refer this example. @LiliDeng can you support min memory in simple requirements? So it doesn't need to use node_requirement in this case.

bamuir: Also having same troubles with features inside node_requirement.

I think I can add an optional parameter to node_requirement, which would then pass to _create_test_case_requirement, would you like me to push that change for you to see?

bamuir: Pushed the change that could work. I feel limited here though. For example, would be nice to also add a test using Data disks, which seems possible via simple_requirements. However, I would still want the node memory and core constraints.

squirrelsc: Please add the `min_memory_mb` supports to simple_requirement, instead of node. The features are part of node, not at the node_requirement level.

squirrelsc: > Pushed the change that could work. I feel limited here though. For example, would be nice to also add a test using Data disks, which seems possible via simple_requirements. However, I would still want the node memory and core constraints.

The `node_requirement` is more capable than simple_requirement, but it's complex. If you add the min_memory_mb to `simple_requirement`, the simple_requirement fits the requirements. But the node_requirement cannot add supported_features, because it should be in the node, not separatedly.

microsoft/testsuites/vm_extensions/azure_disk_encryption.py
```diff
@@ -0,0 +1,193 @@
...
+
+@TestSuiteMetadata(
+    area="vm_extension",
+    category="functional",
+    description="Tests for the Azure Disk Encryption (ADE) extension",
+    requirement=simple_requirement(
+        min_memory_mb=8 * 1024,
+        supported_features=[AzureExtension],
+        supported_platform_type=[AZURE],
+        supported_os=[Ubuntu, CBLMariner, CentOs, Oracle, Redhat],
```
bamuir: @squirrelsc / @LiliDeng my test seems to be failing now after the change adding min_memory to simple_requirements.

I am seeing my test skipped for unsupported OS when using a supported OS. I tried Mariner, RHEL, and Ubuntu. Here is the output and last command:

squirrelsc: Can you check os matrix like other extensions? They use a distro list, and use isinstance to check.

bamuir: If I am looking at the right setup (azureperformancediagnostics.py), this means we would have to come in here and manually update this support any time we make a support addition in ADE, or any time a new image comes out? Would be cool if we could set a minimum supported OS Major version, so that for example when RHEL 10 comes out, we do not all have to modify the code to allow RHEL 10 testing. 

It seems at least for minor versions this will work well for our case, and we have some older major versions we don't support so we can specify those. But just a thought for future improvement if the supported_os feature is not working.

bamuir: Update pushed. Did some additional refactoring to clean up my helper methods

squirrelsc: If you can support latest versions without change the code, that would be great. You can modify the logic to have minimal versions like `os.version >= "8.0.0"`

microsoft/testsuites/vm_extensions/azure_disk_encryption.py
```diff
@@ -0,0 +1,221 @@
...
+        min_memory_mb=8 * 1024,
+        supported_features=[AzureExtension],
+        supported_platform_type=[AZURE],
+    ),
+)
+class AzureDiskEncryption(TestSuite):
+    def before_case(self, log: Logger, **kwargs: Any) -> None:
+        node = kwargs["node"]
+        if not self._is_supported_linux_distro(node):
+            raise SkippedException(
```
squirrelsc: Raise it like below, so the os information can be bring out automatically.

`raise SkippedException(UnsupportedDistroException(node.os))`

bamuir: Updated, included support for min supported versions as well.

microsoft/testsuites/vm_extensions/azure_disk_encryption.py
```diff
@@ -0,0 +1,218 @@
...
+            Redhat: [7, 8, 9],
+            CentOs: [7, 8],
+            Oracle: [8],
+            Ubuntu: [18, 20, 22],
+            CBLMariner: [2],
+        }
+
+        for distro, versions in supported_major_versions.items():
+            if isinstance(node.os, distro):
+                min_supported_version = min(versions)
```
squirrelsc: If so, one version is enough to support the logic.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #2918, opened 2023-08-22
MANA: handle ubuntu backport updates and enforce kernel versions

mcgov:
Canonical has backported 6.2+ mana changes to Ubuntu 5.15, releasing in 1045.
Handle installing the correct kernel versions on Ubuntu < 23.04 (ships with 6.2).

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -267,6 +267,40 @@ def enable_uio_hv_generic_for_nic(node: Node, nic: NicInfo) -> None:
         )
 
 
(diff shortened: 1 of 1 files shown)


mcgov:
https://github.com/microsoft/lisa/pull/2916#discussion_r1302096370

@squirrelsc responding in draft PR version.
VMs can run on MANA without the driver, but DPDK is not guaranteed to function correctly without the requirements.
This is enforcing the tests to not run on images which don't satisfy the requirements. These checks will likely become more relaxed as more images add those backported patches.

mcgov:
This PR is not needed anymore, there's a different version to handle rdma-core installation for kernels < v6.4 on the way.

PR #2919, opened 2023-08-23
DPDK VPP: bugfixes for new versions and SUSE

mcgov:
SLES doesn't have vpp packages available for some versions, so check if there are none after adding the repository and refreshing. - Skip is there is not a supported package available.
- Add interface check to look for 'VSC' as well as 'VirtualFunction'

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkvpp.py
+++ b/microsoft/testsuites/dpdk/dpdkvpp.py
@@ -84,6 +84,7 @@ class DpdkVpp(Tool):
         vpp_detected_interface = (
             "GigabitEthernet" in vpp_interface_output
             or "VirtualFunctionEthernet" in vpp_interface_output
+            or "NetVSC" in vpp_interface_output
         )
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkvpp.py
```diff
@@ -84,6 +84,7 @@ def run_test(self) -> None:
         vpp_detected_interface = (
             "GigabitEthernet" in vpp_interface_output
             or "VirtualFunctionEthernet" in vpp_interface_output
+            or "NetVSC" in vpp_interface_output
```
LiliDeng, line 87: NetVSC is expected or not, can you double confirm? from the text, I understand it is synthetic network?

mcgov: There is a dpdk PMD called netvsc as well

LiliDeng (approved):

PR #2926, opened 2023-08-30
try to fix folder exist issue

LiliDeng:


lisa/testsuite.py
```diff
@@ -672,6 +673,7 @@ def __save_serial_log(self, environment: Environment, log_path: Path) -> None:
                 log_dir.mkdir(parents=True)
                 serial_console.get_console_log(log_dir, force_run=True)
 
+    @retry(exceptions=FileExistsError, tries=30, delay=1)
```
squirrelsc: Delay shorter to get it running faster, 0.1 second is long enough. Even 0 is ok.

LiliDeng: updated

squirrelsc (approved):

PR #2929, opened 2023-08-30
Kselftest: Run using Linux Stable repo

adityagesh:
Run kselftest from linux stable - master branch

Currently tested for Ubuntu and Mariner

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest.py
+++ b/microsoft/testsuites/kselftest/kselftest.py
@@ -11,11 +11,53 @@ from lisa.base_tools.uname import Uname
 from lisa.executable import Tool
 from lisa.messages import SubTestMessage, TestStatus, create_test_result_message
 from lisa.node import Node
-from lisa.operating_system import CBLMariner
+from lisa.operating_system import CBLMariner, Ubuntu
 from lisa.testsuite import TestResult
 from lisa.tools import Cp, Git, Ls, Make, RemoteCopy, Tar
+from lisa.tools.chmod import Chmod
+from lisa.tools.mkdir import Mkdir
 from lisa.tools.whoami import Whoami
 from lisa.util import LisaException, UnsupportedDistroException, find_groups_in_lines
+
+_UBUNTU_OS_PACKAGES = [
+    "git",
+    "build-essential",
+    "bison",
+    "flex",
+    "libelf-dev",
+    "xz-utils",
+    "libssl-dev",
+    "bc",
+    "ccache",
+    "libncurses-dev",
+    "gcc-multilib",
+    "libc6-i386",
+    "libc6-dev-i386",
+]
+
+_MARINER_OS_PACKAGES = [
+    "bison",
+    "flex",
+    "build-essential",
+    "openssl-devel",
+    "bc",
+    "dwarves",
+    "rsync",
+    "libcap-devel",
+    "libcap-ng-devel",
+    "fuse",
+    "fuse-devel",
+    "popt-devel",
+    "numactl-devel",
+    "libmnl-devel",
+    "libinput",
+    "mesa-libgbm-devel",
+    "glibc-static",
+    "clang",
+    "glibc-devel",
+    "binutils",
+    "kernel-headers",
+]
 
 
 @dataclass
@@ -26,10 +68,8 @@ class KselftestResult:
 
 
 class Kselftest(Tool):
-    _MARINER_KERNEL_SRC_REPO = "https://github.com/microsoft/CBL-Mariner-Linux-Kernel"
-    _KERNEL_REPO_NAME = "CBL-Mariner-Linux-Kernel"
-    _KSELFTEST_TAR_PATH = (
-        "/build/kselftest/kselftest_install/kselftest-packages/kselftest.tar.gz"
+    _KSELF_TEST_SRC_REPO = (
+        "git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git"
     )
 
     # kselftest result log has "ok" and "not ok" prefixes, use regex to filter them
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -64,66 +65,89 @@ def __init__(
...
+            self.node.tools[Tar].extract(
+                str(self._remote_tar_path), str(self._kself_installed_dir), sudo=True
             )
+            self._log.debug(f"Extracted tar from path {self._remote_tar_path}!")
+        else:
+            mkdir = self.node.tools[Mkdir]
+            mkdir.create_directory(str(self._kself_installed_dir))
+            if isinstance(self.node.os, Ubuntu):
+                # cache is used to speed up recompilation
+                self.node.os.install_packages(
```
praveenkumr: Is there a way, where we can maintain this list somewhere as json or similar, based upon the Distro and programatically install, instead of keeping track everywhere we create / import new testcases/suits to lisa.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -64,66 +65,89 @@ def __init__(
...
 
     # install common dependencies
     def _install(self) -> bool:
-        if not isinstance(self.node.os, CBLMariner):
+        if not (
+            isinstance(self.node.os, Ubuntu) or isinstance(self.node.os, CBLMariner)
+        ):
             raise UnsupportedDistroException(
-                self.node.os, "kselftests are supported on Mariner VMs only."
+                self.node.os, "kselftests are supported on Ubuntu and Mariner only"
```
squirrelsc: The error message of `UnsupportedDistroException` has OS information already, it doesn't need to mention here. The message should be like `kselftests in LISA doesn't support this os`.

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: It's ok to add it here. You can move it to top of the file like configuration. Tests have different dependencies, and it won't be changed frequently. So it's better to track it in test code.

adityagesh: Updated

microsoft/testsuites/kselftest/kselftest.py
adityagesh: Thanks, Updated

adityagesh:
@squirrelsc @LiliDeng Any more comments on this PR?

squirrelsc:
@LiliDeng LGTM

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -64,66 +106,48 @@ def __init__(
...
+                self.node.os.install_packages(_MARINER_OS_PACKAGES)
 
             uname = self.node.tools[Uname]
             uname_result = uname.get_linux_information(force_run=False)
-            version = uname_result.kernel_version
             git = self.node.tools[Git]
 
             # If version.patch is zero, clone major.minor kernel from
             # upstream stable kernel. If version.patch is non-zero, clone
             # major.minor.patch kernel corresponding to the distro in use
```
LiliDeng: change comments as well

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -138,30 +162,44 @@ def _install(self) -> bool:
...
     def run_all(
-        self, test_result: TestResult, environment: Environment, log_path: str
+        self,
+        test_result: TestResult,
+        environment: Environment,
+        log_path: str,
+        run_test_as_root: bool = False,
     ) -> List[KselftestResult]:
+        # Executing kselftest as root may cause
+        # VM to hang
```
LiliDeng, line 199: why?

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -64,66 +106,48 @@ def __init__(
...
+            self._log.debug(f"Extracted tar from path {self._remote_tar_path}!")
+        else:
+            mkdir = self.node.tools[Mkdir]
+            mkdir.create_directory(str(self._kself_installed_dir))
+            if isinstance(self.node.os, Ubuntu):
+                # cache is used to speed up recompilation
+                self.node.os.install_packages(_UBUNTU_OS_PACKAGES)
+            elif isinstance(self.node.os, CBLMariner):
+                # clone kernel, build kernel, then build kselftests
+                self.node.os.install_packages(_MARINER_OS_PACKAGES)
```
LiliDeng, line 144: does it support all ubuntu distro?

microsoft/testsuites/kselftest/kselftest.py
adityagesh: I had a specific case where I observed this, I did not save the reference. Let me try to find it.

My understanding is that without sudo that test does not run.
image

microsoft/testsuites/kselftest/kselftest.py
adityagesh: Updated to restrict to 18.04 and above.

microsoft/testsuites/kselftest/kselftest.py
adityagesh: Removed

microsoft/testsuites/kselftest/kselftest.py
adityagesh: https://www.spinics.net/lists/linux-kselftest/msg15551.html

LiliDeng (dismissed):

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -138,32 +162,47 @@ def _install(self) -> bool:
...
+                shell=True,
+                cwd=PurePosixPath(kernel_path, "tools/testing/selftests"),
+                sudo=True,
+                expected_exit_code=0,
+                expected_exit_code_failure_message="fail to build & install kselftest",
+            ).assert_exit_code()
+            # change permissions of kselftest-packages directory
+            # to run test as non root user.
+            self.node.execute(
+                f"chmod -R 777 {self._kself_installed_dir}", sudo=True, shell=True
```
LiliDeng: do we have chmod tool, if yes, please use it

LiliDeng (approved):

PR #2931, opened 2023-08-30
Fix NTTTCP testcases for BSD

dsrivastavv:


microsoft/testsuites/performance/common.py
```diff
@@ -338,8 +345,10 @@ def perf_ntttcp(  # noqa: C901
                 buffer_size = int(65536 / 1024)
             if udp_mode:
                 buffer_size = int(1024 / 1024)
+
             server_result = server_ntttcp.run_as_server_async(
                 server_nic_name,
+                server_ip=server.internal_address,
```
LiliDeng: I recall it may impact nested perf if set the server ip?

dsrivastavv: NTTTCP for freebsd doesn't have a flag to work directly with the nic name, so I need to specify this for freebsd. For other distro, I have modified it to keep the original behavior

LiliDeng (dismissed):

squirrelsc (dismissed):

squirrelsc (approved):

PR #2932, opened 2023-08-31
tools/systemd_analyze: add the plot command

anirudhrb:
The plot command generates an SVG that shows the services started and the startup time taken by each service. Useful for understanding the userspace boot performance.


LiliDeng:
please fix ci check error @anirudhrb

LiliDeng (approved):

PR #2940, opened 2023-09-05
Libvirt: Add MS CLH installer

smit-gardhariya:
This PR contains below Commits:

**libvirt platform: Add MS CLH Installer** : This will add new installer type to install MS cloud-hypervisor for libvirt platform.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/transformers.py
+++ b/lisa/sut_orchestrator/libvirt/transformers.py
@@ -35,6 +35,8 @@ class SourceInstallerSchema(BaseInstallerSchema):
     # source code repo
     repo: str = ""
     ref: str = ""
+    auth_token: str = ""
+    igvm_repo: str = ""
 
 
 @dataclass_json
@@ -378,6 +380,7 @@ class CloudHypervisorSourceInstaller(CloudHypervisorInstaller):
         Ubuntu.__name__: ["gcc"],
         CBLMariner.__name__: ["gcc", "binutils", "glibc-devel"],
     }
+    _build_cmd: str = "cargo build --release"
 
     @classmethod
     def type_name(cls) -> str:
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -114,3 +114,6 @@ class CloudHypervisorNodeSchema(BaseLibvirtNodeSchema):
     # Can be obatained from:
     # https://github.com/cloud-hypervisor/rust-hypervisor-firmware
     firmware: str = ""
+    # guest_vm_type can be NON-CVM or CVM
+    guest_vm_type: str = "NON-CVM"
```
squirrelsc: It should be set by the feature SecurityProfile.

squirrelsc:
@cwize1 Can you help reviewing it?

lisa/sut_orchestrator/libvirt/schema.py
smit-gardhariya: @squirrelsc This is just pipeline parameter used to differentiate what type of libvirt xml we want to create in order to derive guest vm type. It has nothing to do with azure vm security profile. Azure VM here will be standard VM only. 

This will just create xml for cvm/non-cvm guest based on this param value from runbook.

Also this is for guest vm (L2 guest vm), not for L1 vm (azure vm).

squirrelsc: The features folder contains abstract features, they are not for Azure only. The features are for test cases running across platforms. For example, if a CVM test case can run on both Azure and libvirt, the feature can filter and setup right environments for the test case.

smit-gardhariya: Hi Chi, i have removed that commit and will create separate PR for it.

squirrelsc: No problem. Please wait for reviewing from @cwize1

vyadavmsft:
@LiliDeng Can you please review this.

squirrelsc:
> @LiliDeng Can you please review this.

It needs @cwize1 review.

smit-gardhariya:
@cwize1 Could you please help me here with the review :)

squirrelsc (approved):
Approve with PR timeout.

PR #2942, opened 2023-09-05
Add permission grant script for AITL service

paxue:


microsoft/utils/Setup_AITL.ps1
```diff
@@ -0,0 +1,239 @@
+# This script is to create/update a custom role and grant to Service Principal for Partner in Production Env
```
squirrelsc: Add copyright header.

squirrelsc (approved):

Issue #2903, opened 2023-08-16
[Document Request] Allow user specify OS disk size alongside OS image


hcoona:
Hi LISA team,

I'd like to request a new feature to allow user specify OS disk size alongside OS image. Allowing user to specify OS image is a great feature to let user establish a basic environment for testing by burn necessary drivers, configurations, services together into the image. However, the user OS image could be very large and cannot leave enough spaces for the following testing. That's why I want to specify OS disk size.

Regards,
Shuai Zhang

adityagesh:
Hi,

Runbook already has support to override OS Disk Size.
Sample: 

```
platform:
  - type: azure
    azure:
      subscription_id: *******************
    requirement:
      azure:
        marketplace: *******
        location: *********
        osdisk_size_in_gb: 30
```

hcoona:
Thank you! Would be appreciated if we can add it to the document.

![图片](https://github.com/microsoft/lisa/assets/712433/67511d56-d9db-44c7-a7e3-a5ef2e6d1f6a)

![图片](https://github.com/microsoft/lisa/assets/712433/c35b990f-d045-45b9-b775-8073e25f847a)

trafalmuffti:
Obvious presumptions would automate away difficulties in guessing and choosing. People are easily predicted in their guesses for estmiating image size.

LiliDeng:
PR merged

PR #2943, opened 2023-09-06
Add docs for os_disk_size_in_gb

adityagesh:
Closes request #2903

### DIFF
--- a/docs/run_test/platform.rst
+++ b/docs/run_test/platform.rst
@@ -147,6 +147,7 @@ deployment.
             location: "<one or multiple locations, split by comma>"
             vm_size: "<vm size>"
             maximize_capability: "<true or false>"
+            osdisk_size_in_gb: <disk size in gb>
 
 * **virtual_network_resource_group**. Specify if an existing virtual network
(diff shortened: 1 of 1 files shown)


docs/run_test/platform.rst
```diff
@@ -179,6 +180,9 @@ deployment.
   run all test cases. Notice, there are some features are conflict by natural,
   so some test cases may not be picked up. This setting is useful to force run
   perf tests on not designed VM sizes.
+* **osdisk_size_in_gb** is used to specify the minimum size of the OS disk. The size
+  of disk may be larger depending on other requirements set by vhd, shared image gallery
+  or marketplace image.
```
LiliDeng: it is the size of the os disk, not `the minimum`, and when the specified os disk size is less than the default disk size of the  image/vhd/shared image, then default os size will be used. also we need mention when osdisk_size_in_gb will be used, and the range, refer to `https://learn.microsoft.com/en-us/azure/virtual-machines/linux/expand-disks?tabs=ubuntu`

adityagesh: Updated, please check

LiliDeng (approved):

PR #2945, opened 2023-09-06
Fix latency network tests in BSD

dsrivastavv:
FreeBSD 10 and later removed support for XFS, so skipping the testsuite on BSD

### DIFF
--- a/lisa/tools/sockperf.py
+++ b/lisa/tools/sockperf.py
@@ -65,8 +65,18 @@ class Sockperf(Tool):
         r"=\s+(?P<min_latency_us>[0-9]+\.[0-9]+)"
     )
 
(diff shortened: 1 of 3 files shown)


lisa/tools/sockperf.py
```diff
@@ -65,8 +65,18 @@ def can_install(self) -> bool:
...
+        r"Summary: Round trip is (?P<avg_latency_us>[0-9]+\.[0-9]+) usec"
+    )  # noqa: E501
+
+    # Total 1283 observations;
+    sockperf_total_observations = re.compile(
+        r"Total (?P<total_observations>[0-9]+) observations"
+    )  # noqa: E501
+
+    # [Valid Duration] RunTime=0.546 sec; SentMessages=1283; ReceivedMessages=1283
+    sockperf_run_time = re.compile(r"RunTime=(?P<run_time>[0-9]+\.[0-9]+) sec")
```
LiliDeng, line 79: will the units always be usec / sec?

dsrivastavv: It will always be seconds based on the runs i did

LiliDeng (approved):

PR #2948, opened 2023-09-08
DPDK Hugepages: explicit selection of 1GB or 2MB

mcgov:
Mixing hugepage types can cause inconsistent results, move to explicit selection to allow later 2MB / 1GB tests. Old selection was dynamic based on memory size. This means we mix sizes in some cases and exclude sizes in other cases. Move to explicit selection to enable a t4 test for each specific size.

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -87,19 +87,21 @@ class DpdkTestResources:
         self.switch_sriov = True
 
 
-def init_hugepages(node: Node) -> None:
+def init_hugepages(node: Node, enable_gibibyte_hugepages: bool = False) -> None:
     mount = node.tools[Mount]
-    mount.mount(name="nodev", point="/mnt/huge", fs_type=FileSystem.hugetlbfs)
-    mount.mount(
-        name="nodev",
-        point="/mnt/huge-1G",
-        fs_type=FileSystem.hugetlbfs,
-        options="pagesize=1G",
-    )
-    _enable_hugepages(node)
+    if enable_gibibyte_hugepages:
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -273,6 +279,7 @@ def initialize_node_resources(
     variables: Dict[str, Any],
     pmd: str,
     sample_apps: Union[List[str], None] = None,
+    enable_gibibyte_hugepages: bool = False,
```
LiliDeng, line 282: now, all distros won't enable_gibibyte_hugepages

mcgov: This is correct, I'm going to add specific tests for 1GB hugepages later.

mcgov: This commit is just to make the current tests consistent. With this commit we don't create a mount for either type of hugepages unless that size is actually used; we avoid having a dangling mount with 0 pages assigned.

mcgov: Added a second commit to the PR to add the 1GB hugepage tests so we don't lose coverage.

LiliDeng: thanks, please also take care of such image redhat rhel-arm64 8_6-arm64 8.6.2022060901, which doesn't support 1GB hugepage, when 1GB hugepage + such images => skip it

LiliDeng: please handle it in another PR

LiliDeng (approved):

PR #2950, opened 2023-09-13
Removed redundant fde kernel installation code

SRIKKANTH:
Moving .efi files is no longer needed.

### DIFF
--- a/lisa/transformers/kernel_installer.py
+++ b/lisa/transformers/kernel_installer.py
@@ -6,6 +6,7 @@ import re
 from dataclasses import dataclass, field
 from typing import Any, Dict, List, Optional, Type, cast
 
+from assertpy import assert_that
 from dataclasses_json import dataclass_json
 
 from lisa import schema
@@ -122,8 +123,9 @@ class KernelInstallerTransformer(Transformer):
         node = quick_connect(runbook.connection, "installer_node")
 
         uname = node.tools[Uname]
+        kernel_version_before_install = uname.get_linux_information()
         self._log.info(
-            f"kernel version before install: {uname.get_linux_information()}"
+            f"kernel version before install: {kernel_version_before_install}"
         )
         factory = subclasses.Factory[BaseInstaller](BaseInstaller)
         installer = factory.create_by_runbook(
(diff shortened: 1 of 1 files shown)


lisa/transformers/kernel_installer.py
```diff
@@ -140,37 +142,17 @@ def _internal_run(self) -> Dict[str, Any]:
             if installer.runbook.source != "linux-image-azure-fde":
                 posix = cast(Posix, node.os)
                 posix.replace_boot_kernel(installed_kernel_version)
-            else:
```
squirrelsc, line 143: Please check with the code author understand if it's redundant. Is it related to cvm?

SRIKKANTH: @LiliDeng This CVM logic is no longer required and causing failures as it cannot find .efi files under "/usr/lib/linux/efi/".

LiliDeng: will you run old distro which needs this logic?

SRIKKANTH: No, we only use this with proposed kernel kernels. This logic is not needed on all recent CVM kernels on both focal and jammy.

lisa/transformers/kernel_installer.py
```diff
@@ -140,37 +142,17 @@ def _internal_run(self) -> Dict[str, Any]:
...
         node.reboot()
+        kernel_version_after_install = uname.get_linux_information(force_run=True)
         self._log.info(
             f"kernel version after install: "
-            f"{uname.get_linux_information(force_run=True)}"
+            f"{kernel_version_after_install}",
         )
+        assert_that(
+            kernel_version_after_install, "Kernel installation Failed"
+        ).is_equal_to(kernel_version_before_install)
```
LiliDeng, line 152: @squirrelsc long time ago, we had a discussion on this check, do you think we can add the check?

lisa/transformers/kernel_installer.py
```diff
@@ -140,37 +142,14 @@ def _internal_run(self) -> Dict[str, Any]:
             if installer.runbook.source != "linux-image-azure-fde":
```
LiliDeng, line 142: we should remove if

SRIKKANTH: That step is not needed for cvm. So lets keep it.

lisa/transformers/kernel_installer.py
squirrelsc: Yes, it looks good to me.

LiliDeng (approved):

PR #2951, opened 2023-09-13
Bamuir/ade minor version checks

bamuir:
- Mandate 4 cores for full encryption test
- increase time allowed to check for fully encrypted OS
- Check for unsupported OS Major/Minor versions and skip tests

### DIFF
--- a/microsoft/testsuites/vm_extensions/azure_disk_encryption.py
+++ b/microsoft/testsuites/vm_extensions/azure_disk_encryption.py
@@ -241,6 +241,7 @@ class AzureDiskEncryption(TestSuite):
     def _is_unsupported_version(self, node: Node, image: str) -> bool:
         # List of known bad images that should be skipped
         known_bad_images = [
+            # Minimal not supported
             "canonical 0001-com-ubuntu-minimal-kinetic minimal-22_10 22.10.202307010",
             # Missing packages
             "canonical 0001-com-ubuntu-server-focal 20_04-lts 20.04.202007080",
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/azure_disk_encryption.py
```diff
@@ -204,18 +219,81 @@ def _enable_ade_extension(self, node: Node, log: Logger, result: TestResult) ->
...
+            "canonical ubuntuserver 18.04-lts 18.04.202306070",
+            "canonical ubuntuserver 18_04-lts-gen2 18.04.202001210",
+            "canonical ubuntuserver 18_04-lts-gen2 18.04.202004290",
+            "canonical ubuntuserver 18_04-lts-gen2 18.04.202009220",
+            # Mariner is supported after may 2023
+            "microsoftcblmariner cbl-mariner cbl-mariner-2 2.20221122.01",
+            "microsoftcblmariner cbl-mariner cbl-mariner-2 2.20230126.01",
+            "microsoftcblmariner cbl-mariner cbl-mariner-2 2.20230303.02",
+            "microsoftcblmariner cbl-mariner cbl-mariner-2-arm64 2.20230126.01",
+        ]
```
LiliDeng: instead of using the image names, we can use kernel version, since the test image may be gallery image or vhd type,

bamuir: I'm not sure that will work. For example, these 2 images have the same Kernel but we support one and not the other

image

bamuir: removed this check, will use auto triage rules for known bad images

microsoft/testsuites/vm_extensions/azure_disk_encryption.py
```diff
@@ -212,10 +221,48 @@ def _is_supported_linux_distro(self, node: Node) -> bool:
             Ubuntu: 18,
             CBLMariner: 2,
         }
+        # Remove after automatic major version support is released to ADE
+        max_supported_major_versions = {
+            Redhat: 9,
+            CentOs: 8,
+            Oracle: 8,
+            Ubuntu: 22,
```
LiliDeng, line 229: 22.10 doesn't support ADE, we need exclude it

LiliDeng (approved):

PR #2953, opened 2023-09-14
Add aitl python wrapper

squirrelsc:


squirrelsc:
@paxue please refer to this to create initialize script.

PR #2957, opened 2023-09-18
Fix issue that marketplace, shared_gallery, vhd is not None when parameter is empty

lubaihua33:
Before fixing, if the runbook is like below, the azure_runbook,marketplace is not None though its each parameter is empty, this marketplace is invalid, so azure_runbook,marketplace should be None.
    requirement:
      azure:
        marketplace:
          publisher: ""
          offer: ""
          sku: ""
          version: ""
      vhd: ""https://vxxtralus.blob.core.windows.net/vhds/bvtpublisher1/bxxxT_Linux_Genxx_Success/27082023/blob-02.26.20.893.vhd"
      
This PR is to fix this issue in this situation.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -464,7 +464,7 @@ class AzureNodeSchema:
             # this step makes vhd_raw is validated, and
             # filter out any unwanted content.
             self.vhd_raw = vhd.to_dict()  # type: ignore
-        elif self.vhd_raw is not None:
+        elif self.vhd_raw:
             assert isinstance(self.vhd_raw, str), f"actual: {type(self.vhd_raw)}"
             vhd = VhdSchema(vhd_path=self.vhd_raw)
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -326,9 +326,19 @@ def marketplace(self) -> Optional[AzureVmMarketplaceSchema]:
                 marketplace = schema.load_by_type(
                     AzureVmMarketplaceSchema, self.marketplace_raw
                 )
-                # this step makes marketplace_raw is validated, and
-                # filter out any unwanted content.
-                self.marketplace_raw = marketplace.to_dict()  # type: ignore
+                if not all(
```
squirrelsc, line 329: It should raise exception to fix, instead of error with silence.

squirrelsc:
@LiliDeng LGTM, but it needs to test if there is regresion.

LiliDeng (approved):

PR #2958, opened 2023-09-19
Do not catch ReleaseEndOfLifeException

kamalca:
get_environment_information now initializes lspci tool which can trigger EOL exception but it will not be raised again later because the tool won't be initialized again. This fix re-raises the exception.

### DIFF
--- a/lisa/platform_.py
+++ b/lisa/platform_.py
@@ -18,6 +18,7 @@ from lisa.util import (
     InitializableMixin,
     LisaException,
     NotMeetRequirementException,
+    ReleaseEndOfLifeException,
     SkippedException,
     constants,
     hookimpl,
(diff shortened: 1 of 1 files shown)


lisa/platform_.py
```diff
@@ -121,6 +122,11 @@ def get_environment_information(self, environment: Environment) -> Dict[str, str
             information.update(
                 self._get_environment_information(environment=environment)
             )
+        except ReleaseEndOfLifeException as identifier:
+            self._log.exception(
```
squirrelsc: It doesn't need to log out, just raise it. I don't remember if which one can keep the call trace shorter, please pick it.

```python
except ReleaseEndOfLifeException:
    raise

except ReleaseEndOfLifeException as e:
    raise e
```

kamalca:
@squirrelsc Did we decide against this mitigation strategy? Should I close this PR?

Do we have tasks created for more long term mitigation of failures in `get_environment_info`?

squirrelsc:
> @squirrelsc Did we decide against this mitigation strategy? Should I close this PR?
> 
> Do we have tasks created for more long term mitigation of failures in `get_environment_info`?

Yes, please close this PR. @LiliDeng Please make sure the unsupported distro exception doesn't depend on get_env_info.

PR #2959, opened 2023-09-19
MSR: Add platform id regression check

mcgov:
In hyper-v, platform info is passed to the host in an MSR register to the hypervisor. This is used to determine what kind of guest is present. For Linux guests, check that the IsOpenSource flag is set in this info according to:
https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/tlfs/hypercall-interface#guest-os-identity-for-proprietary-operating-systems

### DIFF
--- a/microsoft/testsuites/core/msr.py
+++ b/microsoft/testsuites/core/msr.py
@@ -1,0 +1,149 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import semver
+from assertpy import assert_that
+
+from lisa import (
+    RemoteNode,
+    SkippedException,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    simple_requirement,
+)
+from lisa.operating_system import CBLMariner, Debian, Fedora, Linux, Suse
+from lisa.sut_orchestrator import AZURE
+from lisa.tools import Modprobe
+
+# See docs for hypercall spec, sharing os info
+# is required before making hypercalls.
+#  https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/tlfs/hypercall-interface#guest-os-identity-for-proprietary-operating-systems
+
+# build info: bits 0-15
+BUILD_INFO_MASK = 0x0000_0000_0000_FFFF
+# service pack, bits 16-23
+SERVICE_PACK_MASK = 0x0000_0000_00FF_0000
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/msr.py
```diff
@@ -0,0 +1,91 @@
...
+IS_OPEN_SOURCE_OS = 0x8000_0000_0000_0000
+
+
+@TestSuiteMetadata(
+    area="core",
+    category="functional",
+    description="""
+    Test suite verifies hyper-v platform id is set correctly via hypercall to host.
+    """,
+    requirement=simple_requirement(supported_os=[Linux]),
```
LiliDeng: should we use supported platform in case it fails in other platforms?

microsoft/testsuites/core/msr.py
```diff
@@ -0,0 +1,91 @@
...
+        patch = (msr_register_content & SERVICE_PACK) >> 16
+        os_id = (msr_register_content & OS_ID) >> 40
+        node.log.info(f"OS ID: {hex(os_id)}")
+        node.log.info(
+            "Found os build info: (major, minor, patch, build) "
+            f"{major}.{minor}.{patch}-{build}"
+        )
+        assert_that(msr_register_content & IS_OPEN_SOURCE_OS).described_as(
+            "OS_TYPE not set to OPEN_SOURCE in bitfield. "
+            f"Expected {hex(msr_register_content)} & {hex(IS_OPEN_SOURCE_OS)} != 0."
```
LiliDeng: can we make the description easy to understand?

microsoft/testsuites/core/msr.py
mcgov: Updated

microsoft/testsuites/core/msr.py
mcgov: I'm not sure? It's pretty simple in C terms, just parsing a bitfield. I wrapped the properties into a class which will pretty print, so hopefully it's more clear what's happening

mcgov: There's extra code because I'm sure at some point we'll want to check the other fields. Right now we just need to check the IsOpenSource field for this test. This first OsPlatformInfo struct is used in the first hypercall, most of it is optional but if people mess the fields up they won't get nice things like VM health info.

LiliDeng (approved):

PR #2960, opened 2023-09-20
add setup_AITL.py for role assignment

paxue:


microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+# User sample:  .\Setup_AITL.py
+
+import argparse
+import json
+import subprocess
+import sys
+import time
+
+
+def call_cmd(command: str) -> subprocess.CompletedProcess[str]:
```
squirrelsc: Add underscore, if it's a private method.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+# This script is to create/update a custom role and grant to Service Principal for Partner in Production Env
+# Debug sample: .\Setup_AITL.py -r "AzureLSGDelegator_test" -p "AzureLinuxTestingService" -s "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
```
squirrelsc: use all small case on file name.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+            text=True,
+        )
+        return_code = result.returncode
+        print(f"Return Code: {return_code} for cmd: {command}")
+        # output = result.stdout
+        # print(f"Output:\n{output}") # can open for debug purpose
+        return result
+    except subprocess.CalledProcessError as e:
+        # If the command returns a non-zero exit code, you can handle the error here
+        print(f"Command failed with error: {e.stderr}")
```
squirrelsc: It doesn't need to catch error, if it just exits. Let the exception throw out directly.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+        sys.exit(1)
+
+
+def _compare_str_list(str_list_1: list[str], str_list_2: list[str]) -> bool:
+    if len(str_list_1) != len(str_list_2):
+        return False
+
+    str_list_1.sort()
+    str_list_2.sort()
+    return str_list_1 == str_list_2
```
squirrelsc: try `return sorted(str_list_1) == sorted(str_list_2)`. Even it doesn't need to check if length is the same or not.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def _compare_role_setting(role1_json: str, role2_json: str) -> bool:
+    # input from az role list output, is an array: [role...]
+    if role1_json == "" or role2_json == "":
```
squirrelsc: use below, it handles both empty string and None.

```python
if not role1_json or not role2_json.
```

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+def _compare_str_list(str_list_1: list[str], str_list_2: list[str]) -> bool:
+    if len(str_list_1) != len(str_list_2):
+        return False
+
+    str_list_1.sort()
+    str_list_2.sort()
+    return str_list_1 == str_list_2
+
+
+def _compare_role_setting(role1_json: str, role2_json: str) -> bool:
```
squirrelsc: Don't use role1_list, role2_list. Use expected_definition, actual_definition.

1. Avoid using unmeaningful names like 1, 2.
2. Avoid to use type in the name like _list, unless mutiple types on the same meaning.

microsoft/utils/setup_AITL.py
paxue: At the time of write this function I thought this "call_cmd" can be used outside. But it's OK to turn into private

squirrelsc: Because it will be used independently, so let's keep all things private.

squirrelsc:
Remove the old file in the same PR.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+    # input from az role list output, is an array: [role...]
+    if role1_json == "" or role2_json == "":
+        return False
+
+    role1_array = json.loads(role1_json)
+    role2_array = json.loads(role2_json)
+    if len(role1_array) != 1 or len(role2_array) != 1:
```
squirrelsc: Add comments above, why it's expected to be 1

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/add/action",
+]
+
+target_role = {
+    "Description": "Custmor role is to run test cases and upload logs in Prod env.",
+    "IsCustom": True,
+    "Name": role_name,
+    "Actions": action_perms_list,
+    "DataActions": data_action_perms_list,
+    "AssignableScopes": [subscription_id],
```
squirrelsc: Does it needed? I'm thinking if users want to enable AITL on multiple subscriptions, this may fail.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+]
+data_action_perms_list = [
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/delete",
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/read",
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/write",
+    "Microsoft.Storage/storageAccounts/blobServices/containers/blobs/add/action",
+]
+
+target_role = {
+    "Description": "Custmor role is to run test cases and upload logs in Prod env.",
```
squirrelsc: Mention about AITL in the description. It's not related to "prod" env. User doesn't have that concept.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+
+        is_roles_match = _compare_role_setting(role_before_update, role_after_update)
+        if is_roles_match is False:
+            print("Found changed results.")
+            count += 1
+            change_count += 1
+        else:
+            change_count = 0
+
+        time.sleep(20)
```
squirrelsc: It's too long, 5 seconds is max.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+
+    elapsed_time = time.time() - start_time
+    print(f"Role Propagation took time: {elapsed_time}")
```
squirrelsc: It should tell the result like "Role is propagated in {elapsed_time}."

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+# Main Function
+parser = argparse.ArgumentParser(
```
squirrelsc: Add a method to organize the parser logic.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+)
+
+args = parser.parse_args()
+role_name = args.role
+service_principal_name = args.service_principal
+subscription_id = args.subscriptionId
+
+# Step 1. set the default parameter for subscription_id, actions, dataActions
+if subscription_id == "":
+    sub_cmd_result = call_cmd('az account list --query "[0].id" --output tsv')
```
squirrelsc: It should use current subscription, instead of first one.

microsoft/utils/setup_AITL.py
```diff
@@ -0,0 +1,299 @@
...
+        )
+        _wait_role_propagate(role_name, subscription_id, role_before_update)
+
+# Step 3. assign this role to Service Principal
+get_service_principal_result = call_cmd(
+    f'az ad sp list --display-name "{service_principal_name}" \
+        --filter "servicePrincipalType eq \'Application\'" '
+)
+service_principal_list = json.loads(get_service_principal_result.stdout)
+service_principal_objectid = service_principal_list[0]["objectId"]
```
squirrelsc: It needs to check the service principal exists or not. If not, it needs to raise an exception to ask user to register resource provider firstly.

microsoft/utils/setup_AITL.py
paxue: The 5 sec is not sufficient. when the role is created, it takes about 4~5 min to be ready to assign to the service principal.

microsoft/utils/setup_AITL.py
paxue: The "subscription_id" is needed in the scope. The current design: for one customer, if they have different subscriptions, they need to create different roles with different scopes.
 az role list command has some limitation, if doesn't specify "subscription" in scope, then this role can't be found.

microsoft/utils/setup_AITL.py
squirrelsc: Sorry, I forget to mention, it needs to give the full name.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,297 @@
...
+            sys.exit(1)
+
+        print("Checking change applied...")
+
+        get_role_cmd_result = _call_cmd(
+            f'az role definition list --name "{role_name}" --scope "{subscription_id}'
+        )
+        role_after_update = get_role_cmd_result.stdout
+        if role_after_update.startswith("[]"):  # for role creation
+            time.sleep(5)
```
squirrelsc: It should be like below and not need to sleep and continue.
```python
        if not role_after_update.startswith("[]"):
```

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,297 @@
...
+            print(f"Fail to detect changed custom role after {count} times")
+            sys.exit(1)
+
+        print("Checking change applied...")
+
+        get_role_cmd_result = _call_cmd(
+            f'az role definition list --name "{role_name}" --scope "{subscription_id}'
+        )
+        role_after_update = get_role_cmd_result.stdout
+        if role_after_update.startswith("[]"):  # for role creation
```
squirrelsc: Don't write comments in same line with code.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,297 @@
...
+
+def _wait_role_propagate(
+    role_name: str, subscription_id: str, role_before_update: str = ""
+) -> None:
+    print("In _wait_role_propagate: waiting for the role changes to propagate")
+    start_time = time.time()
+
+    change_count = 0
+    count = 0
+    max_count = 30
```
squirrelsc: use timeout, and interval. So later, it's easy to modify, if the time is not enough.

squirrelsc: Even you can directly check the interval time by compare two datetime value.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,297 @@
...
+        )
+        _wait_role_propagate(role_name, subscription_id, role_before_update)
+
+# Step 3. assign this role to Service Principal
+get_service_principal_result = _call_cmd(
+    f'az ad sp list --display-name "{service_principal_name}" \
+        --filter "servicePrincipalType eq \'Application\'" '
+)
+service_principals = json.loads(get_service_principal_result.stdout)
+if len(service_principals) == 0:
```
squirrelsc: It should be moved before create custom role. So it won't wait custom role.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,297 @@
...
+# Main Function
+args = _init_arg_parser()
+
+role_name = args.role
+service_principal_name = args.service_principal
+subscription_id = args.subscriptionId
+
+# Step 1. set the default parameter for subscription_id, actions, dataActions
+if subscription_id == "":
+    sub_cmd_result = _call_cmd('az account list --query "[?isDefault].id" --output tsv')
```
squirrelsc: Use `az account show`

microsoft/utils/setup_aitl.py
paxue: I don't think this is a good idea. After the creation cmd sent, it takes about 1~2 min to finish the role creation. We need to sleep for waiting. If not sleep, the program will continue checking role all the time before it's finished creation. I am not sure  this is what we want.

squirrelsc: It's not to change the logic, the full change is like below,

```python
        if not role_after_update.startswith("[]"): 
            is_roles_match = _compare_role_setting(role_before_update, role_after_update)
            if is_roles_match is False:
                print("Found changed results.")
                count += 1
                change_count += 1
            else:
                change_count = 0

        time.sleep(5)
```

paxue: I changed to "not role_after_update.startswith("[]"):", but still waiting

microsoft/utils/setup_aitl.py
paxue: I got this cmd from link. Is using "az account show" for code readibility?
https://learn.microsoft.com/en-us/cli/azure/manage-azure-subscriptions-azure-cli

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+        default="",
+        help="Subscription scope in Role assignment(default is current subscription)",
+    )
+
+    return parser.parse_args()
+
+
+def _set_target_role_parameters(
+    subscription_id: str, role_name: str
+) -> tuple[str, str]:
```
squirrelsc: Use `Tuple` from typing to be compatible with 3.8. You can see pipeline errors to fix it.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+    result = subprocess.run(
+        command,
+        shell=True,
+        check=True,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    return_code = result.returncode
+    logging.info(f"Return Code: {return_code} for cmd: {command}")
```
squirrelsc: move it to debug level.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+        logging.info("Checking change applied...")
+        get_role_cmd_result = _call_cmd(
+            f'az role definition list --name "{role_name}" --scope "{subscription_id}"'
+        )
+        role_after_update = get_role_cmd_result.stdout
+
+        # for role creation completed
+        if role_after_update.startswith("[]"):
+            count = 0
+            logging.info("--- No role found.")
```
squirrelsc: move it to debug level.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+        else:
+            is_same = _compare_role_setting(role_before_update, role_after_update)
+            if is_same:
+                # if no changes then reset count 0
+                count = 0
+                logging.info("--- Role setting Not Updated.")
```
squirrelsc: move it to debug level.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+            else:
+                count += 1
+                logging.info("--- Found changed results.")
```
squirrelsc: move it to debug level.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+        if count >= required_continuous_change_count:
+            break
+        time.sleep(sleep_interval)
+
+    elapsed_time = time.time() - start_time
+    logging.info(f"Role is propagated in {elapsed_time} sec")
+
+
+def _init_arg_parser() -> Namespace:
+    parser = ArgumentParser(
```
squirrelsc, line 154: Add debug parameter to show debug level log.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+
+        is_same_role = _check_same_role_existed(exist_role, target_role_json)
+        if is_same_role is True:
+            logging.info(
+                f"Input Role {role_name} has same setting with exist one, won't update"
+            )
+        else:
+            # only update when different.
+            logging.info(f"Updating existed role: {role_name} with parameters: ")
+            logging.info(target_role_json)
```
squirrelsc: move it to debug level.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+                f"az role assignment create --role {role_name} --assignee-object-id "
+                f" {service_principal_objectid} --scope {subscription_id}"
+            )
+
+            logging.info(
+                f"Succeeded to Assign role to ServicePrincipal: in {i+1} times trial"
+            )
+            break
+        except subprocess.CalledProcessError as e:
+            logging.info(
```
squirrelsc: move it to debug level.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+                f"Failed in {i+1} times retry: Assign role to ServicePrincipal"
+            )
+            logging.info(f"Error Info: {e.stderr}")
```
squirrelsc: move it to debug level.

microsoft/utils/setup_aitl.py
```diff
@@ -0,0 +1,365 @@
...
+) -> None:
+    logging.info(
+        f"Assign role {role_name} to ServicePrincipal {service_principal_name} "
+        f"with ID: {service_principal_objectid}"
+    )
+
+    max_tries = 10
+    for i in range(max_tries):
+        try:
+            time.sleep(5)
```
squirrelsc: Don't sleep before take actions. Only sleep when it needs to retry.

squirrelsc (approved):

PR #2963, opened 2023-09-20
enable serial console log for idrac and add ClusterCapabilities

LiliDeng:


lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -80,19 +107,67 @@ def type_schema(cls) -> Type[schema.TypedSchema]:
...
-        self._change_boot_order_once("VCD-DVD")
         assert self.client.iso_http_url, "iso_http_url is required for idrac client"
+        self._change_boot_order_once("VCD-DVD")
         if self.get_power_state() == "Off":
             self._log.debug("System is already off.")
         else:
-            self.reset("GracefulShutdown")
+            self.reset("ForceOff")
         self._insert_virtual_media(self.client.iso_http_url)
+        check_till_timeout(
```
squirrelsc: insert media after it's off. I'm not sure whether weird behaviors when inserting and turn off happening together.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
```diff
@@ -11,6 +11,12 @@
 from ..schema import ClusterSchema
 
 
+class ClusterCapabilities:
```
squirrelsc: Do you want to get cluster/server capability, or client capability? For rack manager, it doesn't make sense to get cluster (rack manager) capability.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -80,19 +107,67 @@ def type_schema(cls) -> Type[schema.TypedSchema]:
...
+            response.dict["ProcessorSummary"]["LogicalProcessorCount"]
+        )
+        cluster_capabilities.free_memory_mb = (
+            int(response.dict["MemorySummary"]["TotalSystemMemoryGiB"]) * 1024
+        )
+        self.logout()
+        return cluster_capabilities
+
+    def get_serial_console_log(self) -> str:
+        response = self.redfish_instance.post(
```
squirrelsc, line 147: Make sure it's cleaned up, when deploy a new environment.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -80,19 +107,67 @@ def type_schema(cls) -> Type[schema.TypedSchema]:
...
             self._log.debug("System is already off.")
         else:
-            self.reset("GracefulShutdown")
+            self.reset("ForceOff")
         self._insert_virtual_media(self.client.iso_http_url)
+        check_till_timeout(
+            lambda: self.get_power_state() == "Off",
+            timeout_message="wait for client into 'Off' state",
+        )
+        self._enable_serial_console()
```
squirrelsc: It should be done when initialize platform, not deploy.

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
```diff
@@ -70,3 +70,9 @@ def reset(self, operation: str) -> None:
             rm_node.execute(f"set system {operation} -i {client.management_port}")
 
         self._log.debug(f"client has been {operation} successfully")
+
+    def get_cluster_capabilities(self) -> ClusterCapabilities:
+        cluster_capabilities = ClusterCapabilities()
+        cluster_capabilities.core_count = 8
```
squirrelsc: For rack manager, we can get it by two ways.

1. write values in runbook
2. Load real values, when environment is booted up.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -54,16 +54,17 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self.build_factory = Factory[Build](Build)
 
     def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
-        return self._configure_node_capabilities(environment, log)
-
-    def _deploy_environment(self, environment: Environment, log: Logger) -> None:
         # currently only support one cluster
         assert self._baremetal_runbook.cluster, "no cluster is specified in the runbook"
-        cluster_instance = self._baremetal_runbook.cluster[0]
+        self.cluster_instance = self._baremetal_runbook.cluster[0]
```
squirrelsc: rename it to cluster_runbook. Keep it private, if it's used only by subclasses.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
LiliDeng: I want to get client capability, to create node capabilities

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
LiliDeng: if use 2, probably some cases will be skipped before deployment?

squirrelsc: That's ok, we have best tries. But we can still leave default value here, maybe a max value to prevent any case is blocked.

lisa/sut_orchestrator/baremetal/cluster/cluster.py
squirrelsc: If so, it should be named as ClientCapabilities, and the method should consider to input client information.

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
pupacha: Would all of the clients be having same capabilities for Rack Manager case ? If not, `get_cluster_capabilities` should return a list.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -41,7 +43,7 @@ def _stop(
         if self.cluster.get_power_state() == "Off":
             self._log.debug("System is already off.")
             return
-        self.cluster.reset("GracefulShutdown")
+        self.cluster.reset("ForceOff")
```
pupacha: Since its a feature and can be used inside a testcase, I think we should avoid "ForceOff" here. May be use "GracefulShutdown" and wait until get_power_state() is "Off", if this works as expected?

pupacha (dismissed):

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -52,18 +52,20 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+        self._cluster_runbook = self._baremetal_runbook.cluster[0]
+        self.cluster = self.cluster_factory.create_by_runbook(self._cluster_runbook)
 
-        self.cluster = self.cluster_factory.create_by_runbook(cluster_instance)
+    def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
         assert self.cluster.runbook.client, "no client is specified in the runbook"
+        if self.cluster.runbook.client[0].capabilities:
+            client_capabilities = self.cluster.runbook.client[0].capabilities
+        else:
+            client_capabilities = self.cluster.get_client_capabilities()
```
squirrelsc: Pass in the client object, when calling get_client_capabilities.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -50,14 +56,47 @@ def _start(self, wait: bool = True) -> None:
             self._log.debug("System is already powered on.")
             return
         self.cluster.reset("On")
+        check_till_timeout(
```
squirrelsc: Add it to `reset` method, so it won't be repeated.

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -50,14 +56,47 @@ def _start(self, wait: bool = True) -> None:
...
+        self.cluster = platform.cluster  # type: ignore
+        self.cluster.logout()
+
+    def _get_console_log(self, saved_path: Optional[Path]) -> bytes:
+        self._login()
+        if saved_path:
+            screenshot_file_name: str = "serial_console"
+            decoded_data = base64.b64decode(self.cluster.get_server_screen_shot())
+            screenshot_raw_name = saved_path / f"{screenshot_file_name}.png"
+            img_file = open(screenshot_raw_name, "wb")
```
squirrelsc: Use `with` make sure the handle is closed.

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
```diff
@@ -54,19 +53,32 @@ def type_schema(cls) -> Type[schema.TypedSchema]:
...
+            self.rm_node.execute(f"set system {operation} -i {client.management_port}")
         self._log.debug(f"client has been {operation} successfully")
+
+    def get_client_capabilities(self) -> ClientCapabilities:
+        cluster_capabilities = ClientCapabilities()
+        cluster_capabilities.core_count = 0
+        cluster_capabilities.free_memory_mb = 0
+        return cluster_capabilities
+
+    def clean_up(self) -> None:
```
squirrelsc: cleanup

lisa/sut_orchestrator/baremetal/platform_.py
LiliDeng: update

lisa/sut_orchestrator/baremetal/cluster/idrac.py
LiliDeng: update

lisa/sut_orchestrator/baremetal/cluster/idrac.py
LiliDeng: update

lisa/sut_orchestrator/baremetal/cluster/rackmanager.py
LiliDeng: update

lisa/sut_orchestrator/baremetal/cluster/idrac.py
```diff
@@ -80,26 +108,81 @@ def type_schema(cls) -> Type[schema.TypedSchema]:
...
+            "/redfish/v1/Dell/Managers/iDRAC.Embedded.1/DellLCService/Actions/"
+            "DellLCService.ExportServerScreenShot",
+            body={"FileType": file_type},
+        )
+        self._wait_for_completion(response)
+        return str(response.dict["ServerScreenShotFile"])
+
+    def reset(self, operation: str, force_run: bool = False) -> None:
+        if not force_run and operation in self.state_dict.keys():
+            if self.get_power_state() == self.state_dict[operation]:
```
squirrelsc: Get the dict value one time, and use it in next places. So the logic would be easier to understand, and lines shorter.

squirrelsc (dismissed):

PR #2964, opened 2023-09-20
Azure: fix resolving wrong latest version.

squirrelsc:
Fix issue like 8.5.8 is before 8.5.18, when resolving "latest" tag on version.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1,6 +1,5 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-
 import copy
 import json
 import logging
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1904,8 +1904,10 @@ def _resolve_marketplace_image(
                             f"{marketplace.offer} {marketplace.sku} in {location}"
                         )
                     else:
+                        # use the same sort approach as Az CLI.
+                        versioned_images.sort(key=lambda x: parse(x.name), reverse=True)
                         # any one should be the same to get purchase plan
```
kamalca: Does this comment still add any useful information? This function is not being used for the sole purpose of getting the purchase plan information.

squirrelsc: Good catch

LiliDeng (approved):

PR #2966, opened 2023-09-21
Add password extension feature

lubaihua33:
This PR is to resolve the scenario that some images need to input a password when running the sudo command. Like primekey, vmware-inc images, they have a service to change the sudoers. The password and username which is set in the arm template and configured in the image by Cloud-init might be invalid. So need to reset a password if encounter this situation.

### DIFF
--- a/lisa/features/password_extension.py
+++ b/lisa/features/password_extension.py
@@ -1,0 +1,25 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.feature import Feature
+
+FEATURE_NAME_PASSWORD_EXTENSION = "PasswordExtension"
+
+
+class PasswordExtension(Feature):
+    @classmethod
+    def name(cls) -> str:
+        return FEATURE_NAME_PASSWORD_EXTENSION
+
+    @classmethod
+    def can_disable(cls) -> bool:
+        raise NotImplementedError()
+
+    def enabled(self) -> bool:
+        raise NotImplementedError()
+
+    def is_supported(self) -> bool:
+        raise NotImplementedError()
+
+    def reset_password(self, username: str, password: str) -> None:
+        raise NotImplementedError()
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -22,6 +22,7 @@ from lisa.util import (
     RequireUserPasswordException,
     constants,
     fields_to_dict,
+    generate_strong_password,
     get_datetime_path,
     hookimpl,
     hookspec,
@@ -683,6 +684,8 @@ class RemoteNode(Node):
         password_extension = self.features[PasswordExtension]
         username = self._connection_info.username
         password = self._connection_info.password
+        if not password:
+            password = generate_strong_password()
         try:
             password_extension.reset_password(username, str(password))
         except Exception as identifier:
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -12,6 +12,7 @@ from lisa.executable import Tools
 from lisa.feature import Features
 from lisa.nic import Nics, NicsBSD
 from lisa.operating_system import BSD, OperatingSystem
+from lisa.secret import add_secret
 from lisa.tools import Chmod, Df, Echo, Lsblk, Mkfs, Mount, Reboot, Uname
 from lisa.tools.mkfs import FileSystem
 from lisa.util import (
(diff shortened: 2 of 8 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1116,7 +1117,7 @@ def _create_deployment_parameters(
             arm_parameters.admin_key_data = get_public_key_data(
                 self.runbook.admin_private_key_file
             )
-        else:
+        if self.runbook.admin_password:
```
squirrelsc: remove the logic, to set both always.

lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -472,12 +472,16 @@
                         {
                             "name": "publicKeyData",
                             "type": "string"
+                        },
+                        {
+                            "name": "disable_password_authentication",
```
squirrelsc, line 477: Does it impact password only or password first scenario, like Windows?

lisa/node.py
```diff
@@ -160,24 +154,49 @@ def check_sudo_password_required(self) -> None:
...
+                                "The password might be invalid for running sudo command"
+                            )
                     password_prompt = result.stdout.replace(
                         f"{constants.LISA_TEST_FOR_SUDO}", ""
                     )
                     password_prompts.append(password_prompt)
                     self.log.debug(f"password prompt {i}: {password_prompt}")
                 ssh_shell.password_prompts = password_prompts
 
+    def _execute_sudo_test(self) -> ExecutableResult:
```
squirrelsc: 1. Return bool value to indicate success or not.
2. Shorter name like `test_sudo`

lisa/node.py
```diff
@@ -160,24 +154,49 @@ def check_sudo_password_required(self) -> None:
...
+        process = self._execute(
+            f"echo {constants.LISA_TEST_FOR_SUDO}",
+            shell=True,
+            sudo=True,
+            no_info_log=True,
+        )
+        return process.wait_result(10)
+
+    def _is_password_reset_successful(self) -> bool:
```
squirrelsc: `_reset_password`

lisa/node.py
```diff
@@ -160,24 +154,49 @@ def check_sudo_password_required(self) -> None:
...
+        from lisa.features import PasswordExtension
```
squirrelsc: Move to top, it doesn't need to be imported here.

lisa/node.py
```diff
@@ -160,24 +154,49 @@ def check_sudo_password_required(self) -> None:
...
+
+        if self.features.is_supported(PasswordExtension):
+            ssh_shell = cast(SshShell, self.shell)
+            username = ssh_shell.connection_info.username
+            password = ssh_shell.connection_info.password
```
squirrelsc: If the password doesn't exist, we can generate one randomly.

lisa/node.py
```diff
@@ -160,24 +154,49 @@ def check_sudo_password_required(self) -> None:
...
                 password_prompts = []
                 for i in range(1, 3):
-                    process = self._execute(
-                        f"echo {constants.LISA_TEST_FOR_SUDO}",
-                        shell=True,
-                        sudo=True,
-                        no_info_log=True,
-                    )
-                    result = process.wait_result(10)
+                    result = self._execute_sudo_test()
```
squirrelsc: It needs a refactoring on `support_sudo`. The `check_sudo_password_required` should be moved to remote node. The remote node overrides `support_sudo`, if the node.support_sudo return false. It will be clearer, and it doesn't need many logic on cast remote node.

```python
# Node class
    def support_sudo(self) -> bool:
        self.initialize()

        # check if sudo supported
        if self._support_sudo is None:
            if self.is_posix:
                result = process.wait_result(10)
                if result.exit_code == 0:
                    self._support_sudo = True
                else:
                    self._support_sudo = False
                    self.log.info("node doesn't support sudo, may cause failure later.")
            else:
                # set Windows to true to ignore sudo asks.
                self._support_sudo = True

        return self._support_sudo

# Remote node class
    def support_sudo(self) -> bool:
        if self._support_sudo is None:
                result = super.support_sudo
                if not result:
                    self.check_sudo_password_required()

File: lisa/sut_orchestrator/azure/arm_template.json (hunk 2 above)
Comment (lubaihua33) [reply lisa/sut_orchestrator/azure/arm_template.json:477]: This is in fun getLinuxConfiguration, so it is only for Linux.  In fun getOsProfile, only when the key data is not empty and is Linux, then it calls getLinuxConfiguration fun.

So this won't impact the password only or the password first scenario.

File: lisa/features/password_extension.py [hunk 8]
```diff
@@ -0,0 +1,28 @@
...
+    def enabled(self) -> bool:
+        raise NotImplementedError()
+
+    def is_supported(self) -> bool:
+        raise NotImplementedError()
+
+    def reset_password(self, username: str, password: str) -> None:
+        raise NotImplementedError()
+
+    def generate_password(self) -> str:
```
squirrelsc: `generate_password` should be implement here, because it's shared.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2464,3 +2465,56 @@ def create_setting(
...
+            type_="VMAccessForLinux",
+            type_handler_version="1.5",
+            auto_upgrade_minor_version=True,
+            protected_settings=reset_user_password,
+        )
+        assert_that(result["provisioning_state"]).described_as(
+            "Expected the extension to succeed"
+        ).is_equal_to("Succeeded")
+
+    def generate_password(self) -> str:
```
squirrelsc: I implemented a version, please pick it from my branch. It can be put in utils.

https://github.com/microsoft/lisa/commit/79569fd7f4c4f8e484ea9c79b72ca402e94a54eb

lisa/sut_orchestrator/azure/platform_.py

lisa/node.py
lubaihua33: If move to top, it has a circle import error.

lisa/features/password_extension.py
lubaihua33: Deleted this function.

lisa/sut_orchestrator/azure/features.py
lubaihua33: Have picked. Added some changes because '\' and '-' are not allowed in the password according to the Azure doc.

lisa/features/password_extension.py
squirrelsc: use cherry pick to pick the commit, instead of merging it into existing commits. You can reorder commits, to make sure the committed order is valid.

lisa/sut_orchestrator/azure/features.py
squirrelsc: use cherry pick to pick the commit, instead of merging it into existing commits. You can reorder commits, to make sure the committed order is valid.

squirrelsc:
@LiliDeng LGTM. Please find a right time to merge. Please skip release during China national holidays.

LiliDeng:
@lubaihua33 please rebase the code, thanks.

LiliDeng (approved):

PR #2968, opened 2023-09-22
Run `environment_connected` phase after deploy_environment

pupacha:
This change will allow to run transformers with `environment_connected` phase after a new environment is deployed as well

### DIFF
--- a/lisa/runners/lisa_runner.py
+++ b/lisa/runners/lisa_runner.py
@@ -279,6 +279,11 @@ class LisaRunner(BaseRunner):
                     environment.status == EnvironmentStatus.Deployed
                 ), f"actual: {environment.status}"
                 self._reset_awaitable_timer("deploy")
+                transformer.run(
+                    self._runbook_builder,
+                    phase=constants.TRANSFORMER_PHASE_ENVIRONMENT_CONNECTED,
+                    environment=environment,
+                )
             except ResourceAwaitableException as identifier:
                 if self._is_awaitable_timeout("deploy"):
                     self._log.info(


lisa/runners/lisa_runner.py
```diff
@@ -279,6 +279,11 @@ def _deploy_environment_task(
                     environment.status == EnvironmentStatus.Deployed
                 ), f"actual: {environment.status}"
                 self._reset_awaitable_timer("deploy")
+                transformer.run(
```
squirrelsc, line 282: It's not the place to run this phase, because the environment is deployed, but not connected. Which scenario needs this?

pupacha: I'm looking to run transformer whenever the platform deploys a new environment. Is there a better place where I can put this `transformer.run()` ?

squirrelsc: Kameron already supports to run the transformer after initialized. I don't know what's difference with your scenario. Can you explain the details, or use the transformer type directly?

pupacha: I created a sample transformer and tried using it with main (which has Kameron's code). And it doesn't work with the scenario I explained above. From what I understand, Kameron's code triggers TRANSFORMER_PHASE_ENVIRONMENT_CONNECTED phase only with `_initialize_environment_task` at https://github.com/microsoft/lisa/blob/main/lisa/runners/lisa_runner.py#L312 . 
And `_initialize_environment_task` is not called everytime a new environment is created by platform. I might be wrong, but from what I understand, `_initialize_environment_task` is called when `environment` is specified in the runbook.

pupacha: Actually, `environment_connected` does not sound apt for the usecase I was mentioning about. It has to be more like `environment_deployed`. But creating one more phase, I feel, is overkill scaling. Hence, thought of using the same phase here as well. Is it fine if I can rename the existing `TRANSFORMER_PHASE_ENVIRONMENT_CONNECTED` to `TRANSFORMER_PHASE_ENVIRONMENT_DEPLOYED`.

squirrelsc: The "connected" phase won't be executed, if a test case runs on deployed phase. And an environment won't call to the transformer, if a deployed test case run. Because the environment won't be connected by runner. I will update the code to mark environment is dirty, if a deployed phase test cases run.

I don't see the "scenario" which you need the change. So, I cannot evaluate if a new transformer needed, or you can use this one. In Kameron's scenario, the transformer is used for distro upgrading testing. It doesn't need to run smoke tests, which runs on deployed phase. So, the connected phase works in his scenario.

pupacha: My scenario is that I want to run certain scripts, config changes etc on a newly created node by the platform before the test execution. To give more details, we have an image and we want to run tests but a change in some kernel config.
Basically I need following workflow.

`platform.deploy_env()` -> `Make config change (using transformer ?)` -> `Run tests on the created environment`

squirrelsc: > Actually, `environment_connected` does not sound apt for the usecase I was mentioning about. It has to be more like `environment_deployed`. But creating one more phase, I feel, is overkill scaling. Hence, thought of using the same phase here as well. Is it fine if I can rename the existing `TRANSFORMER_PHASE_ENVIRONMENT_CONNECTED` to `TRANSFORMER_PHASE_ENVIRONMENT_DEPLOYED`.

Again, please sharing the scenario, before proposing the solution. That's how we can evaluate the right solution. I have discussed with Kameron a lot. Renaming it will break smoke tests, because smoke tests need to be first to connect to an environment. So if there is any environment failure, it should be first to know it.

pupacha: Yes, shared it above.

squirrelsc: > My scenario is that I want to run certain scripts, config changes etc on a newly created node by the platform before the test execution. To give more details, we have an image and we want to run tests but a change in some kernel config. Basically I need following workflow.
> 
> `platform.deploy_env()` -> `Make config change (using transformer ?)` -> `Run tests on the created environment`

Try to exclude all smoke tests, see if it works. The scenario is similar with Kameron's. The smoke tests is not effective in this case, because the transformer needs to connect to the environment firstly. So image should be work, before any test cases. After I marked env dirty, if a test case is run on deployed phase, you don't need to exclude the smoke tests. But the smoke tests check the environment before any change applied, so the transformer won't be executed.

pupacha: Ok, I will run the test `verify_cpu_count` and check if it works for my use case. Thanks.

pupacha: It actually worked. Thanks @squirrelsc for the suggestion. I think now I sort of understand what was happening. I wasn't aware of how exactly the test task dispatching worked. But now, kind of got an overview.

If I understand correctly, `environment_connected` transformer can be used for my usecase given the tests do not have `environment_status=EnvironmentStatus.Deployed` as requirement.

squirrelsc: Yes, I will submit a PR today. So, you can include smoke tests as well. But smoke test won't run the transformers of connected phase.

kamalca: > My scenario is that I want to run certain scripts, config changes etc on a newly created node by the platform before the test execution. To give more details, we have an image and we want to run tests but a change in some kernel config. Basically I need following workflow.
> 
> `platform.deploy_env()` -> `Make config change (using transformer ?)` -> `Run tests on the created environment`

@pupacha Can you use the azure deploy and export to vhd transformers for your use case? (See the kernel validation pipelines for examples)

It is typically better to apply config changes once and run testing on the updated image than to apply the changes for every test.

pupacha: Right. That is a way for making things work for `azure` platform but we are using the new `baremetal` platform, its a costly operation to deploy and regenerate vhd/iso after config change on a baremetal server. Hence, this requirement to run transformer before test execution starts.

PR #2970, opened 2023-09-22
Ultra disk testcases added.

SRIKKANTH:
Functional and perf test cases for ultradisk are added.

### DIFF
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -144,6 +144,7 @@ DiskType = Enum(
         "unknown",
         "nvme",
         "premiumssd",
+        "ultradisk",
     ],
 )
 
--- a/microsoft/testsuites/core/provisioning.py
+++ b/microsoft/testsuites/core/provisioning.py
@@ -16,6 +16,8 @@ from lisa import (
     TestSuite,
     TestSuiteMetadata,
     create_timer,
+    schema,
+    search_space,
     simple_requirement,
 )
 from lisa.environment import EnvironmentStatus
@@ -189,6 +191,31 @@ class Provisioning(TestSuite):
             log_path,
             reboot_in_platform=True,
             case_name="verify_reboot_in_platform",
+        )
+
+    @TestCaseMetadata(
+        description="""
+        This case runs smoke test on a node provisioned with an ultra datadisk.
+        The test steps are same as `smoke_test`.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            disk=schema.DiskOptionSettings(
+                data_disk_type=schema.DiskType.UltraSSDLRS,
+                data_disk_count=search_space.IntRange(min=1),
+            ),
+            environment_status=EnvironmentStatus.Deployed,
+            supported_features=[SerialConsole],
+        ),
+    )
+    def verify_deployment_provision_ultra_datadisk(
+        self, log: Logger, node: RemoteNode, log_path: Path
+    ) -> None:
+        self._smoke_test(
+            log,
+            node,
+            log_path,
+            "verify_deployment_provision_ultra_datadisk",
         )
 
     @TestCaseMetadata(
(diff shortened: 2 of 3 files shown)


lisa/messages.py
```diff
@@ -144,6 +144,7 @@ class PerfMessage(MessageBase):
         "unknown",
         "nvme",
         "premiumssd",
+        "ultradisk",
```
kamalca, line 147: We have Enums for Disk Type in several other places. Can we resuse code so that less changes need to be made in the future?

microsoft/testsuites/core/provisioning.py
```diff
@@ -191,6 +193,31 @@ def verify_reboot_in_platform(
             case_name="verify_reboot_in_platform",
         )
 
+    @TestCaseMetadata(
+        description="""
+        This case runs smoke test on a node provisioned with ultra disk.
+        The test steps are same as `smoke_test`.
+        """,
+        priority=3,
```
kamalca: Does it make sense to have priority different than other provisioning tests?

microsoft/testsuites/core/provisioning.py
```diff
@@ -191,6 +193,31 @@ def verify_reboot_in_platform(
...
+        requirement=simple_requirement(
+            disk=schema.DiskOptionSettings(
+                data_disk_type=schema.DiskType.UltraSSDLRS,
+                os_disk_type=schema.DiskType.StandardHDDLRS,
```
kamalca: Do we need to place a requirement on `os_disk_type`? Maybe we should let it follow the default / let LISA decide.

microsoft/testsuites/core/provisioning.py
```diff
@@ -191,6 +193,31 @@ def verify_reboot_in_platform(
...
+                data_disk_iops=search_space.IntRange(min=160000),
```
kamalca: Do we need this high of an IOPS requirement? Do we need any IOPS requirement?

This will boot with 1TB disk.

kamalca (changes requested):

lisa/messages.py
SRIKKANTH: The other place where I found the disk types declared are "schema.DiskType". But these names are a bit different compared to ones we have here. The disk types used in messages.py are used in db table. Changing them to different ones might break storage perf powerbi reports if anyone is using the column. @LiliDeng please share your thoughts.

microsoft/testsuites/core/provisioning.py
SRIKKANTH: @johnsongeorge-w any thoughts on this?

microsoft/testsuites/core/provisioning.py
SRIKKANTH: 160,000 IOPS is the max supported by a single ultradisk. Lets use it.

squirrelsc: It's a provisioning test case, it doesn't need to be max value.

SRIKKANTH: What would be a good size to use then?

microsoft/testsuites/core/provisioning.py
SRIKKANTH: It was a copy paste error. Changed it to 1.

microsoft/testsuites/core/provisioning.py
squirrelsc: It doesn't need to set iops requirements on provisioning test case.

SRIKKANTH: Fixed it. Removed the IOPS requirement.

lisa/messages.py
SRIKKANTH: @johnsongeorge-w FYI.

microsoft/testsuites/core/provisioning.py
SRIKKANTH: Removed it.

lisa/messages.py
kamalca: Okay, I think it is okay to keep the current implementation if it is representing values in a different database.

kamalca (approved):

LiliDeng (approved):

PR #2973, opened 2023-09-24
xfstests: use correct exclude file format

LiliDeng:


microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -130,15 +130,20 @@ class Xfstesting(TestSuite):
     # VM will hung during running case xfs/520
     # commit d0c7feaf8767 ("xfs: add agf freeblocks verify in xfs_agf_verify")
     # TODO: will figure out the detailed reason of every excluded case.
-    EXCLUDED_TESTS = (
-        "generic/211 generic/430 generic/431 generic/434 /xfs/438 xfs/490"
+    excluded_tests = (
```
squirrelsc, line 133: Upper case should be ok, because we use it like a constant.

squirrelsc (approved):

microsoft/testsuites/xfstests/xfstesting.py
LiliDeng: https://github.com/microsoft/lisa/pull/2973/files#diff-82592cd45538925e263695b707663401ce51a014b3d9e43f2c406f60b28e2d8dR145 redefine it, so it can't be a constant

squirrelsc: It should use another variable, instead of modifying the original variable. Don't modify variables, which share among test  cases. For example, in combinator with multiple images, if oracle runs before other distros, the value is effective there. That's why it uses upper case, to let people know, don't modify it.

You can modify it in `_execute_xfstests`, like redhat.

LiliDeng: ok, merge too quick, will raise a new PR for it tomorrow.

squirrelsc: I haven't noticed the class variable was changed... If I noticed it, I won't approve this change.

PR #2974, opened 2023-09-25
Add read/write permission for "Microsoft.Compute/images"

lubaihua33:


squirrelsc:
@paxue Please be aware of it, when applying new permissions.

squirrelsc (approved):

PR #2977, opened 2023-09-25
Mark env dirty, if it runs deployed phase case. Fixed aitl script errors.

squirrelsc:
If a test case is specified to run on deployed phase, it won't run transformers on connected phase. So, mark the environment to dirty, to prevent side effects on other test cases.


squirrelsc:
@LiliDeng, I have a slight check, it looks no regression. It deploys more environments for each deployed test cases.

PR #2980, opened 2023-09-26
MANA: add driver survey test

mcgov:
mana_en drivers should be present in kernels >= 5.15. 
mana_ib drivers should be present in kernels >= 6.2. 
Add a test to survey images and check for mana drivers in images.

Test checks kernel config and modprobe to verify the driver is present as either a builtin or module. It checks kernel logs to verify the driver had loaded as expected.

### DIFF
--- a/microsoft/testsuites/core/mana.py
+++ b/microsoft/testsuites/core/mana.py
@@ -1,0 +1,164 @@
+import re
+
+from lisa import (
+    LisaException,
+    Node,
+    SkippedException,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+    simple_requirement,
+)
+from lisa.operating_system import Linux, Posix
+from lisa.sut_orchestrator import AZURE
+from lisa.tools import Dmesg, KernelConfig, Modprobe
+
+
+@TestSuiteMetadata(
+    area="mana_driver",
+    category="functional",
+    description="""
+    This test suite checks for the presence of mana and mana_ib drivers
+    """,
+)
+class ManaDriverCheck(TestSuite):
+    MANA_DRIVER_MESSAGE_PATTERN = re.compile(r"\[\s*[0-9.]+\s*\]\s+mana\s+.*")
+    MANA_IB_DRIVER_MESSAGE_PATTERN = re.compile(r"\[\s*[0-9.]+\s*\]\s+mana_ib\s+.*")
+
+    @TestCaseMetadata(
+        description="""
+        This test case checks for the mana ethernet driver.
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            supported_platform_type=[AZURE], supported_os=[Posix]
+        ),
+    )
+    def verify_mana_en_driver_present(self, node: Node) -> None:
+        if not node.nics.is_mana_device_present():
+            raise SkippedException(
+                "MANA driver tests should be run in a MANA VM environment"
+            )
+
+        distro = node.os
+        if isinstance(distro, Linux):
+            if distro.get_kernel_information().version < "5.15.0":
+                raise SkippedException("MANA is not available for kernels < 5.15")
+
+        mana_driver_name = "mana"
+        dmesg = node.tools[Dmesg]
+        modprobe = node.tools[Modprobe]
+        kconfig = node.tools[KernelConfig]
+
+        # Check kconfig for mana
+        mana_is_builtin = kconfig.is_built_in(mana_driver_name)
+        mana_is_module = kconfig.is_built_as_module(mana_driver_name)
+        if mana_is_builtin:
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/mana.py
```diff
@@ -0,0 +1,164 @@
...
+)
+class ManaDriverCheck(TestSuite):
+    MANA_DRIVER_MESSAGE_PATTERN = re.compile(r"\[\s*[0-9.]+\s*\]\s+mana\s+.*")
+    MANA_IB_DRIVER_MESSAGE_PATTERN = re.compile(r"\[\s*[0-9.]+\s*\]\s+mana_ib\s+.*")
+
+    @TestCaseMetadata(
+        description="""
+        This test case checks for the mana ethernet driver.
+        """,
+        priority=4,
```
squirrelsc: P4 means it's seldom chance to run. If it's not high cost tests like longhaul or stress tests, they don't need to be P4. And for the impact, what if it's failed? P4 means less impact.

microsoft/testsuites/core/mana.py
```diff
@@ -0,0 +1,164 @@
...
+        mana_is_builtin = kconfig.is_built_in(mana_driver_name)
+        mana_is_module = kconfig.is_built_as_module(mana_driver_name)
+        if mana_is_builtin:
+            node.log.info(f"{mana_driver_name} is reported as builtin in kconfig")
+        elif mana_is_module:
+            node.log.info(
+                f"{mana_driver_name} is reported as a loadable kernel module in kconfig"
+            )
+        else:
+            node.log.warn(
```
squirrelsc: Don't print warning level log, no one checks it. Either fail test case or use info level log.

microsoft/testsuites/core/mana.py
mcgov: You're right, it should be alright to run since it skips unless we're on MANA.

microsoft/testsuites/core/mana.py
mcgov: Works for me :) moved to Info

microsoft/testsuites/core/mana.py
```diff
@@ -0,0 +1,164 @@
...
+            if distro.get_kernel_information().version < "5.15.0":
+                raise SkippedException("MANA is not available for kernels < 5.15")
+
+        mana_driver_name = "mana"
+        dmesg = node.tools[Dmesg]
+        modprobe = node.tools[Modprobe]
+        kconfig = node.tools[KernelConfig]
+
+        # Check kconfig for mana
+        mana_is_builtin = kconfig.is_built_in(mana_driver_name)
```
LiliDeng: we can add a new method for the common code used by these two cases

microsoft/testsuites/core/mana.py
```diff
@@ -0,0 +1,164 @@
...
+@TestSuiteMetadata(
+    area="mana_driver",
+    category="functional",
+    description="""
+    This test suite checks for the presence of mana and mana_ib drivers
+    """,
+)
+class ManaDriverCheck(TestSuite):
+    MANA_DRIVER_MESSAGE_PATTERN = re.compile(r"\[\s*[0-9.]+\s*\]\s+mana\s+.*")
+    MANA_IB_DRIVER_MESSAGE_PATTERN = re.compile(r"\[\s*[0-9.]+\s*\]\s+mana_ib\s+.*")
```
LiliDeng, line 27: add the sample messages

microsoft/testsuites/core/mana.py
```diff
@@ -0,0 +1,164 @@
...
+
+        distro = node.os
+        if isinstance(distro, Linux):
+            if distro.get_kernel_information().version < "5.15.0":
+                raise SkippedException("MANA is not available for kernels < 5.15")
+
+        mana_driver_name = "mana"
+        dmesg = node.tools[Dmesg]
+        modprobe = node.tools[Modprobe]
+        kconfig = node.tools[KernelConfig]
```
LiliDeng, line 42: we can check config even there is no mana devices

microsoft/testsuites/core/mana.py
mcgov: combined, thanks!

microsoft/testsuites/core/mana.py
mcgov: updated

microsoft/testsuites/core/mana.py
mcgov: added a test just kconfig only

mcgov:
block merge on testing after refactor...

microsoft/testsuites/core/mana.py
LiliDeng: I don't see the updated with sample messages

PR #2981, opened 2023-09-26
MANA: enable rdma-core source installation

mcgov:
First commit:
- Enable a source build of rdma-core; useful for MANA testing where backported kernel patches are available but backported rdma-core and dpdk are not available in the package manager.
- small refactors: 
  - move pmd setup to it's own exposed function, used in upcoming commit for l3fwd
  - fix some variable names for clarity
  - allow sample_apps to be used in init_nodes_concurrent

Wget commit:
- add 5 retries to wget tool.

### DIFF
--- a/lisa/base_tools/wget.py
+++ b/lisa/base_tools/wget.py
@@ -2,6 +2,8 @@ import re
 from typing import TYPE_CHECKING, Optional, Tuple, Type
 from urllib.parse import urlparse
 
+from retry import retry
+
 from lisa.executable import Tool
 from lisa.tools.ls import Ls
 from lisa.tools.mkdir import Mkdir
@@ -31,6 +33,7 @@ class Wget(Tool):
         posix_os.install_packages([self])
         return self._check_exists()
 
+    @retry(LisaException, tries=5, delay=2, backoff=1.5)
     def get(
         self,
         url: str,
--- a/microsoft/testsuites/dpdk/dpdktestpmd.py
+++ b/microsoft/testsuites/dpdk/dpdktestpmd.py
@@ -2,7 +2,6 @@
 # Licensed under the MIT license.
 
 import re
-import time
 from pathlib import PurePosixPath
 from typing import Any, List, Tuple, Type, Union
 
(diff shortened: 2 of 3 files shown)


lisa/base_tools/wget.py
```diff
@@ -64,18 +72,26 @@ def get(
...
+                timeout=timeout,
+            )
+            download_file_path = self._get_downladed_file_path(command_result.stdout)
+            if download_file_path:
+                break
+            else:
+                retry -= 1
+                time.sleep(5)
+
+        if not download_file_path:
```
squirrelsc: You can add a retry decorator on this method.

mcgov: That was giving me a type checking warning, I just noticed it's used elsewhere so maybe it's alright? I'll switch.

lisa/base_tools/wget.py
```diff
@@ -31,6 +33,7 @@ def install(self) -> bool:
         posix_os.install_packages([self])
         return self._check_exists()
 
+    @retry(LisaException, tries=5, delay=2, backoff=1.5)
```
LiliDeng, line 36: it may waste time when encounter a real issue, like the url doesn't exist

mcgov: To address your worry or wasted time: It wastes much more developer time to skip/fail when there's throttling or another network hiccup. Our more complex runs can take an hour or more to set up if we're applying multiple update steps that involve restarts and status checks, it's ok to spend 7-45 seconds retrying if there is throttling or some other network hiccup. They do happen, I wouldn't have added this retry if I hadn't hit this problem previously.
For one test running once, it's less likely you'll hit a problem. It's different when you have thousands of tests running on an NxN matrix of host builds and hardware.

It would help me if you have an actionable suggestion, such as 'could we reduce the retries to 2 tries' or 'can we reduce the backoff to 1.25 instead of 1.5'

mcgov:
If possible, let's prefer merging the stacked PR that includes these commits:
https://github.com/microsoft/lisa/pull/3038

LiliDeng:
#3048 has been merged

PR #2983, opened 2023-09-27
Extra config to support Lunar proposed kernel installation.

SRIKKANTH:
Below steps are needed in order to install proposed kernels on Lunar. 

Create "/etc/apt/preferences.d/proposed.pref" with below config:
 
Package: *
Pin: release a=*-proposed
Pin-Priority: 500


LiliDeng (approved):

squirrelsc:
If it's a new feature from Lunar, the logic should bypass the new config on old versions. So, when new version arrives, it doesn't need to add one by one.

PR #2984, opened 2023-09-28
Add Git Bisect Tool

adityagesh:
Create git bisect tool from Git with
methods start, good, bad


lisa/tools/git.py
```diff
@@ -311,6 +296,13 @@ def get_current_branch(self, cwd: pathlib.PurePath) -> str:
         )
         return filter_ansi_escape(result.stdout)
 
+    def get_current_commit_hash(self, cwd: pathlib.PurePath) -> str:
```
squirrelsc, line 299: This looks not related with the PR/commit. But it's ok to include it.

squirrelsc (approved):

PR #2992, opened 2023-10-03
IP: add is_device_up

mcgov:
Add a function to check current device status using `ip -br link show`

### DIFF
--- a/lisa/tools/ip.py
+++ b/lisa/tools/ip.py
@@ -83,6 +83,12 @@ class Ip(Tool):
         r"\s+dev\s+"  # looking for the device for the default route
         r"([a-zA-Z0-9]+)"  # capture device
     )
+    # -br flag gives easily parsible output
+    # ex:
+    # eth0             UP             00:15:5d:ff:20:68 ...
+    __ip_br_show_regex = re.compile(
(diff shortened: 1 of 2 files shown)


lisa/tools/ip.py
```diff
@@ -83,6 +83,9 @@ class Ip(Tool):
         r"\s+dev\s+"  # looking for the device for the default route
         r"([a-zA-Z0-9]+)"  # capture device
     )
+    __ip_br_show_regex = re.compile(
```
squirrelsc, line 89: Move examples above the regexp.

lisa/tools/ip.py
```diff
@@ -127,6 +133,20 @@ def _get_matched_dict(self, result: str) -> Dict[str, str]:
             "ip_addr": matched.group("ip_addr"),
         }
 
+    def is_device_up(self, nic_name: str) -> bool:
+        device_info = self.node.execute(
+            f"ip -br link show {nic_name}",
```
LiliDeng, line 138: put it into ip tool

mcgov: Not sure what you mean, pull the regex into the tool?
or use self.run?

lisa/tools/ip.py
mcgov: moved

mcgov:
closing, testing this without the other rdma-core PR is getting too complicated

PR #2995, opened 2023-10-05
MSR: check rdmsr exists after installation

mcgov:
Covering two edge cases:
- msr-tools package is available but rdmsr is not in that package
- msr-tools is not available but rdmsr is available.

Fix is just to check if the binary is available and skip if it's not.

### DIFF
--- a/microsoft/testsuites/core/msr.py
+++ b/microsoft/testsuites/core/msr.py
@@ -184,6 +184,7 @@ class Msr(TestSuite):
         ).is_not_zero()
 
 
-# NOTE: further work: checking the kernel version matches checking for known manufacturer ids, etc.
-#       implementing this platform ID info is not required for use with hyper-v but is for
-#       hv guest extensions and azure platform health reporting.
+# NOTE: further work: checking the kernel version matches checking for known
+#       manufacturer ids, etc.
+#       implementing this platform ID info is not required for use with hyper-v
+#       but is for hv guest extensions and azure platform health reporting.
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/core/msr.py
```diff
@@ -116,16 +122,42 @@ def verify_hyperv_platform_id(self, node: RemoteNode) -> None:
...
+            arch_msr_offset = HvOsPlatformInfo.HV_REGISTER_GUEST_OSID[arch_id]
+        except KeyError as missing_key:
+            raise SkippedException(f"Arch {missing_key} is not supported by msr test")
+
+        # try installing msr-tools if rdmsr isn't already insalled.
+        if node.execute("command -v rdmsr", shell=True, sudo=True).exit_code != 0:
+            try:
+                distro.install_packages("msr-tools")
+            except AssertionError:
+                raise SkippedException(
```
LiliDeng, line 138: for redhat rhel-arm64 8_6-arm64 8.6.2022060901 image, case fails for below exception. it is not AssertionError type
```failed. MissingPackagesException: Package manager could not install packages: msr-tools```

mcgov: Nice, will fix. I guess this is a bug for the other distros? There doesn't seem to be a common failure mode for installing packages. Some return without installing, some throw an AssertionError, some throw a MissingPackagesException (I think this is the correct failure mode)

mcgov: fixed :)

LiliDeng (approved):

PR #2996, opened 2023-10-05
libvirt platform: improve logging

pupacha:
Added few debug logs to help investigate failures better.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -594,8 +594,10 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
 
         # Create libvirt domain (i.e. VM).
         xml = self._create_node_domain_xml(environment, log, node)
+        log.debug(f"Domain xml for {node_context.vm_name} - {xml}")
         node_context.domain = self.libvirt_conn.defineXML(xml)
 
+        log.debug(f"Creating libvirt domain - {node_context.vm_name}")
         self._create_domain_and_attach_logger(
             node_context,
         )
@@ -1103,10 +1105,14 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         while True:
             addr = self._try_get_node_ip_address(environment, log, node)
             if addr:
+                log.debug(f"VM {node_context.vm_name} booted with IP - {addr}")
                 return addr
 
             if time.time() > timeout:
-                raise LisaException(f"no IP addresses found for {node_context.vm_name}")
+                raise LisaException(
+                    f"no IP addresses found for {node_context.vm_name}."
+                    " Guest OS might have failed to boot"
+                )
 
     # Try to get the IP address of the VM.
     def _try_get_node_ip_address(


LiliDeng (dismissed):

squirrelsc:
@cwize1 can you help reviewing changes in libvirt?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1103,10 +1105,14 @@ def _get_node_ip_address(
...
             addr = self._try_get_node_ip_address(environment, log, node)
             if addr:
+                log.debug(f"Domain {node_context.vm_name} booted with IP - {addr}")
                 return addr
 
             if time.time() > timeout:
-                raise LisaException(f"no IP addresses found for {node_context.vm_name}")
+                raise LisaException(
+                    f"no IP addresses found for {node_context.vm_name}."
+                    " Domain creation might have failed"
```
cwize1: What do you mean by "domain creation"?

Typically, when I see this error, it is because the OS failed to boot.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -594,8 +594,10 @@ def _create_node(
 
         # Create libvirt domain (i.e. VM).
         xml = self._create_node_domain_xml(environment, log, node)
+        log.debug(f"Domain xml for {node_context.vm_name} - {xml}")
```
cwize1, line 597: This log line seems unnecessary. If the user ever needs to view this, they can just use `virsh dumpxml`.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -594,8 +594,10 @@ def _create_node(
...
         node_context.domain = self.libvirt_conn.defineXML(xml)
 
+        log.debug(f"Creating domain - {node_context.vm_name}")
```
cwize1: Only really libvirt uses the term "domain". I would say "libvirt domain" or "libvirt VM".

cwize1 (changes requested):

lisa/sut_orchestrator/libvirt/platform.py
pupacha: `Domain` is Libvirt's term to mean a VM/Guest. Since we are using Libvirt internally for VM creation, I added the message this way.

lisa/sut_orchestrator/libvirt/platform.py
pupacha: Right. But that is only when the Guest VM is running. It doesn't help much in cases when VM's are cleanup. But I felt this would be of no harm and is good to have.

lisa/sut_orchestrator/libvirt/platform.py
pupacha: As you suggested in other comment, "Libvirt domain" would be apt to use. Will change it.

lisa/sut_orchestrator/libvirt/platform.py
cwize1: Are you aware of the `keep-environment` flag?

cwize1 (approved):

LiliDeng (approved):

lisa/sut_orchestrator/libvirt/platform.py
pupacha: Yes I'm aware, but we cannot use it in our autmated pipelines because of resource constraints.

PR #3000, opened 2023-10-09
Bug Fix: Azure GPU Extension

adityagesh:
If GPU installation using Azure GPU Extension runs more than once on a Linux VM, the installation fails with error 'Microsoft.HpcCompute.NvidiaGpuDriverLinux' already added or specified in input.
In this case, SKIP the case instead of failing

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -547,13 +547,24 @@ class Gpu(AzureFeatureMixin, features.Gpu):
         if release not in supported_versions.get(type(self._node.os), []):
             raise UnsupportedOperationException("GPU Extension not supported")
         extension = self._node.features[AzureExtension]
-        result = extension.create_or_update(
-            type_="NvidiaGpuDriverLinux",
-            publisher="Microsoft.HpcCompute",
-            type_handler_version="1.6",
-            auto_upgrade_minor_version=True,
-            settings={},
-        )
+        try:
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -547,13 +547,23 @@ def _install_driver_using_platform_feature(self) -> None:
...
+                type_handler_version="1.6",
+                auto_upgrade_minor_version=True,
+                settings={},
+            )
+        except Exception as e:
+            if (
+                "'Microsoft.HpcCompute.NvidiaGpuDriverLinux' already added"
+                " or specified in input" in str(e)
+            ):
+                raise SkippedException("GPU Extension is already installed")
```
LiliDeng: @adityagesh thanks for fixing this. If throw skipexception here, what is the test result? should we just log one line and let test continue?

adityagesh: In this case, the test would be skipped.

The only test case that directly invokes the VM Extension installation is "verify_gpu_extension_installation"
The other test cases checks if the driver exists and then tries to install using extension, so this bug would never occur.

But, what you said makes sense, the Feature should not be raising a SKIPPED exception. I'll make the change as you suggested.

LiliDeng (approved):

squirrelsc (dismissed):
Don’t skip test out of test case level

PR #3002, opened 2023-10-09
Collect disk controller type as information 

SRIKKANTH:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -766,6 +769,11 @@ def _get_node_information(self, node: Node) -> Dict[str, str]:
         information["platform"] = self.type_name()
         return information
 
+    def _get_disk_controller_type(self, node: Node) -> str:
```
squirrelsc, line 784: Why is it needed? If the information is decided by deployment already, it doesn't need to be saved into information.

SRIKKANTH: So far there is no way we can distinguish test executions between SCSI & NVMe disc controllers. There are instances where we need to cover both SCSI and NVMe disc controller behavior. For example storage perf test numbers executed with NVMe and SCSI will be different.

squirrelsc: For perf test cases, it needs to update the `disk_type` in `DiskPerformanceMessage`. The dict information shouldn't be overused, because it's low performance to query from database.

SRIKKANTH: > Why is it needed? If the information is decided by deployment already, it doesn't need to be saved into information.

How can we know what disc controller type is used in a test from the data base? Can you provide any better idea where we can keep the disc controller type? 

> For perf test cases, it needs to update the `disk_type` in `DiskPerformanceMessage`. The dict information shouldn't be overused, because it's low performance to query from database.

This requirement is not just specific to  perf  tests. Even in perf tests with `disk_type` in `DiskPerformanceMessage` we can only mention disc types but not controller type. For example in case of ultra disk IO perf, allowed IOPS are different with SCSI and NVMe, where disc_type is ultradisk for both of them. We need a mechanism to distinguish these results based on the controller type. Please see: https://learn.microsoft.com/en-us/azure/virtual-machines/ebdsv5-ebsv5-series#ebdsv5-series-scsi 

Please let me know if there any better way to keep disc controller type, if you find information column is not best one.

squirrelsc: You can add a new field to save the controller type in DiskPerformanceMessage.

squirrelsc:
@LiliDeng LGTM

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -432,6 +434,7 @@ def __init__(self, runbook: schema.Platform) -> None:
         # It has to be defined after the class definition is loaded. So it
         # cannot be a class level variable.
         self._environment_information_hooks = {
+            KEY_HARDWARE_DISK_CONTROLLER_TYPE: self._get_disk_controller_type,
```
LiliDeng, line 438: Where do you use it?

LiliDeng:
it throws exception when run any cases + canonical ubuntuserver 16.04-lts 16.04.201702020

SRIKKANTH:
> it throws exception when run any cases + canonical ubuntuserver 16.04-lts 16.04.201702020

Its fixed.

LiliDeng (approved):

PR #3003, opened 2023-10-11
GPU Cases: Add support for Mariner 2

adityagesh:


lisa/features/gpu.py
```diff
@@ -186,6 +189,19 @@ def _install_grid_driver(self, driver_url: str) -> None:
...
+                match_lines="^GRUB_CMDLINE_LINUX",
+                regexp="lockdown=integrity",
+                replacement="lockdown=0",
+                file="/etc/default/grub",
+                sudo=True,
+            )
+            self._node.execute(
+                "grub2-mkconfig -o /boot/grub2/grub.cfg", shell=True, sudo=True
+            )
+            self._node.reboot()
```
adityagesh: @squirrelsc  @LiliDeng 

Is this approach fine? The GRID driver is unsigned and is failing to install because lockdown=integrity is set in Mariner 2 by default

LiliDeng: you can confirm with mariner team, what is the customers' scenario when customers want to install gpu driver.

adityagesh: > you can confirm with mariner team, what is the customers' scenario when customers want to install gpu driver.

Okay, I am able to install CUDA driver without this. Only having issue with GRID Driver.

lisa/features/gpu.py
```diff
@@ -275,6 +291,14 @@ def _install_cuda_driver(self, version: str) -> None:
                     self._node.os.install_packages("cuda-drivers-465")
                 else:
                     self._node.os.install_packages("cuda-drivers-515")
+        elif isinstance(self._node.os, CBLMariner):
+            # Replace "5.15.131.1-2.cm2" with "5.15.131.1.2.cm2"
+            kernel_ver = self._node.os.get_kernel_information().raw_version.replace(
+                "-", "."
+            )
+            self._node.os._install_package_from_url(
+                f"https://packages.microsoft.com/cbl-mariner/2.0/prod/nvidia/x86_64/Packages/c/cuda-525.85.12-3_{kernel_ver}.x86_64.rpm"
```
LiliDeng: does it support arm64? if yes, replace x86_64 into actual cpu type.

adityagesh: No, it only supports x86. I will add that condition as well.

lisa/features/gpu.py
squirrelsc: It's ok to remove sign in test environments.

adityagesh: I will Skip GRID Driver for Mariner, since it is not in the supported list.
https://learn.microsoft.com/en-us/azure/virtual-machines/linux/n-series-driver-setup#nvidia-grid-drivers

adityagesh:
Removed GRID driver installation for Mariner and now installs using CUDA @LiliDeng

LiliDeng (approved):

PR #3009, opened 2023-10-17
verify_timesync_ptp: Read chrony config as sudo

adityagesh:
Fix for "Failed. AssertionError: [Chrony config file should use the symlink /dev/ptp_hyperv.] Expected **<cat: /etc/chrony.conf: Permission denied**> to contain item <ptp_hyperv>, but did not"

### DIFF
--- a/microsoft/testsuites/core/timesync.py
+++ b/microsoft/testsuites/core/timesync.py
@@ -115,7 +115,9 @@ class TimeSync(TestSuite):
         #  instead of /dev/ptp0 or /dev/ptp1.
         for chrony_config in self.chrony_path:
             if node.shell.exists(PurePosixPath(chrony_config)):
-                chrony_results = cat.run(f"{chrony_config}")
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/timesync.py
```diff
@@ -115,7 +115,9 @@ def verify_timesync_ptp(self, node: Node) -> None:
         #  instead of /dev/ptp0 or /dev/ptp1.
         for chrony_config in self.chrony_path:
             if node.shell.exists(PurePosixPath(chrony_config)):
```
LiliDeng, line 118: @adityagesh thanks for the fix, please have one more fix for the loop, it should not assert for each file.

adityagesh: Updated, can you please check if this logic is fine?

LiliDeng (approved):

PR #3014, opened 2023-10-18
Add 'enabled' for testcase in tier.yml

adityagesh:
- Add 'enable' for testcase in tier.yml
- Add function for git bisect log

### DIFF
--- a/lisa/tools/git.py
+++ b/lisa/tools/git.py
@@ -381,8 +381,12 @@ class GitBisect(Git):
         result = self.run(f"bad {ref}", cwd=cwd, force_run=True)
         result.assert_exit_code(message=f"failed to run bisect bad {result.stdout}")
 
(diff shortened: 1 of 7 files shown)


microsoft/runbook/tiers/tdebug.yml
```diff
@@ -15,3 +15,4 @@ testcase:
     retry: $(retry)
     use_new_environment: $(use_new_environment)
     ignore_failure: $(ignore_failure)
+    enabled: $(enable_tests)
```
LiliDeng: the file name should be debug.yml?

adityagesh: Removed the change in that file. I thought it was part of the tier files.

I think it is named tdebug because of the source file structure? 

image

LiliDeng (approved):

PR #3016, opened 2023-10-19
fix hwclock issue

LiliDeng:


squirrelsc (dismissed):

lisa/tools/hwclock.py
```diff
@@ -30,24 +30,27 @@ def set_rtc_clock_to_system_time(self) -> None:
         cmd_result = self.run("--systohc", shell=True, sudo=True)
         cmd_result.assert_exit_code()
 
-    def set_datetime(self, time: datetime) -> None:
+    def set_datetime(self, time: str) -> None:
         self.run(
-            f"--set --date='{time}'",
+            f"--set --date={time}",
```
squirrelsc: A string is not friendly and reuse. You can create a new method called `set_to_next_year`

squirrelsc: If it's datetime formatting issue, use a library to get right format.

LiliDeng: updated

squirrelsc (approved):

PR #3018, opened 2023-10-20
Enable LTP test case on WSL

squirrelsc:
See details in commits.

### DIFF
--- a/lisa/tools/lsblk.py
+++ b/lisa/tools/lsblk.py
@@ -160,9 +160,16 @@ class Lsblk(Tool):
         lsblk_entries = json.loads(output)["blockdevices"]
 
         for lsblk_entry in lsblk_entries:
+            disk_mountpoint = lsblk_entry["mountpoint"]
+            if disk_mountpoint == "/mnt/wslg/distro":
+                # WSL mounts the system disk to /mnt/wslg/distro, and it's not
+                # the default returned by lsblk. WSLg distro mountpoint is not
+                # accessible, so replace it to "/"
+                disk_mountpoint = "/"
+
             disk_info = DiskInfo(
                 name=lsblk_entry["name"],
-                mountpoint=lsblk_entry["mountpoint"],
+                mountpoint=disk_mountpoint,
                 size=int(lsblk_entry["size"]),
                 dev_type=lsblk_entry["type"],
                 fstype=lsblk_entry["fstype"],
--- a/lisa/tools/lsblk.py
+++ b/lisa/tools/lsblk.py
@@ -1,13 +1,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import json
 import re
 from dataclasses import dataclass, field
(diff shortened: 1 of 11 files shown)


LiliDeng:
@squirrelsc please rebase and resolve the conflict

squirrelsc:
> @squirrelsc please rebase and resolve the conflict

Sure. I found shell doesn't supported WSL. I will add more commits.

squirrelsc:
@LiliDeng Ready to merge

lisa/tools/lsblk.py
```diff
@@ -155,57 +151,37 @@ def get_disks(self, force_run: bool = False) -> List[DiskInfo]:
         disks: List[DiskInfo] = []
 
         # parse output of lsblk
+        # -b print SIZE in bytes rather than in human readable format
+        # -J output in JSON format
+        # -o list of columns to output
         output = self.run(
-            "-b -P -o NAME,SIZE,TYPE,MOUNTPOINT,FSTYPE", sudo=True, force_run=force_run
+            "-b -J -o NAME,SIZE,TYPE,MOUNTPOINTS,FSTYPE", sudo=True, force_run=force_run
```
LiliDeng: MOUNTPOINTS doesn't work on linux. It is `MOUNTPOINT`

squirrelsc: I tested it with Ubuntu 22.04 WSL. Which distro doesn't work? The MOUNTPOINT returns just one mount point. In WSL, it doesn't return the root path `/` by default.

LiliDeng: I tested it on marketplace image `Canonical 0001-com-ubuntu-server-focal 20_04-lts latest`

squirrelsc: ok, I will change it back to MOUNTPOINT for compatibility. It should also work.

LiliDeng (approved):

lisa/tools/lsblk.py
```diff
@@ -155,57 +151,42 @@ def get_disks(self, force_run: bool = False) -> List[DiskInfo]:
         disks: List[DiskInfo] = []
 
         # parse output of lsblk
+        # -b print SIZE in bytes rather than in human readable format
+        # -J output in JSON format
+        # -o list of columns to output
         output = self.run(
-            "-b -P -o NAME,SIZE,TYPE,MOUNTPOINT,FSTYPE", sudo=True, force_run=force_run
+            "-b -J -o NAME,SIZE,TYPE,MOUNTPOINT,FSTYPE", sudo=True, force_run=force_run
```
LiliDeng, line 158: there is a regression for this change.
On some images like `openlogic centos 7_9-arm64 7.9.2023010302`, lsblk doesn't support -J parameter, will fix it tomorrow.

squirrelsc: Thank you, Lili! We can use -J by default, if it doesn't work, use previous table format.

PR #3019, opened 2023-10-21
Need to replace boot kernel if the installer type is "source"

lubaihua33:


lisa/transformers/kernel_installer.py
```diff
@@ -138,10 +138,14 @@ def _internal_run(self) -> Dict[str, Any]:
         self._log.info(f"installed kernel version: {installed_kernel_version}")
 
         # for ubuntu cvm kernel, there is no menuentry added into grub file
-        if hasattr(installer.runbook, "source"):
-            if installer.runbook.source != "linux-image-azure-fde":
-                posix = cast(Posix, node.os)
-                posix.replace_boot_kernel(installed_kernel_version)
+        if (
```
squirrelsc: hasattr is not a good check, use `isinstance` instead. Add comments in the code. Only if the comments applies to changes, it won't be put in the code.

lubaihua33: Sorry, Chi. I didn't get your point. This logic checks if it has "type" in the runbook. I'm not sure how `instance ` can achieve this function.

Have added comments.

squirrelsc: For example, the `source` field is in RepoInstallerSchema, so it can be like `isinstance(installer, RepoInstaller)`. The `type` should be always here, and it doesn't need to be checked.

lubaihua33: Done

LiliDeng (dismissed):

lisa/transformers/kernel_installer.py
```diff
@@ -138,10 +138,18 @@ def _internal_run(self) -> Dict[str, Any]:
         self._log.info(f"installed kernel version: {installed_kernel_version}")
 
         # for ubuntu cvm kernel, there is no menuentry added into grub file
-        if hasattr(installer.runbook, "source"):
-            if installer.runbook.source != "linux-image-azure-fde":
-                posix = cast(Posix, node.os)
-                posix.replace_boot_kernel(installed_kernel_version)
+        # when the installer's type is "source", it needs to add the menuentry into
+        # grub file. Otherwise, the node might not boot into the new kernel especially
+        # the new kernel is not the latest version.
```
LiliDeng: does it mean the installed kernel version is lower than current kernel version?

squirrelsc:
@LiliDeng LGTM

lisa/transformers/kernel_installer.py
lubaihua33: Yes. I have modified the comment. Your description is much clearer.

LiliDeng (dismissed):

LiliDeng:
@lubaihua33 please resolve the conflict

LiliDeng (approved):

PR #3029, opened 2023-11-01
Enable Availability Zones

kamalca:
Enabling availability zones will allow us to deploy Ultra Disks in all regions.

Done:
* Minimal changes to deploy with zones on LISA
* Avoid conflicts between availability sets and availability zones
* Capability checks for VM SKUs
* Handle Ultra Disk requirements / failures related to availability zones

Not Done:
* Others?
* Deploy to more than one zone (right now it picks one out of the available zones)


squirrelsc:
How does the availability zones work with availability set?

kamalca:
> How does the availability zones work with availability set?

They are mutually exclusive. You cannot use availability sets with availability zones.
With these minimal changes, you would have to remove any mention of availability sets in the runbook for availability zones to work. I am hoping to improve this so that one runbook can take both availability set and availability zone fields but select which one based on `use_availability_set` or some other flag.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1213,6 +1216,23 @@ def _create_deployment_parameters(
...
+            arm_parameters.use_availability_sets = (
+                self._azure_runbook.use_availability_sets
+            )
+            if (
+                node.capability.disk.data_disk_type == schema.DiskType.UltraSSDLRS
+                and not arm_parameters.use_availability_sets
+                and not arm_parameters.availability_zones
+            ):
+                arm_parameters.use_availability_sets = False
+                arm_parameters.availability_zones = node.capability.zones.items
```
kamalca: @squirrelsc node.capability.zones is not being populated from the node_space created in _resource_sku_to_capability(). Do you know what I am doing wrong?

squirrelsc: Please remove the cached files, so the cache will be refreshed immediately.

kamalca: I did that as well. I see that the node_space in _resource_sku_to_capability() is set correctly. Still node.capability.zones is not set

squirrelsc: Ok, I got understand it now. The current logic updates the settings in create_deployment_parameters, it's not designed to save back to the disk. It needs to write in `_resource_sku_to_capability` as you mentioned, so it could be saved to the cached file, and be used next time.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1715,6 +1737,9 @@ def _resource_sku_to_capability(  # noqa: C901
                         azure_raw_capabilities[
                             location_capability.name
                         ] = location_capability.value
+                node_space.zones = search_space.SetSpace(
```
kamalca: @squirrelsc This is where I set the zones. It is writing to cache fine. Still doesn't show up in the other location though.

Is there a capability merge function or something that I missed?

kamalca: "Standard_D4s_v4": {
            "location": "westus3",
            "vm_size": "Standard_D4s_v4",
            "capability": {
                "type": "requirement",
                "name": "westus3_Standard_D4s_v4",
              ...
                "zones": {
                    "is_allow_set": true,
                    "items": [
                        "1",
                        "3",
                        "2"
                    ]
                },

squirrelsc: I may not understand correctly. So if it writes to the cache, it should be able to loaded back to the vm size information. For other loctions, do you mean the locations, which you haven't remove the cache? You can delete the whole cache folder to make sure all locations are refreshed.

kamalca: Yes, I deleted all caches. The new cache is correct but it is not loaded into node.capability.zones

image

kamalca: Does it have to do with where this node space is coming from?

``` python
 for node_space in environment.runbook.nodes_requirement:
```

lisa/schema.py
```diff
@@ -831,6 +831,10 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
         default_factory=partial(search_space.IntRange, min=0),
         metadata=field_metadata(decoder=search_space.decode_count_space),
     )
+    zones: search_space.SetSpace[int] = field(
```
squirrelsc: The zones should be in AzureNodeSchema, because it's not like a general concept.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Yes, please try if it's in node_space. The node.capability may not be set at this moment.

lisa/schema.py
kamalca: If I move zones then how can I set the VM capability in `_resource_sku_to_capability` since that returns a `NodeSpace`

lisa/features/availability.py
```diff
@@ -0,0 +1,172 @@
...
+availability_type_priority: List[AvailabilityType] = [
+    AvailabilityType.NoRedundancy,
+    AvailabilityType.AvailabilitySet,
+    AvailabilityType.AvailabilityZone,
+]
+
+
+@dataclass_json()
+@dataclass
+class AvailabilityArmParameter:
```
squirrelsc: It's Azure specified, so it should be in Azure features file.

lisa/util/constants.py
```diff
@@ -106,6 +106,12 @@
 ENVIRONMENT_KEEP_NO = "no"
 ENVIRONMENT_KEEP_FAILED = "failed"
 
+AVAILABILITY_DEFAULT = "default"
```
squirrelsc, line 109: What's different with default and none?

lisa/node.py
```diff
@@ -1076,6 +1079,15 @@ def list(self) -> Iterable[Node]:
         for node in self._list:
             yield node
 
+    def list_features_by_type(
+        self, type_name: Type[T_SETTINGS]
+    ) -> Iterable[T_SETTINGS]:
+        for node in self._list:
+            if node.capability.features:
```
squirrelsc: 1. Why this method is needed? The node.features should be enough.
1. One node should have only one feature with the same name. If it needs to present multiple information, it should be in the feature itself. So if a method get type_name as an argument, it should return only one feature.

lisa/util/constants.py
```diff
@@ -106,6 +106,12 @@
...
+AVAILABILITY_NONE = "none"
+AVAILABILITY_SET = "availability_set"
+AVAILABILITY_ZONE = "availability_zone"
+AVAILABILITY_VMSS = "vmss"
```
squirrelsc: It looks not used.

lisa/util/constants.py
kamalca: Default will enable availability sets if tip session is provided, or enable availability zones if zones are provided. None will not use any redundancy option regardless of whether tip session info or availability zones are provided.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2009,8 +2016,11 @@ def create_setting(
 
     @classmethod
     def _enable_hibernation(cls, *args: Any, **kwargs: Any) -> None:
-        parameters: Any = kwargs.get("arm_parameters")
-        if parameters.use_availability_sets:
+        parameters = cast(AzureArmParameter, kwargs.get("arm_parameters"))
+        if (
+            parameters.availability_options.availability_type
```
kamalca, line 2022: @squirrelsc this logic is running in Hibernation.on_before_deployment. Are there any guarantees about what order the Feature.on_before_deployment will be run? If Hibernation runs before Availability, the arm parameters might not be accurate for this check.

lisa/features/availability.py
kamalca: Isn't the whole concept of availability zones and availability sets specific to Azure? I imagine this feature won't be enabled in any other platform, so what is the purpose of separating logic into different files?

lisa/node.py
kamalca: environment.nodes does not have features, it only has capability.features so I couldn't make use of __getitem__ in Features or get_feature_settings.

This function returns multiple type settings because it is implemented in Nodes. It gets the feature settings for all the nodes.

lisa/schema.py
squirrelsc: Try `node_space.get_extended_runbook(AzureNodeSchema, type_name=AZURE)` to get extended node space.

lisa/features/availability.py
squirrelsc: Availability may not be a general name, but there is a common concept to run service with high availability. For example, when servers are upgrading for a service, the servers can be upgraded by availability sets. It also works for faults. If the servers are in two sets, the service is continuing. The zones may be redundant for electric incidences. Check details here, https://techcommunity.microsoft.com/t5/itops-talk-blog/understanding-availability-sets-and-availability-zones/ba-p/1992518

lisa/node.py
squirrelsc: I see. It looks the availability feature is for an environment, not a node. So, it's confusing. For simple, we can assume every node has the same feature, and only pick it from first node. In future, if it's too complex, we can consider supporting features on the environment level.

lisa/sut_orchestrator/azure/features.py
squirrelsc: You can change it to get availability set information from `environment`, instead of the `arm_parameters`.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1724,11 +1731,16 @@ def _resource_sku_to_capability(  # noqa: C901
         azure_raw_capabilities: Dict[str, str] = {}
         if resource_sku.location_info:
             for location_info in resource_sku.location_info:
+                azure_raw_capabilities["availability_zones"] = location_info.zones
```
squirrelsc, line 1769: The logic going to more complex. Can you use a simple json example to explain the structure in comment?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1724,11 +1731,16 @@ def _resource_sku_to_capability(  # noqa: C901
...
                 for zone_details in location_info.zone_details:
                     for location_capability in zone_details.capabilities:
                         azure_raw_capabilities[
                             location_capability.name
                         ] = location_capability.value
+                        if zone_details.additional_properties["Name"]:
```
squirrelsc, line 1776: Does it override previous information from location_info.zones? Please add code comments to explain it.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2120,7 +2130,8 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
 
         assert len(environment.nodes._list) == len(arm_parameters.nodes)
         for node, node_parameters in zip(environment.nodes._list, arm_parameters.nodes):
-            assert node.capability.features
+            if node.capability.features is None:
```
squirrelsc: When is it None?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2151,6 +2162,135 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
...
+        )
+        if node_availability_capabilities and not is_maximize_capability:
+            # Get min capability of all the nodes
+            cap_intersection = node_availability_capabilities[0]
+            for cap in node_availability_capabilities:
+                cap_intersection.intersect(cap)
+
+            assert isinstance(cap_intersection.availability_type, search_space.SetSpace)
+
+            if is_ultra_disk:
```
squirrelsc, line 2258: Add comments to explain the logic.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2151,6 +2162,135 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
...
+
+        return availability_settings
+
+    @classmethod
+    def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
+        environment = cast(Environment, kwargs.get("environment"))
+        arm_parameters = cast(AzureArmParameter, kwargs.get("arm_parameters"))
+        params = arm_parameters.availability_options
+
+        is_ultra_disk = any(
```
squirrelsc: This line can be moved to close when it's used. So the logic is easier to read together.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2151,6 +2162,135 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
...
+            [
+                node.data_disk_type == _disk_type_mapping[schema.DiskType.UltraSSDLRS]
```
squirrelsc: It's better to apply checks on node or node space, instead of arm_parameters. Arm_parameters is more like an output.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2151,6 +2162,135 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
...
+                for node in arm_parameters.nodes
+            ]
+        )
+        is_maximize_capability = any(
+            [node.maximize_capability for node in arm_parameters.nodes]
+        )
+
+        node_availability_capabilities: List[AvailabilitySettings] = list(
+            environment.nodes.list_features_by_type(AvailabilitySettings)
```
squirrelsc: Use the feature from first node, it should be good enough now.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2151,6 +2162,135 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
...
+            ), (
+                f"Availability Type "
+                f"'{params.availability_type}' "
+                "is not supported in the current configuration. Please select one of "
+                f"{[type.value for type in cap_intersection.availability_type.items]}. "
+                "Or consider changing the disk type or location."
+            )
+
+        if params.availability_type == AvailabilityType.AvailabilitySet:
+            params.availability_zones.clear()
```
squirrelsc, line 2304: Add comments like "availability set conflicts with availability zones, and the availability set is higher priority than zones"...

lisa/features/availability.py
squirrelsc: We can also treat it as an Azure feature, and put it in azure/features like AzureExtensions.

lisa/sut_orchestrator/azure/platform_.py
kamalca: Adding the example:
``` JSON
"locationInfo": [
    {
    "location": "southcentralus",
    "zoneDetails": [
        {
        "Name": [
            "3"
        ],
        "capabilities": [
            {
            "name": "UltraSSDAvailable",
            "value": "True"
            }
        ],
        "name": null
        }
    ],
    "zones": [
        "2",
        "3",
        "1"
    ]
    }
],
```

lisa/sut_orchestrator/azure/features.py
kamalca: Is it safe to assume that all the nodes in an envirnment will be the same setup? For example, data_disk_type? It could be the case that one node support availability zones, but not all, right?

lisa/sut_orchestrator/azure/features.py
kamalca: With the current implementation, this check in Hibernation is dependent on the code run in Availability.on_befroe_deployment. Checking the environment would also not be accurate.

Availability was implemented this way because it has a dependency on data_disk_type / Ultra Disk usage.

lisa/sut_orchestrator/azure/features.py
kamalca: Environment.nodes was empty. What other variable contains maximize_capability?

lisa/sut_orchestrator/azure/features.py
squirrelsc: Yes. We can assume all nodes are the same configurations for now. I tried to support different schemas in the same env, but no scenario and high cost.

lisa/features/availability.py
kamalca: I moved AvailabilityArmParameter to common because I couldn't find a way to import it in AzureArmParameter without a circular dependency. Common is also where it is primarily used, so I think this is an acceptable location.

lisa/sut_orchestrator/azure/features.py
kamalca: I removed this unnecessary change.

lisa/features/availability.py
squirrelsc: It's ok.

lisa/sut_orchestrator/azure/features.py
squirrelsc: try `environment.runbook.nodes_requirement`

lisa/sut_orchestrator/azure/features.py
squirrelsc: This may raise exception during deployment. If you find it, it can be skipped in hooks.py.

lisa/sut_orchestrator/azure/features.py
kamalca: What is the difference between
`environment.runbook.nodes_requirement[0].disk` and
`environment.capability.nodes[0].disk`?
Should I use one vs. the other?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1267,14 +1274,16 @@ def _create_deployment_parameters(
...
-        for f in features_settings.values():
-            feature_type = next(
-                x for x in self.supported_features() if x.name() == f.type
-            )
+        for feature_type, setting in [
+            (t, s)
+            for t in self.supported_features()
+            for s in features_settings.values()
+            if t.name() == s.type
+        ]:
```
kamalca, line 1286: Is this an acceptable solution? It will guarantee the order of on_before_deployment will match the order of the list returned by AzurePlatform.supported_features()
Do you foresee this change causing any problems? Is it a bad idea to rely on a guaranteed order of execution?

This felt like a better solution to me than trying to catch specific deployment errors in hooks.py

lisa/sut_orchestrator/azure/features.py
kamalca: Please see my comment below
> Is this an acceptable solution? It will guarantee the order of on_before_deployment will match the order of the list returned by AzurePlatform.supported_features() Do you foresee this change causing any problems? Is it a bad idea to rely on a guaranteed order of execution?
> 
> This felt like a better solution to me than trying to catch specific deployment errors in hooks.py

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: Yes, please also leave comments here and `supported_features` to specify the order is respected, and write comments explicitly explain the hibernation depends on Availability in the feature list.

lisa/sut_orchestrator/azure/features.py
squirrelsc: environment.runbook.nodes_requirement is almost here in Azure, but nodes are created later.

squirrelsc: Sorry, I mixed it with places. In `on_before_deployment`, the nodes should be ready to use now. So use Node is better than node_requirements, in case it breaks in a combined ready/azure environment. The combined env is supported, but not used in real scenarios.

squirrelsc (approved):

squirrelsc:
@LiliDeng LGTM

PR #3032, opened 2023-11-03
Add retry_with_excluded_exceptions

LiliDeng:


lisa/util/__init__.py
```diff
@@ -750,3 +753,25 @@ def check_till_timeout(
         sleep(interval)
     if timer.elapsed() >= timeout:
         raise LisaException(f"timeout: {timeout_message}")
+
+
+def retry_with_excluded_exceptions(
+    tries: int,
+    delay: float,
+    exceptions_to_skip: Tuple[Type[Exception], ...] = (),
```
squirrelsc: Use List instead of Tuple. The values should use `[ ]`, not `( )`.

lisa/util/__init__.py
```diff
@@ -750,3 +753,25 @@ def check_till_timeout(
         sleep(interval)
     if timer.elapsed() >= timeout:
         raise LisaException(f"timeout: {timeout_message}")
+
+
+def retry_with_excluded_exceptions(
```
squirrelsc: The "excluded" and "skip" in the parameters should be consistent. The name could be like `retry_without_exceptions`.

lisa/util/__init__.py
```diff
@@ -750,3 +753,25 @@ def check_till_timeout(
...
+    tries: int,
+    delay: float,
+    exceptions_to_skip: Tuple[Type[Exception], ...] = (),
+) -> Callable[[Callable[..., T]], Callable[..., T]]:
+    def decorator(func: Callable[..., T]) -> Callable[..., T]:
```
squirrelsc, line 766: It can call the retry package internally. If it's a skipped exception, raise it, and retry for others.

lisa/util/__init__.py
squirrelsc: name it like `skipped_exception`, so just one underscore.

lisa/util/__init__.py

lisa/util/__init__.py
LiliDeng: updated

lisa/util/__init__.py
```diff
@@ -750,3 +757,37 @@ def check_till_timeout(
...
+    delay: float = 0,
+    max_delay: Optional[float] = None,
+    backoff: float = 1,
+    jitter: Union[Tuple[float, float], float] = 0,
+    logger: Optional["Logger"] = None,
+) -> Callable[[Callable[..., T]], Callable[..., T]]:
+    def decorator(func: Callable[..., T]) -> Callable[..., T]:
+        @wraps(func)
+        @retry(
+            exceptions=RetryException,
```
squirrelsc, line 774: Should it be `Exception` directly?

lisa/util/__init__.py
```diff
@@ -750,3 +757,37 @@ def check_till_timeout(
...
+            logger=logger,
+        )
+        def wrapper(*args: object, **kwargs: object) -> T:
+            try:
+                return func(*args, **kwargs)
+            except Exception as e:
+                if any(isinstance(e, ex_type) for ex_type in skipped_exceptions):
+                    raise
+                else:
+                    raise RetryException("Retry for non-skipped exceptions") from e
```
squirrelsc: It looks we shouldn't add extra message in the exception, it shouldn't be printed in the exception.

LiliDeng: yes, it is unexpected to throw the RetryException, then this design doesn't work, I always see RetryException thrown out with / without the extra message, if don't use RetryException, it always retry whenever throw exception.

squirrelsc: I mean the added message "Retry for non-skipped exceptions" is not necessary. The original Exception message is enough.

LiliDeng: I know, during testing, the final thrown out exception type is RetryException which is unexpected
like `FAILED   failed. RetryException: `

squirrelsc: Ok, if so, let's go back to original implementation without reuse the `retry` module.

lisa/util/__init__.py
```diff
@@ -750,3 +752,49 @@ def check_till_timeout(
...
+                        sleep(current_delay)
+                        # Apply jitter if it's specified as a tuple (min, max)
+                        if isinstance(jitter, tuple):
+                            current_delay += random.uniform(jitter[0], jitter[1])
+                        elif jitter > 0:
+                            current_delay += random.uniform(0, jitter)
+                        current_delay *= backoff
+
+                        if logger:
+                            logger.warning(
```
squirrelsc: logger to info level.

squirrelsc (approved):

PR #3037, opened 2023-11-08
Fix gallery has no attribute 'name' issue

lubaihua33:
When new creates a gallery, the return value of "check_or_create_gallery" is a dict. "gallery.name" has error "AttributeError: 'dict' object has no attribute 'name'" .

### DIFF
--- a/lisa/sut_orchestrator/azure/transformers.py
+++ b/lisa/sut_orchestrator/azure/transformers.py
@@ -533,7 +533,7 @@ class SharedGalleryImageTransformer(Transformer):
             runbook.gallery_location,
             self._log,
         )
-        gallery = check_or_create_gallery(
+        check_or_create_gallery(
             platform,
             runbook.gallery_resource_group_name,
             runbook.gallery_name,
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -573,7 +573,11 @@ def _internal_run(self) -> Dict[str, Any]:
             runbook.gallery_image_location,
         )
 
-        sig_url = f"{gallery.name}/{runbook.gallery_image_name}/{gallery_image_version}"
+        sig_url = (
+            f"{runbook.gallery_name}/"
```
LiliDeng, line 577: is it mandatory? if no, it may be an empty value.

lubaihua33: If there is no `gallery_name `configured in the runbook, `SigTransformerSchema `has a default value "`lisa_default_gallery`". So it won't be empty.

LiliDeng (approved):

PR #3038, opened 2023-11-08
DPDK L3Fwd test (stacked PR)

mcgov:
This PR adds the l3fwd test to dpdkutil.py, dpdksuite.py, and dpdkperf.py

It is a stacked PR, building on the content from:
[Routing tables: az route and linux kernel routing with ip route by mcgov · Pull Request #3020 · microsoft/lisa (github.com)](https://github.com/microsoft/lisa/pull/3020)
[MANA: enable rdma-core source installation by mcgov · Pull Request #2981 · microsoft/lisa (github.com)](https://github.com/microsoft/lisa/pull/2981)

### DIFF
--- a/lisa/util/constants.py
+++ b/lisa/util/constants.py
@@ -163,7 +163,3 @@ LISA_TEST_FOR_SUDO = "lisa test for sudo"
 SIGINT = 2
 SIGTERM = 15
 SIGKILL = 9
-
-# azure routing table magic subnet prefix
-# signals 'route all traffic on this subnet'
-AZ_ROUTE_ALL_TRAFFIC = "0.0.0.0/0"
--- a/lisa/util/constants.py
+++ b/lisa/util/constants.py
@@ -159,6 +159,11 @@ NETWORK_PERFORMANCE_TOOL_DPDK_TESTPMD = "dpdk-testpmd"
 # Test for command with sudo
 LISA_TEST_FOR_SUDO = "lisa test for sudo"
 
+# linux signals
 SIGINT = 2
 SIGTERM = 15
 SIGKILL = 9
+
+# azure routing table magic subnet prefix
+# signals 'route all traffic on this subnet'
+AZ_ROUTE_ALL_TRAFFIC = "0.0.0.0/0"
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -19,7 +19,7 @@ from lisa import (
     search_space,
 )
 from lisa.features import Gpu, Infiniband, IsolatedResource, Sriov
-from lisa.operating_system import BSD, CBLMariner, Windows
+from lisa.operating_system import BSD, CBLMariner, Ubuntu, Windows
 from lisa.testsuite import simple_requirement
 from lisa.tools import Echo, Git, Ip, Kill, Lsmod, Make, Modprobe
 from lisa.util.constants import SIGINT
@@ -709,10 +709,12 @@ class Dpdk(TestSuite):
         ),
         priority=3,
         requirement=simple_requirement(
+            supported_os=[Ubuntu],
             min_core_count=8,
             min_count=3,
             min_nic_count=3,
             network_interface=Sriov(),
+            unsupported_features=[Gpu, Infiniband],
         ),
     )
     def verify_dpdk_l3fwd_ntttcp_tcp(
@@ -734,6 +736,7 @@ class Dpdk(TestSuite):
             """,
         priority=3,
         requirement=simple_requirement(
+            supported_os=[Ubuntu],
             min_core_count=8,
             min_count=3,
             min_nic_count=3,
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -698,10 +698,16 @@ class Dpdk(TestSuite):
     @TestCaseMetadata(
         description=(
             """
-                Run the L3 forwarding test for DPDK
+                Run the L3 forwarding test for DPDK.
+                This test creates a DPDK port forwarding setup between
+                two NICs on the same VM. It forwards packets from a sender on
+                subnet_a to a receiver on subnet_b. Without l3fwd,
+                packets will not be able to jump the subnets.  This imitates
+                a network virtual appliance setup, firewall, or other data plane
+                tool for managing network traffic with DPDK.
         """
         ),
-        priority=4,
+        priority=3,
         requirement=simple_requirement(
             min_core_count=8,
             min_count=3,
@@ -718,9 +724,15 @@ class Dpdk(TestSuite):
 
     @TestCaseMetadata(
         description="""
-          Run the L3 forwarding test for DPDK
-        """,
-        priority=4,
+                Run the l3fwd test using GiB hugepages.
+                This test creates a DPDK port forwarding setup between
+                two NICs on the same VM. It forwards packets from a sender on
+                subnet_a to a receiver on subnet_b. Without l3fwd,
+                packets will not be able to jump the subnets. This imitates
+                a network virtual appliance setup, firewall, or other data plane
+                tool for managing network traffic with DPDK.
+            """,
+        priority=3,
         requirement=simple_requirement(
             min_core_count=8,
             min_count=3,
--- a/microsoft/testsuites/dpdk/dpdksuite.py
+++ b/microsoft/testsuites/dpdk/dpdksuite.py
@@ -23,7 +23,10 @@ from lisa.operating_system import BSD, CBLMariner, Ubuntu, Windows
 from lisa.testsuite import simple_requirement
 from lisa.tools import Echo, Git, Ip, Kill, Lsmod, Make, Modprobe
 from lisa.util.constants import SIGINT
-from microsoft.testsuites.dpdk.common import DPDK_STABLE_GIT_REPO
+from microsoft.testsuites.dpdk.common import (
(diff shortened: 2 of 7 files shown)


lisa/nic.py
```diff
@@ -227,6 +227,9 @@ def get_secondary_nic(self) -> NicInfo:
         # will assert if none is present.
         return self.get_nic_by_index(1)
 
+    def get_tertiary_nic(self) -> NicInfo:
```
squirrelsc: It doesn't need to be a method. The logic is simple, and it's used only in dpdk..

mcgov: refactored to make this cleaner in general, labeling the resources by subnet.
https://github.com/microsoft/lisa/blob/723286d80ae540934f7a7518a6a0dcb4cda9dc4b/microsoft/testsuites/dpdk/dpdkutil.py#L767

lisa/util/constants.py
```diff
@@ -159,6 +159,11 @@
...
 LISA_TEST_FOR_SUDO = "lisa test for sudo"
 
+# linux signals
 SIGINT = 2
 SIGTERM = 15
 SIGKILL = 9
+
+# azure routing table magic subnet prefix
+# signals 'route all traffic on this subnet'
+AZ_ROUTE_ALL_TRAFFIC = "0.0.0.0/0"
```
squirrelsc: It's not a platform level consts. Put it in dpdk.

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -693,6 +695,49 @@ def verify_dpdk_send_receive_gb_hugepages_netvsc(
         except UnsupportedPackageVersionException as err:
             raise SkippedException(err)
 
+    @TestCaseMetadata(
+        description=(
+            """
+                Run the L3 forwarding test for DPDK
+        """
+        ),
+        priority=4,
```
squirrelsc: Why set it to P4?

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -693,6 +695,49 @@ def verify_dpdk_send_receive_gb_hugepages_netvsc(
         except UnsupportedPackageVersionException as err:
             raise SkippedException(err)
 
+    @TestCaseMetadata(
+        description=(
+            """
+                Run the L3 forwarding test for DPDK
```
squirrelsc: Describe more details, on the test logic and test purpose.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -610,3 +619,455 @@ def _parallel_cleanup(node: Node) -> None:
...
+    l3fwd_app_name = "dpdk-l3fwd"
+    send_side = 1
+    receive_side = 2
+    # arbitrarily pick fwd/snd/recv nodes.
+    forwarder, sender, receiver = environment.nodes.list()
+    if (
+        not isinstance(forwarder.os, Ubuntu)
+        or forwarder.os.information.version < "22.4.0"
+    ):
+        raise SkippedException("l3fwd test not compatible, use Ubuntu >= 22.04")
```
LiliDeng, line 767: skip the case in the beginning to save time

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -610,3 +619,455 @@ def _parallel_cleanup(node: Node) -> None:
...
+    )
+    receiver.features[NetworkInterface].create_route_table(
+        nic_name=subnet_b_nics[forwarder].name,
+        route_name="fwd-tx",
+        subnet_mask=ipv4_to_lpm(subnet_b_nics[receiver].ip_addr),
+        em_first_hop=AZ_ROUTE_ALL_TRAFFIC,
+        next_hop_type="VirtualAppliance",
+        dest_hop=subnet_b_nics[forwarder].ip_addr,
+    )
+
```
LiliDeng, line 853: can we use loop for above code?

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -610,3 +619,455 @@ def _parallel_cleanup(node: Node) -> None:
...
+    try:
+        fwd_proc.wait_output("L3FWD: entering main loop", timeout=30)
+    except LisaException:
+        raise LisaException(
+            "L3fwd did not start. Check command output for incorrect flags, "
+            "core dumps, or other setup/init issues."
+        )
+
+    # start ntttcp client and server
+    ntttcp_threads_count = 64
```
LiliDeng, line 973: is it expected that fwd_proc running in the background during run ntttcp? if so, I saw that ntttcp starts to run after fwd_proc finishes running

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: My logic was:

- It's a test that is checking for packet forwarding performance regressions for network virtual appliances. 
- It requires lots of resources and will take a long time to run. 
- It will run on a very narrow set of images and doesn't need to be picked up in every pipeline.

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: This is the beginning

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: 1. What's the impact, if the test case failed? P4 means it's rare case to happen, and if it happens, it's little harmful.
2. How long does it take? From the test case name, it should be a validation test case. If a few packages are forwarded successfully, it should pass. It doesn't need long time to run. If it's a perf or stress test case, I agree it may need longer time.
3. If some images doesn't support this test case, it should be skipped, instead of use the priority to control it.

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: I don't get what the benefit would be. Each node is calling the function with arguments from the other nodes and different subnets so we would just be removing the argument labels in the function call.

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: How so? Wait_output doesn't kill the process, it just checks stdout. The flow is:
```
fwd.start_async() # no timeout
fwd.wait_output('started', timeout=30)  # blocks, throw if timeout
server.start_async(timeout=30) 
client.start(timeout=10) # blocks
server.wait_result()
fwd.kill()
```

lisa/util/constants.py
mcgov: 😆 I put it in here because I pictured you saying "this is an Azure constant not a DPDK one"

mcgov: I'll move it though

squirrelsc: Either it's Azure or DPDK one, it shouldn't be in the base LISA constants..

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: These are good points.

1. I see what you mean now, I was worrying about pipelines wasting time deploying 3 VMs just to skip the test. It's an important test if it fails and catches a bug. I'll pull to 3.
2. The setup takes a while. You have to build DPDK each time to run it, l3fwd is a sample app and I don't think it's normally part of the DPDK package. This is a tricky test case. It is both a validation case (we check for 0 or no throughput) and a perf case (we're using ntttcp, checking the perf numbers, and creating a perf message). It might make more sense to put it in the perf suite?
3. The test does skip unsupported images at runtime. I don't want to maintain this case on every target we normally support for the DPDK tests, it's too complex. We will add support for more kernels later if we have an ask. Is there a way to get LISA to skip deploying the environment if the image isn't Ubuntu 22.04 or above? I thought we didn't have a way to enforce that before runtime.

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: This might be confusing because l3fwd doesn't log much while it's running? You can verify the process is up and killed after the other two exit though. Packets won't flow between the subnets without l3fwd running.

microsoft/testsuites/dpdk/dpdksuite.py
squirrelsc: 1. P3 looks good to me.
2. For installation, make sure they are running in parallel, so install 3 nodes use similar time as 2 nodes.
3. It can be split to two cases, one is for functionality validation, another is for perf.
4. Distro detection cannot happen before a real deployment, but it should be ok to spend some time to deploy but not used. It's not too long in the whole progress.

mcgov: moved to pri3 and added to description

lisa/util/constants.py
mcgov: Moved to dpdk/common

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: I can't speed the DPDK build up, we are only building DPDK on one node: the forwarder.
Now that you mention it there might be some build optimizations I could apply to all the DPDK installations? I will look into that in another PR.

microsoft/testsuites/dpdk/dpdksuite.py

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: @squirrelsc split perf and functional tests and removed perf message post in functional test.

LiliDeng:
#3048 has been merged

PR #3040, opened 2023-11-09
Fix junit errors on subtests result, improving xfs reliability and refactoring on subtests

squirrelsc:
See commits for details


LiliDeng:
I have tested xfstests, total run time reduces from 4 hours into 2.5 hours. will test the junit later.

LiliDeng (approved):

PR #3041, opened 2023-11-10
DiskEncryptionExtension: Mark node dirty after case

adityagesh:
Disk Encryption case may cause node to be in bad state some time after the case completes. Hence mark the node as dirty to avoid causing issues for other cases

### DIFF
--- a/microsoft/testsuites/vm_extensions/azure_disk_encryption.py
+++ b/microsoft/testsuites/vm_extensions/azure_disk_encryption.py
@@ -56,6 +56,14 @@ class AzureDiskEncryption(TestSuite):
             if node.os.is_package_in_repo(package):
                 node.os.install_packages(package)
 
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/azure_disk_encryption.py
```diff
@@ -56,6 +56,14 @@ def before_case(self, log: Logger, **kwargs: Any) -> None:
             if node.os.is_package_in_repo(package):
                 node.os.install_packages(package)
 
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
+        # Disk Encryption may cause node to be in bad state
+        # some time after the test case is done. This can cause
+        # subsequent test cases to fai when environment is reused.
```
LiliDeng: fai => fail

adityagesh: updated

PR #3042, opened 2023-11-13
Kernel RPMInstaller: Make input path as local path to Node

pupacha:
Made input file path as local path to Node. Also, added force, nodeps options while installing the RPM.

### DIFF
--- a/lisa/base_tools/rpm.py
+++ b/lisa/base_tools/rpm.py
@@ -20,6 +20,12 @@ class Rpm(Tool):
         posix_os.install_packages(package_name)
         return self._check_exists()
 
+    def is_valid_package(self, file: str) -> bool:
+        cmd_result = self.run(
+            f'-qp "{file}"',
+        )
+        return cmd_result.exit_code == 0
+
     def get_file_size(self, file: str) -> int:
         cmd_result = self.run(
             "--queryformat='%{SIZE}' " f"-qp {file}",
@@ -29,9 +35,16 @@ class Rpm(Tool):
         )
         return int(cmd_result.stdout)
 
-    def install_local_package(self, file: str) -> None:
+    def install_local_package(
+        self, file: str, force: bool = False, nodeps: bool = False
+    ) -> None:
+        parameters = f'-ivh "{file}"'
+        if force:
+            parameters += " --force"
+        if nodeps:
+            parameters += " --nodeps"
         self.run(
-            f"-ivh {file}",
+            parameters,
             sudo=True,
             expected_exit_code=0,
             expected_exit_code_failure_message=(f"failed to install {file}"),
--- a/lisa/transformers/rpm_kernel_installer.py
+++ b/lisa/transformers/rpm_kernel_installer.py
@@ -1,6 +1,5 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-import os
 from dataclasses import dataclass, field
 from pathlib import PurePath
 from typing import Type
@@ -18,7 +17,7 @@ from .kernel_installer import BaseInstaller, BaseInstallerSchema
 @dataclass_json()
 @dataclass
 class RPMInstallerSchema(BaseInstallerSchema):
-    # kernel rpm - local absolute path
+    # kernel rpm - Node's local absolute path
     kernel_rpm_path: str = field(
         default="",
         metadata=field_metadata(
@@ -44,10 +43,11 @@ class RPMInstaller(BaseInstaller):
             )
         runbook: RPMInstallerSchema = self.runbook
         kernel_rpm_path: str = runbook.kernel_rpm_path
-        err: str = f"can not find kernel rpm file: {kernel_rpm_path}"
-        assert os.path.exists(kernel_rpm_path), err
-        assert kernel_rpm_path.endswith(
-            ".rpm"
+        assert self._node.shell.exists(
+            PurePath(kernel_rpm_path)
+        ), f"Node does not contain kernel rpm file: {kernel_rpm_path}"
+        assert self._node.tools[Rpm].is_valid_package(
+            kernel_rpm_path
         ), f"Provided file {kernel_rpm_path} is not an rpm"
 
     def install(self) -> str:
@@ -55,15 +55,9 @@ class RPMInstaller(BaseInstaller):
         runbook: RPMInstallerSchema = self.runbook
         kernel_rpm_path: str = runbook.kernel_rpm_path
 
-        filename = os.path.basename(kernel_rpm_path)
-        remote_path = node.working_path / filename
-        node.shell.copy(
-            PurePath(kernel_rpm_path),
-            remote_path,
-        )
-
         rpm = node.tools[Rpm]
-        rpm.install_local_package(str(remote_path))
+        rpm.install_local_package(kernel_rpm_path, force=True, nodeps=True)
+        filename = PurePath(kernel_rpm_path).name
         installed_kernel_version = filename[: -len(".rpm")]
 
         return installed_kernel_version


lisa/base_tools/rpm.py
```diff
@@ -31,8 +31,25 @@ def get_file_size(self, file: str) -> int:
...
     def install_local_package(self, file: str) -> None:
         self.run(
-            f"-ivh {file}",
+            f'-ivh "{file}"',
             sudo=True,
             expected_exit_code=0,
             expected_exit_code_failure_message=(f"failed to install {file}"),
         )
+
+    def install_local_packages_in_dir(
```
squirrelsc: Let's use the install single file method. It may need to select which packages install. So we can control this list in transformer.

lisa/transformers/rpm_kernel_installer.py
```diff
@@ -18,10 +19,18 @@
 @dataclass_json()
 @dataclass
 class RPMInstallerSchema(BaseInstallerSchema):
-    # kernel rpm - local absolute path
-    kernel_rpm_path: str = field(
+    # kernel rpms list - local absolute path
+    rpm_list: List[str] = field(
```
squirrelsc: It needs to have two transformers. This one is for kernel installation, there are extra steps. It needs to add another transformer called `PackageInstallationTransformer`, and derive a transformer called `RpmPackageInstallationTransformer`.

lisa/transformers/rpm_kernel_installer.py
```diff
@@ -43,27 +52,28 @@ def validate(self) -> None:
...
-            PurePath(kernel_rpm_path),
-            remote_path,
-        )
+        time_now = datetime.utcnow().strftime("%Y_%m_%d_%H_%M_%S")
+        # append time to folder name just in case we have multiple kernel installers
+        remote_path = node.working_path / f"kernel_rpms_{time_now}"
+        node.tools[Mkdir].create_directory(str(remote_path))
+        copy_tool = node.tools[RemoteCopy]
+        for rpm_file in runbook.rpm_list:
+            copy_tool.copy_to_remote(
```
squirrelsc: The copy should be done by the copy transformer, don't hard code it here. In some cases, it may need to support local testing.

lisa/transformers/rpm_kernel_installer.py
pupacha: Sure, I will add the suggested transformers for installation of non-kernel packages. But even for kernel, there could be multiple kernel related packages one wants to install along with the kernel itself for it to work properly. For example, following could be the list of rpms.
* kernel-some_version.rpm
* kernel-devel-some_version.rpm
* kernel-debuginfo-some_version.rpm
* kernel-docs-some_version.rpm
* kernel-tools-some_version.rpm
etc
Hence, the `List` type to support installing these set of kernel packages together.

lisa/transformers/rpm_kernel_installer.py
pupacha: This copy is different from the copy transformer. This doesn't target a user given destination. I used `copy_to_remote` instead of `scp` tool, just because of its easy to use nature.

lisa/transformers/rpm_kernel_installer.py
squirrelsc: Only the first one is the kernel RPM, others are not. Other packages can be installed by another transformer, before or after the kernel transformer.

lisa/base_tools/rpm.py
pupacha: The input given to `install_local_packages_in_dir` is a remote directory (inside `node.working_path`) which is created by the transformer. The rpms are copied into this directory by the transformer itself. So, control of the list is with the transformer already.

lisa/transformers/rpm_kernel_installer.py
pupacha: Sure. I was of the opinion that kernel related packages should be installed together.
I will then keep the Kernel RPM installer as it is and add the new installers as suggested. Thanks.

lisa/base_tools/rpm.py
pupacha: Will remove this function. As discussed in the other comment, this wont be needing.

lisa/transformers/rpm_kernel_installer.py
squirrelsc: What's mean a "user given destination"? The "copy transformer" is `FileUploaderTransformer`. The destination is defined in the transformer, it's the middle transition location to make the whole things works. The logic you implemented is similar with `FileUploaderTransformer`, but it assumes the rpm kernel installation only work, if the rpm files are locally. The assumption shouldn't be here to prevent other scenarios, like install by rpm in build, or test with local rpm packages.

pupacha: I meant there's no specific destination that needs to be taken through runbook. This is just an scp of the given local files to a remote directory under `node.working_path`. 

> but it assumes the rpm kernel installation only work, if the rpm files are locally. The assumption shouldn't be here to prevent other scenarios, like install by rpm in build, or test with local rpm packages.

Yes, I had this assumption to address the immediate use case. But since there is also `FileUploaderTransformer` which can take care of copying the files, Do you suggest continuing with this assumption of local file or change the assumption to remote file? We could also add a parameter `is_remote` to explicitly set through runbook. I want to maintain some uniformity and use same approach for the new `PackageInstallationTransformer` as well.

squirrelsc: the rpm installer should install only local files. If some files need to be copied to the tested node, it should be done by the file uploader transformer. It needs to minimize the functionality of each transformers, so they can be composted together for various scenario.
Sorry, I haven't found it assumes the packages are remote. It should be changed to use file uploader and local only rpm, and hope it doesn't bring too much work to your pipelines.

pupacha: Ok, got it. Then I will change the current Kernel RPMInstaller to use the path as local to the node (remote). And as suggested, FileUploaderTransformer can be used to copy the files. Will follow same approach for new PackageInstallationTransformer.
No, this doesn't bring much work in the pipelines. Thanks for the inputs.

lisa/transformers/rpm_kernel_installer.py
pupacha: Hi @squirrelsc . I'm thinking of following schema for the `PackageInstallationTransformer`. Please let me know if this looks good to you.

```
- type: package_installer
  phase: <PHASE> # should also support 'environment_connected'
  connection:
    .
    .
  reboot: <true|false>
  installer:
    type: local # install local packages
    file_list: <LIST>
    package_manager: 
      type: <rpm|apt|..>
      <additional parameters specific to each manager. For example: force_install>

  #  # Other extendable installers could be
  # installer:
  #   type: remote # install remote packages
  #   packages: <LIST>
         
```

squirrelsc: 1. The `package_installer` is like a base class. It doesn't need to have a layer in the runbook. So the type is like `rpm_installer`. The package_manager and other properties are set in the `RpmPackageInstallerTransformer`, which inherits from `PackageInstallerTransformer`. The `KernelInstaller` is not an example of the `PackageInstallerTransformer`, because the `KernelInstaller` compose different types. But `PackageInstallerTransformer` just extends methods and interface, the derived package installers are the same methods.
2. For reboot, the default value should be `False`.

lisa/base_tools/rpm.py
```diff
@@ -20,6 +20,12 @@ def install(self) -> bool:
         posix_os.install_packages(package_name)
         return self._check_exists()
 
+    def is_rpm(self, file: str) -> bool:
```
squirrelsc: `is_valid_package`

lisa/transformers/rpm_kernel_installer.py
pupacha: Yes, for reboot, `False` will be the default value.
As you suggested and discussed offline, I will go ahead with simpler `rpm_installer` as transformer type.

lisa/base_tools/rpm.py
pupacha: Done.

squirrelsc:
We can use this PR to modify existing installer, and next PR to add the new RPM transformer.

pupacha:
> We can use this PR to modify existing installer, and next PR to add the new RPM transformer.

Sure, will create a new PR for new transformers. I'm done with changes for existing installer.

lisa/transformers/rpm_kernel_installer.py
```diff
@@ -44,26 +50,19 @@ def validate(self) -> None:
...
-        remote_path = node.working_path / filename
-        node.shell.copy(
-            PurePath(kernel_rpm_path),
-            remote_path,
-        )
-
         rpm = node.tools[Rpm]
-        rpm.install_local_package(str(remote_path))
-        installed_kernel_version = filename[: -len(".rpm")]
+        rpm.install_local_package(kernel_rpm_path)
```
squirrelsc: We can continue to assume the file name has the version information. To pass in the version information makes the transformer is harder to use.

pupacha: Older code was like
`installed_kernel_version = filename[: -len(".rpm")]`
I will keep it as it is then. It will assume that files have `.rpm` extension.

squirrelsc (approved):

PR #3045, opened 2023-11-14
verify_repository_installed: support freebsd

LiliDeng:


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -737,7 +737,7 @@ def verify_repository_installed(self, node: Node) -> None:  # noqa: C901
                     f"{id_} repository should be present",
                 ).is_true()
         else:
-            raise LisaException(f"Unsupported distro type : {type(node.os)}")
+            raise SkippedException(f"Unsupported distro type : {type(node.os)}")
```
squirrelsc: 1. It should use the unsupported os exception.
2. The error message is not clear enough. Is it not supported by LISA, or the repo is not supported? I know some distro doesn't have repo, it's ok to skip the test case for this case. But is there any distro unsupported by LISA?

LiliDeng: updated

squirrelsc (approved):

PR #3050, opened 2023-11-16
Fix GPU cases for Ubuntu 2304 and 2310

adityagesh:
Check commits for description


lisa/features/gpu.py
```diff
@@ -67,8 +67,6 @@ class Gpu(Feature):
         "libelf-dev",
         "linux-tools-$(uname -r)",
         "linux-cloud-tools-$(uname -r)",
-        "python3",
```
squirrelsc, line 70: Make sure no regression on existing versions.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -531,6 +532,19 @@ def get_supported_driver(self) -> List[ComputeSDK]:
             )
         return driver_list
 
+    # GRID driver is supported on a limited number of distros.
+    # https://learn.microsoft.com/en-us/azure/virtual-machines/linux/n-series-driver-setup#nvidia-grid-drivers # noqa: E501
```
squirrelsc, line 543: Follow the document to implement the allowed list instead of unsupported list.

lisa/features/gpu.py
adityagesh: Checked load_gpu case for 1604, 1804, 2004, 2204, 2304, 2310

squirrelsc:
@LiliDeng LGTM

lisa/sut_orchestrator/azure/features.py
```diff
@@ -556,9 +570,10 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._is_nvidia = True
 
     def _install_driver_using_platform_feature(self) -> None:
+        # https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/hpccompute-gpu-linux
         supported_versions: Dict[Any, List[str]] = {
```
LiliDeng, line 578: will it impact all of gpu test cases?

adityagesh: As per documentation GPU installation using extension is only supported in these. The support for the older distro's are probably removed now.

adityagesh: Do you want me to revert and keep it as the old values itself?

LiliDeng (approved):

PR #3051, opened 2023-11-16
Add new transformer `rpm_package_installer`

pupacha:
Added a new transformer that can install a list of rpm package files that are already present on the node.

### DIFF
--- a/lisa/transformers/package_installer.py
+++ b/lisa/transformers/package_installer.py
@@ -1,0 +1,91 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from dataclasses import dataclass, field
+from pathlib import PurePath
+from typing import Any, Dict, List, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import schema
+from lisa.operating_system import RPMDistro
+from lisa.tools import Rpm
+from lisa.transformers.deployment_transformer import (
+    DeploymentTransformer,
+    DeploymentTransformerSchema,
+)
+from lisa.util import UnsupportedDistroException, field_metadata
+
+
+@dataclass_json()
+@dataclass
+class PackageInstallerSchema(DeploymentTransformerSchema):
+    # Node's local absolute path of directory where package files are located
+    directory: str = field(default="", metadata=field_metadata(required=True))
+    # list of package files to be installed
+    files: List[str] = field(default_factory=list)
+    # reboot after installation
+    reboot: bool = field(default=False)
+
+
+class PackageInstaller(DeploymentTransformer):
+    @classmethod
+    def type_name(cls) -> str:
+        return "package_installer"
+
+    def _validate_package(self, file: str) -> None:
+        raise NotImplementedError()
+
+    def _install_package(self, file: str) -> None:
+        raise NotImplementedError()
+
+    def _validate(self) -> None:
+        runbook: PackageInstallerSchema = self.runbook
+        directory: PurePath = PurePath(runbook.directory)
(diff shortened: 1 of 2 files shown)


lisa/transformers/package_installer.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def validate(self) -> None:
+        raise NotImplementedError()
+
+
+@dataclass_json()
+@dataclass
+class RPMPackageInstallerTransformerSchema(PackageInstallerSchema):
+    # Node's local absolute path of directory where rpm files are located
+    directory_path: str = field(default="", metadata=field_metadata(required=True))
```
squirrelsc: "directory" and "path" are similar. Use `path` is enough.

lisa/transformers/package_installer.py
```diff
@@ -0,0 +1,89 @@
...
+class PackageInstallerSchema(DeploymentTransformerSchema):
+    ...
+
+
+class PackageInstaller(DeploymentTransformer):
+    @classmethod
+    def type_name(cls) -> str:
+        return "package_installer"
+
+    def validate(self) -> None:
```
squirrelsc, line 30: When the validate will be called? It's not in the flow of transformer. It doesn't need here.

squirrelsc: The validate is not called in the transformer flow. If you need it to be called in the base PackageInstaller, it needs to implement `_internal_run` in this class. You can think about wrap two methods for derived class, like `_validate_package` and `_install_package`. So the loop, reboot logic are implemented in the base class.

lisa/transformers/package_installer.py
pupacha: `path` might mislead into path of the file. How about just `directory` ?

lisa/transformers/package_installer.py
pupacha: Sure, this is a good idea. It will need change in the schema of the base transformer as well. I somehow missed to call `validate` in the `_internal_run` method.

lisa/transformers/package_installer.py
squirrelsc: path is shorter. But it's ok to me for `directory`.

pupacha: Thanks. going with `directory`

lisa/transformers/package_installer.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return RPMPackageInstallerTransformerSchema
+
+    @property
+    def _output_names(self) -> List[str]:
+        return []
+
+    def _validate(self) -> None:
```
squirrelsc, line 83: The subclasses can implement methods on single package, like `_validate_package`, `_install_package`. So there can be advanced logic.

lisa/transformers/package_installer.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class RPMPackageInstallerTransformer(PackageInstaller):
+    @classmethod
+    def type_name(cls) -> str:
+        return "rpm_package_installer"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return RPMPackageInstallerTransformerSchema
```
squirrelsc: If there is no extra parameters, it doesn't need to declare a new subclass of schema. That's why the type and schema are two fields, instead of one.

pupacha: Sure, removed it.

squirrelsc (approved):

PR #3057, opened 2023-11-17
add check_or_set_storage_account_blob_public_access

LiliDeng:
Fix failed. ResourceExistsError: Public access is not permitted on this storage account. when run 
```
            container_client.set_container_access_policy(
                signed_identifiers={}, public_access="container"
            )
```


lisa/sut_orchestrator/azure/common.py
```diff
@@ -2353,3 +2354,23 @@ def is_cloud_init_enabled(node: Node) -> bool:
     ) and ls_tool.path_exists("/var/lib/cloud/instance", sudo=True):
         return True
     return False
+
+
+def check_or_set_storage_account_blob_public_access(
```
squirrelsc, line 2359: Can we ask the runtime team to use their public storage account? It would be risk to all subscriptions, which runs this test case.

PR #3062, opened 2023-11-22
No need to check the authorized_keys file if it is empty

lubaihua33:
Cloud-init installs an empty authorized_keys file under the users' folder if the VM is deployed with the "Password" authentication type. If the authorized_keys file is empty, we don't need to report it.

### DIFF
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -990,11 +990,13 @@ class AzureImageStandard(TestSuite):
                 # then it is harmless, otherwise fail the case
                 # command="echo 'Please login as the user \"USERNAME\" rather than the user \"root\".';echo;sleep 10;exit 142"  # noqa: E501
                 key_content = node.tools[Cat].read(file_path, sudo=True)
-                key_match = key_pattern.findall(key_content)
-                if not (key_match and key_match[0]):
-                    assert_that(
-                        file_exists, f"{file_path} is detected. It should be deleted."
-                    ).is_false()
+                if key_content:
+                    key_match = key_pattern.findall(key_content)
+                    if not (key_match and key_match[0]):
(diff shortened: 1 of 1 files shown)


LiliDeng (dismissed):

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -985,7 +985,8 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
             passwd_matches = re.split(":", passwd_raw_output)
             file_path = f"{passwd_matches[5]}/.ssh/authorized_keys"
             file_exists = node.tools[Ls].path_exists(file_path, sudo=True)
-            if file_exists:
+            stat = node.tools[Stat]
```
LiliDeng: Please check stat support freebsd or not

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -985,7 +985,8 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
+            if file_exists and stat.get_total_size(file_path, sudo=True) > 0:
```
LiliDeng: we can move stat into if block to reduce the time, just use this tool if file exists

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: Thank you for this comment. After double checking, the output of the command " stat --format='%s' " in freebsd is different. 

I changed to check if the key_content is empty.

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: I changed to check if the key_content is empty.

LiliDeng (approved):

PR #3071, opened 2023-11-27
add an ignorable call trace

LiliDeng:


squirrelsc (dismissed):

lisa/features/serial_console.py
```diff
@@ -33,6 +33,8 @@ class SerialConsole(Feature):
         re.compile(
             r"^(.*ipt_CLUSTERIP: ClusterIP.*loaded successfully.*)$", re.MULTILINE
         ),
+        # the call trace is annoying but does not affect correct operation of the VM.
```
squirrelsc: Because it's a RIP. Provide more information, like which image has this, and what does it mean.

LiliDeng: updated

squirrelsc (approved):

PR #3073, opened 2023-11-28
Add cd-rom 'empty tray' regression test

mcgov:
Adding a regression test to check virtual cd/dvd tray reports 'empty' after initial provisioning is complete.
A recent bug would cause the tray to report CDS_DISK_OK which means 'the drive has a disk inside and it's ready to use' even though there is no ISO/CD/DVD in the 'tray'. 
This test uses a small portable C program to check that the behavior is fixed after the initial provisioning process; all linux distros will support these commands; they've been around as long as CD-ROM support has.

### DIFF
--- a/microsoft/testsuites/cdrom/cdrom.py
+++ b/microsoft/testsuites/cdrom/cdrom.py
@@ -1,14 +1,16 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 import pathlib
+from typing import Any
 
 from assertpy import assert_that
 
 from lisa import Node, SkippedException, TestCaseMetadata, TestSuite, TestSuiteMetadata
-from lisa.operating_system import Linux
+from lisa.operating_system import CBLMariner, Debian, Linux
 from lisa.sut_orchestrator import AZURE
 from lisa.testsuite import simple_requirement
 from lisa.tools import Gcc
+from lisa.util.logger import Logger
 
 
 @TestSuiteMetadata(
@@ -57,6 +59,10 @@ class CdromSuite(TestSuite):
         node.tools[Gcc].compile(
             filename=str(node_source_path), output_name=str(node_exec_path)
         )
+        assert_that(node.shell.exists(node_exec_path)).described_as(
+            "Test bug! The build is broken on this distro. "
+            f"Could not find {str(node_exec_path)}"
+        ).is_true()
 
         # and run the test if the device is present.
         result = node.execute(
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/cdrom/cdrom.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+
+    @TestCaseMetadata(
+        description="""
+            Test to check the installation ISO is unloaded
+            after provisioning and rebooting a new VM.
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            supported_os=[Linux], supported_platform_type=[AZURE]
```
squirrelsc, line 37: It doesn't need to specify Linux, it's default settings.

microsoft/testsuites/cdrom/cdrom.py
```diff
@@ -0,0 +1,80 @@
...
+        # and run the test if the device is present.
+        result = node.execute(
+            cmd=str(node_exec_path),
+            shell=True,
+            sudo=True,
+        )
+        output = result.stdout.strip()
+
+        # first check for failure to open the cdrom device
+        assert_that(result.exit_code).described_as(
```
squirrelsc, line 76: I feel it doesn't need to assert for -1 separately. The raw output should be able to explain the error.

microsoft/testsuites/cdrom/cdrom.py
```diff
@@ -0,0 +1,80 @@
...
+    # CD drive status constants from linux uapi
+    # see https://github.com/torvalds/linux/blob/master/include/uapi/linux/cdrom.h#L417
+    _expected_device_status = (
+        "CDS_NO_DISC"  # drive tray is closed, no disk in the drive
+    )
+
+    @TestCaseMetadata(
+        description="""
+            Test to check the installation ISO is unloaded
+            after provisioning and rebooting a new VM.
```
LiliDeng, line 33: what is the impact if the iso is not unloaded?

mcgov: I am not aware of the full scope of the issue, I believe Michael pointed the problem out and the host OS team has a fix ready. I cleaned up the very basic test that was shared with me and implemented it as a LISA test for regression checks going forward.
I would imagine any program which waits for the tray to be empty would get stuck forever though. Think 'Remove install ISO...'

microsoft/testsuites/cdrom/cdrom.py
mcgov: Does supported_os=[Linux] not enforce that other OS's are unsupported?
I've only tested on Linux and don't see the need to support BSD. It was a host side fix so whichever OS checks the status of the drive should get the same result.

microsoft/testsuites/cdrom/cdrom.py
mcgov: I disagree, it's a subtle distinction but not being able to open /dev/cdrom implies a test bug (priviledges or distro differences) while the second assertion definitively implies a host bug. 

Having different error messages for each is a time saver when you're triaging through hundreds of tests. This is a bigger issue but there are a lot of error messages that LISA raises that I wish were more specific about _what should be done_ when you see the error raised.

LiliDeng:
please fix compile issue on mariner @mcgov

mcgov:
> please fix compile issue on mariner @mcgov

Fixed! Thanks for pointing this out. I've decided to only support Mariner and Debian to reduce future build-breaking headaches. I've also added a check for a failed build and throw a specific error message for that case so it should be easier to triage in the future.

LiliDeng (approved):

PR #3074, opened 2023-11-28
Set USE_MS_GUEST_KERNEL for CLH testcase only if YES

smit-gardhariya:
We will set the env var USR_MS_GUEST_KERNEL only
if it is set to YES as new CLH dev_cli script had some changes which expect this env var to be set only
when required. Setting default value to NO would not work with it anymore.

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -230,7 +230,8 @@ class CloudHypervisorTests(Tool):
                 auth_token=self.ms_access_token,
             )
             self.env_vars["GUEST_VM_TYPE"] = self.clh_guest_vm_type
-            self.env_vars["USE_MS_GUEST_KERNEL"] = self.use_ms_guest_kernel
+            if self.use_ms_guest_kernel:
+                self.env_vars["USE_MS_GUEST_KERNEL"] = self.use_ms_guest_kernel
         else:
             git.clone(self.upstream_repo, clone_path)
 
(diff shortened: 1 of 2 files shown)


LiliDeng (approved):

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -230,7 +230,8 @@ def _install(self) -> bool:
                 auth_token=self.ms_access_token,
             )
             self.env_vars["GUEST_VM_TYPE"] = self.clh_guest_vm_type
-            self.env_vars["USE_MS_GUEST_KERNEL"] = self.use_ms_guest_kernel
+            if self.use_ms_guest_kernel:
```
pupacha, line 233: we can remove the condition from `ch_tests.py` and have directly here as

```
if self.use_ms_guest_kernel == "YES":
    self.env_vars["USE_MS_GUEST_KERNEL"] = self.use_ms_guest_kernel
```

smit-gardhariya: It would be better to put some check on value before setting it. We already have that check before executing the testcase where we set the class variable.

pupacha: In `ch_tests.py`, you are setting the variable only if `use_ms_guest_kernel` is "YES". And in `ch_tests_tool.py`, you are checking if the variable is set.  So why not just a single condition in `ch_tests_tool.py` that does "YES" check?

PR #3075, opened 2023-11-28
Create Sysbench tool and testcases using it for CPU/Memory/FileIO performance

smit-gardhariya:
This PR has 2 commits.

1. Create Sysbench Tool in LISA
2. Create testcases for CPU/Memory/FileIO performance using Sysbench tool

### DIFF
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -150,7 +150,32 @@ DiskType = Enum(
 
 
 @dataclass
-class DiskPerformanceMessage(PerfMessage):
+class SysbenchPerfMessage:
+    # Sysbench common params
+    min_latency_ms: float = 0.00
+    max_latency_ms: float = 0.00
+    avg_latency_ms: float = 0.00
+    percentile_95_latency_ms: float = 0.00
+    sum_latency_ms: float = 0.00
+    total_time: float = 0.00
+    events_avg: float = 0.00
+    events_stddev: float = 0.00
+    execution_time_avg: float = 0.00
+    execution_time_stddev: float = 0.00
+    total_events: Decimal = Decimal(0)
+    threads: int = 0
+    events: int = 0
+    time_limit_sec: int = 0
+
+
+@dataclass
+class CPUPerformanceMessage(PerfMessage, SysbenchPerfMessage):
+    benchmark: str = ""
+    cpu_speed: Decimal = Decimal(0)
+
+
+@dataclass
+class DiskPerformanceMessage(PerfMessage, SysbenchPerfMessage):
     disk_setup_type: DiskSetupType = DiskSetupType.raw
     block_size: int = 0
     disk_type: DiskType = DiskType.nvme
@@ -167,6 +192,35 @@ class DiskPerformanceMessage(PerfMessage):
     write_lat_usec: Decimal = Decimal(0)
     randwrite_iops: Decimal = Decimal(0)
     randwrite_lat_usec: Decimal = Decimal(0)
+
+    # Sysbench FileIO params
+    read_mib_per_sec: Decimal = Decimal(0)
+    write_mib_per_sec: Decimal = Decimal(0)
+    fsyncs_per_sec: Decimal = Decimal(0)
+    file_fsync_all: str = ""
+    file_fsync_end: str = ""
+    total_file: int = 0
+    file_total_size_in_gb: int = 0
+    file_async_backlog: int = 0
+    file_fsync_freq: int = 0
+    file_merged_requests: int = 0
+    file_rw_ratio: float = 0
+    file_ops: str = ""
+    file_io_mode: str = ""
+    file_fsync_mode: str = ""
+
+
+@dataclass
+class MemoryPerformanceMessage(PerfMessage, SysbenchPerfMessage):
+    total_mib_transferred: int = 0
+    block_size_in_kb: int = 0
+    memory_total_size_in_gb: int = 0
+    mib_per_second: float = 0.0
+    operations_per_second: float = 0
+    scope: str = ""
+    hugetlb_on: bool = False
+    access_mode: str = ""
+    operation: str = ""
 
 
 @dataclass
--- a/lisa/tools/sysbench.py
+++ b/lisa/tools/sysbench.py
@@ -1,0 +1,562 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import inspect
+import re
+from typing import Any, Dict, List, cast
+
+from lisa import notifier
+from lisa.executable import Tool
+from lisa.messages import (
+    CPUPerformanceMessage,
+    DiskPerformanceMessage,
+    MemoryPerformanceMessage,
+    create_perf_message,
+)
+from lisa.operating_system import Posix
+from lisa.util import LisaException, find_group_in_lines, find_groups_in_lines
+from lisa.util.process import ExecutableResult
+
+
+class Sysbench(Tool):
+    @property
+    def command(self) -> str:
+        return "sysbench"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        posix_os.install_packages("sysbench")
+
+        return self._check_exists()
+
+    def __run_sysbench_test(
+        self,
+        test_type: str,
+        threads: int,
+        events: int,
+        time_limit: int,
+        validation: bool = True,
+        extra_args: str = "",
+    ) -> ExecutableResult:
+        validate: str = "off"
+        if validation:
+            validate = "on"
+        args = (
+            f"{test_type} run --threads={threads} --events={events}"
+            f" --validate={validate} --percentile=95"
+            f" --verbosity=5 --histogram=on"
+            f" --time={time_limit} --debug=on {extra_args}"
+        )
+
+        result: ExecutableResult = self.run(
+            args,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="Sysbench Test failed",
+        )
+        return result
+
+    def run_cpu_perf(
+        self,
+        test_result: Any,
+    ) -> None:
+        threads: int = 1
+        events: int = 0
+        cpu_max_prime: int = 10000
+        time_limit: int = 10
+
+        extra_args: str = f" --cpu-max-prime={cpu_max_prime}"
+        perf_msg: Dict[Any, Any] = {}
+        testcase_name: str = inspect.stack()[1][3]
+        res = self.__run_sysbench_test(
+            test_type="cpu",
+            threads=threads,
+            events=events,
+            time_limit=time_limit,
+            extra_args=extra_args,
+        )
+        perf_msg = self.__process_cpu_perf_result(res.stdout)
+        perf_msg["benchmark"] = f"prime_{cpu_max_prime}"
+
+        self.__send_subtest_msg(
+            message_type=CPUPerformanceMessage,
+            test_name=testcase_name,
+            test_result=test_result,
+            other_fields=perf_msg,
+        )
+
+    def run_fileio_perf(
+        self,
+        test_result: Any,
+        total_file: int = 128,
+        io_ops: Any = None,
+    ) -> None:
+        threads: int = 1
+        events: int = 0
+        time_limit: int = 10
+        file_io_mode: str = "sync"
+        file_fsync_all: str = "off"
+        file_fsync_mode: str = "fsync"
+        file_fsync_end: bool = False
+        block_size_in_kb: int = 16
+        file_total_size_in_gb: int = 2
+        file_async_backlog: int = 128
+        file_fsync_freq: int = 100
+        file_merged_requests: int = 0
+        file_rw_ratio: float = 1.5
+
+        total_size = file_total_size_in_gb * 1024 * 1024 * 1024
+        block_size = block_size_in_kb * 1024
+        valid_io_mode: List[str] = ["sync", "async", "mmap"]
+        valid_fsync_mode: List[str] = ["fsync", "fdatasync"]
+        valid_test_mode: List[str] = [
+            "seqwr",
+            "seqrd",
+            "rndrd",
+            "rndwr",
+            "seqrewr",
+            "rndrw",
+        ]
+
+        perf_msg: Dict[Any, Any] = {}
+        testcase_name: str = inspect.stack()[1][3]
+        fsync_end: str = "on" if file_fsync_end else "off"
+
+        for operation in io_ops:
+            if operation not in valid_test_mode:
+                raise LisaException(
+                    f"Invalid io_ops: '{operation}' "
+                    f"Valid values: {', '.join(valid_test_mode)}"
+                )
+        if not io_ops:
+            # Run default for all IO test_modes
+            io_ops = valid_test_mode
+
+        if file_io_mode not in valid_io_mode:
+            raise LisaException(
+                f"Invalid file_io_mode. Valid file_io_mode: {valid_io_mode}"
+            )
+
+        if file_fsync_mode not in valid_fsync_mode:
+            raise LisaException(
+                "Invalid file_fsync_mode. Valid file_fsync_mode: {valid_fsync_mode}"
+            )
+
+        for operation in io_ops:
+            extra_args: str = (
+                f" --file-test-mode={operation} --file-num={total_file}"
+                f" --file-block-size={block_size}"
+                f" --file-total-size={total_size}"
+                f" --file-io-mode={file_io_mode}"
+                f" --file-async-backlog={file_async_backlog}"
+                f" --file-fsync-freq={file_fsync_freq}"
+                f" --file-fsync-all={file_fsync_all}"
+                f" --file-fsync-end={fsync_end}"
+                f" --file-fsync-mode={file_fsync_mode}"
+                f" --file-merged-requests={file_merged_requests}"
+                f" --file-rw-ratio={file_rw_ratio}"
+            )
+
+            prepare_cmd: str = (
+                f"fileio prepare --file-total-size={total_size} --file-num={total_file}"
+            )
+            self.run(
+                prepare_cmd,
+                force_run=True,
+            )
+
+            res = self.__run_sysbench_test(
+                test_type="fileio",
+                threads=threads,
+                events=events,
+                time_limit=time_limit,
+                validation=False,
+                extra_args=extra_args,
+            )
+
+            cleanup_cmd: str = (
+                f"fileio cleanup --file-total-size={total_size} --file-num={total_file}"
+            )
+            self.run(
+                cleanup_cmd,
+                force_run=True,
+            )
+
+            perf_msg = self.__process_fileio_perf_result(
+                res.stdout,
+                operation,
+            )
+            perf_msg["file_fsync_all"] = file_fsync_all
+            perf_msg["file_fsync_end"] = file_fsync_end
+            perf_msg["total_file"] = total_file
+            perf_msg["file_total_size_in_gb"] = file_total_size_in_gb
+            perf_msg["file_async_backlog"] = file_async_backlog
+            perf_msg["file_fsync_freq"] = file_fsync_freq
+            perf_msg["file_merged_requests"] = file_merged_requests
+            perf_msg["file_rw_ratio"] = file_rw_ratio
+            perf_msg["file_ops"] = operation
+            perf_msg["file_io_mode"] = file_io_mode
+            perf_msg["file_fsync_mode"] = file_fsync_mode
+
+            self.__send_subtest_msg(
+                message_type=DiskPerformanceMessage,
+                test_name=testcase_name,
+                test_result=test_result,
+                other_fields=perf_msg,
+            )
+
+    def run_memory_perf(
+        self,
+        test_result: Any,
+        memory_oper: Any = None,
+        memory_access_mode: Any = None,
+    ) -> None:
+        hugetlb: str = "off"
+        threads: int = 1
+        events: int = 0
+        time_limit: int = 10
+        memory_block_size_in_kb: int = 1
+        memory_total_size_in_gb: int = 32
+        memory_scope: str = "global"
+
+        valid_mem_scope: List[str] = ["global", "local"]
+        valid_mem_operation: List[str] = ["read", "write", "none"]
+        valid_mem_access_mode: List[str] = ["seq", "rnd"]
+        perf_msg: Dict[Any, Any] = {}
+        testcase_name: str = inspect.stack()[1][3]
+
+        # Calculate block/total memory size in bytes
+        block_size = memory_block_size_in_kb * 1024
+        total_mem_size = memory_total_size_in_gb * 1024 * 1024 * 1024
+
+        if not memory_oper:
+            # Run with default operation if not passed
+            memory_oper = valid_mem_operation
+
+        if not memory_access_mode:
+            # Run with default access mode if not passed
+            memory_access_mode = valid_mem_access_mode
+
+        if memory_scope not in valid_mem_scope:
+            raise LisaException(
+                f"Invalid memory_scope. Valid memory_scope: {valid_mem_scope}"
+            )
+
+        for opr in memory_oper:
+            if opr not in valid_mem_operation:
+                raise LisaException(
+                    f"Invalid memory_oper: '{opr}'"
+                    f"Valid memory_oper: {valid_mem_operation}"
+                )
+        for acc in memory_access_mode:
+            if acc not in valid_mem_access_mode:
+                raise LisaException(
+                    f"Invalid memory_access_mode: {acc}"
+                    f"Valid memory_access_mode: {valid_mem_access_mode}"
+                )
+
+        for op in memory_oper:
+            for access_mode in memory_access_mode:
+                extra_args: str = (
+                    f" --memory-block-size={block_size}"
+                    f" --memory-total-size={total_mem_size}"
+                    f" --memory-scope={memory_scope} --memory-hugetlb={hugetlb}"
+                    f" --memory-oper={op} --memory-access-mode={access_mode}"
+                )
+
+                res = self.__run_sysbench_test(
+                    test_type="memory",
+                    threads=threads,
+                    events=events,
+                    time_limit=time_limit,
+                    extra_args=extra_args,
+                )
+
+                perf_msg = self.__process_memory_perf_result(res.stdout)
+                perf_msg["block_size_in_kb"] = memory_block_size_in_kb
+                perf_msg["memory_total_size_in_gb"] = memory_total_size_in_gb
+                perf_msg["scope"] = memory_scope
+                perf_msg["hugetlb_on"] = hugetlb
+                perf_msg["access_mode"] = access_mode
+                perf_msg["operation"] = op
+                perf_msg["threads"] = threads
+                perf_msg["events"] = events
+                perf_msg["time_limit_sec"] = time_limit
+
+                self.__send_subtest_msg(
+                    message_type=MemoryPerformanceMessage,
+                    test_name=testcase_name,
+                    test_result=test_result,
+                    other_fields=perf_msg,
+                )
+
+    def __process_perf_result(
+        self,
+        data: str,
+    ) -> Dict[Any, Any]:
+        # Sample Output
+        # ================
+        # General statistics:
+        # total time:                          10.0005s
+        # total number of events:              27617
+
+        # Latency (ms):
+        #         min:                                    0.33
+        #         avg:                                    0.36
+        #         max:                                   10.14
+        #         95th percentile:                        0.43
+        #         sum:                                 9988.94
+
+        # Threads fairness:
+        #     events (avg/stddev):           27617.0000/0.00
+        #     execution time (avg/stddev):   9.9889/0.00
+
+        # DEBUG: Verbose per-thread statistics:
+
+        # DEBUG:   thread #  0: min: 0.0003s  avg: 0.0004s  max: 0.0101s  events: 27617
+        # DEBUG:                  total time taken by event execution: 9.9889s
+
+        result: Dict[Any, Any] = {}
+
+        # Extract total time using regular expression
+        total_time_pattern = re.compile(r"total time:\s+(?P<total_time>[\d.]+s)")
+        match = find_group_in_lines(
+            lines=data,
+            pattern=total_time_pattern,
+            single_line=False,
+        )
+        if match:
+            result["total_time"] = match["total_time"]
+
+        # Extract total number of events using regular expression
+        total_events_pattern = re.compile(
+            r"total number of events:\s+(?P<total_events>\d+)"
+        )
+        match = find_group_in_lines(
+            lines=data,
+            pattern=total_events_pattern,
+            single_line=False,
+        )
+        if match:
+            result["total_events"] = match["total_events"]
+
+        # Extract latency information using regular expressions
+        latency_param = "latency_ms"
+        latency_metrics_pattern = re.compile(
+            r"(?P<metric>min|avg|max|95th percentile|sum):\s+(?P<value>[\d.]+)\s"
+        )
+        matches = find_groups_in_lines(
+            lines=data,
+            pattern=latency_metrics_pattern,
+            single_line=False,
+        )
+        for match in matches:
+            metric = match["metric"].strip().replace(" ", "_")
+            if match["metric"] == "95th percentile":
+                metric = "percentile_95"
+            result[f"{metric}_{latency_param}"] = float(match["value"])
+
+        # Extract thread event avg/stddev
+        thread_events_pattern = re.compile(
+            r"events \(avg/stddev\):\s+(?P<event_avg>[\d.]+)/(?P<event_stddev>[\d.]+)"
+        )
+        thread_events = find_groups_in_lines(
+            lines=data,
+            pattern=thread_events_pattern,
+            single_line=False,
+        )
+        if thread_events:
+            result["events_avg"] = thread_events[0]["event_avg"]
+            result["events_stddev"] = thread_events[0]["event_stddev"]
+
+        # Extract execution time avg/stddev
+        thread_exec_time_pattern = re.compile(
+            r"execution time \(avg/stddev\):\s+(?P<avg>[\d.]+)/(?P<std_dev>[\d.]+)"
+        )
+        exec_time = find_groups_in_lines(
+            lines=data,
+            pattern=thread_exec_time_pattern,
+            single_line=False,
+        )
+        if exec_time:
+            result["execution_time_avg"] = exec_time[0]["avg"]
+            result["execution_time_stddev"] = exec_time[0]["std_dev"]
+
+        return result
+
+    def __process_memory_perf_result(
+        self,
+        data: str,
+    ) -> Dict[Any, Any]:
+        result: Dict[Any, Any] = self.__process_perf_result(data)
+
+        operations_per_second = None
+        total_mib_transferred = None
+        mib_per_second = None
+
+        # Extract Total operations and operations per second
+        # Sample Output
+        # Total operations: 65730837 (6571036.01 per second)
+        total_operations_pattern = re.compile(
+            r"Total operations: (?P<operation>\d+) \((?P<per_sec>[\d.]+) per second\)"
+        )
+        match = find_groups_in_lines(
+            lines=data,
+            pattern=total_operations_pattern,
+            single_line=False,
+        )
+        if match:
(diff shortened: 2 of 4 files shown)


lisa/tools/sysbench.py
```diff
@@ -0,0 +1,217 @@
...
+        args = (
+            f"{test_type} run --threads={threads} --events={events}"
+            f" --validate={validation} --percentile={percentile}"
+            f" --verbosity={verbosity} --histogram={histogram}"
+            f" --time={time_limit} --debug={debug} {extra_args}"
+        )
+
+        result: ExecutableResult = self.run(
+            f"{args}",
+        )
```
LiliDeng, line 58:
```
        result: ExecutableResult = self.run(args)
```

smit-gardhariya: done

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+    """,
+)
+class SysbenchPerfTestSuite(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            Runs Sysbench test for cpu
+        """,
+        priority=3,
+    )
+    def verify_sysbench_cpu_perf(
```
squirrelsc: The test case names should be like `perf_sysbench_cpu`.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+            res = process_cpu_perf_result(output.stdout, subtest_name)
+
+        self._send_subtest_msg(
+            result,
+            res.name,
+            res.status,
+            res.message,
+        )
+
+        if res.status == TestStatus.FAILED:
```
squirrelsc: `assert_that` makes it more accurate, so it should expect PASS, instead of not FAILED. like `assert_that(res.status).described_as(...).is_equal_to(TestStatus.PASSED)`

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+        res: SysbenchTestResult = SysbenchTestResult()
+
+        output: ExecutableResult = node.tools[Sysbench].run_cpu_perf()
+
+        if output.exit_code != 0:
+            res.name = subtest_name
+            res.message = f"{subtest_name} failed with " f"exit code {output.exit_code}"
+            res.status = TestStatus.FAILED
+        else:
+            res = process_cpu_perf_result(output.stdout, subtest_name)
```
squirrelsc: If the pattern is always used like "run and process". It can be in `run_cpu_perf`, so the caller is simple.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+
+        self._send_subtest_msg(
```
squirrelsc: Subtests is for multiple test results, if there is only one result, it doesn't need to send a subtest result.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+    total_time = re.search(
+        r"total time:\s+([\d.]+s)",
+        non_debug_text,
+    )
+    if total_time:
+        result["total_time"] = total_time.group(1)
+
+    # Extract total number of events using regular expression
+    total_events = re.search(
+        r"total number of events:\s+(\d+)",
```
squirrelsc: 1. Add examples on each regexp pattern.
2. Move them out of the function, and group them together, so it's easy to maintain together.
3. Use named captures to improve readability and easier to maintain.
4. Use methods in utils/__init__.py, like `find_group_in_lines`, `find_groups_in_lines`.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+
+    # Find all non-DEBUG lines in the text
+    non_debug_text = "\n".join(
+        re.findall(
+            non_debug_pattern,
+            data,
+            re.MULTILINE,
+        )
+    )
+    debug_text = "\n".join(
```
squirrelsc: Why does it need to differentiate debug text?

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+@TestSuiteMetadata(
+    area="sysbench",
+    category="performance",
+    description="""
+    This test suite is for executing the sysbench tests
+    """,
+)
+class SysbenchPerfTestSuite(TestSuite):
```
squirrelsc: `SysbenchTestSuite`

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        subtest_name: str = "default_cpu_perf_check"
+        res: SysbenchTestResult = SysbenchTestResult()
+
+        output: ExecutableResult = node.tools[Sysbench].run_cpu_perf()
+
+        if output.exit_code != 0:
```
squirrelsc: In LISA, it's done in one line, including mark the test case fails. Please refer to how other test cases implemented.

`output.assert_exit_code(expected_exit_code=0, message="...")`

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def run_cpu_perf(
+        self,
+        threads: int = 1,
+        events: int = 0,
+        cpu_max_prime: int = 10000,
+        verbosity: int = 5,
+        time_limit: int = 10,
+        percentile: int = 95,
+        debug: str = "on",
```
squirrelsc: It doesn't need to expose all arguments. For example, the debug is always "on", so it should be in the final command arguments only.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,406 @@
...
+    notifier,
+)
+from lisa.messages import TestStatus, send_sub_test_result_message
+from lisa.testsuite import TestResult
+from lisa.tools import Sysbench
+from lisa.util.process import ExecutableResult
+
+
+@dataclass
+class SysbenchTestResult:
```
squirrelsc: It needs to either use existing perf message or add new perf messages. The general test result is not easy to analyze results.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: I added regex examples for reference and using named captures now.

`find_group_in_lines `is not giving me correct result. I have tried below thing for `total number of events`

I passed `single_lines=False` which gives me `{'total_events': '2'}` whereas correct result is  `'total_events': '23274'`

I also printed `find ` variable value in `find_patterns_groups_in_lines `function which seems correct. Issue seems at `results[index].append(dict(zip(pattern.groupindex, find)))`

```
find_group_in_lines(
            lines=data,
            pattern=total_events_pattern,
            single_line=False,
        )
```

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: We have min/max/avg metrics for latency and debug parts (when we run with debug=on)

So reg-ex to get the min/max/avg, we can re-use if we split up the debug text. I added the sample output for reference now.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: This would not let push the failed message for subtest and testcase would throw the error from there.

We already have assert_that on status. exit code check if to determine which message should be send as subtest result.

We are using subtest here because, as of now, we have default run for cpu performance using sysbench, but in future we might run the same with different parameters. In that case, subtest (having different parameters) would make more sense.

lisa/tools/sysbench.py
smit-gardhariya: Default debug would always be on. I wanted to expose all the command line args for sysbench as function parameters so any new requirement where we need to run with say debug=off can be accommodated.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: We are using subtest here because, as of now, we have default run for cpu performance using sysbench, but in future we might run the same with different parameters. Just like we do for fileio/memory to run with different modes. `run_cpu_perf` function exposes all the args of sysbench cpu command line as parameters.

In that case, subtest (having different parameters) would make more sense.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Yes pattern is run the command and process the stdout to extract the result according to testcase.

I wrote result-extraction in such a way that we can re-use it for common parameters. All `process_*_perf_result` would call `process_perf_result` to get common result metrics.

Also in future we might visit the testcase where we would be running sysbench with more permutation combination of parameters for cpu/memory/fileio. This way, it would require minimal changes.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done for all 3 testcases.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Could you please explain a bit here.

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: The `process_cpu_perf_result` can be called in `run_cpu_perf`. So all logic are handled by tool level. The subtest results should be generated whenever the tool runs. It can still add other changes in either method.

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: You can add `$` at the end to match all numbers in last pattern. The (\d+) should match all numbers in greedy mode, but I don't know why it doesn't work.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done. Please check the latest push.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Yes i Tried that too but same result.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: This is removed with latest push

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,183 @@
...
+
+        self._send_subtest_msg(
+            result,
+            subtest_name,
+            status,
+            message,
+        )
+
+        msg = f"Sysbench-cpu test failed with exit code {output.exit_code}"
+        assert_that(status).described_as(msg).is_equal_to(TestStatus.PASSED)
```
squirrelsc: The whole part from sending subresults to assert the exit code can be moved into the tool. The test case level is for test definition and configurations.

smit-gardhariya: I want to make tool functions more generic, so we have more flexibility to pass parameter from testcase.

There are testcases like memory/fileio which will have more that 1 subtest and we want to fail the testcase if anyone of the sub-tests is failed. W.r.t this scenario we will always need one check/assertion on the testcase level.

smit-gardhariya: I have made changes to push the sending subtest results and assert the exit code to tool level. Please check

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: Good catch! I will fix it. Before that, please remove the single_line=False. It doesn't need for this pattern.

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: Do you mean reuse the output? The patterns are different, so they won't conflict. For example, the non-debug test, you can write it like `r"(?P<metric>min|avg|max|95th percentile|sum):\s+(?P<value>[\d.]+)\n"`, the debug text, you can use like `r"(?P<metric>min|avg|max):\s+(?P<value>[\d.]+)s"`, and capture the events separately.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,125 @@
...
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        subtest_name: str = "default_sysbench_cpu_perf_test"
+        output: SysbenchTestResult = node.tools[Sysbench].run_cpu_perf(
+            subtest_name=subtest_name,
+            test_result=result,
+        )
+        failure_msg = f"Sysbench-cpu test failed with exit code {output.exit_code}"
+        assert_that(output.exit_code).described_as(failure_msg).is_equal_to(0)
```
squirrelsc: The assertion can be moved into the run_cpu_perf as well. So if the tool failed, the test case fails too.

lisa/tools/sysbench.py
squirrelsc: LISA is to simplify usage of commands. Only exposing an argument, when it's used in some place.

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: The result is still here. If move the assertion into the tool, it doesn't need to have.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,125 @@
...
+            "seqwr": "write",
+            "seqrd": "read",
+            "rndrd": "read",
+            "rndwr": "write",
+            "seqrewr": "all",
+            "rndrw": "all",
+        }
+        sysbench = node.tools[Sysbench]
+        failed_subtests = []
+        for mode in io_mode_ops.keys():
```
squirrelsc: The loop should be in the tool. if one of them failed, maybe it doesn't need to run following, if they may have similar issues.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,563 @@
...
+
+        return sysbench_result
+
+    def run_fileio_perf(
+        self,
+        test_mode: str,
+        ops: str,
+        test_result: Any,
+        subtest_name: str = "",
+        threads: int = 1,
```
squirrelsc: The same, too many arguments, so it's hard to use.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,563 @@
...
+        events: int = 0,
+        verbosity: int = 5,
+        time_limit: int = 10,
+        percentile: int = 95,
+        debug: str = "on",
+        histogram: str = "on",
```
squirrelsc: The "on/off" should be wrapped as bool to avoid typo issues.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,125 @@
...
+        self,
+        log: Logger,
+        node: Node,
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        io_mode_ops = {
+            "seqwr": "write",
```
squirrelsc: It looks the map is easy to infer from left to right. So it doesn't need to be two parameters to input. It can input only one str list, and the value is evaluated in the tool.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,563 @@
...
+        validation: str = "on",
+    ) -> SysbenchTestResult:
+        extra_args: str = f" --cpu-max-prime={cpu_max_prime}"
+        result: Dict[Any, Any] = {}
+        message: str = ""
+        status = TestStatus.QUEUED
+
+        if not subtest_name:
+            subtest_name = "default_sysbench_cpu_perf_test"
```
squirrelsc: 1. What exact sysbench test executed? Is there a test name from sysbench? If sysbench has only one cpu test, "cpu" is enough. "sysbench" is redundant, because it's in parent test case name.
2. Define the default value on method argument, so it doesn't need to set here.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,563 @@
...
+        # File operations:
+        #     reads/s:                      2717.15
+        #     writes/s:                     1811.43
+        #     fsyncs/s:                     45.39
+
+        # Throughput:
+        #     read, MiB/s:                  42.46
+        #     written, MiB/s:               28.30
+
+        if ops == "write" or ops == "all":
```
squirrelsc: Try to reuse `DiskPerformanceMessage` for disk results. If `DiskPerformanceMessage` misses throughput or other fields, you can add them. The `DiskPerformanceMessage` needs to collect more information like disk_type.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,563 @@
...
+        # 64190.27 MiB transferred (6417.03 MiB/sec)
+        total_mib_transferred_pattern = re.compile(
+            r"(?P<total_mib>[\d.]+) MiB transferred \((?P<per_sec>[\d.]+) MiB/sec\)"
+        )
+        match = total_mib_transferred_pattern.search(data)
+        if match:
+            total_mib_transferred = float(match.group("total_mib"))
+            mib_per_second = float(match.group("per_sec"))
+
+        result["total_operations"] = total_operations
```
squirrelsc: Follow other perf messages to create a mem perf message. The result dict is not easy to query or compare in dashboards.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,563 @@
...
+        result["total_mib_transferred"] = total_mib_transferred
+        result["mib_per_second"] = mib_per_second
+
+        return result
+
+    def __process_cpu_perf_result(
+        self,
+        data: str,
+    ) -> Dict[Any, Any]:
+        result: Dict[Any, Any] = self.__process_perf_result(data)
```
squirrelsc, line 438: create a CPU perf message.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Yes, i tried without the single_line=False and result was empty. I am not using find_group_in_lines in the code as of now.

lisa/tools/sysbench.py
smit-gardhariya: Ok then i will remove some common args with default value as suggested

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: This SysbenchTestResult, we return to testcase for exit code purpose. I have to pass the exit code to fail the parent level testcase. I can return only exit code. Will make that change

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Sure Chi

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: I want to give leverage to testcase what operation it want to perform for performance testing. Say someone just want to run memory-read perf test or fileio with rndrd mode only. This way testcase would be more flexible

lisa/tools/sysbench.py
smit-gardhariya: Yes i will make some of then as default until someone need it.

lisa/tools/sysbench.py
smit-gardhariya: sure

lisa/tools/sysbench.py
smit-gardhariya: Done. Please check latest push

lisa/tools/sysbench.py
smit-gardhariya: Done. Please check latest push

lisa/tools/sysbench.py
smit-gardhariya: Done. Please check latest push

lisa/tools/sysbench.py
smit-gardhariya: Removed subtest_name.

Using create_perf_message as suggested so there wont be subtest now. We will push multiple result for all iterations for given testcase just like Disk/Network perf test we have.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done. Please check latest push

lisa/tools/sysbench.py
smit-gardhariya: Made some common params always on and removed from fn param.

Used boolean for on/off as suggested

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: I merged the fix, please use `find_group_in_lines` or similar in all places.

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: How about this comment? Is it fixed? The non-debug logic is not necessary too.

lisa/messages.py
```diff
@@ -125,6 +125,22 @@ class PerfMessage(MessageBase):
     role: str = ""
     test_result_id: str = ""
 
+    # Sysbench common params
```
squirrelsc: The PerfMessage is the base class for all perf messages. Please either move them into existing perf message, or new perf messages. The fields shouldn't be tool oriented, they should be metrics oriented.

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: The sysbench tool is to hold the test logic, decide pass or fail. The test case level is to modify and pass in different configurations.

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: You can add a test case variable making the list configurable. But it doesn't block to move the loop into the tool. If you need to add more test cases, it doesn't need to copy the loop into the test cases.

lisa/tools/sysbench.py
squirrelsc: `validation` should be changed to.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,481 @@
...
+        }
+        perf_msg: Dict[Any, Any] = {}
+        testcase_name: str = inspect.stack()[1][3]
+        fsync_end: str = "on" if file_fsync_end else "off"
+
+        if operation not in valid_test_mode.keys():
+            raise LisaException(
+                f"Invalid test_mode. Valid test_mode: {valid_test_mode.keys()}"
+            )
+        elif file_io_mode not in valid_io_mode:
```
squirrelsc: The three `if/elif` blocks are not related, so use `if` is clearer.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,481 @@
...
+        other_fields: Dict[str, Any],
+    ) -> None:
+        subtest_msg = create_perf_message(
+            message_type,
+            self.node,
+            test_result,
+            test_name,
+            other_fields,
+        )
+        print(subtest_msg)
```
squirrelsc: remove the debug lines. You can add console notifier to print all messages in debug level.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,481 @@
...
+        valid_test_mode: Dict[str, str] = {
+            "seqwr": "write",
+            "seqrd": "read",
+            "rndrd": "read",
+            "rndwr": "write",
+            "seqrewr": "all",
+            "rndrw": "all",
+        }
+        perf_msg: Dict[Any, Any] = {}
+        testcase_name: str = inspect.stack()[1][3]
```
squirrelsc, line 119: How about put io mode into test case name? I'm not sure if it's easier to compare in dashboard, if the values are separated.

lisa/messages.py
```diff
@@ -168,6 +190,38 @@ class DiskPerformanceMessage(PerfMessage):
     randwrite_iops: Decimal = Decimal(0)
     randwrite_lat_usec: Decimal = Decimal(0)
 
+    # Sysbench FileIO params
+    read_io_per_sec: Decimal = Decimal(0)
```
squirrelsc: Please try best to reuse above fields, like `read_io_per_sec` should be `read_iops` above. So the data from different tools can be compared together.

lisa/messages.py
```diff
@@ -168,6 +190,38 @@ class DiskPerformanceMessage(PerfMessage):
...
+    file_merged_requests: int = 0
+    file_rw_ratio: float = 0
+    file_ops: str = ""
+    file_io_mode: str = ""
+    file_fsync_mode: str = ""
+
+
+@dataclass
+class MemoryPerformanceMessage(PerfMessage):
+    total_operations: Decimal = Decimal(0)
```
squirrelsc: `operations_per_sec` is easier to understand, if it includes the time unit. And similar to other fields.

lisa/messages.py
```diff
@@ -149,6 +165,12 @@ class PerfMessage(MessageBase):
 )
 
 
+@dataclass
+class CPUPerformanceMessage(PerfMessage):
+    cpu_max_prime: Decimal = Decimal(0)
```
squirrelsc: Is it a test parameter to calculate max prime? If so, use a field like "benchmark", and value is "prime". So it's easy to extend to other perf methodologies.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,481 @@
...
+        return result
+
+    def __send_subtest_msg(
+        self,
+        message_type: Any,
+        test_result: Any,
+        test_name: str,
+        other_fields: Dict[str, Any],
+    ) -> None:
+        subtest_msg = create_perf_message(
```
squirrelsc, line 555: The `tool` of perf messages should be set to "sysbench".

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,481 @@
...
+        perf_msg["total_file"] = total_file
+        perf_msg["file_total_size_in_gb"] = file_total_size_in_gb
+        perf_msg["file_async_backlog"] = file_async_backlog
+        perf_msg["file_fsync_freq"] = file_fsync_freq
+        perf_msg["file_merged_requests"] = file_merged_requests
+        perf_msg["file_rw_ratio"] = file_rw_ratio
+        perf_msg["file_ops"] = operation
+        perf_msg["file_io_mode"] = file_io_mode
+        perf_msg["file_fsync_mode"] = file_fsync_mode
+        perf_msg["test_case_name"] = testcase_name
```
squirrelsc: The test_case_name is added below, so it doesn't need to be set in dict.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,481 @@
...
+            other_fields=perf_msg,
+        )
+
+    def run_memory_perf(
+        self,
+        test_result: Any,
+        threads: int = 1,
+        events: int = 0,
+        time_limit: int = 10,
+        memory_block_size_in_kb: int = 1,
```
squirrelsc: Those arguments are not set in test case level too, they should be removed, unless they are used. Similar for fild and cpu.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: We have removed non-debug now. We dont need to push verbose/debug metrics now.

I have just filtered the non-debug lines now

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done

lisa/tools/sysbench.py
smit-gardhariya: Done

lisa/tools/sysbench.py
smit-gardhariya: Done

lisa/tools/sysbench.py
smit-gardhariya: done

lisa/messages.py
smit-gardhariya: Done

lisa/messages.py
smit-gardhariya: There are 2 field for memory perf message. One is `total_operation` and other is `operations_per_second`.

lisa/messages.py
smit-gardhariya: It is max number till which sysbench will try to get the prime numbers to generate CPU stress. Default is 10000 i.e. sysbench cpu run would try to get prime number upto 10000.

lisa/tools/sysbench.py
smit-gardhariya: done

lisa/tools/sysbench.py
smit-gardhariya: Removed now

lisa/messages.py
smit-gardhariya: These are the common one for sysbench. I put them here as i dont want to have redundant params for all child PerfMessage classes.

I created new perf message dataclass now for sysbench for all those common params and inherited it for Disk/CPU/Memory perf dataclasses

lisa/tools/sysbench.py
smit-gardhariya: Done

smit-gardhariya: That is correct that we are not using them as of now as we are running with default values. There is plan to run the CPU/Memory perf test using sysbench in near future for our validation stack which would required to run sysbech to be run with different parameter.

Those args are necessary to expose all command-line parameter of sysbench.

I would suggest if we can keep this so we dont need or need minimal changes at tool level. Please let me know your thoughts on this

microsoft/testsuites/sysbench/sysbenchperf.py
squirrelsc: Please remove the `non_debug_pattern = r"^(?!.*DEBUG: ).*$"` also.

microsoft/testsuites/sysbench/sysbenchperf.py
```diff
@@ -0,0 +1,103 @@
...
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        memory_operation: List[str] = ["read", "write"]
+        memory_access_mode: List[str] = ["seq", "rnd"]
+        sysbench = node.tools[Sysbench]
+
+        for op in memory_operation:
```
squirrelsc: similar to fileio, please move the loop into the tool.

lisa/messages.py
squirrelsc: The perf messages should be metrics oriented, instead of tool oriented. It's easy to create dashboard and explain by metrics concepts. It's the same to sysbench common fields. It looks many common fields are not help a lot as a "common" field, the tested metrics are the key information of a perf message. We need to abstract the values to perf metrics.

For example, it's easy to understand x_latency_ms in network tests, but hard to understand in CPU tests, and I'm not sure what does it mean in the mem tests. The "events" number is highly depends on the test scenarios and settings.

lisa/tools/sysbench.py
```diff
@@ -0,0 +1,561 @@
...
+        valid_mem_access_mode: List[str] = ["seq", "rnd"]
+        perf_msg: Dict[Any, Any] = {}
+        testcase_name: str = inspect.stack()[1][3]
+        hugetlb: str = "off"
+
+        if memory_scope not in valid_mem_scope:
+            raise LisaException(
+                f"Invalid memory_scope. Valid memory_scope: {valid_mem_scope}"
+            )
+        elif memory_oper not in valid_mem_operation:
```
squirrelsc: Similar with fileio, use 3 `if`, instead of `elif` on unrelated checks.

lisa/tools/sysbench.py
squirrelsc: It's still here, not be removed.

lisa/messages.py
squirrelsc: Please check all other fields too.

1. Make sure it's a common concept, instead of sysbench concept.
2. Create it in the Disk message, if it doesn't exist. But not group them like "sysbench params".

lisa/messages.py
```diff
@@ -168,6 +193,36 @@ class DiskPerformanceMessage(PerfMessage):
...
+    file_rw_ratio: float = 0
+    file_ops: str = ""
+    file_io_mode: str = ""
+    file_fsync_mode: str = ""
+
+
+@dataclass
+class MemoryPerformanceMessage(PerfMessage, SysbenchPerfMessage):
+    total_operations: Decimal = Decimal(0)
+    total_mib_transferred: int = 0
```
squirrelsc, line 215: Should it be `total_mib_transferred_per_sec`? It's easy to cross check on different test settings.

lisa/messages.py
squirrelsc: If so, the `total_operation` doesn't need here. It's hard to compare cross different settings. We don't need to save all details in sysbench output.

lisa/messages.py
squirrelsc: It's a test approach. If it's set as a column name, it's hard to extend other tests in the same CPU table. So, the columns should be like below.

benchmark: prime_10000
max_cpu_speed: ???

lisa/tools/sysbench.py
squirrelsc: Let's add them back in "near future", when it's used. That would be clearer than add "all" now.

lisa/messages.py
smit-gardhariya: We have 2 fields for this. `total_mib_transferred `and `mib_per_second`

Here is the reference output. We are fetching both the figure under above metrics.

# 64190.27 MiB transferred (6417.03 MiB/sec)

lisa/tools/sysbench.py
smit-gardhariya: Done

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done

lisa/messages.py
squirrelsc: It looks similar to total operations, it's not a metrics to measure perf. We don't need to collect all information. The comparable metrics is important. Other data can be found in log for troubleshooting.

lisa/tools/sysbench.py
smit-gardhariya: Done

lisa/messages.py
smit-gardhariya: Done

lisa/messages.py
smit-gardhariya: Removed it now

lisa/tools/sysbench.py
smit-gardhariya: Done now. I added it later again and still pushed it.

microsoft/testsuites/sysbench/sysbenchperf.py
smit-gardhariya: Done

lisa/messages.py
smit-gardhariya: If you look at the sysbench output it will always give latency metrics like min/max/avg. That was the reason to put it under common sysbench metrics

lisa/messages.py
smit-gardhariya: Will check on this

PR #3078, opened 2023-11-29
gpu extension: handle rhui issue before installing extension

LiliDeng:


lisa/operating_system.py
```diff
@@ -1666,37 +1666,40 @@ def capture_system_information(self, saved_path: Path) -> None:
             saved_path / "redhat-release.txt",
         )
 
-    @retry(tries=10, delay=5)
-    def _initialize_package_installation(self) -> None:
-        information = self._get_information()
+    def handle_rhui_issue(self) -> None:
+        # there are some images contain multiple rhui packages, like below:
```
squirrelsc, line 1670: Should it be fixed in the distro?

LiliDeng: the gpu driver extension should contain this fix, I will try to contact with the team

squirrelsc:
@LiliDeng LGTM

PR #3080, opened 2023-11-30
verify_services_state: exactly match the error message

LiliDeng:


squirrelsc (approved):

microsoft/testsuites/network/sriov.py
```diff
@@ -80,6 +100,16 @@ def verify_services_state(self, node: Node) -> None:
                 lambda: node.tools[Systemctl].state() == "running",
                 timeout_message="wait for systemctl status to be running",
             )
+        except LisaTimeoutException:
+            udevd_status = node.tools[Journalctl].logs_for_unit("systemd-udevd")
+            matched = self._device_rename_pattern.search(udevd_status)
+            if matched:
+                raise LisaException(
+                    f"found {matched[0]} message "
```
squirrelsc, line 108: The first line may not be a complete sentence, with "found" and "message". So just include the raw error message, like below,

```python
                    f"{matched[0]}. "
                    "There is a race condition when rename VF nics, "
                    "it causes boot delay, it should be fixed in "
                    "systemd - 245.4-4ubuntu3.21"

LiliDeng: Fixed in #3098

PR #3084, opened 2023-12-01
Support environment_connected phase for kernel_installer

pupacha:


lisa/transformers/kernel_installer.py
```diff
@@ -121,14 +120,13 @@ def _output_names(self) -> List[str]:
...
         runbook: KernelInstallerTransformerSchema = self.runbook
-        assert runbook.connection, "connection must be defined."
         assert runbook.installer, "installer must be defined."
 
         message = KernelBuildMessage()
         build_sucess: bool = False
         boot_success: bool = False
 
-        node = quick_connect(runbook.connection, "installer_node")
+        node = self._node
```
squirrelsc, line 129: Please make sure the installer still works with the original way.

pupacha: Yes, I verified installer works the original way with rpm kernel installer. @squirrelsc

squirrelsc (approved):

PR #3087, opened 2023-12-04
Enable TDX stateless deployment

kamalca:
TDX differs from SEV-SNP CVM because it can have encryption type 'NonPersistedTPM'. This is called 'Stateless' TDX, and it also requires secure boot to be set to false, meanwhile regular CVM and Stateful TDX require secure boot to be set to true.


lisa/features/security_profile.py
```diff
@@ -64,7 +68,7 @@ def __hash__(self) -> int:
         return hash(self._get_key())
 
     def _get_key(self) -> str:
-        return f"{self.type}/{self.security_profile}"
+        return f"{self.type}/{self.security_profile}" f"/{self.encrypt_disk}"
```
LiliDeng: f"{self.type}/{self.security_profile}/{self.encrypt_disk}"

LiliDeng (approved):

PR #3089, opened 2023-12-05
XDP Tool: Use v1.2.0 for Ubuntu1804

adityagesh:
v1.4.1 requires clang-11. Ubuntu 1804 has clang-10.

### DIFF
--- a/microsoft/testsuites/xdp/xdptools.py
+++ b/microsoft/testsuites/xdp/xdptools.py
@@ -41,7 +41,6 @@ class XdpTool(Tool):
     """
 
     _xdp_tools_repo = "https://github.com/xdp-project/xdp-tools.git"
-    _xdp_tools_tag = "v1.4.1"
     _default_command = "xdptool"
 
     #     [test_ether_deny]             PASS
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/xdp/xdptools.py
```diff
@@ -82,6 +81,13 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         super()._initialize(*args, **kwargs)
         self._command: PurePath = PurePath(self._default_command)
         self._gro_lro_settings: Dict[str, DeviceGroLroSettings] = {}
+        # v1.4.1 requires clang-11
+        self._xdp_tools_tag = "v1.4.1"
```
squirrelsc, line 85: use the version class to parse and compatible with higher versions.

adityagesh: @squirrelsc , sorry I did not understand your comment.
Are you suggesting to use Version class for the git tag?

or are you referring to this comparison `self.node.os.information.version == "18.4.0"`

squirrelsc: Oh, yes, I misunderstood it. The logic on os version should be like 'self.node.os.information.version <= "18.4.0"`

adityagesh: The reason for not using <= is because during the dependency installation, the below code already exists

```
if self.node.os.information.version < "18.4.0":
                raise UnsupportedDistroException(self.node.os)
```

Updated

squirrelsc: If so, you can add comments to explain why only one version specified here. Comments is to help others understand.

squirrelsc (approved):

PR #3090, opened 2023-12-05
Drop newer/older DPDK nff-go support

mcgov:
Drop support for debian 10 in the nff-go test. It's EOL and I don't plan to support it going forward.


squirrelsc (dismissed):

squirrelsc (approved):
It's a low priority test case for regression testing, only supported on specific distro versions. We may retire it when no distro supported.

PR #3092, opened 2023-12-05
Add support for PATCH and UPDATE commands on job templates in AITL wrapper

frmunozp:
This PR extends support for the jobTemplates resource type in the AITL wrapper. The following changes are made:

- For job templates, PATCH and PUT (updates) commands are accepted. PATCH commands will modify only the specific property listed in the request body, while PUT commands either create a new job template or replace the existing one. 
- The help menu will now show job template examples if it is requested for job templates.
- A new "jobTemplates" template has been added. It includes to Selections object to demonstrate how to add different test cases, priorities, or areas in a single template. This has come up in the past and the goal is that the example will help provide clarity.

### DIFF
--- a/microsoft/utils/aitl/aitl.py
+++ b/microsoft/utils/aitl/aitl.py
@@ -20,17 +20,19 @@ _fmt = "%(asctime)s.%(msecs)03d[%(thread)d][%(levelname)s] %(name)s %(message)s"
 _datefmt = "%Y-%m-%d %H:%M:%S"
 _api_version = "2023-08-01-preview"
 
-_examples = """
-Examples:
-    Create a job:
-        python -m aitl job create -s {subscription_id} -r {resource_group} -n {job_name} -b @./tier0.json
 
-    List jobs:
-        python -m aitl job list -s {subscription_id} -r {resource_group}
+def _generate_example(resource_type: str = "job") -> str:
+    return f"""
+        Examples for {resource_type.capitalize() if resource_type == 'job' else 'Job Template' }:
+        Create a {resource_type}:
+            python -m aitl {resource_type} create -s {{subscription_id}} -r {{resource_group}} -n {{template_name}} -b {'@./tier0.json' if resource_type == 'job' else '@./template.json'}
 
-    Get a job:
-        python -m aitl job get -s {subscription_id} -r {resource_group} -n {job_name}
-"""  # noqa: E501
+        List {resource_type}s:
+            python -m aitl {resource_type} list -s {{subscription_id}} -r {{resource_group}}
+
+        Get a {resource_type}:
+            python -m aitl {resource_type} get -s {{subscription_id}} -r {{resource_group}} -n {{template_name}}
+    """  # noqa: E501
 
 
 def _initialize() -> None:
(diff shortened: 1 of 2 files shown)


microsoft/utils/aitl/template.json
```diff
@@ -0,0 +1,39 @@
...
+                    0
+                ],
+                "caseTags": [],
+                "action": "include",
+                "testIterations": 1,
+                "retryCount": 0,
+                "useNewEnvironment": false,
+                "ignoreFailure": false
+            },
+            {
```
squirrelsc: Remove this part, it's confusing. One item is enough.

microsoft/utils/aitl/aitl.py
```diff
@@ -79,19 +93,19 @@ def _add_resource_parser(
...
 ) -> None:
     cmd_parser: ArgumentParser = parser.add_parser(
-        name=resource, epilog=_examples, formatter_class=RawTextHelpFormatter
+        name=resource, epilog=_generate_example(resource), formatter_class=RawTextHelpFormatter
     )
     sub_parser = cmd_parser.add_subparsers(dest="action", required=True)
-    for action in ["create", "list", "get", "delete", "update"]:
-        if not support_update and action == "update":
+    for action in ["create", "list", "get", "delete", "update", "patch"]:
+        if not support_update and (action == "update" or action == "patch"):
```
squirrelsc: action in ["update", "patch"]

microsoft/utils/aitl/aitl.py
```diff
@@ -283,10 +297,10 @@ def _process_create_job(**kwargs: Any) -> Any:
     action = kwargs.pop("action")
     resource = kwargs.get("resource")
 
-    if action == "create":
+    if action == "create" or action == "update":
```
squirrelsc: `action in ["create", "update"]`

microsoft/utils/aitl/template.json
frmunozp: I added two because I get many questions from partners on how to add several different type of test parameters (for example, a case priority and a case name) in a single request. This is meant to serve as an example for our partners.

squirrelsc: It's an array, so it's enough. It may create more question, if adding a specified test case name in another item or adding two identical items...

frmunozp: I updated it to just one selection object.

microsoft/utils/aitl/aitl.py
frmunozp: Updated!

microsoft/utils/aitl/aitl.py
frmunozp: Updated

microsoft/utils/aitl/aitl.py
```diff
@@ -103,7 +117,7 @@ def _add_resource_parser(
             action_parser, support_name=support_name, required_name=required_name
         )
 
-        if resource in {"job", "template"} and action == "create":
+        if resource in {"job", "template"} and action in {"create", "update", "patch"}:
```
squirrelsc: It should use`[ ]` as an array. `{ }` is used for dict, I'm not sure what type it is here.

squirrelsc: Please also fix build errors.

squirrelsc:
Use `git rebase -i` to merge commits into one. So we can "rebase and merge", instead of "squash and merge".

squirrelsc (approved):

PR #3094, opened 2023-12-06
PythonVenv tool

adityagesh:


lisa/tools/python.py
```diff
@@ -89,3 +90,66 @@ def exists_package(self, package_name: str) -> bool:
...
+    def get_venv_path(self) -> PurePath:
+        if not hasattr(self, "_venv_path"):
+            self._venv_path = self.create_venv()
+        return self._venv_path
+
+    def create_venv(self, venv_path: str = "", sudo: bool = False) -> PurePath:
+        _venv_path: PurePath = (
+            PurePath(venv_path) if venv_path else self.node.working_path / "venv"
+        )
+        cmd_result = super().run(
```
adityagesh: @squirrelsc @LiliDeng  super() does not give 'Python' Tool. Is this expected behavior? 

image

squirrelsc: Not sure why it doesn't work, but it should be. In this case, the `run()` is not overrides, so you can call `self.run()` too.

adityagesh: self.run uses 'command' property from PythonVenv class in 'run_async'. Here I have used to run a command from Python class.
Similarly using 'super().command' explicitly as well.

Should I replace the usage with node.execute('python3 ....')?

lisa/tools/python.py
```diff
@@ -89,3 +90,66 @@ def exists_package(self, package_name: str) -> bool:
...
+        return [Python]
+
+    # if get_venv_path is invoked before create_venv, it will create
+    # a venv in the node working path
+    def get_venv_path(self) -> PurePath:
+        if not hasattr(self, "_venv_path"):
+            self._venv_path = self.create_venv()
+        return self._venv_path
+
+    def create_venv(self, venv_path: str = "", sudo: bool = False) -> PurePath:
```
adityagesh: @squirrelsc  Should virtual env be created in _initialize()? But this would mean it user cannot pass the installation path

lisa/tools/python.py
squirrelsc: I think `super().run` may not take the command from super class. Because the venv class does not inherit much logic from base class. So it doesn't need to inherit from Python. It can use a python tool inside.

lisa/tools/python.py
squirrelsc: Why do we need venv, instead of direct install in the current Python?

adityagesh: Observed this error in GPU test suite for Ubuntu 2310

2023-12-11 03:50:58.204[139645954287168][DEBUG] lisa.env[generated_0].node[0].cmd[7803] cmd: ['pip3', 'install', '-q', 'torch'], cwd: None, shell: False, sudo: False, nohup: False, posix: True, remote: True, encoding: utf-8
2023-12-11 03:50:59.501[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout error: externally-managed-environment
2023-12-11 03:50:59.501[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout × This environment is externally managed
2023-12-11 03:50:59.502[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout ╰─> To install Python packages system-wide, try apt install
2023-12-11 03:50:59.503[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     python3-xyz, where xyz is the package you are trying to
2023-12-11 03:50:59.503[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     install.
2023-12-11 03:50:59.505[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     If you wish to install a non-Debian-packaged Python package,
2023-12-11 03:50:59.507[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     create a virtual environment using python3 -m venv path/to/venv.
2023-12-11 03:50:59.507[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
2023-12-11 03:50:59.508[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     sure you have python3-full installed.
2023-12-11 03:50:59.509[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     If you wish to install a non-Debian packaged Python application,
2023-12-11 03:50:59.510[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     it may be easiest to use pipx install xyz, which will manage a
2023-12-11 03:50:59.511[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     virtual environment for you. Make sure you have pipx installed.
2023-12-11 03:50:59.512[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout     See /usr/share/doc/python3.11/README.venv for more information.
2023-12-11 03:50:59.514[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
2023-12-11 03:50:59.515[139646265976384][DEBUG] lisa.env[generated_0].node[0].cmd[7803].stdout hint: See PEP 668 for the detailed specification.

adityagesh: I thought it is better to have cases install packages in venv to avoid conflicts. We can also remove the packages after the testcase to clean up space.

squirrelsc: You can pass in parameters by node.tools.create. It will create a new tool instance every time, so the venv doesn't conflict.

adityagesh: Sorry, I did not understand. Are you suggesting we can do that using current python tool?
Or specifying path to this venv tool in the PR?

squirrelsc: You can do it when creating the venv tool. So you can pass an argument, and receiving it when calling `__init__`.

adityagesh: @squirrelsc , I have made the changes, can you please review if this is the expected flow?

Testing is in progress.

lisa/tools/python.py
```diff
@@ -89,3 +94,79 @@ def exists_package(self, package_name: str) -> bool:
...
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Python]
+
+    def _create_venv(self, venv_path: str = "", sudo: bool = False) -> PurePath:
```
squirrelsc: To make sure no side effects among test cases, make `venv_path` as a required argument. So the path of test case should be passed in. Or if some test cases share the same env, they can use the same string. But the `"venv"` is not safe as a default.

squirrelsc: When the `sudo` is needed? The venv should be a user path always.

lisa/tools/python.py
```diff
@@ -89,3 +94,79 @@ def exists_package(self, package_name: str) -> bool:
...
+        )
+        assert_that(
+            cmd_result.exit_code, f"fail to create venv: {_venv_path}"
+        ).is_equal_to(0)
+        self._venv_path = _venv_path
+        return self._venv_path
+
+    def __init__(self, node: "Node", venv_path: str = "") -> None:
+        super().__init__(node)
+        self.python: Python = self.node.tools[Python]
```
squirrelsc: use `self._python`, because it's a private variable.

lisa/tools/python.py
```diff
@@ -89,3 +94,79 @@ def exists_package(self, package_name: str) -> bool:
...
+        return self._check_exists()
+
+    # if get_venv_path is invoked before create_venv, it will create
+    # a venv in the node working path
+    def get_venv_path(self) -> PurePath:
+        if not hasattr(self, "_venv_path"):
+            self._venv_path = self._create_venv(self._venv_installation_path)
+        return self._venv_path
+
+    def install_packages(self, packages_name: str, sudo: bool = False) -> None:
```
squirrelsc: sudo is not necessary. venv is to avoid install in root permission.

lisa/tools/python.py
```diff
@@ -89,3 +94,79 @@ def exists_package(self, package_name: str) -> bool:
...
+            f"-m pip uninstall {package_name} -y", force_run=True, sudo=sudo
+        )
+        return result.exit_code == 0
+
+    def delete_venv(self) -> None:
+        if hasattr(self, "_venv_path"):
+            self.node.execute(f"rm -rf {self._venv_path}")
+            delattr(self, "_venv_path")
+        else:
+            self._log.warning("venv path not found, nothing to delete")
```
squirrelsc: use `info` level. Avoid using warn,

adityagesh: updated

lisa/tools/python.py
adityagesh: removed all usages of sudo

lisa/tools/python.py
adityagesh: updated

lisa/tools/python.py
adityagesh: `_venv_path: PurePath = (
            PurePath(venv_path)
            if venv_path
            else self.node.working_path / datetime.utcnow().isoformat()
)`

Will this work? Cases that want to reuse the env can pass venv_path, otherwise the default would always be unique.

adityagesh:
Tested in canonical 0001-com-ubuntu-server-focal 20_04-lts-gen2 latest, redhat rhel 8-lvm-gen2 8.1.20200318, canonical 0001-com-ubuntu-server-mantic 23_10-gen2 latest,  microsoftcblmariner cbl-mariner cbl-mariner-2 2.20230924.01

lisa/tools/python.py
squirrelsc: The timestamp is not an idempotent way. If some cases want to share a path, you can hardcode a path name in a common logic of those test cases.

adityagesh: Removed default path

squirrelsc:
LGTM. Please mark it to ready, and ask Lili to review and merge.

adityagesh:
@LiliDeng , PR is ready for review

adityagesh:
@LiliDeng , rebased

LiliDeng:
> @LiliDeng , rebased

I ran the case `verify_gpu_cuda_with_pytorch` against `canonical 0001-com-ubuntu-server-jammy 22_04-lts latest`, I encountered below error.
The error is `not found command: No such command: homel****tlisa_working2024022720240227-160635-872gpu_pytorch/bin/python3`

lisa/tools/python.py
```diff
@@ -89,3 +94,74 @@ def exists_package(self, package_name: str) -> bool:
...
+        return (
+            venv.exit_code == 0 and "No module named ensurepip" not in ensurepip.stdout
+        )
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Posix):
+            self.node.os.install_packages("python3-venv")
+        return self._check_exists()
+
+    def get_venv_path(self) -> PurePath:
```
squirrelsc, line 123: Move public methods above private methods.

lisa/tools/python.py
```diff
@@ -89,3 +94,74 @@ def exists_package(self, package_name: str) -> bool:
...
+        return [Python]
+
+    def _create_venv(self, venv_path: str) -> PurePath:
+        cmd_result = self._python.run(
+            f"-m venv {venv_path}", force_run=True, shell=True
+        )
+        assert_that(
+            cmd_result.exit_code, f"fail to create venv: {venv_path}"
+        ).is_equal_to(0)
+        self._venv_path = PurePath(venv_path)
```
squirrelsc: Not sure if it works for both Linux and Windows. Use node.get_pure_path to get right version purepath of the node.

lisa/tools/python.py
```diff
@@ -89,3 +94,74 @@ def exists_package(self, package_name: str) -> bool:
     def uninstall_package(self, package_name: str) -> bool:
         result = self.run(f"uninstall {package_name} -y", force_run=True, sudo=True)
         return result.exit_code == 0
+
+
+class PythonVenv(Tool):
+    @property
+    def command(self) -> str:
+        path = self.get_venv_path().joinpath(f"bin/{self._python.command}")
```
squirrelsc: Because get_venv_path returns a purepath already, and it's for this kind of scenario.

```python
path = self.get_venv_path() / "bin" / self._python.command
return str(path)
```

lisa/tools/python.py
adityagesh: updated

lisa/tools/python.py
adityagesh: updated

squirrelsc:
@LiliDeng LGTM

adityagesh:
@squirrelsc added one more commit to clean package cache.

Redhat VMs have less storage in root partition. Noticed the GPU Pytorch failing while pip is downloading packages.
Should I cherry pick this commit to another PR?

microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -271,6 +280,9 @@ def verify_gpu_cuda_with_pytorch(
         pythonvenv_path = work_path + "/gpu_pytorch"
         pythonvenv = node.tools.create(PythonVenv, venv_path=pythonvenv_path)
 
+        if isinstance(node.os, Linux):
```
squirrelsc, line 285: Add comments to explain. If the os disk is not enough, you can use resource disk to save files.

squirrelsc:
> @squirrelsc added one more commit to clean package cache.
> 
> Redhat VMs have less storage in root partition. Noticed the GPU Pytorch failing while pip is downloading packages. Should I cherry pick this commit to another PR?

It's not related to this PR, it should be a separated one.

microsoft/testsuites/gpu/gpusuite.py
adityagesh: The disk full issue happens while the .whl package is being downloading by pip (the tmp files being stored in root disk).

Facing the issue even after specifying cace dir in pip

 envs = {"TMPDIR": str(cache_dir)}
        cmd_result = self.run(
            f"-m pip install -q {packages_name} --cache-dir={cache_dir}",
            force_run=True,
            update_envs=envs,
        )

squirrelsc: Thank you for the explanation, please add the comment in the code. So we will know why it needs to clean in future.

adityagesh: Added comment, raised package cleanup as a separate PR 
https://github.com/microsoft/lisa/pull/3200

squirrelsc (approved):

PR #3095, opened 2023-12-07
ntttcp: move install lagscope out of try block

LiliDeng:


squirrelsc:
What's the reason to move it? Please add in the commit message.

LiliDeng:
remake the qcow2 image

PR #3097, opened 2023-12-07
register preview feature

LiliDeng:


squirrelsc:
The feature may not be enabled automatically, we can just explain in the error message. Provide instructions to enable a feature, and let users to contact support team or their Microsoft contact to enable it. So it just needs to update error message in `azure/hooks.py`.

LiliDeng:
> The feature may not be enabled automatically, we can just explain in the error message. Provide instructions to enable a feature, and let users to contact support team or their Microsoft contact to enable it. So it just needs to update error message in `azure/hooks.py`.

so just give a message, not register through lisa, I address your comments in https://github.com/microsoft/lisa/pull/3099, if my understanding is correct, then I will close this PR.

PR #3099, opened 2023-12-08
skip case when preview feature is not enabled

LiliDeng:


LiliDeng:
@squirrelsc there is one concern that, whether the original exception is HttpResponseError type.

squirrelsc:
> @squirrelsc there is one concern that, whether the original exception is HttpResponseError type.

Did you have a chance to repro it? How about the validation service subscription?

squirrelsc (approved):

LiliDeng:
> > @squirrelsc there is one concern that, whether the original exception is HttpResponseError type.
> 
> Did you have a chance to repro it? How about the validation service subscription?

I have tried with dev and test, but unregister the feature doesn't take effect, let me try with your suggestion. thanks.

LiliDeng:
> > @squirrelsc there is one concern that, whether the original exception is HttpResponseError type.
> 
> Did you have a chance to repro it? How about the validation service subscription?

I have tested it.

PR #3102, opened 2023-12-08
temp fix - skip test when fail to retrieve user object ID

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -2074,7 +2075,7 @@ def get_identity_id(
     response = requests.get(graph_api_url, headers=headers, timeout=10)
 
     if response.status_code != 200:
-        raise LisaException(
+        raise SkippedException(
```
squirrelsc, line 2078: Please check the error message, make sure it skips with known issue only.

PR #3103, opened 2023-12-11
get object id when use identity in vm level

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -150,6 +150,8 @@
     "image_version",
 ]
 
+METADATA_ENDPOINT = "http://169.254.169.254/metadata/instance?api-version=2021-02-01"
```
squirrelsc, line 154: What's is this address for? How to make sure the IP works always?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2052,9 +2054,74 @@ def get_tenant_id(credential: Any) -> Any:
     return subscription.tenant_id
 
 
+def get_metadata() -> Any:
```
squirrelsc: The name is ambiguous, because it's a function in the common file. What "metadata" is it for?

lisa/sut_orchestrator/azure/common.py
LiliDeng: IMDS is a REST API that's available at a well-known, non-routable IP address (169.254.169.254).

lisa/sut_orchestrator/azure/common.py
LiliDeng: it is Azure VM metadata

lisa/sut_orchestrator/azure/common.py
squirrelsc: Please add code comments above the line to explain it.

lisa/sut_orchestrator/azure/common.py
squirrelsc: Rename the method to be clearer.

lisa/sut_orchestrator/azure/common.py
LiliDeng: done

lisa/sut_orchestrator/azure/common.py
LiliDeng: done

squirrelsc (approved):

PR #3104, opened 2023-12-11
Hwclock tool: Try installing from util-linux-extra

adityagesh:
In canonical 0001-com-ubuntu-server-mantic hwclock binary is present in package 'util-linux-extra'.

### DIFF
--- a/lisa/tools/hwclock.py
+++ b/lisa/tools/hwclock.py
@@ -24,6 +24,9 @@ class Hwclock(Tool):
         posix_os: Posix = cast(Posix, self.node.os)
         package_name = "util-linux"
         posix_os.install_packages(package_name)
+        if not self._check_exists():
+            package_name = "util-linux-extra"
+            posix_os.install_packages(package_name)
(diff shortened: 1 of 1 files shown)


lisa/tools/hwclock.py
```diff
@@ -22,8 +22,13 @@ def can_install(self) -> bool:
...
         posix_os: Posix = cast(Posix, self.node.os)
-        package_name = "util-linux"
-        posix_os.install_packages(package_name)
+        try:
+            package_name = "util-linux"
+            posix_os.install_packages(package_name)
+            self._check_exists()
+        except LisaException:
+            package_name = "util-linux-extra"
+            posix_os.install_packages(package_name)
```
LiliDeng, line 29: package_name = "util-linux"
            posix_os.install_packages(package_name)
            if not self._check_exists():
                package_name = "util-linux-extra"
                posix_os.install_packages(package_name)

LiliDeng (approved):

PR #3105, opened 2023-12-12
Add cached_disk_size in Azure disk settings for Ephemeral os disk type

lubaihua33:
Ephemeral disk type requires the image OS disk’s size to be less than or equal to the temp/cache size of the VM size chosen.
https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks#size-requirements

Add cached_disk_size in the Azure disk setting so Lisa can choose the right VM size for verify_deployment_provision_ephemeral_managed_disk test case.

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -454,6 +454,12 @@ class DiskOptionSettings(FeatureSettings):
             decoder=partial(search_space.decode_set_space_by_type, base_type=DiskType)
         ),
     )
+    os_disk_size: search_space.CountSpace = field(
+        default_factory=partial(search_space.IntRange, min=0),
+        metadata=field_metadata(
+            allow_none=True, decoder=search_space.decode_count_space
+        ),
+    )
     data_disk_type: Optional[
         Union[search_space.SetSpace[DiskType], DiskType]
     ] = field(  # type:ignore
@@ -538,6 +544,7 @@ class DiskOptionSettings(FeatureSettings):
         return (
             self.type == o.type
             and self.os_disk_type == o.os_disk_type
+            and self.os_disk_size == o.os_disk_size
             and self.data_disk_type == o.data_disk_type
             and self.data_disk_count == o.data_disk_count
             and self.data_disk_caching_type == o.data_disk_caching_type
@@ -551,6 +558,7 @@ class DiskOptionSettings(FeatureSettings):
     def __repr__(self) -> str:
         return (
             f"os_disk_type: {self.os_disk_type},"
+            f"os_disk_size: {self.os_disk_size},"
             f"data_disk_type: {self.data_disk_type},"
             f"count: {self.data_disk_count},"
             f"caching: {self.data_disk_caching_type},"
@@ -598,10 +606,11 @@ class DiskOptionSettings(FeatureSettings):
 
     def _get_key(self) -> str:
         return (
-            f"{super()._get_key()}/{self.os_disk_type}/{self.data_disk_type}/"
-            f"{self.data_disk_count}/{self.data_disk_caching_type}/"
-            f"{self.data_disk_iops}/{self.data_disk_throughput}/"
-            f"{self.data_disk_size}/{self.disk_controller_type}"
+            f"{super()._get_key()}/{self.os_disk_type}/{self.os_disk_size}/"
+            f"{self.data_disk_type}/{self.data_disk_count}/"
+            f"{self.data_disk_caching_type}/{self.data_disk_iops}/"
+            f"{self.data_disk_throughput}/{self.data_disk_size}/"
+            f"{self.disk_controller_type}"
         )
 
     def _call_requirement_method(
@@ -621,6 +630,10 @@ class DiskOptionSettings(FeatureSettings):
             value.os_disk_type = getattr(
                 search_space, f"{method.value}_setspace_by_priority"
             )(self.os_disk_type, capability.os_disk_type, disk_type_priority)
+        if self.os_disk_size or capability.os_disk_size:
+            value.os_disk_size = search_space_countspace_method(
+                self.os_disk_size, capability.os_disk_size
+            )
         if self.data_disk_type or capability.data_disk_type:
             value.data_disk_type = getattr(
                 search_space, f"{method.value}_setspace_by_priority"
--- a/lisa/search_space.py
+++ b/lisa/search_space.py
@@ -328,6 +328,9 @@ class SetSpace(RequirementMixin, Set[T]):
         super().remove(element)
         self.items.remove(element)
 
+    def isunique(self, element: T) -> bool:
+        return len(self.items) == 1 and self.items[0] == element
+
     def update(self, *s: Iterable[T]) -> None:
         super().update(*s)
         self.items.extend(*s)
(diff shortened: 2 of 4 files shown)


LiliDeng (dismissed):

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1792,9 +1794,10 @@ def _resource_sku_to_capability(  # noqa: C901
             # Check if CachedDiskBytes is greater than 30GB
             # We use diff disk as cache disk for ephemeral OS disk
             cached_disk_bytes = azure_raw_capabilities.get("CachedDiskBytes", 0)
-            cached_disk_bytes_gb = int(cached_disk_bytes) / 1024 / 1024 / 1024
+            cached_disk_bytes_gb = int(int(cached_disk_bytes) / 1024 / 1024 / 1024)
```
squirrelsc, line 1797:
```suggestion
            cached_disk_bytes_gb = int(cached_disk_bytes / 1024 / 1024 / 1024)
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2729,10 +2756,29 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
                 features.ArchitectureSettings(arch=sig.architecture)  # type: ignore
             )
+            if (
+                node_space.disk
+                and node_space.disk.os_disk_type
+                and schema.DiskType.Ephemeral in node_space.disk.os_disk_type
+            ):
+                assert isinstance(node_space.disk, features.AzureDiskOptionSettings)
+                node_space.disk.cached_disk_size = search_space.IntRange(
+                    min=self._get_sig_os_disk_size(azure_runbook.shared_gallery)
```
squirrelsc: You can use a method called `get_os_disk_size` to call the three methods by different disk type. So the check logic doesn't need to be repeated three times here.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1206,6 +1206,12 @@ def get_all_nics_ip_info(self) -> List[IpInfo]:
 @dataclass()
 class AzureDiskOptionSettings(schema.DiskOptionSettings):
     has_resource_disk: Optional[bool] = None
+    cached_disk_size: search_space.CountSpace = field(
```
squirrelsc: Can it be `os_disk_size`, and put it in common DiskOptionSettings?

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: cached_disk_bytes  is a str. the change like this has an error.

lisa/sut_orchestrator/azure/platform_.py

lisa/search_space.py
```diff
@@ -328,6 +328,9 @@ def remove(self, element: T) -> None:
         super().remove(element)
         self.items.remove(element)
 
+    def isunique(self, element: T) -> bool:
```
squirrelsc, line 331: if it's single item, it should follow other examples. Change the Setspace to the single item, instead of comparing it with a complex way.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2736,6 +2750,47 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
         else:
             ...
 
+        if (
+            node_space.disk
```
squirrelsc: `node_space.disk` can be asserted. So the if logic is simpler.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2736,6 +2750,47 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
+            and node_space.disk.os_disk_type
```
squirrelsc: It should be able to assert too.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2736,6 +2750,47 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
+                )
+            )
+        ):
+            assert isinstance(node_space.disk, features.AzureDiskOptionSettings)
+            node_space.disk.cached_disk_size = search_space.IntRange(
+                min=self._get_os_disk_size(azure_runbook)
+            )
+
+    def _get_os_disk_size(self, azure_runbook: AzureNodeSchema) -> int:
+        if not azure_runbook:
```
squirrelsc: Add comments above, why returns zero.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2736,6 +2750,47 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
...
+            if image_info and image_info.os_disk_image:
+                return _get_disk_size_in_gb(
+                    image_info.os_disk_image.additional_properties
+                )
+        elif azure_runbook.shared_gallery:
+            azure_runbook.shared_gallery = self._parse_shared_gallery_image(
+                azure_runbook.shared_gallery
+            )
+            return self._get_sig_os_disk_size(azure_runbook.shared_gallery)
+        elif azure_runbook.vhd and azure_runbook.vhd.vhd_path:
```
squirrelsc: it should be `else` and assert the vhd.

lisa/sut_orchestrator/azure/features.py
lubaihua33: os_disk_size might be not accurate in describing the cached disk size which is related to VM size.

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: If asserted node_space.disk, most of the unit tests will fail because of node_space.disk assert error.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/platform_.py
lubaihua33: Have added some changes. default value 30 is returned now. Have added the comments.

lisa/sut_orchestrator/azure/platform_.py

lisa/sut_orchestrator/azure/features.py
squirrelsc: `os_disk_size` is a requirement on the disk settings. The cached disk size is calculated in Azure, based on os_disk_size.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2736,6 +2750,44 @@ def _add_image_features(self, node_space: schema.NodeSpace) -> None:
         else:
             ...
```
squirrelsc: Please help remove the useless `else: ...`

lisa/schema.py
```diff
@@ -454,6 +454,12 @@ class DiskOptionSettings(FeatureSettings):
             decoder=partial(search_space.decode_set_space_by_type, base_type=DiskType)
         ),
     )
+    os_disk_size: search_space.CountSpace = field(
```
squirrelsc, line 457: Make sure `_call_requirement_method`, `__eq__`, `__repr__`, `__get_key` proceeded the new field.

lubaihua33: Done

lisa/sut_orchestrator/azure/features.py
lubaihua33: Done

squirrelsc (approved):

PR #3107, opened 2023-12-13
skip case if 'Public access is not permitted on this storage account'

LiliDeng:


microsoft/testsuites/vm_extensions/runtime_extensions/common.py
```diff
@@ -108,9 +109,20 @@ def retrieve_storage_blob_url(
...
+            try:
+                container_client.set_container_access_policy(
+                    signed_identifiers={}, public_access="container"
+                )
+            except ResourceExistsError as ex:
+                if (
+                    "public access is not permitted on this storage account"
+                    in str(ex).lower()
+                ):
+                    raise SkippedException(
```
squirrelsc, line 121: Please bring the raw exception together.

microsoft/testsuites/vm_extensions/runtime_extensions/common.py
```diff
@@ -108,9 +109,20 @@ def retrieve_storage_blob_url(
...
+                        "Public access is not permitted on this storage account "
+                        f"{storage_account_name}, exception is '{ex}'"
```
squirrelsc: Don't use quotes on previous exception message, it may not be good on format.

```suggestion
                        f"{storage_account_name}. {ex}"
```

LiliDeng: updated

squirrelsc (approved):

PR #3109, opened 2023-12-17
skip case when specified vm size can't offer enough data disks

LiliDeng:


squirrelsc (dismissed):

squirrelsc:
Please also output debug log, if an image need more disks.

squirrelsc (approved):

PR #3110, opened 2023-12-20
DPDK: Refactor rdma-core source build, allow arbitrary package build

mcgov:
Fulfilling an ask to be able to test multiple rdma-core versions to allow development and regression tests. Refactors the rdma-core bits into their own own class.

Setting rdma_core_source to either a git or tar.gz link allows you to build from source. For git either set rdma_core_ref or it will pick the latest version sorted tag.

### DIFF
--- a/lisa/util/__init__.py
+++ b/lisa/util/__init__.py
@@ -48,7 +48,7 @@ global_ssh_key_access_lock = Lock()
 # source -
 # https://github.com/django/django/blob/stable/1.3.x/django/core/validators.py#L45
 __url_pattern = re.compile(
-    r"^(?:http|s?ftp)s?://"  # http:// or https://
+    r"^(?:http|https|sftp|ftp)://"  # http:// or https://
     r"(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)"
     r"+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|"  # ...domain
     r"localhost|"  # localhost...
@@ -564,9 +564,7 @@ def find_group_in_lines(
     return result
 
 
-def deep_update_dict(
-    src: Dict[str, Any], dest: Optional[Dict[str, Any]]
-) -> Dict[str, Any]:
+def deep_update_dict(src: Dict[str, Any], dest: Dict[str, Any]) -> Dict[str, Any]:
     if (
         dest is None
         or isinstance(dest, int)
@@ -580,7 +578,7 @@ def deep_update_dict(
 
     if isinstance(result, dict):
         for key, value in src.items():
-            if dest and isinstance(value, dict) and key in dest:
+            if isinstance(value, dict) and key in dest:
                 value = deep_update_dict(value, dest[key])
             result[key] = value
     elif isinstance(src, dict):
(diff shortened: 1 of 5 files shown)


lisa/util/__init__.py
```diff
@@ -598,6 +601,58 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
     return is_url
 
 
+def is_valid_source_code_package(
```
squirrelsc: Why is it called "source code package"? What's examples? It looks check a valid url.

mcgov: The function checks for a valid url, expected protocol, expected package origin and expected file name + extension.
 
I did this for the following reasons:
- it's good to enforce https.
- It will help me or other users when we inevitably pass a url in the wrong spot. Getting an error message about the package name or source is better than some confusing message about `failed to make [wrong package name]` where you have to pull the full log and look at the stack trace to see what code called `make`.
- Prevent users from just passing any tarball or git repo to the rdma-core install code and have it 'make install' with sudo.
- Prevent passing an unexpected package type and getting a weird error message about failure to extract it.
 
After getting some feedback from new users I'm trying to intentionally raise more explicit error messages and check assumptions more. `Could not make /path/whatever ` is less useful than `Failed: expected a package like 'https://...rdma-core.vxx.tar.gz and received https://...dpdk.tar.gz`

squirrelsc: I mean the method is not only used for "source code packages". It can be used for other name patterns. But the method name is called `is_valid_source_code_package`, it's limited, and no examples, no comments, so others are hard to reuse the method.

lisa/util/__init__.py
```diff
@@ -598,6 +601,58 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
...
+) -> bool:
+    # avoid using a mutable default parameter
+    if not allowed_protocols:
+        allowed_protocols = [
+            "https",
+            "sftp",
+        ]
+    # first, check if it's a url.
+    if not is_valid_url(url=source_url, raise_error=False):
+        return False
```
squirrelsc, line 640: False is not easy to troubleshoot. You can add an argument called "raise_error" like other examples. If it's True, raise exceptions, otherwise log the exception and return False.

lisa/util/__init__.py
mcgov: Ah ok, so maybe like `check_url_filename`?

lisa/util/__init__.py
```diff
@@ -47,7 +48,7 @@
 # source -
 # https://github.com/django/django/blob/stable/1.3.x/django/core/validators.py#L45
 __url_pattern = re.compile(
-    r"^(?:http|ftp)s?://"  # http:// or https://
+    r"^(?:http|s?ftp)s?://"  # http:// or https://
```
squirrelsc: 1. Update the comments to make it consistent with the patterns.
2. Expand the patterns make it easy to read, and avoid invalid pattern like `sftps`, `ftps`. `(?:http|https|sftp|ftp)`

lisa/util/__init__.py
```diff
@@ -563,7 +564,9 @@ def find_group_in_lines(
     return result
 
 
-def deep_update_dict(src: Dict[str, Any], dest: Dict[str, Any]) -> Dict[str, Any]:
+def deep_update_dict(
+    src: Dict[str, Any], dest: Optional[Dict[str, Any]]
```
squirrelsc: Why make dest is none? If the dest is None, it doesn't need to update.

lisa/util/__init__.py
```diff
@@ -598,6 +601,98 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
     return is_url
 
 
+def _raise_or_log_failure(log: Logger, raise_error: bool, failure_msg: str) -> bool:
+    if raise_error:
+        raise LisaException(failure_msg)
+    else:
+        log.info(failure_msg)
```
squirrelsc: It should be debug level by default, because it's not worth to raise an exception. It means it's ignorable error.

lisa/util/__init__.py
```diff
@@ -598,6 +601,108 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
...
+# big function to check the parts of a url
+# allow raising exceptions or log and return a bool
+# allows checks for:
+#   expected domains
+#   protocols (require https, sftp, etc)
+#   filenames (pattern matching)
+def check_url(
+    log: Logger,
+    source_url: str,
+    expected_package_name_pattern: Optional[Pattern[str]] = None,
```
squirrelsc: `expected_path_pattern`

lisa/util/__init__.py
```diff
@@ -598,6 +601,108 @@ def is_valid_url(url: str, raise_error: bool = True) -> bool:
...
+    allowed_protocols: Optional[List[str]] = None,
+    expected_domains: Optional[List[str]] = None,
```
squirrelsc: `expected_domain_pattern`. One pattern should be enough to validate all.

lisa/util/__init__.py
squirrelsc: Either remove the comments or make it consistent with patterns.

lisa/util/__init__.py
mcgov: I want to emphasize it's just the filename portion, I'll make it expected_filename_pattern

lisa/util/__init__.py
mcgov: I started with just a regex match but it was a pain with annoying edge cases, so I started writing this version. My goal is making it easier for the caller, this way you can pass a couple of domains, a couple of protocols, and a filename pattern (or just a string) and restrict future users from downloading whatever tarball from wherever later with little room for error.

mcgov: long story short I'm writing it for me so I don't mess it up next time I go to do this 😅

lisa/util/__init__.py
squirrelsc: If it's a regexp pattern, you can just check the end parts. If it'a name like `filename`, it limits usages to relative/full path validation.

lisa/util/__init__.py
squirrelsc: You can test regexp at someplace like https://regex101.com. if it's string comparison, it will meet more problems actually. Like how to allow/disallow subdomains, or attack like "invalidmicrosoft.com"?

lisa/util/__init__.py
mcgov: reverted to original code

lisa/util/__init__.py
mcgov: Good point!

mcgov: Honestly, wish I hadn't even tried to handle this at this point.

mcgov: okedokey, moved to a pattern for domains,

PR #3111, opened 2023-12-20
Update LTP to the newest 20230929 version

paxue:
Fix test bug 47596061:
LTP oom test cases will consume lots of memory, which makes SSH be terminated when out of memory.
https://microsoft.visualstudio.com/OS/_workitems/edit/47596061

### DIFF
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -50,7 +50,7 @@ class Ltp(Tool):
     _RESULT_LTP_ARCH_REGEX = re.compile(r"Machine Architecture: (.*)\s+")
 
     LTP_DIR_NAME = "ltp"
-    DEFAULT_LTP_TESTS_GIT_TAG = "20200930"
+    DEFAULT_LTP_TESTS_GIT_TAG = "20230929"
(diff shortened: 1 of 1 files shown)


LiliDeng:
I have tested the change on Azure platform. Let know whether it has the impact for other platforms. @squirrelsc

squirrelsc:
We can fix it by upgrading the LTP version to latest. The test case will pass.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -40,6 +40,7 @@ class LtpTestsuite(TestSuite):
         timeout=_TIME_OUT,
         requirement=simple_requirement(
             min_core_count=8,
+            min_memory_mb=28 * 1024,
```
squirrelsc: Remove this requirement, since it's not needed now.

squirrelsc: It may cause other problems, like failed on small disk sizes.

paxue: removed

squirrelsc (approved):

squirrelsc:
Update PR title to make it accurate.

PR #3112, opened 2023-12-20
'purchase_plan' support for vhd images.

SRIKKANTH:
'purchase plan' is needed for vhds created using marketplace images with purchase plans.
This change allows purchase plans passed through runbooks for vhds.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -260,6 +260,9 @@ class AzureNodeSchema:
     # Required by shared gallery images which are present in
     # subscription different from where LISA is run
     subscription_id: str = ""
+    purchase_plan_raw: Optional[Union[Dict[Any, Any], str]] = field(
+        default=None, metadata=field_metadata(data_key="purchase_plan")
+    )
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -561,6 +561,11 @@ def from_node_runbook(cls, runbook: AzureNodeSchema) -> "AzureNodeArmParameter":
         if "vhd" in parameters:
             parameters["vhd_raw"] = parameters["vhd"]
             del parameters["vhd"]
+        if "purchase_plan" in parameters:
```
squirrelsc: It should follow other fields, to be stripped in `AzureNodeSchema.__post_init__`.

SRIKKANTH: @squirrelsc I have modified the PR to support vhds with purchase plans. Please review.

squirrelsc:
@LiliDeng LGTM. In future, the purchaseplan should be moved to AzureImageSchema.

LiliDeng (approved):

PR #3114, opened 2023-12-21
Increase sleep time in stress_mshv_vm_create test

pupacha:
CVM guest takes longer time to boot. 20 seconds is the empirical boot time for a guest with 1G memory. Also, CVM guest process killing takes slightly longer time. Hence, increase the sleep timeout accordingly.

### DIFF
--- a/microsoft/testsuites/mshv/mshv_root_stress_tests.py
+++ b/microsoft/testsuites/mshv/mshv_root_stress_tests.py
@@ -186,7 +186,8 @@ class MshvHostStressTestSuite(TestSuite):
             sleep_time = 10
             if guest_vm_type == "CVM":
                 # CVM guest take little more time to boot
-                sleep_time = 100
+                # 20 seconds per VM (with default 1024M)
+                sleep_time = 20 * vm_count
             time.sleep(sleep_time)
 
             for i in range(len(procs)):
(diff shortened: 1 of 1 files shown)


LiliDeng (approved):

microsoft/testsuites/mshv/mshv_root_stress_tests.py
```diff
@@ -186,7 +186,8 @@ def _mshv_stress_vm_create(
             sleep_time = 10
             if guest_vm_type == "CVM":
                 # CVM guest take little more time to boot
-                sleep_time = 100
+                # 20 seconds per VM (with default 1024M)
+                sleep_time = 20 * vm_count
```
squirrelsc, line 190: Sleep is not a right fix. It needs to find a way to detect the state with an interval/timeout pattern.

pupacha: That is done intentionally. We want to launch multiple Cloud-hypervisor processes (VMs) but this is asynchronous and there is no straightforward way to determine if guest finished booting.  `sleep` is easiest way to wait until guest boots.

squirrelsc: The "sleep" looks the easiest way, but it causes problems a lot. Either a longer time, or wait the long time for all cases. If others copy your solution, the problem is worse. To wait a VM started, you can use `wait_tcp_port_ready` method. The `reboot` tool use `who` or `uptime` to check last boot time to check if reboot happened.

pupacha: The VMs we launch, here from this testcase, are without any networking support, so there's no way to check network readiness. Also, `sleep` is only for a special vm type i.e, `CVM` and this has to be given as an explicit input testcase variable. By default, testcase does wait for long time.
I agree that there has to be a better way to check if VM finished its botting. If in future, cloud-hypervisor tool is used more and scenarios like these comes up more, we will plan to implement VM readiness checks.

PR #3115, opened 2023-12-22
Revert "fix the issue when the node is not initialized"

LiliDeng:
Push the change into main by mistake...

This reverts commit befc1dbaf9ddcdcfd3408fa71ee6e6965ad8eed2.


LiliDeng:
@squirrelsc let know whether the change is good, if yes, I will close the revert PR. Sorry I push into main by mistake. I will check any policy can block push to main directly.

squirrelsc:
It needs to understand deeper, why the node is not initialized. It's ok, if you can push to master directly, because you are the owner. You can setup your own forked repo. It can reduce the chance to push incorrectly.

PR #3116, opened 2023-12-22
shell: fix mkdir with parents for Windows nodes

anirudhrb:
`spurplus` is unable to create parent directories when the target node is a Windows node. This is because it expects only POSIX paths, and its parent creation logic doesn't work well with Windows paths. In these cases, `spurplus` throws `OSError`.

To fix this, catch the `OSError` exception and fallback to using the `mkdir` command to create the directory.

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -422,6 +422,16 @@ class SshShell(InitializableMixin):
             if "Channel closed." in str(identifier):
                 assert isinstance(path_str, str)
                 self.spawn(command=["mkdir", "-p", path_str])
+        except OSError as e:
+            if not self.is_posix and parents:
+                # spurplus doesn't handle Windows style paths properly. As a result,
+                # it is unable to create parent directories when parents=True is
+                # passed. So, mkdir ultimately fails. In such cases, use command
+                # instead. On Windows, mkdir creates parent directories by default;
+                # no additional parameter is needed.
+                self._inner_shell.run(command=["mkdir", path_str])
+            else:
+                raise e
 
     def exists(self, path: PurePath) -> bool:
         """Check if a target directory/file exist


lisa/util/shell.py
```diff
@@ -422,6 +422,16 @@ def mkdir(
             if "Channel closed." in str(identifier):
                 assert isinstance(path_str, str)
                 self.spawn(command=["mkdir", "-p", path_str])
+        except OSError as e:
+            if not self.is_posix and parents:
+                # spurplus doesn't handle Windows style paths properly. As a result,
```
squirrelsc, line 427: You can check `self.is_posix` when calling the `_inner_shell.mkdir`. so if it's windows, always set parents to False.

anirudhrb: If we always set parents to false for Windows, how will the parent directories get created?

squirrelsc: As you added " On Windows, mkdir creates parent directories by default; no additional parameter is needed." So it can set to False always.

anirudhrb: Yeah, but internally spurplus doesn't use the `mkdir` console command to create directories. AFAIK, it uses the `mkdir` _SFTP command_ and I don't think the SFTP command automatically creates parent directories.

squirrelsc: If so, it can call `mkdir` always, not to use SFTP command on Windows.

anirudhrb: That is also possible. However, `mkdir` fails if the directory already exists. So that needs to be checked. There is also the `exist_ok` parameter that needs to be considered in such cases. spurplus already handles these. So, I chose this approach.

This also follows the existing pattern in the function which is:
```
invoke spurplus
if error
    invoke direct command
```

squirrelsc: I see, if so, we use current approach. Thank you for detailed explanation!

squirrelsc (dismissed):

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

LiliDeng:
testing

LiliDeng (commented):
testing

LiliDeng (changes requested):
testing

LiliDeng (approved):

PR #3117, opened 2023-12-25
Add bash prompt check when execute commands

lubaihua33:
Some images have bash prompts in stdout, E.g. yaseensmarket1645449809728 wordpress-red-hat images. The bash prompt is like:
        # ----------------------------------------------------------------------
        # Use the this command 'sudo bash ~/getcert.sh' to install a certificate
        # ----------------------------------------------------------------------
 This PR adds a bash prompt check and filters the bash prompt.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -112,6 +112,7 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
         self.capture_boot_time: bool = False
         self.capture_azure_information: bool = False
         self.capture_kernel_config: bool = False
+        self.has_checked_bash_prompt: bool = False
 
     @property
     def shell(self) -> Shell:
@@ -288,6 +289,8 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
         encoding: str = "",
     ) -> Process:
         self.initialize()
+        if isinstance(self, RemoteNode):
+            self._check_bash_prompt()
 
         return self._execute(
             cmd,
@@ -735,6 +738,26 @@ class RemoteNode(Node):
             self.log.debug(f"password prompt {i}: {password_prompt}")
         ssh_shell = cast(SshShell, self.shell)
         ssh_shell.password_prompts = password_prompts
+
+    def _check_bash_prompt(self) -> None:
+        # Check if there is bash prompt in stdout of command. If yes, the prompt
+        # should be filtered from the stdout. E.g. image yaseensmarket1645449809728
+        # wordpress-red-hat images.
+        if not self.has_checked_bash_prompt:
+            process = self._execute(f"echo {constants.LISA_TEST_FOR_BASH_PROMPT}")
+            result = process.wait_result(10)
+            if result.stdout.endswith(f"{constants.LISA_TEST_FOR_BASH_PROMPT}"):
+                bash_prompt = result.stdout.replace(
+                    constants.LISA_TEST_FOR_BASH_PROMPT, ""
+                )
+                if bash_prompt:
+                    self.log.debug(
+                        "detected bash prompt, it will be removed from every output: "
+                        f"{bash_prompt}"
+                    )
+                    ssh_shell = cast(SshShell, self.shell)
+                    ssh_shell.bash_prompt = bash_prompt
+                self.has_checked_bash_prompt = True
 
     def _reset_password(self) -> bool:
         from lisa.features import PasswordExtension
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -397,15 +397,9 @@ class Process:
         if self._is_posix and self._sudo:
             self._result.stdout = self._filter_sudo_result(self._result.stdout)
 
-        if (
-            isinstance(self._shell, SshShell)
-            and self._shell._inner_shell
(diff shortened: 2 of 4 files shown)


lisa/node.py
```diff
@@ -288,6 +289,8 @@ def execute_async(
         encoding: str = "",
     ) -> Process:
         self.initialize()
+        if isinstance(self, RemoteNode):
```
squirrelsc, line 292: Did you test with local? Maybe local should have the same logic.

lisa/node.py
```diff
@@ -736,6 +739,26 @@ def _check_password_and_store_prompt(self) -> None:
         ssh_shell = cast(SshShell, self.shell)
         ssh_shell.password_prompts = password_prompts
 
+    def _check_bash_prompt(self) -> None:
+        # Check if there is bash prompt in stdout of command. If yes, the prompt
+        # should be filtered from the stdout. E.g. image yaseensmarket1645449809728
+        # wordpress-red-hat images.
+        if not self.has_checked_bash_prompt:
+            self.has_checked_bash_prompt = True
```
squirrelsc: Move it to the end of the `if` block. It will be more stable with some race condition, like run two commands at the same time.

lisa/node.py
```diff
@@ -736,6 +739,26 @@ def _check_password_and_store_prompt(self) -> None:
...
+                f"echo {constants.LISA_TEST_FOR_BASH_PROMPT}",
+                shell=True,
+                sudo=True,
+                no_info_log=True,
+            )
+            result = process.wait_result(10)
+            bash_prompt = result.stdout.replace(
+                f"{constants.LISA_TEST_FOR_BASH_PROMPT}", ""
+            )
+            self.log.debug(f"bash prompt: {bash_prompt}")
```
squirrelsc: The log should be like: "detected bash prompt, it will be removed from every output:".

lisa/node.py
```diff
@@ -736,6 +739,26 @@ def _check_password_and_store_prompt(self) -> None:
...
+        if not self.has_checked_bash_prompt:
+            self.has_checked_bash_prompt = True
+            process = self._execute(
+                f"echo {constants.LISA_TEST_FOR_BASH_PROMPT}",
+                shell=True,
+                sudo=True,
+                no_info_log=True,
+            )
+            result = process.wait_result(10)
+            bash_prompt = result.stdout.replace(
```
squirrelsc: It needs to check if the content is end with the test prompt. Not to use replace.

lisa/node.py
```diff
@@ -736,6 +739,26 @@ def _check_password_and_store_prompt(self) -> None:
...
+    def _check_bash_prompt(self) -> None:
+        # Check if there is bash prompt in stdout of command. If yes, the prompt
+        # should be filtered from the stdout. E.g. image yaseensmarket1645449809728
+        # wordpress-red-hat images.
+        if not self.has_checked_bash_prompt:
+            self.has_checked_bash_prompt = True
+            process = self._execute(
+                f"echo {constants.LISA_TEST_FOR_BASH_PROMPT}",
+                shell=True,
+                sudo=True,
```
squirrelsc: Not use sudo by default.

lisa/util/process.py
```diff
@@ -519,19 +513,32 @@ def _filter_profile_error(self, raw_input: str) -> str:
         if (
             isinstance(self._shell, SshShell)
             and self._shell.spawn_initialization_error_string
+            and self._shell._inner_shell
+            and self._shell._inner_shell._spur._shell_type
+            == spur.ssh.ShellTypes.minimal
         ):
             raw_input = raw_input.replace(
-                rf"{self._shell.spawn_initialization_error_string}\n", ""
+                f"{self._shell.spawn_initialization_error_string}\n", ""
```
squirrelsc: For every minor fix, use a separated commit.

squirrelsc (dismissed):

lisa/node.py
lubaihua33: I tried to install lisav3 in the image which has the bash prompt. However, it failed for lacking some dependency packages. It is not easy to reproduce with local. When we encounter the same issue in the local scenario, I can add the same logic with local node by then.

lisa/node.py
```diff
@@ -736,6 +739,26 @@ def _check_password_and_store_prompt(self) -> None:
...
+    def _check_bash_prompt(self) -> None:
+        # Check if there is bash prompt in stdout of command. If yes, the prompt
+        # should be filtered from the stdout. E.g. image yaseensmarket1645449809728
+        # wordpress-red-hat images.
+        if not self.has_checked_bash_prompt:
+            process = self._execute(f"echo {constants.LISA_TEST_FOR_BASH_PROMPT}")
+            result = process.wait_result(10)
+            if result.stdout.endswith(f"{constants.LISA_TEST_FOR_BASH_PROMPT}"):
+                bash_prompt = result.stdout.replace(
+                    f"{constants.LISA_TEST_FOR_BASH_PROMPT}", ""
```
LiliDeng:
```
result.stdout.replace(
                    constants.LISA_TEST_FOR_BASH_PROMPT, ""
```

lubaihua33: Done

LiliDeng (approved):

PR #3120, opened 2023-12-26
update hibernation

LiliDeng:


lisa/features/startstop.py
```diff
@@ -48,3 +48,6 @@ def restart(self, wait: bool = True) -> None:
         self._log.info("restarting")
         self._restart(wait=wait)
         self._node.close()
+
+    def status(self) -> str:
+        return ""
```
squirrelsc: raise NotImplementedError()

lisa/features/startstop.py
```diff
@@ -48,3 +48,6 @@ def restart(self, wait: bool = True) -> None:
         self._log.info("restarting")
         self._restart(wait=wait)
         self._node.close()
+
+    def status(self) -> str:
```
squirrelsc: If it's a method call it like `get_status`, or it can be a property called `status`.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +182,20 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
...
+        platform: AzurePlatform = self._platform  # type: ignore
+        compute_client = get_compute_client(platform, api_version="2021-07-01")
+        status = (
+            compute_client.virtual_machines.get(
+                self._resource_group_name, self._vm_name, expand="instanceView"
+            )
+            .instance_view.statuses[1]
+            .display_status
+        )
+        if isinstance(status, str):
```
squirrelsc: use assert, and return the value.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -182,6 +182,20 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
         if wait:
             wait_operation(operation, failure_identity="Start/Stop")
 
+    def get_status(self) -> str:
+        platform: AzurePlatform = self._platform  # type: ignore
+        compute_client = get_compute_client(platform, api_version="2021-07-01")
```
squirrelsc: try if latest version is ok, then use the latest version, not hard code and copy the previous versions.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1923,6 +1923,17 @@ def _resource_sku_to_capability(  # noqa: C901
             node_space.network_interface.max_nic_count = 1
             node_space.network_interface.nic_count = search_space.IntRange(min=1, max=1)
 
+        assert resource_sku.family, "'resource_sku.family' must not be 'None'"
```
squirrelsc: It should be in feature's `create_setting`. If it's not supported, return None. Refer to Gpu feature.

lisa/tools/hibernation_setup.py
```diff
@@ -74,9 +73,13 @@ def _install(self) -> bool:
         return self._check_exists()
 
     def _check(self, pattern: Pattern[str]) -> int:
-        dmesg = self.node.tools[Dmesg]
-        dmesg_output = dmesg.get_output(force_run=True)
-        matched_lines = find_patterns_in_lines(dmesg_output, [pattern])
+        cat = self.node.tools[Cat]
+        log_output = ""
+        if (self.node.execute("ls -lt /var/log/syslog", sudo=True)).exit_code == 0:
```
squirrelsc: use the ls tool.

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
     entry_before_hibernation = hibernation_setup_tool.check_entry()
     exit_before_hibernation = hibernation_setup_tool.check_exit()
     received_before_hibernation = hibernation_setup_tool.check_received()
     uevent_before_hibernation = hibernation_setup_tool.check_uevent()
-    startstop = node.features[StartStop]
+
+    # only set up hibernation setup tool for the first time
     hibernation_setup_tool.start()
-    startstop.stop(state=features.StopState.Hibernate)
+    sleep(300)
```
squirrelsc: Don't hardcode sleep.

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
+    try:
+        startstop.stop(state=features.StopState.Hibernate)
+    except Exception as identifier:
```
squirrelsc: let's use `e` as the variable name in future.

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
+        try:
+            dmesg = node.tools[Dmesg]
+            content = dmesg.get_output(force_run=True)
+            log.debug(content)
```
squirrelsc: Don't output it again. the dmesg tool should output it already.

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
+        except Exception as identifier_dmesg:
+            log.debug(
```
squirrelsc: `type(ex)` can return type.

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
+                f"identifier_dmesg exception is {identifier_dmesg.__class__.__name__}: "
```
squirrelsc: The error message should be more descriptive. In the power test case, it should say like "error on get dmesg output: ".

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
+                f"{identifier_dmesg}"
+            )
+        log.debug(f"exception is {identifier.__class__.__name__}: {identifier}")
```
squirrelsc: It doesn't need to output debug message, because the exception is raised in next line. It's more obvious than debug log already.

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
     timer = create_timer()
     while timeout > timer.elapsed(False):
-        is_ready, _ = wait_tcp_port_ready(
-            node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_ADDRESS],
-            node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_PORT],
-            log=log,
-            timeout=10,
-        )
-        if not is_ready:
+        if "VM deallocated" == startstop.get_status():
```
squirrelsc: The status should be abstracted in Azure as an enumeration, not use the string directly.

microsoft/testsuites/power/common.py
```diff
@@ -32,33 +32,66 @@ def is_distro_supported(node: RemoteNode) -> None:
...
+        raise LisaException("VM is not in deallocated status after hibernation")
+
     startstop.start()
+    dmesg = node.tools[Dmesg]
+    content = dmesg.get_output(force_run=True)
+    if "Hibernate inconsistent memory map detected" in content:
+        raise LisaException(
+            "fail to hibernate for 'Hibernate inconsistent memory map detected'"
+        )
+    if "Call Trace:" in content and "Out of memory" not in content:
```
squirrelsc: It should be checked by the `check_kernel_errors`

microsoft/testsuites/power/common.py
```diff
@@ -133,11 +166,11 @@ def run_network_workload(environment: Environment) -> Decimal:
...
-        kill.by_name("iperf3")
-        kill.by_name("fio")
-        kill.by_name("stress-ng")
+    try:
+        for node in environment.nodes.list():
+            kill = node.tools[Kill]
+            kill.by_name("iperf3")
+            kill.by_name("fio")
+            kill.by_name("stress-ng")
+    finally:
```
squirrelsc: finally/pass do nothing. If it needs to eat the exception, it should be an `except`.

microsoft/testsuites/power/common.py
```diff
@@ -133,11 +166,11 @@ def run_network_workload(environment: Environment) -> Decimal:
...
 def cleanup_env(environment: Environment) -> None:
-    remote_node = cast(RemoteNode, environment.nodes[0])
-    startstop = remote_node.features[StartStop]
-    startstop.start()
-    for node in environment.nodes.list():
-        kill = node.tools[Kill]
-        kill.by_name("iperf3")
-        kill.by_name("fio")
-        kill.by_name("stress-ng")
+    try:
```
squirrelsc: Why remove the `start`?

microsoft/testsuites/power/stress.py
```diff
@@ -51,8 +52,8 @@ def before_case(self, log: Logger, **kwargs: Any) -> None:
     def stress_hibernation(self, environment: Environment, log: Logger) -> None:
         node = cast(RemoteNode, environment.nodes[0])
         is_distro_supported(node)
-        for _ in range(0, self._loop):
-            verify_hibernation(node, log)
+        for index in range(1, self._loop):
```
squirrelsc: why change this part? The `1` is a magic number, add comments above to explain it.

microsoft/testsuites/power/power.py
```diff
@@ -200,9 +200,82 @@ def verify_hibernation_with_memory_workload(
...
+        requirement=simple_requirement(
+            min_nic_count=8,
+            network_interface=Synthetic(),
+            supported_features=[HibernationEnabled()],
+        ),
+    )
+    def verify_hibernation_synthetic_network_max_nics(
+        self, environment: Environment, log: Logger
+    ) -> None:
+        node = cast(RemoteNode, environment.nodes[0])
```
squirrelsc: use `node: Node` in parameter can return the first node...

microsoft/testsuites/power/power.py
```diff
@@ -200,9 +200,82 @@ def verify_hibernation_with_memory_workload(
...
+    def verify_hibernation_max_data_disks(
+        self, environment: Environment, log: Logger
+    ) -> None:
+        node = cast(RemoteNode, environment.nodes[0])
+        is_distro_supported(node)
+        disk = node.features[Disk]
+        data_disks_before_hibernation = disk.get_raw_data_disks()
+        verify_hibernation(node, log)
+        data_disks_after_hibernation = disk.get_raw_data_disks()
+        assert_that(
```
squirrelsc:
```suggestion
        assert_that(data_disks_after_hibernation).describe_as(...
            ).is_length(data_disks_before_hibernation)
```

lisa/sut_orchestrator/azure/common.py
```diff
@@ -154,6 +155,12 @@
 METADATA_ENDPOINT = "http://169.254.169.254/metadata/instance?api-version=2021-02-01"
 
 
+class AzureVMStatus(Enum):
```
squirrelsc: It looks like the hibernation test case is Azure only now. Is the "deallocated" state common in other platforms? if so, it can be moved to common startstop feature.

microsoft/testsuites/power/power.py
squirrelsc: Use `is_length` instead of `len(...)`, so it can print the values, if it's not right.

lisa/features/startstop.py
```diff
@@ -48,3 +48,6 @@ def restart(self, wait: bool = True) -> None:
         self._log.info("restarting")
         self._restart(wait=wait)
         self._node.close()
+
+    def get_status(self) -> str:
```
squirrelsc: It needs to define the return value as an enum in the common class, and map it to str in the Azure class.

microsoft/testsuites/power/common.py
```diff
@@ -32,69 +32,78 @@ def is_distro_supported(node: RemoteNode) -> None:
...
+        if startstop.get_status() == AzureVMStatus.Deallocated.value:
+            is_ready = False
             break
     if is_ready:
-        raise LisaException("VM still can be accessed after hibernation")
+        raise LisaException("VM is not in deallocated status after hibernation")
+
     startstop.start()
+    dmesg = node.tools[Dmesg]
+    dmesg.check_kernel_errors(force_run=True, throw_error=not ignore_call_trace)
```
squirrelsc: The `ignore_call_trace` is used here only once. It can use the same name `throw_error`, or called `throw_error_on_kernel`. So it doesn't need `not`.

lisa/features/startstop.py
```diff
@@ -7,6 +7,12 @@
 FEATURE_NAME_STARTSTOP = "StartStop"
 
 
+class VMStatus(str, Enum):
+    Deallocated = "VM deallocated"
```
squirrelsc: The `Deallocated`, `Running`, `ProvisionSucceeded` are considered as common concepts, so they should be here. But the value like `VM deallocated`, `VM running`, `Provisioning succeeded` are Azure specific, so they should be a map in Azure feature.

LiliDeng: updated

squirrelsc (approved):

PR #3122, opened 2023-12-27
Add verify_supports_nvme

kamalca:
This test case checks whether the NVMe drivers are present such that the image can boot with DiskControllerType NVMe.

We needed a test that will do the same thing as nvme_enabled in the info section. This will enable AzQualify to check images before tagging them.
We cannot simply do verify_deployment_provision with disk_controller_type set to NVMe or use verify_nvme_disk_controller_type because deployment will fail if the image is not yet tagged as NVMe.

### DIFF
--- a/microsoft/testsuites/core/storage.py
+++ b/microsoft/testsuites/core/storage.py
@@ -33,10 +33,24 @@ from lisa.features.security_profile import (
 from lisa.node import Node
 from lisa.operating_system import BSD, Posix, Windows
 from lisa.schema import DiskControllerType, DiskOptionSettings, DiskType
+from lisa.search_space import SetSpace
 from lisa.sut_orchestrator import AZURE
+from lisa.sut_orchestrator.azure.common import AzureNodeSchema, VhdSchema
 from lisa.sut_orchestrator.azure.features import AzureDiskOptionSettings, AzureFileShare
+from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
 from lisa.sut_orchestrator.azure.tools import Waagent
-from lisa.tools import Blkid, Cat, Dmesg, Echo, Lsblk, Mount, NFSClient, Swap, Sysctl
+from lisa.tools import (
+    Blkid,
+    Cat,
+    Dmesg,
+    Echo,
+    Lsblk,
+    Lsinitrd,
+    Mount,
+    NFSClient,
+    Swap,
+    Sysctl,
+)
 from lisa.tools.blkid import PartitionInfo
 from lisa.tools.journalctl import Journalctl
 from lisa.tools.kernel_config import KernelConfig
@@ -284,6 +298,57 @@ class Storage(TestSuite):
     )
     def verify_nvme_disk_controller_type(self, node: RemoteNode) -> None:
         self._verify_disk_controller_type(node, DiskControllerType.NVME)
+
+    @TestCaseMetadata(
+        description="""
+        This test verifies the nvme drivers are either built in or present in initramfs.
+        This test can be used to verify an image before it is tagged as NVMe capable.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            supported_platform_type=[AZURE],
+        ),
+    )
+    def verify_supports_nvme(self, node: RemoteNode, environment: Environment) -> None:
+        has_nvme_core = node.tools[KernelConfig].is_built_in("CONFIG_NVME_CORE") or (
+            node.tools[KernelConfig].is_built_as_module("CONFIG_NVME_CORE")
+            and node.tools[Lsinitrd].has_module("nvme-core.ko")
+        )
+        has_nvme = node.tools[KernelConfig].is_built_in("CONFIG_BLK_DEV_NVME") or (
+            node.tools[KernelConfig].is_built_as_module("CONFIG_BLK_DEV_NVME")
+            and node.tools[Lsinitrd].has_module("nvme.ko")
+        )
+
+        # Pass condition
+        if has_nvme_core and has_nvme:
+            return
+
+        assert isinstance(environment.platform, AzurePlatform)
+        node_runbook: AzureNodeSchema = node.capability.get_extended_runbook(
+            AzureNodeSchema, AZURE
+        )
+        image = node_runbook.image
+        # Failure conditions
+        if image and not isinstance(image, VhdSchema):
+            image.load_from_platform(environment.platform)
+            if isinstance(image.disk_controller_type, DiskControllerType):
+                assert DiskControllerType.NVME != image.disk_controller_type, (
+                    "Image is tagged as NVMe capable, but NVMe drivers are missing."
+                    f"nvme-core: {has_nvme_core}, nvme: {has_nvme}"
+                )
+            elif isinstance(image.disk_controller_type, SetSpace):
+                assert (
+                    DiskControllerType.NVME not in image.disk_controller_type.items
+                ), (
+                    "Image is tagged as NVMe capable, but NVMe drivers are missing."
+                    f"nvme-core: {has_nvme_core}, nvme: {has_nvme}"
+                )
+
+        # Skip if image is not tagged as NVMe capable
+        raise SkippedException(
+            "NVME drivers are not present. "
+            f"nvme-core: {has_nvme_core}, nvme: {has_nvme}"
+        )
 
     @TestCaseMetadata(
         description="""


microsoft/testsuites/core/storage.py
```diff
@@ -270,6 +281,27 @@ def verify_scsi_disk_controller_type(self, node: RemoteNode) -> None:
     def verify_nvme_disk_controller_type(self, node: RemoteNode) -> None:
         self._verify_disk_controller_type(node, DiskControllerType.NVME)
 
+    @TestCaseMetadata(
+        description="""
+        This test verifies the nvme drivers are either built in or present in initramfs.
+        This test can be used to verify an image before it is tagged as NVMe capable.
```
squirrelsc, line 305: Make sure the test case is skipped, if it's not NVMe capable.

microsoft/testsuites/core/storage.py
```diff
@@ -270,6 +282,28 @@ def verify_scsi_disk_controller_type(self, node: RemoteNode) -> None:
...
+    def verify_supports_nvme(self, node: RemoteNode) -> None:
+        _has_nvme_core = node.tools[KernelConfig].is_built_in("CONFIG_NVME_CORE") or (
+            node.tools[KernelConfig].is_built_as_module("CONFIG_NVME_CORE")
+            and node.tools[Lsinitrd].has_module("nvme-core.ko")
+        )
+        _has_nvme = node.tools[KernelConfig].is_built_in("CONFIG_BLK_DEV_NVME") or (
+            node.tools[KernelConfig].is_built_as_module("CONFIG_BLK_DEV_NVME")
+            and node.tools[Lsinitrd].has_module("nvme.ko")
+        )
+        if not (_has_nvme_core and _has_nvme):
```
squirrelsc: Where is the validation line? When will the test case fail?

kamalca: This test case only checks NVMe capability for the image. You asked me to skip the test is the image is not NVMe capable. What would you like the behavior to be?

squirrelsc: As discussed offline, the logic should be like,
expected: False, actual: False -> Skipped
expected: False, actual: True -> Skipped (because this value is not a concern in this scenario)
expected: True, actual: False -> Failed
expected: True, actual: True -> Passed

In most cases, the expected values can be obtained from the test environments by vm_size or other information. If it's not, you can pass in a test variable. The default expected value should be False to prevent the test case fails, if the tag is not cared.

microsoft/testsuites/core/storage.py
```diff
@@ -270,6 +282,44 @@ def verify_scsi_disk_controller_type(self, node: RemoteNode) -> None:
...
 
+    @TestCaseMetadata(
+        description="""
+        This test verifies the nvme drivers are either built in or present in initramfs.
+        This test can be used to verify an image before it is tagged as NVMe capable.
+        """,
+        priority=1,
+    )
+    def verify_supports_nvme(self, node: RemoteNode, environment: Environment) -> None:
+        _has_nvme_core = node.tools[KernelConfig].is_built_in("CONFIG_NVME_CORE") or (
```
squirrelsc: local variable doesn't need to start with `_`.

microsoft/testsuites/core/storage.py
```diff
@@ -270,6 +282,44 @@ def verify_scsi_disk_controller_type(self, node: RemoteNode) -> None:
...
+            node.tools[KernelConfig].is_built_as_module("CONFIG_BLK_DEV_NVME")
+            and node.tools[Lsinitrd].has_module("nvme.ko")
+        )
+
+        # Pass condition
+        if _has_nvme_core and _has_nvme:
+            return
+
+        if environment.platform:
+            location = node.capability._extended_runbook.location
```
squirrelsc: It looks need an extra field on disk settings called `supported_disk_controller_type`. So, the checks can be on `node.features[Disk].supported_disk_controller_type`.

kamalca: LISA doesn't currently gather any data about the image features, so I'm not sure the best way to implement this. Thoughts?

kamalca: I feel like there would need to be an image equivalent to `_resource_sku_to_capability` (ie. `_marketplace_to_capability`) which will allow the image features determine the settings and behaviors of NodeSpace and Features.

squirrelsc: Ok, I thought it's to get features on vm sizes. Yes, we need to support image info in test case as well. There is an `AzureImageSchema` class in Azure. We can create a common ImageSchema in `NodeSpace` and add needed information in platform.

microsoft/testsuites/core/storage.py
```diff
@@ -270,6 +284,47 @@ def verify_scsi_disk_controller_type(self, node: RemoteNode) -> None:
...
+        # Pass condition
+        if has_nvme_core and has_nvme:
+            return
+
+        if isinstance(environment.platform, AzurePlatform):
+            node_runbook: AzureNodeSchema = node.capability.get_extended_runbook(
+                AzureNodeSchema, AZURE
+            )
+            location = node_runbook.location
+            marketplace = node_runbook.marketplace
```
LiliDeng: this case may run against vhd? if so, we need handle this situation.

kamalca:
Rebased on PR https://github.com/microsoft/lisa/pull/3255

microsoft/testsuites/core/storage.py
```diff
@@ -277,6 +292,57 @@ def verify_scsi_disk_controller_type(self, node: RemoteNode) -> None:
...
+        if has_nvme_core and has_nvme:
+            return
+
+        assert isinstance(environment.platform, AzurePlatform)
+        node_runbook: AzureNodeSchema = node.capability.get_extended_runbook(
+            AzureNodeSchema, AZURE
+        )
+        image = node_runbook.image
+        # Failure conditions
+        if image and not isinstance(image, VhdSchema):
```
squirrelsc, line 332: VHD should be able to set in the runbook to specify it supports nvme.

kamalca: We can't rely on people manually specifying things in the runbook because it causes confusion. This means we need a good default value. If we don't include NVMe in VHD default, people will be confused why they can't deploy VHD with NVMe disk controller type. On the other hand, if default value has NVMe (as it currently does) then there will be confusing failures on this test case. The test case will assume VHD supports NVMe by default when this isn't necessarily true.
If someone wants to check if their VHD supports VHD all they have to do is confirm that this test case is pass and not skip. VHDs aren't tagged with any info so there shouldn't be an expectation that LISA will check if they are correctly tagged.

squirrelsc: Can you list the situations in a table? With current logic, the test case is always disabled for vhd.

kamalca: |                             | Marketplace | SIG / CGI | VHD                 |
|-----------------------------|-------------|-----------|---------------------|
| Has drivers                 | PASS        | PASS      | PASS                |
| Missing drivers, not tagged | SKIP        | SKIP      | SKIP                |
| Missing drivers, tagged capable     | FAIL        | FAIL      | VHD is never tagged |

kamalca: The main purpose of this test case is "Tag the image if and only if the test case passes" to be used before the image is published publicly to the marketplace. We don't really care about the skip vs. failure condition unless an image published in the marketplace is tagged but doesn't have drivers. This test case does not really care about VHDs because they aren't tagged.

squirrelsc: It maybe a case, that we need to verify a VHD is good with the tag before it's published. If the runbook is not modified, the vhd can be skipped to verify. It needs to be capable to verify vhd support that or not by adding tag in runbook. How about create unsupported NVMe capable for vhd by default?

kamalca: Here is how I see the options
|                            | Deploying VHD on NVMe | Running verify_support_nvme  |
|----------------------------|-----------------------|------------------------------|
| Default support NVMe       |         Works!        | Why is test case failing? :( |
| Default don't support NVMe |   Doesn't deploy :(   |       Catches Failures       |

I could add a tag which it separate from `disk_controller_type` for the image, but it would be quite niche use case. It would be confusing and I don't think it would really be used.

squirrelsc: If we add a test variable for the test case, can it solve the problem?

PR #3123, opened 2024-01-02
Fix the regression issue of PR2908

lubaihua33:
This PR also fixes the regression issue of PR [https://github.com/https://github.com/microsoft/lisa/pull/2908].


lubaihua33:
@glima Could you please check if this fix has any regression for the scenario you mentioned in https://github.com/microsoft/lisa/pull/2908 ?

squirrelsc (approved):

PR #3124, opened 2024-01-03
increase pytorch/CUDA space from 8GB to 16GB

paxue:
Old Ubuntu 18.04.2023031501 version has issue with 8GB disk space installation. It has been solved in 18.04.2023063001. To be safe, increase to 16GB.

### DIFF
--- a/microsoft/testsuites/gpu/gpusuite.py
+++ b/microsoft/testsuites/gpu/gpusuite.py
@@ -262,8 +262,8 @@ class GpuTestSuite(TestSuite):
         _install_driver(node, log_path, log)
         _check_driver_installed(node, log)
 
-        # Step 1, pytorch and CUDA needs 8GB space to download and install
-        torch_required_space = 8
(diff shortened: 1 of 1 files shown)


squirrelsc:
@LiliDeng LGTM

paxue:
> Old Ubuntu 18.04.2023031501 version has issue with 8GB disk space installation. It has been solved in 18.04.2023063001. To be safe, increase to 16GB.

After debug, the root cause is pytorch and CUDA requires more than 9GB. 

Old 18.04.2023031501 version image used less space in /home folder, which made /home folder in sda1 has 9GB (> 8GB) free space. This would make pytorch and cuda installed on the /home folder but failed due to out of space.

The new 18.04.2023063001 version image used more space and left only 5.6GB (< 8GB) in /home folder, which made pytorch and cuda installed on the /mnt folder on sdb1 and passed.

LiliDeng (approved):

PR #3125, opened 2024-01-03
HyperV tool improvements

anirudhrb:
- New parameter to specify VM generation
- New parameter to disable secure boot
- New parameter to configure COM ports
- Enclose some arguments to `New-VM` in double quotes. These arguments can potentially contain spaces and other special chars.

All these will be used in the upcoming Hyper-V platform.

### DIFF
--- a/lisa/tools/hyperv.py
+++ b/lisa/tools/hyperv.py
@@ -3,6 +3,7 @@
 
 import re
 import time
+from typing import Dict, Optional
 
 from assertpy import assert_that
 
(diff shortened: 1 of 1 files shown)


lisa/tools/hyperv.py
```diff
@@ -91,6 +103,21 @@ def create_vm(
                 force_run=True,
             )
 
+        # configure COM ports if specified
+        if com_ports is None:
```
squirrelsc: It doesn't need to specify an empty com_ports, if it's not set.
```suggestion
        if com_ports:
            ...
```

anirudhrb: Changed.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

LiliDeng (approved):

PR #3127, opened 2024-01-05
Added working path required by wsl

paulli2017:
Kselftest under WSL for both local and remote situation has working path issue that is not implemented in Node class. The current change is to fix this issue and the test passed and Kselftest works as expected under local and azure situation.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -449,6 +449,19 @@ class Node(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixi
         """
         raise NotImplementedError()
 
+    def _get_remote_working_path(self) -> PurePath:
+        if self.is_posix:
+            remote_root_path = Path("$HOME")
+        else:
+            remote_root_path = Path("%TEMP%")
+
+        working_path = remote_root_path.joinpath(
+            constants.PATH_REMOTE_ROOT, constants.RUN_LOGIC_PATH
+        ).as_posix()
+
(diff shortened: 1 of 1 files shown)


paulli2017:

lisa/node.py
```diff
@@ -449,6 +449,19 @@ def get_working_path(self) -> PurePath:
         """
         raise NotImplementedError()
 
+    def get_remote_working_path(self) -> PurePath:
```
squirrelsc: name it like `_get_remote_working_path`, so it's a protected method.

lisa/node.py
```diff
@@ -658,17 +671,7 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
-        else:
-            remote_root_path = Path("%TEMP%")
-
-        working_path = remote_root_path.joinpath(
-            constants.PATH_REMOTE_ROOT, constants.RUN_LOGIC_PATH
-        ).as_posix()
-
-        # expand environment variables in path
-        return self.get_pure_path(self.expand_env_path(working_path))
+        return super().get_remote_working_path()
```
squirrelsc: `self.get_remote_working_path()` is clearer.

lisa/node.py
paulli2017: the change is done, thanks

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3128, opened 2024-01-05
fix ssh issue on old distro when use key

LiliDeng:
tested on below images
```
openlogic centos 6.10 6.10.20190808
credativ Debian 8 8.0.201804150
canonical 0001-com-ubuntu-server-focal 20_04-lts latest
```

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -148,6 +148,19 @@ def try_connect(
     # So try with paramiko firstly.
     paramiko_client = paramiko.SSHClient()
 
+    # Put ssh-rsa in the first element of preferred public keys
+    # to resolve ssh issues on old distros.
+    paramiko.Transport.preferred_pubkeys = (  # type: ignore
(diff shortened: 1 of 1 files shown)


squirrelsc (dismissed):

LiliDeng:
@squirrelsc not sure why it complains `lisa/util/shell.py:152:5: error: "Type[Transport]" has no attribute "preferred_pubkeys"`, any idea?

lisa/util/shell.py
```diff
@@ -148,6 +148,20 @@ def try_connect(
     # So try with paramiko firstly.
     paramiko_client = paramiko.SSHClient()
 
+    # Put ssh-rsa in the first element of preferred public keys
+    # to resolve ssh issues on old distros.
```
squirrelsc, line 152: The comments need to explain why use `hasattr`.

squirrelsc (approved):

PR #3129, opened 2024-01-08
AITL: exclude high cost ultra disks test cases

LiliDeng:


squirrelsc:
Please explain the test cases are high cost, so they are excluded in example templates.

LiliDeng:
> Please explain the test cases are high cost, so they are excluded in example templates.

I updated the commit message.

squirrelsc (approved):

PR #3131, opened 2024-01-09
Onboard Mariner to LISA Infiniband Test Suite

henryli001:
The PR enables LISA to run the Infiniband test suite on Mariner using the inbox Mellanox Core driver provided by the Mariner kernel. 

** Uncovered a bug in latest WALinuxAgent source (v2.9.1.1) where IP address fails to be assigned for IB interface. Have filed a [bug](https://github.com/Azure/WALinuxAgent/issues/3025) for WAlinuxAgent team to triage. Temporary workaround is to use WAlinuxAgent source v2.9.0.4. 

Current Test Status:
1. Successfully passing P1 verify_ping_pong and P2 verify_hpc_over_sriov test
2. P2 verify_ib_naming is skipped since it only applies for HPC images
3. P2 verify_hpc_over_nd is skipped since it requires the hv_network_direct kernel module, which is deprecated in newer images
4. P4 tests are currently failing for Mariner and are still under investigation

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -409,7 +409,7 @@ class Infiniband(Feature):
                 else:
                     raise UnsupportedKernelException(
                         node.os,
-                        "Cannot install OFED drivers without " "kernel-devel package",
+                        "Cannot install OFED drivers without kernel-devel package",
                     )
 
                 extra_params = (
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -346,7 +346,8 @@ class Infiniband(Feature):
         kernel_version = node.tools[Uname].get_linux_information().kernel_version
         self._install_dependencies()
 
-        if not isinstance(node.os, CBLMariner):
+        if isinstance(node.os, CBLMariner):
+            return
             # Install OFED
             ofed_version = self._get_ofed_version()
             if isinstance(node.os, Oracle):
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -12,7 +12,7 @@ from lisa.base_tools import Cat, Sed, Uname, Wget
 from lisa.feature import Feature
 from lisa.features import Disk
 from lisa.operating_system import CBLMariner, CentOs, Oracle, Redhat, Ubuntu
-from lisa.tools import Ls, Lspci, Make, Service
+from lisa.tools import Firewall, Ls, Lspci, Make, Service
 from lisa.tools.tar import Tar
 from lisa.util import (
     LisaException,
@@ -177,6 +177,9 @@ class Infiniband(Feature):
             self.install_ofed()
 
         node = self._node
+        # Turn off firewall
+        firewall = node.tools[Firewall]
+        firewall.stop()
         # Disable SELinux
         if not isinstance(node.os, CBLMariner):
             sed = node.tools[Sed]
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -11,7 +11,7 @@ from retry import retry
 from lisa.base_tools import Cat, Sed, Uname, Wget
 from lisa.feature import Feature
 from lisa.features import Disk
-from lisa.operating_system import CentOs, Oracle, Redhat, Ubuntu
+from lisa.operating_system import CentOs, Oracle, Redhat, Ubuntu, CBLMariner
 from lisa.tools import Firewall, Ls, Lspci, Make, Service
 from lisa.tools.tar import Tar
 from lisa.util import (
@@ -177,17 +177,15 @@ class Infiniband(Feature):
             self.install_ofed()
 
         node = self._node
-        # Turn off firewall
-        firewall = node.tools[Firewall]
-        firewall.stop()
         # Disable SELinux
-        sed = node.tools[Sed]
-        sed.substitute(
-            regexp="SELINUX=enforcing",
-            replacement="SELINUX=disabled",
-            file="/etc/selinux/config",
-            sudo=True,
-        )
+        if not isinstance(node.os, CBLMariner):
+            sed = node.tools[Sed]
+            sed.substitute(
+                regexp="SELINUX=enforcing",
+                replacement="SELINUX=disabled",
+                file="/etc/selinux/config",
+                sudo=True,
+            )
 
         # for non-hpc images, add net.ifnames=0 biosdevname=0 in boot kernel parameter
         # to make ib device name consistent across reboots
@@ -277,6 +275,18 @@ class Infiniband(Feature):
             "pciutils",
             "lsof",
         ]
+        cblmariner_required_packages = [
+            "pciutils-libs",
+            "pciutils",
+            "rdma-core",
+            "rdma-core-devel",
+            "libibverbs",
+            "libibverbs-utils",
+            "build-essential",
+            "libfabric-devel",
+            "ucx-ib",
+            "ucx-devel",
+        ]
         if isinstance(node.os, CentOs):
             node.execute(
                 "yum install -y https://partnerpipelineshare.blob.core.windows.net"
@@ -324,6 +334,8 @@ class Infiniband(Feature):
                 if node.os.is_package_in_repo(package):
                     ubuntu_required_packages.append(package)
             node.os.install_packages(ubuntu_required_packages)
+        elif isinstance(node.os, CBLMariner):
+            node.os.install_packages(cblmariner_required_packages)
         else:
             raise UnsupportedDistroException(
                 node.os,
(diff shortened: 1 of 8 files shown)


lisa/features/infiniband.py
```diff
@@ -177,17 +177,15 @@ def setup_rdma(self) -> None:
             self.install_ofed()
 
         node = self._node
-        # Turn off firewall
-        firewall = node.tools[Firewall]
-        firewall.stop()
```
LiliDeng, line 182: why remove this

lisa/sut_orchestrator/azure/features.py
```diff
@@ -190,23 +183,6 @@ def _execute(self, wait: bool, operator: str, **kwargs: Any) -> None:
...
-                )
-                .instance_view.statuses[1]
-                .display_status
-            )
-            assert isinstance(status, str), f"actual: {type(status)}"
-            assert self.azure_vm_status_map.get(status) is not None, "unknown vm status"
-            return cast(VMStatus, self.azure_vm_status_map.get(status))
-        except Exception as e:
-            raise LisaException(f"fail to get status of vm {self._vm_name}") from e
-
```
LiliDeng: rebase code, don't remove these

microsoft/runbook/azure.yml
```diff
@@ -1,15 +1,17 @@
-name: azure default
+name: azure infiniband
```
LiliDeng: no need update this

lisa/sut_orchestrator/azure/features.py
henryli001: This was caused by wrongdoing on my end when migrating changes across local lisa repositories. Have fixed this.

lisa/features/infiniband.py
henryli001: It's not removed but instead was moved to setup_rdma in lisa/sut_orchestrator/azure/features.py. Mariner requires a reboot after modifying waagent.conf to let WALinuxAgent enable IPoIB After the reboot, the firewall disabling no longer persist and thus I moved these two lines to a later point after the reboot.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -81,16 +85,50 @@ def deprovision(self) -> None:
         self.run("-deprovision --force", sudo=True, expected_exit_code=0)
 
     def upgrade_from_source(self) -> None:
-        git = self.node.tools[Git]
-        git.clone(self._src_url, cwd=self.node.working_path)
+        # Pin the WALinuxAgent to v2.9.0.4 because the latest v2.9.1.1
```
LiliDeng: add a method parameter version for 2.9.0.4, if no version specified, use latest code, otherwise checkout into this version in waagent source code

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -81,16 +85,50 @@ def deprovision(self) -> None:
...
+            gzip=True,
+            sudo=True,
+        )
+
         python_cmd, _ = self.get_python_cmd()
         for package in list(["python-setuptools", "python3-setuptools"]):
             if self.node.os.is_package_in_repo(package):  # type: ignore
                 self.node.os.install_packages(package)  # type: ignore
+
+        if not self.node.tools[Ls].path_exists(path="/usr/bin/python"):
```
LiliDeng: these logic should be put in self.get_python_cmd

lisa/sut_orchestrator/azure/tools.py
henryli001: Re-implemented this function based on the comment.

One follow-up question is whether there are any successful test passes with WALinuxAgent v2.9.1.1? In the previous experiment testing Ubuntu with WALinuxAgent v2.9.1.1, it fails to create an IP over IB.

lisa/sut_orchestrator/azure/tools.py
henryli001: Moved to this function

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -80,9 +84,14 @@ def deprovision(self) -> None:
         # self.run("-deprovision+user --force", sudo=True)
         self.run("-deprovision --force", sudo=True, expected_exit_code=0)
 
-    def upgrade_from_source(self) -> None:
+    def upgrade_from_source(self, source_version: str) -> None:
```
LiliDeng: source_version: str = ""

lisa/sut_orchestrator/azure/features.py
```diff
@@ -689,22 +690,38 @@ def setup_rdma(self) -> None:
...
+            )
+            sed.substitute(
+                regexp="# AutoUpdate.Enabled=y",
+                replacement="AutoUpdate.Enabled=y",
+                file="/etc/waagent.conf",
+                sudo=True,
+            )
+
+        if isinstance(self._node.os, CBLMariner):
+            self._node.reboot()
```
LiliDeng: add comments for why mariner need reboot here

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -100,12 +100,16 @@ def verify_hpc_over_sriov(self, log: Logger, node: Node) -> None:
...
-            "mlx_compat",
             "rdma_cm",
             "iw_cm",
             "ib_cm",
         ]
 
+        # mlx_compat is only needed if using Mellanox OFED. CBLMariner uses
+        # the inbox Mellanox core driver provided by kernel which does not need this
+        if not isinstance(node.os, CBLMariner):
+            expected_modules.append("mlx_compat")
```
LiliDeng: Please rebase code, https://github.com/microsoft/lisa/pull/3134 has removed this module.

lisa/features/infiniband.py
```diff
@@ -324,10 +329,12 @@ def _install_dependencies(self) -> None:
...
                     ubuntu_required_packages.append(package)
             node.os.install_packages(ubuntu_required_packages)
+        elif isinstance(node.os, CBLMariner):
+            node.os.install_packages(cblmariner_required_packages)
         else:
             raise UnsupportedDistroException(
                 node.os,
-                "Only CentOS 7.6-8.3 and Ubuntu 18.04-20.04 distros are "
+                "Only CentOS 7.6-8.3, Ubuntu 18.04-22.04 and CBLMariner distros are "
                 "supported by the HCP team",
```
squirrelsc: Fix the typo "HCP" to "HPC".

lisa/features/infiniband.py
```diff
@@ -339,94 +346,97 @@ def install_ofed(self) -> None:
...
 
-        # Install OFED
-        ofed_version = self._get_ofed_version()
-        if isinstance(node.os, Oracle):
-            os_class = "ol"
-        elif isinstance(node.os, Redhat):
-            os_class = "rhel"
-        else:
-            os_class = node.os.name.lower()
+        if not isinstance(node.os, CBLMariner):
```
squirrelsc: Change to below, and add comments to explain why Mariner skipped.
```suggestion
        if isinstance(node.os, CBLMariner):
            return
```

lisa/sut_orchestrator/azure/features.py
```diff
@@ -689,22 +690,38 @@ def setup_rdma(self) -> None:
         if len(devices) > 1:
             # upgrade waagent to latest version to resolve
             # multiple ib devices not getting ip address issue
-            waagent.upgrade_from_source()
+            waagent.upgrade_from_source("v2.9.0.4")
```
squirrelsc, line 695: What's this version? Is it a beta version or released version? Is this step to upgrade or downgrade? Hardcoded version is going to old. Is there a better way to handle it?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -689,22 +690,38 @@ def setup_rdma(self) -> None:
...
-        )
-        waagent.restart()
+        if isinstance(self._node.os, CBLMariner):
+            sed.append(
+                text="OS.EnableRDMA=y",
+                file="/etc/waagent.conf",
+                sudo=True,
+            )
+        else:
+            sed.substitute(
```
squirrelsc: Those logic is still valid in Mariner, because the original string is not found.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -689,22 +690,38 @@ def setup_rdma(self) -> None:
...
-        )
-        sed.substitute(
-            regexp="# AutoUpdate.Enabled=y",
-            replacement="AutoUpdate.Enabled=y",
-            file="/etc/waagent.conf",
-            sudo=True,
-        )
-        waagent.restart()
+        if isinstance(self._node.os, CBLMariner):
+            sed.append(
```
squirrelsc: Avoid to append blindly. If this environment runs tests more than once, this line will be appended twice. Use cat.read to check if the line doesn't exist, and then append.

lisa/sut_orchestrator/azure/features.py
henryli001: This is a officially released version but not the latest one. And this step is still upgrading.

The latest walinuxagent source does not work when testing Ubuntu 22.04 marketplace image (canonical 0001-com-ubuntu-server-jammy 22_04-lts 22.04.202307010). No IP address was assigned to ib0. But it works when switching back to the older v2.9.0.4. 

Needs further digging on this one.

lisa/sut_orchestrator/azure/features.py
henryli001: Good catch! I'll make a fix for this.

lisa/sut_orchestrator/azure/features.py
henryli001: Fixed

lisa/sut_orchestrator/azure/tools.py
henryli001: Fixed

lisa/sut_orchestrator/azure/features.py
henryli001: Comments added

lisa/sut_orchestrator/azure/features.py
henryli001: The same error can be reproed on both Ubuntu 2204 and Mariner v2 marketplace images when using the latest WALinuxAgent source version. But this error won't show up if switching to the prior 2.9.0.4 version.

Was able to discover error logs in waagent.log related to setting up RDMA device and thus based on current investigation, this error seems like a regression in the latest WALinuxAgent source code. Have filed a [bug](https://github.com/Azure/WALinuxAgent/issues/3025) in the WALinuxAgent repository for the WALinuxAgent team to triage. 

For now, the workaround solution is to temporarily upgrade WALinuxAgent to the prior v2.9.0.4 until a conclusion is provided by the WALinuxAgent team regarding this matter.

lisa/features/infiniband.py
henryli001: I fixed the typo and refined the error log since CBLMariner is not supported by the HPC team yet. There's no CBLMariner HPC image.

lisa/features/infiniband.py
```diff
@@ -424,7 +437,7 @@ def install_ofed(self) -> None:
         node.execute(
             "/etc/init.d/openibd force-restart",
             expected_exit_code=0,
-            expected_exit_code_failure_message="SetupRDMA: failed to restart driver",
+            expected_exit_code_failure_message="SetupRDMA: failed to " "restart driver",
```
LiliDeng: ```expected_exit_code_failure_message="SetupRDMA: failed to restart driver",```

lisa/sut_orchestrator/azure/features.py
```diff
@@ -689,22 +690,51 @@ def setup_rdma(self) -> None:
...
-            file="/etc/waagent.conf",
+            file=waagent_config_path,
             sudo=True,
         )
-        waagent.restart()
+
+        # CBLMariner uses the Mellanox inbox driver instead of the Mellanox OFED driver,
+        # which requires a reboot of the VM for WALinuxAgent to enable RDMA.
+        if isinstance(self._node.os, CBLMariner):
+            self._node.reboot()
```
LiliDeng: restart the waagent should be enough, could you please double check?

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -80,9 +84,14 @@ def deprovision(self) -> None:
         # self.run("-deprovision+user --force", sudo=True)
         self.run("-deprovision --force", sudo=True, expected_exit_code=0)
 
-    def upgrade_from_source(self) -> None:
+    def upgrade_from_source(self, source_version: str = "") -> None:
         git = self.node.tools[Git]
         git.clone(self._src_url, cwd=self.node.working_path)
+        if source_version:
+            git.checkout(
+                ref=f"tags/{source_version}",
```
LiliDeng: ```ref=source_version,``` should be enough?

lisa/sut_orchestrator/azure/features.py
henryli001: I actually tried just restarting waagent.service without rebooting and that does not work. Suspect that it might be related to using the inbox driver. This [instruction](https://techcommunity.microsoft.com/t5/azure-compute-blog/configuring-infiniband-for-ubuntu-hpc-and-gpu-vms/ba-p/1221351) also says to reboot the VM after editing waagent conf for the inbox driver scenario.

lisa/sut_orchestrator/azure/tools.py
henryli001: Good catch, this was fixed.

lisa/features/infiniband.py
henryli001: An error caused by black. Have fixed it.

lisa/sut_orchestrator/azure/features.py
squirrelsc: Ok, please keep tracking the bug.

lisa/sut_orchestrator/azure/features.py
LiliDeng: I think the comments is unclear, probably the reboot is for loading some modules, not for waagent.

henryli001: After further investigation, seems like the reboot is not required. The actual root-cause is that required kernel modules: ib_uverbs, rdma_ucm, ib_umad and ib_ipoib need to be loaded so that walinuxagent can successfully enable RDMA. Re-factor the implementation to check and load these four kernel modules and then restart the service.

LiliDeng (dismissed):

LiliDeng:
@squirrelsc let know any comments.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -689,21 +689,48 @@ def setup_rdma(self) -> None:
         if len(devices) > 1:
             # upgrade waagent to latest version to resolve
             # multiple ib devices not getting ip address issue
-            waagent.upgrade_from_source()
+
+            # Upgrade to v2.9.0.4 since the latest v2.9.1.1 version
+            # does not successfully assign IP over the IB interface
+            waagent.upgrade_from_source("v2.9.0.4")
```
squirrelsc, line 695: Forget to mention, please mark the node to dirty, so the environment won't be used by other test cases.

henryli001: Fixed

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3134, opened 2024-01-10
Remove mlx_compat

kamalca:
No longer have mlx_compat as a required module. It is not always required depending on how HPC support was installed.

### DIFF
--- a/microsoft/testsuites/hpc/infinibandsuite.py
+++ b/microsoft/testsuites/hpc/infinibandsuite.py
@@ -100,7 +100,6 @@ class InfinibandSuite(TestSuite):
             "ib_uverbs",
             "ib_core",
             "mlx5_core",
-            "mlx_compat",
             "rdma_cm",
             "iw_cm",
             "ib_cm",


LiliDeng:
> No longer have mlx_compat as a required module. It is not always required depending on the OFED driver version and how HPC was installed.

Can you elaborate more on this? thanks!

kamalca:
> > No longer have mlx_compat as a required module. It is not always required depending on the OFED driver version and how HPC was installed.
> 
> Can you elaborate more on this? thanks!

I haven't been able to find great documentation on what exactly the mlx_compat driver is responsible for, but I have deduced that it at least provides some symbols for the mlx_core and other mlx drivers. For the OFED drivers, mlx_compat is the highest level dependency so it must be loaded first, then you can load mlx_core, then the mlx_ib drivers.

image

Without the mlx_compat driver you would get a missing symbols error. However, there are ways to enable infiniband support other than OFED drivers, for example the mlx_core drivers can be built with the kernel, and would then not need the mlx_compat module. Some of the HPC images do not have this mlx_compat driver. I do not know exactly how the HPC teams set up these images, but if they required the mlx_compat module, then the mlx_core modules would fail to load. As long as the mlx_core module is loaded, it doesn't really matter whether it depended on the mlx_compat driver or not, we can still test the infiniband functionality.

LiliDeng (approved):

PR #3135, opened 2024-01-10
Skip cases where there is no initrd

kamalca:
Initrd is a dependency, not a failure condition, so we can skip the test case if the dependency is not met.

CVM does not have initrd so it will fail with "Could not find initrd file path". This does not mean that the modules are missing.

Ubuntu Mantic fails with "modules.dep" could not be found". The initrd might be structured differently in the new version of Ubuntu.

### DIFF
--- a/microsoft/testsuites/core/hv_module.py
+++ b/microsoft/testsuites/core/hv_module.py
@@ -19,7 +19,7 @@ from lisa.operating_system import BSD, Redhat
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
 from lisa.sut_orchestrator.azure.tools import LisDriver
 from lisa.tools import KernelConfig, Lsinitrd, Lsmod, Modinfo, Modprobe
-from lisa.util import SkippedException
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/hv_module.py
```diff
@@ -102,9 +102,14 @@ def verify_initrd_modules(self, environment: Environment) -> None:
...
-            if not lsinitrd.has_module(module_file_name=hv_modules_file_names[module]):
-                missing_modules.append(module)
+        try:
+            for module in hv_modules_file_names:
+                if not lsinitrd.has_module(
+                    module_file_name=hv_modules_file_names[module]
+                ):
+                    missing_modules.append(module)
+        except (LisaException, AssertionError) as e:
+            raise SkippedException(str(e))
```
squirrelsc: Add comments above to explain why exceptions are considered as skipped, instead of failed.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3136, opened 2024-01-11
tools/hyperv: support extra args for create_vm

anirudhrb:
Allow passing per-command extra args to create_vm(). The key of the dictionary is the command and the value is the string containing the extra arguments to that command.

This is useful for passing experimental arguments from the runbook.

This feature will be utilized in the upcoming Hyper-V platform.

### DIFF
--- a/lisa/tools/hyperv.py
+++ b/lisa/tools/hyperv.py
@@ -68,40 +68,39 @@ class HyperV(Tool):
 
         powershell = self.node.tools[PowerShell]
 
-        if not extra_args:
-            extra_args = {}
-        else:
-            # convert all keys to lowercase to avoid case sensitivity
-            extra_args = {k.lower(): v for k, v in extra_args.items()}
-
-        extra_args_str = extra_args["new-vm"] if "new-vm" in extra_args else ""
-
         # create a VM in hyperv
-        powershell.run_cmdlet(
-            f'New-VM -Name "{name}" -Generation {generation} '
-            f"-MemoryStartupBytes {memory}MB -BootDevice VHD "
-            f'-VHDPath "{guest_image_path}" -SwitchName "{switch_name}" '
-            f"{extra_args_str}",
+        self._run_hyperv_cmdlet(
+            "New-VM",
+            f'-Name "{name}" -Generation {generation} -MemoryStartupBytes {memory}MB '
+            f'-BootDevice VHD -VHDPath "{guest_image_path}" '
+            f'-SwitchName "{switch_name}"',
+            extra_args=extra_args,
             force_run=True,
         )
 
-        if "set-vmprocessor" in extra_args:
-            powershell.run_cmdlet(
-                f"Set-VMProcessor -VMName {name} {extra_args['set-vmprocessor']}",
+        if extra_args is not None and "set-vmprocessor" in extra_args:
+            self._run_hyperv_cmdlet(
+                "Set-VMProcessor",
+                f"-VMName {name}",
+                extra_args=extra_args,
                 force_run=True,
             )
 
         # set cores and memory type
-        powershell.run_cmdlet(
-            f"Set-VM -Name {name} -ProcessorCount {cores} -StaticMemory "
+        self._run_hyperv_cmdlet(
+            "Set-VM",
+            f"-Name {name} -ProcessorCount {cores} -StaticMemory "
             "-CheckpointType Disabled",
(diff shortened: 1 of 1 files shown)


lisa/tools/hyperv.py
```diff
@@ -61,20 +61,36 @@ def create_vm(
...
         if stop_existing_vm:
             self.delete_vm(name)
 
         powershell = self.node.tools[PowerShell]
 
+        if not extra_args:
+            extra_args = {}
+        else:
+            # convert all keys to lowercase to avoid case sensitivity
+            extra_args = {k.lower(): v for k, v in extra_args.items()}
```
squirrelsc: This may be used in other places, so it needs to refine the code to add a method like `_run_hyperv_cmdlet(cmd:str, args: str, extra_args:Optional[Dict[str, str]]= None, force_run:bool = False)`

anirudhrb: Sure, I will add that function. But do we want to convert _all_ `run_cmdlet()` calls in this tool to `_run_hyperv_cmdlet()` right away? Or do it only for `create_vm()` now and later, whenever a use-case comes up, we can do it for other functions?

squirrelsc: How about direct implement it in `run_cmdlet`. It use split to take first part, and append `extra_args`.

anirudhrb: To me it doesn't make much sense to pass a dictionary of args that might or might not be applicable for the command being run with `run_cmdlet`. Ideally, the processing of extra args should take place in the upper layers.

IMO we are over-engineering things here. `extra_args` should be used *only* for experimental features. For non-experimental features, a proper new parameter should be introduced to `create_vm` instead of abusing `extra_args`. So, the use-case for `extra_args` is limited and we don't need to generalize it that much. We can get away with supporting experimental arguments _where needed_ instead of adding in a generic way because it will likely *not* be used widely. So, we can simply add a new parameter named `vm_processor_experimental_args` (corresponding to `Set-VMProcessor`) and handle it appropriately within the function. And if experimental args are needed in more places in the future (unlikely, IMO), we can add them in a similar manner.

What do you think? Do you foresee a lot of usage of experimental args?

squirrelsc: Two scenarios need extra args. One is the experimental features, and another is to provide flexibility on unforeseen scenarios. IMO, all above designs are not perfect fit to both goals. I'm thinking how to make it more extendable on tools, like features.

It looks two points in above explanation: 1) support the extra args in run_cmdlet or hyperv only. I think we can leave it to hyperv instead of in the powershell tool. 2) use extra_args or vm_processor_experimental_args. The extra_args can save time on support future related functionalities. It's configurable in runbook without code changes.

anirudhrb: > provide flexibility on unforeseen scenarios

Do you have any examples from the past of such "unforeseen scenarios"? Is there similar flexibility in other platforms?

Generally, new functionalities should be implemented by making LISA code changes and adding the required parameters instead of hiding the new arguments in the runbook. Experimental features are the only justifiable case for hiding arguments in runbook.

squirrelsc: Some examples like how the tag and properties of VM is used internally, and on_before_deployment features. HyperV uses PowerShell commands, it equals the arm template on Azure.

anirudhrb: Hyper-V PowerShell commands and ARM template are not equivalent. The PowerShell commands are code while ARM template is configuration. It is easier to programmatically modify configuration than code.

I can see how `on_before_deployment` can use `extra_args`. However, adding `Feature`s will necessarily need code change and it cannot be avoided by specifying `extra_args` in the runbook. Which was the main reason (I guess) for preferring generic `extra_args`. So, `on_before_deployment` doesn't really need `extra_args`. All it needs is an API between the `Feature` and the platform. In the case of Azure, this API is the ARM template. For Hyper-V, we can introduce a new configuration class called `HypervVMConfig` that the feature can modify in `on_before_deployment` and the platform implementation will take care of passing the correct parameters when invoking the tool.

> how the tag and properties of VM is used internally

Sorry, I didn't understand this use-case. Could you please elaborate? How does it benefit by having extra args in the runbook?

anirudhrb: Resolved offline.

squirrelsc (approved):

LiliDeng (approved):

PR #3139, opened 2024-01-12
add capture_vm_information and catch exception if no serial console enabled

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -747,44 +747,51 @@ def _delete_boot_diagnostic_container(
...
         information: Dict[str, Any] = {}
-        for key, method in self._environment_information_hooks.items():
-            node.log.debug(f"detecting {key} ...")
-            try:
-                value = method(node)
-                if value:
-                    information[key] = value
-            except Exception as identifier:
-                node.log.exception(f"error on get {key}.", exc_info=identifier)
+        if node.capture_vm_information:
```
squirrelsc: Reduce indent levels
```suggestion
        if node.capture_vm_information is False:
            return information
```

lisa/node.py
```diff
@@ -113,6 +113,7 @@ def __init__(
         self.capture_azure_information: bool = False
         self.capture_kernel_config: bool = False
         self.has_checked_bash_prompt: bool = False
+        self.capture_vm_information: bool = True
```
squirrelsc: It doesn't need to save on node level. Keep it in platform runbook should be enough.

squirrelsc (dismissed):

squirrelsc (approved):

Issue #3086, opened 2023-12-01
Print only missing requirements or features in 'No available VM size' error spew


mcgov:
LISA can throw an error and skip deployment if the environment requirements are not met based on the runbook or test requirements. The current error message is super verbose and not helpful for debugging what happened. An example:

```
[...]

PR #3141, opened 2024-01-16
#3086

srivamsidandu:
Solved this issue by Printing only missing requirements or features in 'No available VM size' error spew


squirrelsc:
Cannot understand why this PR raised, and it's not related to the mentioned issuse.

PR #3142, opened 2024-01-16
Skip OFED install failure

kamalca:
No not treat a failure to install the OFED drivers as a test case failure.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -411,16 +411,19 @@ class Infiniband(Feature):
         if not self._is_legacy_device():
             extra_params += " --skip-unsupported-devices-check"
 
-        node.execute(
-            f"{self.resource_disk_path}/{ofed_folder}/mlnxofedinstall "
-            f"--add-kernel-support {extra_params} "
-            f"--tmpdir {self.resource_disk_path}/tmp",
-            expected_exit_code=0,
-            expected_exit_code_failure_message="SetupRDMA: failed to install "
-            "OFED Drivers",
-            sudo=True,
-            timeout=1200,
-        )
+        try:
+            node.execute(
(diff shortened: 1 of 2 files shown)


lisa/features/infiniband.py
```diff
@@ -411,16 +411,22 @@ def install_ofed(self) -> None:
...
+                f"--add-kernel-support {extra_params} "
+                f"--tmpdir {self.resource_disk_path}/tmp",
+                expected_exit_code=0,
+                expected_exit_code_failure_message="SetupRDMA: failed to install "
+                "OFED Drivers",
+                sudo=True,
+                timeout=1200,
+            )
+        except AssertionError as e:
+            raise UnsupportedKernelException(
```
LiliDeng: it will catch all the exceptions?

kamalca: Yes, it will raise UnsupportedKernelException instead of AssertionError for all non-zero exit codes from the OFED driver install command.

LiliDeng: it misleads us if the real exception doesn't belong to `UnsupportedKernelException`

kamalca: @LiliDeng Which exception should I use for this scenario: prerequisite installation failed, not the test itself. `MissingPackagesException`? `NotMeetRequirementException`?

LiliDeng: @squirrelsc any comments?

squirrelsc: `MissingPackagesException` provides more details than `NotMeetRequirementException`.

kamalca: Is this ready to be merged?

squirrelsc: LGTM

LiliDeng (approved):

PR #3143, opened 2024-01-16
Fixed the test result file path issue for kselftest under WSL

paulli2017:
The current test result file path is relative to local path of shell, which caused missing test result file issue when kselftest is done. The change is to set the output test file at the guest node directory, which is to meet the logic of checking test result file at later stage when kselftest is finished.

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest.py
+++ b/microsoft/testsuites/kselftest/kselftest.py
@@ -204,29 +204,34 @@ class Kselftest(Tool):
     ) -> List[KselftestResult]:
         # Executing kselftest as root may cause
         # VM to hang
+
+        # get username
+        username = self.node.tools[Whoami].get_username()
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -204,29 +203,28 @@ def run_all(
     ) -> List[KselftestResult]:
         # Executing kselftest as root may cause
         # VM to hang
+
+        # output test result
+        result_file_name = "kselftest-results.txt"
+        result_file = f"/home/{result_file_name}"
```
squirrelsc: Put the result file under user's folder, instead of root of /home.

squirrelsc:
It should be 1 commit. please rebase to main to remove extra 2 commits.

LiliDeng (approved):

PR #3146, opened 2024-01-19
Add posix judgement for checking bash prompt

lubaihua33:


squirrelsc:
@LiliDeng LGTM

lubaihua33:
We don't need to add this change. So close this PR.

PR #3148, opened 2024-01-22
Added support of kselftest from tar file source under WSL

paulli2017:
Kselftest from tarball file source failed under WSL situation due to the non-existed destination directory. The change is to create the target directory before the tarball file is copied to remote directory.

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest.py
+++ b/microsoft/testsuites/kselftest/kselftest.py
@@ -124,6 +124,8 @@ class Kselftest(Tool):
             )
 
         if self._tar_file_path:
+            mkdir = self.node.tools[Mkdir]
+            mkdir.create_directory(self._remote_tar_path.parent.as_posix())
             self.node.shell.copy(PurePath(self._tar_file_path), self._remote_tar_path)
             self.node.tools[Tar].extract(
                 str(self._remote_tar_path), str(self._kself_installed_dir), sudo=True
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -123,15 +123,17 @@ def _install(self) -> bool:
                 self.node.os, "kselftests in LISA does not support this os"
             )
 
+        # Create the kselftest installation directory
+        # for both tar file source and distro build.
+        mkdir = self.node.tools[Mkdir]
+        mkdir.create_directory(self._kself_installed_dir.as_posix())
```
squirrelsc: It needs to create directory in Tar tool. So, when Tar is used in other places, the non-exist dirs can be created too.

paulli2017: From my debug, Tar tool directory is the same as the parent of _kself_installed_dir. As I tried to re-use the current code, I just took advantage of the implementation of non-tar tool code and shared it among these two routes. I can change that if it causes issues to other test cases.

squirrelsc: From our design principles, tools should complete their goal with robust behavior. If the dest dir doesn't exist, it needs to create the dir.

paulli2017: sure, I will make the change

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3151, opened 2024-01-23
Add Windows variants of cp and unzip tools

anirudhrb:

PR #3152, opened 2024-01-24
Hyper-V platform

anirudhrb:
A new platform that provisions VMs on a Windows host using Hyper-V's PowerShell APIs.

### DIFF
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -63,7 +63,6 @@ if platform.system() == "Linux":
         print(f"libvirt package is not installed. [{e}]")
 
 try:
-    import lisa.sut_orchestrator.hyperv.features  # noqa: F401
     import lisa.sut_orchestrator.hyperv.platform_  # noqa: F401
     import lisa.sut_orchestrator.hyperv.schema  # noqa: F401
 except ModuleNotFoundError as e:
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -62,12 +62,10 @@ if platform.system() == "Linux":
     except ModuleNotFoundError as e:
         print(f"libvirt package is not installed. [{e}]")
 
-try:
-    import lisa.sut_orchestrator.hyperv.platform_  # noqa: F401
-    import lisa.sut_orchestrator.hyperv.schema  # noqa: F401
-except ModuleNotFoundError as e:
-    print(f"hyperv package is not installed. [{e}]")
+# Hyper-V platform
+import lisa.sut_orchestrator.hyperv.platform_  # noqa: F401
 
+# Transformers
 import lisa.transformers.dom0_kernel_installer  # noqa: F401
 import lisa.transformers.dump_variables  # noqa: F401
 import lisa.transformers.file_uploader  # noqa: F401
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -66,6 +66,7 @@ if platform.system() == "Linux":
 import lisa.sut_orchestrator.hyperv.platform_  # noqa: F401
 
 # Transformers
+import lisa.transformers.disable_cloud_components  # noqa: F401
 import lisa.transformers.dom0_kernel_installer  # noqa: F401
 import lisa.transformers.dump_variables  # noqa: F401
 import lisa.transformers.file_uploader  # noqa: F401
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -62,6 +62,13 @@ if platform.system() == "Linux":
     except ModuleNotFoundError as e:
         print(f"libvirt package is not installed. [{e}]")
 
+try:
+    import lisa.sut_orchestrator.hyperv.features  # noqa: F401
+    import lisa.sut_orchestrator.hyperv.platform_  # noqa: F401
+    import lisa.sut_orchestrator.hyperv.schema  # noqa: F401
+except ModuleNotFoundError as e:
+    print(f"hyperv package is not installed. [{e}]")
+
 import lisa.transformers.dom0_kernel_installer  # noqa: F401
 import lisa.transformers.dump_variables  # noqa: F401
 import lisa.transformers.file_uploader  # noqa: F401
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 from dataclasses import dataclass
-from pathlib import PureWindowsPath
+from pathlib import PurePath
 from typing import Optional
 
 from lisa import Node, RemoteNode
@@ -13,11 +13,11 @@ from lisa.util.process import Process
 class NodeContext:
     vm_name: str = ""
     host: Optional[RemoteNode] = None
-    working_path = PureWindowsPath()
+    working_path = PurePath()
     serial_log_process: Optional[Process] = None
 
     @property
-    def console_log_path(self) -> PureWindowsPath:
+    def console_log_path(self) -> PurePath:
         return self.working_path / f"{self.vm_name}-console.log"
 
 
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -14,7 +14,6 @@ class NodeContext:
     vm_name: str = ""
     host: Optional[RemoteNode] = None
     working_path = PureWindowsPath()
-    vhd_path = PureWindowsPath()
     serial_log_task_mgr: Optional["TaskManager[None]"] = None
 
     @property
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -12,7 +12,7 @@ from lisa.util.parallel import TaskManager
 @dataclass
 class NodeContext:
     vm_name: str = ""
-    server_node: Optional[RemoteNode] = None
+    host: Optional[RemoteNode] = None
     vhd_local_path = PurePosixPath()  # Local path on the machine where LISA is running
     vhd_remote_path = PureWindowsPath()  # Path on the hyperv server
     console_log_path = PureWindowsPath()  # Remote path with serial console output
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 from dataclasses import dataclass
-from pathlib import PurePosixPath, PureWindowsPath
+from pathlib import PureWindowsPath
 from typing import Optional
 
 from lisa import Node, RemoteNode
@@ -13,10 +13,13 @@ from lisa.util.parallel import TaskManager
 class NodeContext:
     vm_name: str = ""
     host: Optional[RemoteNode] = None
-    vhd_local_path = PurePosixPath()  # Local path on the machine where LISA is running
-    vhd_remote_path = PureWindowsPath()  # Path on the hyperv server
-    console_log_path = PureWindowsPath()  # Remote path with serial console output
+    working_dir = PureWindowsPath()
+    vhd_path = PureWindowsPath()
     serial_log_task_mgr: Optional["TaskManager[None]"] = None
+
+    @property
+    def console_log_path(self) -> PureWindowsPath:
+        return self.working_dir / f"{self.vm_name}-console.log"
 
 
 def get_node_context(node: Node) -> NodeContext:
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -6,7 +6,7 @@ from pathlib import PureWindowsPath
 from typing import Optional
 
 from lisa import Node, RemoteNode
-from lisa.util.parallel import TaskManager
+from lisa.util.process import Process
 
 
 @dataclass
@@ -14,7 +14,7 @@ class NodeContext:
     vm_name: str = ""
     host: Optional[RemoteNode] = None
     working_path = PureWindowsPath()
-    serial_log_task_mgr: Optional["TaskManager[None]"] = None
+    serial_log_process: Optional[Process] = None
 
     @property
     def console_log_path(self) -> PureWindowsPath:
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
 from dataclasses import dataclass
 from pathlib import PurePosixPath, PureWindowsPath
 from typing import Optional
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -1,12 +1,14 @@
 from dataclasses import dataclass
 from pathlib import PurePosixPath, PureWindowsPath
+from typing import Optional
 
-from lisa import Node
+from lisa import Node, RemoteNode
 
 
 @dataclass
 class NodeContext:
     vm_name: str = ""
+    server_node: Optional[RemoteNode] = None
     vhd_local_path = PurePosixPath()  # Local path on the machine where LISA is running
     vhd_remote_path = PureWindowsPath()  # Path on the hyperv server
     console_log_path = PureWindowsPath()  # Remote path with serial console output
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -13,13 +13,13 @@ from lisa.util.parallel import TaskManager
 class NodeContext:
     vm_name: str = ""
     host: Optional[RemoteNode] = None
-    working_dir = PureWindowsPath()
+    working_path = PureWindowsPath()
     vhd_path = PureWindowsPath()
     serial_log_task_mgr: Optional["TaskManager[None]"] = None
 
     @property
     def console_log_path(self) -> PureWindowsPath:
-        return self.working_dir / f"{self.vm_name}-console.log"
+        return self.working_path / f"{self.vm_name}-console.log"
 
 
 def get_node_context(node: Node) -> NodeContext:
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -1,8 +1,9 @@
 from dataclasses import dataclass
+from pathlib import PurePosixPath, PureWindowsPath
 
 
 @dataclass
 class NodeContext:
     vm_name: str = ""
-    vhd_local_path: str = ""  # Local path on the machine where LISA is running
-    vhd_remote_path: str = ""  # Path on the hyperv server
+    vhd_local_path = PurePosixPath()  # Local path on the machine where LISA is running
+    vhd_remote_path = PureWindowsPath()  # Path on the hyperv server
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -1,9 +1,16 @@
 from dataclasses import dataclass
 from pathlib import PurePosixPath, PureWindowsPath
 
+from lisa import Node
+
 
 @dataclass
 class NodeContext:
     vm_name: str = ""
     vhd_local_path = PurePosixPath()  # Local path on the machine where LISA is running
     vhd_remote_path = PureWindowsPath()  # Path on the hyperv server
+    console_log_path = PureWindowsPath()  # Remote path with serial console output
+
+
+def get_node_context(node: Node) -> NodeContext:
+    return node.get_context(NodeContext)
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -6,6 +6,7 @@ from pathlib import PurePosixPath, PureWindowsPath
 from typing import Optional
 
 from lisa import Node, RemoteNode
+from lisa.util.parallel import TaskManager
 
 
 @dataclass
@@ -15,6 +16,7 @@ class NodeContext:
     vhd_local_path = PurePosixPath()  # Local path on the machine where LISA is running
     vhd_remote_path = PureWindowsPath()  # Path on the hyperv server
     console_log_path = PureWindowsPath()  # Remote path with serial console output
+    serial_log_task_mgr: Optional["TaskManager[None]"] = None
 
 
 def get_node_context(node: Node) -> NodeContext:
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -1,0 +1,8 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class NodeContext:
+    vm_name: str = ""
+    vhd_local_path: str = ""  # Local path on the machine where LISA is running
+    vhd_remote_path: str = ""  # Path on the hyperv server
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -167,7 +167,6 @@ class HypervPlatform(Platform):
         vm_name_prefix = f"lisa-{test_suffix}"
 
         hv = self.server_node.tools[HyperV]
-        pwsh = self.server_node.tools[PowerShell]
         default_switch = hv.get_first_switch()
 
         if environment.runbook.nodes_requirement is None:
@@ -209,21 +208,7 @@ class HypervPlatform(Platform):
 
             self.server_node.shell.copy(node_context.vhd_local_path, remote_path)
             if is_zipped:
-                extraction_path = remote_path.parent.joinpath("extracted")
-                self.server_node.tools[PowerShell].run_cmdlet(
-                    f"Expand-Archive -Path {remote_path} "
-                    f"-DestinationPath {extraction_path} -Force"
-                )
-                extracted_vhd = self.server_node.tools[PowerShell].run_cmdlet(
-                    f"Get-ChildItem -Path {extraction_path} "
-                    f"| Select -First 1 -ExpandProperty Name"
-                )
-                extracted_vhd = str(extraction_path.joinpath(extracted_vhd))
-                pwsh.run_cmdlet(
-                    f"Copy-Item -Path {extracted_vhd} "
-                    f"-Destination {node_context.vhd_remote_path}"
-                )
-                self.server_node.shell.remove(remote_path)
+                self._unzip_vhd(node_context, remote_path)
 
             assert isinstance(node.capability.core_count, int)
             assert isinstance(node.capability.memory_mb, int)
@@ -258,6 +243,26 @@ class HypervPlatform(Platform):
             node.set_connection_info(
                 address=ip_addr, username=username, password=password
             )
+
+    def _unzip_vhd(
+        self, node_context: NodeContext, zipped_vhd_path: PureWindowsPath
+    ) -> None:
+        pwsh = self.server_node.tools[PowerShell]
+        extraction_path = zipped_vhd_path.parent.joinpath("extracted")
+        pwsh.run_cmdlet(
+            f"Expand-Archive -Path {zipped_vhd_path} "
+            f"-DestinationPath {extraction_path} -Force"
+        )
+        extracted_vhd = pwsh.run_cmdlet(
+            f"Get-ChildItem -Path {extraction_path} "
+            f"| Select -First 1 -ExpandProperty Name"
+        )
+        extracted_vhd = str(extraction_path.joinpath(extracted_vhd))
+        pwsh.run_cmdlet(
+            f"Copy-Item -Path {extracted_vhd} "
+            f"-Destination {node_context.vhd_remote_path}"
+        )
+        self.server_node.shell.remove(zipped_vhd_path)
 
     def _delete_environment(self, environment: Environment, log: Logger) -> None:
         self._delete_nodes(environment, log)
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -181,8 +181,6 @@ class HypervPlatform(Platform):
 
             vm_name = f"{vm_name_prefix}-{i}"
 
-            print(f"{i} {node_runbook} {node_space.core_count} {node_space.memory_mb}")
-
             node = environment.create_node_from_requirement(node_space)
             assert isinstance(node, RemoteNode)
             node.name = vm_name
@@ -206,7 +204,10 @@ class HypervPlatform(Platform):
                 )
                 is_zipped = True
 
+            log.debug("Copying VHD to server")
             self.server_node.shell.copy(node_context.vhd_local_path, remote_path)
+            log.debug("Finished copying VHD to server")
+
             if is_zipped:
                 self._unzip_vhd(node_context, remote_path)
 
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -141,3 +141,13 @@ class HypervPlatform(Platform):
             node.set_connection_info(
                 address=ip_addr, username=username, password=password
             )
+
+    def _delete_environment(self, environment: Environment, log: Logger) -> None:
+        self._delete_nodes(environment, log)
+
+    def _delete_nodes(self, environment: Environment, log: Logger) -> None:
+        hv = self.server_node.tools[HyperV]
+        for node in environment.nodes.list():
+            node_ctx = get_node_context(node)
+            log.debug(f"Deleting VM {node_ctx.vm_name}")
+            hv.delete_vm(node_ctx.vm_name)
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -74,6 +74,10 @@ class HypervPlatform(Platform):
         if not environment.runbook.nodes_requirement:
             return True
 
+        if len(environment.runbook.nodes_requirement) > 1:
+            log.error("Multiple nodes are currently not supported.")
+            return False
+
         nodes_capabilities = self._create_node_capabilities()
 
         nodes_requirement = []
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -1,8 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-import random
-import string
 from pathlib import PurePath, PureWindowsPath
 from typing import Any, List, Optional, Type, cast
 
@@ -182,8 +180,7 @@ class HypervPlatform(Platform):
         if environment.runbook.nodes_requirement is None:
             return  # nothing to deploy?
 
-        test_suffix = "".join(random.choice(string.ascii_uppercase) for _ in range(5))
-        vm_name_prefix = f"lisa-{test_suffix}"
+        vm_name_prefix = f"lisa-{environment.name}"
 
         hv = self._server.tools[HyperV]
         default_switch = hv.get_first_switch()
@@ -199,7 +196,7 @@ class HypervPlatform(Platform):
                 HypervNodeSchema, type(self).type_name()
             )
 
-            vm_name = f"{vm_name_prefix}-{i}"
+            vm_name = f"{vm_name_prefix}-n{i}"
 
             node = environment.create_node_from_requirement(node_space)
             assert isinstance(node, RemoteNode)
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -75,11 +75,6 @@ class HypervPlatform(Platform):
         return server_node
 
     def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
-        return self._configure_node_capabilities(environment, log)
-
-    def _configure_node_capabilities(
-        self, environment: Environment, log: Logger
-    ) -> bool:
         if not environment.runbook.nodes_requirement:
             return True
 
@@ -92,6 +87,9 @@ class HypervPlatform(Platform):
 
             requirement = node_space.generate_min_capability(nodes_capabilities)
             nodes_requirement.append(requirement)
+
+        if not self._check_host_capabilities(nodes_requirement, log):
+            return False
 
         environment.runbook.nodes_requirement = nodes_requirement
         return True
@@ -120,10 +118,10 @@ class HypervPlatform(Platform):
     def _check_host_capabilities(
         self,
         nodes_requirements: List[schema.NodeSpace],
-        host_capabilities: _HostCapabilities,
         log: Logger,
     ) -> bool:
         total_required_memory_mib = 0
+        host_capabilities = self._host_capabilities
 
         for node_requirements in nodes_requirements:
             # Calculate the total amount of memory required for all the VMs.
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -41,7 +41,7 @@ class HypervPlatform(Platform):
         self._host_capabilities = self._get_host_capabilities(self._log)
         self._source_vhd: Optional[PureWindowsPath] = None
         self._source_factory = Factory[Source](Source)
-        self._artifact_paths: Optional[List[PureWindowsPath]] = None
+        self._source_files: Optional[List[PureWindowsPath]] = None
 
     def _get_hyperv_runbook(self) -> HypervPlatformSchema:
         hyperv_runbook = self.runbook.get_extended_runbook(HypervPlatformSchema)
@@ -182,14 +182,14 @@ class HypervPlatform(Platform):
         return node_capabilities
 
     def _download_sources(self) -> None:
-        if self._artifact_paths:
+        if self._source_files:
             return
 
         if not self._hyperv_runbook.source:
             return
 
         source = self._source_factory.create_by_runbook(self._hyperv_runbook.source)
-        self._artifact_paths = source.download(self._server)
+        self._source_files = source.download(self._server)
 
     def _prepare_source_vhd(self, node_runbook: HypervNodeSchema) -> None:
         if self._source_vhd:
@@ -197,8 +197,8 @@ class HypervPlatform(Platform):
 
         if node_runbook.vhd and node_runbook.vhd.vhd_path:
             self._source_vhd = PureWindowsPath(node_runbook.vhd.vhd_path)
-        elif self._artifact_paths:
-            for artifact_path in self._artifact_paths:
+        elif self._source_files:
+            for artifact_path in self._source_files:
                 if artifact_path.suffix == ".vhd" or artifact_path.suffix == ".vhdx":
                     self._source_vhd = artifact_path
                     break
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -9,7 +9,7 @@ from lisa.environment import Environment
 from lisa.node import RemoteNode
 from lisa.platform_ import Platform
 from lisa.tools import Cp, HyperV, Mkdir, PowerShell
-from lisa.util import LisaException
+from lisa.util import LisaException, constants
 from lisa.util.logger import Logger, get_logger
 from lisa.util.subclasses import Factory
 
@@ -210,7 +210,8 @@ class HypervPlatform(Platform):
         if environment.runbook.nodes_requirement is None:
             return  # nothing to deploy?
 
-        vm_name_prefix = f"lisa-{environment.name}"
+        normalized_name = constants.NORMALIZE_PATTERN.sub("-", constants.RUN_NAME)
+        vm_name_prefix = f"{normalized_name}-e{environment.id}"
 
         hv = self._server.tools[HyperV]
         default_switch = hv.get_default_external_switch()
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -44,8 +44,6 @@ class HypervPlatform(Platform):
         hyperv_runbook = self.runbook.get_extended_runbook(HypervPlatformSchema)
         assert hyperv_runbook, "platform runbook cannot be empty"
         self._hyperv_runbook = hyperv_runbook
-        print("Hyperv platform initialize")
-        print(f"{self._hyperv_runbook}")
 
         if len(self._hyperv_runbook.servers) > 1:
             self._log.warning(
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -14,7 +14,7 @@ from lisa.tools import HyperV, PowerShell
 from lisa.util.logger import Logger, get_logger
 
 from .. import HYPERV
-from .context import get_node_context
+from .context import NodeContext, get_node_context
 from .features import StartStop
 from .schema import HypervNodeSchema, HypervPlatformSchema
 from .serial_console import SerialConsole, SerialConsoleLogger
@@ -231,8 +231,9 @@ class HypervPlatform(Platform):
             com1_pipe_name = f"{vm_name}-com1"
             com1_pipe_path = f"\\\\.\\pipe\\{com1_pipe_name}"
 
-            self.console_logger.start_logging(
-                com1_pipe_name, node_context.console_log_path
+            log.info(f"Serial logs at {node_context.console_log_path}")
+            node_context.serial_log_task_mgr = self.console_logger.start_logging(
+                com1_pipe_name, node_context.console_log_path, log
             )
 
             extra_args = {x.command: x.args for x in self._hyperv_runbook.extra_args}
@@ -266,6 +267,11 @@ class HypervPlatform(Platform):
         for node in environment.nodes.list():
             node_ctx = get_node_context(node)
             vm_name = node_ctx.vm_name
+
             log.debug(f"Deleting VM {vm_name}")
             hv.delete_vm(vm_name)
-            self.console_logger.stop_logging(f"{vm_name}-com1")
+
+            # The script that logs the serial console output exits gracefully
+            # on its own after the VM is deleted. So, wait for that to happen.
+            assert node_ctx.serial_log_task_mgr
+            node_ctx.serial_log_task_mgr.wait_for_all_workers()
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -15,7 +15,6 @@ from lisa.util.logger import Logger, get_logger
 
 from .. import HYPERV
 from .context import NodeContext, get_node_context
-from .features import StartStop
 from .schema import HypervNodeSchema, HypervPlatformSchema
 from .serial_console import SerialConsole, SerialConsoleLogger
 
@@ -39,7 +38,7 @@ class HypervPlatform(Platform):
 
     @classmethod
     def supported_features(cls) -> List[Type[feature.Feature]]:
-        return [StartStop, SerialConsole]
+        return [SerialConsole]
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         hyperv_runbook = self.runbook.get_extended_runbook(HypervPlatformSchema)
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -266,7 +266,7 @@ class HypervPlatform(Platform):
             com1_pipe_path = f"\\\\.\\pipe\\{com1_pipe_name}"
 
             log.info(f"Serial logs at {node_context.console_log_path}")
-            node_context.serial_log_task_mgr = self._console_logger.start_logging(
+            node_context.serial_log_process = self._console_logger.start_logging(
                 com1_pipe_name, node_context.console_log_path, log
             )
 
@@ -303,9 +303,9 @@ class HypervPlatform(Platform):
             )
 
     def _delete_environment(self, environment: Environment, log: Logger) -> None:
-        self._delete_nodes(environment)
+        self._delete_nodes(environment, log)
 
-    def _delete_nodes(self, environment: Environment) -> None:
+    def _delete_nodes(self, environment: Environment, log: Logger) -> None:
         def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
             hv = self._server.tools[HyperV]
             vm_name = node_ctx.vm_name
@@ -315,10 +315,12 @@ class HypervPlatform(Platform):
             else:
                 hv.delete_vm_async(vm_name)
 
-            # The script that logs the serial console output exits gracefully
-            # on its own after the VM is deleted. So, wait for that to happen.
-            assert node_ctx.serial_log_task_mgr
-            node_ctx.serial_log_task_mgr.wait_for_all_workers()
+            assert node_ctx.serial_log_process
+            result = node_ctx.serial_log_process.wait_result()
+            log.debug(
+                f"{vm_name} serial log process exited with {result.exit_code}. "
+                f"stdout: {result.stdout}"
+            )
 
         run_in_parallel(
             [
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 from functools import partial
-from pathlib import PurePath, PureWindowsPath
+from pathlib import PurePath
 from typing import Any, List, Optional, Type, cast
 
 from lisa import feature, schema, search_space
@@ -41,7 +41,7 @@ class HypervPlatform(Platform):
         self._hyperv_runbook = self._get_hyperv_runbook()
         self._server = self._initialize_server_node()
         self._host_capabilities = self._get_host_capabilities(self._log)
-        self._source_vhd: Optional[PureWindowsPath] = None
+        self._source_vhd: Optional[PurePath] = None
         self._source_factory = Factory[Source](Source)
         self._source_files: Optional[List[PurePath]] = None
 
@@ -200,11 +200,11 @@ class HypervPlatform(Platform):
             return
 
         if node_runbook.vhd and node_runbook.vhd.vhd_path:
-            self._source_vhd = PureWindowsPath(node_runbook.vhd.vhd_path)
+            self._source_vhd = PurePath(node_runbook.vhd.vhd_path)
         elif self._source_files:
             for artifact_path in self._source_files:
                 if artifact_path.suffix == ".vhd" or artifact_path.suffix == ".vhdx":
-                    self._source_vhd = PureWindowsPath(artifact_path)
+                    self._source_vhd = PurePath(artifact_path)
                     break
 
         if not self._source_vhd:
@@ -248,12 +248,12 @@ class HypervPlatform(Platform):
             node_context.vm_name = vm_name
             node_context.host = self._server
 
-            node_context.working_path = PureWindowsPath(
+            node_context.working_path = PurePath(
                 self._server.working_path / f"{vm_name}"
             )
 
             vm_vhd_name = f"{vm_name}{self._source_vhd.suffix}"
-            vhd_path = PureWindowsPath(node_context.working_path / f"{vm_vhd_name}")
+            vhd_path = PurePath(node_context.working_path / f"{vm_vhd_name}")
 
             self._server.tools[Mkdir].create_directory(str(node_context.working_path))
 
@@ -294,7 +294,7 @@ class HypervPlatform(Platform):
             )
 
     def _resize_vhd_if_needed(
-        self, vhd_path: PureWindowsPath, node_runbook: HypervNodeSchema
+        self, vhd_path: PurePath, node_runbook: HypervNodeSchema
     ) -> None:
         pwsh = self._server.tools[PowerShell]
         vhd_size = int(pwsh.run_cmdlet(f"(Get-VHD -Path {vhd_path}).Size"))
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -10,7 +10,7 @@ from lisa import feature, schema, search_space
 from lisa.environment import Environment
 from lisa.node import RemoteNode
 from lisa.platform_ import Platform
-from lisa.tools import HyperV, PowerShell
+from lisa.tools import Cp, HyperV, Ls, PowerShell, Unzip
 from lisa.util.logger import Logger, get_logger
 
 from .. import HYPERV
@@ -252,21 +252,18 @@ class HypervPlatform(Platform):
     def _unzip_vhd(
         self, node_context: NodeContext, zipped_vhd_path: PureWindowsPath
     ) -> None:
-        pwsh = self.server_node.tools[PowerShell]
         extraction_path = zipped_vhd_path.parent.joinpath("extracted")
-        pwsh.run_cmdlet(
-            f"Expand-Archive -Path {zipped_vhd_path} "
-            f"-DestinationPath {extraction_path} -Force"
+        self.server_node.tools[Unzip].extract(
+            str(zipped_vhd_path), str(extraction_path)
         )
-        extracted_vhd = pwsh.run_cmdlet(
-            f"Get-ChildItem -Path {extraction_path} "
-            f"| Select -First 1 -ExpandProperty Name"
-        )
-        extracted_vhd = str(extraction_path.joinpath(extracted_vhd))
-        pwsh.run_cmdlet(
-            f"Copy-Item -Path {extracted_vhd} "
-            f"-Destination {node_context.vhd_remote_path}"
-        )
+
+        extracted_files = self.server_node.tools[Ls].list(str(extraction_path))
+        assert len(extracted_files) == 1
+
+        extracted_vhd = PureWindowsPath(extracted_files[0])
+        extracted_vhd = extraction_path.joinpath(extracted_vhd)
+
+        self.server_node.tools[Cp].copy(extracted_vhd, node_context.vhd_remote_path)
         self.server_node.shell.remove(zipped_vhd_path)
 
     def _resize_vhd_if_needed(
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -77,6 +77,9 @@ class HypervPlatform(Platform):
         pwsh = self.server_node.tools[PowerShell]
         default_switch = hv.get_first_switch()
 
+        if environment.runbook.nodes_requirement is None:
+            return  # nothing to deploy?
+
         for i, node_space in enumerate(environment.runbook.nodes_requirement):
             node_runbook = node_space.get_extended_runbook(
                 HypervNodeSchema, type(self).type_name()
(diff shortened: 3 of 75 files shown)


squirrelsc:
Because it's a big PR, please make every commit meaningful.

lisa/mixin_modules.py
```diff
@@ -62,6 +62,12 @@
     except ModuleNotFoundError as e:
         print(f"libvirt package is not installed. [{e}]")
 
+try:
+    import lisa.sut_orchestrator.hyperv.platform_  # noqa: F401
+    import lisa.sut_orchestrator.hyperv.schema  # noqa: F401
```
squirrelsc: I think the schema file is referenced in platform, so it doesn't need to be listed here.

lisa/mixin_modules.py
```diff
@@ -62,6 +62,12 @@
...
+except ModuleNotFoundError as e:
+    print(f"hyperv package is not installed. [{e}]")
```
squirrelsc: There is no extra known Python package for HyperV, so this try/except should be removed.

lisa/tools/hyperv.py
```diff
@@ -156,6 +156,12 @@ def create_vm(
         if not is_ready:
             raise LisaException(f"VM {name} did not start")
 
+    def get_first_switch(self) -> str:
```
squirrelsc: 1. It may need to check if the first listed external switch.
1. Because it's used like "default" switch, so it should be `get_default_switch` too.
1. It may need more information from switches, so we need to define an object for switch.

lisa/tools/ls.py
```diff
@@ -99,7 +99,7 @@ def path_exists(self, path: str, sudo: bool = False) -> bool:
         return output.strip() == "True"
 
     def list(self, path: str, sudo: bool = False) -> List[str]:
-        command = f'Get-ChildItem -Path "{path}" `| Select-Object -ExpandProperty Name'
+        command = f'Get-ChildItem -Path "{path}" | Select-Object -ExpandProperty Name'
```
squirrelsc: There were some changes forth and back on how to encoding the string. Please make sure it's right way and no regression on other places.

lisa/util/shell.py
```diff
@@ -412,6 +412,7 @@ def mkdir(
         path_str = self._purepath_to_str(path)
         self.initialize()
         assert self._inner_shell
+
```
squirrelsc: remove the extra blank line.

lisa/sut_orchestrator/hyperv/serial_console.py
```diff
@@ -0,0 +1,79 @@
...
+        server_node = node_ctx.server_node
+        console_log_path = node_ctx.console_log_path
+        console_log_local_path = server_node.local_log_path / f"{vm_name}-console.log"
+        server_node.shell.copy_back(console_log_path, PurePath(console_log_local_path))
+
+        with open(console_log_local_path, mode="r", encoding="utf-8") as f:
+            log = f.read()
+
+        # Remove ANSI control codes.
+        log = re.sub("\x1b\\[[0-9;]*[mGKF]", "", log)
```
squirrelsc: Use `filter_ansi_escape` in util

lisa/sut_orchestrator/hyperv/schema.py
```diff
@@ -0,0 +1,37 @@
...
+class HypervPlatformSchema:
+    servers: List[HypervServer] = field(default_factory=list)
+    extra_args: List[ExtraArgs] = field(default_factory=list)
+
+
+@dataclass_json
+@dataclass
+class HypervNodeSchema:
+    hyperv_generation: int = 2
+    vhd: str = ""
```
squirrelsc: In Azure platform, we're moving to an unified image source object, called AzureImageSchema. Please add an `ImageSchema` in root `schema.py`, and inherit a `VHDImageSchema` in sut_orchestrator folder. So later, when Azure is refactoring, it can use this class.

lisa/tools/ls.py
anirudhrb: Do you recall the reason behind the back tick (`)? Because during my testing it just throws an error. "Get-ChildItem: A positional parameter cannot be found that accepts argument 'Select-Object'"

My understanding is that backtick is used for escaping things. Here, we don't want to escape `|`. It needs to be interpreted as the pipe operator instead of a literal `|`.

squirrelsc: It's used to escape `|`, because there was several layers of text wrappers. But after I encode the command to base64, it may not be a concern.

lisa/sut_orchestrator/hyperv/schema.py
anirudhrb: I don't have the full context on this. What should be the fields in `ImageSchema` and `VHDImageSchema`?

squirrelsc: The `ImageSchema` can be empty, and we will add more fields later like `cpu_arch`, `data_disk`. There may be another layer called `AzureImageSchema`, add data_path, vhd_gen. For `VhdAzureImageSchema`, it needs another file to support CVM testing. Check related classed in Azure to see how they are used.

anirudhrb:
> Because it's a big PR, please make every commit meaningful.

I have tried my best to make the commits meaningful and small. The commits show the full development history including some back and forth which explains the high number of commits. It will be a fairly high effort task for me to re-organize the commits further.

I have also raised some PRs separately like the changes in Hyper-V tool. So that diff no longer appears here. One such PR that is pending is https://github.com/microsoft/lisa/pull/3151. Once that is merged that diff disappears too. I'll try to merge more such changes outside the platform folder.

After that, the remaining diff should be largely in sut_orchestrator/hyperv and more manageable considering the code in that is heavily inspired from libvirt platform.

squirrelsc:
> > Because it's a big PR, please make every commit meaningful.
> 
> I have tried my best to make the commits meaningful and small. The commits show the full development history including some back and forth which explains the high number of commits. It will be a fairly high effort task for me to re-organize the commits further.
> 
> I have also raised some PRs separately like the changes in Hyper-V tool. So that diff no longer appears here. One such PR that is pending is #3151. Once that is merged that diff disappears too. I'll try to merge more such changes outside the platform folder.
> 
> After that, the remaining diff should be largely in sut_orchestrator/hyperv and more manageable considering the code in that is heavily inspired from libvirt platform.

ok, please not to resolve comments. I will use that to track of a comment fully addressed. If you need to track fixes, please use the emotion button.

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
```diff
@@ -0,0 +1,41 @@
...
+Write-Host "PipeName: $PipeName"
+Write-Host "OutFile: $OutFile"
+
+$oStream = New-Object System.IO.FileStream $OutFile, 'OpenOrCreate', 'Write', 'Read'
+$pClient = New-Object System.IO.Pipes.NamedPipeClientStream(".", $PipeName)
+$reconnect = $false
+
+try {
+    while (1) {
+        $timeout = 60000
```
squirrelsc: Make it configurable

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        $pClient.Connect($timeout)
+        if (!$pClient.IsConnected) {
+            Write-Host "Failed to connect to pipe. Exiting..."
+            exit
+        }
+
+        Write-Host "Connected to pipe"
+        $pClient.CopyTo($oStream)
```
squirrelsc, line 53: Is it a block operation?

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
```diff
@@ -0,0 +1,41 @@
...
+
+        Write-Host "Disconnected from pipe. Reconnecting..."
+        $reconnect = $true
```
squirrelsc, line 66: When will it be disconnected, if above is block operation?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+
+    @classmethod
+    def supported_features(cls) -> List[Type[feature.Feature]]:
+        return [SerialConsole]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        hyperv_runbook = self.runbook.get_extended_runbook(HypervPlatformSchema)
+        assert hyperv_runbook, "platform runbook cannot be empty"
+        self._hyperv_runbook = hyperv_runbook
+        print("Hyperv platform initialize")
```
squirrelsc: Don't use print, always use logger.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+        print(f"{self._hyperv_runbook}")
+
+        if len(self._hyperv_runbook.servers) > 1:
+            self._log.warning(
```
squirrelsc: info level is enough.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+                "Multiple servers are currently not supported. "
+                "Only the server host will be used."
+            )
+
+        server = self._hyperv_runbook.servers[0]
+        self.server_node = RemoteNode(
+            runbook=schema.Node(name="hyperv-server"),
+            index=-1,
+            logger_name="hyperv-server",
```
squirrelsc: "hyperv" is enough. Make prefix shorter to show more meaningful information without drag to right.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+            parent_logger=get_logger("hyperv-platform"),
```
squirrelsc: "hyperv"

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+        environment.runbook.nodes_requirement = nodes_requirement
+        return True
+
+    def _get_host_capabilities(self, log: Logger) -> _HostCapabilities:
+        host_cap = _HostCapabilities()
+
+        free_mem_bytes = self.server_node.tools[PowerShell].run_cmdlet(
+            "(Get-CimInstance -ClassName Win32_OperatingSystem).FreePhysicalMemory"
+        )
+        host_cap.free_memory_kib = int(free_mem_bytes) // 1024
```
squirrelsc: use mb is same to other places.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+            requirement = node_space.generate_min_capability(nodes_capabilities)
+            nodes_requirement.append(requirement)
+
+        environment.runbook.nodes_requirement = nodes_requirement
+        return True
+
+    def _get_host_capabilities(self, log: Logger) -> _HostCapabilities:
+        host_cap = _HostCapabilities()
+
+        free_mem_bytes = self.server_node.tools[PowerShell].run_cmdlet(
```
squirrelsc: Use free mem is not stable, because VM could by dynamic memory. As we planned, we support one VM in one server in first stage, so just hard code a value like 16 GB for now. Later. It needs to detect system memory, and assign VMs based on it.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+            "(Get-CimInstance -ClassName Win32_OperatingSystem).FreePhysicalMemory"
+        )
+        host_cap.free_memory_kib = int(free_mem_bytes) // 1024
+
+        lp_count = self.server_node.tools[PowerShell].run_cmdlet(
```
squirrelsc: Host information can be cached to prevent loading when VM created every time.

lisa/sut_orchestrator/hyperv/context.py
```diff
@@ -0,0 +1,23 @@
...
+from typing import Optional
+
+from lisa import Node, RemoteNode
+from lisa.util.parallel import TaskManager
+
+
+@dataclass
+class NodeContext:
+    vm_name: str = ""
+    server_node: Optional[RemoteNode] = None
```
squirrelsc: `host`

anirudhrb:
> > > Because it's a big PR, please make every commit meaningful.
> > 
> > 
> > I have tried my best to make the commits meaningful and small. The commits show the full development history including some back and forth which explains the high number of commits. It will be a fairly high effort task for me to re-organize the commits further.
> > I have also raised some PRs separately like the changes in Hyper-V tool. So that diff no longer appears here. One such PR that is pending is #3151. Once that is merged that diff disappears too. I'll try to merge more such changes outside the platform folder.
> > After that, the remaining diff should be largely in sut_orchestrator/hyperv and more manageable considering the code in that is heavily inspired from libvirt platform.
> 
> ok, please not to resolve comments. I will use that to track of a comment fully addressed. If you need to track fixes, please use the emotion button.

Sorry! I probably pressed the wrong button by mistake. I didn't mean to resolve anything here.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Hardcoding a value will place an artificial limit on memory. Even if only one VM is supported, it should be able to use as much memory on the host as possible and should not be limited to hardcoded value like "16 GB".

How about I add comment here calling out the limitations of this logic and say that this logic needs to be updated when supporting multiple VMs?

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Isn't warning more proper here since we are _warning_ the user that the way they have configured the runbook is not supported? Also, Libvirt platform uses warning for this.

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
anirudhrb: Yes.

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
anirudhrb: It gets briefly disconnected when the VM is rebooted.
I will add some comments to this file.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+        self, environment: Environment, log: Logger
+    ) -> bool:
+        if not environment.runbook.nodes_requirement:
+            return True
+
+        host_capabilities = self._get_host_capabilities(log)
+        nodes_capabilities = self._create_node_capabilities(host_capabilities)
+
+        nodes_requirement = []
+        for node_space in environment.runbook.nodes_requirement:
```
squirrelsc, line 89: return false, if there are multiple nodes_requirement.

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: Yes, you can specify a max capability of cpu/mem, and use it.

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: Warning is not a clear expectation, because it highly depends on if user see it or not. Either raise exception to stop run, or use info to log it.

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
squirrelsc: if it's a block operation, the loop doesn't need a sleep. if there is some case it's not blocked, only sleep, if the loop time is shorter than 1 second. It makes sure minimal loss of log.

anirudhrb: Which sleep are you referring to? This loop doesn't have any sleep. If we get disconnected, we try to immediately connect again.

squirrelsc: Oh, sorry. I mixed with other code snippets.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,295 @@
...
+                remote_path = PureWindowsPath(
+                    self.server_node.working_path / f"{vm_name}-vhd.zip"
+                )
+                is_zipped = True
+
+            log.debug("Copying VHD to server")
+            self.server_node.shell.copy(node_context.vhd_local_path, remote_path)
+            log.debug("Finished copying VHD to server")
+
+            if is_zipped:
```
squirrelsc: When will it be zipped? It should be handled in transformer, and unzip one time.

anirudhrb: The idea is to allow specifying a zipped VHD in the runbook. This speeds up the copy of the VHD to the host node. We will copy in zipped format and afterwards unzip it on the host node. `is_zipped` will be true if we detect a .zip extension in the VHD path provided in the runbook.

anirudhrb: Tested with our ~2GB VHD that, when zipped, comes down to ~300MB. So, copying over the network is much faster.

anirudhrb: I agree it should be one time. I will make the changes.

Not sure about transformer though. If this is a transformer its output will be a variable that gives the _remote_ VHD path. Then the platform schema needs to be modified to be able to accept a remote VHD path.

lisa/tools/ls.py
anirudhrb: Yeah, Base64 encoding probably eliminated the need for escaping.
Since this is an independent change I will extract it into a separate PR.

lisa/mixin_modules.py
anirudhrb: Done

lisa/util/shell.py
anirudhrb: Done

lisa/sut_orchestrator/hyperv/serial_console.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
anirudhrb: Done

lisa/sut_orchestrator/hyperv/serial_console_logger.ps1
anirudhrb: Added some comments.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: These prints were just for debugging. Removed.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: changed.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done

lisa/mixin_modules.py
anirudhrb: Fixed.

lisa/tools/ls.py
anirudhrb: https://github.com/microsoft/lisa/pull/3159

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: So, no changes are needed here, right? Host capabilities are already being set to the max cpu and max free mem available.

lisa/sut_orchestrator/hyperv/context.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Changed.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: I made it one time.

Waiting to hear further thoughts from you on the transformer. It will require runbook schema change.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+            if not node_space.check(nodes_capabilities):
+                return False
+
+            requirement = node_space.generate_min_capability(nodes_capabilities)
+            nodes_requirement.append(requirement)
+
+        environment.runbook.nodes_requirement = nodes_requirement
+        return True
+
+    def _init_host_capabilities(self, log: Logger) -> _HostCapabilities:
```
squirrelsc: `_init_host_capabilities` should be `_get_host_capabilities`, because it returns capability, and not save somewhere.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+    ) -> bool:
+        total_required_memory_mib = 0
+
+        for node_requirements in nodes_requirements:
+            # Calculate the total amount of memory required for all the VMs.
+            assert isinstance(node_requirements.memory_mb, int)
+            total_required_memory_mib += node_requirements.memory_mb
+
+        # Ensure host has enough memory for all the VMs.
+        if total_required_memory_mib > host_capabilities.free_memory_mib:
```
squirrelsc: The core count should be handled like this too. It needs to set 1:1 mapping between virtual core and actual logic cores. So, the perf is acceptable.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+        server_node.set_connection_info(
+            address=server.address, username=server.username, password=server.password
+        )
+
+        return server_node
+
+    def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
+        return self._configure_node_capabilities(environment, log)
+
+    def _configure_node_capabilities(
```
squirrelsc: This method is not shared and meaningful. The name is ambiguous to describe the purpose or executed logic. So, the content should be in `_prepare_environment`.

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: There is no such a transformer, but you can create one to extract zip file, and return a path to vhd variable. It doesn't need to change platform schema, because the changes happen on variables.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+            assert self._common_vhd_path
+
+            node.name = vm_name
+
+            node_context = get_node_context(node)
+            node_context.vm_name = vm_name
+            node_context.host = self.server_node
+
+            node_context.working_dir = PureWindowsPath(
+                self.server_node.working_path / f"{vm_name}"
```
squirrelsc: vm_name may be conflict, if it's specified in runbook. So the path should include test run part.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+            )
+
+            vm_vhd_name = f"{vm_name}.{self._common_vhd_path.suffix}"
+            node_context.vhd_path = PureWindowsPath(
+                self.server_node.working_path / f"{vm_vhd_name}"
```
squirrelsc: use above working_path from node_context.

lisa/sut_orchestrator/hyperv/context.py
```diff
@@ -0,0 +1,26 @@
...
+
+from lisa import Node, RemoteNode
+from lisa.util.parallel import TaskManager
+
+
+@dataclass
+class NodeContext:
+    vm_name: str = ""
+    host: Optional[RemoteNode] = None
+    working_dir = PureWindowsPath()
```
squirrelsc: working_path as other places.

lisa/sut_orchestrator/hyperv/context.py
```diff
@@ -0,0 +1,26 @@
...
+    vhd_path = PureWindowsPath()
```
squirrelsc: Why does it need to save vhd_path?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+
+    @classmethod
+    def supported_features(cls) -> List[Type[feature.Feature]]:
+        return [SerialConsole]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._hyperv_runbook = self._get_hyperv_runbook()
+        self.server_node = self._initialize_server_node()
+        self._host_capabilities = self._init_host_capabilities(self._log)
+        self._common_vhd_path: Optional[PureWindowsPath] = None
```
squirrelsc: Name it `_source_vhd`. It will include more information, when using VhdSchema.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+class _HostCapabilities:
+    def __init__(self) -> None:
+        self.core_count = 0
+        self.free_memory_mib = 0
+
+
+class HypervPlatform(Platform):
+    def __init__(
```
squirrelsc: If there is no extra logic, it doesn't need to rewrite `__init__`

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+    def type_name(cls) -> str:
+        return HYPERV
+
+    @classmethod
+    def supported_features(cls) -> List[Type[feature.Feature]]:
+        return [SerialConsole]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._hyperv_runbook = self._get_hyperv_runbook()
+        self.server_node = self._initialize_server_node()
```
squirrelsc: It's used in platform only, so it should be `_server`

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+    def _initialize_server_node(self) -> RemoteNode:
+        assert self._hyperv_runbook, "hyperv runbook cannot be empty"
+
+        if len(self._hyperv_runbook.servers) > 1:
+            self._log.info(
+                "Multiple servers are currently not supported. "
+                "Only the first server will be used."
+            )
+
+        server = self._hyperv_runbook.servers[0]
```
squirrelsc: `server_runbook`

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: It also needs to reserve some memory to host, like 2GB.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+
+        log.debug(
+            f"Host capabilities: {host_cap.core_count} cores, "
+            f"{host_cap.free_memory_mib} MiB free memory"
+        )
+
+        return host_cap
+
+    # Check that the VM requirements can be fulfilled by the host.
+    def _check_host_capabilities(
```
squirrelsc: Is this method used?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+                f"Host only has {host_capabilities.free_memory_mib} MiB free."
+            )
+            return False
+
+        return True
+
+    def _create_node_capabilities(self) -> schema.NodeSpace:
+        host_capabilities = self._host_capabilities
+        node_capabilities = schema.NodeSpace()
+        node_capabilities.name = "Hyper-V"
```
squirrelsc: Should it be used as VM name? It should be set to empty, or use an unique value, when vm is created.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+            is_allow_set=True,
+            items=[
+                schema.FeatureSettings.create(SerialConsole.name()),
+            ],
+        )
+
+        return node_capabilities
+
+    def _deploy_environment(self, environment: Environment, log: Logger) -> None:
+        self._deploy_nodes(environment, log)
```
squirrelsc: This wrapped method is not meaningful.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+            vhd_remote_path = self._unzip_vhd(vhd_remote_path)
+
+        self._common_vhd_path = vhd_remote_path
+
+    def _deploy_nodes(self, environment: Environment, log: Logger) -> None:
+        if environment.runbook.nodes_requirement is None:
+            return  # nothing to deploy?
+
+        test_suffix = "".join(random.choice(string.ascii_uppercase) for _ in range(5))
+        vm_name_prefix = f"lisa-{test_suffix}"
```
squirrelsc: Refer to Azure, generate vm name from the test env name or specified from runbook.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+
+        hv = self.server_node.tools[HyperV]
+        default_switch = hv.get_first_switch()
```
squirrelsc: It needs to create an external switch, if there is not. So a clean environment can be used to run cases too.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+
+        extra_args = {
+            x.command.lower(): x.args for x in self._hyperv_runbook.extra_args
+        }
+
+        self.console_logger = SerialConsoleLogger(self.server_node)
```
squirrelsc: name `_console_logger`

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+
+        for i, node_space in enumerate(environment.runbook.nodes_requirement):
+            node_runbook = node_space.get_extended_runbook(
+                HypervNodeSchema, type(self).type_name()
+            )
+
+            vm_name = f"{vm_name_prefix}-{i}"
```
squirrelsc: Follow Azure's name conversion.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,323 @@
...
+            self.server_node.tools[Cp].copy(
+                self._common_vhd_path, node_context.vhd_path
+            )
+
+            self._resize_vhd_if_needed(node_context.vhd_path, node_runbook)
+
+            assert isinstance(node.capability.core_count, int)
+            assert isinstance(node.capability.memory_mb, int)
+
+            com1_pipe_name = f"{vm_name}-com1"
```
squirrelsc, line 267: The name could be ```com1_serial_log = f"{vm_name}-serial-log"```

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: The unzip should happen _after_ copying to the server node. Which means the transformer will have to update the variable with the _remote path_ of the extracted VHD. So, the VHD field in runbook can be interpreted in two ways: it can be a _remote path_ if unzip transformer is used, else it will be _local path_. So there has to be some change in the schema to indicate whether the vhd path is remote or local.

squirrelsc: It doesn't need to be a remote path. In lisa, there are local node and remote node. When you operate a node, it's always local path. There is an uploader transformer already, but it needs to figure out how to work with host nodes, instead of vm nodes. Otherwise, it needs an unzip transformer. Another way is refer to BareMetal solution, use a build part, and copy images from a specified path, and execute other logic.

anirudhrb: It needs to be a remote path because that is the whole point of doing this. We want to copy the zipped file over the network and unzip it on the remote node. This way we copy a smaller size over the network thus saving time.

This is the sequence of operations we need:
1. Copy the zipped VHD to the *server node*. Let's say it is copied to `C:\Users\Administrator\lisatest\image.zip`. This path is on the server node i.e. it is a remote path.
2. Unzip the VHD _on the server node_. Let's say it got unzipped at `C:\Users\Administrator\lisatest\image.vhd`.
3. Now use `image.vhd` as the source VHD for all VMs. To do this, the platform needs to be aware of the path to `image.vhd` i.e. `C:\Users\Administrator\lisatest\image.vhd`.

As you can see in this example, the final VHD path is a remote path. If a transformer performs the above operations, it needs to modify the `$(vhd)` variable to have the *remote path* so that the platform is aware of where to find the source VHD.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: This is the name of the pipe corresponding to the COM port. IMO it doesn't make sense to add `-serial-log` to it. The name `{vm_name}-com1` says that this pipe corresponds to the COM1 port of the VM with name `vm_name`.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: removed.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Missed calling this method. Fixed now.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: changed.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: renamed.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: removed.

lisa/sut_orchestrator/hyperv/context.py
anirudhrb: changed.

lisa/sut_orchestrator/hyperv/context.py
anirudhrb: You're right we are not using the stored value later. Removed it.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,325 @@
...
+
+    def _get_host_capabilities(self, log: Logger) -> _HostCapabilities:
+        host_cap = _HostCapabilities()
+
+        free_mem_bytes = self._server.tools[PowerShell].run_cmdlet(
+            "(Get-CimInstance -ClassName Win32_OperatingSystem).FreePhysicalMemory"
+        )
+        host_cap.free_memory_mib = int(free_mem_bytes) // 1024
+
+        host_cap.free_memory_mib -= 2048  # reserve 2 GiB for host
```
squirrelsc: Because it reserves some to host, the name is not accurate. How about `allocable_memory_mib`?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,325 @@
...
+
+        log.debug(
+            f"Host capabilities: {host_cap.core_count} cores, "
+            f"{host_cap.free_memory_mib} MiB free memory"
+        )
+
+        return host_cap
+
+    # Check that the VM requirements can be fulfilled by the host.
+    def _check_host_capabilities(
```
squirrelsc: The method name is too general, call it like `_is_host_resources_enough`.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: There is no name field in the Hyper-V runbook schema.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done. Azure's name is using resource group. I have used environment.name instead of that.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Using existing external switch as discussed offline.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done.

lisa/tools/hyperv.py
anirudhrb: Done all three. Since we are looking for external switch I named it `get_default_external_switch`.

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: Ok. We will add name in runbook later. It's used to test with the existing VM.

anirudhrb: Run ID is already part of the `working_path`. What more do we need?

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: This is not per VM. Baremetal platform is also using a fixed name "baremetal".

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Introduced VHDSource in the latest iteration.

lisa/sut_orchestrator/hyperv/schema.py
anirudhrb: Please see the latest change after introducing VHDSource. As of now, we don't need any ImageSchema. We can add later on a need basis when we want to introduce metadata fields. No point in introducing empty & unused classes.

squirrelsc: Source/build is related to copy files. ImageSchema is to describe image metadata. The metadata is used to decide VM settings, when create vm, or a test case should be picked or not. We have some settings already, cpu arch, vhd gen, cvm... The empty class is to reduce cost of this change, it doesn't mean not needed.

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: Make sure environment is a part of the `working_path` too.

lisa/sut_orchestrator/hyperv/schema.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclass_json
+@dataclass
+class ExtraArgs:
+    command: str
+    args: str
+
+
+@dataclass_json()
+@dataclass
+class VHDSourceSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
```
squirrelsc: Name it `BuildSchema`, so it's more close to BareMetal. It doesn't need have vhd in the name.

lisa/sut_orchestrator/hyperv/schema.py
```diff
@@ -0,0 +1,52 @@
...
+@dataclass
+class LocalVHDSourceSchema(VHDSourceSchema):
+    vhd_path: str = ""
+    extract: bool = False
+
+
+@dataclass_json
+@dataclass
+class HypervPlatformSchema:
+    vhd_source: VHDSourceSchema
```
squirrelsc: `build`

anirudhrb: That name doesn't make sense here. The names in the schema should be such that its use is clear and obvious to the user. Naming it "vhd_source" makes it obvious to the user that they are supposed to specify the VHD details here. Simply calling it "build" gives no such hint to the user and is confusing.

squirrelsc: ok, we can call it source, the schema is like below. The source object should provide a method to query all destination files together.
 
```yaml
source
  type: 
  files:
    - source: # absolute or relative path, depends on implementation. It can be a name pattern or folder for multiple files.
      destination: # if it's relative path, the root is lisa working path.
      overwrite: true/false
      unzip: true/false
```

anirudhrb: 1. Is destination optional? For things like VHD, VMGS etc. destination should not be needed. The platform will manage the destination.
2. What values can `type` take? Will it be `local`, `ado`, `azure_storage` etc or something else?
3. From the above schema it looks like multiple files can be specified under `files`. If so, how will the platform know which one is VHD, VMGS or some other file?
4. I didn't quite understand this part: "source object should provide a method to query all destination files together". Do you have an example?

squirrelsc: > Is destination optional? For things like VHD, VMGS etc. destination should not be needed. The platform will manage the destination.

Yes, it's optional. If it's not provided, it's a folder called "sources" under the lisa *root* working path. 

> What values can type take? Will it be local, ado, azure_storage etc or something else?

Yes, it's the type of source. It's like other LISA subclasses. The whole `source` object is optional, if vhd paths are provided in the ImageSchema.

> From the above schema it looks like multiple files can be specified under files. If so, how will the platform know which one is VHD, VMGS or some other file?

It based on the file extension. We assume there is only one VHD and corresponding vmgs files in all specified files. The `source` object will be also used in other platforms in future.

> I didn't quite understand this part: "source object should provide a method to query all destination files together". Do you have an example?

I mean the `source` object in code runtime, it should have a property like `files`, which returns `List[str]`. So, when call source.files, it returns all destination files. It can be used to look for vhd file.

anirudhrb: > The whole `source` object is optional, if vhd paths are provided in the ImageSchema.

Okay, so if I want to specify a local vhd path, I don't need a `source`. I can specify in `ImageSchema`. But if I want to specify a local `.zip` file that contains the VHD, I would need to specify a `source` with `type: local` & `unzip: true` in the runbook. Right?

What happens if both are specified i.e. vhd path in ImageSchema and `source`? Should one take precedence, or should we error out?

squirrelsc: > What happens if both are specified i.e. vhd path in ImageSchema and source? Should one take precedence, or should we error out?

If vhd path is specified, it should be able to find from the source files. It's used to be clearer on which file is taken, but not required. If not, LISA should fail to see the vhd is not found from source.

all combinations of source and vhd in ImageSchema:

1. source: yes, vhd: yes. vhd should exist in source.files.
2. source: yes, vhd: no. platform detects vhd path from source.files.
3. source: no, vhd: yes. use vhd path directly. Assume it's in HyperV host already. (copied by other ways).
4. source: no, vhd: no. fail by lack of vhd path.

lisa/sut_orchestrator/hyperv/schema.py
anirudhrb: I guess this comment can be resolved since we discussed this topic in the other comment.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: environment name is part of vm name. `lisa-{environment.name}-n{index}`. This is like Azure platform: since we don't have resource group, I am using environment name.

lisa/sut_orchestrator/hyperv/schema.py
anirudhrb: Added.

lisa/sut_orchestrator/hyperv/schema.py
anirudhrb: Please see the latest changes.

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: use hyperv for all lower case. I will look at it later.

anirudhrb: changed.

lisa/sut_orchestrator/hyperv/schema.py
```diff
@@ -0,0 +1,69 @@
...
+@dataclass
+class ExtraArgs:
+    command: str
+    args: str
+
+
+@dataclass_json
+@dataclass
+class SourceFileSchema:
+    source: str = field(default="", metadata=field_metadata(required=True))
```
squirrelsc: Because it's required, it doesn't need to provide default value.

lisa/sut_orchestrator/hyperv/schema.py
```diff
@@ -0,0 +1,69 @@
...
+    # if destination is not provided by user, the source implementation will decide
+    # the destination.
+    destination: Optional[str] = None
+    unzip: bool = False
```
squirrelsc, line 35: We can set it to true by default. There is rare case where needs zipped files.

lisa/tools/hyperv.py
```diff
@@ -156,6 +164,18 @@ def create_vm(
         if not is_ready:
             raise LisaException(f"VM {name} did not start")
 
+    def get_default_external_switch(self) -> Optional[VMSwitch]:
```
squirrelsc: Move it to another PR.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,317 @@
...
+    def supported_features(cls) -> List[Type[feature.Feature]]:
+        return [SerialConsole]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._hyperv_runbook = self._get_hyperv_runbook()
+        self._server = self._initialize_server_node()
+        self._host_capabilities = self._get_host_capabilities(self._log)
+        self._source_vhd: Optional[PureWindowsPath] = None
+        self._source_factory = Factory[Source](Source)
+        self._artifact_paths: Optional[List[PureWindowsPath]] = None
```
squirrelsc: call it `_source_files`, so it's consistent with other places.

lisa/sut_orchestrator/hyperv/schema.py
anirudhrb: If the file is not a zip file, then no unzip is needed. So, unzip should be false.

lisa/tools/hyperv.py
anirudhrb: https://github.com/microsoft/lisa/pull/3170

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/schema.py
anirudhrb: fixed.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -0,0 +1,317 @@
...
+        self._delete_nodes(environment, log)
+
+    def _delete_nodes(self, environment: Environment, log: Logger) -> None:
+        hv = self._server.tools[HyperV]
+        for node in environment.nodes.list():
+            node_ctx = get_node_context(node)
+            vm_name = node_ctx.vm_name
+
+            log.debug(f"Deleting VM {vm_name}")
+            hv.delete_vm(vm_name)
```
squirrelsc: Make sure it's async deletion by default. It would be fast.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def _download_single_file(
+        self,
+        local_path: Path,
+        destination: Optional[str],
+        unzip: bool,
+        server: RemoteNode,
+    ) -> List[PureWindowsPath]:
+        if unzip:
+            return self._download_zipped_file(local_path, destination, server)
```
squirrelsc: The logic can be 1) copy, 2) unzip. So the copy logic is sharable, regardless unzip or not.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,190 @@
...
+
+        destination_is_dir = destination and (
```
squirrelsc, line 88: Use `is_file` of `ls` tool can check if it's a dir or file.

squirrelsc: Let's assume the destination is always a dir.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,190 @@
...
+
+        server.tools[Mkdir].create_directory(
+            str(dest_path) if destination_is_dir else str(dest_path.parent)
+        )
+
+        self._log.debug(f"Copying {local_path} to server")
+        server.shell.copy(local_path, dest_path)
+        self._log.debug("Finished copying.")
+
+        return [dest_path]
```
squirrelsc: Because this method is to return "single file", so don't return a list.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,190 @@
...
+        local_files = []
+        if local_path.is_dir():
+            local_files = [f for f in local_path.iterdir() if f.is_file()]
+        else:
+            local_files = [local_path]
+
+        downloaded_paths: List[PureWindowsPath] = []
+
+        for f in local_files:
+            downloaded_paths += self._download_single_file(
```
squirrelsc: Make sure it keeps original path, for example,
source is like 
```
dir1
   dir2
      file1
   file2
```
the dest should be same like,
```
destdir1
   dir2
      file1
   file2
```

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,190 @@
...
+
+        defined_destination = None
+        if destination:
+            if PureWindowsPath(destination).is_absolute():
+                defined_destination = PureWindowsPath(destination)
+            else:
+                defined_destination = PureWindowsPath(server.working_path / destination)
+
+        tmp_dest = PureWindowsPath(
+            server.working_path / "zipped_sources_tmp" / local_path.name
```
squirrelsc: Just copy the zip file with the same way as normal file. After that, unzip the file at current location.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,190 @@
...
+
+        return [final_destination]
+
+    def _unzip_file(
+        self,
+        server: RemoteNode,
+        zipped_file: PureWindowsPath,
+        destination_dir: Optional[PureWindowsPath] = None,
+    ) -> PureWindowsPath:
+        extraction_path = zipped_file.parent.joinpath(f"extracted-{uuid.uuid4()}")
```
squirrelsc: use the default path in the zip file. So the unzipped path is predictable.

anirudhrb: This is a temporary staging path for unzip. After unzip, it is copied to the final destination.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: Currently I'm not supporting nested directories & recursive copy. Let's add on need basis otherwise this is becoming unnecessarily complex.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: That single file can be a zip file which may have multiple files inside.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: If as per the above comment, we assume destination is always a directory, this may be simplified. I'll see.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: After assuming destination is a dir, the code has simplified. Please review again.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: After assuming destination is a dir, the code has simplified. Please review again.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: After assuming destination is a dir, the code has simplified. Please review again.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: You mean don't wait for it to finish? Or do it in parallel?

squirrelsc: Both. Like Azure, use a flag to control whether wait for deletion.

lisa/sut_orchestrator/hyperv/source.py
squirrelsc: use `remote_copy` tool, it should support Windows already.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,121 @@
...
+        if destination:
+            if PureWindowsPath(destination).is_absolute():
+                defined_destination = PureWindowsPath(destination)
+            else:
+                defined_destination = PureWindowsPath(server.working_path / destination)
+
+        if unzip:
+            dest_path = (
+                PureWindowsPath(server.working_path)
+                / "zipped_sources_tmp"
```
squirrelsc: Just put the zip file as its normal path. And unzip it there.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: `RemoteCopy` tool will need to be modified to return the list of destination file paths.

anirudhrb: Also, there seems to be a bug. The `RemoteCopy` tool calls `is_file()` of `Ls` tool to check if the source is a file. But the output of that command is not correctly captured by lisa. Here are the logs:

```
2024-02-12 07:40:54.317[139862791419456][DEBUG] lisa.local.cmd[2525] cmd: ['sh', '-c', 'command -v sudo'], cwd: None, shell: True, sudo: False, nohup: False, posix: True, remote: False, encoding: utf-8
2024-02-12 07:40:54.320[139863021057600][DEBUG] lisa.local.cmd[2525].stdout /usr/bin/sudo                                         
2024-02-12 07:40:54.320[139862791419456][DEBUG] lisa.local.cmd[2525] execution time: 0.002 sec, exit code: 0
2024-02-12 07:40:54.320[139862791419456][DEBUG] lisa.local.cmd[1136] cmd: ['sudo', 'sh', '-c', 'ls -al /home/anirudh/vhds/l1vh/dom0-2401.3.0-ci.86884940-l1vh.vhdx.zip'], cwd: None, shell: True, sudo: True, nohup: False, posix: True, remote: False, encoding: utf-8                                                                                                                               
-rw-rw-r-- 1 anirudh anirudh 386787215 Jan  4 12:14 /home/anirudh/vhds/l1vh/dom0-2401.3.0-ci.86884940-l1vh.vhdx.zip
2024-02-12 07:40:55.267[139862791419456][DEBUG] lisa.local.cmd[1136] execution time: 0.947 sec, exit code: 0 
```

Notice how the output log of the `ls` command is not prefixed with timestamp etc. like the other logs (e.g. `2024-02-12 07:40:54.320[139863021057600][DEBUG] lisa.local.cmd[2525].stdout`). So this output is not captured by lisa and `result.stdout` is empty. Due to this `is_file()` return `False`.

If I pass `sudo=False` to `is_file()`, it works fine. The problem is only with `sudo=True`.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Looks like we don't have a `run_cmdlet_async()` in PowerShell tool, so that needs to be implemented first.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: https://github.com/microsoft/lisa/pull/3172

lisa/sut_orchestrator/hyperv/source.py
squirrelsc: I will take a look. The local command has different behaviors than remote.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: https://github.com/microsoft/lisa/pull/3177

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,114 @@
...
+        return SourceSchema
+
+    def download(self, server: RemoteNode) -> List[PureWindowsPath]:
+        raise NotImplementedError()
+
+
+class LocalSource(Source):
+    def __init__(self, runbook: LocalSourceSchema) -> None:
+        super().__init__(runbook)
+        self.local_runbook: LocalSourceSchema = self.runbook
```
squirrelsc: use `_local_runbook`, because it's a private variable.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,114 @@
...
+        self._log = get_logger("local", self.__class__.__name__)
+        self.downloaded_files: Optional[List[PureWindowsPath]] = None
```
squirrelsc: This is not used out of `download` method. so it should be a variable in the method.

lisa/sut_orchestrator/hyperv/source.py
squirrelsc: The fix PR is merged, please try remote_copy. It can share more logic and handle future scenarios.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: This is to cache the result and avoid downloading multiple times. If this is not `None`, it means we have already downloaded the files and can return them directly.

anirudhrb: Done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: Done.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,81 @@
...
+class Source(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
+    def __init__(self, runbook: SourceSchema) -> None:
+        super().__init__(runbook=runbook)
+        self._log = get_logger("source", self.__class__.__name__)
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return SourceSchema
+
+    def download(self, server: RemoteNode) -> List[PureWindowsPath]:
```
squirrelsc: use `PurePath`, in future, we will consolidate source classes.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,81 @@
...
+        raise NotImplementedError()
+
+
+class LocalSource(Source):
+    def __init__(self, runbook: LocalSourceSchema) -> None:
+        super().__init__(runbook)
+        self._local_runbook: LocalSourceSchema = self.runbook
```
squirrelsc: The runbook is saved in `subclasses.BaseClassWithRunbookMixin`. When it's needed, get it from `self._runbook` with the right type.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,81 @@
...
+        self._log = get_logger("local", self.__class__.__name__)
```
squirrelsc: The logger name "local" is confusing. It means where LISA runs in other places. Pass a parent logger in, and give a subname like "source" in the `Source` base class.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _download_file(
+        self, file: SourceFileSchema, server: RemoteNode
+    ) -> List[PureWindowsPath]:
+        local_path = PurePath(file.source)
+
+        downloaded_paths: List[PurePath] = []
+
+        if file.destination:
+            if PureWindowsPath(file.destination).is_absolute():
```
squirrelsc: Use `server.get_pure_path`

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,81 @@
...
+                dest = PureWindowsPath(server.working_path) / file.destination
+        else:
+            dest = PureWindowsPath(server.working_path) / "sources"
+
+        downloaded_paths = server.tools[RemoteCopy].copy_to_remote(
+            local_path, dest, recurse=True
+        )
+
+        if file.unzip:
+            assert len(downloaded_paths) == 1
```
squirrelsc: Add a message, once the assert is failed.

anirudhrb: done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,84 @@
...
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return SourceSchema
+
+    def download(self, server: RemoteNode) -> List[PurePath]:
+        raise NotImplementedError()
+
+
+class LocalSource(Source):
+    def __init__(self, runbook: LocalSourceSchema, **kwargs: Any) -> None:
```
squirrelsc: the `__init__` method do nothing now, it can be deleted totally.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -0,0 +1,84 @@
...
+from lisa.util.logger import get_logger
+
+from .schema import LocalSourceSchema, SourceFileSchema, SourceSchema
+
+
+class Source(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
+    def __init__(self, runbook: SourceSchema, **kwargs: Any) -> None:
+        super().__init__(runbook=runbook)
+        self._log = get_logger(
+            "source", self.__class__.__name__, parent=kwargs.get("parent_logger")
```
squirrelsc: parent could be moved to `__init__` like `def __init__(self, runbook: SourceSchema, parent_logger: Logger, **kwargs: Any) -> None:`

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -202,7 +204,7 @@ def _prepare_source_vhd(self, node_runbook: HypervNodeSchema) -> None:
         elif self._source_files:
             for artifact_path in self._source_files:
                 if artifact_path.suffix == ".vhd" or artifact_path.suffix == ".vhdx":
-                    self._source_vhd = artifact_path
+                    self._source_vhd = PureWindowsPath(artifact_path)
```
squirrelsc: All used place can be PurePath. `PureWindowsPath` is used only when initializing.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: This class is created by factory. So, we will need to add it in few other places too like `create_by_runbook()`, `create_with_runbook()` etc...

squirrelsc: The `create_by_runbook` has `**kwargs: Any`, so it should work.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: Done. Using PurePath everywhere.

lisa/sut_orchestrator/hyperv/source.py
anirudhrb: done.

lisa/sut_orchestrator/hyperv/source.py
```diff
@@ -1,25 +1,23 @@
...
 from lisa.util import InitializableMixin, subclasses
-from lisa.util.logger import get_logger
+from lisa.util.logger import Logger, get_logger
 
 from .schema import LocalSourceSchema, SourceFileSchema, SourceSchema
 
 
 class Source(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
-    def __init__(self, runbook: SourceSchema, **kwargs: Any) -> None:
+    def __init__(self, runbook: SourceSchema, parent_logger: Logger) -> None:
```
squirrelsc: Keep `**kwargs: Any` here and when calling super(). So if other functionality implemented in base classes, it won't break.

anirudhrb: Added. It is also worth noting that the `Source` in baremetal platform doesn't have `kwargs` and someone needs to add it there too.

squirrelsc: @LiliDeng Please add kwargs in BareMetal Typed schemas.

squirrelsc:
@LiliDeng LGTM. Please remember to use squash merge.

LiliDeng (approved):

PR #3154, opened 2024-01-25
upgrade paramiko version to fix vulnerability

lesscodingmorehappiness:


microsoft/runbook/tiers/tier.yml
```diff
@@ -4,7 +4,7 @@ extension:
   - ../../testsuites
 variable:
   - name: tier
-    value: 0
+    value: 1
```
squirrelsc: Revert this change.

pyproject.toml
```diff
@@ -12,10 +12,12 @@ classifiers = [
     "Topic :: Software Development :: Testing",
 ]
 dependencies = [
+    "temppathlib >= 1.2.0",
```
squirrelsc: Don't use `>=`. Check if both temppathlib and icontract should be in the list or not.

lisa/util/shell.py
```diff
@@ -31,6 +51,638 @@
...
+        confirm: bool = True,
+    ):
+        """See paramiko.SFTP documentation."""
+        return self.__wrap(method=lambda sftp: sftp.put(localpath, remotepath, callback, confirm))
+
+    def get(self, remotepath: str, localpath: str, callback: Optional[Callable[[int, int], Any]] = None):
+        """See paramiko.SFTP documentation."""
+        return self.__wrap(method=lambda sftp: sftp.get(remotepath, localpath, callback))
+
+class SshShellWithSFTP(icontract.DBC):
```
squirrelsc, line 204: What this is used to fix? Can we just include the core parts?

lesscodingmorehappiness: yes I'm trying to only include the functions being used in lisa, I don't want to break current lisa functions or introduce unused contents.

BTW please ignore this draft for now I'm trying to fix some linting alerts

squirrelsc:
Fixed by upgrading spurplus.

PR #3156, opened 2024-01-27
Case insensitive sku family check

kamalca:
Make the checks for sku family case insensitive to avoid bugs.

Some of the SKU families don't follow capitalization conventions (ie. "standardDCEDV5Family" has a capital V). I figured it would be better to do a case-insensitive approach.

Note: I needed to also update the api version to get 'NonpersistedTPM' (Stateless TDX) working again.


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3157, opened 2024-01-27
Azure: Use runbook locations as allowed_locations

pupacha:
set `locations` attribute of `AzurePlatformSchema` as default `allowed_locations` when `location` field is not provided in runbook.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -366,8 +366,14 @@ class AzurePlatformSchema:
                     "should be specified either both or not."
                 )
 
-        if not self.locations:
-            self.locations = LOCATIONS
+        self.locations_list: List[str] = []
+        if self.locations:
+            if isinstance(self.locations, str):
+                self.locations_list = [self.locations]
+            else:
+                self.locations_list = self.locations
+        else:
+            self.locations_list = LOCATIONS
 
     @property
     def cloud(self) -> Cloud:
@@ -518,7 +524,9 @@ class AzurePlatform(Platform):
         is_success: bool = False
 
         # get eligible locations
-        allowed_locations = _get_allowed_locations(nodes_requirement)
+        allowed_locations = _get_allowed_locations(
+            nodes_requirement, self._azure_runbook
+        )
         log.debug(f"allowed locations: {allowed_locations}")
 
         # Any to wait for resource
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2963,7 +2973,7 @@ def _get_allowed_locations(nodes_requirement: List[schema.NodeSpace]) -> List[st
         existing_locations = existing_locations_str.split(",")
         existing_locations = [x.strip() for x in existing_locations]
     else:
-        existing_locations = LOCATIONS[:]
+        existing_locations = azure_runbook.locations_list
```
squirrelsc, line 2976: What's difference with this change or not?

pupacha: In the case when 'location' is not passed under azure requirement, with this change, we will now be able to pass list of locations to `locations` attribute of `AzurePlatformSchema`, which LISA will be using. See  https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/platform_.py#L521.

``` 
platform:
  - type: azure
    .
    .
    .
    azure:
      resource_group_name: $(resource_group)
      locations: $(azure_locations_list)
    requirement:
      .
      .
      .
      azure:
        # location: $(location)
        vm_size: $(vm_size)
```

With this change, above runbook will be rightfully supported. Without this change, LISA will not use passed `locations` attribute anywhere and uses the default set `LOCATIONS` at https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/platform_.py#L2966

squirrelsc: Let's just remove locations from platform level. It looks confusing, and not used.

pupacha: We want to use this parameter from the platform level. With a single location input at requirements level, there are chances for quota issues from Azure. `locations` from platform level provides a way to give a set of locations, not worrying about the quota, as LISA would move onto the next location from the `locations` list.
Also, this parameter is present from the beginning, just that lisa was not doing what it is intended to do. If the name looks confusing, can we rename it to something like `locations_set` or `allowed_locations` ?

squirrelsc: As above lines, the locations of requirements support multiple locations too. The default list is needed when node requirement doesn't specify location.

pupacha: I couldn't actually get it. The `location` of Azure Requirement expects just an str https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/common.py#L260 . If there is already a way to pass list of locations in runbook, could you please provide a sample runbook?

pupacha: With fix from https://github.com/microsoft/lisa/pull/3162, `location` of Azure Requirement can take comma separated string and addresses the usecase of passing multiple locations.

pupacha:
https://github.com/microsoft/lisa/pull/3162 will let `location` take comma separated strings as input.

PR #3158, opened 2024-01-29
Changed the timeout number

paulli2017:
test for kernel validation pipeline test

### DIFF
--- a/lisa/sut_orchestrator/azure/Enable-SSH.ps1
+++ b/lisa/sut_orchestrator/azure/Enable-SSH.ps1
@@ -31,6 +31,11 @@ $acl | Set-Acl
 Start-Service sshd
 
 # enable key authentication and restart for effective.
-$sshd_config = "C:\ProgramData\ssh\sshd_config" 
-(Get-Content $sshd_config) -replace '#PubkeyAuthentication', 'PubkeyAuthentication' | Out-File -encoding ASCII $sshd_config
+$ssh_server = Get-Service sshd | Where-Object { $_.Name -like 'sshd' -and $_.Status -like "Running" }
+if ($ssh_server)
+{
+    Write-Output "sshd server is running"
+    $sshd_config = "C:\ProgramData\ssh\sshd_config"
+    (Get-Content $sshd_config) -replace '#PubkeyAuthentication', 'PubkeyAuthentication' | Out-File -encoding ASCII $sshd_config
+}
 Restart-Service sshd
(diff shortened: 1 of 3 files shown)


lisa/schema.py
```diff
@@ -1577,7 +1577,7 @@ class Runbook:
     tags: Optional[List[str]] = None
     concurrency: int = 1
     # minutes to wait for resource
-    wait_resource_timeout: float = 5
+    wait_resource_timeout: float = 10
```
squirrelsc: You can configure it in runbook. And don't mention internal work in GitHub.

lisa/sut_orchestrator/azure/Enable-SSH.ps1
```diff
@@ -31,6 +31,11 @@ $acl | Set-Acl
 Start-Service sshd
 
 # enable key authentication and restart for effective.
-$sshd_config = "C:\ProgramData\ssh\sshd_config" 
-(Get-Content $sshd_config) -replace '#PubkeyAuthentication', 'PubkeyAuthentication' | Out-File -encoding ASCII $sshd_config
+$ssh_server = Get-Service sshd | Where-Object { $_.Name -like 'sshd' -and $_.Status -like "Running" }
+if ($ssh_server)
```
squirrelsc, line 35: What does it mean, if ssh_server is not here? It may cause key auth changes are skipped.

lisa/tools/reboot.py
```diff
@@ -149,7 +149,7 @@ def command(self) -> str:
     def _check_exists(self) -> bool:
         return True
 
-    def reboot(self, time_out: int = 600) -> None:
+    def reboot(self, time_out: int = 1200) -> None:
```
squirrelsc, line 152: It doesn't need to change the default value. Specify the value when it's called.

lisa/tools/reboot.py
```diff
@@ -174,7 +174,7 @@ def reboot(self, time_out: int = 600) -> None:
                         constants.ENVIRONMENTS_NODES_REMOTE_PORT
                     ],
                     log=self._log,
-                    timeout=20,
+                    timeout=1200,
```
squirrelsc: 20 should be enough, in this loop it will try multiple times.

LiliDeng:
@paulli2017 hi, are you planning to update this pull request?

paulli2017:
This change is not needed anymore

PR #3160, opened 2024-01-29
Skip if irqbalance is not installed

kamalca:


microsoft/testsuites/network/sriov.py
```diff
@@ -694,6 +695,10 @@ def verify_irqbalance(self, environment: Environment) -> None:
         server_node = cast(RemoteNode, environment.nodes[0])
         client_node = cast(RemoteNode, environment.nodes[1])
 
+        ls = server_node.tools[Ls]
+        if not ls.path_exists("/etc/default/irqbalance"):
+            raise SkippedException("irqbalance is not installed, skip this test case.")
```
squirrelsc: Do you have reasons, why irqbalance is not installed? It may not be completed for all cases, but add them helping anyway.

LiliDeng:
@kamalca please double check for different case, in mariner, the config file is under /etc/sysconfig/irqbalance, thank you.

kamalca:
Thank you for the thoughtful comments. I've implemented what I believe is a better implementation. It runs `irqbalance --debug` directly can checks the output of the command, instead of relying on systemd configuration files which can be different in different distros.

Let me know if you have any concerns about this implementation.

microsoft/testsuites/network/sriov.py
```diff
@@ -694,10 +695,23 @@ def verify_irqbalance(self, environment: Environment) -> None:
...
+
+        # Get the irqbalance version if we can
+        if isinstance(server_node.os, Posix):
+            try:
+                environment.log.debug(
+                    "irqbalance version: "
+                    f"{server_node.os.get_package_information('irqbalance')}"
+                )
+            except Exception as e:
+                print(e)
```
LiliDeng: don't use print

microsoft/testsuites/network/sriov.py
```diff
@@ -694,10 +695,23 @@ def verify_irqbalance(self, environment: Environment) -> None:
         server_node = cast(RemoteNode, environment.nodes[0])
         client_node = cast(RemoteNode, environment.nodes[1])
 
-        server_node.tools[Sed].append(
-            "IRQBALANCE_ARGS='--debug'", "/etc/default/irqbalance", True
-        )
-        server_node.tools[Service].restart_service("irqbalance")
+        if server_node.execute("command -v irqbalance", shell=True).exit_code != 0:
```
LiliDeng: in sles, it needs sudo permission

microsoft/testsuites/network/sriov.py
```diff
@@ -694,10 +695,27 @@ def verify_irqbalance(self, environment: Environment) -> None:
...
+                "command -v irqbalance", shell=True, sudo=True
+            ).exit_code
+            != 0
+        ):
+            raise SkippedException("irqbalance is not installed")
+
+        # Get the irqbalance version if we can
+        if isinstance(server_node.os, Posix):
+            try:
+                environment.log.debug(
```
LiliDeng: use log: Logger in the test case definition

microsoft/testsuites/network/sriov.py
```diff
@@ -679,9 +679,10 @@ def verify_sriov_ethtool_offload_setting(self, environment: Environment) -> None
         this is the case.
 
         Steps,
-        1. Set irqbalance logging level to debug.
-        2. Generate some network traffic.
-        3. Check check logs for “Selecting irq xxx for rebalancing”.
+        1. Stop irqbalance service.
+        2. Start irqbalance as a background process with debug mode.
+        3. Generate some network traffic.
+        4. Check check irqbalance output for “Selecting irq xxx for rebalancing”.
```
LiliDeng: ```4. Check irqbalance output for "Selecting irq xxx for rebalancing".```

microsoft/testsuites/network/sriov.py
```diff
@@ -694,10 +695,27 @@ def verify_irqbalance(self, environment: Environment) -> None:
...
+        ):
+            raise SkippedException("irqbalance is not installed")
+
+        # Get the irqbalance version if we can
+        if isinstance(server_node.os, Posix):
+            try:
+                environment.log.debug(
+                    "irqbalance version: "
+                    f"{server_node.os.get_package_information('irqbalance')}"
+                )
```
LiliDeng, line 712: can irqbalance --version give the version we want?

microsoft/testsuites/network/sriov.py
kamalca: Changed.

microsoft/testsuites/network/sriov.py
kamalca: `--version` doesn't give build info

kamalca: For example, the original irqbalance issue was seen in `1.8.0+1build1` but not `1.8.0-1ubuntu0.1`. If we ran `irqbalance --version` they would both return 1.8.0.

LiliDeng (approved):

PR #3162, opened 2024-02-01
Azure: fix error on resolving image version with multiple locations.

squirrelsc:
The original code hasn't use the latest runbook to resolve image. So the location is not set to single value.


LiliDeng (dismissed):

pupacha (dismissed):

pupacha:
Thanks for this fix @squirrelsc

squirrelsc:
@LiliDeng, please help merging it.

PR #3163, opened 2024-02-01
Support Bicep

LiliDeng:
I have tested for different scenarios in internal pipelines.


LiliDeng:
I have tested for different scenarios in internal pipelines.

lisa/sut_orchestrator/azure/bicepconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "analyzers": {
+    "core": {
+      "rules": {
+        "no-unused-params": {
```
squirrelsc, line 5: Is there any unused params? They should be removed.

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -0,0 +1,317 @@
...
+  plan: nodes[i].purchase_plan
+  properties: {
+    availabilitySet: (use_availability_set ? getAvailabilitySetId(availability_set_name_value) : null)
+    hardwareProfile: {
+      vmSize: nodes[i].vm_size
+    }
+    osProfile: (((!empty(nodes[i].vhd)) && (!empty(nodes[i].vhd.vmgs_path))) ? null : getOsProfile(nodes[i].short_name, admin_username, (!empty(admin_password)), admin_password, ((!empty(admin_key_data)) && nodes[i].is_linux), getLinuxConfiguration('/home/${admin_username}/.ssh/authorized_keys', admin_key_data, empty(admin_password))))
+    storageProfile: {
+      imageReference: (((!empty(nodes[i].vhd)) && (!empty(nodes[i].vhd.vmgs_path))) ? null : (((!empty(nodes[i].vhd)) && (!empty(nodes[i].vhd.vhd_path))) ? getOsDiskVhd(nodes[i].name) : ((!empty(nodes[i].shared_gallery)) ? getOsDiskSharedGallery(nodes[i].shared_gallery) : getOsDiskMarketplace(nodes[i]))))
+      osDisk: (((!empty(nodes[i].vhd)) && (!empty(nodes[i].vhd.vmgs_path))) ? getOSDisk('${nodes[i].name}-disk') : ((nodes[i].os_disk_type == 'Ephemeral') ? getEphemeralOSImage(nodes[i]) : getOSImage(nodes[i])))
```
squirrelsc, line 287: For those logic, put them in new function. So the template is straightforward.

squirrelsc: Is it possible to use functions for those logic?

LiliDeng: I have tried to use condition ? a : b in function, seems it doesn't work. I will try again.

lisa/sut_orchestrator/azure/bicepconfig.json
LiliDeng: There are no unused params.
This config file is for using experimental feature - functions, other contents are default ones. I can change the level into error.

LiliDeng: I changed the level into error

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: I put ```((!empty(nodes[i].vhd)) && (!empty(nodes[i].vhd.vmgs_path)))``` into func

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -0,0 +1,323 @@
...
+    vTpmEnabled: true
+  }
+  securityType: security_profile.security_type
+}
+
+func getAvailabilitySetId(availability_set_name string) object => {
+  id: resourceId('Microsoft.Compute/availabilitySets', availability_set_name)
+}
+
+func hasVmgsPath(node object) bool => bool((!empty(node.vhd)) && (!empty(node.vhd.vmgs_path)))
```
squirrelsc: It should be called `isCvm`. The function names can be used to explain the business logic.

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -0,0 +1,323 @@
...
+  name: nodes[i].name
+  location: nodes[i].location
+  tags: vm_tags
+  plan: nodes[i].purchase_plan
+  properties: {
+    availabilitySet: availability_set_value
+    hardwareProfile: {
+      vmSize: nodes[i].vm_size
+    }
+    osProfile: (hasVmgsPath(nodes[i]) ? null : getOsProfile(nodes[i].short_name, admin_username, (!empty(admin_password)), admin_password, ((!empty(admin_key_data)) && nodes[i].is_linux), getLinuxConfiguration('/home/${admin_username}/.ssh/authorized_keys', admin_key_data, empty(admin_password))))
```
squirrelsc: The `hasVmgsPath` should be called in `getOsProfile`. `getOsProfile` should return a final result. So the line here should be like:
```bicep
osProfile: getOsProfile(nodes[i].short_name, admin_username, admin_password, admin_key_data)
```

lisa/sut_orchestrator/azure/arm_template.bicep
squirrelsc: Not only this place, all `? :` logic should be moved to functions, not in the general template. Think it like code. Make it a clear hierarchy.

lisa/sut_orchestrator/azure/README.rst
```diff
@@ -0,0 +1,36 @@
+Azure Deployment Template
```
squirrelsc: Move it to doc folder, and refer it in dev parts.

lisa/sut_orchestrator/azure/README.rst
```diff
@@ -0,0 +1,36 @@
...
+
+   Make sure you have the Bicep CLI installed. You can install it `here <https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-cli>`__. Put bicepconfig.json in the same directory as the Bicep template.
+
+3. **Deployment Options for Debugging**:
+   
+   - **Deploy with ARM Template and Parameters**: Use the following commands to deploy the generated ARM template with parameters:
+   
+     .. code-block:: bash
+     
+        az group create -l westus3 -n biceptest
```
squirrelsc: Is it really needed? And it's hard to follow, because it needs a parameters file, but not exists here. You can specify after that, run LISA as normal see if any error and new changes applied or not.

squirrelsc: Actually, comparing with previous steps, one more step is to build json before call LISA.

lisa/sut_orchestrator/azure/README.rst
```diff
@@ -0,0 +1,36 @@
+Azure Deployment Template
+==========================
+
+`Bicep <https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/>`__ provides a more concise and readable way to define Azure infrastructure compared to ARM templates. However, as of now, the Azure Python SDK doesn't support direct deployment of Bicep templates. Therefore, we need to generate an ARM template from the Bicep template and deploy it using the Azure SDK.
```
squirrelsc: To fast respond to regression on the big change, please add a flag in dev section, called `use_bicep`. The default value is `false`. And make the generated json another name like `autogen_arm_template.json`.

lisa/sut_orchestrator/azure/README.rst
LiliDeng: I add it under Deployment Options for Debugging section

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: when use below code, it has error during deployment

```
func getOsProfile(node object, admin_username string, admin_password string, admin_key_data string) object|null => isCvm(node) ? null 
: {
  computername: node.short_name
  adminUsername: admin_username
  adminPassword: (empty(admin_password) ? null : admin_password)
  linuxConfiguration: (((!empty(admin_key_data)) && node.is_linux) ? getLinuxConfiguration('/home/${admin_username}/.ssh/authorized_keys', admin_key_data, empty(admin_password)) : null)
}
```

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: updated

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: I tried to wrap them in one function, please take a look, thanks

lisa/sut_orchestrator/azure/README.rst
LiliDeng: you mean add use_bicep into runbook, by default, set it as false?

lisa/sut_orchestrator/azure/arm_template.bicep
squirrelsc: 1. What's the error? What's different template generated, if use a function here?
2. For `?:`, it's better to have a separate line to use variable name to get it clear logic.

squirrelsc: Try if it works with two level of functions. First level is to return the object, called `generateOsProfile`. The second level calls the first one, called `getOsProfile` like below,

```bicep
func getOsProfile(node object, admin_username string, admin_password string, admin_key_data string) any =>  isCvm(node) ? null: generateOsProfile(node, admin_username, admin_password, admin_key_data)
```

This table shows any type is useful for dynamic types.
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/data-types#data-type-assignability

lisa/sut_orchestrator/azure/README.rst
squirrelsc: It's not recommended to debugging with az commands, because the reader may not familiar with az commands, and it doesn't save time or provide more information by using lisa directly.

lisa/sut_orchestrator/azure/README.rst
squirrelsc: Yes, add a use_bicep. The default value depends on how we confident on the change. If no regression on wide scenarios, the default value is true, otherwise set to false. If it's set to true, and a regression is reported in some pipelines, the value could be set to false there. It doesn't need to wait we fix the regression.

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: updated

lisa/sut_orchestrator/azure/README.rst
LiliDeng: updated

lisa/sut_orchestrator/azure/README.rst
LiliDeng: updated

lisa/sut_orchestrator/azure/README.rst
LiliDeng: updated

docs/run_test/azure_template.rst
```diff
@@ -0,0 +1,18 @@
+Azure Deployment Template
```
squirrelsc: It should be in the `write_test` parts. Add a link to reach this page in the `extension.rst`.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -365,6 +368,9 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
                     "should be specified either both or not."
                 )
 
+        if self.use_bicep is True:
```
squirrelsc: It doesn't need to handle this flag here. Just use it in `_load_template` method.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1046,7 +1052,9 @@ def _initialize_credential(self) -> None:
 
     def _load_template(self) -> Any:
         if self._arm_template is None:
-            template_file_path = Path(__file__).parent / "arm_template.json"
+            template_file_path = (
```
squirrelsc: How about existing code, which instruments in the template? Are they effected? Including the internal nsg hook?

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -0,0 +1,349 @@
...
+func getVMOsDisk(node object) object => isCvm(node) ? getOSDisk('${node.name}-disk')
+: ((node.os_disk_type == 'Ephemeral')
+? getEphemeralOSImage(node)
+: getOSImage(node))
+
+func getDataDisk(nodeName string, dataDisk object, index int) object => (dataDisk.type == 'UltraSSD_LRS')
+? getAttachDisk(dataDisk, '${nodeName}-data-disk-${index}', index)
+: getCreateDisk(dataDisk, '${nodeName}-data-disk-${index}', index)
+
+resource nodes_vms 'Microsoft.Compute/virtualMachines@2022-08-01' = [for i in range(0, node_count): {
```
squirrelsc, line 310: Move resources together. Don't mix func and resources.

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -0,0 +1,349 @@
...
+      createOption: 'ImportSecure'
+      storageAccountId: resourceId(shared_resource_group_name, 'Microsoft.Storage/storageAccounts', vhd_storage_name)
+      securityDataUri: nodes[i].vhd.vmgs_path
+      sourceUri: nodes[i].vhd.vhd_path
+    }
+  }
+  zones: (use_availability_zones ? availability_zones : null)
+}]
+
+resource nodes_data_disks 'Microsoft.Compute/disks@2022-03-02' = [for i in range(0, (length(data_disks) * node_count)): if (data_disks[(i % length(data_disks))].type == 'UltraSSD_LRS') {
```
squirrelsc: Move `if` to next line, so it's easy to read. Add comments to explain the different logic of UltraSSD_LRS. Can the logic of ultraSSD_LRS be in a function?

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: I have tested it, nsg hook works well in both scenario.

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: done

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: done

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: done

docs/run_test/azure_template.rst
LiliDeng: I don't understand why it is put in `extension.rst`

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: - the `if` line can't be the single line, it will throw error. 
- I changed the `for` and `if` into a new line. please check, and the logic can't be put under the function, since the resource will check the properties' names.
- add comments for the logic of UltraSSD_LRS.

docs/write_test/azure_template.rst
```diff
@@ -0,0 +1,18 @@
+Azure Deployment Template
```
squirrelsc, line 1: Provide a link to reach this document in the `extension.rst`.

docs/run_test/azure_template.rst
squirrelsc: I think it's best place, but you can find another place.

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -0,0 +1,354 @@
...
+      createOption: 'ImportSecure'
+      storageAccountId: resourceId(shared_resource_group_name, 'Microsoft.Storage/storageAccounts', vhd_storage_name)
+      securityDataUri: nodes[i].vhd.vmgs_path
+      sourceUri: nodes[i].vhd.vhd_path
+    }
+  }
+  zones: (use_availability_zones ? availability_zones : null)
+}]
+
+/*
```
squirrelsc: This format of comments is for the whole resource. But this resource is for all data disks. Use `/* comments */` and comment above the `for loop` line.

LiliDeng: update

docs/run_test/azure_template.rst
LiliDeng: updated

lisa/sut_orchestrator/azure/arm_template.bicep
LiliDeng: during the testing, I found when use null or {}, both don't work, I have raised an issue in bicep github, and wait for the response.
`func getImageReference(node object) object => isCvm(node) 
? null
: generateImageReference(node)`

when use `null`, the error is `'The provided value for the output of the user-defined function '__bicep.getImageReference' is not valid. The value '' is not part of the allowed value(s): '{}'.'",`

when use `{}`, the error is `The value of parameter imageReference.publisher is invalid`

squirrelsc: How about `json('null')`?
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/linter-rule-simplify-json-null

LiliDeng: thanks, I tried `json('null')`, it has the same error with using `null`

`The value '' is not part of the allowed value(s): '{}'.'`

LiliDeng: Fixed this issue. And user-defined types (including nullable types) are only available in **ARM language version 2.0**, which uses a dict for resources instead of an array. The template is still deployable.

squirrelsc (dismissed):

LiliDeng:
test scenario

- regression testing without bicep enabled

- enable bicep
> gen1 and gen2 marketplace image
> gen1 and gen2 shared gallery image
> gen1 and gen2 vhd
> cvm vhd
> case with data disks and nvme disks
>  cover the path which need load and update template

squirrelsc (approved):

PR #3165, opened 2024-02-05
Change default Architecture

kamalca:
For maximize_capability, the capability needs to accept both x64 and Arm64 architectures. When it just defaults to x64, this blocks deployment of undefined Arm64 SKUs with ARM images.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -6,6 +6,7 @@ import json
 import re
 import string
 from dataclasses import dataclass, field
+from enum import Enum
 from functools import partial
 from pathlib import Path
 from random import randint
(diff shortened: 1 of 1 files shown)


kamalca:
I have some concerns with how these changes will impact this code: https://github.com/microsoft/lisa/blob/f99fb4aa47a9725485933534b1b50fd70fe140f3/lisa/sut_orchestrator/azure/features.py#L1995-L1997

I will test the behavior of resize tests on undefined ARM SKU. Let me know if I should test any other scenarios.

squirrelsc:
@LiliDeng LGTM

kamalca:
> I have some concerns with how these changes will impact this code:
> 
> https://github.com/microsoft/lisa/blob/f99fb4aa47a9725485933534b1b50fd70fe140f3/lisa/sut_orchestrator/azure/features.py#L1995-L1997
> 
> I will test the behavior of resize tests on undefined ARM SKU. Let me know if I should test any other scenarios.

This code is irrelevant for undefined SKUs. The test case already fails on the following line
https://github.com/microsoft/lisa/blob/62ce27e3ee409404570a0bccf86e22d211626f14/lisa/sut_orchestrator/azure/features.py#L1942

I don't foresee any other issues. So LGTM.

LiliDeng (approved):

PR #3166, opened 2024-02-06
VhdTransformer: Skip failed prepare

kamalca:
For lab environment, the entire waagent / cloud-init package needs to be removed. This will cause VhdTransformer to fail when getting waagent tool.

### DIFF
--- a/lisa/sut_orchestrator/azure/transformers.py
+++ b/lisa/sut_orchestrator/azure/transformers.py
@@ -93,6 +93,9 @@ class VhdTransformerSchema(schema.Transformer):
     # restore environment or not
     restore: bool = False
 
+    # deprovision the VM or not
+    deprovision: bool = True
+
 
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -165,10 +165,13 @@ def _prepare_virtual_machine(self, node: RemoteNode) -> None:
         if not runbook.public_address:
             runbook.public_address = node.public_address
 
-        # prepare vm for exporting
-        wa = node.tools[Waagent]
-        node.execute("export HISTSIZE=0", shell=True)
-        wa.deprovision()
+        try:
```
squirrelsc: Use a switch like `skip_wa_deprovision`. So it's applicable and restrict for different scenarios.

squirrelsc (approved):

PR #3167, opened 2024-02-07
Availability default and max_cap

kamalca:
Add a catch for if maximize_capability is set and the availability type is set to default. It should use availability sets if the tags or properties are set.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -2397,6 +2397,15 @@ class Availability(AzureFeatureMixin, features.Availability):
                 "Or consider changing the disk type or location."
             )
 
+        # If the availability_type is still set to Default, then
+        # resolve the default without considering capabilities
+        if params.availability_type == AvailabilityType.Default:
+            params.availability_type = (
+                AvailabilityType.AvailabilitySet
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -2397,6 +2397,15 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
                 "Or consider changing the disk type or location."
             )
 
+        # If maximize_capability is set, only use AvailabilitySet if
```
squirrelsc: The logic doesn't check for max cap, how to make sure it applied only max cap is set?

kamalca: If we enter the 'not max cap' logic above, there is no way for the availability_type to still be default.

squirrelsc: Explicitly logic is clearer, if change above condition to `if is_maximize_capability`, and put the new logic there.

kamalca: Maybe it would be better if I changed the comment?
I wrote the code in a way such that if for any reason the value is still default when it shouldn't be, then resolve the default.

squirrelsc (approved):

PR #3168, opened 2024-02-07
DisableCloudComponents

kamalca:
Transformer to disable cloud-init and walinuxagent so that an exported VHD can boot in a lab environment.


kamalca:
@squirrelsc Transformer moved from private repo

squirrelsc (approved):

squirrelsc (approved):

PR #3173, opened 2024-02-12
azure: prioritize location with quota availability

pupacha:
Current code returns the awaitable_candidates of the first location from the allowed_locations list even if other locations have available_candidates. This change will iterate through all the allowed_locations and returns a matching available_candidate. If no matching is found, it iterates through locations again and gives an awaitable_candidate.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -521,34 +521,34 @@ class AzurePlatform(Platform):
         all_awaitable: bool = False
         errors: List[str] = []
 
-        for location in allowed_locations:
-            caps, error = self._get_azure_capabilities(
-                location=location, nodes_requirement=nodes_requirement, log=log
-            )
+        caps, error = self._get_azure_capabilities(
+            allowed_locations=allowed_locations,
+            nodes_requirement=nodes_requirement,
+            log=log,
+        )
 
-            if error:
-                errors.append(error)
+        if error:
+            errors.append(error)
 
-            # If returns non-zero length array, it means found either available
-            # or awaitable for all nodes.
-            if caps:
-                all_awaitable = True
+        # If returns non-zero length array, it means found either available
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -521,34 +521,34 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
         all_awaitable: bool = False
         errors: List[str] = []
 
-        for location in allowed_locations:
```
squirrelsc, line 524: This loop should check all locations already. Why move the loop into `_get_azure_capabilities`

pupacha:
@squirrelsc I noticed LISA doesn't use other locations even if there is available quota, when multiple locations are passed. With these changes, LISA would prefer a location with available quota.

lisa/sut_orchestrator/azure/platform_.py
pupacha: The loop was in `_prepare_environment` method. and `_get_azure_capabilities` method was returning an `awaitable_candidate` (if applicable) for the first location of the list. And even if the other locations have enough available quota, it was using the awaitable_candidate.

squirrelsc: The L538 `if all(isinstance(x, schema.NodeSpace) for x in caps):` is to check if all are available resources. If it's not, the loop takes the next location. Please check why it doesn't work as expected.

pupacha: IIUC, the check is to verify the object type. How does L538 verify if the quota is available or not? 
This https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/platform_.py#L2592 could possibly return awaitable result.

pupacha:
https://github.com/microsoft/lisa/pull/3174 fixes the same issue mentioned in this PR. Closing PR.

PR #3176, opened 2024-02-14
shell: fix output missed on local Linux node.

squirrelsc:
With the fix, pty is only enabled for remote Linux.


anirudhrb:
I tested this change and it fixed the issue for me.

PR #3179, opened 2024-02-20
Deploy VM in TiP with Availability_Type None

adityagesh:
To Deploy VM in TiP with Availability_type None requires TipNode.SessionId to be defined in vm_tags.
This tag however has to be removed in case of availability set deployment.


squirrelsc:
@adityagesh don't mention internal information in Github like TiP.

PR #3181, opened 2024-02-21
Disable Cloud Components: install_packages

kamalca:
Different lab testing environments have different package dependencies for the OS. This will allow packages to be installed as part of the process to prepare the OS for lab environment.

### DIFF
--- a/lisa/transformers/repo_package_installer.py
+++ b/lisa/transformers/repo_package_installer.py
@@ -1,0 +1,56 @@
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import schema
+from lisa.operating_system import Posix
+from lisa.transformers.deployment_transformer import (
+    DeploymentTransformer,
+    DeploymentTransformerSchema,
+)
+
+
+@dataclass_json()
+@dataclass
+class RepoPackageInstallerTransformerSchema(DeploymentTransformerSchema):
+    install_packages: List[str] = field(default_factory=list)
+
+
+class RepoPackageInstallerTransformer(DeploymentTransformer):
+    """
+    This Transformer installs packages from a repository.
+    """
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "install_repo_packages"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return RepoPackageInstallerTransformerSchema
+
+    @property
+    def _output_names(self) -> List[str]:
+        return []
+
+    def _internal_run(self) -> Dict[str, Any]:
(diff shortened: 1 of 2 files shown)


lisa/transformers/disable_cloud_components.py
```diff
@@ -43,6 +44,13 @@ def _internal_run(self) -> Dict[str, Any]:
         sed = node.tools[Sed]
         ls = node.tools[Ls]
 
+        if (
```
squirrelsc: It should be another transformer. So they can be organized easily.

kamalca: Should I try including it in the existing package_installer transformer? Right now it is designed for RPMs and other package files. I originally didn't include it in this tranformer because it would require quite a bit of changes for it to make sense. Also if I create another tranformer named package_installer or something similar it would be confusing.

squirrelsc: That one is to install local packages. Maybe you can create a new one called `install_repo_packages`

lisa/transformers/repo_package_installer.py
```diff
@@ -0,0 +1,49 @@
...
+    @property
+    def _output_names(self) -> List[str]:
+        return []
+
+    def _internal_run(self) -> Dict[str, Any]:
+        runbook: RepoPackageInstallerTransformerSchema = self.runbook
+        assert isinstance(runbook, RepoPackageInstallerTransformerSchema)
+        node = self._node
+
+        if (
```
squirrelsc: Raise exceptions, if it's not this case. At least logging, instead of do nothing silently.

kamalca: I raised an exception for non-Posix. But the other checks are just for `[]` or `[""]`. I didn't want to run install packages on nothing because that will upgrade all the packages. But it shouldn't be an error if the runbook has an option to install packages but provides nothing.

lisa/transformers/repo_package_installer.py
```diff
@@ -0,0 +1,47 @@
...
+    @property
+    def _output_names(self) -> List[str]:
+        return []
+
+    def _internal_run(self) -> Dict[str, Any]:
+        runbook: RepoPackageInstallerTransformerSchema = self.runbook
+        assert isinstance(runbook, RepoPackageInstallerTransformerSchema)
+        node = self._node
+
+        assert isinstance(node.os, Posix)
```
squirrelsc: Add failure message.

lisa/transformers/repo_package_installer.py
squirrelsc: It's ok not to raise an exception. It should be logged at least if nothing installed. That helps debugging issues like package string is not passed in from pipeline.

squirrelsc (dismissed):

squirrelsc (approved):

PR #3183, opened 2024-02-22
disk: handle when passed disktype is empty

LiliDeng:


lisa/schema.py
```diff
@@ -473,7 +473,18 @@ class DiskOptionSettings(FeatureSettings):
             ],
         ),
         metadata=field_metadata(
-            decoder=partial(search_space.decode_set_space_by_type, base_type=DiskType)
+            decoder=lambda input: (
+                search_space.decode_set_space_by_type(data=input, base_type=DiskType)
```
squirrelsc: Write a method, it would be clearer than lambda.

lisa/search_space.py
```diff
@@ -710,3 +710,15 @@ def create_set_space(
     else:
         set_space = None
     return set_space
+
+
+def decode_empty_count_space(
```
squirrelsc: It should be called like `decode_nullable_set_space`. It's not a count space or always empty one.

lisa/search_space.py
```diff
@@ -710,3 +710,15 @@ def create_set_space(
...
+    data: Any, base_type: Any, allowed_values: Any, is_allow_set: bool = False
```
squirrelsc: `allowed_values` should be `default_values`

lisa/search_space.py
LiliDeng: updated

lisa/search_space.py
LiliDeng: updated

squirrelsc (approved):

PR #3184, opened 2024-02-22
Add is_bash in SshShell to flag the posix system without bash support

lubaihua33:
Rack manager server doesn't support bash. When try_connect, the stdout_content has "Unknown syntax: cmd" string. So, I add is_bash in SshShell to flag the system like Rack manager server. 
When it needs minimal shell type, if not is_bash, the function generate_run_command needs to be overridden. This part change can refer to https://github.com/microsoft/lisa/pull/2908

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -267,6 +267,9 @@ class SshShell(InitializableMixin):
         else:
             self.is_posix = True
             shell_type = spur.ssh.ShellTypes.sh
+            # it doesn't support bash. Use minimal shell type
+            if stdout_content and "Unknown syntax" in stdout_content:
+                shell_type = spur.ssh.ShellTypes.minimal
 
         sock = self._establish_jump_boxes(
             address=self.connection_info.address,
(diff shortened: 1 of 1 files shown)


lisa/util/shell.py
```diff
@@ -261,12 +261,15 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
         stdout.close()
 
+        self.is_bash = True
         if stdout_content and "Windows" in stdout_content:
             self.is_posix = False
             shell_type = WindowsShellType()
         else:
             self.is_posix = True
             shell_type = spur.ssh.ShellTypes.sh
+            if stdout_content and "Unknown syntax" in stdout_content:
```
squirrelsc, line 271: How about assign shell_type to minimal here, and use the variable called `_shell_type`?

lubaihua33: _shell_type is used for?
I have added some changes, but have an error. I'm not sure how to fix that, can you help?

squirrelsc: If the selected shell type doesn't need to be used in next lines, it doesn't need have a variable.

lisa/util/shell.py
```diff
@@ -267,6 +267,14 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         else:
             self.is_posix = True
             shell_type = spur.ssh.ShellTypes.sh
+            # it doesn't support bash. Use minimal shell type
+            if stdout_content and "Unknown syntax" in stdout_content:
+                shell_type = spur.ssh.ShellTypes.minimal
+                func_type = type(spur.ssh.ShellTypes.minimal.generate_run_command)
+                shell_type.generate_run_command = func_type(
```
squirrelsc: The mypy error is caused by the shell_type is not an instance. So the replacement of generate_run_command should happen after `self._inner_shell = spurplus.SshShell(spur_ssh_shell=spur_ssh_shell, sftp=sftp)`. The `self._inner_shell` is an instance.

lubaihua33: Thank you Chi. I have added the changes.

lubaihua33: Thank you Chi. I have added the changes.

squirrelsc (approved):

goenkam:
This works for the Overlake case, can we merge. 

image

PR #3185, opened 2024-02-23
Bicep - enable it by default

LiliDeng:


squirrelsc (approved):

LiliDeng:
will merge once internal code is ready

PR #3187, opened 2024-02-26
add resource_group_managed_by support in runbook, prepare for RG-scoped roles

paxue:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1344,6 +1346,7 @@ def check_or_create_resource_group(
     cloud: Cloud,
     resource_group_name: str,
     location: str,
+    managed_by: str,
```
squirrelsc: Give a default value `""`. So it doesn't need to be provided, if it's not needed.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1356,9 +1359,16 @@ def check_or_create_resource_group(
...
+                managed_by_para = None
+            else:
+                log.info(f"Using managed_by resource group： '{managed_by}'")
+                managed_by_para = managed_by
+
             with global_credential_access_lock:
                 rm_client.resource_groups.create_or_update(
-                    resource_group_name, {"location": location}
+                    resource_group_name,
+                    {"location": location, "managed_by": managed_by_para},
```
squirrelsc: The dict can be assigned to a variable firstly, so you can base on if managed_by is set or not to update the dict.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1356,9 +1359,19 @@ def check_or_create_resource_group(
         if not az_shared_rg_exists:
             log.info(f"Creating Resource group: '{resource_group_name}'")
 
+            if managed_by == "":
```
squirrelsc:
```python
rg_properties = {"location": location}
if managed_by:
  log.debug(f"Using managed_by resource group: '{managed_by}'")
  rg_properties["managed_by"] = managed_by
```

LiliDeng:
I have tested it, no regression issue found.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3190, opened 2024-02-27
Add no_redundancy_vm_tags to Azure Platform

adityagesh:
Add no_redundancy_vm_tags to add VM tags when availability_type is set as none in Azure platform

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -695,6 +695,7 @@ class AvailabilityArmParameter:
     availability_set_tags: Dict[str, str] = field(default_factory=dict)
     availability_set_properties: Dict[str, Any] = field(default_factory=dict)
     availability_zones: List[int] = field(default_factory=list)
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -375,7 +376,7 @@
             },
             "name": "[parameters('nodes')[copyIndex('vmCopy')]['name']]",
             "location": "[parameters('nodes')[copyIndex('vmCopy')]['location']]",
-            "tags": "[parameters('vm_tags')]",
+            "tags": "[union(parameters('vm_tags'), variables('no_redundancy_vm_tags'))]",
```
squirrelsc: For this change, make it ready before calling into arm template. It doesn't need to union in the arm template.

LiliDeng: please check #3163 , I will send the notification mail ASAP.

PR #3195, opened 2024-03-01
Make data_path nullable

kamalca:
Making data_path nullable will allow us to add it to a runbook without specifying a default value.


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3200, opened 2024-03-07
Implement package cleanup

adityagesh:
Implement package cleanup for Linux.


adityagesh:
Used in https://github.com/microsoft/lisa/pull/3094

squirrelsc (approved):

PR #3202, opened 2024-03-07
Use sudo with cat /etc/waagent.conf

kanchansenlaskar:
This change fixes the InfinibandSuite.verify_hpc_over_sriov test.
The change is tested with image: microsoftcblmariner cbl-mariner cbl-mariner-2 2.20240301.01

### DIFF
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -115,7 +115,9 @@ class Waagent(Tool):
         waagent_conf_file = self._get_waagent_conf_path()
 
         config = {}
-        cfg = self.node.tools[Cat].run(waagent_conf_file, force_run=force_run).stdout
+        cfg = self.node.tools[Cat].read(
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/tools.py
```diff
@@ -115,7 +115,11 @@ def _get_configuration(self, force_run: bool = False) -> Dict[str, str]:
         waagent_conf_file = self._get_waagent_conf_path()
 
         config = {}
-        cfg = self.node.tools[Cat].run(waagent_conf_file, force_run=force_run).stdout
+        cfg = (
+            self.node.tools[Cat]
+            .run(waagent_conf_file, force_run=force_run, sudo=True)
+            .stdout
+        )
```
LiliDeng, line 120: thanks for the fix, could you please use `cat.read`

kanchansenlaskar: done

LiliDeng (approved):

PR #3203, opened 2024-03-08
verify_key_vault_extension: Limit keyvault name to 24 chars

adityagesh:
Limit name to 24 chars

### DIFF
--- a/microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
+++ b/microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
@@ -101,7 +101,10 @@ class AzureKeyVaultExtensionBvt(TestSuite):
         resource_group_name = runbook.shared_resource_group_name
         application_id = runbook.service_principal_client_id
         node_context = get_node_context(node)
-        vault_name = f"lisa-kve-{platform.subscription_id[-5:]}-{node_context.location}"
+        # A vault's name must be between 3-24 alphanumeric characters.
+        vault_name = (
+            f"lisa-kv{platform.subscription_id[-6:]}{node_context.location[:11]}"
+        )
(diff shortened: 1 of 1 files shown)


LiliDeng (dismissed):

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -101,7 +101,10 @@ def verify_key_vault_extension(
         resource_group_name = runbook.shared_resource_group_name
         application_id = runbook.service_principal_client_id
         node_context = get_node_context(node)
-        vault_name = f"lisa-kve-{platform.subscription_id[-5:]}-{node_context.location}"
+        # A vault's name must be between 3-24 alphanumeric characters.
```
squirrelsc, line 104: Do you see it's failed some locations? The key vault name must be global unique, so if the location name is too long to remove sub id, it's easy to conflict between subscriptions.

squirrelsc: Please follow below example to get last chars of locations. I reviewed last 11 chars last time, it has no conflict. It doesn't need to keep the full location name. The prefix need at least one extra char to know it's from kve, like `"list-kv{sub id[:6]}{location[:11]}"`

```python
def get_storage_account_name(
    subscription_id: str, location: str, type_: str = "s"
) -> str:
    subscription_id_postfix = subscription_id[-8:]
    # name should be shorter than 24 character
    return f"lisa{type_}{location[:11]}{subscription_id_postfix}"
```

squirrelsc:
@LiliDeng LGTM. It will leave some old kv not used, but it should be ok.

LiliDeng (approved):

PR #3204, opened 2024-03-11
Added unsupported OS requirement to verify_ringbuffer_settings_change

r-dailey:
This test case should be skipped as FreeBSD does not support modifying ringbuffer sizes like in this test.

### DIFF
--- a/microsoft/testsuites/network/networksettings.py
+++ b/microsoft/testsuites/network/networksettings.py
@@ -91,6 +91,9 @@ class NetworkSettings(TestSuite):
 
         """,
         priority=1,
+        requirement=simple_requirement(
+            unsupported_os=[BSD],
+        ),
     )
     def verify_ringbuffer_settings_change(self, node: Node) -> None:
         ethtool = node.tools[Ethtool]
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/network/networksettings.py
```diff
@@ -91,6 +91,9 @@ class NetworkSettings(TestSuite):
 
         """,
         priority=1,
+        requirement=simple_requirement(
+            unsupported_os=[BSD],
```
squirrelsc: Please exclude Windows also. Windows is by default excluded. But after the change, it's included now.

squirrelsc:
> FreeBSD does not use ethtool so this test case should be skipped as it is testing an ethtool capability.

The reason is FreeBSD doesn't support to set ringbuffer, so it's skipped. Please make description and git commit messages accurate.

microsoft/testsuites/network/networksettings.py
r-dailey: Added Windows to the unsupported OS list.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3205, opened 2024-03-11
Updated gcc.py to enable gcc installation on FreeBSD

r-dailey:
Updated gcc.py to enable gcc installation on FreeBSD

### DIFF
--- a/lisa/tools/gcc.py
+++ b/lisa/tools/gcc.py
@@ -60,9 +60,6 @@ class Gcc(Tool):
             # Tools including sockperf need gcc11 on
             # FreeBSD to compile.
             posix_os.install_packages("lang/gcc")
-            # self.node.execute(
-            #     "ln -s /usr/local/bin/gcc11 /usr/local/bin/gcc", sudo=True
-            # )
         else:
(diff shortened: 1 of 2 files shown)


lisa/tools/gcc.py
```diff
@@ -59,10 +59,7 @@ def _install(self) -> bool:
         if isinstance(posix_os, BSD):
             # Tools including sockperf need gcc11 on
             # FreeBSD to compile.
-            posix_os.install_packages("gcc11-11.3.0_8")
-            self.node.execute(
-                "ln -s /usr/local/bin/gcc11 /usr/local/bin/gcc", sudo=True
-            )
+            posix_os.install_packages("lang/gcc")
```
squirrelsc, line 60: What's the default gcc version installed? The above comment says it needs gcc11 at least. Make sure the default version is higher than it. And the comment can be removed, because we use latest gcc version.

r-dailey: Default version installed is 12.2. I removed the comment and my testing changes.

LiliDeng (approved):

PR #3206, opened 2024-03-12
util: make deepcopy in set_filtered_fields

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1170,9 +1170,11 @@ def _create_deployment_parameters(
             "availability_zones",
             "availability_type",
         ]
-        set_filtered_fields(self._azure_runbook, arm_parameters, copied_fields)
+        # the runbook may be updated, so make a copy to avoid the original runbook
+        azure_runbook = copy.deepcopy(self._azure_runbook)
+        set_filtered_fields(azure_runbook, arm_parameters, copied_fields)
         set_filtered_fields(
-            self._azure_runbook,
+            azure_runbook,
```
squirrelsc: I don’t see the new run book is used later. Is it needed to set back?

LiliDeng: this place will change the origin runbook. https://github.com/microsoft/lisa/blob/main/lisa/sut_orchestrator/azure/features.py#L2448

LiliDeng: using deepcopy in set_filtered_fields also works

squirrelsc: Change in the set_filtered_fields is right, but it needs to check regressions carefully.

squirrelsc (approved):

PR #3207, opened 2024-03-12
tool: AzureBlobOperator

LiliDeng:


lisa/tools/blobdownload.py
```diff
@@ -0,0 +1,115 @@
...
+        connection_string: str,
+        container_name: str,
+        blob_name: str,
+        file_name: str = "",
+    ) -> PurePath:
+        if file_name:
+            file_path = self.get_tool_path().joinpath(file_name)
+        else:
+            file_path = self.get_tool_path().joinpath(blob_name)
+        download_script = (
```
squirrelsc: This tool should be in Azure folder like other Azure tools. So it won't be implemented like this way.

lisa/tools/blobdownload.py
```diff
@@ -0,0 +1,115 @@
...
+from pathlib import PurePath
+from typing import Any, List
+
+from lisa.executable import Tool
+from lisa.util import LisaException
+
+from .python import Pip, Python
+
+
+class BlobDownloader(Tool):
```
squirrelsc: Change current blob download parts to use this tool.

lisa/tools/blobdownload.py
```diff
@@ -0,0 +1,115 @@
...
+        return self._check_exists()
+
+    def download_blob(
+        self,
+        connection_string: str,
+        container_name: str,
+        blob_name: str,
+        file_name: str = "",
+    ) -> PurePath:
+        if file_name:
```
squirrelsc: When is the file name or blob name provided? Can we just provide one way?

lisa/tools/blobdownload.py
LiliDeng: @squirrelsc could you please elaborate this `So it won't be implemented like this way.`? Thanks. 
I have tried two ways, either download the blob into the lisa run machine, then upload into the test target; or use this wrap python code, to download the blob directly into the test target.

lisa/tools/blobdownload.py
LiliDeng: there are no places to download blob in current code

lisa/tools/blobdownload.py
LiliDeng: update

lisa/tools/blobdownload.py
squirrelsc: > there are no places to download blob in current code

what means "no places to download blob?" If you mean the folder to save the downloaded things, you can use the local working folder from local node.

pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "spurplus ~= 2.3.4",
     "websockets ~= 10.3",
     "charset_normalizer ~= 2.1.1",
+    "msrestazure ~= 0.6.4",
```
squirrelsc: Don't use it in default dependencies. As comments in another PR. It needs to create another installation option or put it in Azure staffs.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1232,36 +1232,49 @@ def generate_sas_token(
...
+    if connection_string:
+        blob_service_client = BlobServiceClient.from_connection_string(
+            connection_string
+        )
+    else:
+        assert subscription_id, "subscription_id is required."
+        assert cloud, "cloud is required."
+        assert account_name, "account_name is required."
+        assert resource_group_name, "resource_group_name is required."
+        shared_key_credential = get_storage_credential(
```
squirrelsc:
```suggestion
        storage_credential = get_storage_credential(
```

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1232,36 +1232,49 @@ def generate_sas_token(
...
-        shared_key_credential,
-    )
+    blob_service_client: BlobServiceClient
+    if connection_string:
+        blob_service_client = BlobServiceClient.from_connection_string(
+            connection_string
+        )
+    else:
+        assert subscription_id, "subscription_id is required."
+        assert cloud, "cloud is required."
```
squirrelsc: We should give the public cloud as default.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1232,36 +1232,49 @@ def generate_sas_token(
...
-        f"https://{account_name}.blob.{cloud.suffixes.storage_endpoint}",
-        shared_key_credential,
-    )
+    blob_service_client: BlobServiceClient
+    if connection_string:
+        blob_service_client = BlobServiceClient.from_connection_string(
+            connection_string
+        )
+    else:
+        assert subscription_id, "subscription_id is required."
```
squirrelsc: The message should say like "subscription_id is required, if the connection_string is not set."

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1232,36 +1232,49 @@ def generate_sas_token(
...
-    subscription_id: str,
-    cloud: Cloud,
-    account_name: str,
-    resource_group_name: str,
+    credential: Optional[Any] = None,
+    subscription_id: Optional[str] = None,
+    cloud: Optional[Cloud] = None,
+    account_name: Optional[str] = None,
+    resource_group_name: Optional[str] = None,
+    connection_string: Optional[str] = None,
```
squirrelsc, line 1242: Is connection string the common concept? should it be a sas url?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1232,36 +1232,49 @@ def generate_sas_token(
...
-    account_name: str,
     container_name: str,
-    resource_group_name: str,
+    credential: Optional[Any] = None,
+    subscription_id: Optional[str] = None,
+    cloud: Optional[Cloud] = None,
+    account_name: Optional[str] = None,
+    resource_group_name: Optional[str] = None,
+    connection_string: Optional[str] = None,
+    create_if_not_exist: bool = True,
```
squirrelsc: `allow_create`

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -579,3 +580,91 @@ def get_pool_records(self, pool_id: int, force_run: bool = False) -> Dict[str, s
...
+        account_name: str,
+        container_name: str,
+        blob_name: str,
+        credential: Optional[Any] = None,
+        subscription_id: Optional[str] = None,
+        cloud: Optional[Cloud] = None,
+        resource_group_name: Optional[str] = None,
+        connection_string: Optional[str] = None,
+    ) -> PurePath:
+        from lisa import constants
```
squirrelsc: Move them to top, avoid to import in methods.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -579,3 +580,91 @@ def get_pool_records(self, pool_id: int, force_run: bool = False) -> Dict[str, s
...
+        )
+        blob_client = container_client.get_blob_client(blob_name)
+        if not blob_client.exists():
+            raise LisaException(
+                f"Blob {blob_name} not found in container {container_name}"
+            )
+        try:
+            blob_data = blob_client.download_blob()
+            with open(file_path, "wb") as f:
+                f.write(blob_data.readall())
```
squirrelsc: Use `chunks` to return as an iteration. readall will cause OOM or performance issue with 30GB vhd.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -579,3 +580,91 @@ def get_pool_records(self, pool_id: int, force_run: bool = False) -> Dict[str, s
...
+            self._log.debug("Blob downloaded successfully.")
+        except Exception as e:
+            raise LisaException("An error occurred during blob download.") from e
+        return file_path
+
+    def list_blobs(
```
squirrelsc, line 554: Is it used somewhere? Change them use this method.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -579,3 +580,91 @@ def get_pool_records(self, pool_id: int, force_run: bool = False) -> Dict[str, s
             records[content_split[i]] = content_split[i + 1]
 
         return records
+
+
+class BlobDownloader(Tool):
```
squirrelsc: `AzureBlobOperator`. We can move other methods like upload here.

lisa/sut_orchestrator/azure/common.py
LiliDeng: if it is a sas url, use requests / wget can reach the same goal.

squirrelsc: I mean if the **connection string** is a common concept in blob service client? If so, it's ok to use it as a variable name.

lisa/sut_orchestrator/azure/tools.py
LiliDeng: if move them to the top, it will have a circular import error.

pyproject.toml
LiliDeng: I moved it to the default dependencies, because in CI check, the example parts will fail because lack of msrest package. so I need update noxfile.py in example method to use the split installation option?

lisa/sut_orchestrator/azure/tools.py
squirrelsc: Try import constants from `lisa.util`

pyproject.toml
squirrelsc: Please check the calltrace, where it's needed to be imported. If the examples run with Azure, it should be installed. So you can put it in Azure firstly. If you cannot figure it out, please share the call trace here.

LiliDeng: Here is the call trace.
```
nox > lisa --debug
  File "/home/runner/work/lisa/lisa/.nox/example/bin/lisa", line 5, in <module>
    from lisa.main import cli
  File "/home/runner/work/lisa/lisa/lisa/__init__.py", line 6, in <module>
    from lisa.environment import Environment
  File "/home/runner/work/lisa/lisa/lisa/environment.py", line 19, in <module>
    from lisa.node import Node, Nodes
  File "/home/runner/work/lisa/lisa/lisa/node.py", line 24, in <module>
    from lisa.nic import Nics, NicsBSD
  File "/home/runner/work/lisa/lisa/lisa/nic.py", line 15, in <module>
    from lisa.tools import Cat, Ip, KernelConfig, Ls, Lspci, Modprobe, Tee
  File "/home/runner/work/lisa/lisa/lisa/tools/__init__.py", line 93, in <module>
    from .reboot import Reboot
  File "/home/runner/work/lisa/lisa/lisa/tools/reboot.py", line 13, in <module>
  File "/home/runner/work/lisa/lisa/lisa/features/__init__.py", line 14, in <module>
    from .gpu import Gpu, GpuEnabled, GpuSettings
  File "/home/runner/work/lisa/lisa/lisa/features/gpu.py", line 17, in <module>
    from lisa.sut_orchestrator.azure.tools import LisDriver
  File "/home/runner/work/lisa/lisa/lisa/sut_orchestrator/azure/tools.py", line 9, in <module>
    from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD, Cloud  # type: ignore
ModuleNotFoundError: No module named 'msrestazure'
```

lisa/sut_orchestrator/azure/common.py
LiliDeng: in BlobServiceClient.from_connection_string the parameter called
```
        :param str conn_str:
            A connection string to an Azure Storage account.
```

it can be below two types

```
Complete Connection String
DefaultEndpointsProtocol=https;AccountName=<your_account_name>;AccountKey=<your_account_key>;EndpointSuffix=core.windows.net
```

```
SAS Token Connection String
BlobEndpoint=https://<your_account_name>.blob.core.windows.net/;SharedAccessSignature=<your_sas_token>
```

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/tools.py
LiliDeng: it works. but `from lisa.sut_orchestrator.azure.common import get_or_create_storage_container` still can't be moved to the top.

lisa/sut_orchestrator/azure/tools.py

lisa/tools/blobdownload.py
LiliDeng: Currently there is no existing code to use download blob.

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/common.py

pyproject.toml
squirrelsc: The general gpu feature referred an Azure tool, it's not right. If the code related to Azure, it shouldn't be in a general feature. Please move the logic from general gpu feature to azure extended gpu. But is the LisDriver still useful?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1232,50 +1232,68 @@ def generate_sas_token(
...
+    blob_service_client: BlobServiceClient
+    if connection_string:
+        blob_service_client = BlobServiceClient.from_connection_string(
+            connection_string
+        )
+    else:
+        assert (
+            subscription_id
+        ), "subscription_id is required, if connection_string is not set."
+        assert account_name, "account_name is required."
```
squirrelsc: The same as above, mention if connection string is not set.

lisa/sut_orchestrator/azure/tools.py
squirrelsc: I don't see cycle imports from common to other azure modules. What's error here?

lisa/sut_orchestrator/azure/tools.py
squirrelsc: Check blob API, there is a chunks method. We don't need to custom a chunk method.

pyproject.toml
LiliDeng: after moving lisdriver out of azure tool (since it also can be installed on hyperv vm), now the call trace is 
```
  File "/home/runner/work/lisa/lisa/.nox/example/bin/lisa", line 5, in <module>
    from lisa.main import cli
  File "/home/runner/work/lisa/lisa/lisa/main.py", line 16, in <module>
    import lisa.mixin_modules  # noqa: F401
  File "/home/runner/work/lisa/lisa/lisa/mixin_modules.py", line 69, in <module>
    import lisa.transformers.disable_cloud_components  # noqa: F401
  File "/home/runner/work/lisa/lisa/lisa/transformers/disable_cloud_components.py", line 8, in <module>
  File "/home/runner/work/lisa/lisa/lisa/sut_orchestrator/azure/tools.py", line 12, in <module>
    from msrestazure.azure_cloud import AZURE_PUBLIC_CLOUD, Cloud  # type: ignore
ModuleNotFoundError: No module named 'msrestazure'
```

lisa/sut_orchestrator/azure/tools.py

lisa/sut_orchestrator/azure/common.py

lisa/sut_orchestrator/azure/tools.py
LiliDeng: now, it worked. updated

pyproject.toml
squirrelsc: It needs to move `import lisa.transformers.disable_cloud_components` into `try/except` of Azure modules. The general GPU feature shouldn't use lisdriver. The related logic should be in Azure GPU feature.

lisa/tools/lisdriver.py
```diff
@@ -0,0 +1,85 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        if (
+            isinstance(self.node.os, Redhat)
+            and self.node.os.information.version < "7.8.0"
+        ):
+            return True
+
+        raise UnsupportedDistroException(
```
squirrelsc, line 39: Based on the supports up to Redhat 7.8. Is it end of life? Can we remove the lisdriver tool?

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -586,3 +514,117 @@ def get_pool_records(self, pool_id: int, force_run: bool = False) -> Dict[str, s
...
+
+                blob_size = blob_properties["size"]
+                downloaded_size = 0
+                with open(file_path, "wb") as file:
+                    download_stream = blob_client.download_blob()
+                    for chunk in download_stream.chunks():
+                        file.write(chunk)
+                        downloaded_size += len(chunk)
+                        percentage = (downloaded_size / blob_size) * 100
+                        self._log.debug(
```
squirrelsc: Make sure it won't print too many lines. For example, it's ok to update every 10 seconds.

lisa/sut_orchestrator/azure/tools.py
```diff
@@ -586,3 +514,117 @@ def get_pool_records(self, pool_id: int, force_run: bool = False) -> Dict[str, s
...
+                            f"Downloaded {downloaded_size}/{blob_size} bytes "
+                            f"({percentage:.2f}% complete)"
+                        )
+                        time.sleep(2)
```
squirrelsc: Why add sleep here? It should run as fast as possible to download.

lisa/tools/lisdriver.py
LiliDeng: it is EOL, `7.8 (retired  September 30, 2020)`, but in our testing, sometimes we pick up the old distro for testing

lisa/sut_orchestrator/azure/tools.py

lisa/sut_orchestrator/azure/tools.py
LiliDeng: removed

pyproject.toml
LiliDeng: updated, thanks.

lisa/sut_orchestrator/azure/tools.py
LiliDeng: I put download and list logic into azure common file, and refer from there

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2065,17 +2089,121 @@ def check_blob_exist(
...
+                    file.write(chunk)
+                    downloaded_size += len(chunk)
+                    percentage = (downloaded_size / blob_size) * 100
+                    current_time = time()
+                    if current_time >= next_log_time:
+                        log.debug(
+                            f"Downloaded {downloaded_size}/{blob_size} bytes "
+                            f"({percentage:.2f}% complete)"
+                        )
+                        next_log_time += log_interval
```
squirrelsc: The next logging time should be at least 10 seconds from now.
`next_log_time = current_time + log_interval`

squirrelsc (approved):

PR #3213, opened 2024-03-14
tool: ado_artifact_download

LiliDeng:


pyproject.toml
```diff
@@ -27,6 +27,8 @@ dependencies = [
     "spurplus ~= 2.3.4",
     "websockets ~= 10.3",
     "charset_normalizer ~= 2.1.1",
+    "requests",
```
squirrelsc, line 30: Move the tool into Azure folder, and the dependencies should be in Azure too.

lisa/tools/ado_artifact_download.py
```diff
@@ -0,0 +1,127 @@
...
+        file_extension = artifact_path.suffix
+        if file_extension == ".zip":
+            with zipfile.ZipFile(str(artifact_path), "r") as zip_ref:
+                zip_ref.extractall(str(artifact_path.parent))
+        source_path = os.path.splitext(str(artifact_path))[0]
+        self._log.info(f"Artifact extracted to {str(source_path)}")
+        return source_path
+
+    @classmethod
+    def _windows_tool(cls) -> Optional[Type[Tool]]:
```
squirrelsc: There is no special logic for Windows, so it doesn't need the windows extension.

pyproject.toml
squirrelsc: Create a new installation part called "ado" like azure, aws. And check if it's possible to add it as a depenency of azure.

lisa/sut_orchestrator/baremetal/source.py
```diff
@@ -54,79 +51,14 @@ def download(self, timeout: int = 600) -> List[Path]:
...
-                download_file.write(response.content)
-            self._log.debug(f"downloaded artifact to {artifact_path}")
-            if artifact.extract:
-                source_path = self.extract(artifact_path)
-                artifacts_path.append(Path(source_path))
-            else:
-                artifacts_path.append(artifact_path)
+        ado = self._local_node.tools[ADOArtifactsDownloader]
+        artifacts_path = ado.download(
+            personal_access_token,
```
squirrelsc: For the methods with many arguments, give names when calling it. So, if the method add more arguments in the middle, it won't bring errors.

lisa/sut_orchestrator/baremetal/source.py
```diff
@@ -38,6 +32,9 @@ def __init__(self, runbook: ADOSourceSchema) -> None:
         super().__init__(runbook)
         self.ado_runbook: ADOSourceSchema = self.runbook
         self._log = get_logger("ado", self.__class__.__name__)
+        from lisa.node import local
```
squirrelsc, line 35: I think it won't bring cycle import. So it should be put on the top.

squirrelsc: This is used only in download method, so it doesn't need a class level variable. Even it's needed, if there is no cycle import, the import should be at the top of the file.

lisa/tools/__init__.py
```diff
@@ -120,6 +121,7 @@
 from .wsl import Wsl
 
 __all__ = [
+    "ADOArtifactsDownloader",
```
squirrelsc: Move it into another folder called `advanced_tools`. And don't use `__init__.py` there. So, when a tool is used, it will be imported.

lisa/tools/ado_artifact_download.py
LiliDeng: in the baremetal, it used `local().tools[ADOArtifactsDownloader]`, when the local is windows, this code is needed.

lisa/sut_orchestrator/baremetal/source.py
LiliDeng: updated.

lisa/sut_orchestrator/baremetal/source.py
LiliDeng: updated

lisa/tools/__init__.py
LiliDeng: updated, please check, thanks.

lisa/tools/ado_artifact_download.py
squirrelsc: There is no special logic for Windows, that's why I said it doesn't need. Only if windows logic is different from the default implementation, it needs a derived Windows tool. For this tool, only `command` echo different words.

lisa/advanced_tools/ado_artifact_download.py
```diff
@@ -0,0 +1,127 @@
...
+                download_file.write(response.content)
+            self._log.debug(f"downloaded artifact to {artifact_path}")
+            if artifact.extract:
+                source_path = self.extract(artifact_path)
+                artifacts_path.append(Path(source_path))
+            else:
+                artifacts_path.append(artifact_path)
+        return artifacts_path
+
+    def extract(self, artifact_path: Path) -> str:
```
squirrelsc: This method doesn't need to be moved. It's not related to ado artifacts.

lisa/tools/ado_artifact_download.py
LiliDeng: if remove this, I met below issue:
```
LisaException: cannot find [windowsadoartifactsdownloader] on [local], Windows, Remote(False) and installation of [windowsadoartifactsdownloader] isn't enabled in lisa.
```

lisa/advanced_tools/ado_artifact_download.py
LiliDeng: updated

lisa/tools/ado_artifact_download.py
squirrelsc: Can you share the full call trace? The windows tool specification is optional, not required.

LiliDeng: > Can you share the full call trace? The windows tool specification is optional, not required.

fixed

squirrelsc (dismissed):

LiliDeng:
@squirrelsc please review again, I have found and fixed one issue during testing. Thanks.

squirrelsc (approved):

PR #3214, opened 2024-03-14
updated pip install for freebsd

r-dailey:
Updated the install process for Freebsd to enable more tests

### DIFF
--- a/lisa/tools/python.py
+++ b/lisa/tools/python.py
@@ -11,7 +11,7 @@ from lisa.executable import Tool
 if TYPE_CHECKING:
     from lisa.node import Node
 
-from lisa.operating_system import Posix, BSD
+from lisa.operating_system import BSD, Posix
 from lisa.tools.mkdir import Mkdir
 from lisa.util import UnsupportedDistroException, get_matched_str
 
@@ -54,11 +54,10 @@ class Pip(Tool):
 
     def _install(self) -> bool:
         package_name = "python3-pip"
-        assert isinstance(self.node.os, Posix)
         if isinstance(self.node.os, BSD):
-            self.node.os.install_packages("py39-pip")
-        else:
-            self.node.os.install_packages(package_name)
+            package_name = "py39-pip"
+        assert isinstance(self.node.os, Posix)
+        self.node.os.install_packages(package_name)
         return self._check_exists()
 
     def install_packages(self, packages_name: str, install_path: str = "") -> None:
--- a/lisa/tools/python.py
+++ b/lisa/tools/python.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 import re
 from pathlib import PurePath
-from typing import TYPE_CHECKING, List, Type
+from typing import TYPE_CHECKING, List, Type, cast
 
 from assertpy import assert_that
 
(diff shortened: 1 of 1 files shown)


lisa/tools/python.py
```diff
@@ -52,8 +52,12 @@ def dependencies(self) -> List[Type[Tool]]:
 
     def _install(self) -> bool:
         package_name = "python3-pip"
+        posix_os: Posix = cast(Posix, self.node.os)
```
squirrelsc: Why does it needs to cast? the self.node.os should be able to used for freebsd too.

r-dailey: Thats a good question. I don't know why. I just followed the format that I saw in a different install function that already had different install paths for different distros. I will look into it and remove the cast if it is not needed.

r-dailey: Updated this to use the self.node.os and updated the pip name for the tool specific to freebsd to enable it to detect the install

lisa/tools/python.py
```diff
@@ -11,7 +11,7 @@
 if TYPE_CHECKING:
     from lisa.node import Node
 
-from lisa.operating_system import Posix
+from lisa.operating_system import Posix, BSD
```
LiliDeng: `from lisa.operating_system import BSD, Posix`

lisa/tools/python.py
```diff
@@ -53,7 +55,10 @@ def dependencies(self) -> List[Type[Tool]]:
     def _install(self) -> bool:
         package_name = "python3-pip"
         assert isinstance(self.node.os, Posix)
-        self.node.os.install_packages(package_name)
+        if isinstance(self.node.os, BSD):
+            self.node.os.install_packages("py39-pip")
+        else:
+            self.node.os.install_packages(package_name)
```
LiliDeng:
```
if isinstance(self.node.os, BSD):
    package_name = "py39-pip"
self.node.os.install_packages(package_name)
```

lisa/tools/python.py
r-dailey: Reordered the import statements.

lisa/tools/python.py
r-dailey: updated the if statement to minimize the code size.

lisa/tools/python.py
```diff
@@ -52,6 +54,8 @@ def dependencies(self) -> List[Type[Tool]]:
 
     def _install(self) -> bool:
         package_name = "python3-pip"
+        if isinstance(self.node.os, BSD):
+            package_name = "py39-pip"
```
squirrelsc: Is there a version for all py3? py39 may have compatibility issue with newer python versions like 3.10, 3.11.

r-dailey: This is the most recent version of pip I saw available through the pkg utility. I will look into whether or not there is a generic one.

squirrelsc (dismissed):

squirrelsc:
@LiliDeng LGTM

lisa/tools/python.py
r-dailey: Changed it to installing devel/py-pip so that it is not version dependent.

LiliDeng (approved):

PR #3216, opened 2024-03-15
kernel_install: add check_kernel_version

LiliDeng:


lisa/transformers/kernel_installer.py
```diff
@@ -74,6 +74,7 @@ class KernelInstallerTransformerSchema(DeploymentTransformerSchema):
         default=None, metadata=field_metadata(required=True)
     )
     raise_exception: Optional[bool] = True
+    check_kernel_version: Optional[bool] = True
```
squirrelsc, line 81: Why is it needed? Please add it in code comments.

squirrelsc (approved):

PR #3218, opened 2024-03-19
os: Update for Microsoft Azure Linux

LiliDeng:


lisa/operating_system.py
```diff
@@ -1858,7 +1858,8 @@ def name_pattern(cls) -> Pattern[str]:
 class CBLMariner(RPMDistro):
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
-        return re.compile("^Common Base Linux Mariner|mariner$")
+        # Microsoft Azure Linux 3.0
+        return re.compile("^Common Base Linux Mariner|mariner|Microsoft$")
```
xordux: Thank you for the changes. Perhaps it will be better to be more specific with the new name:
```suggestion
        return re.compile("^Common Base Linux Mariner|mariner|Microsoft Azure Linux$")
```

LiliDeng: yeah, but here https://github.com/microsoft/lisa/blob/main/lisa/operating_system.py#L113 it only matches the first string `Microsoft`

jslobodzian (approved):

squirrelsc (approved):

PR #3220, opened 2024-03-19
Disabling ASA extension for AzSecPack

suresh-thelkar:
Disabling ASA extension for AzSecPack

### DIFF
--- a/microsoft/testsuites/vm_extensions/azsecpack.py
+++ b/microsoft/testsuites/vm_extensions/azsecpack.py
@@ -123,7 +123,7 @@ class AzSecPack(TestSuite):
         # Install agent extension
         try:
             self._install_monitor_agent_extension(node)
-            self._install_security_agent_extension(node, log)
+            # self._install_security_agent_extension(node, log)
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/azsecpack.py
```diff
@@ -123,7 +123,7 @@ def verify_azsecpack(self, node: Node, log: Logger, result: TestResult) -> None:
         # Install agent extension
         try:
             self._install_monitor_agent_extension(node)
-            self._install_security_agent_extension(node, log)
+            # self._install_security_agent_extension(node, log)
```
squirrelsc, line 126: 1. Please explain the reason to comment out it in the PR.
2. Remove the line directly, instead of commenting out.

suresh-thelkar: This is to test a specific custom image. In order to run a pipeline I had to raise this PR and then use its GitHub reference in the pipeline. This is the reason I marked the PR as draft.

PR #3222, opened 2024-03-19
updated verify_timesync_ntp to use freebsd specifics

r-dailey:
Made use of hwclock conditional on the os type to prevent failures from trying to install it. Also added a conditional to call the appropriate utility in freebsd for setting the rtc to system time for this test case.

### DIFF
--- a/lisa/tools/hwclock.py
+++ b/lisa/tools/hwclock.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 from datetime import datetime
-from typing import cast
+from typing import Optional, Type, cast
 
 from dateutil.parser import parser
 from retry import retry
@@ -12,6 +12,10 @@ from lisa.util import LisaException
 
 
 class Hwclock(Tool):
+    @classmethod
+    def _freebsd_tool(cls) -> Optional[Type[Tool]]:
+        return HwclockFreebsd
+
     @property
     def command(self) -> str:
         return "hwclock"
@@ -62,3 +66,19 @@ class Hwclock(Tool):
                 f" error: {command_result.stderr}"
             )
         return parser().parse(command_result.stdout)
+
+
+class HwclockFreebsd(Hwclock):
+    @property
+    def command(self) -> str:
+        return ""
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def set_rtc_clock_to_system_time(self) -> None:
+        self.node.execute(cmd="adjkerntz -i", expected_exit_code=0)


microsoft/testsuites/core/timesync.py
```diff
@@ -325,12 +325,16 @@ def verify_timesync_ntp(self, node: Node) -> None:
...
+            hwclock = node.tools[Hwclock]
         service = node.tools[Service]
         # 1. Stop systemd-timesyncd if this service exists.
         service.stop_service("systemd-timesyncd")
         # 2. Set rtc clock to system time.
-        hwclock.set_rtc_clock_to_system_time()
+        if not isinstance(node.os, BSD):
+            hwclock.set_rtc_clock_to_system_time()
+        else:
+            node.execute("adjkerntz -i")
```
squirrelsc: It should be implemented as an alternative to hwclock. You can only implement `set_rtc_clock_to_system_time` method firstly.

r-dailey: Should that be in the hwclock.py file then or should it live somewhere else?

squirrelsc: in the hwclock.py like other BSD tools

r-dailey: Added it to Hwclock.py and confirmed it solves the install issue.

lisa/tools/hwclock.py
```diff
@@ -62,3 +66,20 @@ def get(self, no_error_log: bool = True) -> datetime:
...
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def set_rtc_clock_to_system_time(self) -> None:
+        cmd_result = self.node.execute("adjkerntz -i")
+        cmd_result.assert_exit_code()
```
squirrelsc: The error code assertion can be in same call with `execute`. Check its parameters.

r-dailey: I updated the command to have the expected exit code in the execute line. I was previously just following the format of the existing function in Hwclock which has the exit value assertion on a separate line. Should I change that one to match as well?

squirrelsc: You don't need to clean up existing code, there are a lot of place...

squirrelsc:
@LiliDeng LGTM, @r-dailey please fix check failures.

squirrelsc:
@r-dailey, please use `git -i rebase` to organize commits. For simple changes, we can do squashed merge, but for complex PRs, we want to keep meaningful commits.

LiliDeng:
@r-dailey use isort tool to fix below issue, thanks.

image

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3223, opened 2024-03-20
add CVM check case for verify_os_partition_identifier

lesscodingmorehappiness:
For fixing bug https://microsoft.visualstudio.com/OS/_workitems/edit/49527142


lisa/tools/blkid.py
```diff
@@ -33,6 +35,9 @@ class Blkid(Tool):
     # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01" # noqa: E501
     _get_partition_name = re.compile(r"\s*(?P<name>\S+):.*")
 
+    # /dev/sda1: LABEL="Temporary Storage" UUID="9ED4084BD408285B" TYPE="ntfs" PARTUUID="03e90eae-01" # noqa: E501
```
squirrelsc: The original logic is low efficient, run multiple regexp on the same string. Please merge them, and call `find_group_in_lines`.
Use https://regex101.com/ to debug regexp.

microsoft/testsuites/core/storage.py
```diff
@@ -294,21 +300,34 @@ def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
         )
         os_partition_info = node.tools[Blkid].get_partition_info_by_name(os_partition)
 
+        # check if cvm
+        is_cvm = False
+        settings = Feature.get_feature_settings(
+            node.features[SecurityProfile]._settings
```
squirrelsc, line 306: 1. Avoid to use feature to check if it's CVM or not. Is there a way to check it from the VM itself?
2. The _settings is a private method. Add a method in SecurityProfile feature to return the value.

microsoft/testsuites/core/storage.py
```diff
@@ -677,9 +696,17 @@ def _get_mtab_mount_point_regex(self, mount_point: str) -> Pattern[str]:
         return regex
 
     def _check_root_partition_in_log(
-        self, log: str, os_partition_info: PartitionInfo
+        self, log: str, os_partition_info: PartitionInfo, is_cvm: bool = False
     ) -> bool:
-        if (
+        if is_cvm:
+            if not get_matched_str(
```
squirrelsc, line 702: Add code comments to explain the newly added logic.

microsoft/testsuites/core/storage.py
lesscodingmorehappiness: 1. I checked with Karmeron there's no existing mechanism in LISA to determine if it's a CVM from the VM itself.
2. will update

microsoft/testsuites/core/storage.py
lesscodingmorehappiness: added

microsoft/testsuites/core/storage.py
squirrelsc: > * I checked with Karmeron there's no existing mechanism in LISA to determine if it's a CVM from the VM itself.

Do you mean not exists in LISA or in the VM? If it's not in LISA, but in VM, we can add it.

lesscodingmorehappiness: I did some search and try and it seems not feasible to directly get this info from VM itself:
1. there's no info from dmesg/journalctl clearly indicates this is CVM;
2. the image/SKU name may contain keywords like 'confidential' but it looks not safe to me
3. az cli can get the vm metadata which contains security type to show if it's cvm or not, but this metadata seems to be stored on azure side not the vm itself. A similar command 'curl -H Metadata:true' running from vm only shows secureBootEnabled and virtualTpmEnabled info but no securityType

squirrelsc: Thanks a lot for a deep investigation.

lisa/features/security_profile.py
```diff
@@ -114,6 +114,9 @@ def can_disable(cls) -> bool:
     def enabled(self) -> bool:
         return True
 
+    def get_settings(self) -> schema.FeatureSettings:
+        return self._settings
```
squirrelsc: It can be moved to base feature class, but use a deep copy to make it read only. Also add comments it's a read only value, the changes won't be effctive.

microsoft/testsuites/core/storage.py
```diff
@@ -294,21 +300,34 @@ def verify_os_partition_identifier(self, log: Logger, node: RemoteNode) -> None:
...
         os_partition_info = node.tools[Blkid].get_partition_info_by_name(os_partition)
 
+        # check if cvm
+        is_cvm = False
+        settings = Feature.get_feature_settings(
+            node.features[SecurityProfile].get_settings()
+        )
+        assert isinstance(settings, SecurityProfileSettings)
+        if SecurityProfileType.CVM == settings.security_profile:
+            log.debug("This is CVM.")
```
squirrelsc, line 310: The description should be more technical like "Confidential computing is detected to be enabled."

lisa/features/security_profile.py
lesscodingmorehappiness: Moved

microsoft/testsuites/core/storage.py
lesscodingmorehappiness: Modified

squirrelsc:
@LiliDeng LGTM
@lesscodingmorehappiness Please use `git -i rebase` to organize commits, make sure they are meaningful. So we can use rebase and merge.

PR #3231, opened 2024-03-26
Add test to check and log mdatp install/config info

mcgov:
Adding an mdatp test to check configuration info in a VM if defender is installed.

### DIFF
--- a/microsoft/testsuites/mdatp/check-mdatp.sh
+++ b/microsoft/testsuites/mdatp/check-mdatp.sh
@@ -7,22 +7,24 @@
 # checks Azure Extension directories using globs, since they're not 
 # well-known.
 
-# Returns nonzero value on detection of any defender info
-# Return value is a bitmask showing the types of info detected.
-# Bits are ordered by seriousness of the issue:
-# Least significant bit is bad
-# Most significant bit is worst.
+# Returns nonzero value on detection of any defender info.
+# Return value are ordered by seriousness of the issue:
+# Lower value is bad
+# Highest value is worst.
+
+# Publishers: Re-run and remediate until you get a 0 return code.
+
 EXIT_CODE=0
 # if any mdatp install in /etc/opt is found
-EXIT_MDATP_AGENT_INSTALLED=1
+EXIT_MDATP_AGENT_INSTALLED=251
 # if mdatp az extension is installed
-EXIT_MDE_INSTALLED=2
+EXIT_MDE_INSTALLED=252
 # if any log dirs are found
-EXIT_MDATP_LOGS_FOUND=4
+EXIT_MDATP_LOGS_FOUND=253
 # if an installation log is found
-EXIT_MDATP_INSTALL_LOGS_FOUND=8
+EXIT_MDATP_INSTALL_LOGS_FOUND=254
 # if an onboarding blob is found
-EXIT_ONBOARD_INFO_FOUND=16
+EXIT_ONBOARD_INFO_FOUND=255
 MDATP_OPT_DIR='/etc/opt/microsoft/mdatp'
 MDATP_LOG_DIR='/var/log/microsoft/mdatp'
 ERROR_MSG_HEADER="----------------------------------------------------------------------"
@@ -43,7 +45,7 @@ check_unexpected_file () {
 for mde_dir in /var/lib/waagent/Microsoft.Azure.AzureDefenderForServers.MDE.Linux* ; do
     # for all the versioned folders we find...
     if [ -e "$mde_dir" ]; then
-        EXIT_CODE=$((EXIT_CODE|EXIT_MDE_INSTALLED))
+        EXIT_CODE=$EXIT_MDE_INSTALLED
         if [ -d "$mde_dir" ]; then
             # find regular files, skip printing them if they are binary
             find_printable_files "$mde_dir"
@@ -58,7 +60,7 @@ done
 for log_dir in /var/log/azure/Microsoft.Azure.AzureDefenderForServers.MDE.Linux* ; do
     # for all the versioned folders we find...
     if [ -e "$log_dir" ]; then
-        EXIT_CODE=$((EXIT_CODE|EXIT_MDATP_LOGS_FOUND))
+        EXIT_CODE=$EXIT_MDATP_LOGS_FOUND
         echo "checking $log_dir..."
         if [ -d "$log_dir" ]; then
             find_printable_files "$log_dir"
@@ -73,7 +75,7 @@ done
 for log_dir in /var/lib/GuestConfig/extension_logs/Microsoft.Azure.AzureDefenderForServers.MDE.Linux* ; do
     # for all the versioned folders we find...
     if [ -e "$log_dir" ]; then
-        EXIT_CODE=$((EXIT_CODE|EXIT_MDATP_LOGS_FOUND))
+        EXIT_CODE=$EXIT_MDATP_LOGS_FOUND
         echo "checking $log_dir..."
         if [ -d "$log_dir" ]; then
             find_printable_files "$log_dir"
@@ -85,13 +87,13 @@ done
 
 # check for mde agent install in /etc/opt
 if [  -e "$MDATP_OPT_DIR" ]; then
-    EXIT_CODE=$((EXIT_CODE|EXIT_MDATP_AGENT_INSTALLED))
+    EXIT_CODE=$EXIT_MDATP_AGENT_INSTALLED
     find_printable_files "$MDATP_OPT_DIR"
 fi
 
 # check for install or runtime logs in /var/log
 if [  -e "$MDATP_LOG_DIR" ]; then
-    EXIT_CODE=$((EXIT_CODE|EXIT_MDATP_LOGS_FOUND))
+    EXIT_CODE=$EXIT_MDATP_LOGS_FOUND
     find_printable_files "$MDATP_LOG_DIR"
 fi
 
@@ -100,7 +102,7 @@ if [ -f "$MDATP_LOG_DIR/install.log" ]; then
     echo "$ERROR_MSG_HEADER" >&2
     echo "ERROR: mdatp install logs are present in this image!" >&2
     echo "Publishers should remove this data before publishing public images." >&2
-    EXIT_CODE=$((EXIT_CODE|EXIT_MDATP_INSTALL_LOGS_FOUND))
+    EXIT_CODE=$EXIT_MDATP_INSTALL_LOGS_FOUND
 fi
 
 # special log line for mdatp_onboard.json
@@ -108,7 +110,7 @@ if [ -f "$MDATP_OPT_DIR/mdatp_onboard.json" ]; then
     echo "$ERROR_MSG_HEADER" >&2
     echo "ERROR: mdatp onboarding info is present in this image!" >&2
     echo "Publishers should remove this data before publishing public images." >&2
-    EXIT_CODE=$((EXIT_CODE|EXIT_ONBOARD_INFO_FOUND))
+    EXIT_CODE=$EXIT_ONBOARD_INFO_FOUND
 fi
 
 # returns nonzero value if defender info is found
--- a/microsoft/testsuites/mdatp/check-mdatp.sh
+++ b/microsoft/testsuites/mdatp/check-mdatp.sh
@@ -1,0 +1,115 @@
+#! /bin/sh
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+# check the following folders for printable files, print them
+# checks Azure Extension directories using globs, since they're not 
+# well-known.
+
+# Returns nonzero value on detection of any defender info
+# Return value is a bitmask showing the types of info detected.
+# Bits are ordered by seriousness of the issue:
+# Least significant bit is bad
+# Most significant bit is worst.
+EXIT_CODE=0
+# if any mdatp install in /etc/opt is found
+EXIT_MDATP_AGENT_INSTALLED=1
+# if mdatp az extension is installed
+EXIT_MDE_INSTALLED=2
+# if any log dirs are found
+EXIT_MDATP_LOGS_FOUND=4
+# if an installation log is found
+EXIT_MDATP_INSTALL_LOGS_FOUND=8
+# if an onboarding blob is found
+EXIT_ONBOARD_INFO_FOUND=16
+MDATP_OPT_DIR='/etc/opt/microsoft/mdatp'
+MDATP_LOG_DIR='/var/log/microsoft/mdatp'
+ERROR_MSG_HEADER="----------------------------------------------------------------------"
+
+find_printable_files () {
+    found_dir="$1"
+    sudo find "$found_dir" -type f -exec file '{}' ';' -exec grep -Iq . '{}' ';' -exec head -c 1024 '{}' ';'
+}
+
+check_unexpected_file () {
+    surprise_file="$1"
+    echo "Found unexpected regular file: $surprise_file" >&2
+    # see what it is, print if it's printable
+    file "$surprise_file"; grep -Iq . "$surprise_file" && cat "$surprise_file"
+}
+
+# check for MDE extension installation, folder is version labeled, so use a shell glob
+for mde_dir in /var/lib/waagent/Microsoft.Azure.AzureDefenderForServers.MDE.Linux* ; do
+    # for all the versioned folders we find...
+    if [ -e "$mde_dir" ]; then
+        EXIT_CODE=$((EXIT_CODE|EXIT_MDE_INSTALLED))
+        if [ -d "$mde_dir" ]; then
+            # find regular files, skip printing them if they are binary
+            find_printable_files "$mde_dir"
+        else
+            # not expecting to find a regular file instead of a dir, but...
+            check_unexpected_file "$mde_dir"
+        fi
+    fi
+done
+
+# check for ARM log files
+for log_dir in /var/log/azure/Microsoft.Azure.AzureDefenderForServers.MDE.Linux* ; do
+    # for all the versioned folders we find...
+    if [ -e "$log_dir" ]; then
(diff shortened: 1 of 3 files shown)


microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,57 @@
...
+            shell=True,
+            sudo=True,
+        )
+        script_output = result.stdout.strip()
+        exit_code = result.exit_code
+        if exit_code != 0:
+            raise SkippedException("Did not find mdatp installation on this VM.")
+
+        node.log.info("Found mdatp installation!")
+        node.log.info(f"Found config information: {script_output}")
```
squirrelsc: From the logic, it looks the test case always passed or skipped. It doesn't provide useful information.

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,57 @@
...
+from lisa.operating_system import Linux
+from lisa.sut_orchestrator import AZURE
+from lisa.testsuite import simple_requirement
+
+
+@TestSuiteMetadata(
+    area="mdatp",
+    category="functional",
+    description="""
+        Tests to check the behavior of the virtual cdrom device in VMs.
```
squirrelsc: it's inconsistent with test case description. Is it testing the cdrom or defender? What's the "defender"? Is it "Microsoft Defender"?

microsoft/testsuites/mdatp/mdatp.py
mcgov: > From the logic, it looks the test case always passed or skipped
This is true
>  It doesn't provide useful information.
I disagree. Right now this test either skips because no config info is present or passes if config info is found. We can use this to look for images which ship with mdatp onboarding or config info. Right now we don't know if this is a bug or not, Lili and Baihua and I are on a chat with that team. When we have enough info to validate the info, we'll add a fail condition where if there is config info and it's unexpected, we fail the test.

microsoft/testsuites/mdatp/mdatp.py
mcgov: Ah thanks, will fix

mcgov: edited to keep it consistent. I'm using mdatp as a catch-all term for defender for cloud (MDC) and defender endpoint (MDE) which are two flavors of of ms defender

microsoft/testsuites/mdatp/check-mdatp.sh
```diff
@@ -0,0 +1,29 @@
+#! /bin/sh
+
+# check the following folders for printable files, print them
+FOUND_DEFENDER=1
+MDATP_OPT_DIR='/etc/opt/microsoft/mdatp/'
+MDATP_EXTENSION_DIR='/var/lib/waagent/Microsoft.Azure.AzureDefenderForServers.MDE.Linux'
```
okopel: The extension directory is startsWith "/var/lib/waagent/Microsoft.Azure.AzureDefenderForServers.MDE.Linux"
The real path is /var/lib/waagent/Microsoft.Azure.AzureDefenderForServers.MDE.Linux-1.X.X.X" (version)

(MDE.Linux extension owner)

okopel: Do you want to add MDATP_EXTENSION_LOG_DIR also?
For ARM = /var/log/azure/Microsoft.Azure.AzureDefenderForServers.MDE.Linux
For ARC = /var/lib/GuestConfig/extension_logs/Microsoft.Azure.AzureDefenderForServers.MDE.Linux-<version>

okopel (changes requested):

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: You can define a test variable. If it's enforced and not found, the test case will fail. For example, "verify_mdatp_config_enable_assert".

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,95 @@
...
+            in script_output
+        )
+        error_header = ""
+        if found_onboarding_info:
+            error_header += "mdatp onboarding information is present in this image!!"
+        if found_install_logs:
+            error_header += "mdatp installation logs are present in this image!!"
+        if found_mde_extension:
+            error_header += "MDE extension data was found in /var/lib/waagent!"
+        if any([found_onboarding_info, found_install_logs, found_mde_extension]):
```
squirrelsc: It looks either way, the test case will be failed, if the mdatp is found.

mcgov: Yes. I don't like this solution because people could intentionally publish an image with defender on it, but not have any leaked onboarding credentials in the image. We can fix it later if that ends up being a common use case and we have tons of false positives.

microsoft/testsuites/mdatp/check-mdatp.sh
mcgov: yes :) Thank you!

mcgov: Ahhhhhh thank you!!

mcgov: I was wondering why we hadn't seen any of these yet 😜

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: From the logic, the test case either skipped or failed. No path to be passed. The pass rate is always 0. Is that intentional?

microsoft/testsuites/mdatp/check-mdatp.sh
mcgov: added

mcgov: fixed, using globs to grab whatever is there.

microsoft/testsuites/mdatp/mdatp.py
mcgov: It was, but now that you mention it, let's change it.

mcgov: fixed :)

squirrelsc: Probability yes... See below simplified logic, it looks like the test case will fail, when mdatp is detected. It's hard to understand why, and what issues the test case is to prevent.

```python
if any([found_onboarding_info, found_install_logs, found_mde_extension]):
    fail(...
else:
    fail(...
```

microsoft/testsuites/mdatp/mdatp.py
mcgov: I don't think we need it, for now this should always result in a fail if info is present.

microsoft/testsuites/mdatp/mdatp.py
mcgov: It's so you get a different error message if there's a known problem or an unknown problem. It makes it easier to filter results and jump to ones that might need fixing.

mcgov: I'll add some more comments, you're right tho :)

mcgov: commented

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,100 @@
...
+    category="functional",
+    description="""
+        Test to verify there are no pre-installed copies of mdatp.
+    """,
+)
+class MdatpSuite(TestSuite):
+    @TestCaseMetadata(
+        description="""
+           Check for mdatp endpoint/cloud install, dump config info.
+           Fails if mdatp is present in the image without explicitly
```
squirrelsc: What means "without explicitly installing it"? Does the logic detect it, and pass in this case?

mcgov: I guess it's an unneccesary adjective...
It means we are chekcing for the presence of mdatp without installing it. The test doesn't install it, so we can safely assume if it's present when the test runs, we didn't install it.

mcgov: I've changed a few things to make the failure path more clear, add comments about changes that might be needed later, and changed the description based on your feedback.

squirrelsc:
@LiliDeng LGTM

microsoft/testsuites/mdatp/check-mdatp.sh
```diff
@@ -0,0 +1,113 @@
...
+            # find regular files, skip printing them if they are binary
+            find_printable_files "$mde_dir"
+        else
+            # not expecting to find a regular file in /var/log/azure, but...
+            check_unexpected_file "$mde_dir"
+        fi
+    fi
+done
+
+# check for azure extension install dir
```
okopel: check for **Mde agent** install dir
No extension dir

microsoft/testsuites/mdatp/check-mdatp.sh
```diff
@@ -0,0 +1,113 @@
...
+            find_printable_files "$log_dir"
+        else
+            # not expecting to find a regular file in /var/log/azure, but...
+            check_unexpected_file "$log_dir"
+        fi
+    fi
+done
+
+# check for ARC mde extension logs
+for log_dir in /var/lib/GuestConfig/extension_logs/Microsoft.Azure.AzureDefenderForServers.MDE.Linux-* ; do
```
okopel: Note that the **log folders** on **ARC** look like:
- /var/lib/GuestConfig/extension_logs/Microsoft.Azure.AzureDefenderForServers.MDE.Linux-1.0.3.12
- /var/lib/GuestConfig/extension_logs/Microsoft.Azure.AzureDefenderForServers.MDE.Linux-1.0.3.13

and **log folders** in **ARM** like:
- /var/log/azure/Microsoft.Azure.AzureDefenderForServers.MDE.Linux 
without version's folder, just files.

The **Mde-Extension** files (no logs) on both Arm and Arc look like
- /var/lib/waagent/Microsoft.Azure.AzureDefenderForServers.MDE.Linux-1.0.3.12/
- /var/lib/waagent/Microsoft.Azure.AzureDefenderForServers.MDE.Linux-1.0.3.13/

microsoft/testsuites/mdatp/check-mdatp.sh
mcgov: Thanks! Fixed the comment

microsoft/testsuites/mdatp/check-mdatp.sh
mcgov: Thank you! I've made each section more clear and added an exit code to differentiate the extension install from the agent install. The test fails either way but we'll get a nice error message detailing what was present.

srkoripe (approved):
LGTM from mdatp side. Pls get a review from the MDC agent side too if possible

mcgov:
> LGTM from mdatp side. Pls get a review from the MDC agent side too if possible

Already reviewed! Thanks so much. @LiliDeng can we merge this ASAP?

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,125 @@
...
+            error_header += "MDE extension installation was found in this image! "
+        if not any(
+            [
+                found_onboarding_info,
+                found_install_logs,
+                found_mdatp_installed,
+                found_mde_extension_install,
+            ]
+        ):
+            raise SkippedException(
```
squirrelsc: Should it mean passed?

mcgov: No. If the script returns anything but 0 and none of those four things above are detected, something is broken. We didn't detect any of the stuff we're looking for, but the script didn't return 0 indicating a clean run.  We shouldn't pass a test when something is broken, so it skips. We'll be able to filter for those results and look closer at them, if this ever actually occurs. Out of the 6000 images I've tested so far, it has never occurred.

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,125 @@
...
+        Test to verify there are no pre-installed copies of mdatp.
+    """,
+)
+class MdatpSuite(TestSuite):
+    @TestCaseMetadata(
+        description="""
+           Check for mdatp endpoint/cloud install, dump config info.
+           Fails if mdatp is installed in the image.
+           Raises specific error messages depending on the type of info
+           foud
```
LiliDeng: found

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,125 @@
...
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            unsupported_os=[Windows], supported_platform_type=[AZURE]
```
LiliDeng, line 42: how about BSD?

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: it needs at least one path to be Passed, or there is no way to calculate pass rate, evaluate the quality of the test case.

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,125 @@
...
+            shell=True,
+            sudo=True,
+            no_debug_log=True,
+        )
+        script_output = result.stdout.strip()
+        exit_code = result.exit_code
+        if exit_code is None:
+            raise SkippedException("exit code was None after running check-mdatp")
+        if exit_code == 0:
+            node.log.info("No mdatp onboarding info found, image is clean.")
```
mcgov, line 72: @squirrelsc the pass code path is here.

microsoft/testsuites/mdatp/mdatp.py
mcgov: I think we covered this already, there is a path to passing.

microsoft/testsuites/mdatp/mdatp.py
mcgov: It's supported.

mcgov: Not that there are any images in the marketplace for this, but MacOS is also supported.
The shell script will work on any unix-like posix shell, not just bash. I tried to avoid anything bash-specific to maximize compatibility.

microsoft/testsuites/mdatp/mdatp.py
mcgov: Thanks! Fixed

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -0,0 +1,134 @@
...
+        # and run the script to check and dump defender info
+        result = node.execute(
+            cmd=f"{str(script_path)}",
+            shell=True,
+            sudo=True,
+            no_debug_log=True,
+        )
+        script_output = result.stdout.strip()
+        exit_code = result.exit_code
+        if exit_code is None:
```
squirrelsc: it's impossible to be None from LISA logic. If it needs, assert is not None.

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: From the message, "no recognized error code found." if no error, it should be passed, why skip it? If as you said, it won't happen, if it happens, that's something unexpected. It should be failed. It doesn't make sense to skip here.

mcgov: Does SKIP make an image fail in AzCertify or something? 
If we get an unexpected error code it implies we're running on a non-posix system, like maybe we've ended up on ReactOS or something. I think SerenityOS has a posix shell so that one will probably work, even though defender isn't supported. I don't feel good about giving a pass when we haven't found what we're looking for and the script probably hasn't even run correctly.

I changed this code path because of feedback from you, it was a hard fail if the return code was nonzero. You said this needs to be 100% reliable to use with AzCertify, so I modified the script such that we don't use bash-specifics, we look for specific things in the shell script, we don't rely on string parsing, and we allow for a skip case in the event we can't validate anything correctly.

I've taken time to make sure this shell script works with all unix shells including BSD, XNU/MacOS/OSX, and Linux.  If this returns anything but the 5 error codes in the script: we're not running in an OS which supports defender **as of 4/11/2024**. 

If we're assuming this test is going to be around longer than 10 years, I don't think we should give new unexpected OS's or unexpected shells the same passing grade as a known, validated OS (or non-compliant shell).

Having said that, you have made me rethink the error codes. It would probably be better to return something higher than 127 so we don't conflict with errno? I'm going to change that, it would be nice to get real exit codes for OS errors. Then we can change the message to "Unexpected exit code found: "

microsoft/testsuites/mdatp/mdatp.py
mcgov: fixed, thanks!

microsoft/testsuites/mdatp/mdatp.py
mcgov: Fixed 🙂

mcgov: By the way, is there a way to raise a PASS with a failure message? I find I often want to add a 'pass' note and it would be helpful for analyzing results after the test run. We can discuss offline.

LiliDeng: > By the way, is there a way to raise a PASS with a failure message? I find I often want to add a 'pass' note and it would be helpful for analyzing results after the test run. We can discuss offline.

`PassedException` may meet your requirement.

mcgov: @squirrelsc I was wrong, after re-writing this skip message... it should be a fail! Thanks for chatting w me offline and helping me understand your thought process, concerns, and what actions were needed.

mcgov: My thought is, if there's suddenly a new posix-shell incompatible image in the marketplace... we need to take this into account. We also don't want to allow people to skip the check by fiddling with the shell defaults.

squirrelsc:
@LiliDeng Current version LGTM.

LiliDeng (approved):

PR #3232, opened 2024-03-26
Fix serial disk add testcase logic

SRIKKANTH:
Current logic is adding and removing a single disk instead of adding 1 disk at a time till max supported disks.

Modified the logic to add 1 disk at a time till max allowed disks and for removal all attached disks one after the other.

### DIFF
--- a/microsoft/testsuites/core/storage.py
+++ b/microsoft/testsuites/core/storage.py
@@ -586,11 +586,13 @@ class Storage(TestSuite):
 
         # get partition info before adding data disk
         partitions_before_adding_disk = lsblk.get_disks(force_run=True)
-
+        added_disk_count = 0
+        disks_added = []
         for _ in range(disks_to_add):
             # add data disk
-            log.debug("Adding 1 managed disk")
-            disks_added = disk.add_data_disk(1, disk_type, size)
+            disks_added.append(disk.add_data_disk(1, disk_type, size))
+            added_disk_count += 1
+            log.debug(f"Added {added_disk_count}'th managed disk")
 
             # verify that partition count is increased by 1
             # and the size of partition is correct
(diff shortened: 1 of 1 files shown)


LiliDeng:
> Current logic is adding and removing a single disk instead of adding 1 disk at a time till max supported disks.
> 
> Modified the logic to add 1 disk at a time till max allowed disks and for removal all attached disks one after the other.

Is there any test issue for current code? Can you clarify the reason for changing code? Thanks.

SRIKKANTH:
> > Current logic is adding and removing a single disk instead of adding 1 disk at a time till max supported disks.
> > Modified the logic to add 1 disk at a time till max allowed disks and for removal all attached disks one after the other.
> 
> Is there any test issue for current code? Can you clarify the reason for changing code? Thanks.

Current testcase simply verifies adding removing a single disk instead of actually checking if its allowed to attach one disk at a time till max supported disks.

Is this expected behavior? 

If yes why the add remove iteration count is equal to max disks allowed on the SKU why not add remove with a fixed number of times. 

I think we should validate adding disks serially till max allowed disks and remove all the attached disks in the end.

adityagesh:
Your change aligns with the description of the test case (pasting it below)

```
        1. Get maximum number of data disk for the current vm_size.
        2. Get the number of data disks already added to the vm.
        3. Serially add and remove the data disks and verify that the added
        disks are present in the vm.
```

However, the name of the function is misleading, how about renaming it similar to something like hot_add_disk_serial_one_by_one? Similar to how we have for the nic

Thoughts on this? @LiliDeng

microsoft/testsuites/core/storage.py
```diff
@@ -567,11 +567,14 @@ def _hot_add_disk_serial(
 
         # get partition info before adding data disk
         partitions_before_adding_disk = lsblk.get_disks(force_run=True)
-
+        added_disk_count = 0
+        disks_added = []
         for _ in range(disks_to_add):
             # add data disk
             log.debug("Adding 1 managed disk")
```
LiliDeng: remove this log

SRIKKANTH: Removed.

LiliDeng (approved):

PR #3238, opened 2024-03-28
mariner 3.0: fix compile issue for xdpdump

LiliDeng:
I have tested on mariner 2.0, 3.0 and ubuntu 20.04. no regression issue.

### DIFF
--- a/microsoft/testsuites/xdp/xdpdump.py
+++ b/microsoft/testsuites/xdp/xdpdump.py
@@ -89,6 +89,18 @@ class XdpDump(Tool):
         ).is_equal_to(self._code_path.parent)
         git.init_submodules(cwd=self._code_path)
 
+        # This workaround addresses a discrepancy in the bpf_perf_event_read_simple
+        # function declaration and definition detected by the latest version of
+        # the clang C compiler. Specifically, the return value is incorrectly specified
+        # as an int in the .h file but as an enum in the C file.
+        sed = self.node.tools[Sed]
+        sed.substitute(
+            "LIBBPF_API int bpf_perf_event_read_simple",
+            "LIBBPF_API enum bpf_perf_event_ret bpf_perf_event_read_simple",
+            f"{self._code_path.parent}/libbpf/src/libbpf.h",
+            sudo=True,
+        )
+
         # create a default version for exists checking.
         make = self.node.tools[Make]
         make.make(


jslobodzian (approved):

microsoft/testsuites/xdp/xdpdump.py
```diff
@@ -89,6 +89,15 @@ def _install(self) -> bool:
         ).is_equal_to(self._code_path.parent)
         git.init_submodules(cwd=self._code_path)
 
+        # fix the build issue, higher compiler version has different return
```
jslobodzian: This is a work-around for a code issue detected by the latest clang C compiler. Clang is properly detecing and reporting that there is a mismatch between the function declaration of bpf_perf_event_read_simple and the function's definition.   Specifically, the return value is incorrectly specified as an int in the .h file  and is reported as an enum in the .c file.  The sed command fixes the function declaration to match the definition.

The affected code is in libbpf.h.  Unfortunately that code is not owned by Microsoft.  Also, this code is included as a git submodule.  The submodule is pinned to 6 year old code.  To fix this properly the xdpdump repository needs to be updated to take advantage of the latest libbpf code.

jslobodzian (approved):

microsoft/testsuites/xdp/xdpdump.py
LiliDeng: update the comments

anphel31 (approved):

microsoft/testsuites/xdp/xdpdump.py
```diff
@@ -89,6 +89,18 @@ def _install(self) -> bool:
         ).is_equal_to(self._code_path.parent)
         git.init_submodules(cwd=self._code_path)
 
+        # This workaround addresses a discrepancy in the bpf_perf_event_read_simple
```
squirrelsc, line 92: Does it work for all distros?

LiliDeng: I tested on mairner 2.0 and ubuntu 20.04 which don't change the compiler recently, I will cover more distro.

LiliDeng: I have tested the earliest images which pass this test case.

squirrelsc (approved):

PR #3242, opened 2024-03-31
Added hot add disks at random lun testcases

SRIKKANTH:
Below testcases added:
1. verify_hot_add_disk_serial_random_lun_standard_ssd
2. verify_hot_add_disk_serial_random_lun_premium_ssd

### DIFF
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 from functools import partial
-from typing import Any, List, Optional, Type
+from typing import Any, Dict, List, Optional, Type
 
 from assertpy.assertpy import assert_that
 
@@ -56,6 +56,7 @@ class Disk(Feature):
         count: int,
         disk_type: schema.DiskType = schema.DiskType.StandardHDDLRS,
         size_in_gb: int = 20,
+        lun: int = -1,
     ) -> List[str]:
         raise NotImplementedError
 
@@ -66,6 +67,9 @@ class Disk(Feature):
         self.disks: List[str] = []
 
     def get_resource_disk_mount_point(self) -> str:
+        raise NotImplementedError
+
+    def get_luns(self) -> Dict[str, int]:
         raise NotImplementedError
 
     # Get boot partition of VM by looking for "/boot" and "/boot/efi"
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -1629,51 +1629,18 @@ class Disk(AzureFeatureMixin, features.Disk):
         vm = get_vm(azure_platform, self._node)
         return vm.storage_profile.disk_controller_type
 
-    def get_raw_data_disks(self) -> List[str]:
-        # Handle BSD case
-        if isinstance(self._node.os, BSD):
-            return self._get_raw_data_disks_bsd()
-
-        # disk_controller_type == NVME
-        node_disk = self._node.features[Disk]
-        if node_disk.get_os_disk_controller_type() == schema.DiskControllerType.NVME:
-            # Getting OS disk nvme namespace and disk controller used by OS disk.
-            # Sample os_boot_partition:
-            # name: /dev/nvme0n1p15, disk: nvme, mount_point: /boot/efi, type: vfat
-            os_boot_partition = node_disk.get_os_boot_partition()
-            if os_boot_partition:
-                os_disk_namespace = get_matched_str(
-                    os_boot_partition.name,
-                    self.NVME_NAMESPACE_PATTERN,
-                )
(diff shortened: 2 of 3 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -1629,51 +1629,17 @@ def get_hardware_disk_controller_type(self) -> Any:
...
-            for name_space in nvme_namespaces:
-                if (
-                    name_space.startswith(os_disk_controller)
-                    and name_space != os_disk_namespace
-                ):
-                    disk_array.append(name_space)
-            return disk_array
-
-        # disk_controller_type == SCSI
+    def get_azure_scsi_discs(self) -> List[str]:
```
squirrelsc: `_get_scsi_disks`

lisa/features/disks.py
```diff
@@ -68,6 +69,9 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
     def get_resource_disk_mount_point(self) -> str:
         raise NotImplementedError
 
+    def get_linux_device_luns(self) -> Dict[str, int]:
```
squirrelsc: `get_luns`

SRIKKANTH: The method gets the luns from within the VM. So I felt its better if we can put that in the same. 
get_linux_luns seems more dev friendly to me.

squirrelsc: I know the logic applies to `linux` only, but we can extend it if needed. So the name could be called `get_luns`.

SRIKKANTH: This method specifically brought to cross check if luns from linux are matching with azure or not. If there is any such method needed for azure we can name it accordingly.

lisa/sut_orchestrator/azure/features.py
squirrelsc: 1. it's not used in other places, so it should add `_` to specify it's a private method.
2. all other places names are "disk", so keep it consistent to avoid using "disc".

lisa/features/disks.py
squirrelsc: This is in base class, maybe extended only in downstream. So make all method names to be right at first time.

lisa/features/disks.py
```diff
@@ -56,6 +56,7 @@ def add_data_disk(
         count: int,
         disk_type: schema.DiskType = schema.DiskType.StandardHDDLRS,
         size_in_gb: int = 20,
+        user_lun: int = -1,
```
squirrelsc: all values are set by caller/user, so call it `lun`.

microsoft/testsuites/core/storage.py
```diff
@@ -581,18 +643,21 @@ def _hot_add_disk_serial(
...
-
         # get partition info before adding data disk
         partitions_before_adding_disk = lsblk.get_disks(force_run=True)
         added_disk_count = 0
         disks_added = []
-        for _ in range(disks_to_add):
+        # Assuming existing disks added sequentially from lun = 0 to
+        # (current_data_disk_count - 1)
+        free_luns = list(range(current_data_disk_count, max_data_disk_count))
+        if randomize_luns:
```
squirrelsc, line 658: Avoid to use random in the test case, it makes the test case behavior inconsistent. You can use a fixed seed to make sure the random is reproable every time.

lisa/features/disks.py
SRIKKANTH: Changed the method name to get_luns

lisa/features/disks.py
SRIKKANTH: renamed user_lun to lun

microsoft/testsuites/core/storage.py
SRIKKANTH: Used seed instead of just random

lisa/sut_orchestrator/azure/features.py
squirrelsc: Some variables still using "disc", please change them to "disk".

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1629,51 +1629,18 @@ def get_hardware_disk_controller_type(self) -> Any:
...
-            for name_space in nvme_namespaces:
-                if (
-                    name_space.startswith(os_disk_controller)
-                    and name_space != os_disk_namespace
-                ):
-                    disk_array.append(name_space)
-            return disk_array
-
-        # disk_controller_type == SCSI
+    def _get_scsi_data_discs(self) -> List[str]:
```
squirrelsc: use "disk", not "disc"

squirrelsc:
@LiliDeng, please help take a look.

LiliDeng (approved):

PR #3246, opened 2024-04-02
DPDK: Skip failsafe test on MANA

mcgov:
Add a skip for unsupported failsafe pmd config on MANA

### DIFF
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -310,6 +310,9 @@ def initialize_node_resources(
     extra_nics: Union[List[NicInfo], None] = None,
 ) -> DpdkTestResources:
     _set_forced_source_by_distro(node, variables)
+    if pmd == "failsafe" and node.nics.is_mana_device_present():
+        raise SkippedException("Failsafe PMD test on MANA is not supported.")
+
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -310,6 +310,9 @@ def initialize_node_resources(
     extra_nics: Union[List[NicInfo], None] = None,
 ) -> DpdkTestResources:
     _set_forced_source_by_distro(node, variables)
+    if pmd == "failsafe" and node.nics.is_mana_device_present():
+        raise SkippedException("Failsafe PMD test on MANA is not supported.")
```
squirrelsc, line 314: Raise UnsupportedOsException in tools, and catch it in test cases to skip.

mcgov: This file is for shared code, every DPDK test calls `initialize_node_resources`. I would prefer to have two lines that apply to every test rather than wrapping every test in the same try/catch.

squirrelsc: That's easy to make some test cases should be failed to be skipped. But the tool is in test suite folder, if you think it's ok, you can do like it.

squirrelsc:
@LiliDeng LGTM

PR #3251, opened 2024-04-04
NIC: fix module information for Mana

olshanov:
There are no mana_en driver exists in upstream. The mana driver could be omitted since modprobe will remove it anyway.

### DIFF
--- a/lisa/nic.py
+++ b/lisa/nic.py
@@ -130,9 +130,7 @@ class Nics(InitializableMixin):
     _device_module_map = {
         "mlx5_core": ModuleInformation(["mlx5_ib"], "CONFIG_MLX5_CORE"),
         "mlx4_core": ModuleInformation(["mlx4_en", "mlx4_ib"], "CONFIG_MLX4_CORE"),
-        "mana": ModuleInformation(
-            ["mana", "mana_en", "mana_ib"], "CONFIG_MICROSOFT_MANA"
-        ),
+        "mana": ModuleInformation(["mana_ib"], "CONFIG_MICROSOFT_MANA"),
     }
 
     def __init__(self, node: "Node"):


lisa/nic.py
```diff
@@ -130,9 +130,7 @@ class ModuleInformation:
     _device_module_map = {
         "mlx5_core": ModuleInformation(["mlx5_ib"], "CONFIG_MLX5_CORE"),
         "mlx4_core": ModuleInformation(["mlx4_en", "mlx4_ib"], "CONFIG_MLX4_CORE"),
-        "mana": ModuleInformation(
-            ["mana", "mana_en", "mana_ib"], "CONFIG_MICROSOFT_MANA"
-        ),
+        "mana": ModuleInformation(["mana_ib"], "CONFIG_MICROSOFT_MANA"),
```
squirrelsc, line 133: When removing mod, there is a parameter `ignore_error`. It can be used to ignore non-exists mods, instead of removing them from here.

olshanov: `mana_en` doesn't exist now and as far as I can see never existed. Why keep it here?

squirrelsc: It should exist in other scenarios or images, maybe in test stages.

olshanov: https://github.com/torvalds/linux/commits/master/drivers/net/ethernet/microsoft/mana/Makefile

There only 2 commits that change driver makefile and it always build as mana.ko.
All other images that I saw have `mana.ko` built-in, in that case ModuleInformation isn't used.
I will build `mana` driver same way, it will pass tests that I need. Just reject this PR.

squirrelsc:
Found a workaround, and the change is not necessary.

PR #3255, opened 2024-04-08
Add Image Feature Requirements

kamalca:
Each feature can now implement a function `create_image_requirement` to determine a requirement based on the raw_capabilities of the image. The requirement is then merged with the platform requirement.

### DIFF
--- a/lisa/feature.py
+++ b/lisa/feature.py
@@ -87,6 +87,16 @@ class Feature(InitializableMixin):
         """
         return None
 
+    @classmethod
+    def create_image_requirement(
+        cls, image: schema.ImageSchema
+    ) -> Optional[schema.FeatureSettings]:
+        """
+        It's called in the platform to check if an image restricts the feature or not.
+        If it's restricted, create a setting.
+        """
+        return None
+
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         """
         override for initializing
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -1620,6 +1620,11 @@ class ImageSchema:
     pass
 
 
+class ArchitectureType(str, Enum):
+    x64 = "x64"
+    Arm64 = "Arm64"
+
+
 def load_by_type(schema_type: Type[T], raw_runbook: Any, many: bool = False) -> T:
     """
     Convert dict, list or base typed schema to specified typed schema.
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -12,7 +12,17 @@ from functools import lru_cache, partial
 from pathlib import Path, PurePath
 from threading import Lock
 from time import sleep, time
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Tuple,
+    Type,
+    Union,
+)
 
 import requests
 from assertpy import assert_that
@@ -24,7 +34,11 @@ from azure.keyvault.certificates import (
 )
 from azure.keyvault.secrets import SecretClient
 from azure.mgmt.compute import ComputeManagementClient
-from azure.mgmt.compute.models import VirtualMachine
+from azure.mgmt.compute.models import (
+    CommunityGalleryImage,
+    GalleryImage,
+    VirtualMachine,
+)
 from azure.mgmt.keyvault import KeyVaultManagementClient
 from azure.mgmt.keyvault.models import (
     AccessPolicyEntry,
@@ -77,6 +91,7 @@ from retry import retry
 from lisa import feature, schema, search_space
 from lisa.environment import Environment, load_environments
 from lisa.feature import Features
+from lisa.features.security_profile import SecurityProfileType
 from lisa.node import Node, RemoteNode, local
 from lisa.secret import PATTERN_HEADTAIL, PATTERN_URL, add_secret, replace
 from lisa.tools import Ls
@@ -194,12 +209,67 @@ class AzureVmPurchasePlanSchema:
 
 @dataclass_json
 @dataclass
-class AzureImageSchema:
+class AzureImageSchema(schema.ImageSchema):
+    architecture: Union[
+        schema.ArchitectureType, search_space.SetSpace[schema.ArchitectureType]
+    ] = field(  # type: ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            is_allow_set=True,
+            items=[schema.ArchitectureType.x64, schema.ArchitectureType.Arm64],
+        ),
+        metadata=field_metadata(
+            decoder=partial(
+                search_space.decode_nullable_set_space,
+                base_type=schema.ArchitectureType,
+                is_allow_set=True,
+                default_values=[
+                    schema.ArchitectureType.x64,
+                    schema.ArchitectureType.Arm64,
+                ],
+            )
+        ),
+    )
+    disk_controller_type: Optional[
+        Union[
+            search_space.SetSpace[schema.DiskControllerType], schema.DiskControllerType
+        ]
+    ] = field(  # type:ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            is_allow_set=True,
+            items=[schema.DiskControllerType.SCSI, schema.DiskControllerType.NVME],
+        ),
+        metadata=field_metadata(
+            decoder=partial(
+                search_space.decode_nullable_set_space,
+                base_type=schema.DiskControllerType,
+                is_allow_set=True,
+                default_values=[
+                    schema.DiskControllerType.SCSI,
+                    schema.DiskControllerType.NVME,
+                ],
+            )
+        ),
+    )
+    hyperv_generation: Optional[
+        Union[search_space.SetSpace[int], int]
+    ] = field(  # type:ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            is_allow_set=True,
+            items=[1, 2],
+        ),
+        metadata=field_metadata(
+            decoder=partial(search_space.decode_set_space_by_type, base_type=int)
+        ),
+    )
     network_data_path: Optional[
         Union[search_space.SetSpace[schema.NetworkDataPath], schema.NetworkDataPath]
     ] = field(  # type: ignore
         default_factory=partial(
             search_space.SetSpace,
+            is_allow_set=True,
             items=[
                 schema.NetworkDataPath.Synthetic,
                 schema.NetworkDataPath.Sriov,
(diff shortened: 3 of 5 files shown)


lisa/schema.py
```diff
@@ -870,6 +870,10 @@ class NodeSpace(search_space.RequirementMixin, TypedSchema, ExtendableSchemaMixi
         ),
     )
 
+    # Platform may add image features to
+    # keep track of platform requirements.
+    _image_features: Optional[Dict[str, Any]] = None
```
squirrelsc: The image features should be included in `ImageSchema`, and extended in `AzureImageSchema`. It's not a part of node space like this.

lisa/feature.py
```diff
@@ -87,6 +87,16 @@ def create_setting(
         """
         return None
 
+    @classmethod
+    def create_image_requirement(
```
squirrelsc, line 91: The input should be `ImageSchema`. The method should be called like `create_settings_from_image`. The Azure data structure should be parsed to ImageSchema in the extended AzureImageSchema methods.

kamalca: @squirrelsc 
* Is there a reason `AzureImageSchema` does not inherit from Image schema?
* Are you fine with `ImageSchema` containing a free format dict field?

I don't think there is a way to completely get around needing a dict. It is similar to the `raw_capabilities` of `create_setting` where azure returns a bunch of key value pairs and the feature needs to implement the logic of how to interpret the raw capabilities.

squirrelsc: 1. `AzureImageSchema` should inherit from `ImageSchema`
2. It doesn't need to have a free format dict in `constructors` of `ImageSchema` and its subsclasses.

There should be a static method like `create_from_raw_info`. For example, the class `AzureMarketplaceImageSchema` inherits from `AzureImageSchema`, you can parse the raw dict in the `create_from_raw_info` to generate an `AzureMarketplaceImageSchema` object.

kamalca: But a lot of the parsing needs to happen on the platform level because you need the compute client and other platform level information like which features are supported.

Also, how can I store all the feature data in the AzureMarketplaceImageSchema? Do I add a FeatureSpace to AzureImageSchema? It needs to be capable of storing information for every single Feature.

squirrelsc: The flow is like,

1. The platform fetches data from Azure SDK for Image.
2. Platform calls the corresponding `ImageSchema` subclass and pass in the raw image information. It supposes have all information, which needs to create features.
3. With above step, a structured `ImageSchema` is created. Call `create_settings_from_image` of each possible image feature to create settings. This logic is shared for all kinds of Azure image (market place, gallery image, community image, vhd).
4. Merge the image features with node capabilities, so it's the real capability, and ready to merge with requirements from test case and platform.

kamalca: The current implementation does not generate image capabilities, it generates image requirements which are merged with platform requirements, not platform capabilities.

You previously recommended that I modify the existing `_add_image_features` for my implementation, but this function is not positioned to perform the flow you are now suggesting. It sounds like you are now suggesting a flow which allows LISA to select an image based on other requirements; this is a much bigger ask then what I have currently implemented.

I think we should have a meeting to discuss how to move forward with this.

squirrelsc: We can discuss offline. Let me take some back, we call it image requirements, instead of image capability. It's not to make image selectable. We follow the current process to merge image features with test case/runbook requirements firstly, then find out a capability.

kamalca:
@LiliDeng Do you have a Community Gallery Image I can test my changes on?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -193,9 +206,60 @@ class AzureVmPurchasePlanSchema:
     publisher: str
 
 
+class ArchitectureType(str, Enum):
```
squirrelsc: I remember there are some similar declared, please reuse them. If not exist, move it to schema.py for reusing.

kamalca: It is not created, it is moved from azure/features.py to prevent import loop. Architecture is only implemented in Azure Platform so it would be a bit weird to move it to schema.py.

LiliDeng:
> @LiliDeng Do you have a Community Gallery Image I can test my changes on?

community_gallery_image: "southcentralus/FreeBSDCGTest-d8a43fa5-745a-4910-9f71-0c9da2ac22bf/FreeBSD-14.0-RC3-amd64-ufs-gen1-testing/latest"

please specify the location southcentralus

lisa/sut_orchestrator/azure/common.py
```diff
@@ -576,7 +973,7 @@ def get_image_name(self) -> str:
             assert isinstance(
                 self.shared_gallery_raw, dict
             ), f"actual type: {type(self.shared_gallery_raw)}"
-            if self.shared_gallery.resource_group_name:
+            if self.shared_gallery_raw.get("resource_group_name", ""):
```
squirrelsc: Why get it from raw?

lisa/sut_orchestrator/azure/common.py
squirrelsc: It's a general setting, and will be used by all platforms. So it makes sense to be in schema.py.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2294,6 +2295,13 @@ def create_setting(
             security_profile=search_space.SetSpace(True, capabilities)
         )
 
+    @classmethod
+    def create_image_requirement(
+        cls, image: schema.ImageSchema
+    ) -> Optional[schema.FeatureSettings]:
+        assert isinstance(image, AzureImageSchema)
```
squirrelsc: Follow other type assertion to print `f"actual: {type(image)}"`.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -243,13 +420,121 @@ def __hash__(self) -> int:
...
+            "ConfidentialVmSupported",
+        ):
+            capabilities.append(SecurityProfileType.SecureBoot)
+            capabilities.append(SecurityProfileType.CVM)
+            capabilities.append(SecurityProfileType.Stateless)
+        self.security_profile = security_profile = search_space.SetSpace(
+            True, capabilities
+        )
+
+    def _parse_version(self, platform: "AzurePlatform") -> "SharedImageGallerySchema":
```
squirrelsc: 1. Remove the underscore, because it's called out of this class.
2. Call it `resolve_version` better, because there is logic to change the version from latest to specific versions, not just parse.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -243,13 +429,121 @@ def __hash__(self) -> int:
...
+                    time = gallery_image.publishing_profile.published_date
+                    assert image, "'image' must not be 'None'"
+                    assert image.name, "'image.name' must not be 'None'"
+                    self.image_version = image.name
+                elif gallery_image.publishing_profile.published_date > time:
+                    time = gallery_image.publishing_profile.published_date
+                    assert image, "'image' must not be 'None'"
+                    assert image.name, "'image.name' must not be 'None'"
+                    self.image_version = image.name
+        return copy.copy(self)
```
squirrelsc: Because the `self` is modified already, so it doesn't need to return a new object. An inplace update is enough.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2994,13 +2908,13 @@ def _get_os_disk_size(self, azure_runbook: AzureNodeSchema) -> int:
...
+                self
             )
             return self._get_sig_os_disk_size(azure_runbook.shared_gallery)
         elif azure_runbook.community_gallery_image:
-            azure_runbook.community_gallery_image = self._parse_community_gallery_image(
-                azure_runbook.community_gallery_image
+            azure_runbook.community_gallery_image = (
+                azure_runbook.community_gallery_image._parse_version(self)
             )
             return self._get_cgi_os_disk_size(azure_runbook.community_gallery_image)
```
squirrelsc, line 2859: Maybe next step, the methods should be moved to ImageSchema.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2994,13 +2908,13 @@ def _get_os_disk_size(self, azure_runbook: AzureNodeSchema) -> int:
                 # if no image info, use default size 30
                 return 30
         elif azure_runbook.shared_gallery:
-            azure_runbook.shared_gallery = self._parse_shared_gallery_image(
-                azure_runbook.shared_gallery
+            azure_runbook.shared_gallery = azure_runbook.shared_gallery._parse_version(
```
squirrelsc: Can it be called automatically, so it doesn't need to be called explicitly here.

kamalca: @squirrelsc
I could move all the 'os disk size' logic to be in the AzureImageSchema's.
This will add to the size/complexity of the PR, but I can do that if you think it is worth it.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -193,25 +206,141 @@ class AzureVmPurchasePlanSchema:
...
+                    SecurityProfileType.Standard,
+                    SecurityProfileType.SecureBoot,
+                    SecurityProfileType.CVM,
+                    SecurityProfileType.Stateless,
+                ],
+            )
+        ),
+    )
+
+    def load_features(self, platform: "AzurePlatform") -> None:
```
squirrelsc: `load_from_platform`. The "feature" is used as a key concept in LISA. Avoid using it in other ways.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -193,25 +206,141 @@ class AzureVmPurchasePlanSchema:
...
+        """Get raw image tags from Azure platform."""
+        raise NotImplementedError()
+
+    def _parse_info(self, raw_features: Dict[str, Any]) -> None:
+        """Parse raw image tags to AzureImageSchema"""
+        raise NotImplementedError()
+
+
+def _get_image_features(image: Any) -> Dict[str, Any]:
+    image_features: Dict[str, Any] = {}
```
squirrelsc: Add comments to explain this is for backward compatibility.

lisa/sut_orchestrator/azure/platform_.py
squirrelsc: You can do it in a commit after the main change.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -193,25 +206,141 @@ class AzureVmPurchasePlanSchema:
...
+    def _get_info(self, platform: "AzurePlatform") -> Optional[Dict[str, Any]]:
+        """Get raw image tags from Azure platform."""
+        raise NotImplementedError()
+
+    def _parse_info(self, raw_features: Dict[str, Any]) -> None:
+        """Parse raw image tags to AzureImageSchema"""
+        raise NotImplementedError()
+
+
+def _get_image_features(image: Any) -> Dict[str, Any]:
```
squirrelsc: it can be called in `load_features` once.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -225,6 +354,63 @@ class AzureVmMarketplaceSchema(AzureImageSchema):
     def __hash__(self) -> int:
         return hash(f"{self.publisher}/{self.offer}/{self.sku}/{self.version}")
 
+    def _get_info(self, platform: "AzurePlatform") -> Optional[Dict[str, Any]]:
+        for location in platform._find_marketplace_image_location():
```
squirrelsc: If the methods `_find_marketplace_image_location` and `_get_image_info` are not called in other places in platform, they should be moved here. If there are many methods involved, we can keep as is, but remove underscore of them.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -243,13 +429,121 @@ def __hash__(self) -> int:
...
+        return copy.copy(self)
+
 
 @dataclass_json()
 @dataclass
 class VhdSchema(AzureImageSchema):
     vhd_path: str = ""
     vmgs_path: Optional[str] = None
 
+    def _get_info(self, platform: "AzurePlatform") -> Optional[Dict[str, Any]]:
```
squirrelsc: Does it miss merge information from runbook?

lisa/sut_orchestrator/azure/common.py
kamalca: What do you mean by this comment?

kamalca: I created this function originally to reduce duplicated code. I will just implement it in each function if that is more clear.

lisa/sut_orchestrator/azure/common.py
kamalca: The problem I ran into with modifying self is that it doesn't call the setter method in AzureNodeSchema, so the object and the raw info become unsynced. It caused problems in other areas.

lisa/sut_orchestrator/azure/common.py
kamalca: Previously VHD only copied hyperv generation from the runbook, but I implemented that logic in platform in a slightly different way. Now the hyperv generation provided in the runbook is the default value and it gets overridden if the image is tagged.

lisa/sut_orchestrator/azure/common.py
kamalca: When raw and object get out of sync, checking object is unsafe. In the command below, the join is getting keys shared_gallery_raw. The `if` check should match what is being used below.

kamalca: I could change `shared_gallery_raw.get` to `shared__gallery.__dict__.get` if you want both of them to pull from the object and not the raw.

lisa/sut_orchestrator/azure/common.py
squirrelsc: I mean add code comments to explain the purpose of this function. I thought it's for consuming existing vhd generation flags in platform level.

lisa/sut_orchestrator/azure/common.py
squirrelsc: It looks a copy doesn't solve the problem, which unsync between object the raw info. It should be solved at the caller level.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -225,6 +354,63 @@ class AzureVmMarketplaceSchema(AzureImageSchema):
...
+        ):
+            self.disk_controller_type = schema.DiskControllerType.NVME
+
+        # Hyper-V Generation
+        try:
+            gen = raw_features.get("hyper_v_generation")
+            if gen:
+                self.hyperv_generation = int(gen.strip("V"))
+        except (TypeError, ValueError, AttributeError):
+            pass
```
squirrelsc: Does it need a debug level log?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -225,6 +354,63 @@ class AzureVmMarketplaceSchema(AzureImageSchema):
...
+        if security_profile == "TrustedLaunchSupported":
+            capabilities.append(SecurityProfileType.SecureBoot)
+        elif security_profile in (
+            "TrustedLaunchAndConfidentialVmSupported",
+            "ConfidentialVmSupported",
+        ):
+            capabilities.append(SecurityProfileType.SecureBoot)
+            capabilities.append(SecurityProfileType.CVM)
+            capabilities.append(SecurityProfileType.Stateless)
+        self.security_profile = security_profile = search_space.SetSpace(
```
squirrelsc: Should it remove the `security_profile =`?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -243,13 +429,121 @@ def __hash__(self) -> int:
...
 
+    def _get_info(self, platform: "AzurePlatform") -> Optional[Dict[str, Any]]:
+        self.resolve_version(platform)
+        sig_info = platform._get_sig(self)
+        return _get_image_tags(sig_info)
+
+    def _parse_info(self, raw_features: Dict[str, Any]) -> None:
+        # CPU Architecture
+        arch = raw_features.get("architecture")
+        if arch == "arm64":
```
squirrelsc: Some logic are similar among marketplace, sig and community image. Create a shared method in AzureImageSchema to reduce duplicate code.

lisa/sut_orchestrator/azure/common.py
squirrelsc: I cannot figure out what happened from the code reviewing. Can you share why the raw and object unsync? We should fix the unsync. The raw info is used in only one scenario, but I forgot when it's used.

lisa/sut_orchestrator/azure/common.py
kamalca: Should I pass in logger from platform? (I'm not super familiar with how loggers are handled)

lisa/sut_orchestrator/azure/common.py
kamalca: More logic could eventually be moved to the AzureImageSchemas but I'm a bit afraid to change any of the behavior in platform because it is impossible for me to test every scenario for bugs.

lisa/sut_orchestrator/azure/common.py
kamalca: I see what you're saying, but I want `_get_info` to return a single type. If I move `_get_image_features` (now called `_get_image_tags` out of `_get_info` then that function would return a MarketplaceImage / GalleryImage / CommunityGalleryImage and it would return None for VHD.

lisa/sut_orchestrator/azure/common.py
kamalca: I fixed the syncing, but how do you want me to move forward with this logic. My original change was because the check `if self.shared_gallery.resource_group_name` does not match the variable being used `[self.shared_gallery_raw.get(k, "") for k in SIG_IMAGE_KEYS]`

I still think both should use the same, either `shared_gallery` or `shared_gallery_raw`

squirrelsc: Ideally, they should be consistent, so `self.shared_gallery.resource_group_name` is better, because it's a var name, not a string.
Let me know, if in some cases it's inconsistent. It needs to understand why not consistent and fix them.

lisa/sut_orchestrator/azure/common.py
squirrelsc: Yes, you can use the logger from platform. The different logs decide the prefix of log lines, and may be put into different log files.

lisa/sut_orchestrator/azure/common.py
squirrelsc: You can override `load_from_platform` for vhd. So, the logic applies to other schemas only.

kamalca: That doesn't address the need for `_get_info` to have a common return type. This function takes the Marketplace / SIG / CGI and outputs a common dictionary. The reason I wrote a separate function is to take advantage of the common structure of the three classes.

squirrelsc: I may not fully understand it. If the `load_from_platform` is overridden in vhd. It doesn't need to take care `_get_info` and `_parse_info` in vhd schema.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -195,24 +210,209 @@ class AzureVmPurchasePlanSchema:
...
+                base_type=schema.DiskControllerType,
+                is_allow_set=True,
+                default_values=[
+                    schema.DiskControllerType.SCSI,
+                    schema.DiskControllerType.NVME,
+                ],
+            )
+        ),
+    )
+    hyperv_generation: Optional[
```
squirrelsc, line 255: Other places are using name `hyper_v_generation`, but here is `hyperv_generation`. Should they are the same?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -243,13 +450,81 @@ def __hash__(self) -> int:
...
+                    self.image_version = image.name
+
 
 @dataclass_json()
 @dataclass
 class VhdSchema(AzureImageSchema):
     vhd_path: str = ""
     vmgs_path: Optional[str] = None
 
+    def _get_info(self, platform: "AzurePlatform") -> Dict[str, Any]:
```
squirrelsc: It doesn't need to override the two methods in vhd schema, just overwrite `load_from_platform` like below.

```python
def load_from_platform(self, platform: "AzurePlatform") -> None:
    pass
```

lisa/sut_orchestrator/azure/common.py
```diff
@@ -243,13 +450,81 @@ def __hash__(self) -> int:
...
+        compute_client = get_compute_client(platform)
+        sig = compute_client.gallery_images.get(
+            resource_group_name=self.resource_group_name,
+            gallery_name=self.image_gallery,
+            gallery_image_name=self.image_definition,
+        )
+        assert isinstance(sig, GalleryImage), f"actual: {type(sig)}"
+        return sig
+
+    def _get_info(self, platform: "AzurePlatform") -> Dict[str, Any]:
```
squirrelsc: move private methods below public methods.

lisa/sut_orchestrator/azure/common.py
kamalca: Images are tagged with `hyper_v_generation` naming convention in LISA has been `hyperv_generation`

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@kamalca could you please run a test with gallery arm64 image?

kamalca:
> @kamalca could you please run a test with gallery arm64 image?

Shared Image Gallery or Community Gallery Image?

LiliDeng:
> > @kamalca could you please run a test with gallery arm64 image?
> 
> Shared Image Gallery or Community Gallery Image?

You need cover all scenarios, I hit one issue when run a test with Shared Image Gallery.

kamalca:
> > > @kamalca could you please run a test with gallery arm64 image?
> > 
> > 
> > Shared Image Gallery or Community Gallery Image?
> 
> You need cover all scenarios, I hit one issue when run a test with Shared Image Gallery.

I fixed a typo. Thank you for helping test the changes.

kamalca:
@LiliDeng Any updates on this getting merged? I have to do more work to continually rebase, merge conflicts, and retest changes the longer this goes unmerged.

LiliDeng (approved):

PR #3257, opened 2024-04-09
notifiler: enable json notifier

LiliDeng:
current dumped json file format is like this, it is an array of all perf messages from one triggered run.

image

### DIFF
--- a/lisa/notifiers/perfdump.py
+++ b/lisa/notifiers/perfdump.py
@@ -1,0 +1,67 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import json
+import os
+from dataclasses import dataclass
+from datetime import datetime
+from decimal import Decimal
+from enum import Enum
+from typing import Any, List, Type, cast
+
+from dataclasses_json import dataclass_json
+
+from lisa import constants, messages, notifier, schema
+
+
+@dataclass_json()
+@dataclass
+class PerfDumpSchema(schema.Notifier):
+    path: str = "lisa.json"
+
+
+class PerfDump(notifier.Notifier):
+    """
+    The Json notifier is used to output the perf test results in JSON format.
+    """
+
+    _first_message_written: bool = False
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "perfdump"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return PerfDumpSchema
+
+    def _received_message(self, message: messages.MessageBase) -> None:
+        if isinstance(message, messages.PerfMessage):
+            message_dict = {}
+            for key, value in message.__dict__.items():
+                if isinstance(value, Enum):
+                    value = value.value
+                elif isinstance(value, datetime) or isinstance(value, Decimal):
+                    value = str(value)
+                message_dict[key] = value
+            if self._first_message_written:
+                self._report_file.write(",\n")
+            else:
+                self._first_message_written = True
+            json.dump(message_dict, self._report_file, indent=4, ensure_ascii=False)
+
+    def _subscribed_message_type(self) -> List[Type[messages.MessageBase]]:
+        return [messages.PerfMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        runbook = cast(PerfDumpSchema, self.runbook)
+        self._report_path = constants.RUN_LOCAL_LOG_PATH / runbook.path
+        self._report_file = open(self._report_path, "a", encoding="utf-8")
+        if os.path.getsize(self._report_path) == 0:
+            self._report_file.write("[")
+
+    def finalize(self) -> None:
+        try:
+            self._report_file.write("]")
+        finally:
+            self._report_file.close()
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -15,6 +15,7 @@ import lisa.notifiers.env_stats  # noqa: F401
 import lisa.notifiers.file  # noqa: F401
 import lisa.notifiers.html  # noqa: F401
 import lisa.notifiers.junit  # noqa: F401
+import lisa.notifiers.perfdump  # noqa: F401
 import lisa.notifiers.text_result  # noqa: F401
 import lisa.runners.lisa_runner  # noqa: F401
 import lisa.sut_orchestrator.ready  # noqa: F401


squirrelsc:
> current dumped json file format is like this, the messages are the array of message dict image

It should dump messages by a list, so it's easy to export to other formats. For perf messages, it doesn't need to know which test case produces them.

lisa/notifiers/json.py
```diff
@@ -0,0 +1,59 @@
...
+from lisa import constants, messages, notifier, schema
+
+
+@dataclass_json()
+@dataclass
+class JsonSchema(schema.Notifier):
+    path: str = "lisa.json"
+
+
+class Json(notifier.Notifier):
```
squirrelsc: It just focuses on perf messages, and dump a better format to consume by other tools. It should be like List[Dict[str, Any]] format. All messages are dumped in the same array.

lisa/notifiers/json.py
```diff
@@ -0,0 +1,59 @@
...
+    def _subscribed_message_type(self) -> List[Type[messages.MessageBase]]:
+        return [messages.PerfMessage]
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        runbook = cast(JsonSchema, self.runbook)
+        self._report_path = constants.RUN_LOCAL_LOG_PATH / runbook.path
+        self._grouped_results: Dict[Any, Any] = {}
+
+    def finalize(self) -> None:
+        with open(self._report_path, "w") as json_file:
```
squirrelsc: It's better to dump messages continually, instead of dump all at the end. It's faster and real time.

LiliDeng:
> > current dumped json file format is like this, the messages are the array of message dict image
> 
> It should dump messages by a list, so it's easy to export to other formats. For perf messages, it doesn't need to know which test case produces them.

you mean we don't need to distinguish the messages for which run? if these messages from two run we can add them into one list.

lisa/notifiers/json.py
LiliDeng: updated

lisa/notifiers/json.py
LiliDeng: updated

lisa/notifiers/json.py
```diff
@@ -0,0 +1,65 @@
...
+            message_dict = {}
+            for key, value in message.__dict__.items():
+                if isinstance(value, Enum):
+                    value = value.value
+                elif isinstance(value, datetime) or isinstance(value, Decimal):
+                    value = str(value)
+                message_dict[key] = value
+
+            if not os.path.exists(self._report_path):
+                with open(self._report_path, "w", encoding="utf-8") as json_file:
```
squirrelsc: Don't open a file multiple times, it's low efficiency. You can open it when initializing, and close it in finalizing.

lisa/notifiers/json.py
```diff
@@ -0,0 +1,65 @@
...
+from lisa import constants, messages, notifier, schema
+
+
+@dataclass_json()
+@dataclass
+class JsonSchema(schema.Notifier):
+    path: str = "lisa.json"
+
+
+class Json(notifier.Notifier):
```
squirrelsc: This class is specified to perf message, so the name shouldn't be called like `Json`, use `PerfDump`.

lisa/notifiers/json.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import json
```
squirrelsc: Does it include platform information like image and vm size?

LiliDeng: it contains vmsize, kernel version, it doesn't contain image info.

```
{
    "type": "Performance",
    "time": "2024-04-10 07:25:38.061040",
    "elapsed": 0,
    "tool": "iperf3",
    "test_case_name": "perf_tcp_iperf_sriov",
    "platform": "azure",
    "location": "westus2",
    "host_version": "20348-10.0-1-0.1606",
    "guest_os_type": "Linux",
    "distro_version": "Ubuntu 20.04.6 LTS",
    "vmsize": "Standard_DS2_v2",
    "kernel_version": "5.15.0-1060-azure",
    "lis_version": "",
    "ip_version": "IPv4",
    "protocol_type": "TCP",
    "data_path": "Sriov",
    "test_date": "2024-04-10 07:17:08.413022",
    "role": "",
    "test_result_id": "lisa_0_0",
    "connections_num": 1,
    "throughput_in_gbps": "0",
    "latency_us": "0",
    "buffer_size": "0",
    "tx_packets": "0",
    "rx_packets": "0",
    "pkts_interrupts": "0",
    "number_of_receivers": 1,
    "number_of_senders": 1,
    "sender_cycles_per_byte": "0",
    "connections_created_time": 0,
    "retrans_segments": 0,
    "receiver_cycles_rer_byte": "0",
    "buffer_size_bytes": "32",
    "tx_throughput_in_gbps": 0.09070724171712838,
    "rx_throughput_in_gbps": 0.09070724171712838,
    "retransmitted_segments": 0,
    "congestion_windowsize_kb": "69.4904296875"
}
```

lisa/notifiers/json.py
LiliDeng: updated

lisa/notifiers/json.py
LiliDeng: updated

lisa/notifiers/perfdump.py
```diff
@@ -0,0 +1,63 @@
...
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        runbook = cast(PerfDumpSchema, self.runbook)
+        self._report_path = constants.RUN_LOCAL_LOG_PATH / runbook.path
+        self._report_file = open(self._report_path, "a", encoding="utf-8")
+        if os.path.getsize(self._report_path) == 0:
+            self._report_file.write("[")
+
+    def finalize(self) -> None:
+        try:
+            self._report_file.write("[")
```
squirrelsc: should it write `]`, not `[`?

lisa/notifiers/perfdump.py
```diff
@@ -0,0 +1,63 @@
...
+            message_dict = {}
+            for key, value in message.__dict__.items():
+                if isinstance(value, Enum):
+                    value = value.value
+                elif isinstance(value, datetime) or isinstance(value, Decimal):
+                    value = str(value)
+                message_dict[key] = value
+
+            json.dump(message_dict, self._report_file, indent=4, ensure_ascii=False)
+            self._report_file.write(",")
```
squirrelsc: The tailing `,` may not be loaded by some libraries. It's better to append `,` before each new line.

LiliDeng: updated

squirrelsc (approved):

PR #3262, opened 2024-04-11
feat: Implement mshv_install test to verify Linux Dom0 boot against fresh Hyper-V bins

Camelron:
Add Linux Dom0 boot smoketest to verify a set of Hyper-V binaries against Linux, the linux-mshv driver.

### DIFF
--- a/microsoft/testsuites/mshv/mshv_install.py
+++ b/microsoft/testsuites/mshv/mshv_install.py
@@ -5,6 +5,8 @@ import re
 from pathlib import Path
 from typing import Any, Dict
 
+import time
+
 from assertpy import assert_that
 
 from lisa import (
@@ -13,15 +15,11 @@ from lisa import (
     Node,
     TestCaseMetadata,
     TestSuite,
-    TestSuiteMetadata,
-    simple_requirement,
+    TestSuiteMetadata
 )
 # from lisa.features import SerialConsole
 from lisa.testsuite import TestResult
-from lisa.tools import Dmesg, Ls, Service, Reboot
-from lisa.util import SkippedException, TcpConnectionException, constants
-from lisa.util.shell import wait_tcp_port_ready
-
+from lisa.tools import Cp, Dmesg, Ls, Reboot
 
 @TestSuiteMetadata(
     area="mshv",
@@ -60,6 +58,18 @@ class MshvHostInstallSuite(TestSuite):
         Path("/home/cloud")
     )
 
+    _init_path_dst_hvix = (
+        Path("/home/cloud") / f"hvix64.exe"
+    )
+
+    _init_path_dst_kdstub = (
+        Path("/home/cloud") / f"kdstub.dll"
+    )
+
+    _init_path_dst_lxhvloader = (
+        Path("/home/cloud") / f"lxhvloader.dll"
+    )
+
     @TestCaseMetadata(
         description="""
         This test case will
@@ -69,7 +79,6 @@ class MshvHostInstallSuite(TestSuite):
         The test expects the MSHV binaries to be installed to be placed under lisa/microsoft/testsuites/mshv/test_data
         before lisa is executed.
         """,
-        priority=0,
         timeout=60,  # 60 seconds
     )
     def verify_mshv_install_succeeds(
@@ -80,50 +89,31 @@ class MshvHostInstallSuite(TestSuite):
         result: TestResult,
     ) -> None:
         # Copy Hvix64.exe, kdstub.dll, lxhvloader.dll into test machine
-        hvix_path = self._init_path_dst / f"hvix64.exe"
-        node.shell.copy(self._test_hvix_file_path, hvix_path)
-        res = node.shell.spawn(
-            command=['sudo', 'cp', hvix_path.as_posix(), self._test_hvix_file_path_dst.as_posix()],
-            allow_error=False,
-        ).wait_for_result()
-        log.info(f"sudo cp result {res.return_code}")
+        node.shell.copy(self._test_hvix_file_path, self._init_path_dst_hvix)
+        test_sha256_cmd = "sudo sha256sum %s" % self._test_hvix_file_path_dst.as_posix()
+        res = node.execute(
+            test_sha256_cmd,
+            shell=True,
+            sudo=True
+        )
+        time.sleep(5)
+        node.tools[Cp].copy(self._init_path_dst_hvix.as_posix(), self._test_hvix_file_path_dst.as_posix(), sudo=True)
+        res = node.execute(
+            test_sha256_cmd,
+            shell=True,
+            sudo=True
+        )
 
-        assert res.return_code == 0
-
-        # kdstub_path = self._init_path_dst / f"kdstub.dll"
-        # node.shell.copy(self._test_kdstub_file_path, kdstub_path)
-        # res = node.shell.spawn(
-        #     command=['sudo', 'cp', kdstub_path.as_posix(), self._test_hvix_file_path_dst.as_posix()],
-        #     allow_error=False,
-        # ).wait_for_result()
-        # assert res.return_code == 0
+        # node.shell.copy(self._test_kdstub_file_path, self._init_path_dst_kdstub)
+        # node.tools[Cp].copy(self._init_path_dst_kdstub.as_posix(), self._test_hvix_file_path_dst.as_posix(), sudo=True)
         
-        lxhvloader_path = self._init_path_dst / f"lxhvloader.dll"
-        node.shell.copy(self._test_lxhvloader_file_path, lxhvloader_path)
-        res = node.shell.spawn(
-            command=['sudo', 'cp', lxhvloader_path.as_posix(), self._test_lxhvloader_file_path_dst.as_posix()],
-            allow_error=False,
-        ).wait_for_result()
-        assert res.return_code == 0
+        node.shell.copy(self._test_lxhvloader_file_path, self._init_path_dst_lxhvloader)
+        node.tools[Cp].copy(self._init_path_dst_lxhvloader.as_posix(), self._test_lxhvloader_file_path_dst.as_posix(), sudo=True)
+
 
         reboot_tool = node.tools[Reboot]
         reboot_tool.reboot_and_check_panic(log_path)
 
-        is_ready, tcp_error_code = wait_tcp_port_ready(
-            node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_ADDRESS],
-            node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_PORT],
-            log=log,
-        )
-
-        if is_ready:
-            # 2. check that mshv comes up
-            mshvUp = node.tools[Ls].path_exists("/dev/mshv", sudo=True)
-            assert mshvUp
-        else:
-            raise TcpConnectionException(
-                node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_ADDRESS],
-                node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_PORT],
-                tcp_error_code,
-                "no panic found in serial log",
-            )
-        return
+        # 2. check that mshv comes up
+        mshvUp = node.tools[Ls].path_exists("/dev/mshv", sudo=True)
+        assert mshvUp
--- a/microsoft/testsuites/mshv/mshv_install.py
+++ b/microsoft/testsuites/mshv/mshv_install.py
@@ -29,45 +29,30 @@ from lisa.tools import Cp, Dmesg, Ls, Reboot
     """,
 )
 class MshvHostInstallSuite(TestSuite):
+    CONFIG_BINPATH = "mshv_binpath"
 
-    _test_hvix_file_path = (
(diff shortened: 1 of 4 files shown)


microsoft/testsuites/mshv/test_data/README.md
```diff
@@ -0,0 +1,7 @@
+To execute this testsuite, one must provide the set of MSHV binaries to test.
```
squirrelsc: Does it apply to all test cases in this mshv folder, or only added new test cases? Either way, the description should be in test suite metadata. So the generated test document contains it.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+            mshvUp = node.tools[Ls].path_exists("/dev/mshv", sudo=True)
+            assert mshvUp
+        else:
+            raise TcpConnectionException(
+                node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_ADDRESS],
+                node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_PORT],
+                tcp_error_code,
+                "no panic found in serial log",
+            )
+        return
```
squirrelsc: It doesn't need return at the end.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+    def verify_mshv_install_succeeds(
+        self,
+        log: Logger,
+        node: Node,
+        log_path: Path,
+        result: TestResult,
+    ) -> None:
+        # Copy Hvix64.exe, kdstub.dll, lxhvloader.dll into test machine
+        hvix_path = self._init_path_dst / f"hvix64.exe"
+        node.shell.copy(self._test_hvix_file_path, hvix_path)
```
squirrelsc: You can change the path from a test visible variable. if the variable is not provided, the test case should be skipped, instead of failed.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+        res = node.shell.spawn(
+            command=['sudo', 'cp', hvix_path.as_posix(), self._test_hvix_file_path_dst.as_posix()],
+            allow_error=False,
+        ).wait_for_result()
+        log.info(f"sudo cp result {res.return_code}")
+
+        assert res.return_code == 0
+
+        # kdstub_path = self._init_path_dst / f"kdstub.dll"
```
squirrelsc: Remove the commented parts.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+        # Copy Hvix64.exe, kdstub.dll, lxhvloader.dll into test machine
+        hvix_path = self._init_path_dst / f"hvix64.exe"
+        node.shell.copy(self._test_hvix_file_path, hvix_path)
+        res = node.shell.spawn(
+            command=['sudo', 'cp', hvix_path.as_posix(), self._test_hvix_file_path_dst.as_posix()],
+            allow_error=False,
+        ).wait_for_result()
+        log.info(f"sudo cp result {res.return_code}")
+
+        assert res.return_code == 0
```
squirrelsc: use `res.assert_exit_code`

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+        self,
+        log: Logger,
+        node: Node,
+        log_path: Path,
+        result: TestResult,
+    ) -> None:
+        # Copy Hvix64.exe, kdstub.dll, lxhvloader.dll into test machine
+        hvix_path = self._init_path_dst / f"hvix64.exe"
+        node.shell.copy(self._test_hvix_file_path, hvix_path)
+        res = node.shell.spawn(
```
squirrelsc: use `node.execute`

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+            command=['sudo', 'cp', hvix_path.as_posix(), self._test_hvix_file_path_dst.as_posix()],
+            allow_error=False,
+        ).wait_for_result()
+        log.info(f"sudo cp result {res.return_code}")
```
squirrelsc: The output is in debug level already.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+        log: Logger,
+        node: Node,
+        log_path: Path,
+        result: TestResult,
+    ) -> None:
+        # Copy Hvix64.exe, kdstub.dll, lxhvloader.dll into test machine
+        hvix_path = self._init_path_dst / f"hvix64.exe"
+        node.shell.copy(self._test_hvix_file_path, hvix_path)
+        res = node.shell.spawn(
+            command=['sudo', 'cp', hvix_path.as_posix(), self._test_hvix_file_path_dst.as_posix()],
```
squirrelsc: use `cp` tool for copying.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+        res = node.shell.spawn(
+            command=['sudo', 'cp', lxhvloader_path.as_posix(), self._test_lxhvloader_file_path_dst.as_posix()],
+            allow_error=False,
+        ).wait_for_result()
+        assert res.return_code == 0
+
+        reboot_tool = node.tools[Reboot]
+        reboot_tool.reboot_and_check_panic(log_path)
+
+        is_ready, tcp_error_code = wait_tcp_port_ready(
```
squirrelsc: wait for ready should be run by above reboot method. it doesn't need call it separately.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Update to new MSHV components over old ones in a pre-configured MSHV image
+        2. Reboot VM, check that mshv comes up
+
+        The test expects the MSHV binaries to be installed to be placed under lisa/microsoft/testsuites/mshv/test_data
+        before lisa is executed.
+        """,
+        priority=0,
```
squirrelsc: It won't be a P0 test case. Please use default P2.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,129 @@
...
+        Path("/home/cloud")
+    )
+
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. Update to new MSHV components over old ones in a pre-configured MSHV image
+        2. Reboot VM, check that mshv comes up
+
+        The test expects the MSHV binaries to be installed to be placed under lisa/microsoft/testsuites/mshv/test_data
```
squirrelsc: Pass in a path as a case variable.

microsoft/testsuites/mshv/mshv_install.py
Camelron: We specifically need sudo for this operation to install in the ESP partition.

Camelron: update: cp tool supports sudo. Addressed.

microsoft/testsuites/mshv/mshv_install.py
Camelron: Hey @squirrelsc, I'm trying to do just that, referencing this change: https://github.com/microsoft/lisa/commit/44c42cfbc038cc7321a1648e6709f09b580b05de#diff-45aed30957bd1da6be270e89e29ddc7a0acc7af49d6f014aef1fa5d7ddedef6bR82 

Is this the right approach? I'm trying to pass a variable "mshv_binpath" from my runbook to this testcase but it doesn't seem to be working.

squirrelsc: It's the right way to use the variable. But in the runbook, you need to define a case_visible variable like this:

```yaml
  - name: mshv_bindings_path
    value: ""
    is_case_visible: True
```

microsoft/testsuites/mshv/test_data/README.md
Camelron: Removed; replaced with mshv_binpath case variable.

microsoft/testsuites/mshv/mshv_install.py
Camelron: Worked, thanks!

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,97 @@
...
+        node.shell.copy(test_lxhvloader_file_path, self._init_path_init_lxhvloader)
+        node.tools[Cp].copy(self._init_path_init_lxhvloader,
+                            self._test_path_dst_lxhvloader, sudo=True)
+
+        reboot_tool = node.tools[Reboot]
+        reboot_tool.reboot_and_check_panic(log_path)
+
+        # 2. check that mshv comes up
+        mshv = node.tools[Ls].path_exists("/dev/mshv", sudo=True)
+        assert mshv
```
squirrelsc: Add message to the assertion. So once it's failed, the error message can indicate what happened, and how to deal with it.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,97 @@
...
+
+        test_hvix_file_path = Path(binpath) / "hvix64.exe"
+        test_kdstub_file_path = Path(binpath) / "kdstub.dll"
+        test_lxhvloader_file_path = Path(binpath) / "lxhvloader.dll"
+
+        log.info(f"binpath: {binpath}")
+
+        # Copy Hvix64.exe, kdstub.dll, lxhvloader.dll into test machine
+        node.shell.copy(test_hvix_file_path, self._test_path_init_hvix)
+        node.tools[Cp].copy(self._test_path_init_hvix,
```
squirrelsc: The `node.tools[Cp]` is used multiple times, so define a variable is clearer.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,97 @@
...
+        """,
+        timeout=120,
+    )
+    def verify_mshv_install_succeeds(
+        self,
+        log: Logger,
+        node: Node,
+        variables: Dict[str, Any],
+        log_path: Path,
+        result: TestResult,
```
squirrelsc: The result is not used, it should be removed.

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,97 @@
...
+        description="""
+        This test case will
+        1. Update to new MSHV components over old ones in a
+            pre-configured MSHV image
+        2. Reboot VM, check that mshv comes up
+
+        The test expects the directory containing MSHV binaries to be passed in
+        the mshv_binpath variable.
+        """,
+        timeout=120,
```
squirrelsc: The default timeout is 600 seconds, does it need to set it to 120 seconds?

microsoft/testsuites/mshv/mshv_install.py
```diff
@@ -0,0 +1,97 @@
...
+from lisa.testsuite import TestResult
+from lisa.tools import Cp, Ls, Reboot
+from lisa.util import SkippedException
+
+
+@TestSuiteMetadata(
+    area="mshv",
+    category="functional",
+    description="""
+    This test suite is to test VM working well after updating MSHV on VM
```
squirrelsc: Give full name of MSHV. The description will be in LISA document, it should be understandable for others.

microsoft/testsuites/mshv/mshv_install.py
Camelron: Storing "node.tools[Cp]" as CpTool or something? Does this really warrant the extra variable?

squirrelsc: Use like `cp = node.tools[Cp]`. So the var cp can be used multiple times.

microsoft/testsuites/mshv/mshv_root_stress_tests.py
```diff
@@ -120,7 +120,7 @@ def stress_mshv_vm_create(
                     test_status=TestStatus.FAILED,
                     test_message=repr(e),
                 )
-        self._save_dmesg_logs(node, log_path)
+        node.tools[CloudHypervisor].save_dmesg_logs(node, log_path)
```
squirrelsc, line 123: Why make this change? The dmesg tool should work with any platform.

squirrelsc:
@LiliDeng mostly LGTM

microsoft/testsuites/mshv/mshv_root_stress_tests.py
Camelron: I figured since I was sharing the routine originally in the root_stress test, I'd put it in the CloudHypervisor tool (to me, this is the "common" code?)

squirrelsc: Ok, good to keep it.

LiliDeng (approved):

PR #3264, opened 2024-04-16
Fix test verify_vmbus_devices_channels for CVM

kanchansenlaskar:
In CVM we don't have these three vmbus device names:
"Data Exchange",
"Synthetic mouse",
"Synthetic keyboard",

Hence added the condition to not expect them for cvm, and updated the test description.

This change is tested with:
redhat rhel 9_3 9.3.2024040517
canonical 0001-com-ubuntu-confidential-vm-jammy 22_04-lts-cvm 22.04.202403280

This does not impact BSD

### DIFF
--- a/microsoft/testsuites/core/lsvmbus.py
+++ b/microsoft/testsuites/core/lsvmbus.py
@@ -11,6 +11,12 @@ from lisa import (
     TestSuiteMetadata,
     simple_requirement,
 )
+from lisa.feature import Feature
+from lisa.features.security_profile import (
+    SecurityProfile,
+    SecurityProfileSettings,
+    SecurityProfileType,
+)
 from lisa.operating_system import BSD, Windows
 from lisa.sut_orchestrator import AZURE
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/lsvmbus.py
```diff
@@ -80,12 +96,13 @@ def verify_vmbus_devices_channels_bsd(self, node: Node) -> None:
             - Operating system shutdown
             - Time Synchronization
             - Heartbeat
-            - Data Exchange
-            - Synthetic mouse
-            - Synthetic keyboard
             - Synthetic network adapter
             - Synthetic SCSI Controller
             - Synthetic IDE Controller (gen1 only)
+        It expects addtional three vmbus device names for non-cvm:
```
squirrelsc: Add indents to be under "1. ".

kanchansenlaskar: resolved

squirrelsc:
@LiliDeng LGTM.

LiliDeng (approved):

PR #3265, opened 2024-04-16
M2VHD: Fix kernel params on ARM

kamalca:
ARM images wait for an early console which is only present in the cloud, not in lab environment. Remove this kernel parameter to allow for boot in lab.

### DIFF
--- a/lisa/transformers/disable_cloud_components.py
+++ b/lisa/transformers/disable_cloud_components.py
@@ -6,7 +6,7 @@ from dataclasses_json import dataclass_json
 
 from lisa import schema
 from lisa.sut_orchestrator.azure.tools import Waagent
-from lisa.tools import Ls, Sed
+from lisa.tools import Cat, Ls, Sed
 from lisa.transformers.deployment_transformer import (
     DeploymentTransformer,
     DeploymentTransformerSchema,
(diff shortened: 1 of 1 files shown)


lisa/transformers/disable_cloud_components.py
```diff
@@ -56,4 +56,16 @@ def _internal_run(self) -> Dict[str, Any]:
                 sudo=True,
             )
 
+        node.tools[Cat].run("/etc/default/grub", force_run=True)
```
squirrelsc: Please add comments above to explain the logic.

lisa/transformers/disable_cloud_components.py
```diff
@@ -45,15 +45,31 @@ def _internal_run(self) -> Dict[str, Any]:
...
 
+        # Remove macaddress binding
         if ls.path_exists("/etc/netplan/50-cloud-init.yaml"):
             sed.delete_lines(
                 "macaddress",
                 PurePosixPath("/etc/netplan/50-cloud-init.yaml"),
                 sudo=True,
             )
 
+        # Remove 'earlycon' from kernel paramteters
```
squirrelsc: typo on parameter.

squirrelsc (approved):

PR #3270, opened 2024-04-19
Downgrade the case priority for its instability

LiliDeng:


squirrelsc (dismissed):

microsoft/testsuites/vm_extensions/cvt.py
```diff
@@ -313,7 +313,7 @@ class CVTTest(TestSuite):
         this test validate the functionality of ASR driver by verifying
         integrity of a source disk with respect to a target disk
         """,
-        priority=3,
+        priority=5,
```
squirrelsc, line 318: Please update in the case descriptions too. So it can show in the test case details, why it's P5.

squirrelsc (approved):

PR #3272, opened 2024-04-22
add logic to assert truncation logic

feng-j678:
[x] add logic to assert truncation_logic in some images, this allowed lisa to pass the below example images. the truncation logic is recent lpe change we deployed to handle vm agent json deserialization error. 
image

running bvt on image: marketplace_image:redhat rhel 8-lvm 8.5.2022031405
image

### DIFF
--- a/microsoft/testsuites/vm_extensions/linux_patch_extension.py
+++ b/microsoft/testsuites/vm_extensions/linux_patch_extension.py
@@ -55,7 +55,10 @@ def _verify_vm_agent_running(node: Node, log: Logger) -> None:
 
 
 def _assert_status_file_result(status_file: Any) -> None:
-    if status_file["error"]["code"] == "PACKAGE_LIST_TRUNCATED":
+    if (
+        len(status_file["error"]["details"]) > 0
+        and status_file["error"]["details"][0]["code"] == "PACKAGE_LIST_TRUNCATED"
+    ):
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/linux_patch_extension.py
```diff
@@ -54,6 +54,21 @@ def _verify_vm_agent_running(node: Node, log: Logger) -> None:
     ).is_true()
 
 
+def _assert_status_file_result(status_file: Any) -> None:
+    if status_file["error"]["code"] == "PACKAGE_LIST_TRUNCATED":
```
LiliDeng: can we check the status code in a loop instead of waiting the operations till time out? since I found when it is in CompletedWithWarnings situation, we have to wait till time out, it costs 1.5 hours.

https://github.com/microsoft/lisa/blob/b17820284813f1d656a4b6814cffca004ffd160f/microsoft/testsuites/vm_extensions/linux_patch_extension.py#L144

feng-j678: I ran bvt on image marketplace_image:redhat rhel 8-lvm 8.5.2022031405, it didn't take up to 1.5hours, however if we do encounter a failure, then it will take up to 1.5hour.

LiliDeng (approved):

PR #3273, opened 2024-04-24
mariner: install perf tool from kernel-tools

LiliDeng:


lisa/tools/perf.py
```diff
@@ -31,12 +31,9 @@ def _install(self) -> bool:
...
             )
-            if (
-                isinstance(self.node.os, Redhat)
-                or isinstance(self.node.os, CentOs)
-                or isinstance(self.node.os, CBLMariner)
-                or isinstance(self.node.os, Suse)
-            ):
+            if isinstance(self.node.os, CBLMariner):
+                self.node.os.install_packages("kernel-tools")
+            elif isinstance(self.node.os, (CentOs, Redhat, Suse)):
```
squirrelsc: Does it work like this way? CentOs is from Redhat, so it doesn't need to be in the list.

LiliDeng: yes, worked, and I have removed CentOS.

squirrelsc (approved):

PR #3274, opened 2024-04-24
fix xfs fail, add upload result when timeout

paxue:
increase the core number to 8 for xfs verify_generic_standard_datadisk
add try logic for exception to upload the test result
reduce the wait timeout in wait by 20s to let wat timeout happen before the whole case timeout.

### DIFF
--- a/microsoft/testsuites/xfstests/xfstesting.py
+++ b/microsoft/testsuites/xfstests/xfstesting.py
@@ -125,7 +125,7 @@ class Xfstesting(TestSuite):
     # commit d0c7feaf8767 ("xfs: add agf freeblocks verify in xfs_agf_verify")
     # TODO: will figure out the detailed reason of every excluded case.
     excluded_tests = (
-        "generic/211 generic/430 generic/431 generic/434 xfs/438 xfs/490"
+        "generic/211 generic/430 generic/431 generic/434 generic/738 xfs/438 xfs/490"
         + " btrfs/007 btrfs/178 btrfs/244 btrfs/262"
         + " xfs/030 xfs/032 xfs/050 xfs/052 xfs/106 xfs/107 xfs/122 xfs/132 xfs/138"
         + " xfs/144 xfs/148 xfs/175 xfs/191-input-validation xfs/289 xfs/293 xfs/424"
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -534,7 +535,7 @@ def _execute_xfstests(
             mount_opts,
         )
         xfstests.set_excluded_tests(excluded_tests)
-        xfstests.run_test(test_type, log_path, result, data_disk, self.TIME_OUT)
+        xfstests.run_test(test_type, log_path, result, data_disk, self.TIME_OUT - 20)
```
squirrelsc: Add comments above to explain the magic number.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -171,11 +171,15 @@ def run_test(
...
         # this is the actual process name, when xfstests runs.
-        pgrep.wait_processes("check", timeout=timeout)
-
-        self.check_test_results(
-            log_path=log_path, test_type=test_type, result=result, data_disk=data_disk
-        )
+        try:
+            pgrep.wait_processes("check", timeout=timeout)
+        finally:
+            self.check_test_results(
```
squirrelsc, line 177: Make sure the `check_test_results` won't raise another exception by incomplete content.

microsoft/testsuites/xfstests/xfstesting.py
paxue: Sure. I will add it

squirrelsc:
@LiliDeng LGTM

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -144,6 +144,7 @@ def before_case(self, log: Logger, **kwargs: Any) -> None:
          standard data disk with xfs type system.
         """,
         requirement=simple_requirement(
+            min_core_count=8,
```
LiliDeng: there is no enough evidence to show the timeout issue is related with core count, let's remove this change, and keep monitor the test results data, then decide the next step.

paxue: removed

LiliDeng (approved):

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -125,7 +125,7 @@ class Xfstesting(TestSuite):
     # commit d0c7feaf8767 ("xfs: add agf freeblocks verify in xfs_agf_verify")
     # TODO: will figure out the detailed reason of every excluded case.
     excluded_tests = (
-        "generic/211 generic/430 generic/431 generic/434 xfs/438 xfs/490"
+        "generic/211 generic/430 generic/431 generic/434 generic/738 xfs/438 xfs/490"
```
squirrelsc, line 128: Add comments above, why 738 is removed. Is the timeout resoled after this test is removed?

paxue: yes, tested for many times, it seems 738 is the blocking issue. I will send another PR to add this comment. After removing 738 it passed on 2 core VMs

PR #3275, opened 2024-04-24
iperf - exclude non json part from output

LiliDeng:


lisa/tools/iperf3.py
```diff
@@ -78,10 +78,7 @@ class Iperf3(Tool):
         r" Gbits/sec.*receiver([\w\W]*?))",
         re.MULTILINE,
     )
-    _json_pattern = re.compile(
-        r"[\w\W]*?(?P<json>{[\w\W]*.*)",
-        re.MULTILINE,
-    )
+    _json_pattern = re.compile(r"\{.*\}", re.DOTALL)
```
squirrelsc, line 84: Add examples above. And from the PR description and commit message, I cannot understand why makes this change.

lisa/tools/iperf3.py
```diff
@@ -466,6 +463,6 @@ def _get_bandwidth(self, result: str, pattern: Pattern[str]) -> Decimal:
 
     def _pre_handle(self, result: str) -> str:
         result = result.replace("-nan", "0")
-        result_matched = self._json_pattern.match(result)
+        result_matched = self._json_pattern.search(result)
```
squirrelsc: Try if it's possible to use lisa internal wrappers.

lisa/tools/iperf3.py
LiliDeng: added

lisa/tools/iperf3.py

squirrelsc (approved):

PR #3277, opened 2024-04-24
add configure_user_permissions

LiliDeng:


lisa/node.py
```diff
@@ -516,6 +528,16 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self.shell.initialize()
         self.os: OperatingSystem = OperatingSystem.create(self)
         self.capture_system_information("started")
+        if (
+            self._first_initialize
+            and self.is_test_target
+            and isinstance(self.os, CBLMariner)
+        ):
+            self.configure_user_permissions()
```
squirrelsc, line 536: Add code comments to explain the logic.

lisa/node.py
```diff
@@ -516,6 +528,16 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self.shell.initialize()
         self.os: OperatingSystem = OperatingSystem.create(self)
         self.capture_system_information("started")
+        if (
```
squirrelsc, line 531: The fix should be in the test case level, in case some mariner case needs to test `wheel` functionality. Please check with test case owner to fix it.

squirrelsc: Another way is to provide a tool called `su`, and handle it there. Even to add a tool, it needs the test case owner to fix it.

PR #3279, opened 2024-04-24
mdatp: Add noexec workaround

mcgov:
LISA ~~has a bug where it~~ (edited) won't consider mount permissions before placing the working path. This means you'll end with a path where you can maybe read and write but not execute anything regardless of the file permissions.

This example workaround demonstrates how to detect the issue and a possible workaround, remounting the partition with the executable permission.

### DIFF
--- a/lisa/tools/df.py
+++ b/lisa/tools/df.py
@@ -33,7 +33,7 @@ class Df(Tool):
         ).stdout
         df_entries = find_patterns_groups_in_lines(out, [self._DF_ENTRY_REGEX])[0]
         if len(df_entries) == 0:
-            self._log.warn(f"Path: {directory} not exist!")
+            self._log.info(f"Path: {directory} not exist!")
             return None
         df_entry = df_entries[0]
         return PartitionInfo(
--- a/lisa/tools/df.py
+++ b/lisa/tools/df.py
@@ -25,7 +25,7 @@ class Df(Tool):
     def can_install(self) -> bool:
         return True
 
-    def get_mount_info_for_dir(
+    def get_partition_by_path(
         self, directory: str, force_run: bool = False
     ) -> Optional[PartitionInfo]:
         out = self.run(
--- a/lisa/tools/df.py
+++ b/lisa/tools/df.py
@@ -25,6 +25,26 @@ class Df(Tool):
     def can_install(self) -> bool:
         return True
 
+    def get_mount_info_for_dir(
+        self, directory: str, force_run: bool = False
+    ) -> Optional[PartitionInfo]:
+        out = self.run(
+            parameters=f"{directory}", force_run=force_run, shell=True, sudo=True
+        ).stdout
+        df_entries = find_patterns_groups_in_lines(out, [self._DF_ENTRY_REGEX])[0]
+        if len(df_entries) == 0:
+            self._log.warn(f"Path: {directory} not exist!")
+            return None
+        df_entry = df_entries[0]
+        return PartitionInfo(
+            name=df_entry["name"],
+            mountpoint=df_entry["mountpoint"],
+            available_blocks=int(df_entry["available"]),
+            used_blocks=int(df_entry["used"]),
+            total_blocks=int(df_entry["total"]),
+            percentage_blocks_used=int(df_entry["percentage_use"]),
+        )
+
     def get_partitions(self, force_run: bool = False) -> List[PartitionInfo]:
         # run df and parse the output
         # The output is in the following format:
--- a/lisa/tools/mount.py
+++ b/lisa/tools/mount.py
@@ -22,18 +22,24 @@ class PartitionInfo(object):
     disk: str
     mount_point: str
     type: str
-
+    options: List[str]
     # /dev/sda1
     # /dev/sdc
     _disk_regex = re.compile(r"\s*\/dev\/(?P<disk>\D+).*")
 
-    def __init__(self, name: str, mount_point: str, fs_type: str) -> None:
+    def __init__(
+        self, name: str, mount_point: str, fs_type: str, options: Optional[List[str]]
+    ) -> None:
         self.name = name
         self.mount_point = mount_point
         self.type = fs_type
         matched = self._disk_regex.fullmatch(name)
         assert matched
         self.disk = matched.group("disk")
+        if options:
+            self.options = options
+        else:
+            self.options = []
 
     def __str__(self) -> str:
         return self.__repr__()
@@ -43,7 +49,8 @@ class PartitionInfo(object):
             f"name: {self.name}, "
             f"disk: {self.disk}, "
             f"mount_point: {self.mount_point}, "
-            f"type: {self.type}"
+            f"type: {self.type}, "
+            f"options: {','.join(self.options)}"
         )
 
 
@@ -57,7 +64,8 @@ class Mount(Tool):
     # /dev/sda1 on / type ext4 (rw,relatime,discard)
     # /dev/sda1 on /mnt/a type ext4 (rw,relatime,discard)
     _partition_info_regex = re.compile(
-        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+type\s+(?P<type>.*)\s+.*"
+        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+type"
+        r"\s+(?P<type>.*)\s+\(?(?P<options>.*)\)?"
     )
     _mount_info_regex = re.compile(
         r"\s*(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+type\s+(?P<type>.*)\s\(+.*"
@@ -65,7 +73,8 @@ class Mount(Tool):
 
     # /dev/da1p1 on /mnt/resource (ufs, local, soft-updates)
     _partition_info_regex_bsd = re.compile(
-        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+(\((?P<type>.*),.*,.*\))"
+        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+"
+        r"(\((?P<type>.*),(?P<options>.*,.*)\))"
     )
 
     @property
@@ -130,11 +139,18 @@ class Mount(Tool):
                 matched = self._partition_info_regex.fullmatch(line)
             if matched:
                 partition_name = matched.group("name")
+                option_match = matched.group("options")
+                if not option_match:
+                    options = []
+                else:
+                    split_options = option_match.split(",")
+                    options = [option.strip() for option in split_options]
                 partition_info.append(
                     PartitionInfo(
-                        f"/dev/{partition_name}",
-                        matched.group("mount_point"),
-                        matched.group("type"),
+                        name=f"/dev/{partition_name}",
+                        mount_point=matched.group("mount_point"),
+                        fs_type=matched.group("type"),
+                        options=options,
                     )
                 )
 
--- a/lisa/tools/mount.py
+++ b/lisa/tools/mount.py
@@ -107,6 +107,27 @@ class Mount(Tool):
         cmd_result = self.node.execute(" ".join(runline), shell=True, sudo=True)
         cmd_result.assert_exit_code()
 
+    @retry(tries=24, delay=5)
+    def remount(
+        self,
+        point: str,
+        options: List[str],
+    ) -> None:
+        runline = [self.command]
+        if isinstance(self.node.os, BSD):
+            # BSD allows updating mount options with -u flag
+            # User must specify all options on the commandline
+            runline.append("-u")
+        else:
+            # Linux 'mount' allows remounting using old mount options
+            # ex: mount -o remount,exec /path
+            # will remount path and replace noexec with exec, but keep other any options
+            options = ["remount"] + options
+        runline.append(f"-o {','.join(options)}")
+        runline.append(f"{point}")
+        cmd_result = self.node.execute(" ".join(runline), shell=True, sudo=True)
+        cmd_result.assert_exit_code()
+
     def umount(
         self, disk_name: str, point: str, erase: bool = True, fs_type: str = ""
     ) -> None:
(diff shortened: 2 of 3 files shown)


lisa/tools/mount.py
```diff
@@ -57,15 +64,17 @@ class Mount(Tool):
...
     )
     _mount_info_regex = re.compile(
         r"\s*(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+type\s+(?P<type>.*)\s\(+.*"
     )
 
     # /dev/da1p1 on /mnt/resource (ufs, local, soft-updates)
     _partition_info_regex_bsd = re.compile(
-        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+(\((?P<type>.*),.*,.*\))"
+        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+"
+        r"(\((?P<type>.*),(?P<options>.*,.*)\))"
```
squirrelsc: Should the "options" be "local" in the example? Is it expected?

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -22,6 +24,59 @@
...
+    # we currently discard the mount options so they're not exposed
+    # in tools.Mount
+    for part in node.tools[Mount].get_partition_info():
+        node.log.debug(f"checking partition {str(part)}...")
+        if part.mount_point == partition.mountpoint:
+            return "noexec" in part.options
+    # This path would mean that df and mount are returning different
+    # info about the mount points on the VM.
+    # There should be no way for this to happen, assert if it does
+    assert f"Could not find partition info for {partition.mountpoint}!"
```
squirrelsc: This assert doesn't work. It asserts a string is not None. So it won't be failed.

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -22,6 +24,59 @@
...
+    # EXAMPLE WORKAROUND:
+    # - Get the mount point for the path.
+    # - Check if the mount point is mounted noexec.
+    # - if yes, remount with exec flag
+    working_path = node.get_working_path()
+    mount_info = node.tools[Df].get_mount_info_for_dir(
+        directory=working_path.as_posix()
+    )
+    if not mount_info:
+        node.log.warn(
```
squirrelsc: Don't use log.warn. Use log.info or raise exception.

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -22,6 +24,59 @@
 EXIT_ONBOARD_INFO_FOUND = 255
 
 
+def working_path_workaround(node: Node) -> None:
```
squirrelsc: Make the method name clearer like `remove_noexec`

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -22,6 +24,59 @@
...
+    # EXAMPLE WORKAROUND:
+    # - Get the mount point for the path.
+    # - Check if the mount point is mounted noexec.
+    # - if yes, remount with exec flag
+    working_path = node.get_working_path()
+    mount_info = node.tools[Df].get_mount_info_for_dir(
+        directory=working_path.as_posix()
+    )
+    if not mount_info:
```
squirrelsc, line 33: Use `if mount_info:`. In most case, the `if` block should be positive.

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -22,6 +24,59 @@
...
+            return "noexec" in part.options
+    # This path would mean that df and mount are returning different
+    # info about the mount points on the VM.
+    # There should be no way for this to happen, assert if it does
+    assert f"Could not find partition info for {partition.mountpoint}!"
+    # appease pylint by returning after assert
+    return False
+
+
+def fix_noexec_working_path(node: Node, partition: PartitionInfo) -> None:
```
squirrelsc: `remount_without_noexec`

lisa/tools/mount.py
mcgov: I'm assuming BSD partitions can contain more than three options in the tuple, similar to Linux partitions.

mcgov: So the pattern should grab the fs type; the first option, then grab the rest as generic 'options'

microsoft/testsuites/mdatp/mdatp.py
mcgov: Oh wow, yep! Thank you so much, I wanted to finish this example so I wrote this around 1AM. Will fix

microsoft/testsuites/mdatp/mdatp.py
mcgov: I don't want to raise an exception here and we can put it in info if you prefer strongly. I don't get why we have three log levels if we can't use them to signal the context of messages. 

This is an unexpected situation that shouldn't be fatal:
- The mount info tools are behaving in a way we don't expect, but...
-  we can still (probably) run the test since this section of code is an example workaround for the .01% of images with this issue.

It's a non-fatal but concerning log mesage, it seems like a good candidate for a 'warn' message to me. It's about the context of the message.

microsoft/testsuites/mdatp/mdatp.py
mcgov: Hmm maybe this should also be a warning too and not an assert.

 I don't want to raise an exception in this code path since almost all images (>99%) don't need this workaround.
The problem also isn't a bug in this test, it's a LISA bug where we are not paying attention to mount options when creating our working paths. We're also not paying attention to file ACLs, but that's another issue. I don't have time or enough context about the inheritances in Node to fix the general case. I'm hoping we can create a bug and someone can hoist this code to the appropriate parent class to ensure all tests get the benefit of this fix.

microsoft/testsuites/mdatp/mdatp.py
mcgov: This does make it way cleaner! Thanks

microsoft/testsuites/mdatp/mdatp.py
mcgov: apply_working_path_noexec_workaround?

I want to ensure it's clear that this is a workaround for LISA's handling of working path creation. I'm expecting that someone (hopefully not me, but maybe me) will revisit this commit later and generalize this workaround and hoist it to one of the Node parent classes.

lisa/tools/mount.py
squirrelsc: To capture all rest, it's better to end with `(?P<options>.*)\))` . It handles more than 2 options.

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: If it's no way to happened, raise an exception to make it obvious and fix it. It's a kind of safeguard code.

squirrelsc: BTW, it's not a LISA bug, some systems are noexec to be more secure by design. LISA perfers to use commands, instead of scripts. If the script is rewritten in native LISA, there is no such error.
LISA shouldn't mark a mount point exec blindly, it may miss some test points.

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: Please use info level.
The warn method is implemented in Python package, not in LISA. In the practical, most people don't care warn level in the routine pipelines. When developing a new case, it needs to read every new log line to make sure it's clear to others.

lisa/tools/df.py
```diff
@@ -25,6 +25,26 @@ def command(self) -> str:
...
 
+    def get_mount_info_for_dir(
+        self, directory: str, force_run: bool = False
+    ) -> Optional[PartitionInfo]:
+        out = self.run(
+            parameters=f"{directory}", force_run=force_run, shell=True, sudo=True
+        ).stdout
+        df_entries = find_patterns_groups_in_lines(out, [self._DF_ENTRY_REGEX])[0]
+        if len(df_entries) == 0:
+            self._log.warn(f"Path: {directory} not exist!")
```
squirrelsc: Don't print warn line. It's not helpful for troubleshooting. Raise an exception, and we can know which images need to be fixed.

lisa/tools/df.py
```diff
@@ -25,6 +25,26 @@ def command(self) -> str:
     def can_install(self) -> bool:
         return True
 
+    def get_mount_info_for_dir(
```
squirrelsc: Like other methods, call it `get_partition_by_path`. "mount info" is not used in this class. Avoid bringing new unnecessary concepts.

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: The name should focus on action taken, not put everything in the name. The "workaround" has no detail, it's not needed.
The working_path is confusing, because noexec is not an attribute on a path. The method is to "remove" the "noexec". You can see it to "apply" "noexec_workaround", but "remove noexec" is clearer on "how to apply workaround".

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: Sorry, from the content, `remount_with_exec` is clear.

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -22,6 +24,70 @@
...
+
+
+def remount_without_noexec(node: Node, partition: PartitionInfo) -> None:
+    # Fix issue where working path is created in a partition marked
+    # 'noexec' meaning file permissions for x are ignored.
+    # Attempt to remount the parition to allow executables.
+    # The other fix would just be adding a disk, creating a parition
+    # marking it executable, and using that as the working path.
+    node.tools[Mount].mount(
+        options="remount,exec",
```
squirrelsc: How about existing other options, like "local" in the example? It may need to get all options, and remove "noexec", and then add "remount" and "exec".

lisa/tools/mount.py
mcgov: will do!

microsoft/testsuites/mdatp/mdatp.py
mcgov: > LISA shouldn't mark a mount point exec blindly, it may miss some test points.
I wasn't clear on what the bug is, I think.

[My assumption on what working path is for is is based on the comments about what the working path is for:](https://vscode.dev/github/microsoft/lisa/blob/main/lisa/node.py#L187)

My impression is that the working path should be a place for tests to place things. Tests should be able to read and write things in the working path. I think a lot of test authors would assume they will be able to execute things in there as well. I think LISA should ensure that the working path (wherever it ends up) is r/w/x. LISA might have to be flexible about using available space or maybe even adding space.

> BTW, it's not a LISA bug, some systems are noexec to be more secure by design. LISA perfers to use commands, instead of scripts. If the script is rewritten in native LISA, there is no such error.

I don't think it's fair to blame the test for not checking whether the partition returned by` get_working_path()` is mounted with noexec. There was no mention of this mount option in the code before this PR, even in comments. The DF and Mount tool ignore mount options, so even if I had known about noexec and wanted to check the mount options before, I couldn't have using the tools provided by LISA.

I'm not trying to blame or shame anyone; bugs are a part of software development. Maybe bug is too strong of a word? LISA not checking the mount options of a partition before placing a working path is more of an undocumented non-intuitive behavior. 

It's not strange we haven't run into it before: it's only present on 1/6000th of the images in the marketplace.

microsoft/testsuites/mdatp/mdatp.py
mcgov: I'll stop bringing this up, maybe I can make a feature request and ask people who give me feedback about LISA's logging to upvote it 😜

microsoft/testsuites/mdatp/mdatp.py
mcgov: > If it's no way to happened, raise an exception to make it obvious and fix it. It's a kind of safeguard code.

It's not that there is no way for it to happen, it's just that it is a non-fatal error. This section of code can fail and the test can still pass, this section of code can fail and the test can still fail. We'll still catch the failures since they will still show up as failures, we'll allow the tests who should pass to pass.

lisa/tools/df.py
mcgov: fixed

microsoft/testsuites/mdatp/mdatp.py
squirrelsc: I got feedback from a team, they use ado pipeline, and look at their error/warn lines. They don't want LISA has error/warning lines, which are not fixed.

microsoft/testsuites/mdatp/mdatp.py
mcgov: fixed

lisa/tools/df.py
mcgov: fixed

microsoft/testsuites/mdatp/mdatp.py
mcgov: refactored to handle bsd/linux remount arg differences and preserve options on both

microsoft/testsuites/mdatp/mdatp.py
```diff
@@ -22,6 +23,56 @@
...
+            options = [option for option in partition.options if option != "noexec"]
+        else:
+            # Linux allows remounting as exec with other options preserved.
+            # just have to pass 'exec' to remount
+            options = ["exec"]
+        # initiate the remount.
+        node.tools[Mount].remount(point=mountpoint, options=options)
+
+    # otherwise, no issues with partition, nothing to fix.
+    return
```
squirrelsc: It doesn't need to return at the end, if no return parameter defined.

lisa/tools/df.py
squirrelsc: Haven't fixed.

microsoft/testsuites/mdatp/mdatp.py
mcgov: Isn't this more of a style choice then? Removed.

microsoft/testsuites/mdatp/mdatp.py
mcgov: ok

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3280, opened 2024-04-25
Fixing nic initialization for freebsd nodes

r-dailey:
Added a different check to the nic initialization code to check for a freebsd operating system and updated the nic provisioning test case to use the built in method instead of call the Nic class directly.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -22,7 +22,7 @@ from lisa import schema
 from lisa.executable import Tools
 from lisa.feature import Features
 from lisa.nic import Nics, NicsBSD
-from lisa.operating_system import BSD, OperatingSystem
+from lisa.operating_system import OperatingSystem
 from lisa.secret import add_secret
 from lisa.tools import Chmod, Df, Echo, Lsblk, Mkfs, Mount, Reboot, Uname, Wsl
 from lisa.tools.mkfs import FileSystem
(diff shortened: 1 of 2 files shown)


lisa/node.py
```diff
@@ -1215,7 +1215,8 @@ def create_nics(node: Node) -> Nics:
     """
     Returns a Nics object for the node based on the OS type.
     """
-    if isinstance(node.os, BSD):
+    os = node.execute(cmd="uname", no_error_log=True).stdout
+    if os.__contains__("FreeBSD"):
```
squirrelsc: `"FreeBSD" in os`

lisa/node.py
```diff
@@ -1215,7 +1215,8 @@ def create_nics(node: Node) -> Nics:
     """
     Returns a Nics object for the node based on the OS type.
     """
-    if isinstance(node.os, BSD):
+    os = node.execute(cmd="uname", no_error_log=True).stdout
```
squirrelsc, line 1220: Add comments above to explain why we don't use `isinstance`. So in future, it won't be changed back.

squirrelsc:
@r-dailey, please fix the flake8 errors.

squirrelsc (dismissed):

squirrelsc (approved):

PR #3281, opened 2024-04-26
Initialize package before adding repo for SUSE image to avoid system …

lubaihua33:
Initialize package before adding repo for SUSE image to avoid system locked

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -1988,6 +1988,7 @@ class Suse(Linux):
         repo_name: Optional[str] = None,
         keys_location: Optional[List[str]] = None,
     ) -> None:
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -2004,6 +2004,7 @@ def clean_package_cache(self) -> None:
     def add_azure_core_repo(
         self, repo_name: Optional[AzureCoreRepo] = None, code_name: Optional[str] = None
     ) -> None:
+        self._initialize_package_installation()
```
LiliDeng: thanks for the fix, put it in `self.add_repository()`

LiliDeng (approved):

PR #3286, opened 2024-04-30
Reduce min data_disk_count to 4 in IO perf test.

SRIKKANTH:
Having min data_disk_count =16 in these testcases is blocking tests on smaller sizes.

### DIFF
--- a/microsoft/testsuites/performance/storageperf.py
+++ b/microsoft/testsuites/performance/storageperf.py
@@ -102,7 +102,7 @@ class StoragePerformance(TestSuite):
                 data_disk_type=schema.DiskType.PremiumSSDLRS,
                 os_disk_type=schema.DiskType.PremiumSSDLRS,
                 data_disk_iops=search_space.IntRange(min=5000),
-                data_disk_count=search_space.IntRange(min=16),
+                data_disk_count=search_space.IntRange(min=4),
             ),
         ),
     )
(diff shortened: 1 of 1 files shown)


LiliDeng:
@johnsongeorge-w please review, thanks.

johnsongeorge-w (changes requested):
Set the default for disk count and min iops as 16 and 5000 respectively in the default runbook.

SRIKKANTH:
> Set the default for disk count and min iops as 16 and 5000 respectively.
@johnsongeorge-w This is what the default in the test case which is blocking us the test from running on smaller tests. 
I am reducing the min disc count to 4 which is the minimum number of discs supported any azure VM SKU.

johnsongeorge-w:
> > Set the default for disk count and min iops as 16 and 5000 respectively.
> > @johnsongeorge-w This is what the default in the test case which is blocking us the test from running on smaller tests.
> > I am reducing the min disc count to 4 which is the minimum number of discs supported any azure VM SKU.

I'm talking about the defaults in runbook

SRIKKANTH:
> > > Set the default for disk count and min iops as 16 and 5000 respectively.
> > > @johnsongeorge-w This is what the default in the test case which is blocking us the test from running on smaller tests.
> > > I am reducing the min disc count to 4 which is the minimum number of discs supported any azure VM SKU.
> 
> I'm talking about the defaults in runbook
The default runbook ("runbook\partial\platform_azure.yml") has values data_disk_iops=0, data_disk_count=0. 
Changing these values will result in creation of each test VM with that config. 
I don't think its a good idea.

PR #3287, opened 2024-05-01
Enable NVIDIA persistence mode by default for GPU test 

henryli001:
Enable nvidia-persistence mode after installing the NVIDIA GPU driver by default. Fixed this [bug](https://microsoft.visualstudio.com/OS/_workitems/edit/49246027/)


microsoft/testsuites/gpu/gpusuite.py
```diff
@@ -415,6 +415,14 @@ def _install_driver(node: Node, log_path: Path, log: Logger) -> None:
 
     __install_driver_using_sdk(node, log, log_path)
 
+    # Enabling nvidia persistence mode which is recommended by NVIDIA
```
LiliDeng, line 418: do you have a doc link which can put in comments?

PR #3288, opened 2024-05-02
VMAccess Fix - Add user to wheel group in Mariner 2.0 before running 'su'

mayankdaruka-msft:
VMAccess tests have been failing on Mariner 2.0 because the non-root user is executing a 'su' command to determine the validity of the username-password combination for an account, without having permissions to do so.

In Mariner, the user has to be added to wheel group before running a 'su' command. This PR introduces a fix to add the user to wheel group first.

### DIFF
--- a/microsoft/testsuites/vm_extensions/runtime_extensions/vmaccess.py
+++ b/microsoft/testsuites/vm_extensions/runtime_extensions/vmaccess.py
@@ -13,11 +13,14 @@ from lisa import (
     TestSuiteMetadata,
     simple_requirement,
 )
-from lisa.operating_system import BSD
+from lisa.operating_system import BSD, CBLMariner
 from lisa.sut_orchestrator import AZURE
 from lisa.sut_orchestrator.azure.features import AzureExtension
 from microsoft.testsuites.vm_extensions.runtime_extensions.common import (
     create_and_verify_vmaccess_extension_run,
+)
+from lisa.tools import (
+    Usermod,
 )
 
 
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/runtime_extensions/vmaccess.py
```diff
@@ -77,9 +80,16 @@ def _validate_password(
...
+    if isinstance(node.os, CBLMariner):
+        if node.os.information.version >= "2.0.0":
+            # In Mariner 2.0, there is a security restriction that only allows wheel group users to use 'su' command
+            # Add current user (specified during VM creation) to wheel group in Mariner
+            node.tools[Usermod].add_user_to_group("wheel", sudo=True)
+
     # simple command to determine if username password combination is valid/invalid
     node.execute(
-        cmd=f'echo "{password}" | su --command true - {username}',
+        cmd=f'echo "{password}" | su --command true {username}',
```
squirrelsc, line 99: Does it need to remove the `-`?

squirrelsc: Please call `add_secret(password)` to add the password into secrets, so it won't appear in the log.

mayankdaruka-msft: It should be without the `-`. That seems to not work on Mariner as well.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3293, opened 2024-05-13
use user delegation sas token 

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1226,70 +1223,44 @@ def get_storage_credential(
     return {"account_name": account_name, "account_key": key.value}
 
 
-def generate_blob_sas_token(
-    credential: Any,
-    subscription_id: str,
-    cloud: Cloud,
-    account_name: str,
-    resource_group_name: str,
+def generate_user_delegation_sas_token(
```
squirrelsc, line 1226: It shouldn't use sas token anymore. The file should be able to access without sas, if the user has permission.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1517,12 +1470,12 @@ def copy_vhd_to_storage(
                 if not os.path.exists(azcopy_path):
                     raise LisaException(f"{azcopy_path} does not exist")
 
-                sas_token = generate_sas_token(
+                sas_token = generate_user_delegation_sas_token(
```
squirrelsc, line 1473: The sas token is not necessary anymore. The user, who runs lisa, should have permission to create vhds in the lisa subscription.

lisa/sut_orchestrator/azure/common.py
LiliDeng: I didn't find a way to copy vhd from one storage account into another storage account without sas in the same subscription, issue raised here https://github.com/Azure/azure-sdk-for-python/issues/35571 
Maybe we can convert the vhd into shared gallery image, then make replication when find a qualified location.

squirrelsc:
Make sure it doesn't break existing runbook. If some fields are not necessary, print deprecating message with instruction. We remove them after 1 or 3 months.

squirrelsc (approved):

PR #3297, opened 2024-05-17
Boot diagnostics: enable with managed storage account

LiliDeng:


lisa/sut_orchestrator/azure/arm_template.json
```diff
@@ -2,12 +2,6 @@
     "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
-        "storage_name": {
```
squirrelsc, line 5: What's useful of this file now? Should we start to delete it?

squirrelsc (approved):

lisa/sut_orchestrator/azure/arm_template.json
LiliDeng: I think so, let me raise another PR to delete it.

PR #3312, opened 2024-06-10
waagent: support finding python/waagent in /usr on Flatcar

jepio:
Current Flatcar images (since 3637.0.0) contain waagent and python in /usr/{s,}bin. Search for python in the candidate path and use that to decide whether to call waagent directly or use the module path prefixed with /usr/share/oem.

### DIFF
--- a/lisa/sut_orchestrator/azure/tools.py
+++ b/lisa/sut_orchestrator/azure/tools.py
@@ -32,6 +32,7 @@ class Waagent(Tool):
         "/usr/libexec/platform-python",
         # for flatcar
         "/usr/share/oem/python/bin/python3",
+        "/usr/share/oem/python/bin/python",
     ]
     _src_url = "https://github.com/Azure/WALinuxAgent/"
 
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/tools.py
```diff
@@ -221,7 +223,7 @@ def _get_waagent_conf_path(self) -> str:
         if result.exit_code == 0:
             waagent_path = result.stdout
         else:
-            if isinstance(self.node.os, CoreOs):
+            if isinstance(self.node.os, CoreOs) and "/usr/share/oem/" in python_cmd:
```
squirrelsc: Please add it as a sub condition under CoreOs, and add comments in code to explain why the new check condition needs.

jepio: Thanks for the review @squirrelsc, I've removed this change from the PR since I realized this block will correctly discover the config path in both cases:
```
        result = self.node.execute(
            f'{python_cmd} -c "from azurelinuxagent.common.osutil import get_osutil;'
            'print(get_osutil().agent_conf_file_path)"',
            sudo=use_sudo,
        )
        if result.exit_code == 0:
            waagent_path = result.stdout
```

LiliDeng (approved):

PR #3314, opened 2024-06-11
Tool to check supported hugepagesize and fix in dpdk tests while mounting with 1G hugepage size

kanchansenlaskar:
Some dpdk tests are failing during mounting as hugepage size of 1G is not supported all the time.
A tool has been written to check the hugepage sizes available. If 1G hugepage is not supported, it will skip the tests.

This change has been tested (both positive and negative scenario)

### DIFF
--- a/lisa/tools/hugepages.py
+++ b/lisa/tools/hugepages.py
@@ -1,0 +1,112 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import re
+from enum import Enum
+from typing import Any, Set
+
+from lisa.executable import Tool
+from lisa.tools.echo import Echo
+from lisa.tools.free import Free
+from lisa.tools.ls import Ls
+from lisa.tools.lscpu import Lscpu
+from lisa.tools.mkfs import FileSystem
+from lisa.tools.mount import Mount
+from lisa.util import NotEnoughMemoryException, UnsupportedOperationException
+
+PATTERN_HUGEPAGE = re.compile(
+    r"^.*hugepages-(?P<hugepage_size_in_kb>\d+)kB.*",
+)
+
+
+class HugePageSize(Enum):
+    HUGE_1GB = 1048576
+    HUGE_2MB = 2048
+    HUGE_512MB = 524288
+    HUGE_16GB = 16777216
+
+
+class Hugepages(Tool):
+    @property
+    def command(self) -> str:
+        return ""
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._hugepages_dir_path = "/sys/devices/system/node/node*/hugepages"
+
+    def get_hugepage_sizes_in_kb(self) -> Set[int]:
+        """
+        This function get the hugepage sizes in kB available on the system.
+        This is done by listing the hugepages directory
+        and extracting the size from the directory name.
+        e.g. [
+        "'/sys/devices/system/node/node0/hugepages/hugepages-16777216kB/'",
+        "'/sys/devices/system/node/node0/hugepages/hugepages-2048kB/'",
+        "'/sys/devices/system/node/node0/hugepages/hugepages-524288kB/'"
+        ]
+        """
+        hugepage_sizes: Set[int] = set()
+        ls = self.node.tools[Ls]
+        # if not ls.path_exists(self._hugepages_dir_path, sudo=True):
+        #     raise SkippedException(f"path: {self._hugepages_dir_path} does not exist")
+        hugepage_size_dir_names = ls.list_dir(self._hugepages_dir_path, sudo=True)
+        for hugepage_size_dir_name in hugepage_size_dir_names:
+            matched_hugepage = PATTERN_HUGEPAGE.match(hugepage_size_dir_name)
+            if not matched_hugepage:
+                raise UnsupportedOperationException(
+                    "No supported hugepage size found in the"
+                    f"list: {hugepage_size_dir_name}"
+                )
+            hugepage_sizes.add(int(matched_hugepage.group("hugepage_size_in_kb")))
+        return hugepage_sizes
+
+    def _enable_hugepages(self, hugepage_size_kb: HugePageSize) -> None:
+        echo = self.node.tools[Echo]
+        meminfo = self.node.tools[Free]
+        nics_count = len(self.node.nics.get_nic_names())
+        numa_nodes = self.node.tools[Lscpu].get_numa_node_count()
+
+        request_space_kb = (nics_count - 1) * 1024 * 1024 * numa_nodes * 2
+        free_memory_kb = meminfo.get_free_memory_kb()
+
+        if free_memory_kb < request_space_kb:
+            raise NotEnoughMemoryException(
+                f"Not enough {hugepage_size_kb.value} KB pages "
+                "available for DPDK! "
+                f"Requesting {request_space_kb} KB found {free_memory_kb} "
+                "KB free."
+            )
+
+        request_pages = request_space_kb // hugepage_size_kb.value
+        for i in range(numa_nodes):
+            # nr_hugepages will be written with the number calculated
+            # based on 2MB hugepages if not specified, subject to change
+            # this based on further discussion
+            echo.write_to_file(
+                f"{request_pages}",
+                self.node.get_pure_path(
+                    f"/sys/devices/system/node/node{i}/hugepages/"
+                    f"hugepages-{hugepage_size_kb.value}kB/nr_hugepages"
+                ),
+                sudo=True,
+            )
+
+    def init_hugepages(self, hugepage_size: HugePageSize) -> None:
+        mount = self.node.tools[Mount]
+        hugepage_sizes = self.get_hugepage_sizes_in_kb()
+        if hugepage_size.value not in hugepage_sizes:
+            raise UnsupportedOperationException(
+                f"Supported hugepage sizes in kB are: {hugepage_sizes}."
+                f" {hugepage_size.value} kB size is not available"
+            )
+        if not mount.check_mount_point_exist(f"/mnt/huge-{hugepage_size.value}kb"):
+            mount.mount(
+                name="nodev",
+                point=f"/mnt/huge-{hugepage_size.value}kb",
+                fs_type=FileSystem.hugetlbfs,
+                options=f"pagesize={hugepage_size.value}KB",
+            )
+        self._enable_hugepages(hugepage_size_kb=hugepage_size)
--- a/lisa/util/__init__.py
+++ b/lisa/util/__init__.py
@@ -207,6 +207,21 @@ class UnsupportedKernelException(LisaException):
         return message
 
 
+class NotEnoughMemoryException(LisaException):
+    """
+    This exception is used to indicate that that the system does not have enough memory.
+    """
+
+    def __init__(self, message: str = "") -> None:
+        self._extended_message = message
+
+    def __str__(self) -> str:
+        message = "Not enough memory"
+        if self._extended_message:
+            message = f"{message}. {self._extended_message}"
+        return message
+
+
 class UnsupportedCpuArchitectureException(LisaException):
     """
     This exception is used to indicate that a test case does not support the
--- a/microsoft/testsuites/dpdk/dpdkutil.py
+++ b/microsoft/testsuites/dpdk/dpdkutil.py
@@ -13,10 +13,12 @@ from lisa import (
     LisaException,
     Logger,
     Node,
+    NotEnoughMemoryException,
     RemoteNode,
     SkippedException,
     UnsupportedDistroException,
     UnsupportedKernelException,
+    UnsupportedOperationException,
     constants,
     notifier,
 )
@@ -28,7 +30,7 @@ from lisa.tools import (
     Dmesg,
     Echo,
     Firewall,
-    Free,
+    Hugepages,
     Ip,
     KernelConfig,
     Kill,
@@ -36,12 +38,11 @@ from lisa.tools import (
     Lsmod,
     Lspci,
     Modprobe,
-    Mount,
     Ntttcp,
     Ping,
     Timeout,
 )
-from lisa.tools.mkfs import FileSystem
+from lisa.tools.hugepages import HugePageSize
 from lisa.util.constants import SIGINT
 from lisa.util.parallel import TaskManager, run_in_parallel, run_in_parallel_async
 from microsoft.testsuites.dpdk.common import (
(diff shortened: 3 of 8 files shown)


lisa/tools/hugepages.py
```diff
@@ -0,0 +1,130 @@
...
+        return self._hugepage_sizes
+
+    def _enable_hugepages(self, enable_gibibyte_hugepages: bool = False) -> None:
+        echo = self.node.tools[Echo]
+
+        meminfo = self.node.tools[Free]
+        nics_count = len(self.node.nics.get_nic_names())
+
+        numa_nodes = self.node.tools[Lscpu].get_numa_node_count()
+        request_pages_2mb = (nics_count - 1) * 1024 * numa_nodes
```
mcgov: As long as we're rewriting this; we should rip out this 2MiB or 1GiB assumption. ARM64 apparently has multiple other hugepage options, so we need to change how we enable them.

Consider how you would rewrite these `_enable_hugepages`  and `init_hugepages` functions now that we know there are more than 2 possible sizes of hugepages. We still want to check that we have enough memory; there's probably an easier way to do that. I started as a C dev and I have noticed I tend to write C even when I'm writing other languages 😅 Bad habit.

A good first step would be getting rid of the `enable_gibibyte_hugepages` boolean. That option is there because of the assumption that we're either using 2MiB or 1GiB hugepages. So now we can just rip it out. 

We don't want to assume the hugepage sizes; but, we *can* assume that **callers** will know what size they want. Tests will be testing a single size at a time; so it's good to assume that people will build tests where there is lots of shared code except one part where they call like "`run_test( .... , hugepage_size=(some constant) )`" . You could try something like creating an Enum of different sizes so the caller can check like...
```
supported_sizes = hugepages.get_hugepage_sizes() 
if not hugepages.HUGEPAGE_2MB in supported_sizes:
   raise SkippedException("2MiB hugepages not supported on this platform")
```
I'm sure @squirrelsc has thoughts, so let's check if this sounds good with him so we don't make you rewrite it twice 😜

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,130 @@
...
+        request_pages_1gb = (nics_count - 1) * numa_nodes
+        memfree_2mb = meminfo.get_free_memory_mb()
+        memfree_1mb = meminfo.get_free_memory_gb()
```
mcgov: typo lol. I know I wrote it, but we can fix it now.

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # request 2iGB memory per nic, 1 of 2MiB pages and 1 GiB page
+        # check there is enough memory on the device first.
+        # default to enough for one nic if not enough is available
+        # this should be fine for tests on smaller SKUs
+
+        if enable_gibibyte_hugepages:
```
mcgov: There HAS to be some easier way to do this. Don't feel obligated to rewrite this verbose nonsense I wrote exactly.

Like, attempting to enable and checking the return code of the write to sysfs? Or attempting to enable and checking if you got the new amount you asked for? I'm sure there's a manpage somewhere with advice on how to check whether the size you asked for was enabled properly.

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,130 @@
...
+                        "hugepages-2048kB/nr_hugepages"
+                    ),
+                    sudo=True,
+                )
+
+    def init_hugepages(self, enable_gibibyte_hugepages: bool = False) -> None:
+        mount = self.node.tools[Mount]
+        if enable_gibibyte_hugepages:
+            hugepage_sizes_in_kb = self.get_hugepage_sizes_in_kb()
+            if "1048576" not in hugepage_sizes_in_kb:
```
mcgov: A good start, but 'naked' string constants are a bad pattern. Maybe make an Enum and make get_hugepage_sizes return the Enum type instead of a string array.

mcgov (changes requested):

lisa/tools/hugepages.py
kanchansenlaskar: check if included in the tool

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,115 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._hugepage_sizes: Set[int] = set()
+        self._hugepages_dir_path = "/sys/devices/system/node/node*/hugepages"
+
+    def get_hugepage_sizes_in_kb(self) -> None:
```
mcgov: It looks odd to me to have a `get_...` function which doesn't return a value. It would make more sense for this to return List[HugePageSize]

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,115 @@
...
+        hugepage_size_dir_names = ls.list_dir(self._hugepages_dir_path, sudo=True)
+        for hugepage_size_dir_name in hugepage_size_dir_names:
+            matched_hugepage = PATTERN_HUGEPAGE.match(hugepage_size_dir_name)
+            if not matched_hugepage:
+                raise LisaException(
+                    f"no hugepage size found in {hugepage_size_dir_name}"
+                )
+            self._hugepage_sizes.add(int(matched_hugepage.group("hugepage_size_in_kb")))
+
+    def _enable_hugepages(
```
mcgov: This shouldn't have a default value. It's an internal function; we should require the caller to provide a value.

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,115 @@
...
+                f"Requesting {request_space_kb} KB found {free_memory_kb} "
+                "KB free."
+                "Test may fail if it cannot allocate memory."
+            )
+        request_pages = request_space_kb // hugepage_size_kb.value
+        for i in range(numa_nodes):
+            # nr_hugepages will be written with the number calculated
+            # based on 2MB hugepages if not specified, subject to change
+            # this based on further discussion
+            echo.write_to_file(
```
mcgov, line 88: Check if the file exists first

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,115 @@
...
+        echo = self.node.tools[Echo]
+        meminfo = self.node.tools[Free]
+        nics_count = len(self.node.nics.get_nic_names())
+        numa_nodes = self.node.tools[Lscpu].get_numa_node_count()
+
+        request_space_kb = (nics_count - 1) * 1024 * 1024 * numa_nodes * 2
+        free_memory_kb = meminfo.get_free_memory_kb()
+
+        if free_memory_kb < request_space_kb:
+            self.node.log.debug(
```
mcgov: If there isn't enough space the allocation will probably fail and the test results will be useless; maybe you should raise an exception here. NotEnoughMemoryException or something (not sure if that type exists yet)

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,115 @@
...
+            # based on 2MB hugepages if not specified, subject to change
+            # this based on further discussion
+            echo.write_to_file(
+                f"{request_pages}",
+                self.node.get_pure_path(
+                    f"/sys/devices/system/node/node{i}/hugepages/"
+                    f"hugepages-{hugepage_size_kb.value}kB/nr_hugepages"
+                ),
+                sudo=True,
+            )
```
mcgov, line 95: Also; check the return code here. If the write failed or we didn't get the value we asked for; raise an exception.

lisa/tools/hugepages.py
```diff
@@ -0,0 +1,115 @@
...
+    def init_hugepages(self, hugepage_size: Optional[HugePageSize] = None) -> None:
+        mount = self.node.tools[Mount]
+        if hugepage_size:
+            self.get_hugepage_sizes_in_kb()
+            if hugepage_size.value not in self._hugepage_sizes:
+                raise SkippedException(
+                    f"Supported hugepage sizes in kB are: {self._hugepage_sizes}."
+                    f" {hugepage_size.value} kB size is not available"
+                )
+            mount.mount(
```
mcgov, line 106: I'm not sure if this is needed on all platforms. You could try checking if sysfs already has the file available for enabling the hugepage size first?

My memory is that this was only a 'don't care' function call before because it was needed on old CentOS versions or something and I was being lazy. Now that we're rewriting it; you can code it better than I did 😜

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -546,7 +475,7 @@ def verify_dpdk_send_receive(
     test_duration: int = variables.get("dpdk_test_duration", 15)
     kill_timeout = test_duration + 5
     test_kits = init_nodes_concurrent(
```
mcgov, line 485: remember that if you're going to raise exceptions if the hugepage allocation fails; you'll need to catch and handle here. We should catch NotEnoughMemoryException and raise a SkippedException(err) that wraps the that exception. Any others; we should allow to bubble up since they imply bugs.

mcgov (commented):
Better! But not quite there yet. You need to handle some errors that will pop up. Some platforms won't support certain hugepage types so we should expect to fail during the enabling phase for some tests.

lisa/tools/hugepages.py
kanchansenlaskar: it is being checked in the earlier step

lisa/tools/hugepages.py
kanchansenlaskar: it doesn't return a code but it does perform an assert inside

lisa/tools/hugepages.py
kanchansenlaskar: Added the step to check if it is mounted already

lisa/util/__init__.py
```diff
@@ -207,6 +207,21 @@ def __str__(self) -> str:
...
+class NotEnoughMemoryException(LisaException):
+    """
+    This exception is used to indicate that that the system does not have enough memory.
+    """
+
+    def __init__(self, message: str = "") -> None:
+        self._extended_message = message
+
+    def __str__(self) -> str:
+        message = "Not enough memory"
```
mcgov, line 219: Ideally this would be more descriptive, I'm picturing seeing a bunch of tests failing with "Not enough memory" as the only failure message. Providing an error msg like "Not enough memory for requested operation: Requested XXX, Available: YYY" would be more useful.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -494,12 +423,12 @@ def verify_dpdk_build(
     log: Logger,
     variables: Dict[str, Any],
     pmd: str,
+    hugepage_size: HugePageSize,
     multiple_queues: bool = False,
-    gibibyte_hugepages: bool = False,
 ) -> DpdkTestResources:
     # setup and unwrap the resources for this test
     test_kit = initialize_node_resources(
```
mcgov: Do you need to try/catch here? Since not enough memory means we can't run the test, we need to catch NotEnoughMemoryException and raise it as a SkippedException somewhere. You can encapsulate Exceptions within each other. Idk is this the exact syntax but something along the lines of:
```
except NotEnoughMemoryException as err:
    raise SkippedException(err)
 ```
 
 It's a little wonky, some tests call initialize_node_resources directly so you might need to try/except in a few spots other than init_nodes_concurrent

mcgov:
Almost there! Last couple of changes to handle and then we're good.

microsoft/testsuites/dpdk/dpdkutil.py
kanchansenlaskar: Kindly clarify on why we shall skip if there is not memory? If there is not enough memory there has to be a test bug at least to perform an environment change or any other decision to be taken up.
I intentionally kept it like that so that the test fails and brings it to our notice.

lisa/util/__init__.py
kanchansenlaskar: _extended_message will take care of that, with a much more customized and better message

microsoft/testsuites/dpdk/dpdkutil.py
kanchansenlaskar: Like i mentioned in the previous reply, I have kept it intensionally like that so that the test actually fails with the reasoning

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: If there isn't enough memory, we're running on a system which doesn't support that test. We cannot determine if DPDK is working because the environment doesn't have enough resources to even start DPDK. There are a lot of sizes of VMs and machines which won't have enough memory for the test; that's alright. We just don't want to run the test on those. We **will** notice if all the DPDK tests get skipped with a NotEnoughMemory error. We don't want that polluting test results, because the test **did not fail**; rather, the test did not run enough to start DPDK.

mcgov: Same behavior with trying to run a given hugepage size test on a given platform. If the architecture doesn't support a certain hugepage size; we shouldn't fail the test. It's a skip. It's not a 'PASS' because we didn't test anything. It's not a FAIL either.

lisa/util/__init__.py
mcgov: Nice! You nailed it 😂

microsoft/testsuites/dpdk/dpdkutil.py
kanchansenlaskar: Are we sure that we won't come across a scenario where the setup actually supports yet doesn't have enough memory to carry out the operation?
Clarifying with you because this is a separate tool that we are writing here, and there can be tests with many scenarios.
Do let me know your thought on this question

kanchansenlaskar: discussed offline, made the change to skip it

microsoft/testsuites/dpdk/dpdkutil.py
kanchansenlaskar: same as above

LiliDeng (dismissed):

LiliDeng:
@kanchansenlaskar could you please run against `oracle oracle-linux ol93-arm64-lvm-gen2 9.3.2`

kanchansenlaskar:
> @kanchansenlaskar could you please run against `oracle oracle-linux ol93-arm64-lvm-gen2 9.3.2`

lisa.util.SkippedException: Unsupported system: 'Red Hat Enterprise Linux release 9.3 (Plow)'. This OS is not supported by the DPDK test suite for Azure.

kanchansenlaskar:
> @kanchansenlaskar could you please run against `oracle oracle-linux ol93-arm64-lvm-gen2 9.3.2`

perf_xdp_rx_drop_multithread_sriov - fails with and without my change. This is not using any of my libraries which I wrote/changed. This is some other testcase which isn't part of the PR which I am fixing.

mcgov (approved):
LGTM, if there's no test regressions I'm good with merging.

LiliDeng (approved):

PR #3324, opened 2024-06-24
remove dependency of storage account

LiliDeng:


squirrelsc:
What's the file `lisa/features/files/platform_mpi-09.01.04.03r-ce.bin` for? Can it be downloaded somewhere?

LiliDeng:
> What's the file `lisa/features/files/platform_mpi-09.01.04.03r-ce.bin` for? Can it be downloaded somewhere?

It is an installation file for mpirun.
I didn't find a public place for it, seems it needs login into ibm website then download. So I used this way to get rid of storage account.

refer 
image

squirrelsc:
> > What's the file `lisa/features/files/platform_mpi-09.01.04.03r-ce.bin` for? Can it be downloaded somewhere?
> 
> It is an installation file for mpirun. I didn't find a public place for it, seems it needs login into ibm website then download. So I used this way to get rid of storage account.
> 

Please make the file url as an argument, if the url is not provided, skip this test case.

squirrelsc (approved):

PR #3329, opened 2024-07-02
Add subscription in vhdSchema to support that vhd is in another subsc…

lubaihua33:
Add subscription in vhdSchema to support that vhd is in another subscription.

This PR doesn't contain the scenario Sig transformer. It will in another PR.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -520,6 +520,7 @@ class SharedImageGallerySchema(AzureImageSchema):
 class VhdSchema(AzureImageSchema):
     vhd_path: str = ""
     vmgs_path: Optional[str] = None
+    subscription: str = ""
 
     def load_from_platform(self, platform: "AzurePlatform") -> None:
         return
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -2225,9 +2236,12 @@ def get_deployable_vhd_path(
     """
     matches = SAS_URL_PATTERN.match(vhd_path)
     if not matches:
-        vhd_details = get_vhd_details(platform, vhd_path)
+        vhd_details = get_vhd_details(platform, vhd_path, vhd_subscription)
         vhd_location = vhd_details["location"]
-        if location == vhd_location:
+        if (
+            location == vhd_location
+            and vhd_details["subscription"] == platform.subscription_id
```
LiliDeng, line 2242: vhd_details may not contain subscription

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2216,7 +2223,11 @@ def _generate_sas_token_for_vhd(
 
 @lru_cache(maxsize=10)  # noqa: B019
 def get_deployable_vhd_path(
-    platform: "AzurePlatform", vhd_path: str, location: str, log: Logger
+    platform: "AzurePlatform",
+    vhd_path: str,
+    location: str,
+    log: Logger,
+    vhd_subscription: str = "",
```
squirrelsc: name it `subscription`.

lisa/sut_orchestrator/azure/common.py
lubaihua33: Have added.

lisa/sut_orchestrator/azure/common.py

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3330, opened 2024-07-02
Add Property maximize_feature_capability

adityagesh:
maximize_feature_capability can accept coma seperated strings. This is used to enable a feature capability

### DIFF
--- a/docs/run_test/platform.rst
+++ b/docs/run_test/platform.rst
@@ -154,6 +154,7 @@ deployment.
             location: "<one or multiple locations, split by comma>"
             vm_size: "<vm size>"
             maximize_capability: "<true or false>"
+            maximize_feature_capability: "<FeatureName1>,<FeatureName2>"
             osdisk_size_in_gb: <disk size in gb>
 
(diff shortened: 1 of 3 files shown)


mcgov (commented):
I love this option, though this option might be reduant. Have you tested whether ignore_capability satifies this requirement? 

Less important, but I have always hated the nomenclature of 'maximize_capability'. We're overriding LISA's VM feature metadata when resolving whether an environment satisfies a test's requirements. We're not really maximizing any value, it's ignoring things. Maybe it's just me. 
Maybe a name like `override_capability_metadata` or `force_enable_capability' or something that makes it clear that we're actually ignoring something or changing the way we evaluate an operation.
No code comments other than these nit picks, it looks great 👍

adityagesh:
Thanks for reviewing @mcgov 
ignore_capability does not satisfy the requirement. We can use it to run the test case, but the feature will not be enabled.

I agree with your thoughts on the naming, maximize_capability is indeed confusion. I named this similar to maximize_capability since it serve very similar purpose.

However, I agree that we should rename to something like 'force_enable_feature'. I will rename it before merging this patch

PR #3331, opened 2024-07-02
Add PCI passthrough support for cloud-hypervisor platform

smit-gardhariya:
Add PCI passthrough support for cloud-hypervisor platform
Add parameter to pass options while loading module using modprobe

### DIFF
--- a/lisa/sut_orchestrator/libvirt/context.py
+++ b/lisa/sut_orchestrator/libvirt/context.py
@@ -7,7 +7,7 @@ from lisa.environment import Environment
 from lisa.node import Node
 
 from .console_logger import QemuConsoleLogger
-from .schema import DiskImageFormat
+from .schema import DiskImageFormat, LibvirtHostPciDeviceSchema
 
 
 @dataclass
@@ -56,6 +56,12 @@ class NodeContext:
 
     console_logger: Optional[QemuConsoleLogger] = None
     domain: Optional[libvirt.virDomain] = None
+
+    # Device pass through configuration
+    is_device_passthrough_set: bool = False
+    device_passthrough_configs: List[LibvirtHostPciDeviceSchema] = field(
+        default_factory=list,
+    )
 
 
 def get_environment_context(environment: Environment) -> EnvironmentContext:
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -34,7 +34,9 @@ from lisa.tools import (
     Iptables,
     Journalctl,
     Ls,
+    Lspci,
     Mkdir,
+    Modprobe,
     QemuImg,
     Sed,
     Service,
@@ -60,6 +62,7 @@ from .schema import (
     BaseLibvirtNodeSchema,
     BaseLibvirtPlatformSchema,
     DiskImageFormat,
+    LibvirtHostPciDeviceSchema,
 )
 from .serial_console import SerialConsole
 from .start_stop import StartStop
@@ -503,6 +506,13 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
 
                 node_context.data_disks.append(data_disk)
 
+        # If Device_passthrough is set in runbook,
+        # Configure device passthrough params for node_context
+        self._configure_device_passthrough(
+            node_runbook.device_passthrough,
+            node_context,
+        )
+
     def restart_domain_and_attach_logger(self, node: Node) -> None:
         node_context = get_node_context(node)
         domain = node_context.domain
@@ -942,6 +952,8 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         on_crash.text = "destroy"
 
         devices = ET.SubElement(domain, "devices")
+        if node_context.is_device_passthrough_set:
+            devices = self._add_device_passthrough(devices, node_context)
 
         serial = ET.SubElement(devices, "serial")
         serial.attrib["type"] = "pty"
@@ -1341,3 +1353,104 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
             )
             with open(str(libvirt_log_local_path), "w") as f:
                 f.write(libvirt_log)
+
+    def _configure_device_passthrough(
+        self,
+        device_configs: Optional[List[LibvirtHostPciDeviceSchema]],
+        node_context: NodeContext,
+    ) -> None:
+        if device_configs:
+            for config in device_configs:
+                if self._device_exists(config):
+                    if self._is_kernel_driver_vfio_pci(config):
+                        raise LisaException(
+                            "Kernel Driver associated to Device already is 'vfio-pci'"
+                            f", Device: {config}"
+                        )
+                    node_context.device_passthrough_configs.append(config)
+                else:
+                    raise LisaException(f"Device not found on host: {config}")
+
+            self._check_passthrough_support(self.host_node)
+
+            node_context.is_device_passthrough_set = True
+            modprobe = self.host_node.tools[Modprobe]
+            allow_unsafe_interrupt = modprobe.load(
+                modules="vfio_iommu_type1",
+                options="allow_unsafe_interrupts=1",
+            )
+            if not allow_unsafe_interrupt:
+                raise LisaException("Allowing unsafe interrupt failed")
+
+    def _add_device_passthrough(
+        self,
+        devices: ET.Element,
+        node_context: NodeContext,
+    ) -> ET.Element:
+        for config in node_context.device_passthrough_configs:
+            hostdev = ET.SubElement(devices, "hostdev")
+            hostdev.attrib["mode"] = "subsystem"
+
+            assert config.device_type
+            hostdev.attrib["type"] = config.device_type
+
+            assert config.managed
+            hostdev.attrib["managed"] = config.managed
+            if config.device_type == "pci":
+                source = ET.SubElement(hostdev, "source")
+                src_addrs = ET.SubElement(source, "address")
+
+                source_domain = config.host_domain
+                assert source_domain
+                src_addrs.attrib["domain"] = f"0x{source_domain}"
+
+                source_bus = config.host_bus
+                assert source_bus
+                src_addrs.attrib["bus"] = f"0x{source_bus}"
+
+                source_slot = config.host_slot
+                assert source_slot
+                src_addrs.attrib["slot"] = f"0x{source_slot}"
+
+                source_function = config.host_function
+                assert source_function
+                src_addrs.attrib["function"] = f"0x{source_function}"
+
+                driver = ET.SubElement(hostdev, "driver")
+                driver.attrib["name"] = "vfio"
+        return devices
+
+    def _device_exists(self, device_config: LibvirtHostPciDeviceSchema) -> bool:
+        devices_list = self.host_node.tools[Lspci].get_devices()
+        devices_slots = [x.slot for x in devices_list]
+        if device_config.device_type == "pci":
+            bus = device_config.host_bus
+            slot = device_config.host_slot
+            fn = device_config.host_function
+            device_address = f"{bus}:{slot}.{fn}"
+            if device_address in devices_slots:
+                return True
+        return False
+
+    def _check_passthrough_support(self, node: Node) -> None:
+        ls = node.tools[Ls]
+        path = "/dev/vfio/vfio"
+        err = "Host does not support IOMMU"
+        if not ls.path_exists(path=path, sudo=True):
+            raise LisaException(f"{err} : {path} does not exist")
+
+        path = "/sys/kernel/iommu_groups/"
+        if len(ls.list(path=path, sudo=True)) == 0:
+            raise LisaException(f"{err} : {path} does not have any entry")
+
+    def _is_kernel_driver_vfio_pci(
+        self,
+        device_config: LibvirtHostPciDeviceSchema,
+    ) -> bool:
+        lspci = self.host_node.tools[Lspci]
+        bus = device_config.host_bus
+        slot = device_config.host_slot
+        fn = device_config.host_function
+        device_address = f"{bus}:{slot}.{fn}"
+        kernel_module = lspci.get_used_module(device_address)
+        return kernel_module == "vfio-pci"
--- a/lisa/sut_orchestrator/libvirt/schema.py
+++ b/lisa/sut_orchestrator/libvirt/schema.py
@@ -52,6 +52,26 @@ class BaseLibvirtPlatformSchema:
     capture_libvirt_debug_logs: bool = False
 
 
+# Configuration options for device-passthrough for the VM.
+@dataclass_json()
+@dataclass
+class LibvirtHostPciDeviceSchema:
+    # This property determines if libvirt would bind/unbind native driver or not
+    # Default is set to "yes" means libvirt will detach the native device driver and
+    # attach it to vfio-pci when vm is create and attach it back on deletion of vm
+    managed: str = "yes"
+
+    # Type of the device. Default is set to "pci"
+    device_type: str = "pci"
+
+    # Host device details for which we want to perform device-passthrough
+    # we can get it using lspci command
+    host_domain: str = ""
+    host_bus: str = ""
+    host_slot: str = ""
+    host_function: str = ""
+
+
 # Possible disk image formats
 class DiskImageFormat(Enum):
     QCOW2 = "qcow2"
@@ -84,6 +104,9 @@ class BaseLibvirtNodeSchema:
     machine_type: str = ""
     # Whether to enable secure boot.
     enable_secure_boot: bool = False
+
+    # Configuration options for device-passthrough.
+    device_passthrough: Optional[List[LibvirtHostPciDeviceSchema]] = None
 
 
 # QEMU orchestrator's per-node configuration options.
--- a/lisa/sut_orchestrator/libvirt/ch_platform.py
+++ b/lisa/sut_orchestrator/libvirt/ch_platform.py
@@ -118,6 +118,8 @@ class CloudHypervisorPlatform(BaseLibvirtPlatform):
         os_kernel.text = node_context.firmware_path
 
         devices = ET.SubElement(domain, "devices")
+        if node_context.is_device_passthrough_set:
+            devices = self._add_device_passthrough(devices, node_context)
 
         console = ET.SubElement(devices, "console")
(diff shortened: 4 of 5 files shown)


lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -396,7 +398,7 @@ def _configure_nodes(self, environment: Environment, log: Logger) -> None:
                 vm_name_prefix,
             )
 
-    def _configure_node(
+    def _configure_node(  # noqa: C901
```
anirudhrb: What warning does this suppress and why?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1378,73 @@ def _capture_libvirt_logs(self) -> None:
...
+        if device_config.get("device_type") == "pci":
+            bus = device_config.get("source_bus")
+            slot = device_config.get("source_slot")
+            fn = device_config.get("source_function")
+            device_address = f"{bus}:{slot}.{fn}"
+            if device_address in devices_slots:
+                return True
+        return False
+
+    def _check_if_host_support_iommu(self, node: Node) -> None:
```
anirudhrb:
```suggestion
    def _is_passthrough_supported(self, node: Node) -> None:
```

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1378,73 @@ def _capture_libvirt_logs(self) -> None:
...
+
+                source_function = config.get("source_function", "")
+                assert source_function
+                src_addrs.attrib["function"] = f"0x{source_function}"
+
+                driver = ET.SubElement(hostdev, "driver")
+                driver.attrib["name"] = "vfio"
+        return devices
+
+    def _check_passthrough_device_exist(self, device_config: Dict[str, str]) -> bool:
```
anirudhrb:
```suggestion
    def _device_exists(self, device_config: Dict[str, str]) -> bool:
```

This just checks if the specified device exists on the system, right? So, the name can be more concise.

squirrelsc (dismissed):

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -503,6 +505,39 @@ def _configure_node(
 
                 node_context.data_disks.append(data_disk)
 
+        # Set device passthrough params for node context
+        if node_runbook.device_passthrough:
+            self._check_if_host_support_iommu(self.host_node)
+            dev_conf_list = node_runbook.device_passthrough
```
LiliDeng: is it `dev_conf_list = node_runbook.device_passthrough_configs`

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Because of lot of if/else in this function, it will throw `too complex function error` which is C901 for flak8. There is no better way to handle this as those if/else are needed.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: There is no `device_passthrough_configs `in the runbook schema.

```
# Configuration options for device-passthrough.
device_passthrough: Optional[List[LibvirtHostPciDeviceSchema]] = None
```

This is the parent attribute for runbook yaml to pass list of devices

lisa/sut_orchestrator/libvirt/platform.py

lisa/sut_orchestrator/libvirt/platform.py

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: `device_passthrough_configs` is something we set for node_context

squirrelsc (dismissed):

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -52,6 +52,26 @@ class BaseLibvirtPlatformSchema:
...
 
 
+# Configuration options for device-passthrough for the VM.
+@dataclass_json()
+@dataclass
+class LibvirtHostPciDeviceSchema:
+    # This property determines if libvirt would bind/unbind native driver or not
+    # Default is set to "yes" means libvirt will detach the native device driver and
+    # attach it to vfio-pci when vm is create and attach it back on deletion of vm
+    managed_passthrough: str = "yes"
```
pupacha: I think `managed` is good enough here, since we know this class (schema) itself is for passthrough device.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -52,6 +52,26 @@ class BaseLibvirtPlatformSchema:
...
+
+    # Type of the device. Default is set to "pci"
+    device_type: str = "pci"
+
+    # Host device details for which we want to perform device-passthrough
+    # we can get it using lspci command
+    source_domain: str = ""
```
pupacha: I dont think we need the `source_` prefix here

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1378,73 @@ def _capture_libvirt_logs(self) -> None:
...
+        if device_config.get("device_type") == "pci":
+            bus = device_config.get("source_bus")
+            slot = device_config.get("source_slot")
+            fn = device_config.get("source_function")
+            device_address = f"{bus}:{slot}.{fn}"
+            if device_address in devices_slots:
+                return True
+        return False
+
+    def _is_passthrough_supported(self, node: Node) -> None:
```
pupacha: Since the function does not return bool, better naming would be something like `_check_passthrough_support`

lisa/sut_orchestrator/libvirt/context.py
```diff
@@ -57,6 +57,10 @@ class NodeContext:
     console_logger: Optional[QemuConsoleLogger] = None
     domain: Optional[libvirt.virDomain] = None
 
+    # Device pass through configuration
+    is_device_passthrough_set: bool = False
+    device_passthrough_configs: List[Dict[str, str]] = field(default_factory=list)
```
pupacha: Can we make `device_passthrough_configs` type to `List[LibvirtHostPciDeviceSchema]` ? Since `LibvirtHostPciDeviceSchema` is already available.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -503,6 +505,39 @@ def _configure_node(
...
+
+            for config in dev_conf_list:
+                device_config = {
+                    "managed_passthrough": config.managed_passthrough,
+                    "device_type": config.device_type,
+                    "source_domain": config.source_domain,
+                    "source_bus": config.source_bus,
+                    "source_slot": config.source_slot,
+                    "source_function": config.source_function,
+                }
```
pupacha: This conversion to dict is not really needed. We can continue to access the members of `LibvirtHostPciDeviceSchema` directly. Instead of dict, LibvirtHostPciDeviceSchema needs to be passed as function arguments for subsequent fn calls.

lisa/sut_orchestrator/libvirt/platform.py
pupacha: In that case, we can refactor the new code to a new function for example `_configure_node_passthrough_devices` and call this fn from `_config_node`

pupacha (changes requested):

lisa/sut_orchestrator/libvirt/schema.py
smit-gardhariya: these params are also applicable to guest PCI device. Those are optional to pass in libvirt. There will be 2 `address `element for `hostdev`. One under `source `( for host device) and other is for guest vm. So keeping **source_** seems fine to me

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Still one if condition we have to add like below snippet, and this would cause the same error of C901

```
# Set device passthrough params for node context
        if node_runbook.device_passthrough:
            self._configure_device_passthrough(self, node_runbook, node_context)
```

lisa/sut_orchestrator/libvirt/schema.py
smit-gardhariya: done

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: done

lisa/sut_orchestrator/libvirt/context.py
smit-gardhariya: My though was to use JSON so we can add extra params like device address so we don't have to construct every time. Schema would tie the hands to manipulate config object. Anyway, there are only 2 occurrences for constructing the device address for LSPCI checks so i restructured it with schema as of now.

lisa/sut_orchestrator/libvirt/platform.py

lisa/sut_orchestrator/libvirt/platform.py
pupacha: Does it still throw the same error if we move the condition to `_configure_device_passthrough` itself ?

smit-gardhariya: Yes, that is what i mentioned above. Also, condition has to be there in **_configure_node** as I don't want to call the function every time. This should be called when there is passthrough params set from runbook.

lisa/sut_orchestrator/libvirt/schema.py
pupacha: Could you provide a sample xml to understand more? 
Does the other guest related `address` also have same domain, bus, slot, function fields?

smit-gardhariya: No, it is not same. Host/Guest params can be different

```
<domain type="ch">
	<name>lisa-OULKV-0</name>
	<memory unit="MiB">1024</memory>
	<vcpu>4</vcpu>
	<os>
		<type>hvm</type>
		<kernel>/home/cloud/device_passthrough_libvirt/vmlinux.bin</kernel>
		<cmdline>root=/dev/vda1 console=hvc0 rw systemd.journald.forward_to_console=1</cmdline>
	</os>
	<devices>
		<hostdev mode='subsystem' type='pci' managed='yes'>
			<source>
				<address domain='0x0000' bus='0x19' slot='0x00' function='0x0'/>
			</source>
			<address type='pci' domain='0x0000' bus='0x05' slot='0x00' function='0x0'/>
			<driver name='vfio'/>
		</hostdev>
		<console type="pty">
			<target type="serial" port="0" />
		</console>
		<interface type="network">
			<source network="default" />
			<model type="virtio" />
			<driver queues="4" iommu="off" />
		</interface>
		<disk type="file">
			<driver if="virtio" type="raw" queues="4" />
			<target dev="vda" />
			<source file="/home/cloud/device_passthrough_libvirt/focal-server-cloudimg-amd64.raw" />
		</disk>
		<disk type="file">
			<driver if="virtio" type="raw" queues="4" />
			<target dev="vdb" />
			<source file="/home/cloud/device_passthrough_libvirt/ubuntu-cloudinit.img" />
		</disk>
	</devices>
</domain>
```

lisa/sut_orchestrator/libvirt/platform.py
pupacha: Since there would be condition in the function, there would be no harm in calling it everytime. I think it's worth giving it a try, atleast to see if this will eliminate format check error.

smit-gardhariya: That condition if function is to make mypy python 3.8 compliance or it was failing. I would not recommend to call the device passthrough config fn every time

lisa/sut_orchestrator/libvirt/schema.py
pupacha: Thanks, Smit, for helping me understand the schema in detail. 

If we also plan to extend `LibvirtHostPciDeviceSchema` to specify the other guest specific `address` as well, then a better schema would be to have a new class called `PCIDeviceSchema` that will have domain, bus, slot, function. And have say `source_device` and `guest_device` in `LibvirtHostPciDeviceSchema`
```
class PCIDeviceSchema:
  domain: str
  bus: str
  slot: str
  function: str

class LibvirtHostPciDeviceSchema:
  .
  .
  .
  source_device: PCIDeviceSchema
  guest_device: PCIDeviceSchema

```

lisa/sut_orchestrator/libvirt/platform.py
pupacha: As I said earlier, if the following condition is inside the function, there is no harm in calling it every time. 
`if node_runbook.device_passthrough:`

`  # noqa: C901` has to the last option if nothing works. And I see the option I suggested still open.

smit-gardhariya: Ideally, we should skip the redundant call to function. Anyway, i made the change please check

lisa/sut_orchestrator/libvirt/schema.py
smit-gardhariya: Totally Agree. As discussed, this will make guest_device config mandatory to be passed from runbook. We can make it optional but we can extend the platform whenever that need will come (Say some post domain creation testcase which check the LSPCI output for guest to verify). Idea with this PR is to enable PCI passthrough first.

lisa/sut_orchestrator/libvirt/platform.py
anirudhrb: > There is no better way to handle this as those if/else are needed

If we make the function smaller by extracting bits out into functions, this error will go away, right?

smit-gardhariya: > > There is no better way to handle this as those if/else are needed
> 
> If we make the function smaller by extracting bits out into functions, this error will go away, right?

Even if i would put one if condition, it was coming. I wanted to call fn only when needed and avoid redundant call even when device pass through is not required.

I removed that condition also now. Please check the latest push

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1353,103 @@ def _capture_libvirt_logs(self) -> None:
...
                 f.write(libvirt_log)
+
+    def _configure_device_passthrough(
+        self,
+        device_configs: Optional[List[LibvirtHostPciDeviceSchema]],
+        node_context: NodeContext,
+    ) -> None:
+        self._check_passthrough_support(self.host_node)
+        if device_configs:
+            for config in device_configs:
```
pupacha:
```suggestion
        if device_configs:
            self._check_passthrough_support(self.host_node)
            for config in device_configs:
```
check for passthrough support has to be done only after `device_configs` check

lisa/sut_orchestrator/libvirt/platform.py
pupacha: Thanks for this change. Looks like we got rid of the flake8 error now.

smit-gardhariya: Yes that was because of IF we added. Now we removed it and calling fn every time

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: done.

anirudhrb:
Could you please post a sample of how a runbook with passthrough configuration would look?

smit-gardhariya:
> Could you please post a sample of how a runbook with passthrough configuration would look?

`
requirement:
      core_count: $(guest_vcpu)
      memory_mb: $(guest_memory)
      cloud-hypervisor:
        disk_img: $(qcow2)
        disk_img_format: "qcow2"
        firmware: $(firmware_path)
        cloud_init:
        device_passthrough:
          - managed: "yes"
            device_type: "pci"
            host_domain: "0000"
            host_bus: "19"
            host_slot: "00"
            host_function: "0"
          - managed: "yes"
            device_type: "pci"
            host_domain: "0000"
            host_bus: "19"
            host_slot: "00"
            host_function: "1"
`

This would remain same for QEMU/CLH. As of now we are focusing on CLH only. QEMU might require some additional change.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -52,6 +52,26 @@ class BaseLibvirtPlatformSchema:
     capture_libvirt_debug_logs: bool = False
 
 
+# Configuration options for device-passthrough for the VM.
+@dataclass_json()
+@dataclass
+class LibvirtHostPciDeviceSchema:
```
pupacha, line 58: Thanks for addressing all the comments, Smit.

I have another comment in general about the approach.

As I understand it, `device_passthrough` is under node requirements. While this approach works with deploying of a single node, this fails with cases where multiple nodes with passthrough is needed. For example, all network tests will create multiple nodes. And by the current code logic, the deployment of second will fail since all pci devices provided in runbook would be assigned to first node itself.

PCI devices are associated with the host node. And in our context, host node is synonymous to platform. So, we should associate the host pci devices to platform than to node requirement.

Here is what I'm thinking in the form of a sample schema. Introduce 'device_pools' at platform level and define all the host devices there. Pools can be of different types like 'pci_net', 'pci_gpu' etc. All the devices that we want to be used in provisioning guests need to be defined here. And for a node/VM to use a passthrough device, the pool type and devices count has to be mentioned in the node requirement. Platform needs to handle assigning the devices accordingly from the available devices of the pool. Platform needs to maintain the state of the devices (which vm a device is assigned to or not, etc).

```
platform:
  - type: cloud-hypervisor
    cloud-hypervisor:
      network_boot_timeout: $(boot_timeout)
      hosts:
        - address: $(host_address)
          username: $(host_username)
      capture_libvirt_debug_logs: true
      device_pools:
        - type: "pci_net"
           devices:
             - domain: "1"
                bus: "1"
                slot: "1"
                function: "1"
             - domain: "2"
                bus: "2"
                slot: "2"
                function: "2"
        - type: "pci_gpu"
           devices:
             - domain: "3"
                bus: "3"
                slot: "3"
                function: "3"
             - domain: "4"
                bus: "4"
                slot: "4"
                function: "4"
    requirement:
      core_count: $(guest_vcpu)
      memory_mb: $(guest_memory)
      cloud-hypervisor:
        disk_img: $(qcow2)
        disk_img_format: "qcow2"
        cloud_init:
        device_passthrough:
          - host_device_pool: "pci_net" 
            managed: "yes"
            count: 2
```
I haven't put too much thought into schema. What I listed above is the rough idea. I've also discussed this with @anirudhrb . We can brainstorm more on this offline and finalize the schema.

PR #3334, opened 2024-07-02
StartStop use_public_address

kamalca:
The StartStop feature should honor the `use_public_address` settings in the runbook.


squirrelsc (approved):

mcgov:
Verified fix :)

PR #3335, opened 2024-07-03
updated msg_level test case with an unsupported_os

r-dailey:
This test case is not supported on BSD and needs to be skipped and the supported function is also not supported by the os.

### DIFF
--- a/microsoft/testsuites/network/networksettings.py
+++ b/microsoft/testsuites/network/networksettings.py
@@ -484,8 +484,14 @@ class NetworkSettings(TestSuite):
                 and name.
             3. Validate changing the message level flag setting.
             4. Revert back the setting to original value.
+
+            Note: BSD does not support the feature tested here and
+            lacks the hv_netvsc module used to support it.
         """,
         priority=2,
+        requirement=simple_requirement(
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/network/networksettings.py
```diff
@@ -485,7 +485,12 @@ def verify_device_rx_hash_level_change(self, node: Node, log: Logger) -> None:
             3. Validate changing the message level flag setting.
             4. Revert back the setting to original value.
         """,
-        priority=2,
+        priority=2,        
+        # BSD does not support the feature listed here and
+        # lacks the module checked in the support fucntion.
```
squirrelsc: 1. Move it to above description, so it can show with skipped message. 
2. Specify which module is not in BSD.

squirrelsc:
Please merge the second commit "Moved comment to description block" into previous one, because it's not a meaningful step for this PR, when it's merged into main branch. From the perspective of code history, we don't need the history of code reviews.

squirrelsc:
Please also fix failed checks.

squirrelsc (approved):

PR #3336, opened 2024-07-03
add install from artifact support for ltp

paxue:
add "ltp_file_path" var to support ltp install from downloaded artifact

### DIFF
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -1,6 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-
+import os
 import re
 from dataclasses import dataclass
 from pathlib import PurePath, PurePosixPath
@@ -24,9 +24,11 @@ from lisa.tools import (
     Make,
     Mkdir,
     Pgrep,
+    RemoteCopy,
     Rm,
     Swap,
     Sysctl,
+    Tar,
 )
 from lisa.util import LisaException, find_patterns_in_lines
 
@@ -71,10 +73,11 @@ class Ltp(Tool):
     def can_install(self) -> bool:
         return True
 
-    def __init__(self, node: Node, *args: Any, **kwargs: Any) -> None:
+    def __init__(self, node: Node, source_file: str, *args: Any, **kwargs: Any) -> None:
         super().__init__(node, args, kwargs)
         git_tag = kwargs.get("git_tag", "")
         self._git_tag = git_tag if git_tag else self.DEFAULT_LTP_TESTS_GIT_TAG
+        self._source_file = source_file
 
     def run_test(
         self,
@@ -340,14 +343,29 @@ class Ltp(Tool):
         self.node.tools[Mkdir].create_directory(top_src_dir, sudo=True)
         self.node.tools[Chmod].update_folder(top_src_dir, "a+rwX", sudo=True)
 
-        # clone ltp
-        git = self.node.tools[Git]
-        ltp_path = git.clone(
-            self.LTP_GIT_URL, cwd=PurePosixPath(top_src_dir), dir_name=top_src_dir
-        )
+        if self._source_file:
+            self._log.debug(
+                f"Use downloaded source code tar file: {self._source_file}!"
+            )
+            remote_source_file = self.get_tool_path(use_global=True) / os.path.basename(
+                self._source_file
+            )
 
-        # checkout tag
-        git.checkout(ref=f"tags/{self._git_tag}", cwd=ltp_path)
+            copy = self.node.tools[RemoteCopy]
+            copy.copy_to_remote(PurePath(self._source_file), remote_source_file)
+            self.node.tools[Tar].extract(
+                str(remote_source_file), top_src_dir, sudo=True
+            )
+            ltp_path = self.node.get_pure_path(top_src_dir)
+        else:
+            # clone ltp
+            git = self.node.tools[Git]
+            ltp_path = git.clone(
+                self.LTP_GIT_URL, cwd=PurePosixPath(top_src_dir), dir_name=top_src_dir
+            )
+
+            # checkout tag
+            git.checkout(ref=f"tags/{self._git_tag}", cwd=ltp_path)
 
         # build ltp in /opt/ltp since this path is used by some
         # tests, e.g, block_dev test
--- a/microsoft/testsuites/ltp/ltpsuite.py
+++ b/microsoft/testsuites/ltp/ltpsuite.py
@@ -35,6 +35,13 @@ class LtpTestsuite(TestSuite):
     @TestCaseMetadata(
         description="""
         This test case will run Ltp lite tests.
+        1. When ltp_source_file (downloaded ltp code) is specified in .yml,
+        case will use it to extract the tar and run ltp, instead of downloading runtime.
+        Example:
+        - name: ltp_source_file
+          value: <path_to_ltp.tar.xz>
+          is_case_visible: true
+        2. When ltp_source_file not in .yml, clone github with ltp_tests_git_tag
         """,
         priority=3,
         timeout=_TIME_OUT,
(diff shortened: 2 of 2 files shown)


microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -55,6 +61,7 @@ def verify_ltp_lite(
         result: TestResult,
     ) -> None:
         # parse variables
+        file_path = variables.get("ltp_file_path", "")
```
squirrelsc: 1. Call it like `ltp_source_file`.
2. Explain the variable is used to use saved source code, instead of downloading runtime.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -340,14 +344,29 @@ def _install(self) -> bool:
         self.node.tools[Mkdir].create_directory(top_src_dir, sudo=True)
         self.node.tools[Chmod].update_folder(top_src_dir, "a+rwX", sudo=True)
 
-        # clone ltp
-        git = self.node.tools[Git]
-        ltp_path = git.clone(
-            self.LTP_GIT_URL, cwd=PurePosixPath(top_src_dir), dir_name=top_src_dir
-        )
+        if self._tar_file_path:
+            self._log.debug(f"Use downloaded ltp tar: {self._tar_file_path}!")
```
squirrelsc: "downloaded" to "prebuilt".

microsoft/testsuites/ltp/ltp.py
```diff
@@ -340,14 +344,29 @@ def _install(self) -> bool:
...
+            self._remote_tar_path = self.get_tool_path(
+                use_global=True
+            ) / os.path.basename(self._tar_file_path)
+
+            mkdir = self.node.tools[Mkdir]
+            mkdir.create_directory(self._remote_tar_path.parent.as_posix())
+            self.node.shell.copy(PurePath(self._tar_file_path), self._remote_tar_path)
```
squirrelsc: It doesn't need to copy, just extract it to the target place.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -340,14 +344,29 @@ def _install(self) -> bool:
...
+            self.node.tools[Tar].extract(
+                str(self._remote_tar_path), top_src_dir, sudo=True
```
squirrelsc: don't extract it to the src dir. The src dir is to find a free partition for build, but for the prebuilt package, the system disk should be enough. Use the global tool path.

microsoft/testsuites/ltp/ltp.py
paxue: as discussed, this is source code, not prebuilt one

microsoft/testsuites/ltp/ltp.py
paxue: Do you mean direct extract from local env to remote node? That may cause lots of extracted files transfering which is very time consuming.

microsoft/testsuites/ltp/ltp.py
paxue: This should be good now as it's source code.

microsoft/testsuites/ltp/ltp.py
squirrelsc: ok, but use the tool called RemoteCopy

microsoft/testsuites/ltp/ltp.py
```diff
@@ -71,10 +73,13 @@ def dependencies(self) -> List[Type[Tool]]:
...
         return True
 
-    def __init__(self, node: Node, *args: Any, **kwargs: Any) -> None:
+    def __init__(
+        self, node: Node, source_file_path: str, *args: Any, **kwargs: Any
+    ) -> None:
         super().__init__(node, args, kwargs)
         git_tag = kwargs.get("git_tag", "")
         self._git_tag = git_tag if git_tag else self.DEFAULT_LTP_TESTS_GIT_TAG
+        self._tar_file_path = source_file_path
```
squirrelsc: 1. The variable is not used out of this method, so it doesn't need to be saved in the class. Use a local variable.
1. The variable should be similar, not lost it's meaning, like `self._source_file`.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -85,7 +93,9 @@ def verify_ltp_lite(
                 )
 
         # run ltp lite tests
-        ltp: Ltp = node.tools.get(Ltp, git_tag=ltp_tests_git_tag)
+        ltp: Ltp = node.tools.get(
+            Ltp, git_tag=ltp_tests_git_tag, source_file_path=ltp_source_file
```
squirrelsc: The variable name should be `source_file` instead of `source_file_path`, keep them consistent.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -340,14 +345,31 @@ def _install(self) -> bool:
         self.node.tools[Mkdir].create_directory(top_src_dir, sudo=True)
         self.node.tools[Chmod].update_folder(top_src_dir, "a+rwX", sudo=True)
 
-        # clone ltp
-        git = self.node.tools[Git]
-        ltp_path = git.clone(
-            self.LTP_GIT_URL, cwd=PurePosixPath(top_src_dir), dir_name=top_src_dir
-        )
+        if self._tar_file_path:
+            self._log.debug(f"Use downloaded ltp tar: {self._tar_file_path}!")
```
squirrelsc: The comments is not clear, it doesn't specify the downloaded is source code of ltp. Update it like "Use downloaded source code tar file: "

microsoft/testsuites/ltp/ltp.py
```diff
@@ -340,14 +345,31 @@ def _install(self) -> bool:
...
+            self._remote_tar_path = self.get_tool_path(
+                use_global=True
+            ) / os.path.basename(self._tar_file_path)
+
+            mkdir = self.node.tools[Mkdir]
+            mkdir.create_directory(self._remote_tar_path.parent.as_posix())
```
squirrelsc: The remotecopy should create the path inside, so it doesn't need to create before copy. If it's not, please support it in remotecopy.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -340,14 +345,31 @@ def _install(self) -> bool:
...
         self.node.tools[Chmod].update_folder(top_src_dir, "a+rwX", sudo=True)
 
-        # clone ltp
-        git = self.node.tools[Git]
-        ltp_path = git.clone(
-            self.LTP_GIT_URL, cwd=PurePosixPath(top_src_dir), dir_name=top_src_dir
-        )
+        if self._tar_file_path:
+            self._log.debug(f"Use downloaded ltp tar: {self._tar_file_path}!")
+            self._remote_tar_path = self.get_tool_path(
```
squirrelsc: 1. The `_remote_tar_path` is not used in other method, it doesn't need to be a class variable.
2. The name should be like `remote_source_file`.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -340,14 +345,31 @@ def _install(self) -> bool:
...
+            mkdir = self.node.tools[Mkdir]
+            mkdir.create_directory(self._remote_tar_path.parent.as_posix())
+            copy = self.node.tools[RemoteCopy]
+            copy.copy_to_remote(PurePath(self._tar_file_path), self._remote_tar_path)
+
+            self.node.tools[Tar].extract(
+                str(self._remote_tar_path), top_src_dir, sudo=True
+            )
+            ltp_path = self.node.get_pure_path(top_src_dir)
+            self._log.debug(f"Extracted tar from path {self._remote_tar_path}!")
```
squirrelsc: This line of log is not needed, because the tar tool should output it.

microsoft/testsuites/ltp/ltp.py
paxue: I does not quite understand your 1st commet. The "source_file_path" will be parsed in from outside "verify_ltp_lite" in ltpsuite.py file. There should be a var "self._source_file" to take this value in init() and then doing the condition when _install()

squirrelsc (dismissed):

squirrelsc (approved):

PR #3338, opened 2024-07-04
Add method to increase os partition for Redhat LVM

adityagesh:
Newer Redhat Distro uses LVM, Hence using osdisk_size increases the disk size but not the os partition size. This method increases the root partition size

### DIFF
--- a/microsoft/testsuites/power/common.py
+++ b/microsoft/testsuites/power/common.py
@@ -24,7 +24,7 @@ from lisa.util import (
     LisaException,
     SkippedException,
     UnsupportedDistroException,
-    find_groups_in_lines,
+    find_group_in_lines,
 )
 from lisa.util.perf_timer import create_timer
 
@@ -179,13 +179,13 @@ def _expand_os_partition(node: Node, log: Logger) -> None:
         # The output of pvscan -s is like below.:
         #  /dev/sda4
         #  Total: 1 [299.31 GiB] / in use: 1 [299.31 GiB] / in no VG: 0 [0   ]
-        pattern = re.compile(r"(?P<disk>.*)(?P<number>[\d]+)", re.M)
-        matched = find_groups_in_lines(pv_result, pattern)
+        pattern = re.compile(r"(?P<disk>.*)(?P<number>[\d]+)$", re.M)
+        matched = find_group_in_lines(pv_result, pattern)
         if not matched:
             log.debug("No physical volume found. Does not require partition resize.")
             return
-        disk = matched[0].get("disk")
-        number = matched[0].get("number")
+        disk = matched.get("disk")
+        number = matched.get("number")
         node.execute(f"growpart {disk} {number}", sudo=True)
         node.execute(f"pvresize {pv_result.splitlines()[0]}", sudo=True)
         root_partition = node.tools[Mount].get_partition_info("/")[0]
--- a/microsoft/testsuites/power/common.py
+++ b/microsoft/testsuites/power/common.py
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+import re
 from decimal import Decimal
 from typing import cast
 
@@ -9,8 +10,22 @@ from lisa import Environment, Logger, Node, RemoteNode, features
 from lisa.features import StartStop
 from lisa.features.startstop import VMStatus
 from lisa.operating_system import Redhat, Suse, Ubuntu
-from lisa.tools import Dmesg, Fio, HibernationSetup, Iperf3, KernelConfig, Kill, Lscpu
-from lisa.util import LisaException, SkippedException
+from lisa.tools import (
+    Dmesg,
+    Fio,
+    HibernationSetup,
+    Iperf3,
+    KernelConfig,
+    Kill,
+    Lscpu,
+    Mount,
+)
+from lisa.util import (
+    LisaException,
+    SkippedException,
+    UnsupportedDistroException,
+    find_groups_in_lines,
+)
 from lisa.util.perf_timer import create_timer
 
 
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -1814,6 +1814,45 @@ def _update_packages(self, packages: Optional[List[str]] = None) -> None:
...
+            self._log.debug("LVM is not enabled on the VM. Does not require LV resize.")
+            return
+        root_partition = self._node.tools[Mount].get_partition_info("/")[0]
+        os_version = self._node.os.information.version
+        if os_version >= "9.2.0":
+            sed = self._node.tools[Sed]
+            sed.substitute(
+                regexp="# use_devicesfile = 1",
+                replacement="use_devicesfile = 1",
+                file="/etc/lvm/lvm.conf",
```
LiliDeng: add comments for this setting

microsoft/testsuites/power/common.py
```diff
@@ -37,6 +37,8 @@ def verify_hibernation(
     log: Logger,
     throw_error: bool = True,
 ) -> None:
+    if isinstance(node.os, Redhat):
+        node.os.grow_os_partition_size()
```
LiliDeng: add comments for why need grow os size for rhel lvm images

adityagesh: Added

lisa/operating_system.py
adityagesh: Sorry, I forgot this remove this earlier
Deleted now

lisa/operating_system.py
```diff
@@ -1814,6 +1814,35 @@ def _update_packages(self, packages: Optional[List[str]] = None) -> None:
         if self._no_repo_enabled.search(output):
             raise RepoNotExistException(self._node.os)
 
+    def grow_os_partition_size(self) -> None:
```
squirrelsc: `expand_os_partition`

lisa/operating_system.py
```diff
@@ -1814,6 +1814,35 @@ def _update_packages(self, packages: Optional[List[str]] = None) -> None:
...
+        from lisa.tools import Mount
+
+        pv_result = self._node.execute("pvscan -s", sudo=True, shell=True).stdout
```
squirrelsc: What's the pvscan? Does it exist on all distros?

lisa/operating_system.py
```diff
@@ -1814,6 +1814,35 @@ def _update_packages(self, packages: Optional[List[str]] = None) -> None:
...
+        matched = re.compile(r"(?P<disk>.*)(?P<number>[\d]+)", re.M).match(
```
squirrelsc: Add examples above the regexp.

squirrelsc: Use LISA built-in methods from utils for capture groups.

lisa/operating_system.py
squirrelsc: Move the methods to tools, instead of in os class.

adityagesh: Create a new tool? or have it as a function inside Hibernation feature itself?

lisa/operating_system.py
adityagesh: I don't think it exists outside Redhat. This resize is only required for Redhat distro's due to the LVM configuration.
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/logical_volume_manager_administration/physvol_display

I tested in multiple version of RHEL (8.0, 8.3, 9.0, 9.3). the tool was available in all of them

The implementation is inside OS class (Redhat), so that should be alright ?

lisa/operating_system.py
squirrelsc: Yes, you can put it in test case firstly. If other places need it, then refactoring and find a better place.

adityagesh: updated

lisa/operating_system.py
adityagesh: updated

adityagesh: updated

lisa/operating_system.py
adityagesh: updated

microsoft/testsuites/power/common.py
```diff
@@ -150,3 +171,34 @@ def cleanup_env(environment: Environment) -> None:
...
+
+
+def _expand_os_partition(node: Node, log: Logger) -> None:
+    if isinstance(node.os, Redhat):
+        pv_result = node.execute("pvscan -s", sudo=True, shell=True).stdout
+        # The output of pvscan -s is like below.:
+        #  /dev/sda4
+        #  Total: 1 [299.31 GiB] / in use: 1 [299.31 GiB] / in no VG: 0 [0   ]
+        pattern = re.compile(r"(?P<disk>.*)(?P<number>[\d]+)", re.M)
+        matched = find_groups_in_lines(pv_result, pattern)
```
squirrelsc: use `find_group_in_lines`, it returns only one, not a list.

microsoft/testsuites/power/common.py
```diff
@@ -150,3 +171,34 @@ def cleanup_env(environment: Environment) -> None:
...
+            return
+        disk = matched[0].get("disk")
+        number = matched[0].get("number")
+        node.execute(f"growpart {disk} {number}", sudo=True)
+        node.execute(f"pvresize {pv_result.splitlines()[0]}", sudo=True)
+        root_partition = node.tools[Mount].get_partition_info("/")[0]
+        device_name = root_partition.name
+        device_type = root_partition.type
+        cmd_result = node.execute(f"lvdisplay {device_name}", sudo=True)
+        if cmd_result.exit_code == 0:
```
squirrelsc, line 195: what happened, if the exit_code is not 0? Add logic or comments.

microsoft/testsuites/power/common.py
adityagesh: image
It returns a list of dict

microsoft/testsuites/power/common.py
adityagesh: Updated

microsoft/testsuites/power/common.py
squirrelsc: It’s group, not groups

adityagesh: Updated

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3344, opened 2024-07-10
add set_kill_user_processes

LiliDeng:


lisa/operating_system.py
```diff
@@ -1897,6 +1902,15 @@ def add_azure_core_repo(
             sudo=True,
         )
 
+    def set_kill_user_processes(self) -> None:
```
squirrelsc, line 1907: what is this for? Why is it needed?

LiliDeng: > what is this for? Why is it needed?
We need this cause the test process being terminated when the SSH session is reset during the test on mariner image

squirrelsc: Please explain in the commit or code.

squirrelsc (approved):

PR #3346, opened 2024-07-11
Hibernation: Fix Max disks testcase

adityagesh:


microsoft/testsuites/power/power.py
```diff
@@ -254,7 +254,7 @@ def verify_hibernation_max_data_disks(self, node: Node, log: Logger) -> None:
         data_disks_after_hibernation = disk.get_raw_data_disks()
         assert_that(data_disks_before_hibernation).described_as(
             "data disks are inconsistent after hibernation"
-        ).is_length(data_disks_after_hibernation)
+        ).is_length(len(data_disks_after_hibernation))
```
LiliDeng, line 257: is_length is enough?

adityagesh: Int is expected as argument 
image

LiliDeng (approved):

PR #3351, opened 2024-07-16
Allow max_data_disk_count to work with maximize_capability

kamalca:
Currently maximize_capability breaks any test cases that rely on max_data_disk_count. Trying to manually set the value in the runbook creates a 'no quota available' error.

These changes make the default value (0,) instead of None, allowing a passed in value to merge with max_capability value.


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3353, opened 2024-07-18
VirtualClient - add perf_redis and perf_postgresql

LiliDeng:


pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "spurplus ~= 2.3.5",
     "websockets ~= 10.3",
     "charset_normalizer ~= 2.1.1",
+    "pandas",
```
squirrelsc: Is it used by csv parser only? Try built-in csv lib. We need to minimize dependencies on LISA core parts.

https://docs.python.org/3/library/csv.html

microsoft/testsuites/virtualclient/workload.py
```diff
@@ -0,0 +1,120 @@
...
+        clients_data: List[Tuple[str, str, str]] = [
+            (node_info["role"], node_info["node"].internal_address, node_info["role"])
+            for node_info in nodes
+        ]
+
+        processes = {}
+        experiment_id = str(uuid.uuid4())
+        for node_info in nodes:
+            node = node_info["node"]
+            vc_tool = node.tools[VirtualClientTool]
```
squirrelsc: use run in parallel.

microsoft/testsuites/virtualclient/workload.py
```diff
@@ -0,0 +1,120 @@
...
+            This test is to run PostgreSQL workload testing with Virtual Client.
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            unsupported_os=[BSD],
+            min_count=2,
+        ),
+        timeout=3000,
+    )
+    def perf_postgresql(self, log: Logger, result: TestResult) -> None:
```
squirrelsc: name it like `perf_vc_postgresql`

microsoft/testsuites/virtualclient/workload.py
```diff
@@ -0,0 +1,120 @@
...
+            timeout=45,
+        )
+
+    def _work_load(
+        self, result: TestResult, profile_name: str, system: str, timeout: int = 10
+    ) -> None:
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
+        nodes: List[Dict[str, Any]] = [
+            {"role": "server", "node": cast(RemoteNode, environment.nodes[1])},
```
squirrelsc: Use node 0 for first one is easier to read.

microsoft/testsuites/virtualclient/workload.py
```diff
@@ -0,0 +1,120 @@
...
+            profile_name="PERF-POSTGRESQL-HAMMERDB-TPCC",
+            system="Azure",
+            timeout=45,
+        )
+
+    def _work_load(
+        self, result: TestResult, profile_name: str, system: str, timeout: int = 10
+    ) -> None:
+        environment = result.environment
+        assert environment, "fail to get environment from testresult"
```
squirrelsc: The error message should be more accurate: the environment of test result shouldn't be empty.

microsoft/testsuites/virtualclient/workload.py
```diff
@@ -0,0 +1,120 @@
...
+            This test is to run redis workload testing with Virtual Client.
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            unsupported_os=[BSD],
+            min_count=2,
+        ),
+    )
+    def perf_redis(self, log: Logger, result: TestResult) -> None:
+        self._work_load(result=result, profile_name="PERF-REDIS", system="Azure")
```
squirrelsc: `system="Azure"` should be a default value on tool level.

microsoft/testsuites/virtualclient/workload.py
```diff
@@ -0,0 +1,120 @@
...
+            disk.add_data_disk(count=1, size_in_gb=256)
+
+        self._work_load(
+            result=result,
+            profile_name="PERF-POSTGRESQL-HAMMERDB-TPCC",
+            system="Azure",
+            timeout=45,
+        )
+
+    def _work_load(
```
squirrelsc: call it like `_run_work_load`

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,255 @@
...
+        remote_path = self.node.get_pure_path(f"{self._vc_log_path}/metrics.csv")
+        current_user = self.node.tools[Whoami].get_username()
+        self.node.tools[Chown].change_owner(remote_path, current_user)
+        self.node.shell.copy_back(
+            remote_path,
+            local_path,
+        )
+        return local_path
+
+    def create_metrics_message(self, metrics_file: Path) -> None:
```
squirrelsc: The method name should be like `send_metrics_message`, because it sends message, not only create.

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,255 @@
...
+        for name, ip, role in clients_info:
+            client_dict = {"name": name, "ipAddress": ip, "role": role}
+            clients_list.append(client_dict)
+        data = {"clients": clients_list}
+        json_data = json.dumps(data, indent=4)
+        file_path = f"{self.node.working_path}/layout.json"
+        self.node.execute(f"echo '{json_data}' > {file_path}", shell=True)
+        return file_path
+
+    def launch_command_async(
```
squirrelsc: `run_vc_command_async`. The new name `launch` is confusing.

microsoft/testsuites/virtualclient/workload.py
```diff
@@ -0,0 +1,120 @@
+import re
```
squirrelsc: The file name called `workload`, but the test suite called VirtualClient, make them consistent. Rather than "workload", is there any other test suite? If not, call them virtual client.

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,275 @@
...
+from .chmod import Chmod
+from .chown import Chown
+from .git import Git
+from .ln import Ln
+from .parted import Parted
+from .unzip import Unzip
+from .whoami import Whoami
+
+
+class VirtualClientTool(Tool):
```
squirrelsc, line 42: The class should be split into two parts, one is to operate vc on the single node, called like `VirtualClient`. Another one is to run workload on one or more nodes. It passes in multiple `VirtualClient` tools.

microsoft/testsuites/virtualclient/workload.py
squirrelsc: The tool installation is still run separated, it's slow down the test progress.

LiliDeng: image

squirrelsc: It can call `tools.get`, so the private function is not needed.

microsoft/testsuites/virtualclient/virtualclient.py
```diff
@@ -0,0 +1,133 @@
...
+                continue
+            process_result = processes[role].wait_result(timeout=(timeout + 5) * 60)
+            matched = self.exit_code_pattern.match(process_result.stdout)
+            assert matched, "can't find the matched str 'Exit Code'"
+            assert_that(matched.group("exit_code")).described_as(
+                f"Exit code on {role} is unexpected"
+            ).is_equal_to("0")
+
+        for node_info in nodes:
+            role = node_info["role"]
```
squirrelsc: `role` is not used in the loop.

microsoft/testsuites/virtualclient/virtualclient.py
```diff
@@ -0,0 +1,133 @@
...
+        self,
+        result: TestResult,
+        profile_name: str,
+        system: str = "Azure",
+        timeout: int = 10,
+    ) -> None:
+        environment = result.environment
+        assert environment, "the environment of test result shouldn't be empty"
+
+        nodes: List[Dict[str, Any]] = [
```
squirrelsc: Use a class to present the data structure, instead of a dict. It's easier to use and check name errors.

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,275 @@
...
+            f"{self.node.working_path.parent.parent}/{package_name}",
+            self._vc_download_path,
+        )
+
+        self.node.tools[Chmod].chmod(self._vc_raw_tool_path, "777")
+        ln = self.node.tools[Ln]
+        ln.create_link(self._vc_raw_tool_path, "/usr/bin/virtualclient")
+        return self._check_exists()
+
+    def generate_layout(self, clients_info: List[Tuple[str, str, str]]) -> str:
```
squirrelsc: Use named parameters, not a Tuple. It's not self explained.

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,275 @@
...
+            f"{self._vc_download_path}/content/linux-{arch}/VirtualClient"
+        )
+        self._vc_profile_file_path: str = (
+            f"{self._vc_download_path}/content/linux-{arch}/profiles"
+        )
+        self._vc_log_path: str = f"{self._vc_download_path}/content/linux-{arch}/logs"
+
+    def install(self) -> bool:
+        if isinstance(self.node.os, Debian):
+            _ = self.node.tools[Git]
```
squirrelsc: Why debian need git?

LiliDeng: update for every distro to install packages needed by VC

lisa/tools/virtualclient.py
LiliDeng: I don't see other tools design like this, could you please tell me the reason, thanks.

microsoft/testsuites/virtualclient/workload.py
LiliDeng: updated

microsoft/testsuites/virtualclient/virtualclient.py
LiliDeng: updated

lisa/tools/virtualclient.py
LiliDeng: updated

microsoft/testsuites/virtualclient/virtualclient.py
LiliDeng: updated

microsoft/testsuites/virtualclient/virtualclient.py
LiliDeng: deleted

lisa/tools/virtualclient.py
squirrelsc: Other tools operate on one node. But the VirtualClient needs to operate on multiple nodes together. For NTTTCP, the operations of multiple nodes in test case level. But we want to abstract vc tools for general use, so it needs to be supported at the tool level.

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,398 @@
...
+        # Split timestamp and microseconds
+        if "." in timestamp_str:
+            main_time, microseconds = timestamp_str.split(".")
+            # Truncate microseconds to 6 digits
+            microseconds = microseconds[:6]
+            timestamp_str = f"{main_time}.{microseconds}"
+        return datetime.strptime(timestamp_str, "%Y-%m-%dT%H:%M:%S.%f")
+
+
+class WorkloadRunner(Tool):
```
squirrelsc: This class doesn't need to be inherited from tools. It can accept a list of VcTargetInfo, as a construct parameter. So the vc client initialization is done by this class. For example,

```python
class VcRunner:
  def __init__(self, targets: List[VcTargetInfo]):
    ...

  def run(self, profile: str, system: str = "Azure", timeout ...)
```

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,390 @@
...
+
+class VcRunner:
+    # [07/08/2024 13:58:15] Exit Code: 0\r\n
+    exit_code_pattern = re.compile(r"([\w\W]*?)Exit Code: (?P<exit_code>\d+)")
+
+    def __init__(self, targets: List[VcTargetInfo]):
+        self.virtual_clients: List[Any] = run_in_parallel(
+            [partial(target.node.tools.get, VirtualClientTool) for target in targets]
+        )
+        self.vc_target_info = targets
```
squirrelsc: 1. make name be consistent for the local and class variable. For example, you can call both of them `targets`.
2. if a variable is not to be accessed out of the class, add underscore before the name.

microsoft/testsuites/virtualclient/virtualclient.py
```diff
@@ -0,0 +1,92 @@
...
+        """,
+        priority=3,
+        requirement=simple_requirement(
+            supported_os=[Debian],
+            min_count=2,
+        ),
+        timeout=3000,
+    )
+    def perf_vc_postgresql(self, log: Logger, result: TestResult) -> None:
+        environment = result.environment
```
squirrelsc: the environment is supported at the test method like node, you don't need to get it from test result.

microsoft/testsuites/virtualclient/virtualclient.py
```diff
@@ -0,0 +1,92 @@
...
+            disk = node.features[Disk]
+            disk.add_data_disk(count=1, size_in_gb=256)
+
+        self._run_work_load(
+            result=result,
+            profile_name="PERF-POSTGRESQL-HAMMERDB-TPCC",
+            timeout=45,
+        )
+
+    def _run_work_load(
```
squirrelsc, line 61: I think most of the method can be moved to vcrunner, like `def run(self, environment, profile_name, ...)`. The test methods need to specify roles and profile only.

microsoft/testsuites/virtualclient/virtualclient.py
```diff
@@ -0,0 +1,92 @@
...
+        requirement=simple_requirement(
+            supported_os=[Debian],
+            min_count=2,
+        ),
+        timeout=3000,
+    )
+    def perf_vc_postgresql(self, log: Logger, result: TestResult) -> None:
+        environment = result.environment
+        assert environment, "the environment should not be empty"
+        for node in environment.nodes.list():
```
squirrelsc: Can the data disks be set in requirements?

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,402 @@
...
+                "More roles specified than available nodes in the environment"
+            )
+
+        self._targets: List[VcTargetInfo] = []
+        nodes = [cast("RemoteNode", environment.nodes[i]) for i in range(len(roles))]
+
+        for role, node in zip(roles, nodes):
+            self._targets.append(VcTargetInfo(role=role, node=node))
+
+        self._virtual_clients: List[Any] = run_in_parallel(
```
squirrelsc: Put virtual clients into the `VcTargetInfo`, so it's saved with its role/node together, and easy to loop.

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,402 @@
...
+        return {
+            node_info.node.name: {
+                "profile_name": profile_name,
+                "system": system,
+                "client_id": node_info.node.name,
+                "layout_path": layout_file,
+                "timeout": timeout,
+                "experiment_id": experiment_id,
+            }
+            for node_info in self._targets
```
squirrelsc, line 365: As above parameter, when save vc tools into the vc targets, it doesn't need a loop to get all parameters at the same time.

microsoft/testsuites/virtualclient/virtualclient.py
squirrelsc: This method is used in this method only, the `system` is not set in test methods, so it should be removed.

microsoft/testsuites/virtualclient/virtualclient.py
```diff
@@ -0,0 +1,77 @@
...
+            environment=environment,
+            profile_name="PERF-POSTGRESQL-HAMMERDB-TPCC",
+            timeout=45,
+        )
+
+    def _run_work_load(
+        self,
+        environment: Environment,
+        profile_name: str,
+        system: str = "Azure",
```
squirrelsc: Remove the `system`, because it's not set in above test cases.

lisa/tools/virtualclient.py
```diff
@@ -0,0 +1,398 @@
...
+        results = self._execute_commands(client_params)
+        self._wait_for_client_result(results, timeout)
+        self._process_results()
+
+    def _generate_layout_file(
+        self,
+    ) -> str:
+        layout_file = ""
+        if len(self._targets) > 1:
+            for target in self._targets:
```
squirrelsc, line 344: The logic is unclear. 

1. If there is only 1 target, why not generate layout.
2. If there are 3 targets, the layout will be generated multiple times, but only last one is taken.

LiliDeng: the layout file need to be generated in each target, but the path is identical for each target

squirrelsc: 1. what happen when there is only 1 target? Should the profile be generated?
2. you can add comments in code to explain the file path is the same, so return last one is ok.

squirrelsc (approved):

PR #3355, opened 2024-07-22
Account the new kernel config for kexec

0xba1a:
A new kernel config, CONFIG_KEXEC_FILE is introduced in recent kernel version. This config introduces a safer systemcall to load a kexec kernel. Now kexec kernel can be loaded by both the old and the new systemcalls. So, kdump test should check for either of KEXEC or KEXEC_FILE config for determining the kernel's ability.

### DIFF
--- a/lisa/tools/kdump.py
+++ b/lisa/tools/kdump.py
@@ -228,8 +228,8 @@ class KdumpBase(Tool):
                 break
         if not kexec_config_present:
             raise LisaException(
-                "The kernel config CONFIG_KEXEC or CONFIG_KEXEC_FILE is not set. Kdump is"
-                "not supported."
+                "The kernel config CONFIG_KEXEC or CONFIG_KEXEC_FILE is not set. "
+                "Kdump is not supported."
             )
         for config in self.required_kernel_config:
             if not self.node.tools[KernelConfig].is_built_in(config):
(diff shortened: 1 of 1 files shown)


lisa/tools/kdump.py
```diff
@@ -217,6 +221,16 @@ def _install(self) -> bool:
...
+        kexec_config_present = False
+        for config in self.kexec_kernel_configs:
+            if self.node.tools[KernelConfig].is_built_in(config):
+                kexec_config_present = True
+                break
+        if not kexec_config_present:
+            raise LisaException(
+                "The kernel config CONFIG_KEXEC or CONFIG_KEXEC_FILE is not set. Kdump is"
+                "not supported."
+            )
```
LiliDeng, line 233: if not kexec_config_present:
            raise LisaException(
                "The kernel config CONFIG_KEXEC or CONFIG_KEXEC_FILE is not set. "
                "Kdump is not supported."
            )

0xba1a: @LiliDeng Updated.

lisa/tools/kdump.py
```diff
@@ -180,11 +180,15 @@ class KdumpBase(Tool):
...
-        "CONFIG_KEXEC",
         "CONFIG_CRASH_DUMP",
         "CONFIG_PROC_VMCORE",
     ]
 
+    kexec_kernel_configs = [
+        "CONFIG_KEXEC",
+        "CONFIG_KEXEC_FILE"
+    ]
+
```
LiliDeng, line 191:
```
    kexec_kernel_configs = [
        "CONFIG_KEXEC",
        "CONFIG_KEXEC_FILE",
    ]
```

LiliDeng (approved):

PR #3356, opened 2024-07-24
Added FreeBSD Date class

r-dailey:
EDIT: Marked this as a DRAFT PR as my git and linting tools have gotten messed up. I will resolve and reopen the PR when it is fixed. Added a FreeBSD specific class for date to accommodate the freebsd date set formatting

### DIFF
--- a/lisa/tools/date.py
+++ b/lisa/tools/date.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 
 from datetime import datetime
+from typing import Optional, Type
 
 from dateutil.parser import parser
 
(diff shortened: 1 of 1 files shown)


squirrelsc (dismissed):

lisa/tools/date.py
```diff
@@ -32,3 +37,25 @@ def set(self, new_date: datetime) -> None:
             expected_exit_code=0,
             expected_exit_code_failure_message="Failed to set date time.",
         )
+
+
+class DateFreeBSD(Date):
+    @property
```
squirrelsc: It doesn't need to duplicate the same methods. Just write the set method.

r-dailey: Removed all the extra methods and just left the set function.

squirrelsc (approved):

LiliDeng (approved):

PR #3360, opened 2024-07-30
Filter out ANSI characters

kamalca:
ANSI characters which add color and other formatting to the console output prevent regex matching from working throughout LISA on some distros.

### DIFF
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -38,8 +38,8 @@ TIMEOUT_READING_PASSWORD_PATTERNS = [
 
 @dataclass
 class ExecutableResult:
-    stdout: str
-    stderr: str
+    _stdout: str
+    _stderr: str
     exit_code: Optional[int]
     cmd: Union[str, List[str]]
     elapsed: float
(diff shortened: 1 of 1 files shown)


lisa/util/process.py
```diff
@@ -35,11 +35,13 @@
     re.compile(r"Password: .+\r\nsudo: timed out reading password"),
 ]
 
+ANSI_ESCAPE = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
```
squirrelsc: There is an util method in utils already. Please filter out it in the place, instead of the lower level class. I'm not sure if there are side effects on existing code.

kamalca: I made the changes as you requested. However, my fear is that we will waste a lot of time debugging issues related to regex matching being affected by ANSI escape characters. These issues are not obvious from the error messages and take time to debug. We shouldn't rely on LISA contributors to foresee ANSI characters in stdout. It should be easy to write bug free code that works on all distros.

squirrelsc: Sorry, I forgot it. I agree to add it in execute result. @LiliDeng, please be aware check if there are regressions.

kamalca: Okay, I will move it back to the executable result.

squirrelsc (approved):

LiliDeng:
testing

PR #3361, opened 2024-07-31
remove check service principal key requirement logic

paxue:
Remove service principal key requirement logic from exception. Key is not supported anymore

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -335,15 +335,10 @@ class AzurePlatformSchema:
             add_secret(self.service_principal_tenant_id, mask=PATTERN_GUID)
         if self.subscription_id:
             add_secret(self.subscription_id, mask=PATTERN_GUID)
-
-        if self.service_principal_client_id or self.service_principal_key:
-            add_secret(self.service_principal_client_id, mask=PATTERN_GUID)
+        if self.service_principal_key:
             add_secret(self.service_principal_key)
-            if not self.service_principal_client_id or not self.service_principal_key:
-                raise LisaException(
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -339,11 +339,8 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
         if self.service_principal_client_id or self.service_principal_key:
             add_secret(self.service_principal_client_id, mask=PATTERN_GUID)
             add_secret(self.service_principal_key)
-            if not self.service_principal_client_id or not self.service_principal_key:
-                raise LisaException(
-                    "service_principal_client_id and service_principal_key "
-                    "should be specified either both or not."
-                )
+            if not self.service_principal_client_id:
```
squirrelsc: It needs to split checks for client id and key. They don't need to be in the same if block.

squirrelsc: @LiliDeng, with supports of managed identity, even the client id should be optional, is it right?

LiliDeng: with managed identity, yes, no need client id, if use azure cli login, client id is still needed

squirrelsc: If the client id isn't required in all cases, it should be optional.

squirrelsc: With the new logic, the exception is not needed. Because L341 checked it exists already.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3362, opened 2024-08-01
Use subscription client to find the sub id for a storage account

lubaihua33:
This is related with PR https://github.com/microsoft/lisa/pull/3348 and https://github.com/microsoft/lisa/pull/3329.
In this PR, use subscription client to find a subscription which has the storage account. 

"subscription" is removed from SigTransformerSchema and vhdSchema. So that it doesn't need any changes in AITL for AAD vhd path scenario.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -520,7 +520,6 @@ class SharedImageGallerySchema(AzureImageSchema):
 class VhdSchema(AzureImageSchema):
     vhd_path: str = ""
     vmgs_path: Optional[str] = None
-    subscription: str = ""
 
     def load_from_platform(self, platform: "AzurePlatform") -> None:
         return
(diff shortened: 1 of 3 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -2159,18 +2158,41 @@ def find_by_name(resources: Any, type_name: str) -> Any:
     return next(x for x in resources if x["type"] == type_name)
 
 
-def get_vhd_details(
-    platform: "AzurePlatform", vhd_path: str, subscription: str = ""
-) -> Any:
-    if subscription == "":
-        subscription = platform.subscription_id
+def get_vhd_details(platform: "AzurePlatform", vhd_path: str) -> Any:
```
squirrelsc, line 2161: Is this method used for sas token? We still allow sas token, will it be broken?

lubaihua33: Yes. The sas token scenario won't be broken.
The change is for the scenario that the storage account is in other subscriptions. The added code is that if the storage account couldn't be found in the platform.subscription_id, then will search in other subscriptions that the credential allows.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3364, opened 2024-08-02
get identity id: try to read from os environment

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -2734,6 +2734,9 @@ def get_managed_identity_object_id(
 def get_identity_id(
     platform: "AzurePlatform", application_id: Optional[str] = None
 ) -> Any:
+    if not application_id:
+        application_id = os.environ.get("AZURE_CLIENT_ID")
```
squirrelsc: Make sure it won't fail, if the variable doesn't exist in the environment, like `os.environ.get("AZURE_CLIENT_ID", "")`

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
```diff
@@ -63,7 +63,7 @@ def _check_system_status(node: Node, log: Logger) -> None:
         expected_exit_code_failure_message=message,
     )
 
-    certb = "/var/lib/waagent/a/symbolinkb"
+    certb = "/var/lib/waagent/b/symbolinkb"
```
squirrelsc: Why is this char changed? Is it possible to be changed in future? Should we use a wildchar to match more paths?

LiliDeng: it is a test bug, it was written wrong in the beginning.
image

squirrelsc: Should it support both a and b?

lisa/sut_orchestrator/azure/common.py

microsoft/testsuites/vm_extensions/azure_keyvault_vm_extension.py
LiliDeng: updated

squirrelsc (approved):

PR #3367, opened 2024-08-03
Added BareMetal platform support for LTP/Kselftest

paulli2017:
Added BareMetal platform support for LTP/Kselftest

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest.py
+++ b/microsoft/testsuites/kselftest/kselftest.py
@@ -10,12 +10,21 @@ from lisa.base_tools.uname import Uname
 from lisa.executable import Tool
 from lisa.messages import TestStatus, send_sub_test_result_message
 from lisa.node import Node
-from lisa.operating_system import CBLMariner, Ubuntu
+from lisa.operating_system import (
+    CBLMariner,
+    Debian,
+    Linux,
+    NixOS,
+    OtherLinux,
+    Posix,
+    RPMDistro,
+    Suse,
+    Ubuntu,
+)
 from lisa.testsuite import TestResult
 from lisa.tools import Cp, Git, Ls, Make, RemoteCopy, Tar
 from lisa.tools.chmod import Chmod
 from lisa.tools.mkdir import Mkdir
-from lisa.tools.whoami import Whoami
 from lisa.util import LisaException, UnsupportedDistroException, find_groups_in_lines
 
 _UBUNTU_OS_PACKAGES = [
@@ -98,26 +107,44 @@ class Kselftest(Tool):
 
         # tar file path specified in yml
         self._tar_file_path = kselftest_file_path
+        kselftest_install_path = "/tmp/kselftest"
+        kselftest_packages = "kselftest-packages"
         if self._tar_file_path:
-            self._remote_tar_path = self.get_tool_path(
-                use_global=True
+            self._remote_tar_path = self.node.get_pure_path(
+                kselftest_install_path,
             ) / os.path.basename(self._tar_file_path)
 
-        # command to run kselftests
-        self._kself_installed_dir = (
-            self.get_tool_path(use_global=True) / "kselftest-packages"
-        )
+            self._kself_installed_dir = self.node.get_pure_path(
+                kselftest_install_path,
+            ) / (kselftest_packages)
+        else:
+            self._remote_tar_path = self.get_tool_path(
+                use_global=True,
+            ) / os.path.basename(self._tar_file_path)
+
+            self._kself_installed_dir = self.get_tool_path(
+                use_global=True,
+            ) / (kselftest_packages)
 
         self._command = self._kself_installed_dir / "run_kselftest.sh"
 
     # install common dependencies
     def _install(self) -> bool:
-        if not (
-            (
-                isinstance(self.node.os, Ubuntu)
-                and self.node.os.information.version >= "18.4.0"
+        if (
+            (isinstance(self.node.os, Posix) and not isinstance(self.node.os, Linux))
+            or (
+                isinstance(self.node.os, Debian)
+                and not isinstance(self.node.os, Ubuntu)
             )
-            or isinstance(self.node.os, CBLMariner)
+            or (
+                isinstance(self.node.os, Ubuntu)
+                and self.node.os.information.version < "18.4.0"
+            )
+            or (
+                isinstance(self.node.os, RPMDistro)
+                and not isinstance(self.node.os, CBLMariner)
+            )
+            or (isinstance(self.node.os, (Suse, NixOS, OtherLinux)))
         ):
             raise UnsupportedDistroException(
                 self.node.os, "kselftests in LISA does not support this os"
@@ -202,14 +229,12 @@ class Kselftest(Tool):
         test_result: TestResult,
         log_path: str,
         timeout: int = 5000,
-        run_test_as_root: bool = False,
     ) -> List[KselftestResult]:
         # Executing kselftest as root may cause
         # VM to hang
 
-        # get username
-        username = self.node.tools[Whoami].get_username()
-        result_directory = f"/home/{username}"
+        # get result directory
+        result_directory = "/tmp/"
         if os.path.exists(result_directory) is False:
             mkdir = self.node.tools[Mkdir]
             mkdir.create_directory(result_directory)
@@ -218,7 +243,7 @@ class Kselftest(Tool):
         result_file = f"{result_directory}/{result_file_name}"
         self.run(
             f" 2>&1 | tee {result_file}",
-            sudo=run_test_as_root,
+            sudo=True,
             force_run=True,
             shell=True,
             timeout=timeout,
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+
 import os
 import re
 from dataclasses import dataclass
@@ -24,7 +25,6 @@ from lisa.tools import (
     Make,
     Mkdir,
     Pgrep,
-    RemoteCopy,
     Rm,
     Swap,
     Sysctl,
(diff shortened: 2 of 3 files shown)


microsoft/testsuites/ltp/ltp.py
```diff
@@ -53,29 +55,73 @@ class Ltp(Tool):
...
-        return "/opt/ltp/runltp"
+        if self._ltp_tar_file:
+            return "/tmp/ltp/runltp"
+        else:
+            return "/opt/ltp/runltp"
+
+    @property
+    def command_dir(self) -> str:
+        if self._ltp_tar_file:
+            return "/tmp/ltp/"
```
squirrelsc: define the command path as a parameter, instead of set it by the tar file path.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -53,29 +55,73 @@ class Ltp(Tool):
...
+            return "/tmp/ltp/skipfile"
+        else:
+            return "/opt/ltp/skipfile"
+
+    def _check_exists(self) -> bool:
+        return (
+            len(self.node.tools[Ls].list(str(self.command_dir), sudo=True)) > 0
+        )
+
+    def __init__(self, node: Node, ltp_tar_file: str, *args: Any, **kwargs: Any) -> None:
```
squirrelsc: `ltp_tar_file` doesn't describe the use of the variable. It should be called like `prebuilt_package`.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -117,7 +122,7 @@ def _install(self) -> bool:
                 isinstance(self.node.os, Ubuntu)
                 and self.node.os.information.version >= "18.4.0"
             )
-            or isinstance(self.node.os, CBLMariner)
+            or isinstance(self.node.os, CBLMariner) or isinstance(self.node.os, SocOs)
```
squirrelsc: Convert to positive conditions, skip other endorsed distros like RedHat, SUSE, etc.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -209,7 +214,7 @@ def run_all(
 
         # get username
         username = self.node.tools[Whoami].get_username()
-        result_directory = f"/home/{username}"
+        result_directory = f"/tmp/{username}"
```
squirrelsc, line 217: keep the original path, and use a test case parameter to control the result path.

squirrelsc: If the path is moved to tmp, it doesn't need the `{username}` part.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -218,7 +223,7 @@ def run_all(
         result_file = f"{result_directory}/{result_file_name}"
         self.run(
             f" 2>&1 | tee {result_file}",
-            sudo=run_test_as_root,
+            sudo=True,
```
squirrelsc, line 246: Remove the whole argument `run_test_as_root`, because it's not used anymore.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -52,30 +54,58 @@ class Ltp(Tool):
...
-    LTP_RESULT_PATH = "/opt/ltp/ltp-results.log"
-    LTP_OUTPUT_PATH = "/opt/ltp/ltp-output.log"
-    LTP_SKIP_FILE = "/opt/ltp/skipfile"
     COMPILE_TIMEOUT = 1800
     RUN_TIMEOUT = 12000
 
     @property
     def command(self) -> str:
-        return "/opt/ltp/runltp"
+        return self._command_path + "runltp"
```
squirrelsc: Use f string like `f"{self._command_path}runltp"`

microsoft/testsuites/ltp/ltp.py
```diff
@@ -52,30 +54,58 @@ class Ltp(Tool):
...
+
+    @property
+    def output_path(self) -> str:
+        return self._command_path + self.LTP_OUTPUT_FILE
+
+    @property
+    def skip_path(self) -> str:
+        return self._command_path + self.LTP_SKIP_FILE
+
+    def _check_exists(self) -> bool:
```
squirrelsc: The default _check_exists should work, it doesn't need to override.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -52,30 +54,58 @@ class Ltp(Tool):
...
+
+    def __init__(self, node: Node, prebuild_package: str, *args: Any, **kwargs: Any) -> None:
         super().__init__(node, args, kwargs)
         git_tag = kwargs.get("git_tag", "")
         self._git_tag = git_tag if git_tag else self.DEFAULT_LTP_TESTS_GIT_TAG
 
+        self._prebuild_package = prebuild_package
+        if self._prebuild_package:
+            self._command_path = "/tmp/ltp/"
+            self._remote_tar_path = self.node.get_pure_path(self._command_path) / os.path.basename(self._prebuild_package)
```
squirrelsc: The two variables can be set or use when installing. They don't need to be saved in class.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -124,37 +154,36 @@ def run_test(
...
 
         # Minimum 4M swap space is needed by some mmp test
         if self.node.tools[Free].get_swap_size() < 4:
             self.node.tools[Swap].create_swap()
 
         # run ltp tests
-        command = f"{self.command} {parameters}"
-        self.node.execute_async(
-            f"echo y | {command}",
+        self.run(
```
squirrelsc: Use the async process, so the timeout detection is used with pgrep. Because some SSH connection will be disconnected, if no output in 1 minute. If your scenario doesn't support pgrep, use an parameter to control it like set `ltp_use_pgrep_timeout` to false.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -58,6 +58,7 @@ def verify_ltp_lite(
         tests = variables.get("ltp_test", "")
         skip_tests = variables.get("ltp_skip_test", "")
         ltp_tests_git_tag = variables.get("ltp_tests_git_tag", "")
+        ltp_tar_file = variables.get("ltp_file_path", "")
```
squirrelsc: the test variable name should be clear, like `ltp_prebuilt_file`. Two names should be consistent.

paulli2017: The change is done, thanks

microsoft/testsuites/ltp/ltp.py
paulli2017: The change is done, thanks

microsoft/testsuites/ltp/ltp.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: the change is done, thanks

microsoft/testsuites/ltp/ltp.py
paulli2017: The change is done, thanks

microsoft/testsuites/ltp/ltp.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: Please push latest changes.

microsoft/testsuites/ltp/ltp.py
paulli2017: From my debug, it seems there is an issue when self.node.execute_async() is called.  self.node.execute_async() will just start to run the command, but if username and password are needed before the command is actually running, the command will never run. However, if self.run() is called, it will not only call self.node.execute_async(), but also wait for the result. This wait helps deal with the username and password issue before the command is run. So if pgrep is needed, the username and password issue also needs to be solved. Please let me know your suggestion.

squirrelsc: Please check with @lubaihua33 to move the logic to shell level and call it in execute_async.

paulli2017: Sure, I will do that, By the way, that change of execute_async() is to make sure the command will be exectuted anyway, however, the current implementation of pgrep.wait_process() just checks the command and exits when the command is not found, this might cause an issue too. For example, while execute_async() is dealing with username and password, pgrep.wait_process() might think the command is finished as it doesn't detect the process, thus the test is treated as done by mistake, and no test result is to be found, which will fail the test unfortunately. I think pgrep.wait_process() might need to check the start of the command first, and then checks when the command is finished. Please let me know your suggestion.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -99,27 +103,31 @@ def __init__(
...
 
     # install common dependencies
     def _install(self) -> bool:
-        if not (
-            (
-                isinstance(self.node.os, Ubuntu)
-                and self.node.os.information.version >= "18.4.0"
-            )
-            or isinstance(self.node.os, CBLMariner)
+        if (isinstance(self.node.os,
```
squirrelsc: You don't need to list all of them here. 1) The BSD covers FreeBSD and OpenBSD, etc, 2) Some uncommon distros doesn't list here, like NixOS, OtherLinux.

microsoft/testsuites/ltp/ltp.py
squirrelsc: The `execute_async` doesn't mean the method is fully async. The username/password checks are synchronous. Before the checks completed, the `execute_async` doesn't exit, so the pgrep won't start.

paulli2017: that makes sense, thanks for clarification.

squirrelsc: You can make the change in next PR.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -99,27 +103,31 @@ def __init__(
         # tar file path specified in yml
         self._tar_file_path = kselftest_file_path
         if self._tar_file_path:
+            self._remote_tar_path = (
+                self.node.get_pure_path("/tmp/kselftest") / os.path.basename(self._tar_file_path)
```
squirrelsc: `self.node.get_pure_path("/tmp/kselftest")` called twice, use a variable to share the logic.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -85,7 +86,7 @@ def verify_ltp_lite(
                 )
 
         # run ltp lite tests
-        ltp: Ltp = node.tools.get(Ltp, git_tag=ltp_tests_git_tag)
+        ltp: Ltp = node.tools.get(Ltp, ltp_prebuilt_file, git_tag=ltp_tests_git_tag)
```
squirrelsc: 1. Use named parameters as much as possible. 
2. Make parameters names consistent as much as possible. It calls `ltp_prebuilt_file` at this level, so the argument should be like `prebuilt_file` instead of `prebuilt_package`. Versus is ok, but make them consistent.

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: Please check relationships of all distros in the operating_system.py, for example, the Fedora includes Redhat, CentOS.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -52,30 +54,52 @@ class Ltp(Tool):
...
+
+    @property
+    def output_path(self) -> str:
+        return f"{self._command_path}{self.LTP_OUTPUT_FILE}"
+
+    @property
+    def skip_path(self) -> str:
+        return f"{self._command_path}{self.LTP_SKIP_FILE}"
+
+    def __init__(self, node: Node, prebuild_file: str, *args: Any, **kwargs: Any) -> None:
```
squirrelsc: `prebuilt_file`, not `prebuild_file` like other places.

microsoft/testsuites/ltp/ltpsuite.py
squirrelsc: Call it with named parameters: `node.tools.get(Ltp, prebuilt_file=ltp_prebuilt_file, git_tag=ltp_tests_git_tag)`

squirrelsc:
Please resolve conflicts, so the checks can run.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@paulli2017 I saw error with this test matrix.

`canonical 0001-com-ubuntu-server-jammy 22_04-lts-gen2 22.04.202302140 KselftestTestsuite.verify_kselftest: FAILED   failed. SSHException: SSH session not active`

Test runs well with main branch.

PR #3371, opened 2024-08-08
Verify Hibernation: Add check using uptime

adityagesh:
We have observed scenario where hibernation logs are not consistently present, in those cases, we still need to verify if Hibernation was successful. One way to do this is using uptime

### DIFF
--- a/microsoft/testsuites/power/common.py
+++ b/microsoft/testsuites/power/common.py
@@ -20,6 +20,7 @@ from lisa.tools import (
     Lscpu,
     Mount,
 )
+from lisa.tools.uptime import Uptime
 from lisa.util import (
     LisaException,
     SkippedException,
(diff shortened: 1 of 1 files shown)


LiliDeng (dismissed):

microsoft/testsuites/power/common.py
```diff
@@ -71,6 +69,7 @@ def verify_hibernation(
 
     # only set up hibernation setup tool for the first time
     hibernation_setup_tool.start()
+    uptime_before_hibernation: str = node.execute("uptime -s", shell=True).stdout
```
LiliDeng: please use tool uptime

adityagesh: updated

LiliDeng (approved):

PR #3372, opened 2024-08-08
Tools: Fix Echo and improve Modprobe

pupacha:
* Echo: enclose file in quotes
* Modprobe: add a new method `load_with_parameters` and support loading kernel
modules with parameters

### DIFF
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -4,6 +4,7 @@ from typing import Any, List, Optional, Type, Union
 
 from lisa.executable import Tool
 from lisa.tools.kernel_config import KLDStat
+from lisa.util import UnsupportedOperationException
 
 
 class Modprobe(Tool):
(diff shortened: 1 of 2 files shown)


lisa/tools/modprobe.py
```diff
@@ -111,6 +111,31 @@ def load(
         )
         return True
 
+    def load_with_parameters(
```
LiliDeng: add the parameters as option parameter for load method

pupacha: `load` method can take list of modules as input. Parameters can be passed in `Modprobe` only with one module. Hence, a new method.

LiliDeng: You can throw an exception if parameters is specified when modules is a list.

squirrelsc: > `load` method can take list of modules as input. Parameters can be passed in `Modprobe` only with one module. Hence, a new method.

You can just join the mod list and parameters, and let modprobe return non-zero exit-code.

pupacha: > > `load` method can take list of modules as input. Parameters can be passed in `Modprobe` only with one module. Hence, a new method.
> 
> You can just join the mod list and parameters, and let modprobe return non-zero exit-code.

I dont think we should consciously append paremeters to mod list (`-a` option) and let modprobe fail.


> You can throw an exception if parameters is specified when modules is a list.

Sure, I will raise `UnsupportedOperationException`

squirrelsc (approved):

PR #3373, opened 2024-08-09
DPDK MANA CentOS 7 changes

paboldin:
This patchset is built on top of `mcgov/dpdk-mana-merge` branch and provides support to build DPDK and RDMA-CORE on CentOS 7.0 Azure images.

### DIFF
--- a/lisa/base_tools/mv.py
+++ b/lisa/base_tools/mv.py
@@ -2,6 +2,7 @@
 # Licensed under the MIT license.
 
 from lisa.executable import Tool
+from typing import Union
 
 
 class Mv(Tool):
@@ -14,13 +15,15 @@ class Mv(Tool):
         return False
 
     def move(
-        self, src_path: str, dest_path: str, overwrite: bool = False, sudo: bool = False
+        self, src_path: str, dest_path: str, overwrite: bool = False,
+        sudo: bool = False, ignore_error: bool = False
     ) -> None:
         args = "-f" if overwrite else ""
+        expected_exit_code = None if ignore_error else 0
         self.run(
             f"{args} {src_path} {dest_path}",
             sudo=sudo,
             shell=True,
             force_run=True,
-            expected_exit_code=0,
+            expected_exit_code=expected_exit_code,
         )
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -14,7 +14,7 @@ from lisa.operating_system import CBLMariner, Redhat, Ubuntu
 from lisa.tools import Cp, Echo, Git, Make, Sed, Uname
 from lisa.tools.gcc import Gcc
 from lisa.tools.lscpu import Lscpu
-from lisa.util import LisaException, field_metadata, subclasses
+from lisa.util import LisaException, VersionInfo, field_metadata, subclasses, parse_version
 from lisa.util.logger import Logger, get_logger
 
 from .kernel_installer import BaseInstaller, BaseInstallerSchema
@@ -158,11 +158,23 @@ class SourceInstaller(BaseInstaller):
         # modify code
         self._modify_code(node=node, code_path=self._code_path)
 
+
+        result = node.execute(
+            "make kernelversion 2>/dev/null",
+            cwd=self._code_path,
+            shell=True,
+        )
+        result.assert_exit_code(
+            0,
+            f"failed on get kernel version: {result.stdout}",
+        )
+        kernel_version = parse_version(result.stdout)
+
         kconfig_file = runbook.kernel_config_file
         self._build_code(
-            node=node, code_path=self._code_path, kconfig_file=kconfig_file
+            node=node, code_path=self._code_path, kconfig_file=kconfig_file,
+            kernel_version=kernel_version
         )
-
         self._install_build(node=node, code_path=self._code_path)
 
         result = node.execute(
@@ -170,16 +182,15 @@ class SourceInstaller(BaseInstaller):
             cwd=self._code_path,
             shell=True,
         )
-
-        kernel_version = result.stdout
         result.assert_exit_code(
             0,
-            f"failed on get kernel version: {kernel_version}",
+            f"failed on get kernel release: {result.stdout}",
         )
+        kernel_release = result.stdout
 
         # copy current config back to system folder.
         result = node.execute(
-            f"cp .config /boot/config-{kernel_version}",
+            f"cp .config /boot/config-{kernel_release}",
             cwd=self._code_path,
             sudo=True,
         )
@@ -227,7 +238,7 @@ class SourceInstaller(BaseInstaller):
             self._log.debug(f"modifying code by {modifier.type_name()}")
             modifier.modify()
 
-    def _build_code(self, node: Node, code_path: PurePath, kconfig_file: str) -> None:
+    def _build_code(self, node: Node, code_path: PurePath, kconfig_file: str, kernel_version: VersionInfo) -> None:
         self._log.info("building code...")
 
         uname = node.tools[Uname]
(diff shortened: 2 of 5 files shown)


lisa/features/network_interface.py
```diff
@@ -28,6 +28,7 @@ def switch_sriov(
         raise NotImplementedError
 
     def is_enabled_sriov(self) -> bool:
+        return True # FIXME
```
squirrelsc, line 31: Why is it needed? The base class shouldn't be instanced, so this method can be implemented in subclasses.

squirrelsc: If it's preexisting VM in Azure, you can specify the vm name in runbook, so you can use the full capability of Azure platform, instead of using Ready platform.

lisa/transformers/kernel_source_installer.py
```diff
@@ -325,6 +315,23 @@ def _install_build_tools(self, node: Node, build_deps: list[str]) -> None:
                     os.install_packages(package)
             os.group_install_packages("Development Tools")
 
+            # if the kernel requires devtoolset, install its gcc
+            has_devtoolset = [
```
squirrelsc: variable name should be `devtoolsets`, because it saves strings of packages, not bool.

lisa/transformers/kernel_source_installer.py
```diff
@@ -325,6 +315,23 @@ def _install_build_tools(self, node: Node, build_deps: list[str]) -> None:
...
+                    pkg
+                    for pkg in build_deps
+                    if pkg.startswith("devtoolset")
+            ]
+            if has_devtoolset:
+                assert len(has_devtoolset) == 1, "only one devtoolset can be given"
+                devtoolset = has_devtoolset[0]
+                node.os.install_packages([devtoolset])
```
squirrelsc, line 327: Use `node.os.install_packages(devtoolsets)`, and it doesn't need to take 1st from previous line. After assertion, the array is guaranteed 1 item.

squirrelsc: `install_packages` accept `str` as 1 package.

lisa/transformers/kernel_source_installer.py
```diff
@@ -325,6 +315,23 @@ def _install_build_tools(self, node: Node, build_deps: list[str]) -> None:
...
+                node.tools[Mv].move("/bin/gcc", "/bin/gcc_back", overwrite=True, sudo=True)
+                result.assert_exit_code()
```
squirrelsc: add message, so once the assertion failed, the error message give more details.

lisa/transformers/kernel_source_installer.py
```diff
@@ -143,7 +146,7 @@ def install(self) -> str:
         runbook: SourceInstallerSchema = self.runbook
         assert runbook.location, "the repo must be defined."
 
-        self._install_build_tools(node)
+        self._install_build_tools(node, runbook.build_deps)
```
squirrelsc: Can the build_deps be added into this class, instead of configurable? It makes repro harder from others.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -615,7 +622,9 @@ def _install(self) -> bool:
             )
 
         if isinstance(node.os, Redhat) and node.os.information.version < "8.0.0":
+            self._fix_mirrorlist_to_vault(node)
```
squirrelsc, line 627: Why is it called twice? If it's by design, please add code comments.

lisa/base_tools/mv.py
```diff
@@ -14,13 +15,14 @@ def can_install(self) -> bool:
         return False
 
     def move(
-        self, src_path: str, dest_path: str, overwrite: bool = False, sudo: bool = False
+        self, src_path: str, dest_path: str, overwrite: bool = False,
+        sudo: bool = False, expected_exit_code: Union[None, int] = 0
```
squirrelsc: add a parameter `ignore_error`, if it's true, set the `expected_exit_code` to `None` in the method. It's easier to understand than exposing `expected_exit_code` directly.

squirrelsc:
@paboldin Thank you for contribution! Please agree the CLA, so I can trigger checks.

paboldin:
@microsoft-github-policy-service agree company="CloudLinux"

lisa/features/network_interface.py
paboldin: Is there are documentation for this? If not, please provide samples or code pointers so that I can document this.

lisa/transformers/kernel_source_installer.py
paboldin: Can you please elaborate?

Currently it is added as a configurable because different kernel versions might need different build dependencies.

paboldin:
@squirrelsc done, and I believe addressed most of your comments. Removed the top commit with "return True" on SRIOV check.

Thanks for the review!

EDIT: comment -> commit

squirrelsc:
@paboldin , I just found it's not merged into main. Can you direct the changes to main branch?

paboldin:
> @paboldin , I just found it's not merged into main. Can you direct the changes to main branch?

I'm afraid not, underlying branch is necessary.

squirrelsc:
@mcgov what's the plan to send PR for [dpdk-mana-merge](https://github.com/microsoft/lisa/tree/mcgov/dpdk-mana-merge)? If the PR goes to dpdk-mana-merge, it needs review again, when it's merged to main. If we have to do, please make sure each commit is meaningful and well managed.

lisa/transformers/kernel_source_installer.py
squirrelsc: Please take care all `assert_exit_code`, add error message.

lisa/transformers/kernel_source_installer.py
```diff
@@ -286,33 +289,49 @@ def _build_code(self, node: Node, code_path: PurePath, kconfig_file: str) -> Non
...
+            if devtoolsets:
+                assert len(devtoolsets) == 1, f"only one devtoolset can be given, instead of {devtoolsets}"
+                devtoolset = devtoolsets[0]
+                node.os.install_packages(devtoolset)
+                node.tools[Mv].move("/bin/gcc", "/bin/gcc_back", overwrite=True, sudo=True)
+                result.assert_exit_code()
+                result = node.execute(
+                    f"ln -s /opt/rh/{devtoolset}/root/usr/bin/gcc /bin/gcc", sudo=True
+                )
+                result.assert_exit_code(f"can not link {devtoolset} to gcc")
```
squirrelsc: Use named arguments, so it won't be broken, if the interface changed in future.

lisa/transformers/kernel_source_installer.py
squirrelsc: You can add checks for kernel versions, if it's some version, add the deps. If the deps are maintained out of code, it's hard to others to make the build tools works without the right configuration.

paboldin: Yeah, but we fetch the kernel version after installation of the build tools (otherwise `make kernelrelease` might fail). Should I add a second call to install_build_tools after kernel version is known?

squirrelsc: Yes, you can call installation again, after know kernel version. Please add comments to explain the dependencies.

LiliDeng:
@paboldin any plan to update this PR?

paboldin:
@LiliDeng I'm on it. Any more comments for review?

squirrelsc:
@mcgov can you continue the work on [microsoft:mcgov/dpdk-mana-merge](https://github.com/microsoft/lisa/tree/mcgov/dpdk-mana-merge)? There may be many conflicts with main branch.

squirrelsc:
@paboldin Thank you for contribution! Would you split tool, installer changes in separated PRs to main branch? The dpdk is changed a lot, so mcgov may need more time to merge the private branch firstly.

PR #3374, opened 2024-08-09
Added KLDLoad and updated modprobeBSD

r-dailey:
Added a KLDLoad tool to the kernel config tool set. Additionally added a load function to the modprobeBSD tool for managing modules during BSD testing.

### DIFF
--- a/lisa/tools/modprobe.py
+++ b/lisa/tools/modprobe.py
@@ -162,3 +162,55 @@ class ModprobeFreeBSD(Modprobe):
                 return False
 
         return True
+
+    def load(
+        self,
+        modules: Union[str, List[str]],
+        parameters: str = "",
+        dry_run: bool = False,
+    ) -> bool:
+        if parameters:
+            raise UnsupportedOperationException(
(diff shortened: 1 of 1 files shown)


lisa/tools/kernel_config.py
```diff
@@ -133,3 +133,30 @@ def module_exists(
             return True
 
         return False
+
+
+class KLDLoad(Tool):
```
squirrelsc: If the class is used in modprobe only, it should be in modeprobe.

r-dailey: I was following the format of KLDStat as that is a similar tool. KLDStat is only used in modprobe from the references I can see in VScode. Should I move both KLDLoad and KLDStat into modprobe?

squirrelsc: Yes, please move them both.

r-dailey: I missed one use of kldstat in kernelconfigBSD so I left it there. KLDLoad is now in modprobe.

squirrelsc (approved):

PR #3377, opened 2024-08-12
Minimal shell: honor full chain of hooks, on either minimal entry point

glima:
We have 2 points in which LISA might decide to go with a minimal shell profile:

1) When first contacting a node, issuing a 'cmd' command, checking if
   'Windows' is present in the output or not (rule out Windows shell
   type). A negative case will incur in either the POSIX shell type,
   *if 'Unknown syntax' is not thrown as output*, or an early
   decision for the *'minimal' shell type* is taken and the shell object is
   set to follow that config from there on.

2) When running the actual first lisa suite SSH command on the same
   node. It turns out we've seen cases where 'Unknown syntax' is *not*
   output at early node initialization (1), for some reason. Well, the
   exception path of actual command execution will check internal node
   state, that tracks if a minimal profile has been tested before or
   not. If not used yet and under a command execution exception, set the
   node to minimal type and try again was the logic.

It turns out the 'minification' of the node config was only done
partially, in (2). We missed the full logic of overriding its final
SSH command tokenizer.

Another missed detail is that the final tokenizer logic at the minimal
shell case was too greedy in trying to remove stray quotes. Let's keep
injecting quotes on tokens with spaces on them. That will not hurt any
legit commands from any minimal context.

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -41,11 +41,17 @@ _spawn_initialization_error_pattern = re.compile(
 )
 
 
-def minimal_escape_sh(value: str) -> str:
-    return value.replace("'", "'\\''")
+def _minimal_escape_sh(value: str) -> str:
+    # Tokens iterated here will either have spaces in them, after
+    # tokenized by process_*_command()--process.py--or not. For those
+    # with spaces, we want to inject quotes around them again, on the
+    # minimal shell case--but *only* for those.
+    if re.search(r"\s", value):
+        return f"'{value}'"
+    return value
 
 
-def minimal_generate_run_command(  # type: ignore
+def _minimal_generate_run_command(  # type: ignore
     self,
     command_args: str,
     store_pid: bool,
@@ -53,7 +59,7 @@ def minimal_generate_run_command(  # type: ignore
     update_env: Optional[Dict[str, str]] = None,
     new_process_group: bool = False,
 ) -> str:
-    return " ".join(map(minimal_escape_sh, command_args))
+    return " ".join(map(_minimal_escape_sh, command_args))
 
 
 def wait_tcp_port_ready(
@@ -213,6 +219,20 @@ def _spawn_ssh_process(shell: spur.ssh.SshShell, **kwargs: Any) -> spur.ssh.SshP
     return shell.spawn(**kwargs)
 
 
+def _minimize_shell(shell: spur.ssh.SshShell) -> None:
+    """
+    Dynamically override that object's method. Here, we don't enclose every
+    shell token under single quotes anymore. That's an assumption from spur
+    that minimal shells will still be POSIX compliant--not true for some
+    cases for LISA users.
+    """
+    func_type = type(spur.ssh.ShellTypes.minimal.generate_run_command)
+    shell._spur._shell_type.generate_run_command = func_type(
+        _minimal_generate_run_command,
+        shell._spur._shell_type,
+    )
+
+
 class SshShell(InitializableMixin):
     def __init__(self, connection_info: schema.ConnectionInfo) -> None:
         super().__init__()
@@ -267,7 +287,11 @@ class SshShell(InitializableMixin):
         else:
             self.is_posix = True
             shell_type = spur.ssh.ShellTypes.sh
-            # it doesn't support bash. Use minimal shell type
+            # First chance in getting a clue about no POSIX shell
+            # support (still not Windows). Use minimal shell type if
+            # so. Bear in mind we can get silence here (no "Unknown
+            # syntax"), but there will be a second chance of setting a
+            # minimal shell further down the flow
             if stdout_content and "Unknown syntax" in stdout_content:
                 shell_type = spur.ssh.ShellTypes.minimal
 
@@ -299,15 +323,7 @@ class SshShell(InitializableMixin):
         )
         self._inner_shell = spurplus.SshShell(spur_ssh_shell=spur_ssh_shell, sftp=sftp)
         if shell_type == spur.ssh.ShellTypes.minimal:
-            # Dynamically override that object's method. Here, we don't enclose every
-            # shell token under single quotes anymore. That's an assumption from spur
-            # that minimal shells will still be POSIX compliant--not true for some
-            # cases for LISA users.
-            func_type = type(spur.ssh.ShellTypes.minimal.generate_run_command)
-            self._inner_shell._spur._shell_type.generate_run_command = func_type(
-                minimal_generate_run_command,
-                self._inner_shell._spur._shell_type,
-            )
+            _minimize_shell(self._inner_shell)
 
     def close(self) -> None:
         if self._inner_shell:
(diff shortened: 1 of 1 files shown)


squirrelsc:
@lubaihua33 @LiliDeng LGTM, please verify it.

LiliDeng:
@glima please fix CI check error.

```
nox > flake8 
./lisa/util/shell.py:215:1: BLK100 Black would make changes.
./lisa/util/shell.py:215:1: E302 expected 2 blank lines, found 1
./lisa/util/shell.py:320:5: E303 too many blank lines (2)
```

lisa/util/shell.py
```diff
@@ -380,6 +391,7 @@ def spawn(
                 # Except CommandInitializationError then use minimal shell type.
                 if not have_tried_minimal_type:
                     self._inner_shell._spur._shell_type = spur.ssh.ShellTypes.minimal
+                    _minimize_shell(self._inner_shell)
```
lubaihua33, line 402: @glima, adding this line here will make the images like "azhpc-desktop", "javlinltd" timeout when executing some commands. We can use `_spawn_initialization_error_pattern` to distiguish this scenario because those images can match this pattern.
can you add an "else" branch for "if matched:", and move this line `_minimize_shell(self._inner_shell)` under the "else" branch? 
```
if not have_tried_minimal_type:
    self._inner_shell._spur._shell_type = spur.ssh.ShellTypes.minimal
    have_tried_minimal_type = True
    matched = _spawn_initialization_error_pattern.search(
        str(identifier)
    )
    if matched:
        self.spawn_initialization_error_string = matched.group(
            "linux_profile_error"
        )
    else:
        _minimize_shell(self._inner_shell)
```

glima: Hi @lubaihua33. I'm sorry you see timeouts there, but that needs further debugging. There is nothing in setting the generate_run_command() function on the shell to a simple, continuous string with the individual tokens (no added quotes) that would make a Linux shell unhappy, I'm afraid. But there is definitely a reason non-POSIX shells would reject that, and we need the support.

The regexp will match regardless, as it looks for a pattern added by spur: https://github.com/mwilliamson/spur.py/blob/master/spur/errors.py#L11, so this is not really helping in my case.

I will fix the style errors, but this change is not really sound. I'd kindly ask for your debugging as to why it fails on your side, so we get to a better fix for that case.

Thanks.

lubaihua33: Hi @glima 
**Some background:**
The images whose publishers are "azhpc-desktop", "javlinltd" are the POSIX shells. But calling sqawn to run any Linux commands might raise CommandInitializationError. Using minimal shel type can solve this. I added comments here for the reason https://github.com/microsoft/lisa/blob/main/lisa/util/shell.py#L370.

**The reason that fails on my side:**
Overriding the generate_run_command() with the minimal_generate_run_command will remove some quotes for commands. However, removing some quotes for a command might lead to a timeout for a POSIX shell.

For example, 
The return value of the original fun 'generate_run_command' is
`''sudo' 'sh' '-c' 'cat /etc/shadow | grep -v "lisatest"''` (This is good for a POSIX shell)

The return value of the override one minimal_generate_run_command is
`'sudo sh -c cat /etc/shadow | grep -v "lisatest"`'  (There are no quotes for `cat /etc/shadow | grep -v "lisatest"`. If running `sudo sh -c cat /etc/shadow` without quotes for `cat /etc/shadow` in a POSIX shell, only the first string `cat` will be taken as the value of `-c`. And only running `cat `will lead to a timeout in a POSIX shell.

Those systems of "azhpc-desktop" and "javlinltd" images are POSIX shells and will enter CommandInitializationError exception. Running some commands like the above ones will timeout.

That's why it fails (timeout) for the systems of "azhpc-desktop", "javlinltd" images on my side.

**How to fix**
I have no better way to find a middle ground for both your scenario and my scenario. I just found the system of "azhpc-desktop", "javlinltd" images of my scenario can match _spawn_initialization_error_pattern, however the system of your scenario won't match. So I thought if we can use this to distinguish them. 
This change is not really sound. That would be great if you could find a better way.

glima: I see. I guess I was lazy/greedy in the substitutions there, my bad. What if we only cut off the obviously shlex.split() added quotes there, like on this latest update? I think it will make your case behave again. Ty

lubaihua33: I still got timeout. You can run "verify_no_pre_exist_users" test case for image "javlinltd1579185328273 cloverdx-server-vm-image 6-5-0 1.0.0" to reproduce.

glima: Would you please give me a full cmdline for that? I have very little experience with running upstream lisa alone...

lubaihua33: Let me send you the runbook and the command offline.

glima: Just did it and it passes with my last change on the patch:

```
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner ________________________________________
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner       AzureImageStandard.verify_no_pre_exist_users: PASSED
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner test result summary
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner     TOTAL    : 1
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner     QUEUED   : 0
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner     ASSIGNED : 0
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner     RUNNING  : 0
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner     FAILED   : 0
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner     PASSED   : 1
2024-08-19 16:51:31.450[20688][INFO] lisa.RootRunner     SKIPPED  : 0
2024-08-19 16:51:31.450[8828][DEBUG] lisa.notifier[Console] received message [TestRun]: TestRunMessage(type='TestRun', time=datetime.datetime(2024, 8, 19, 16, 51, 31, 450097), elapsed=193.20371299999533, status=<TestRunStatus.SUCCESS: 3>, runbook_name=None, test_project=None, test_pass=None, tags=None, run_name='lisa-None-20240819-164815-879', message='')
2024-08-19 16:51:31.450[20688][INFO] lisa. completed in 195.570 sec
```

lubaihua33: I run this test on your branch again. It still failed. I have sent you the log hope we can find the gap.

glima: My bad, too many copies of the repo led to a slight confusion/mismatch. OK, this is a version that works for all. escape_sh() from spur was doing basically what I do here now. But that function was not just escaping quotes, it was adding quotes around things too. Now I just do that when spaces are found in the input, in the minimal case.

lubaihua33: Thanks for this change. It works.

glima:
@squirrelsc do I get it in?

LiliDeng:
@glima please fix the CI check error, thanks.

```./lisa/util/shell.py:49:19: BLK100 Black would make changes.```

lubaihua33 (approved):

LiliDeng (approved):

PR #3379, opened 2024-08-13
Add device passthrough support for libvirt/hyperv platforms

smit-gardhariya:
Add device passthrough support for libvirt/hyperv platform. Create device pool based on given vendor/device ID list. Platform will get device from pool when needed while creating node and put it back into pool when node is getting deleted. 

There are changes in LISA tools like lspci to provide information like domain and list of devices for given vendor-device id, hyperv tool to have support to perform start/stop/restart/enable-passthrough for a given vm

Example runbooks for platforms are

```
platform:
  - type: cloud-hypervisor
    admin_private_key_file: '<PATH>'
    keep_environment: "no"
    cloud-hypervisor:
      network_boot_timeout: 120
      hosts:
        - address: "<IP>"
          username: "<USERNAME>"
          private_key_file: '<PATH>'
      capture_libvirt_debug_logs: true
      device_pools:
        - type: "pci_net"
          devices:
            - vendor_id: "8086"
              device_id: "1572"
        - type: "pci_gpu"
          devices:
            - vendor_id: "15b3"
              device_id: "1007"
    requirement:
      core_count: 1
      memory_mb: 1024
      cloud-hypervisor:
        disk_img: '<PATH>'
        disk_img_format: "qcow2"
        firmware: '<PATH>'
        cloud_init:
          extra_user_data: ""
        device_passthrough:
          - pool_type: "pci_net" 
            managed: "yes"
            count: 1
        disk_img_resize_gib: 25
```

```
platform:
  - type: hyperv
    admin_username: <USERNAME>
    admin_password: <PASSWD>
    requirement:
      core_count:
        min: 32
      memory_mb:
        min: 65536
      hyperv:
        hyperv_generation: 2
        osdisk_size_in_gb: $(osdisk_size_in_gb)
        device_passthrough:
          - pool_type: "pci_net"
            count: 1
    hyperv:
      source:
        type: local
        files:
          - source: <LOCAL VHD PATH>
            unzip: true
      extra_args:
        - command: "<CMD1>"
          args: "<CMD1 ARGS>"
        - command: "<CMD2>"
          args: "<CMD2 ARGS>"
      servers:
        - address: <IP>
          username: <USERNAME>
          password: <PASSWD>
      device_pools:
        - type: "pci_net"
          devices:
            - vendor_id: "8086"
              device_id: "1572"

```

### DIFF
--- a/lisa/sut_orchestrator/hyperv/context.py
+++ b/lisa/sut_orchestrator/hyperv/context.py
@@ -1,12 +1,22 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import PurePath
-from typing import Optional
+from typing import List, Optional
 
 from lisa import Node, RemoteNode
+from lisa.sut_orchestrator.hyperv.schema import DeviceAddressSchema
+from lisa.sut_orchestrator.util.schema import HostDevicePoolType
 from lisa.util.process import Process
+
+
+@dataclass
+class DevicePassthroughContext:
+    pool_type: HostDevicePoolType = HostDevicePoolType.PCI_NIC
+    device_list: List[DeviceAddressSchema] = field(
+        default_factory=list,
+    )
 
 
 @dataclass
@@ -15,6 +25,11 @@ class NodeContext:
     host: Optional[RemoteNode] = None
     working_path = PurePath()
     serial_log_process: Optional[Process] = None
+
+    # Device pass through configuration
+    passthrough_devices: List[DevicePassthroughContext] = field(
+        default_factory=list,
+    )
 
     @property
     def console_log_path(self) -> PurePath:
--- a/lisa/sut_orchestrator/hyperv/get_assignable_devices.py
+++ b/lisa/sut_orchestrator/hyperv/get_assignable_devices.py
@@ -1,0 +1,359 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+# Refer: https://learn.microsoft.com/en-us/windows-server/virtualization/hyper-v/deploy/deploying-graphics-devices-using-dda  # noqa E501
+import re
+from typing import Dict, List, Optional
+
+from lisa.node import Node
+from lisa.tools import PowerShell
+from lisa.util import LisaException, find_group_in_lines, find_groups_in_lines
+from lisa.util.logger import Logger
+
+from .schema import DeviceAddressSchema
+
+
+class HypervAssignableDevices:
+    PKEY_DEVICE_TYPE = "{3AB22E31-8264-4b4e-9AF5-A8D2D8E33E62}  1"
+    PKEY_BASE_CLASS = "{3AB22E31-8264-4b4e-9AF5-A8D2D8E33E62}  3"
+    PKEY_REQUIRES_RESERVED_MEMORY_REGION = "{3AB22E31-8264-4b4e-9AF5-A8D2D8E33E62}  34" # noqa E501
+    PKEY_ACS_COMPATIBLE_UP_HIERARCHY = "{3AB22E31-8264-4b4e-9AF5-A8D2D8E33E62}  31" # noqa E501
+    PROP_DEVICE_TYPE_PCI_EXPRESS_ENDPOINT = "2"
+    PROP_DEVICE_TYPE_PCI_EXPRESS_LEGACY_ENDPOINT = "3"
+    PROP_DEVICE_TYPE_PCI_EXPRESS_ROOT_COMPLEX_INTEGRATED_ENDPOINT = "4"
+    PROP_DEVICE_TYPE_PCI_EXPRESS_TREATED_AS_PCI = "5"
+    PROP_ACS_COMPATIBLE_UP_HIERARCHY_NOT_SUPPORTED = "0"
+    PROP_BASE_CLASS_DISPLAY_CTRL = "3"
+
+    def __init__(self, host_node: Node, log: Logger):
+        self.host_node = host_node
+        self.log = log
+        self.pwsh = self.host_node.tools[PowerShell]
+        self.pnp_allocated_resources: List[Dict[str, str]] = (
+            self.__load_pnp_allocated_resources()
+        )
+
+    def get_assignable_devices(
+        self,
+        vendor_id: str,
+        device_id: str,
+    ) -> List[DeviceAddressSchema]:
+        device_id_list = self.__get_devices_by_vendor_device_id(
+            vendor_id=vendor_id, device_id=device_id
+        )
+
+        devices: List[DeviceAddressSchema] = []
+        for rec in device_id_list:
+            device_id = rec["device_id"]
+            result = self.__get_dda_properties(device_id=device_id)
+            if result:
+                result.friendly_name = rec["friendly_name"]
+                devices.append(result)
+        return devices
+
+    def __get_devices_by_vendor_device_id(
+        self,
+        vendor_id: str,
+        device_id: str,
+    ) -> List[Dict[str, str]]:
+        """
+        Get the device ID list for given vendor/device ID combination
+        """
+        devices: List[Dict[str, str]] = []
+        device_regex = re.compile(
+            r"Description\s+:\s*(?P<desc>.+)\n.*DeviceID\s+:\s*(?P<device_id>.+)"
+        )
+
+        cmd = (
+            "Get-WmiObject Win32_PnPEntity -Filter "
+            f"\"DeviceID LIKE 'PCI\\\\VEN_{vendor_id}&DEV_{device_id}%'\""
+        )
+        stdout = self.pwsh.run_cmdlet(
+            cmdlet=cmd,
+            force_run=True,
+            sudo=True,
+        )
+
+        devices_str = stdout.strip().split("\r\n\r\n")
+        filtered_devices = [i.strip() for i in devices_str if i.strip() != ""]
+        for device_properties in filtered_devices:
+            res = find_group_in_lines(
+                lines=device_properties,
+                pattern=device_regex,
+                single_line=False,
+            )
+            if not res:
+                raise LisaException("Can not extract DeviceId/Description")
+
+            devices.append({
+                "device_id": res["device_id"].strip(),
+                "friendly_name": res["desc"].strip(),
+            })
+        return devices
+
+    def __get_pnp_device_property(self, device_id: str, property_name: str) -> str:
+        """
+        Retrieve a PnP device property by instance ID and property key.
+        """
+        cmd = (
+            "(Get-PnpDeviceProperty -InstanceId "
+            f"'{device_id}' '{property_name}').Data"
+        )
+
+        output = self.pwsh.run_cmdlet(
+            cmdlet=cmd,
+            sudo=True,
+            force_run=True,
+        )
+        return output.strip()
+
+    def __load_pnp_allocated_resources(self) -> List[Dict[str, str]]:
+        # Command output result (just 2 device properties)
+        # ========================================================
+        # __GENUS          : 2
+        # __CLASS          : Win32_PNPAllocatedResource
+        # __SUPERCLASS     : CIM_AllocatedResource
+        # __DYNASTY        : CIM_Dependency
+        # __RELPATH        : Win32_PNPAllocatedResource.Antecedent="\\\\WIN-2IDCNC2D5V
+        #                   C\\root\\cimv2:Win32_DeviceMemoryAddress.StartingAddress=\
+        #                   "2463203328\"",Dependent="\\\\WIN-2IDCNC2D5VC\\root\\cimv2:
+        #                   Win32_PnPEntity.DeviceID=\"PCI\\\\VEN_8086&DEV_A1A3&
+        #                   SUBSYS_07161028&REV_09\\\\3&11583659&0&FC\""
+        # __PROPERTY_COUNT : 2
+        # __DERIVATION     : {CIM_AllocatedResource, CIM_Dependency}
+        # __SERVER         : WIN-2IDCNC2D5VC
+        # __NAMESPACE      : root\cimv2
+        # __PATH           : \\WIN-2IDCNC2D5VC\root\cimv2:Win32_PNPAllocatedResource.
+        #                   Antecedent="\\\\WIN-2IDCNC2D5VC\\root\\cimv2:Win32_
+        #                   DeviceMemoryAddress.StartingAddress=\"2463203328\"",
+        #                   Dependent="\\\\WIN-2IDCNC2D5VC\\root\\cimv2:Win32_PnP
+        #                   Entity.DeviceID=\"PCI\\\\VEN_8086&DEV_A1A3&SUBSYS_07161028&
+        #                   REV_09\\\\3&11583659&0&FC\""
+        # Antecedent       : \\WIN-2IDCNC2D5VC\root\cimv2:Win32_DeviceMemoryAddress.
+        #                    StartingAddress="2463203328"
+        # Dependent        : \\WIN-2IDCNC2D5VC\root\cimv2:Win32_PnPEntity.DeviceID=
+        #                   "PCI\\VEN_8086&DEV_A1A3&SUBSYS_07161028&REV_09\\3
+        #                    &11583659&0&FC"
+        # PSComputerName   : WIN-2IDCNC2D5VC
+
+        # __GENUS          : 2
+        # __CLASS          : Win32_PNPAllocatedResource
+        # __SUPERCLASS     : CIM_AllocatedResource
+        # __DYNASTY        : CIM_Dependency
+        # __RELPATH        : Win32_PNPAllocatedResource.Antecedent="\\\\WIN-2IDCNC2D5VC
+        #                   \\root\\cimv2:Win32_PortResource.StartingAddress=\"8192\"",
+        #                   Dependent="\\\\WIN-2IDCNC2D5VC\\root\\cimv2:Win32_PnPEntity
+        #                   .DeviceID=\"PCI\\\\VEN_8086&DEV_A1A3&SUBSYS_07161028&REV_09
+        #                   \\\\3&11583659&0&FC\""
+        # __PROPERTY_COUNT : 2
+        # __DERIVATION     : {CIM_AllocatedResource, CIM_Dependency}
+        # __SERVER         : WIN-2IDCNC2D5VC
+        # __NAMESPACE      : root\cimv2
+        # __PATH           : \\WIN-2IDCNC2D5VC\root\cimv2:Win32_PNPAllocatedResource.
+        #                    Antecedent="\\\\WIN-2IDCNC2D5VC\\root\\cimv2:Win32_PortR
+        #                    esource.StartingAddress=\"8192\"",Dependent="\\\\WIN-2ID
+        #                    CNC2D5VC\\root\\cimv2:Win32_PnPEntity.DeviceID=\"PCI\\\\
+        #                    VEN_8086&DEV_A1A3&SUBSYS_07161028&REV_09\\\\3&11
+        #                    583659&0&FC\""
+        # Antecedent       : \\WIN-2IDCNC2D5VC\root\cimv2:Win32_PortResource.
+        #                    StartingAddress="8192"
+        # Dependent        : \\WIN-2IDCNC2D5VC\root\cimv2:Win32_PnPEntity.DeviceID=
+        #                    "PCI\\VEN_8086&DEV_A1A3&SUBSYS_07161028&REV_09\\3&
+        #                    11583659&0&FC"
+        # PSComputerName   : WIN-2IDCNC2D5VC
+
+        stdout = self.pwsh.run_cmdlet(
+            cmdlet="gwmi -query 'select * from Win32_PnPAllocatedResource'",
+            sudo=True,
+            force_run=True,
+        )
+        pnp_allocated_resources = stdout.strip().split("\r\n\r\n")
+        result: List[Dict[str, str]] = []
+        # Regular expression to match the key-value pairs
+        pattern = re.compile(r'(?P<key>\S+)\s*:\s*(?P<value>.*?)(?=\n\S|\Z)', re.DOTALL)
+
+        for rec in pnp_allocated_resources:
+            extract_val = {}
+            matches = find_groups_in_lines(
+                lines=rec.strip(),
+                pattern=pattern,
+                single_line=False,
+            )
+            if matches:
+                for element in matches:
+                    key = element["key"]
+                    val = element["value"]
+                    val = val.replace(" ", "")
+                    val = val.replace("\r\n", "")
+                    extract_val[key] = val
+                result.append(extract_val)
+        return result
+
+    def __get_mmio_end_address(self, start_addr: str) -> Optional[str]:
+        # MemoryType   Name                  Status
+        # ----------   ----                  ------
+        # WindowDecode 0xE1800000-0xE1BFFFFF OK
+        #             0xE2000000-0xE2000FFF OK
+        # WindowDecode 0xD4000000-0xD43FFFFF OK
+        #             0xD4800000-0xD4800FFF OK
+        #             0xFED1C000-0xFED3FFFF OK
+
+        device_mem_addr = self.pwsh.run_cmdlet(
+            cmdlet="gwmi -query 'select * from Win32_DeviceMemoryAddress'",
+            sudo=True,
+            force_run=True,
+        )
+        end_addr_rec = None
+        for rec in device_mem_addr.splitlines():
+            rec = rec.strip()
+            if rec.find(start_addr) >= 0:
+                addr = rec.split("-")
+                start_addr_rec = addr[0].split()[-1]
+                end_addr_rec = addr[1].split()[0].strip()
+
+                err = "MMIO Starting address not matching"
+                assert start_addr == start_addr_rec, err
+                break
+        return end_addr_rec
+
+    def __get_dda_properties(self, device_id: str) -> Optional[DeviceAddressSchema]:
+        """
+        Determine if a PCI device is assignable using Discrete Device Assignment (DDA)
+        If so, get DDA proerprties like locationpath, device-id, friendly name
+        """
+        self.log.debug(f"PCI InstanceId: {device_id}")
+
+        rmrr = self.__get_pnp_device_property(
+            device_id=device_id,
+            property_name=self.PKEY_REQUIRES_RESERVED_MEMORY_REGION,
+        )
+        rmrr = rmrr.strip()
+        if rmrr != "False":
+            self.log.debug(
+                "BIOS requires that this device remain attached to BIOS-owned memory."
+                "Not assignable."
+            )
+            return None
+
+        acs_up = self.__get_pnp_device_property(
+            device_id=device_id,
+            property_name=self.PKEY_ACS_COMPATIBLE_UP_HIERARCHY,
+        )
+        acs_up = acs_up.strip()
+        if acs_up == self.PROP_ACS_COMPATIBLE_UP_HIERARCHY_NOT_SUPPORTED:
+            self.log.debug(
+                "Traffic from this device may be redirected to other devices in "
+                "the system. Not assignable."
+            )
+            return None
+
+        dev_type = self.__get_pnp_device_property(
+            device_id=device_id,
+            property_name=self.PKEY_DEVICE_TYPE
+        )
+        dev_type = dev_type.strip()
+        if dev_type == self.PROP_DEVICE_TYPE_PCI_EXPRESS_ENDPOINT:
+            self.log.debug("Express Endpoint -- more secure.")
+        else:
+            if dev_type == (
+                self.PROP_DEVICE_TYPE_PCI_EXPRESS_ROOT_COMPLEX_INTEGRATED_ENDPOINT
+            ):
+                self.log.debug("Embedded Endpoint -- less secure.")
+            elif dev_type == self.PROP_DEVICE_TYPE_PCI_EXPRESS_LEGACY_ENDPOINT:
+                dev_base_class = self.__get_pnp_device_property(
+                    device_id=device_id,
+                    property_name=self.PKEY_BASE_CLASS,
+                )
+                dev_base_class = dev_base_class.strip()
+                if dev_base_class == self.PROP_BASE_CLASS_DISPLAY_CTRL:
+                    self.log.debug("Legacy Express Endpoint -- graphics controller.")
+                else:
+                    self.log.debug("Legacy, non-VGA PCI device. Not assignable.")
+                    return None
+            else:
+                if dev_type == self.PROP_DEVICE_TYPE_PCI_EXPRESS_TREATED_AS_PCI:
+                    self.log.debug(
+                        "BIOS kept control of PCI Express for this device. "
+                        "Not assignable."
+                    )
+                else:
+                    self.log.debug(
+                        "Old-style PCI device, switch port, etc. "
+                        "Not assignable."
+                    )
+                return None
+
+        # Get the device location path
+        location_path = self.__get_pnp_device_property(
+            device_id=device_id,
+            property_name="DEVPKEY_Device_LocationPaths",
+        )
+        location_path = location_path.strip().splitlines()[0]
+        self.log.debug(f"Device locationpath: {location_path}")
+        assert location_path.find("PCI") == 0, "Location path is wrong"
+
+        cmd = (
+            "(Get-PnpDevice -PresentOnly -InstanceId "
+            f"'{device_id}').ConfigManagerErrorCode"
+        )
+        conf_mng_err_code = self.pwsh.run_cmdlet(
+            cmdlet=cmd,
+            force_run=True,
+            sudo=True,
+        )
+        conf_mng_err_code = conf_mng_err_code.strip()
+        self.log.debug(f"ConfigManagerErrorCode: {conf_mng_err_code}")
+        if conf_mng_err_code == "CM_PROB_DISABLED":
+            self.log.debug(
+                "Device is Disabled, unable to check resource requirements, "
+                "it may be assignable."
+            )
+            self.log.debug("Enable the device and rerun this script to confirm.")
+            return None
+
+        irq_assignements = [
+            i for i in self.pnp_allocated_resources
+            if i["Dependent"].find(device_id.replace("\\", "\\\\")) >= 0
+        ]
+        if irq_assignements:
+            msi_assignments = [
+                i for i in self.pnp_allocated_resources
+                if i["Antecedent"].find("IRQNumber=42949") >= 0
+            ]
+            if not msi_assignments:
+                self.log.debug(
+                    "All of the interrupts are line-based, no assignment can work."
+                )
+                return None
+            else:
+                self.log.debug("Its interrupts are message-based, assignment can work.")
+        else:
+            self.log.debug("It has no interrupts at all -- assignment can work.")
+
+        mmio_assignments = [
+            i for i in self.pnp_allocated_resources
+            if i["Dependent"].find(device_id.replace("\\", "\\\\")) >= 0
+            and i["__RELPATH"].find("Win32_DeviceMemoryAddres") >= 0
+        ]
+        mmio_total = 0
+        if mmio_assignments:
+            for rec in mmio_assignments:
+                antecedent_val = rec["Antecedent"]
+                addresses = antecedent_val.split('"')
+                assert len(addresses) >= 2, "Antecedent: Can't get MMIO Start Address"
+                start_address = hex(int(addresses[1].strip())).upper()
+                start_address_hex = start_address.replace("X", "x")
+                end_address = self.__get_mmio_end_address(start_address_hex)
+                assert end_address, "Can not get MMIO End Address"
+
+                mmio = int(end_address, 16) - int(start_address, 16)
+                mmio_total += mmio
+            if mmio_total:
+                mmio_total = round(mmio_total / (1024 * 1024))
+                self.log.debug(f"Device '{device_id}', Total MMIO = {mmio_total}MB ")
+        else:
+            self.log.debug("It has no MMIO space")
+
+        device = DeviceAddressSchema()
+        device.location_path = location_path
+        device.instance_id = device_id
+        return device
--- a/lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
+++ b/lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
@@ -1,0 +1,220 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from typing import Dict, List, Optional
+
+from lisa.node import RemoteNode
+from lisa.sut_orchestrator.hyperv.get_assignable_devices import HypervAssignableDevices
+from lisa.sut_orchestrator.hyperv.schema import (
+    DeviceAddressSchema,
+    HypervNodeSchema,
+    HypervPlatformSchema,
+)
+from lisa.sut_orchestrator.util.device_pool import BaseDevicePool
+from lisa.sut_orchestrator.util.schema import HostDevicePoolSchema, HostDevicePoolType
+from lisa.tools import HyperV, PowerShell
+from lisa.util import ResourceAwaitableException
+from lisa.util.logger import Logger
+
+from .context import DevicePassthroughContext, NodeContext
+
+
+class HyperVDevicePool(BaseDevicePool):
+    def __init__(
+        self,
+        node: RemoteNode,
+        runbook: HypervPlatformSchema,
+        log: Logger,
+    ) -> None:
+        # Device Passthrough configs
+        # Mapping of Host Device Passthrough
+        self.available_host_devices: Dict[
+            HostDevicePoolType, List[DeviceAddressSchema]
+        ] = {}
+        self.supported_pool_type = [
+            HostDevicePoolType.PCI_NIC,
+            HostDevicePoolType.PCI_GPU,
+        ]
+        self._server = node
+        self._hyperv_runbook = runbook
+        self.log = log
+
+    def create_device_pool(
+        self,
+        pool_type: HostDevicePoolType,
+        vendor_id: str,
+        device_id: str,
+    ) -> None:
+        hv_dev = HypervAssignableDevices(
+            host_node=self._server,
+            log=self.log,
+        )
+        devices = hv_dev.get_assignable_devices(
+            vendor_id=vendor_id,
+            device_id=device_id,
+        )
+        primary_nic_id_list = self.get_primary_nic_id()
+        pool = self.available_host_devices.get(pool_type, [])
+        for dev in devices:
+            if dev.instance_id not in primary_nic_id_list:
+                pool.append(dev)
+        self.available_host_devices[pool_type] = pool
+
+    def request_devices(
+        self,
+        pool_type: HostDevicePoolType,
+        count: int,
+    ) -> List[DeviceAddressSchema]:
+        pool = self.available_host_devices[pool_type]
+        if len(pool) < count:
+            raise ResourceAwaitableException(
+                f"Not enough devices are available under pool: {pool_type}. "
+                f"Required count is {count}"
+            )
+        devices = pool[:count]
+
+        # Update the pool
+        pool = pool[count:]
+        self.available_host_devices[pool_type] = pool
+
+        return devices
+
+    def release_devices(
+        self,
+        node_context: NodeContext,
+    ) -> None:
+        vm_name = node_context.vm_name
+        devices_ctx = node_context.passthrough_devices
+        confing_commands = []
+        for ctx in devices_ctx:
+            for device in ctx.device_list:
+                confing_commands.append(
+                    f"Remove-VMAssignableDevice "
+                    f"-LocationPath '{device.location_path}' -VMName '{vm_name}'"
+                )
+                confing_commands.append(
+                    f"Mount-VMHostAssignableDevice -LocationPath "
+                    f"'{device.location_path}'"
+                )
+                confing_commands.append(
+                    f"Enable-PnpDevice -InstanceId '{device.instance_id}' "
+                    "-Confirm:$false"
+                )
+
+        powershell = self._server.tools[PowerShell]
+        for cmd in confing_commands:
+            powershell.run_cmdlet(
+                cmdlet=cmd,
+                force_run=True,
+            )
+
+    def get_primary_nic_id(self) -> List[str]:
+        powershell = self._server.tools[PowerShell]
+        ip: str = self._server.public_address
+
+        # Get the NIC name via IP.
+        # We will get vEthernet switch interface name, not actual NIC for baremetal
+        cmd = (
+            "(Get-NetAdapter | Get-NetIPAddress | Where-Object "
+            f"{{ $_.IPAddress -eq '{ip}' }}).InterfaceAlias"
+        )
+        interface_name = powershell.run_cmdlet(
+            cmdlet=cmd,
+            force_run=True,
+        )
+
+        # Get the MAC for above interface
+        cmd = (
+            "(Get-NetAdapter | Where-Object "
+            f"{{ $_.Name -eq '{interface_name}' }}).MacAddress"
+        )
+        mac_address = powershell.run_cmdlet(
+            cmdlet=cmd,
+            force_run=True,
+        )
+
+        # Get all interfaces for above MAC Address
+        cmd = (
+            "(Get-NetAdapter | Where-Object "
+            f"{{ $_.MacAddress -eq '{mac_address}' }}).Name"
+        )
+        inf_names_str = powershell.run_cmdlet(
+            cmdlet=cmd,
+            force_run=True,
+        )
+        inf_names: List[str] = inf_names_str.strip().splitlines()
+
+        # Get device id for all above interface names we got
+        pnp_device_id_list: List[str] = []
+        for name in inf_names:
+            cmd = (
+                "(Get-NetAdapter | Where-Object "
+                f"{{ $_.Name -eq '{name}' }}).PnPDeviceID"
+            )
+            interface_device_id = powershell.run_cmdlet(
+                cmdlet=cmd,
+                force_run=True,
+            )
+            interface_device_id = interface_device_id.strip()
+            pnp_device_id_list.append(interface_device_id)
+
+        return pnp_device_id_list
+
+    def configure_device_passthrough_pool(
+        self,
+        device_configs: Optional[List[HostDevicePoolSchema]],
+    ) -> None:
+        super().configure_device_passthrough_pool(
+            device_configs=device_configs,
+        )
+
+    def _assign_devices_to_vm(
+        self,
+        vm_name: str,
+        devices: List[DeviceAddressSchema],
+    ) -> None:
+        # Assign the devices to the VM
+        confing_commands = []
+        for device in devices:
+            confing_commands.append(
+                f"Disable-PnpDevice -InstanceId '{device.instance_id}' -Confirm:$false"
+            )
+            confing_commands.append(
+                f"Dismount-VMHostAssignableDevice -Force "
+                f"-LocationPath '{device.location_path}'"
+            )
+            confing_commands.append(
+                f"Add-VMAssignableDevice -LocationPath '{device.location_path}' "
+                f"-VMName '{vm_name}'"
+            )
+
+        powershell = self._server.tools[PowerShell]
+        for cmd in confing_commands:
+            powershell.run_cmdlet(
+                cmdlet=cmd,
+                force_run=True,
+            )
+
+    def _set_device_passthrough_node_context(
+        self,
+        node_context: NodeContext,
+        node_runbook: HypervNodeSchema,
+        hv: HyperV,
+        vm_name: str,
+    ) -> None:
+        if not node_runbook.device_passthrough:
+            return
+        hv.enable_device_passthrough(name=vm_name)
+
+        for config in node_runbook.device_passthrough:
+            devices = self.request_devices(
+                pool_type=config.pool_type,
+                count=config.count,
+            )
+            self._assign_devices_to_vm(
+                vm_name=vm_name,
+                devices=devices,
+            )
+            device_context = DevicePassthroughContext()
+            device_context.pool_type = config.pool_type
+            device_context.device_list = devices
+            node_context.passthrough_devices.append(device_context)
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -285,6 +285,8 @@ class HypervPlatform(Platform):
                 },
                 extra_args=extra_args,
             )
+            # Start the VM
+            hv.start_vm(name=vm_name, extra_args=extra_args)
 
             ip_addr = hv.get_ip_address(vm_name)
             username = self.runbook.admin_username
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -1,6 +1,5 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
-
 from functools import partial
 from pathlib import PurePath
 from typing import Any, List, Optional, Type, cast
@@ -17,6 +16,7 @@ from lisa.util.subclasses import Factory
 
 from .. import HYPERV
 from .context import NodeContext, get_node_context
+from .hyperv_device_pool import HyperVDevicePool
 from .schema import HypervNodeSchema, HypervPlatformSchema
 from .serial_console import SerialConsole, SerialConsoleLogger
 from .source import Source
@@ -44,6 +44,12 @@ class HypervPlatform(Platform):
         self._source_vhd: Optional[PurePath] = None
         self._source_factory = Factory[Source](Source)
         self._source_files: Optional[List[PurePath]] = None
+
+        self.device_pool = HyperVDevicePool(
+            node=self._server,
+            runbook=self._hyperv_runbook,
+            log=self._log,
+        )
 
     def _get_hyperv_runbook(self) -> HypervPlatformSchema:
         hyperv_runbook = self.runbook.get_extended_runbook(HypervPlatformSchema)
@@ -97,6 +103,12 @@ class HypervPlatform(Platform):
             return False
 
         environment.runbook.nodes_requirement = nodes_requirement
+
+        # If Device_passthrough is set in runbook,
+        # Configure device passthrough params / Refresh the pool
+        self.device_pool.configure_device_passthrough_pool(
+            self._hyperv_runbook.device_pools,
+        )
         return True
 
     def _get_host_capabilities(self, log: Logger) -> _HostCapabilities:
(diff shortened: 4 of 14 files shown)


lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+            for r in range(1, len(keys) + 1):
+                for combo in combinations(keys, r):
+                    if sum(len(pool.get(key, [])) for key in combo) >= count:
+                        results.append(combo)
+                        break
+                if results:
+                    break
+
+        if not results:
+            raise SkippedException(
```
squirrelsc: Skipped means the test condition doesn't fit, it doesn't need to test, and nothing need to be worried.

SkippedException should be used in test suite layer only. It makes test cases skipped. If it's triggered in the platform, it may cause all test cases are skipped.

The issue here is more like no resource or the environment breaks, it shouldn't be ignorable for test cases. So it's better to raise a normal exception and fail all related test cases.

lisa/sut_orchestrator/hyperv/context.py
```diff
@@ -1,21 +1,37 @@
...
+
 @dataclass
 class NodeContext:
     vm_name: str = ""
     host: Optional[RemoteNode] = None
     working_path = PurePath()
     serial_log_process: Optional[Process] = None
 
+    # Device pass through configuration
+    is_device_passthrough_set: bool = False
```
squirrelsc: It looks the bool is not needed. You can just use the list below. If it's empty, it means no set.

lisa/sut_orchestrator/hyperv/get_assignable_devices.ps1
```diff
@@ -0,0 +1,216 @@
+param (
```
squirrelsc: Can you convert it to Python? The mixed scripts are hard to maintain and troubleshooting. Currently, LISA has only 2 PowerShell scripts, one is to enable SSH before connection built, another is for reliability to generate serial log from named pipes.

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -0,0 +1,63 @@
+from typing import Any, List, Optional
+
+from lisa.sut_orchestrator.util.schema import HostDevicePoolSchema, HostDevicePoolType
+from lisa.util import LisaException, SkippedException
+
+
+class BaseDevicePoolImpl:
```
squirrelsc: What's the `Impl` for? Call it like `BaseDevicePool`.

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -0,0 +1,63 @@
...
+    def _create_device_pool(
```
squirrelsc: The prefix `_` means protected or private methods, which means call them only in this class or subclasses. This base class has only private methods, it's not a normal design.

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -0,0 +1,63 @@
...
+        self,
+        pool_type: HostDevicePoolType,
+        vendor_id: str,
+        device_id: str,
+    ) -> None:
+        raise NotImplementedError("'_create_device_pool' is not implemented")
```
squirrelsc: `raise NotImplementedError` is enough. The call trace has more details, and you don't need to maintain the string.

lisa/tools/lspci.py
```diff
@@ -68,10 +75,28 @@ def __str__(self) -> str:
     def parse(self, raw_str: str) -> None:
         matched_pci_device_info = PATTERN_PCI_DEVICE.match(raw_str)
```
squirrelsc: Use LISA built-in regexp wrappers, it will save lines on matches.

lisa/tools/lspci.py
```diff
@@ -184,6 +209,19 @@ def get_gpu_devices(self, force_run: bool = False) -> List[PciDevice]:
         ]
         return gpu_device_list
 
+    def get_devices_by_vendor_device_id(
+        self,
+        vendor_id: str,
+        device_id: str,
+        force_run: bool = False,
+    ) -> List[PciDevice]:
+        full_list = self.get_devices(force_run)
```
squirrelsc: When call methods, use named arguments, so it's compatible with future changes.

lisa/tools/lspci.py
```diff
@@ -36,6 +36,13 @@
     re.MULTILINE,
 )
 
+# With -mnn option, result would be with vendor/device id
+# d8:00.0 "Ethernet controller [0200]" "Mellanox Technologies [15b3]"
+#   "MT27520 Family [ConnectX-3 Pro] [1007]" "Mellanox Technologies [15b3]"
+#   "Mellanox Technologies ConnectX-3 Pro Stand-up dual-port 40GbE MCX314A-BCCT [0006]"
+PATTERN_DEVICE_ID = re.compile(r"\[([^\]]{4})\]")
```
squirrelsc: Use named pattern, it makes the code easier to read.

lisa/tools/hyperv.py
```diff
@@ -170,6 +176,29 @@ def create_vm(
...
+        self._run_hyperv_cmdlet("Stop-VM", f"-Name {name} -Force", force_run=True)
+
+    def restart_vm(
+        self,
+        name: str,
+    ) -> None:
+        # restart vm
+        self._run_hyperv_cmdlet("Restart-VM", f"-Name {name} -Force", force_run=True)
+
+    def enable_device_passthrough(self, name: str, mmio_mb: int = 5120) -> None:
```
squirrelsc, line 190: 1. When is it need to be enabled? Does it need to disable?
2. the default value of mmio_mb, is it recommended? If it uses the physical memory, it maybe too many to impact VM mem usage.

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def _get_primary_nic_id(self) -> List[str]:
+        raise NotImplementedError("'_get_primary_nic_id' is not implemented")
+
+    def _get_devices_from_pool(
```
squirrelsc: 1. how about `request_devices` or `assign_devices`? It looks the method is to request a number of devices, and assign to a VM. The `get` is like a read-only method. The `from_pool` is not needed, because the class is for pool.
2. To align with next `put_devices_into_pool`, this method should directly assign devices, instead of return a list, and assign in another class.

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -0,0 +1,63 @@
...
+        self,
+        pool_type: HostDevicePoolType,
+        count: int,
+    ) -> Any:
+        raise NotImplementedError("'_get_devices_from_pool' is not implemented")
+
+    def _put_devices_into_pool(
```
squirrelsc: How about `release_devices`?

lisa/sut_orchestrator/hyperv/context.py
```diff
@@ -1,21 +1,37 @@
...
+    device_passthrough_context: List[DevicePassthroughContext] = field(
```
squirrelsc: 1. `passthrough_devices` is a better name.
2. Because the devices are mainly used by types, so the type `dict[HostDevicePoolType, List[DeviceAddressScehma]` is easier to use.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +378,255 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
                 for node in environment.nodes.list()
             ]
         )
+
+
+class HyperVDevicePoolImpl(BaseDevicePoolImpl):
```
squirrelsc: `HyperVDevicePool` is enough in the name.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +378,255 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+        stdout = powershell.run_cmdlet(
+            cmdlet=cmdlet,
+            sudo=True,
+            force_run=True,
+        ).strip()
+
+        if stdout.find("The list is empty") >= 0:
+            return
+        else:
+            match = re.search(
```
squirrelsc: Add examples above regexp.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +378,255 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+            f"{self._assignable_devices_script} "
+            f"-vendorId {vendor_id} -deviceId {device_id}"
+        )
+        stdout = powershell.run_cmdlet(
+            cmdlet=cmdlet,
+            sudo=True,
+            force_run=True,
+        ).strip()
+
+        if stdout.find("The list is empty") >= 0:
```
squirrelsc: Move the whole logic into Python, it doesn't need the second time conversion.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +378,255 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+                        self.AVAILABLE_HOST_DEVICES[pool_type] = pool
+
+    def _get_devices_from_pool(
+        self,
+        pool_type: HostDevicePoolType,
+        count: int,
+    ) -> List[DeviceAddressSchema]:
+        pool = self.AVAILABLE_HOST_DEVICES[pool_type]
+        if len(pool) < count:
+            raise SkippedException(
```
squirrelsc: It should raise exceptions, so the runbook and test cases should be changed to avoid test case cannot run, because of lack of resources. You can also raise `ResourceAwaitableException`, the lisa runner will follow the retry mechanism, and it may be assigned next time.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +378,255 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+                f"Dismount-VMHostAssignableDevice -Force "
+                f"-LocationPath '{device.location_path}'"
+            )
+            confing_commands.append(
+                f"Add-VMAssignableDevice -LocationPath '{device.location_path}' "
+                f"-VMName '{vm_name}'"
+            )
+        self._run_powershell_commands(confing_commands)
+
+    def _run_powershell_commands(
```
squirrelsc: It looks not saving much lines, by running command in a batch. And it's hard to troubleshooting by lack of call trace and not check exit code.

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -0,0 +1,63 @@
...
+            pool_types_from_runbook = [config.type for config in device_configs]
+            for pool_type in pool_types_from_runbook:
+                if pool_type not in supported_pool_type:
+                    raise LisaException(
+                        f"Pool type '{pool_type}' is not supported by platform"
+                    )
+            for config in device_configs:
+                vendor_device_list = config.devices
+                if len(vendor_device_list) > 1:
+                    raise SkippedException(
```
squirrelsc: It looks need to raise an exception to fail the case, instead of skipping.

squirrelsc:
@cwize1 please help reviewing for libvirts changes. Thanks.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
             )
             with open(str(libvirt_log_local_path), "w") as f:
                 f.write(libvirt_log)
+
+
+class LibvirtDevicePoolImpl(BaseDevicePoolImpl):
```
cwize1: Give this class its own file.

lisa/sut_orchestrator/util/device_pool.py
cwize1: Please do not use object inheritance lightly. As best as I can tell, the `BaseDevicePoolImpl` class is not providing an abstraction point. It is just an overly complex way of providing helper utilities.

Please remove this class. If you want the helper utilities, then add them as normal flat functions.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+        return addr
+
+    def _get_devices_from_pool(
+        self,
+        pool_type: HostDevicePoolType,
+        count: int,
+    ) -> List[DeviceAddressSchema]:
+        pool = self.AVAILABLE_HOST_DEVICES.get(pool_type, {})
+
+        # iommu_group: str = ""
```
cwize1: Why does this commented out code exist?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
             )
             with open(str(libvirt_log_local_path), "w") as f:
                 f.write(libvirt_log)
+
+
+class LibvirtDevicePoolImpl(BaseDevicePoolImpl):
+    def __init__(
+        self,
+        node: Union[Node, RemoteNode],
```
cwize1: `RemoteNode` inherits from `Node`. So, you don't need the `Union` here. You can just say `Node`.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+        self,
+        device_configs: Optional[List[HostDevicePoolSchema]],
+        supported_pool_type: Optional[List[HostDevicePoolType]],
+    ) -> None:
+        if device_configs:
+            # Check if host support device passthrough
+            self._check_passthrough_support(self.host_node)
+
+            if not supported_pool_type:
+                supported_pool_type = self.SUPPORTED_HOST_DEVICE_POOLTYPE
```
cwize1: Should this be a set intersection between `supported_pool_type` and `SUPPORTED_HOST_DEVICE_POOLTYPE`?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +378,255 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+        node: RemoteNode,
+        assignable_device_script: PurePath,
+        runbook: HypervPlatformSchema,
+    ) -> None:
+        # Device Passthrough configs
+        # Mapping of Host Device Passthrough
+        self.AVAILABLE_HOST_DEVICES: Dict[
+            HostDevicePoolType, List[DeviceAddressSchema]
+        ] = {}
+        self.SUPPORTED_HOST_DEVICE_POOLTYPE = [
```
cwize1: Should this be a static variable instead of a member variable?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +378,255 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+class HyperVDevicePoolImpl(BaseDevicePoolImpl):
+    def __init__(
+        self,
+        node: RemoteNode,
+        assignable_device_script: PurePath,
+        runbook: HypervPlatformSchema,
+    ) -> None:
+        # Device Passthrough configs
+        # Mapping of Host Device Passthrough
+        self.AVAILABLE_HOST_DEVICES: Dict[
```
cwize1: SCREAMING_SNAKE_CASE is for static/global variables. Member variables should use snake_case.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+        for context in device_context:
+            devices = context.device_list
+            for device in devices:
+                err = f"Kernel driver is not vfio-pci for device: {device}"
+                pool_type = context.pool_type.value
+                if context.managed == "yes" and "pci" in pool_type:
+                    is_vfio_pci = self._is_driver_vfio_pci(device)
+                    assert is_vfio_pci, err
+
+    def _check_passthrough_support(self, node: Node) -> None:
```
cwize1: Might be worth renaming `node` to `host_node`, just to clarify this is running on the host and not a test VM.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+            vendor_id=vendor_id,
+            device_id=device_id,
+        )
+        primary_nic_iommu = self._get_primary_nic_id()
+        for item in device_list:
+            device = self.POOL_TYPE_TO_DEVICE_PROPERTY[pool_type]()
+            slot_info = item.slot.split(":")
+
+            device.domain = item.domain
+            device.bus = slot_info[0]
```
cwize1: You should have a `_parse_pci_address_str` function and place it next to the `_get_pci_address_str` function.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+                pool = self.AVAILABLE_HOST_DEVICES.get(pool_type, {})
+                devices = pool.get(iommu_group, [])
+                devices.append(device)
+                pool[iommu_group] = devices
+                self.AVAILABLE_HOST_DEVICES[pool_type] = pool
+
+    def _get_device_iommu_group(self, device: DeviceAddressSchema) -> str:
+        iommu_pattern = re.compile(r"/sys/kernel/iommu_groups/(?P<id>\d+)/devices/.*")
+        device_id = self._get_pci_address_str(device, True)
+        command: str = (
```
cwize1: Don't use shell piping, awk, or grep. Run each command individually and process the output in Python.

Similar comment elsewhere.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+            if not allow_unsafe_interrupt:
+                raise LisaException("Allowing unsafe interrupt failed")
+
+    def _add_device_passthrough_xml(
+        self,
+        devices: ET.Element,
+        node_context: NodeContext,
+    ) -> ET.Element:
+        for context in node_context.device_passthrough_context:
+            print(context.device_list)
```
cwize1: Don't use `print`. Use the logger that is used throughout the rest of the code.

lisa/sut_orchestrator/libvirt/platform.py
cwize1: Rename `node` to `host_node`.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+            )
+
+        devices: List[DeviceAddressSchema] = []
+        selected_pools = results[0]
+        for iommu_grp in selected_pools:
+            devices += pool.pop(iommu_grp)
+        self.AVAILABLE_HOST_DEVICES[pool_type] = pool
+        return devices
+
+    def _put_devices_into_pool(
```
cwize1: Perhaps rename `_put_devices_into_pool` to `_return_devices_to_pool`?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -537,6 +568,13 @@ def _create_domain_and_attach_logger(
         # Start the VM.
         node_context.domain.resume()
 
+        # Once libvirt domain is created, check if driver attached to device
```
cwize1, line 553: Add `if node_context.is_device_passthrough_set:` check here to match the `ch_platform.py`'s code?

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+        kernel_module = lspci.get_used_module(device_addr_str)
+        return kernel_module == "vfio-pci"
+
+    def _set_device_passthrough_node_context(
+        self,
+        node_context: NodeContext,
+        node_runbook: BaseLibvirtNodeSchema,
+    ) -> None:
+        if node_runbook.device_passthrough:
+            node_context.is_device_passthrough_set = True
```
cwize1: Is the `is_device_passthrough_set` variable needed? Could you do the same check by looking at if the `device_passthrough_context` list is empty or not?

lisa/sut_orchestrator/hyperv/context.py
smit-gardhariya: > 1. `passthrough_devices` is a better name.
> 2. Because the devices are mainly used by types, so the type `dict[HostDevicePoolType, List[DeviceAddressScehma]` is easier to use.

@squirrelsc 

1> Sure, I will make the change for name as `passthrough_devices `

2> Idea is to set the list of devices for given node context and put only devices which will be passthrough. We dont need to track pool type here. Whatever is there in list, we will assign it to node. Pool type is needed while requesting device from the pool. Node context does not need the pool type.

lisa/sut_orchestrator/hyperv/get_assignable_devices.ps1
smit-gardhariya: Sure, I will convert this in python.

I have integrated the already existing script in Microsoft document for DDA device. Idea was to not change much in the file.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: Sure, will change the name

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @cwize1 

Done

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @cwize1 

Yes, idea is to put platform level check for supported pool type. I will move this out of constructor as static member.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: Sure will do it

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: We won't need this once PowerShell script would be converted to python. Here we are running PowerShell script and fetch the list of assignable devices from stdout.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: Sure Chi, will use `ResourceAwaitableException`

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @squirrelsc 

Sure, I will remove `_run_powershell_commands()`. There were set of commands we have to execute for each device while assigning/removing it to/from vm. That is why i created function to execute the commands in batch.

We call `run_cmdlet()` of PowerShell tool and by default it has `fail_on_error: bool = True`

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: Sure Chi, Will change the class name. I used Impl as shortform of implementation.

smit-gardhariya: @cwize1 

There are common functionalities between 2 platform when it comes to device pool implementation like creating device pool, get/put devices from/into the pool, getting primary NIC etc.

But the implementation is platform specific as hyperv/libvirt has some fundamental differences when it comes to device passthrough implementation. Due to this nature of the requirement, It seems having proper parent class for device pool make sense here. This would even help to eliminate duplicate code for device pool functionality between the SUT platforms.

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: @squirrelsc 

Only class or child-subclasses are using that method. Each platform has the device pool implementation which inherit the `BaseDevicePool` for common functionality. `BaseDevicePool` methods are protected so child class can use it and that is why i used `"_"` with function names. None of the method are private, otherwise i would have used `"__"`

We used companion class to reduce the ancestors for child classes like `CloudHypervisorPlatform `and `QemuPlatform`

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: Sure Chi, will remove the string msg.

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: @squirrelsc

Sure, will change name to `release_devices `and `request_devices `

Assignment is a bit different for libvirt/hyperv platform. Libvirt construct the XML as per need using `hostdev` and no action is needed on host to be performed before-hand whereas, for hyperv platform, we have to remove/dismount device from host and assign it to vm. So having different function would make sense and easy to maintain for any before-hand action for hyperv DDA.

Same thing applied when you are about to put device back to the pool. You again have to remove device from VM and mound/enable it on the host. Libivirt does this on his own where we have to do this explicitly for hyperv platform. Having separate function would make it easy to maintain.

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: Done

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: Done

lisa/tools/hyperv.py
smit-gardhariya: @squirrelsc 

`enable_device_passthrough` is to set config so VM can DDA supportable. We will execute this when there is a device pool configured from runbook. When Node will be deleted, all the devices would be removed from it and assigned back to host by platform.

MMIO space is virtual one can be assigned to any value. I have seen mostly in MBs but we are setting it as default 5GB as of now. It won't reserve anything on physical machine.

lisa/tools/lspci.py
smit-gardhariya: Done

lisa/tools/lspci.py
smit-gardhariya: Sure Chi, will use `PATTERN_DEVICE_ID = re.compile(r"\[(?P<id>[^\]]{4})\]")`

This is to fetch vendor/device id from vendor/device_info.

lisa/tools/lspci.py
smit-gardhariya: @squirrelsc 

This line was not part for this code change and been this way since beginning of this work. Do you want me to refactor this too.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1 We have removed `is_device_passthrough_set` now.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Corrected

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1

Idea is to implement device pool type support at platform specific class. If not provided, we will take what we have as statically defined

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: That was missed line. Will remove it. Added for troubleshooting.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Again, missed to remove it. Will remove it.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: I have used `ResourceAwaitableException` as you suggested.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1 I followed the name changes suggested by Chi and used like `request_device `and `release_devices`

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Sure Chris

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @squirrelsc @cwize1 Used length of the list now.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1 
Sure, I will create it as you suggested.

There is only one occurrence to parse the string of PCI address to get bus, slot, function. That is why I did not create the separate function for it.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Sure @cwize1

lisa/tools/lspci.py
```diff
@@ -184,6 +209,19 @@ def get_gpu_devices(self, force_run: bool = False) -> List[PciDevice]:
...
+    def get_devices_by_vendor_device_id(
+        self,
+        vendor_id: str,
+        device_id: str,
+        force_run: bool = False,
+    ) -> List[PciDevice]:
+        full_list = self.get_devices(force_run)
+        devices_list = []
+        for device in full_list:
+            if device.device_id == device_id and device.vendor_id == vendor_id:
```
pupacha, line 239: Instead of fetching all the devices and filtering from there, we could simply run `lspci -Dd <vendorID>:<deviceID>` and pare it.

lisa/tools/hyperv.py
```diff
@@ -148,6 +149,11 @@ def create_vm(
                     force_run=True,
                 )
 
+    def start_vm(
```
pupacha, line 152: IIUC, `start_vm` is extracted out from `create_vm`. So, for this commit, either `start_vm()` needs to be present at end of `create_vm` or an explicit call to `start_vm` needs to be made wherever `create_vm` is called.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -179,6 +194,10 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self.__init_libvirt_conn_string()
         self.libvirt_conn = libvirt.open(self.libvirt_conn_str)
 
+        self.device_pool_impl = LibvirtDevicePoolImpl(
```
pupacha: ` self.device_pool` will be sufficient.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -1341,3 +1388,316 @@ def _capture_libvirt_logs(self) -> None:
...
+
+
+class LibvirtDevicePoolImpl(BaseDevicePoolImpl):
+    def __init__(
+        self,
+        node: Union[Node, RemoteNode],
+        runbook: BaseLibvirtPlatformSchema,
+    ) -> None:
+        # Mapping of Host Device Passthrough
+        self.AVAILABLE_HOST_DEVICES: Dict[
```
pupacha: `AVAILABLE_HOST_DEVICES` seems to be common across libvirt and hyperv platforms. The code the actually modifies this member can be moved to `BaseDevicePool` class.
For the part where different actions are performed can be overriden in the inherited classes.

lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -220,6 +239,13 @@ def _configure_environment(self, environment: Environment, log: Logger) -> None:
             self.runbook.admin_private_key_file
         )
 
+        # If Device_passthrough is set in runbook,
+        # Configure device passthrough params
+        self.device_pool_impl._configure_device_passthrough_pool(
```
pupacha: This needs to be called in platform initialization since its not specific to environment.

lisa/sut_orchestrator/hyperv/schema.py
```diff
@@ -62,9 +64,27 @@ class VhdSchema(schema.ImageSchema):
     vhd_path: Optional[str] = None
 
 
+@dataclass_json()
+@dataclass
+class DevicePassthroughSchema:
```
pupacha: Similar class is implemented for Libvirt platform. This can be made common and moved into utils.

pupacha (changes requested):

lisa/sut_orchestrator/hyperv/context.py
squirrelsc: For #2, you need to go through the collection (either dict or list), and find the right one by pool_type, when using the context. So, dict is easier to query than list. If you don't need to know the pool type, and most operations are to iterate all items, list is better.

lisa/sut_orchestrator/util/device_pool.py
squirrelsc: The key point is to have the two methods **paired** behaviors. Either just request/release resources from the central place, or apply/delete (including request/release) resources from the nodes.

lisa/tools/lspci.py
squirrelsc: Yes, do refactoring to make code life longer.

lisa/sut_orchestrator/libvirt/platform.py
cwize1: Even if something is only used once, sometimes moving it into a function can help with readability.

Also, in this case, there is an opposite function. So, moving into a function can help keep related code together.

lisa/sut_orchestrator/hyperv/context.py
smit-gardhariya: Removed. All conditions are now on list length

lisa/sut_orchestrator/hyperv/get_assignable_devices.ps1
smit-gardhariya: @squirrelsc This seems a big task. Can we take it as incremental changes?

lisa/tools/lspci.py
smit-gardhariya: lspci tool is implemented in a way where it will cache the list of all devices. I dont want to run multiple lspci combinations with different args. This way, we can enhance the PciDevice class properties as well.

lisa/tools/lspci.py
smit-gardhariya: @squirrelsc Please check the latest push

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: @squirrelsc Agree. But i created separate function which will be called from request/release part only. This way we can handle the assignment/removal related extra configuration properly

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: @squirrelsc Please check the latest push.

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: @squirrelsc Please check the latest push

lisa/sut_orchestrator/hyperv/context.py
smit-gardhariya: @squirrelsc We set only devices required by node under nodeContext. We dont have to query while reading nodeContext. Whatever the devices are present in the list, we will assign it to the node.

All the logic for pool_type based filtering of devices is implemented in the `request_devices`

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @squirrelsc Please check the latest push

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @squirrelsc Please check the latest push

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: @squirrelsc Please check the latest push

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1 Done. Please check the latest push.

lisa/tools/lspci.py
smit-gardhariya: @squirrelsc Used method provided by `lisa.utils` now. Please check the latest push

lisa/tools/lspci.py
smit-gardhariya: @squirrelsc Used method provided by `lisa.utils` now. Please check the latest push

lisa/sut_orchestrator/libvirt/platform.py

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1 Implemented `_parse_pci_address_str `. Please check

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @squirrelsc Removed the fn now

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1 Please check the latest push.

lisa/tools/hyperv.py
smit-gardhariya: There is only one place where we call create_vm that is under hyperv platform. We are calling start_vm explicitly as well.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: Signature/Structure of the pool is different for both the platform. That is why it make sense to put it under individual platform.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: This is pool configuration, and all the configuration related part is put to `_configure_environment`

lisa/sut_orchestrator/hyperv/schema.py
smit-gardhariya: member type is different.

lisa/tools/hyperv.py
pupacha: https://github.com/microsoft/lisa/pull/3379/commits/2f9cae832618b9a00233494d83d64dc3f99f49ea commit should have `start_vm` called immediately after every call of `create_vm`

lisa/sut_orchestrator/hyperv/get_assignable_devices.ps1
squirrelsc: Please finish it in this PR. It brings complex and hard to maintain parts in LISA. If you want to do integration test early, you can create a dev branch in microsoft/lisa to unblock work.

lisa/sut_orchestrator/util/device_pool.py
squirrelsc: at least the `_configure_device_passthrough_pool` should be public, and move upper than all private methods. A class needs at least one public method, or other class cannot call it at all.

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -0,0 +1,63 @@
...
+        raise NotImplementedError()
+
+    def _request_devices(
+        self,
+        pool_type: HostDevicePoolType,
+        count: int,
+    ) -> Any:
+        raise NotImplementedError()
+
+    def _release_devices(
```
squirrelsc: It's called out of BaseDevicePool, it should be public.

lisa/sut_orchestrator/libvirt/platform.py
pupacha: `_configure_environment` is called for every new environment. 
Device pool schema is associated with the platform itself. So, pool configuration has to be done at platform initialization.
There is no point in calling same pool configuration multiple times from `_configure_environment`

pupacha:
Please change the commits order. Put https://github.com/microsoft/lisa/pull/3379/commits/cd6363b01191bd98c70cb408377eb5b4a6a1afee before https://github.com/microsoft/lisa/pull/3379/commits/19f6308ae81c33f7a1f059278e9664e7da6aef67.

https://github.com/microsoft/lisa/pull/3379/commits/19f6308ae81c33f7a1f059278e9664e7da6aef67 imports classes that are defined in https://github.com/microsoft/lisa/pull/3379/commits/cd6363b01191bd98c70cb408377eb5b4a6a1afee

lisa/sut_orchestrator/hyperv/schema.py
pupacha: I see only additional member (`managed: str = ""`) in Libvirt;s `DevicePassthroughSchema`. Otherwise its same for `hyperv` platform.
You can either ignore the usage of additional member in hyperv platform or we can create something called `BaseDevicePassthroughSlchema` class in common utils and if needed Libvirt can inherit it.

lisa/sut_orchestrator/hyperv/get_assignable_devices.ps1
smit-gardhariya: @squirrelsc I removed powershell file and added python script now. Please check latest push.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @squirrelsc I converted powershell file to python class now. This part it removed. Resolving the comment.

lisa/sut_orchestrator/libvirt/platform.py
smit-gardhariya: @cwize1 Removed the shell piping

lisa/tools/hyperv.py
smit-gardhariya: Done

lisa/sut_orchestrator/hyperv/schema.py
smit-gardhariya: Done

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: @squirrelsc Done. Please check latest push

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/platform.py

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -334,3 +369,227 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+                self._assign_devices_to_vm(
+                    vm_name=vm_name,
+                    devices=devices,
+                )
+                device_context = DevicePassthroughContext()
+                device_context.pool_type = config.pool_type
+                device_context.device_list = devices
+                node_context.passthrough_devices.append(device_context)
+
+    def configure_device_passthrough_pool(
```
squirrelsc: move public methods before private methods.

lisa/sut_orchestrator/hyperv/get_assignable_devices.py
```diff
@@ -0,0 +1,302 @@
...
+            cmdlet="gwmi -query 'select * from Win32_PnPAllocatedResource'",
+            sudo=True,
+            force_run=True,
+        )
+        pnp_allocated_resources = stdout.strip().split("\r\n\r\n")
+        result: List[Dict[str, str]] = []
+        for rec in pnp_allocated_resources:
+            res: Dict[str, str] = {}
+            old_key: str = ""
+            for i in rec.strip().splitlines():
```
squirrelsc: Give some examples for output, so it's easier to compare if the output is expected. Use regexp is clearer to operate strings.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -336,3 +369,227 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
                 for node in environment.nodes.list()
             ]
         )
+
+
+class HyperVDevicePool(BaseDevicePool):
```
pupacha: Please move this class to a new file, just like the libvirt device pool.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -336,3 +369,227 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+        hv_dev = HypervAssignableDevices(
+            host_node=self._server,
+            log=self.log,
+        )
+        devices = hv_dev.get_assignable_devices(
+            vendor_id=vendor_id,
+            device_id=device_id,
+        )
+        primary_nic_id_list = self.get_primary_nic_id()
+        pool = self.available_host_devices.get(pool_type, [])
```
pupacha: This is not needed. You can direct do `pool = []` since `available_host_devices[pool_type]` is set  to empty list at the beginning of this method.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -336,3 +369,227 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
                 for node in environment.nodes.list()
             ]
         )
+
+
+class HyperVDevicePool(BaseDevicePool):
+    SUPPORTED_POOL_TYPE = [
```
pupacha: I see `SUPPORTED_POOL_TYPE` is passed all the way to `configure_device_passthrough_pool` of `BaseDevicePool`. So can you make `SUPPORTED_POOL_TYPE` a data member of `BaseDevicePool` itself and remove it as parameter from `configure_device_passthrough_pool` ?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -336,3 +369,227 @@ def _delete_node(node_ctx: NodeContext, wait_delete: bool) -> None:
...
+        return pnp_device_id_list
+
+    def _set_device_passthrough_node_context(
+        self,
+        node_context: NodeContext,
+        node_runbook: HypervNodeSchema,
+        hv: HyperV,
+        vm_name: str,
+    ) -> None:
+        if node_runbook.device_passthrough:
```
pupacha: nit: you can instead return with `if not` condition. this will make the rest of the code move away from 'if block'.

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
```diff
@@ -0,0 +1,358 @@
...
+        ] = {}
+        self.host_node = host_node
+        self.platform_runbook = runbook
+
+    def configure_device_passthrough_pool(
+        self,
+        device_configs: Optional[List[HostDevicePoolSchema]],
+        supported_pool_type: Optional[List[HostDevicePoolType]],
+    ) -> None:
+        if device_configs:
```
pupacha: nit: return with `if not` condition.

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
```diff
@@ -0,0 +1,358 @@
...
+            addr = f"{domain}:{addr}"
+        return addr
+
+    def _parse_pci_address_str(
+        self,
+        addr: str,
+        with_domain: bool = False,
+    ) -> Any:
+        addr_split = addr.strip().split(":")
+        cnt = 1 if with_domain else 0
```
pupacha: nit: `cnt` is confusing. use `index` or `idx`

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
```diff
@@ -0,0 +1,358 @@
...
+        device_addr_str = self._get_pci_address_str(device_addr)
+        kernel_module = lspci.get_used_module(device_addr_str)
+        return kernel_module == "vfio-pci"
+
+    def _set_device_passthrough_node_context(
+        self,
+        node_context: NodeContext,
+        node_runbook: BaseLibvirtNodeSchema,
+    ) -> None:
+        if node_runbook.device_passthrough:
```
pupacha: nit: return with `if not` condition

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
```diff
@@ -0,0 +1,358 @@
...
+                )
+                iommu_grp = iommu_grp_res.get("id", "")
+                break
+        assert iommu_grp, f"Can not get IOMMU group for device: {device}"
+        return f"iommu_grp_{iommu_grp}"
+
+    def get_primary_nic_id(self) -> List[str]:
+        # This is for baremetal. For azure, we have to get private IP
+        host = self.platform_runbook.hosts[0]
+        host_ip = host.address
```
pupacha: you can also get it by doing `self.host_node.connection_info.get("address")`

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
```diff
@@ -0,0 +1,358 @@
...
+        lspci = self.host_node.tools[Lspci]
+        device_list = lspci.get_devices_by_vendor_device_id(
+            vendor_id=vendor_id,
+            device_id=device_id,
+        )
+        primary_nic_iommu = self.get_primary_nic_id()
+        for item in device_list:
+            device = self.POOLTYPE_TO_DEVICE_SCHEMA[pool_type]()
+            bus, slot, fn = self._parse_pci_address_str(item.slot)
+            device.domain = item.domain
```
pupacha: Would all the mapped types from `POOLTYPE_TO_DEVICE_SCHEMA` have `domain` as member ?
I dont think `POOLTYPE_TO_DEVICE_SCHEMA` is needed for now. You can directly use `DeviceAddressSchema`

pupacha (changes requested):
Please add example runbook in the PR and commit descriptions.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: Supported pool type should be platform specific. Platform should decide pool type it supports. As of now we dont have a specific case for it but keeping it at platform level would make more sense as it would be isolated for both the platform

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: DeviceAddressSchema for both the platform would be different. It wont have domain say for hyperv's DeviceAddressSchema.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: We are calling `create_device_pool` for all vendor device id combination and for all pool type. If something is already there in pool for given pool type, this is the correct way to do it.

As of now we dont support more than one vendor/device id combination for given pool type but if we do so, we need this change

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: Done

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @squirrelsc Done with new file created for hyperv device pool

lisa/sut_orchestrator/hyperv/platform_.py
pupacha: Please check Line 405, you are assigning an empty list to `available_host_devices[pool_type]`

lisa/sut_orchestrator/hyperv/platform_.py
pupacha: You can have the member in BaseClass and override it in int init of platform specific classes.
This way you can do something like `self.supported_pool_type` and avoid parameter to `configure_device_passthrough_pool`

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
pupacha: I commented only about `POOLTYPE_TO_DEVICE_SCHEMA` of libvirt base class.

lisa/sut_orchestrator/hyperv/get_assignable_devices.py
smit-gardhariya: @squirrelsc Done. Used named regex now. Added stdout for 2 records.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: @pupacha Done

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: Yes removed that line now.

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: Done

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: This is used so you can add devices as well which has different kind of properties apart from bdf

This way each kind of pool will have its own device properties schema mapped.

lisa/sut_orchestrator/hyperv/platform_.py
smit-gardhariya: Done

smit-gardhariya:
> Please add example runbook in the PR and commit descriptions.

Added. Please check latest push

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
pupacha: Then we can't assume that `domain` will be a parameter for the obtained device type.

lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
```diff
@@ -0,0 +1,236 @@
...
+from lisa.sut_orchestrator.util.schema import HostDevicePoolSchema, HostDevicePoolType
+from lisa.tools import HyperV, PowerShell
+from lisa.util import ResourceAwaitableException
+from lisa.util.logger import Logger
+
+from .context import DevicePassthroughContext, NodeContext
+
+
+class HyperVDevicePool(BaseDevicePool):
+    POOLTYPE_TO_DEVICE_SCHEMA = {
```
pupacha: This is not used anywhere in the HyperV platform.

lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
```diff
@@ -0,0 +1,236 @@
...
+    def release_devices(
+        self,
+        node_context: NodeContext,
+    ) -> None:
+        self._remove_devices_from_vm(node_context=node_context)
+
+        # Refresh the pool
+        self.configure_device_passthrough_pool(
+            self._hyperv_runbook.device_pools,
+        )
```
pupacha: `configure_device_passthrough_pool` calls `create_device_pool` which gets the existing devices in the pool and I guess adds the devices again into the same list.
Either the devices need to be removed from `available_host_devices` before refreshing, or
`create_device_pool` at line 61 should start with an empty pool.

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: Here, PCI device will have domain. Domain will always be there for PCI device object returned by pci tool. We need this domain to construct the libvirt xml.

pupacha: I understand that PCI device will have domain field.
My question is what is the significance of `POOLTYPE_TO_DEVICE_SCHEMA` ?
IOW, if we are not doing any specific action based on the type it gives, why is it needed at all?
We can add it in later in future if there is a need.

pupacha: I understand that PCI device will have domain field.
My question is what is the significance of `POOLTYPE_TO_DEVICE_SCHEMA` ?
IOW, if we are not doing any specific action based on the type it gives, why is it needed at all?
We can add it in later future if there is a need.

lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
smit-gardhariya: Agree. Made the pool empty before refresh

lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
smit-gardhariya: Removed.

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
smit-gardhariya: Removed the mapping now

lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
pupacha: Why not make the pool empty at line 61 in `create_device_pool` itself?

Also, do we really need to call `configure_device_passthrough_pool` in `release_devices` ?
This is anyways called in `_prepare_environment`. It seems redundant.

squirrelsc:
@LiliDeng LGTM

lisa/sut_orchestrator/hyperv/hyperv_device_pool.py
smit-gardhariya: Anyway, we have to read the runbook to get vendor/device id lists to create pool. There is nothing much extra in `configure_device_passthrough_pool ` apart from checking supported pool type. I dont want to keep same piece of code at 2 different place.

pupacha (dismissed):
LGTM!

LiliDeng:
@smit-gardhariya I ran tier2 cases, saw regression issues on below cases.
```
verify_nvme_rescind
verify_nvme_sriov_rescind
verify_sriov_disable_enable_pci
verify_gpu_rescind_validation
verify_sriov_ethtool_offload_setting
```

smit-gardhariya:
> @smit-gardhariya I ran tier2 cases, saw regression issues on below cases.
> 
> ```
> verify_nvme_rescind
> verify_nvme_sriov_rescind
> verify_sriov_disable_enable_pci
> verify_gpu_rescind_validation
> verify_sriov_ethtool_offload_setting
> ```

Fixed it now @LiliDeng 
Issue was with slot value under lspci tool.

LiliDeng (approved):

PR #3381, opened 2024-08-15
Moved user password support to execute_async() for running test case

paulli2017:
Current run of test cases has two methods: 1. call node.execute.async() and then process.wait_result() 2. call node.execute.async and then pgrep.wait_process(). Both works ok if the test run doesn't need username and password while running on target, otherwise the second method will fail due to the missing username and password. The first method works because it has username and password process in process.wait_result() call. This change list is to move the username and password process in process.wait_result() to node.execute.async() so that both methods of running a test case will work. And this change will not cause behavior changes for the first method, which helps avoid the regression issue.

### DIFF
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -320,19 +320,26 @@ class Process:
             self._shell.close()
             raise
 
+        self.check_and_input_password()
+
     def check_and_input_password(self) -> None:
         if (
             self._sudo
             and isinstance(self._shell, SshShell)
             and self._shell.is_sudo_required_password
         ):
-            if not self._shell.connection_info.password:
-                raise RequireUserPasswordException(
-                    "Running commands with sudo requires user's password,"
-                    " but no password is provided."
-                )
-            self.input(f"{self._shell.connection_info.password}\n")
-            self._log.debug("The user's password is input")
+            timer = create_timer()
+            while self.is_running():
+                time.sleep(0.01)
+                if timer.elapsed(False) > 0.5:
+                    if not self._shell.connection_info.password:
+                        raise RequireUserPasswordException(
+                            "Running commands with sudo requires user's password,"
+                            " but no password is provided."
+                        )
+                    self.input(f"{self._shell.connection_info.password}\n")
+                    self._log.debug("The user's password is input")
+                    break
 
     def input(self, content: str) -> None:
         assert self._process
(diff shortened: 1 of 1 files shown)


squirrelsc:
Please remove test cases changes from this PR. They are in another PR.

lisa/node.py
```diff
@@ -548,8 +550,17 @@ def _execute(
...
             command_splitter=command_splitter,
         )
+        self._check_process_password(process)
         return process
 
+    def _check_process_password(self, process: Process) -> None:
+        timer = create_timer()
+        while process.is_running():
+            time.sleep(0.01)
+            if timer.elapsed(False) > 0.5:
```
squirrelsc: The check logic should be optimized to avoid wait 0.5 second for every command. It impacts the overall performance.
1. Check immediately one time. Please look at the logic inside of check_and_input_password, there are 3 conditions. If no condition met, it should skip following checks.
2. change the timer to 0.2, see if it still works.

paulli2017: that's a good suggestion. I kept this logic here as the change doesn't change the original behavior. As to optimization, I tried to call process.check_and_input_password() without waiting. It works from my test, but I am not sure it covers all the situations. I will update the change for your review. By the way, for suggestion 2, if I only change timer from 0.5 to 0.2, it works too.

squirrelsc: You can check with Baihua, how 0.5 second is from.

paulli2017: After discussion with Baihua, we took your first suggestion and made appropriate changes for check_and_input_password() for safety reason. 0.5 sec wait time can be changed to 0.2 sec from my test, but Baihua preferred to keep it to avoid regression. Please let me know if you have further suggestions.

lisa/node.py
```diff
@@ -548,6 +548,7 @@ def _execute(
             update_envs=update_envs,
             command_splitter=command_splitter,
         )
+        process.check_and_input_password()
```
squirrelsc: It can be called in process.start now.

paulli2017: yes, it can, but start function is responsible for just spawning a shell, I am not sure adding password processing there is appropriate. Does start() function need to care about password? By the way, the code change will be cleaner in process module only.

squirrelsc: > Does start() function need to care about password? 
It depends on how to define "start".

> By the way, the code change will be cleaner in process module only.
When the check password moved into "start", if there is another _execute called, it doesn't need to call start, check password together like this.

paulli2017: The change is done, thanks

squirrelsc (approved):

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3383, opened 2024-08-19
skip test cases which depends on blob public access

LiliDeng:


microsoft/testsuites/vm_extensions/runtime_extensions/common.py
```diff
@@ -96,6 +97,20 @@ def retrieve_storage_blob_url(
     )
     is_public_container = container_name.endswith("-public")
     blob_data = script or f"touch {test_file}"
+    if is_public_container:
```
squirrelsc: Move priority should be enough, it doesn't need to mark them as skipped.

LiliDeng: updated

squirrelsc (approved):

PR #3385, opened 2024-08-20
Update azure_disk_encryption.py

bamuir:
Remove maximum supported version check so that new versions are automatically ran, and we can see failures / know when we need to make changes for new distro versions.


LiliDeng:
@bamuir as discussed offline, we can reopen the PR once the known bugs have been fixed for the latest versions of images.

PR #3386, opened 2024-08-20
save_console_log: catch ChunkedEncodingError

LiliDeng:


lisa/sut_orchestrator/azure/common.py
```diff
@@ -2025,12 +2026,19 @@ def save_console_log(
...
+    try:
+        log_response = requests.get(
+            diagnostic_data.serial_console_log_blob_uri, timeout=60
         )
+        if log_response.status_code == 404:
+            log.debug(
+                "The serial console is not generated. "
+                "The reason may be the VM is not started."
+            )
+    except ChunkedEncodingError as ex:
```
squirrelsc, line 2038: Add another `except` to catch all other exceptions, to protect LISA process in future.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2025,12 +2026,19 @@ def save_console_log(
...
         screenshot_raw_name.unlink()
 
-    log_response = requests.get(diagnostic_data.serial_console_log_blob_uri, timeout=60)
-    if log_response.status_code == 404:
-        log.debug(
-            "The serial console is not generated. "
-            "The reason may be the VM is not started."
+    try:
+        log_response = requests.get(
+            diagnostic_data.serial_console_log_blob_uri, timeout=60
```
squirrelsc, line 2031: Does it need to extend the timeout?

LiliDeng: no need, I didn't see the timeout case happen.

squirrelsc (approved):

PR #3394, opened 2024-08-26
add test cases priority

LiliDeng:


squirrelsc (approved):
P2 is the default value, why does it need to add explicitly?

LiliDeng:
> P2 is the default value, why does it need to add explicitly?

I use regex to get the casename and priority with internal code, also keep them consistent with existing cases.

PR #3400, opened 2024-08-27
Support for detecting VMware ESXi system

LiliDeng:


lisa/executable.py
```diff
@@ -184,38 +188,48 @@ def _freebsd_tool(cls) -> Optional[Type[Tool]]:
...
-                no_info_log=True,
-                sudo=True,
-            )
-            if result.exit_code == 0:
-                self._log.debug(
-                    "executable exists in root paths, "
-                    "sudo always brings in following commands."
+            # Check command existence with 'command -v' and 'which'
+            where_commands = [f"command -v {command}", f"which {command}"]
+            for cmd in where_commands:
```
squirrelsc: You can check OS to decide which where command should be used, instead of calling it twice in some cases.

squirrelsc (approved):

PR #3401, opened 2024-08-29
tool: add mono to allow users to run .NET application across various platforms.

LiliDeng:


lisa/operating_system.py
```diff
@@ -487,6 +487,9 @@ def add_azure_core_repo(
     def set_kill_user_processes(self) -> None:
         raise NotImplementedError("set_kill_user_processes is not implemented")
 
+    def add_key(self, server_name: str, key: str) -> None:
```
squirrelsc: Is it a general method for all distro?

lisa/tools/mono.py
```diff
@@ -0,0 +1,98 @@
...
+        required_packages = [
+            "dirmngr",
+            "gnupg",
+            "apt-transport-https",
+            "ca-certificates",
+        ]
+        posix_os.install_packages(required_packages)
+
+        mono_repo_entry = (
+            "deb https://download.mono-project.com/repo/ubuntu stable-bionic main"
```
squirrelsc: Does it support only bionic? If so, filter out other versions.

lisa/tools/mono.py
```diff
@@ -0,0 +1,98 @@
...
+            value=mono_repo_entry,
+            file=self.node.get_pure_path(
+                "/etc/apt/sources.list.d/mono-official-stable.list"
+            ),
+            sudo=True,
+        )
+
+        posix_os.add_key(
+            server_name="hkp://keyserver.ubuntu.com:80",
+            key="3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF",
```
squirrelsc: Where is it from? Is it public?

lisa/operating_system.py
LiliDeng: I know for fedora, SUSE, they use `rpm --import key`,

lisa/tools/mono.py
LiliDeng: add comments, it is public

lisa/tools/mono.py
LiliDeng: change into distro codename

lisa/operating_system.py
squirrelsc: Let's keep it in Debian only, until we implement another os, and find the common pattern.

lisa/tools/mono.py
squirrelsc: Please also provide the source document.

squirrelsc (approved):

PR #3405, opened 2024-09-03
Replace initrd img and Add kernel config as optional parameter to transformer 'dom0_binaries'

smit-gardhariya:
Copy kernel config under /boot if passed to the transformer of 'dom0_binaries' type

Copy initrd if passed from runbook else create softlink for existing
initrd image with new kernel version in filename

### DIFF
--- a/lisa/transformers/dom0_kernel_installer.py
+++ b/lisa/transformers/dom0_kernel_installer.py
@@ -9,7 +9,7 @@ from dataclasses_json import dataclass_json
 
 from lisa import schema
 from lisa.node import Node
-from lisa.tools import Cp, Echo, Ls, Sed, Tar, Uname
+from lisa.tools import Cp, Echo, Ln, Ls, Sed, Tar, Uname
 from lisa.util import field_metadata
 
 from .kernel_installer import BaseInstaller, BaseInstallerSchema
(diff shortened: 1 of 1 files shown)


lisa/transformers/dom0_kernel_installer.py
```diff
@@ -35,6 +35,14 @@ class BinaryInstallerSchema(BaseInstallerSchema):
         ),
     )
 
+    # kernel config local absolute path
+    kernel_config_path: str = field(
+        default="",
+        metadata=field_metadata(
+            required=True,
```
pupacha, line 42: Does it mean this will be a required field ? If so, we should rather it optional

smit-gardhariya: @pupacha no, default would be empty string even if we dont pass it from runbook. 

If some value is set, then only we will try to copy the file to target system

smit-gardhariya: You can refer initrd parameter in this transformer, We are not passing it from runbook in our stack

pupacha (dismissed):

lisa/transformers/dom0_kernel_installer.py
pupacha: Ok. Thanks for confirmation.

squirrelsc (dismissed):

LiliDeng (approved):

PR #3407, opened 2024-09-05
Added FreeBSD specifics to Disk.py

r-dailey:
Added FreeBSD specifics for disk types to the disks feature to enable testing.

### DIFF
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import re
 from functools import partial
 from typing import Any, Dict, List, Optional, Type
 
@@ -8,9 +9,10 @@ from assertpy.assertpy import assert_that
 
 from lisa import schema
 from lisa.feature import Feature
+from lisa.operating_system import BSD
 from lisa.tools import Mount
 from lisa.tools.mount import PartitionInfo
-from lisa.util import LisaException
+from lisa.util import LisaException, get_matched_str
 
 
 class Disk(Feature):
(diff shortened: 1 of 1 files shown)


lisa/features/disks.py
```diff
@@ -79,20 +79,35 @@ def get_os_boot_partition(self) -> Optional[PartitionInfo]:
         for partition in partition_info:
             if partition.mount_point.startswith("/boot"):
                 boot_partition = partition
+                if "FreeBSD" in self._node.os.name:
```
squirrelsc: use `isinstance` to check if it's BSD. Only the os or some basic classes use the string to avoid cycle reference.

lisa/features/disks.py
```diff
@@ -79,20 +79,35 @@ def get_os_boot_partition(self) -> Optional[PartitionInfo]:
...
+                    cmd = "glabel status"
+                    output = self._node.execute(cmd).stdout.split("\n")
+                    for line in output:
+                        if boot_partition.disk in line:
+                            dev = line.split()[2]
```
squirrelsc: Use regexp and add examples above.

lisa/features/disks.py
r-dailey: Ok I will change it to use isinstance.

lisa/features/disks.py
r-dailey: Done

lisa/features/disks.py
```diff
@@ -79,20 +81,41 @@ def get_os_boot_partition(self) -> Optional[PartitionInfo]:
...
     # Get disk controller type from the VM by checking the boot partition
     def get_os_disk_controller_type(self) -> schema.DiskControllerType:
         boot_partition = self.get_os_boot_partition()
         assert boot_partition, "'boot_partition' must not be 'None'"
-
-        if boot_partition.disk.startswith("nvme"):
-            os_disk_controller_type = schema.DiskControllerType.NVME
-        elif boot_partition.disk.startswith("sd"):
-            os_disk_controller_type = schema.DiskControllerType.SCSI
+        if "FreeBSD" in self._node.os.name:
```
squirrelsc: Use `isinstance` as above.

r-dailey: Ah I missed the second one. Thanks for pointing it out.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3408, opened 2024-09-05
set default value to avoid missing credential exception

LiliDeng:
Reverts microsoft/lisa#3395

it breaks current baremetal run
```
    raise LisaException(
lisa.util.LisaException: at least one of password or private_key_file need to be set when connecting
```

### DIFF
--- a/lisa/sut_orchestrator/baremetal/context.py
+++ b/lisa/sut_orchestrator/baremetal/context.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass, field
+from functools import partial
 
 from lisa import schema
 from lisa.environment import Environment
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/baremetal/context.py
```diff
@@ -14,7 +14,7 @@ class EnvironmentContext:
 
 @dataclass
 class NodeContext:
-    connection: schema.ConnectionInfo = field(default_factory=schema.ConnectionInfo)
+    connection: schema.ConnectionInfo = schema.ConnectionInfo(password="mock")
```
squirrelsc: You can use `partial` to give a default value.

squirrelsc (approved):

PR #3410, opened 2024-09-05
Fix libvirt for mariner if already present

smit-gardhariya:
If libvirt not present, install it and fix if distro is mariner
If libvirt is present, fix it if distro is mariner

We are implementing scenario-2 here. Scenario-1 was already there. To do so, Push fn `_fix_mariner_installation` at common place and call the fn for both the scenarios.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/transformers.py
+++ b/lisa/sut_orchestrator/libvirt/transformers.py
@@ -213,22 +213,8 @@ class LibvirtPackageInstaller(LibvirtInstaller):
         packages_list = self._distro_package_mapping[type(linux).__name__]
         self._log.info(f"installing packages: {packages_list}")
         linux.install_packages(packages_list)
-        self._fix_mariner_installation()
+        _fix_mariner_installation(node=self._node)
         return self._get_version()
-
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/libvirt/transformers.py
```diff
@@ -584,3 +570,20 @@ def _install_libvirt(runbook: schema.TypedSchema, node: Node, log: Logger) -> No
     else:
         libvirt_version = libvirt_installer._get_version()
         log.info(f"Already installed! libvirt version: {libvirt_version}")
+        _fix_mariner_installation(node=node)
```
pupacha, line 573: This is needed only for upstream libvirt right ? So, we can call `_fix_mariner_installation` conditionally based on the version.

smit-gardhariya: @pupacha This is irrespective of libvirt version for us. This seems working fine with our mshv mariner2.0 as well which has inbuilt libvirt of older version.

pupacha (approved):

LiliDeng (approved):

PR #3411, opened 2024-09-06
Added testcase for yum update

SRIKKANTH:
There are issues reported during "yum update". 

This test case is for capturing them.

### DIFF
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -520,6 +520,24 @@ class AzureImageStandard(TestSuite):
 
     @TestCaseMetadata(
         description="""
+        Verify if there is any issues in and after 'os update'
+
+        Steps:
+        1. Run os update command.
+        2. Reboot the VM and see if the VM is still in good state.
+        """,
+        priority=2,
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+    )
+    def verify_os_update(self, node: Node) -> None:
+        if isinstance(node.os, Posix):
+            node.os.update_packages("")
+        else:
+            raise SkippedException(f"Unsupported OS or distro type : {type(node.os)}")
+        node.reboot()
+
+    @TestCaseMetadata(
+        description="""
         This test will check that kvp daemon is installed. This is an optional
         requirement for Debian based distros.
 


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -518,6 +518,28 @@ def verify_yum_conf(self, node: Node) -> None:
...
+        1. Run 'yum update -y' command.
+        2. Get the output of dmesg to see if the VM is still in good state.
+        """,
+        priority=2,
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+    )
+    def verify_yum_update(self, node: Node) -> None:
+        dmesg = node.tools[Dmesg]
+        if isinstance(node.os, Fedora):
+            node.execute(
```
squirrelsc: Leverage the YumConfigManager tool, and assert exit code.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -518,6 +518,28 @@ def verify_yum_conf(self, node: Node) -> None:
...
+                "yum update -y",
+                sudo=True,
+            )
+            dmesg.get_output(force_run=True)
```
squirrelsc: Get output doesn't fail the test case, if abnormal things happen. Use `dmesg.check_kernel_errors` or find another way to check the results.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -518,6 +518,28 @@ def verify_yum_conf(self, node: Node) -> None:
...
+        else:
+            raise SkippedException(f"Unsupported distro type : {type(node.os)}")
+
```
LiliDeng, line 538: We need reboot to check whether it can be bootable with new kernel

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Used reboot as a check for VM's health removed dmesg.

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Used node reboot for VM's health check.

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: This particular test is to cover a VM hang issue with "yum update" on RHEL VMs. 
I'm afraid it cannot be reproduced with "yum-config-manager"

squirrelsc:
@LiliDeng LGTM

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -518,6 +518,28 @@ def verify_yum_conf(self, node: Node) -> None:
...
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+    )
+    def verify_yum_update(self, node: Node) -> None:
+        if isinstance(node.os, Fedora):
+            node.execute(
+                "yum update -y",
+                sudo=True,
+                timeout=1800,
+            )
+            node.reboot()
```
LiliDeng: no need restricts the os, use `node.os.update_packages("")` to invoke the system upgrade

SRIKKANTH: Changed the logic to use "node.os.update_packages("")"

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -518,6 +518,24 @@ def verify_yum_conf(self, node: Node) -> None:
...
+
+        Steps:
+        1. Run 'yum update -y' command.
+        2. Reboot the VM and see if the VM is still in good state.
+        """,
+        priority=2,
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
+    )
+    def verify_yum_update(self, node: Node) -> None:
+        if isinstance(node.os, Fedora):
```
LiliDeng: don't restrict the os, mayby just check whether it is Posix

SRIKKANTH: Replaced "yum update -y" with node.os.update_packages("")

LiliDeng: it also can be run on other distro like ubuntu

LiliDeng: @SRIKKANTH please fix CI check error

SRIKKANTH: @LiliDeng error is fixed now.

LiliDeng (approved):

PR #3414, opened 2024-09-10
fix vulnerabilities reported in S360

lesscodingmorehappiness:


pyproject.toml
```diff
@@ -27,6 +27,7 @@ dependencies = [
     "spurplus ~= 2.3.5",
     "websockets ~= 10.3",
     "charset_normalizer ~= 2.1.1",
+    "zipp ~= 3.19.1",
```
squirrelsc: Where is this from?

lesscodingmorehappiness: it's reported in [S360](https://lens.msftcloudes.com/#/dashboard/2a8a3c11-a136-4310-95c3-c812f53657c9?_g=(ws:e36e780f-b4f7-49f6-b1c8-a426de0d20bc)&params=(filters:!((k:Digest,v:'sha256:5d5c18f0a76c6316116fe50487652d75a48877c80902f17295b3496676f97349'),(k:SubscriptionId,v:'46cad376-9887-4ccc-b869-7b47bfd97b54'),(k:DueDateMonth,v:'2024-08-01T00:00:00.0000000Z')),time:(from:'2024-09-09T19:22:50.8310000Z',to:'2024-09-10T16:02:55.6710000Z')))

squirrelsc: Sorry for unclear ask. I mean where the dependency is from? This section is for common code of LISA. I want to maintain a minimun set of dependencies, because we want to use LISA in some restricted environments. If the zipp dependency is from Azure, the package should be added in the Azure section.

lesscodingmorehappiness: It seems to be some misalert, I don't see this pkg now...I'll send update based on latest scan result

lesscodingmorehappiness: image

microsoft/testsuites/vm_extensions/runtime_extensions/common.py
```diff
@@ -5,7 +5,7 @@
 
 from assertpy import assert_that
 from azure.core.exceptions import ResourceExistsError
-from azure.storage.blob import BlobType
+from azure.storage.blob._models import BlobType
```
squirrelsc: It may not be allowed to reference `_models`. Please try if it's fixed by upgrading storage packages.

squirrelsc:
Please fix check errors and merge them into one meaningful commit.

squirrelsc (approved):

PR #3415, opened 2024-09-11
Add device-passthrough testcase for CH platform

smit-gardhariya:
Add test case to verify if passthrough-devices
are visible to guest for CH platform

### DIFF
--- a/microsoft/testsuites/device_passthrough/functional_tests.py
+++ b/microsoft/testsuites/device_passthrough/functional_tests.py
@@ -1,0 +1,110 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+from pathlib import Path
+from typing import Any, Dict, cast
+
+from lisa import (
+    Environment,
+    Logger,
+    Node,
+    TestCaseMetadata,
+    TestSuite,
+    TestSuiteMetadata,
+)
+from lisa.sut_orchestrator import CLOUD_HYPERVISOR
+from lisa.sut_orchestrator.libvirt.ch_platform import CloudHypervisorPlatform
+from lisa.sut_orchestrator.libvirt.schema import BaseLibvirtNodeSchema
+from lisa.testsuite import TestResult, simple_requirement
+from lisa.tools import Lspci
+from lisa.util import LisaException, SkippedException
+
+
+@TestSuiteMetadata(
+    area="device_passthrough",
+    category="functional",
+    description="""
+    This test suite is for testing device passthrough functional tests.
+    """,
+)
+class DevicePassthroughFunctionalTests(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            Check if passthrough device is visible to guest.
+            This testcase support only on CLOUD_HYPERVISOR
+            platform of LISA. Please refer below runbook snippet.
+
+            platform:
+              - type: cloud-hypervisor
+                cloud-hypervisor:
+                  device_pools:
+                    - type: "pci_net"
+                      devices:
+                        - vendor_id: xxx
+                          device_id: xxx
+                requirement:
+                  cloud-hypervisor:
+                    device_passthrough:
+                      - pool_type: "pci_net"
+                        managed: "yes"
+                        count: 1
+
+            We will check if sufficient devices are visible to guest or not.
+            Platform will create device pool based on given device/vendor id.
+            'device_passthrough' section will tell platform to create node
+            with appropriate num of devices being passthrough. Based on pool_type
+            value, platform will try to get devices from pool and assign it to node.
+
+            Testcase will verify if needed devices are present on node by reading
+            extended runbook. It will get the vendor/device id based on 'pool_type'
+            and check how many devices are present on node for that vendor/device id.
+        """,
+        priority=4,
+        requirement=simple_requirement(
+            supported_platform_type=[CLOUD_HYPERVISOR],
+        ),
+    )
+    def verify_device_passthrough_on_guest(
+        self,
+        log: Logger,
+        node: Node,
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        lspci = node.tools[Lspci]
+        platform = cast(CloudHypervisorPlatform, environment.platform)
+        pool_vendor_device_map = {}
+        assert platform.platform_runbook.device_pools, "Device pool cant be empty"
+        for pool in platform.platform_runbook.device_pools:
+            pool_type = str(pool.type.value)
+            vendor_device_id = {
+                "vendor_id": pool.devices[0].vendor_id,
+                "device_id": pool.devices[0].device_id,
+            }
+            pool_vendor_device_map[pool_type] = vendor_device_id
+
+        assert environment.runbook.nodes_requirement, "requirement cant be empty"
+        for node_space in environment.runbook.nodes_requirement:
+            node_runbook: BaseLibvirtNodeSchema = node_space.get_extended_runbook(
+                BaseLibvirtNodeSchema, CLOUD_HYPERVISOR
+            )
+            if not node_runbook.device_passthrough:
+                raise SkippedException("No device-passthrough is set for node")
+            for req in node_runbook.device_passthrough:
+                pool_type = str(req.pool_type.value)
+                ven_dev_id_of_pool = pool_vendor_device_map[pool_type]
+                ven_id = ven_dev_id_of_pool["vendor_id"]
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/device_passthrough/functional_tests.py
```diff
@@ -0,0 +1,89 @@
...
+class DevicePassthroughFunctionalTests(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            Check if passthrough device is visible to guest.
+            This testcase support only on CLOUD_HYPERVISOR
+            platform of LISA.
+        """,
+        priority=4,
+        requirement=simple_requirement(
+            supported_platform_type=[CLOUD_HYPERVISOR],
```
squirrelsc, line 63: Does it support Hyper-V?

smit-gardhariya: Yes @squirrelsc , Idea is to extend the same test for hyper-v too. As of now, this test is solely for CH platform. 

We need some changes to make the test work on hyper-v platform and I cannot test the change because of some infra issue on hyper-v. So, that is scoped for future one.

microsoft/testsuites/device_passthrough/functional_tests.py
```diff
@@ -0,0 +1,89 @@
...
+    def verify_device_passthrough_on_guest(
+        self,
+        log: Logger,
+        node: Node,
+        environment: Environment,
+        log_path: Path,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        ctx = get_ch_context(node)
```
squirrelsc: The context is not used after assertion.

microsoft/testsuites/device_passthrough/functional_tests.py
```diff
@@ -0,0 +1,89 @@
...
+        assert platform.platform_runbook.device_pools, "Device pool cant be empty"
+        for pool in platform.platform_runbook.device_pools:
+            pool_type = str(pool.type.value)
+            vendor_device_id = {
+                "vendor_id": pool.devices[0].vendor_id,
+                "device_id": pool.devices[0].device_id,
+            }
+            pool_vendor_device_map[pool_type] = vendor_device_id
+
+        assert environment.runbook.nodes_requirement, "requirement cant be empty"
```
squirrelsc, line 87: I'm not understand the logic, why does it need to check platform runbook and node runbook both? They should be expected values, and one of them should be enough to provide the expected value.

microsoft/testsuites/device_passthrough/functional_tests.py
smit-gardhariya: Yes, context is to check if any device-passthrough is set or not. I can check that with extended runbook too. Will remove context now

microsoft/testsuites/device_passthrough/functional_tests.py
smit-gardhariya: Device pool is part of platform runbook whereas what kind of device need to be assigned from pool is part of node runbook.

Platform runbook defines the pool configuration
Node runbook defines how many devices/what kind of devices need to be passthrough under requirement section

smit-gardhariya: Here is the reference @squirrelsc 

```
platform:
  -type: cloud-hypervisor
      cloud-hypervisor:
      device_pools:
        - type: "pci_net"
	  devices:
	    - vendor_id: xxx
	      device_id: xxx
    requirement:
      cloud-hypervisor:
        device_passthrough:
	  -pool_type: "pci_net"
	    managed: "yes"
	    count: 1
```

squirrelsc: Use node.runbook, instead of nodes_requirement. The nodes_requirements may be expanded to multiple nodes.

microsoft/testsuites/device_passthrough/functional_tests.py
squirrelsc: It may hide some bugs. The case should check if the platform defined devices or not, instead of the actual devices exists or not.

microsoft/testsuites/device_passthrough/functional_tests.py
```diff
@@ -0,0 +1,89 @@
...
+    area="device_passthrough",
+    category="functional",
+    description="""
+    This test suite is for testing device passthrough functional tests.
+    """,
+)
+class DevicePassthroughFunctionalTests(TestSuite):
+    @TestCaseMetadata(
+        description="""
+            Check if passthrough device is visible to guest.
```
squirrelsc, line 32: Make it clearer on what's value compared. So it's easy to setup runbook to run this case.

microsoft/testsuites/device_passthrough/functional_tests.py
smit-gardhariya: @squirrelsc 

Flow is like, based on runbook, platform will set the node context under `passthrough_devices` and if that is set, we want to verify that what are the devices assigned to libvirt domain.

Reason on verifying actual devices is present in the VM is

- Make sure devices are visible to guest. If there is a case where libvirt domain may come up but device is not correctly passthrough
- Run functional aspect based on device type (NIC, GPU etc), This is considered for upcoming work. We can extend the same testcase to verify this. Current scope is limited for the testcase, but we will extend this in upcoming work cycle.
- bdf for pci device under Guest VM would be different. So only way to verify is from runbook. Check the node requirement to get the device count and device type (pool type) and get what is the vendor/device id for given device type so We can check on VM that whatever runbook has defined as devices are visible to guest or not.

I am still a bit unclear when you said it may hide some bugs. Could you please explain.

microsoft/testsuites/device_passthrough/functional_tests.py
smit-gardhariya: We are using extended runbook only. Let me check if is there other better way to get the runbook.

microsoft/testsuites/device_passthrough/functional_tests.py
squirrelsc: For most test cases, they verify if the "actual" value is the same as the "expected" value. In this case, node context has the actual value, and runbook has the expected value.
It should be skipped only the "expected" value doesn't meet the test criteria. It shouldn't skip the case, if the "actual" value by the "actual" value. Because if the expected value expects 1, but actual value is 0. It should be failed, instead of skipped.

smit-gardhariya: @squirrelsc 

Sure CHi, I will remove node context check. We can do the same check on extended runbook. 

If "Expected" device from runbook is not present, we are failing the test anyway.

microsoft/testsuites/device_passthrough/functional_tests.py
smit-gardhariya: Sure Chi

smit-gardhariya: Done

microsoft/testsuites/device_passthrough/functional_tests.py
smit-gardhariya: Removed the node context. We will read extended runbook and decide if device passthrough is set or not. If not, skip the test.

microsoft/testsuites/device_passthrough/functional_tests.py
smit-gardhariya: Using extended runbook only to get the expected value.

squirrelsc: Please check Azure platform code. the runbook should be merged, and copied to each node. The nodes_requirement is the raw info to create nodes, which shouldn't be used for each environment.

https://github.com/microsoft/lisa/blob/d50a39f7734f8eb16ecbe3fa37235e00b5e7c407/lisa/sut_orchestrator/azure/platform_.py#L1123

microsoft/testsuites/device_passthrough/functional_tests.py
```diff
@@ -0,0 +1,110 @@
...
+            }
+            pool_vendor_device_map[pool_type] = vendor_device_id
+
+        assert environment.runbook.nodes_requirement, "requirement cant be empty"
+        for node_space in environment.runbook.nodes_requirement:
+            node_runbook: BaseLibvirtNodeSchema = node_space.get_extended_runbook(
+                BaseLibvirtNodeSchema, CLOUD_HYPERVISOR
+            )
+            if not node_runbook.device_passthrough:
+                raise SkippedException("No device-passthrough is set for node")
```
squirrelsc, line 93: This check should happen earlier before any other command.

PR #3416, opened 2024-09-11
Add TestCategory and MachineName to JUnitReport

snipex111:
Add TestCategory and MachineName to JUnitReport

Update ExitCode=0 when 0 test case detected, this is done as "if a pipeline fails/cancelled just after all test cases ran and passed but before the summary generation, in rerun our pipeline optimisation script will return "0 test case" and this should be a valid case and only report should be generated and rerun and pipeline should pass.

### DIFF
--- a/TestControllers/TestController.psm1
+++ b/TestControllers/TestController.psm1
@@ -53,6 +53,7 @@ Class TestController {
 	[string] $RGIdentifier
 	[string] $OsVHD
 	[string] $TestCategory
+	[string] $MachineName
 	[string] $TestNames
 	[string] $TestArea
 	[string] $TestTag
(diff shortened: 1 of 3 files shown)


LiliDeng (dismissed):

snipex111:

TestControllers/TestController.psm1
```diff
@@ -323,47 +330,48 @@ Class TestController {
 			}
 		}
 		if (!$allTests) {
-			Throw "Not able to collect any test cases from XML files"
+			Write-LogWarn "Not able to collect any test cases from XML files"
```
LiliDeng, line 333: why change this?

snipex111: basically , if a pipeline run get cancelled just after all test cases passed, but before the report generation, so in this case if we rerun. "i changed the script in pipeline to run only failed test case."
 so in rerun 0 testcases will be there, but task can be sucessfully executed as we only have report generation left in this rerun.

LiliDeng: can you add return here? and keep following contents not changed?

snipex111: done

LiliDeng (approved):

PR #3417, opened 2024-09-12
add retry_on_deployment

LiliDeng:


lisa/node.py
```diff
@@ -1116,6 +1116,10 @@ def cleanup(self) -> None:
         for node in self._list:
             node.cleanup()
 
+    def clear(self) -> None:
```
squirrelsc: It doesn't need this method. If set env.status to new, the nodes will be reset too. Check the status methods of env.

lisa/runners/lisa_runner.py
```diff
@@ -290,30 +290,46 @@ def _prepare_environments(self) -> None:
...
+                    if self._is_awaitable_timeout("deploy"):
+                        self._log.info(
+                            f"[{environment.name}] timeout on waiting for more "
+                            f"resource: {identifier}, skip assigning case."
+                        )
+                        raise SkippedException(identifier)
+                    else:
+                        # rerun prepare to calculate resource again.
+                        environment.status = EnvironmentStatus.New
+            except Exception as identifier:
```
squirrelsc, line 314: Use the same level as ResourceAwaitableException. The `ResourceAwaitableException` shouldn't be counted in retry.

lisa/runners/lisa_runner.py
```diff
@@ -290,30 +290,46 @@ def _prepare_environments(self) -> None:
...
-                result=test_results[0],
-                exception=identifier,
-            )
-            self._delete_environment_task(environment=environment, test_results=[])
+                    # Log retry information and continue to the next attempt
+                    self._log.info(
+                        f"[{environment.name}] Deployment failed, retrying... "
+                        f"(Attempt {attempt + 1}/{max_retries - 1})"
+                    )
+                    environment.status = EnvironmentStatus.Prepared
```
squirrelsc: `environment.status = EnvironmentStatus.New` is also applicable here.

lisa/runners/lisa_runner.py
```diff
@@ -290,30 +290,46 @@ def _prepare_environments(self) -> None:
...
+                        self._log.info(
+                            f"[{environment.name}] timeout on waiting for more "
+                            f"resource: {identifier}, skip assigning case."
+                        )
+                        raise SkippedException(identifier)
+                    else:
+                        # rerun prepare to calculate resource again.
+                        environment.status = EnvironmentStatus.New
+            except Exception as identifier:
+                if attempt >= max_retries - 1:
```
squirrelsc: `if attempt < max_retries:` then retry. Otherwise, raise attach to test result.

lisa/runners/lisa_runner.py
```diff
@@ -290,30 +290,46 @@ def _prepare_environments(self) -> None:
     def _deploy_environment_task(
         self, environment: Environment, test_results: List[TestResult]
     ) -> None:
-        try:
+        max_retries = test_results[0].runtime_data.retry_on_deployment + 1
```
squirrelsc: The flag should be on environment, not test case. Every time, the retry count may applied to different test cases. So it cannot be set correctly.

lisa/runners/lisa_runner.py
LiliDeng: with set it as new, I saw below exception, during retry deploy

```
2024-09-12 23:52:13.320[15700][ERROR] lisa.runner[0] case failed
  File "C:\code\lisa-lsg\lsg-lisa\lisa\lisa\runners\lisa_runner.py", line 298, in _deploy_environment_task
    self.platform.deploy_environment(environment)
  File "C:\code\lisa-lsg\lsg-lisa\lisa\lisa\platform_.py", line 184, in deploy_environment
    self._deploy_environment(environment, log)
  File "C:\code\lisa-lsg\lsg-lisa\lisa\lisa\sut_orchestrator\azure\platform_.py", line 607, in _deploy_environment
    raise identifier
  File "C:\code\lisa-lsg\lsg-lisa\lisa\lisa\sut_orchestrator\azure\platform_.py", line 578, in _deploy_environment
    location, deployment_parameters = self._create_deployment_parameters(
  File "C:\code\lisa-lsg\lsg-lisa\lisa\lisa\sut_orchestrator\azure\platform_.py", line 1118, in _create_deployment_parameters
    azure_node_runbook = self._create_node_runbook(
  File "C:\code\lisa-lsg\lsg-lisa\lisa\lisa\sut_orchestrator\azure\platform_.py", line 1276, in _create_node_runbook
    raise LisaException("vm_size is not detected before deploy")
lisa.util.LisaException: vm_size is not detected before deploy
```

squirrelsc: Can you check why it works for awaitable but not here?

LiliDeng: if encounter awaitable, I saw the prepare_environment is invoked again, so below two lines work when encounter deployment failure.
```
                    environment.status = EnvironmentStatus.New
                    self._prepare_environment(environment)
```

squirrelsc: It doesn't make sense, the logic in the same function. Why only Awaitable exception call prepare?

LiliDeng: for awaitable exception, it keep running, since is_done is always false, for deployment failure, is_done is true after running once.

image

LiliDeng: for Awaitable exception, is_done is always false, and it will keep running fetch_task, for retry logic, it only happens inside of _deploy_environment_task

lisa/node.py
LiliDeng: removed

lisa/runners/lisa_runner.py

lisa/schema.py
```diff
@@ -1320,6 +1320,12 @@ def reload_requirements(self) -> None:
 class EnvironmentRoot:
     warn_as_error: bool = field(default=False)
     environments: List[Environment] = field(default_factory=list)
+    retry_on_deployment: int = field(
```
squirrelsc: `retry` should be enough, and add comments above.

selftests/azure/test_prepare.py
```diff
@@ -377,7 +377,11 @@ def load_environment(
                 _ = node_req.get_extended_runbook(common.AzureNodeSchema, AZURE)
                 runbook._original_nodes_requirement.append(node_req)
         environment = Environment(
-            is_predefined=True, warn_as_error=False, id_=0, runbook=runbook
+            is_predefined=True,
+            warn_as_error=False,
+            id_=0,
+            runbook=runbook,
+            retry_on_deployment=0,
```
squirrelsc: Think about to add a test case for retry. It should be in runner tests.

lisa/node.py
squirrelsc: Is latest change pushed?

lisa/runners/lisa_runner.py
```diff
@@ -290,30 +290,46 @@ def _prepare_environments(self) -> None:
...
-            except ResourceAwaitableException as identifier:
-                if self._is_awaitable_timeout("deploy"):
+                try:
+                    # Attempt to deploy the environment
+                    self.platform.deploy_environment(environment)
+                    assert (
+                        environment.status == EnvironmentStatus.Deployed
+                    ), f"actual: {environment.status}"
+                    self._reset_awaitable_timer("deploy")
+                    break
```
squirrelsc: add comments above to explain break.

lisa/node.py
LiliDeng: missed this, updated

lisa/schema.py

lisa/runners/lisa_runner.py
LiliDeng: added

lisa/runners/lisa_runner.py

lisa/runners/lisa_runner.py
```diff
@@ -308,12 +309,16 @@ def _deploy_environment_task(
                     # rerun prepare to calculate resource again.
                     environment.status = EnvironmentStatus.New
         except Exception as identifier:
-            self._attach_failed_environment_to_result(
-                environment=environment,
-                result=test_results[0],
-                exception=identifier,
-            )
-            self._delete_environment_task(environment=environment, test_results=[])
+            if self._check_retry_limit():
```
squirrelsc: The name can be more clear like `_need_retry`. So it hints the output is true/false to retry or not.

lisa/runner.py
```diff
@@ -190,6 +193,21 @@ def _reset_awaitable_timer(self, name: str) -> None:
         self._wait_resource_logged = False
         self._wait_resource_timers[name] = _wait_resource_timer
 
+    def _check_retry_limit(self) -> bool:
+        if self._max_retries > 0:
```
squirrelsc: The retry counter should be on the environment, not a global one. It's different with timeout.

LiliDeng: do you mean add current_attempt into environment?

squirrelsc: Yes, it can be called like `tried_times`.

lisa/runners/lisa_runner.py

lisa/runner.py
LiliDeng: updated

lisa/runner.py
```diff
@@ -190,6 +191,22 @@ def _reset_awaitable_timer(self, name: str) -> None:
         self._wait_resource_logged = False
         self._wait_resource_timers[name] = _wait_resource_timer
 
+    def _need_retry(self, environment: Environment) -> bool:
```
squirrelsc, line 194: The logic can be simpler and clearer like below.

```python
if environment.tried_times >= environment.retry:
    if environment.retry > 0:
        self._log.info(
        f"Tried {environment.tried_times + 1} times, but failed again."
            )
    return False

self._log.info(
    f"Retrying... (Attempt {environment.tried_times}/{max_retries})"
)
environment.tried_times += 1
return True
```

LiliDeng: updated

squirrelsc (approved):

PR #3418, opened 2024-09-13
Updating test case logic to check if netsvc_module exists

umfranci:
For the test case 'verify_device_msg_level_change', the path check of the netsvc_module file would usually throw a 'Permission Error' since the command was not using sudo. Hence updating the code to use the 'Ls' tool with sudo to check if the file is present.

Tested the change with both scenarios - one where the path exists and one where it does not.

### DIFF
--- a/microsoft/testsuites/network/networksettings.py
+++ b/microsoft/testsuites/network/networksettings.py
@@ -3,7 +3,6 @@
 
 import re
 import time
-from pathlib import PurePosixPath
 from typing import Any, Dict, List, Tuple, Union, cast
 
 from assertpy import assert_that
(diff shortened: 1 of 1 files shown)


LiliDeng (dismissed):

microsoft/testsuites/network/networksettings.py
```diff
@@ -731,10 +730,10 @@ def _check_msg_level_change_supported(self, node: Node) -> None:
                     cwd=node.working_path,
                 ).stdout
 
-            assert node.shell.exists(
-                PurePosixPath(netvsc_module)
+            ls = node.tools[Ls]
+            assert ls.path_exists(
+                f"{netvsc_module}", sudo=True
```
LiliDeng: f"{netvsc_module}" => `netvsc_module`

LiliDeng (approved):

PR #3419, opened 2024-09-13
io_uring support

SRIKKANTH:
Added 'io_uring' support to LISA io perf tests and a new testcase 'perf_nvme_io_uring'.
io_uring is advanced alternative for libaio with low system overhead and more io depth.

### DIFF
--- a/lisa/tools/fio.py
+++ b/lisa/tools/fio.py
@@ -37,6 +37,15 @@ FIOMODES = Enum(
 )
 
 
+class IoEngine(Enum):
+    IO_URING = "io_uring"
+    LIBAIO = "libaio"
+    POSIXAIO = "posixaio"
+
+    def __str__(self) -> str:
+        return self.value
+
+
 class Fio(Tool):
     fio_repo = "https://github.com/axboe/fio/"
     branch = "fio-3.29"
@@ -93,6 +102,7 @@ class Fio(Tool):
         verify_dump: bool = False,
         verify_fatal: bool = False,
         verify: str = "",
+        ioengine: IoEngine = IoEngine.LIBAIO,
         cwd: Optional[pathlib.PurePath] = None,
     ) -> FIOResult:
         cmd = self._get_command(
@@ -114,6 +124,7 @@ class Fio(Tool):
             verify_dump,
             verify_fatal,
             verify,
+            ioengine,
         )
         result = self.run(
             cmd,
@@ -149,6 +160,7 @@ class Fio(Tool):
         verify_dump: bool = False,
         verify_fatal: bool = False,
         verify: str = "",
+        ioengine: IoEngine = IoEngine.LIBAIO,
         cwd: Optional[pathlib.PurePath] = None,
     ) -> Process:
         cmd = self._get_command(
@@ -170,6 +182,7 @@ class Fio(Tool):
             verify_dump,
             verify_fatal,
             verify,
+            ioengine,
         )
         process = self.run_async(
             cmd,
@@ -244,7 +257,7 @@ class Fio(Tool):
             fio_message.append(fio_result_message)
         return fio_message
 
-    def _get_command(
+    def _get_command(  # noqa: C901
         self,
         name: str,
         filename: str,
@@ -264,10 +277,13 @@ class Fio(Tool):
         verify_dump: bool = False,
         verify_fatal: bool = False,
         verify: str = "",
+        ioengine: IoEngine = IoEngine.LIBAIO,
     ) -> str:
-        ioengine = "posixaio" if isinstance(self.node.os, BSD) else "libaio"
+        if isinstance(self.node.os, BSD):
+            ioengine = IoEngine.POSIXAIO
+
         cmd = (
-            f"--ioengine={ioengine} --filename={filename} "
+            f"--ioengine={ioengine.value} --filename={filename} "
             f"--readwrite={mode} --iodepth={iodepth} "
             f"--name={name}"
         )
--- a/microsoft/testsuites/performance/common.py
+++ b/microsoft/testsuites/performance/common.py
@@ -44,6 +44,7 @@ from lisa.tools import (
     Ssh,
     Sysctl,
 )
+from lisa.tools.fio import IoEngine
 from lisa.tools.ntttcp import (
     NTTTCP_TCP_CONCURRENCY,
     NTTTCP_TCP_CONCURRENCY_BSD,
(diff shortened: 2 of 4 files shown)


lisa/tools/fio.py
```diff
@@ -264,8 +268,11 @@ def _get_command(
         verify_dump: bool = False,
         verify_fatal: bool = False,
         verify: str = "",
+        ioengine: str = "",
     ) -> str:
-        ioengine = "posixaio" if isinstance(self.node.os, BSD) else "libaio"
+        if ioengine == "":
```
squirrelsc: If it's BSD and specified ioengine, it will fail. Move the `"posixaio"` to default value of function declaration, and use `if isinstance(..., BSD)` to always enforce libaio for BSD.

microsoft/testsuites/performance/common.py
```diff
@@ -84,7 +86,9 @@ def perf_disk(
     # fail with EAGAIN.
     # read: https://www.kernel.org/doc/Documentation/sysctl/fs.txt
     # So we set numjob to 256 if numjob is larger than 256.
-    numjob = min(numjob, 256)
+    # This limitation is only needed for 'libaio' ioengine but not for 'io_uring'.
+    if ioengine == "libaio":
```
squirrelsc: Use an enum for the ioengine types to avoid typo.

microsoft/testsuites/performance/nvmeperf.py
```diff
@@ -72,3 +72,52 @@ def perf_nvme(self, node: Node, result: TestResult) -> None:
...
+
+    @TestCaseMetadata(
+        description="""
+        This test case uses fio to test NVMe disk performance
+        with 'io_uring' as ioengine.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            supported_features=[NvmeSettings(disk_count=8)],
```
squirrelsc: Does it need 8 disks to test io_uring?

squirrelsc: The test case should be skipped, if it's not Nvme disk. Please make sure it works like that.

SRIKKANTH: The logic is already in the testcase.

SRIKKANTH: Reduced the requirement to 1 disk.

microsoft/testsuites/performance/common.py
SRIKKANTH: Added 3 new constants for io engines in "lisa\lisa\util\constants.py"

squirrelsc: Constants reduce typos for the string. The enum limit options of this parameter.

microsoft/testsuites/performance/nvmeperf.py
squirrelsc: 1. The requirements can find vm size with nvme by feature implementation. If the checks are in test case, it may not match to right vm size for general distro testing.
2. In the test case, I see assertion, but no skipped exception. It makes case failed, instead of skip for unmatched env.

lisa/tools/fio.py
SRIKKANTH: Fixed as suggested

microsoft/testsuites/performance/common.py
SRIKKANTH: Changed the logic to use enum

microsoft/testsuites/performance/nvmeperf.py
SRIKKANTH: Changed the requirement to supported_features=[Nvme].

microsoft/testsuites/performance/nvmeperf.py
```diff
@@ -72,3 +73,52 @@ def perf_nvme(self, node: Node, result: TestResult) -> None:
...
+        with 'io_uring' as ioengine.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            supported_features=[Nvme],
+        ),
+    )
+    def perf_nvme_io_uring(self, node: Node, result: TestResult) -> None:
+        nvme = node.features[Nvme]
```
LiliDeng, line 61: use a method for common code used in perf_nvme_io_uring and perf_nvme

SRIKKANTH: Added common method "_perf_nvme" as suggested

lisa/tools/fio.py
```diff
@@ -264,10 +278,13 @@ def _get_command(
...
         verify: str = "",
+        ioengine: IoEngine = IoEngine.LIBAIO,
     ) -> str:
-        ioengine = "posixaio" if isinstance(self.node.os, BSD) else "libaio"
+        if isinstance(self.node.os, BSD):
+            ioengine = IoEngine.POSIXAIO
+
         cmd = (
-            f"--ioengine={ioengine} --filename={filename} "
+            f"--ioengine={ioengine.value} --filename={filename} "
```
LiliDeng, line 286: make sure the value is not none here.

image

SRIKKANTH: Fixed this issue.

microsoft/testsuites/performance/nvmeperf.py
```diff
@@ -37,6 +38,25 @@ class NvmePerformace(TestSuite):
         ),
     )
     def perf_nvme(self, node: Node, result: TestResult) -> None:
+        self._perf_nvme(node, IoEngine.LIBAIO, result)
```
squirrelsc, line 42: Update test case comments to specify what io engine is used.

microsoft/testsuites/performance/nvmeperf.py
```diff
@@ -37,6 +39,25 @@ class NvmePerformace(TestSuite):
...
+        description="""
+        This test case uses fio to test NVMe disk performance
+        using 'io_uring' as ioengine.
+        """,
+        priority=3,
+        timeout=TIME_OUT,
+        requirement=simple_requirement(
+            supported_features=[Nvme],
+        ),
+    )
```
LiliDeng, line 54: for perf_nvme, it has requirement for count of nvme disks, for perf_nvme_io_uring, by default it uses 1 nvme disk, is it expected?

LiliDeng (changes requested):

microsoft/testsuites/performance/nvmeperf.py
SRIKKANTH: There are new VMs with just 1 NVMe. So its a valid case.

microsoft/testsuites/performance/nvmeperf.py
SRIKKANTH: Updated.

LiliDeng (approved):

PR #3420, opened 2024-09-14
Support PCI IDs

SRIKKANTH:
Support PCI IDs
This will help in identifying devices even better.
Example: NVMe local devices vs remote storage devices when disc controller type == NVMe.


lisa/tools/lspci.py
```diff
@@ -78,43 +116,21 @@ def __str__(self) -> str:
...
-                lines=self.device_info,
-                pattern=PATTERN_DEVICE_ID,
-                single_line=False,
-            )
-            self.device_id = device_id_raw.get("id", "")
-            assert self.device_id, f"cannot find device id from {raw_str}"
+    def parse(self, raw_str: str, pci_ids: Dict[str, Any]) -> None:
+        matched_pci_device_info = PATTERN_PCI_DEVICE.match(raw_str)
+        if matched_pci_device_info:
+            self.slot = matched_pci_device_info.group("slot")
```
squirrelsc: What's the reason to remove assertions?

lisa/tools/lspci.py
```diff
@@ -142,47 +158,107 @@ def _install(self) -> bool:
...
+                ):
+                    device_type_list.append(device)
+        else:
+            class_names = DEVICE_TYPE_DICT[device_type.upper()]
+            device_type_list = [
+                x for x in devices_list if x.device_class in class_names
+            ]
         return device_type_list
 
+    @retry(KeyError, tries=10, delay=20)
```
squirrelsc: Why retry can fix key errors?

SRIKKANTH: Some times difference seen between outputs of "lspci -n" and "lspci -m" used in this routine.
It usually occurs while checking for VFs right after VM booting when the VF count is 8. 
All VF becomes available only after couple of minutes of VM booting.
This is for re-trying in such cases instead of simply failing.

lisa/tools/lspci.py
SRIKKANTH: Added them back.

lisa/tools/lspci.py
```diff
@@ -142,47 +165,107 @@ def _install(self) -> bool:
...
                 f"pci_type '{device_type}' is not supported to be searched."
             )
-        class_names = DEVICE_TYPE_DICT[device_type.upper()]
         devices_list = self.get_devices(force_run)
-        device_type_list = [x for x in devices_list if x.device_class in class_names]
+        device_type_list = []
+        if use_pci_ids:
+            for device in devices_list:
+                if (
+                    device.controller_id in CONTROLLER_ID_DICT[device_type.upper()]
```
squirrelsc: Define a class instead of 3 dict with the same key. The class can define a method to accept a device, and compare if exists in one dict.

lisa/tools/lspci.py
squirrelsc: Can you check which one of "-n" or "-m" is incorrect, and why it's fixed by rerun? 10*20 means 200 seconds, if this method called many times, it slows down the test cases significantly.

lisa/tools/lspci.py
```diff
@@ -142,47 +165,107 @@ def _install(self) -> bool:
         return self._check_exists()
 
     def get_device_names_by_type(
-        self, device_type: str, force_run: bool = False
+        self, device_type: str, force_run: bool = False, use_pci_ids: bool = False
```
squirrelsc: It looks the new argument is to filter devices by name? If so, you can add a new method called `get_devices_by_allowed_list`, it calls get_device_names_by_type and filter them.

lisa/tools/lspci.py
```diff
@@ -78,43 +116,28 @@ def __str__(self) -> str:
...
 
-    def parse(self, raw_str: str) -> None:
-        matched_pci_device_info_list = find_groups_in_lines(
-            lines=raw_str,
-            pattern=PATTERN_PCI_DEVICE,
-        )
-        if matched_pci_device_info_list:
-            matched_pci_device_info = matched_pci_device_info_list[0]
-            self.slot = matched_pci_device_info.get("slot", "").strip()
+    def parse(self, raw_str: str, pci_ids: Dict[str, Any]) -> None:
```
squirrelsc: This method looks not need to be changed, the filter happens on upper caller too. It doesn't need to filter in this method.

microsoft/testsuites/network/stress.py
```diff
@@ -274,6 +275,8 @@ def stress_sriov_with_max_nics_reboot_from_platform(
             for node in environment.nodes.list():
                 start_stop = node.features[StartStop]
                 start_stop.restart()
+            # Add delay to wait for the network interface ready.
+            sleep(120)
```
squirrelsc: Don't add sleep, and it's so long. Find a way to wait some signal.

lisa/tools/lspci.py
SRIKKANTH: Both -n and -m are correct. Its not an issue of lspci commands. Its the number of PCI devices newly added between these command executions is causing the exception. With out this wait, testcases will be dealing with less number of VF devices than actually attached.

LiliDeng:
Make sure you have tested it. like `verify_device_statistics`

lisa/tools/lspci.py
SRIKKANTH: Found that comparing 3 values is not needed. Changed the logic to compare only the controller id.

lisa/tools/lspci.py
SRIKKANTH: Removed usage of a parameter "use_pci_ids" from all places. Its no longer needed.

lisa/tools/lspci.py
SRIKKANTH: Fixed it.

microsoft/testsuites/network/stress.py
SRIKKANTH: Removed sleep from test cases. As retries in "get_devices" method fixed it.

lisa/tools/lspci.py
```diff
@@ -78,43 +120,25 @@ def __str__(self) -> str:
...
 
     def parse(self, raw_str: str) -> None:
-        matched_pci_device_info_list = find_groups_in_lines(
-            lines=raw_str,
-            pattern=PATTERN_PCI_DEVICE,
-        )
-        if matched_pci_device_info_list:
-            matched_pci_device_info = matched_pci_device_info_list[0]
-            self.slot = matched_pci_device_info.get("slot", "").strip()
+        matched_pci_device_info = PATTERN_PCI_DEVICE.match(raw_str)
```
squirrelsc: use `find_group_in_lines` (not groups), so it returns one dict.

lisa/tools/lspci.py
```diff
@@ -158,18 +185,52 @@ def get_devices_by_type(
...
         devices_list = self.get_devices(force_run)
-        device_type_list = [x for x in devices_list if x.device_class in class_names]
+        device_type_list = []
+        for device in devices_list:
+            if device.device_id in DEVICE_ID_DICT[device_type.upper()]:
+                device_type_list.append(device)
+
         return device_type_list
 
+    @retry(KeyError, tries=10, delay=10)
```
squirrelsc: Add code comments above to explain the retry, and update `delay` to 1 or 2 seconds, so it can run faster.

lisa/tools/lspci.py
SRIKKANTH: Fixed it.

lisa/tools/lspci.py
SRIKKANTH: Added comments about retry.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
see this exception
```Sriov.verify_sriov_basic: FAILED   failed. KeyError: 'cannot find device id from 00:00.0'```

SRIKKANTH:
> see this exception `Sriov.verify_sriov_basic: FAILED failed. KeyError: 'cannot find device id from 00:00.0'`

Fixed the issue.

LiliDeng (approved):

PR #3422, opened 2024-09-18
Add tftp deployment on BareMetal platform

paulli2017:
Added SOC support on BareMetal platform

### DIFF
--- a/lisa/base_tools/sed.py
+++ b/lisa/base_tools/sed.py
@@ -47,12 +47,17 @@ class Sed(Tool):
         self,
         text: str,
         file: str,
+        match_line: str = "",
         sudo: bool = False,
     ) -> None:
         # always force run, make sure it happens every time.
         text = text.replace('"', '\\"')
+        if match_line:
+            append_line = f"{match_line}"
+        else:
+            append_line = "$"
         result = self.run(
-            f"-i.bak '$a{text}' {file}",
+            f"-i.bak '{append_line}a{text}' {file}",
             force_run=True,
             no_error_log=True,
             no_info_log=True,
--- a/lisa/sut_orchestrator/baremetal/bootconfig.py
+++ b/lisa/sut_orchestrator/baremetal/bootconfig.py
@@ -1,0 +1,165 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from dataclasses import dataclass, field
+from pathlib import PurePosixPath
+from typing import Optional, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import schema
+from lisa.node import quick_connect
+from lisa.tools import Cat, Sed
+from lisa.util import (
+    InitializableMixin,
+    LisaException,
+    field_metadata,
+    find_group_in_lines,
+    subclasses,
+)
+from lisa.util.logger import get_logger
+
+from .schema import BootConfigSchema
+
+BOOT_LABEL = "label lisa baremetal"
+
+
+class BootConfig(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
+    def __init__(
+        self,
+        runbook: BootConfigSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.boot_config_runbook: BootConfigSchema = self.runbook
+        self._log = get_logger("boot_config", self.__class__.__name__)
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return BootConfigSchema
+
+    def config(self) -> None:
+        raise NotImplementedError()
+
+
+@dataclass_json()
+@dataclass
+class PxeBootSchema(BootConfigSchema):
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    host_name: str = field(default="", metadata=field_metadata(required=True))
+    image_source: str = field(default="", metadata=field_metadata(required=True))
+    kernel_boot_params: Optional[str] = field(default="")
+
+
+class PxeBoot(BootConfig):
+    def __init__(
+        self,
+        runbook: PxeBootSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.pxe_runbook: PxeBootSchema = self.runbook
+        self._log = get_logger("pxe_boot", self.__class__.__name__)
+        self._boot_dir = "/var/lib/tftpboot/"
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "pxe_boot"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return PxeBootSchema
+
+    def config(self) -> None:
+        assert self.pxe_runbook.connection, "connection is required for dhcp_server"
+
+        self.pxe_runbook.connection.name = "dhcp_server"
+        self._dhcp_server = quick_connect(
+            self.pxe_runbook.connection, logger_name="dhcp_server"
+        )
+
+        node_name = f"{self.pxe_runbook.host_name}"
+        pxe_boot_config_path = self._get_pxe_config_path(node_name)
+        if not pxe_boot_config_path:
+            with open(pxe_boot_config_path, "w") as f:
+                f.write("timeout 150\n\nmenu title selections\n\n")
+
+        boot_image = PurePosixPath(
+            self.pxe_runbook.image_source,
+        ).relative_to(self._boot_dir)
+        boot_entry = f"kernel {boot_image}"
+
+        sed = self._dhcp_server.tools[Sed]
+        # Delete the label if one is existed
+        sed.delete_lines(
+            f"^{BOOT_LABEL}$/,/^$",
+            PurePosixPath(pxe_boot_config_path),
+        )
+        self._log.debug(
+            "Deleted boot entry for LISA if existed"
+            f"{pxe_boot_config_path}, "
+            f"pointing to {boot_entry}"
+        )
+
+        # Add one at the start
+        params = self.pxe_runbook.kernel_boot_params
+        append = f"\\n    append {params}" if params else ""
+        match_str = "/^menu.*/"
+        sed.append(
+            f" \\\\n{BOOT_LABEL}\\n    {boot_entry}{append}",
+            f"{pxe_boot_config_path}",
+            f"{match_str}",
+        )
+
+        self._log.debug(
+            "Added boot entry for LISA at "
+            f"{pxe_boot_config_path}, "
+            f"pointing to {boot_entry}"
+        )
+
+    def _get_pxe_config_path(self, node_name: str) -> str:
+        cat = self._dhcp_server.tools[Cat]
+        output_dhcp_info = cat.read(
+            "/etc/dhcp/dhcpd.conf",
+            force_run=True,
+        )
+        # Here is part of output_dhcp_info
+        # ...
+        # host dev-gp9 {
+        #    hardware ethernet 04:27:28:06:f7:88;
+        #    fixed-address 192.168.3.123;
+        #    option host-name "blade3";
+        # }
+        #
+        # host dev-gp10 {
+        #    hardware ethernet 04:27:28:15:3f:f0;
+        #    fixed-address 192.168.3.117;
+        #    option host-name "blade4";
+        # }
+        # ...
+        # if node 10 is used to be bootup from pxe_server,
+        # configuration file for node 10 on pxe_server needs
+        # to be modified. The configuration file's name for
+        # node 10 is actually based on its physical address:
+        # 04:27:28:15:3f:f0 as you can see from the above,
+        # and it is called 01-04-27-28-15-3f-f0 under bootup
+        # directory. Below is to find the node's physical
+        # address string 04:27:28:15:3f:f0 from output_dhcp_info
+        # and then obtain the configuration file name for
+        # later bootup menu change.
+        host_ref = f"host {node_name}"
+        pattern = host_ref + r"\s+\{\r?\n\s+hardware ethernet\s+(?P<mac>[0-9a-f:]+);"
+        node_pattern = re.compile(pattern, re.M)
+        node_address = find_group_in_lines(
+            lines=output_dhcp_info,
+            pattern=node_pattern,
+            single_line=False,
+        )
+        if node_address:
+            config_file = "01-" + node_address["mac"].replace(":", "-")
+            config_file_fullpath = self._boot_dir + f"pxelinux.cfg/{config_file}"
+        else:
+            raise LisaException(f"Failed to find DHCP entry for {node_name}")
+
+        return config_file_fullpath
--- a/lisa/sut_orchestrator/baremetal/build.py
+++ b/lisa/sut_orchestrator/baremetal/build.py
@@ -3,16 +3,41 @@
 
 import os
 import re
-from pathlib import Path
+from pathlib import Path, PurePath
 from typing import Dict, List, Type
 
 from smb.SMBConnection import SMBConnection  # type: ignore
 
 from lisa import schema
+from lisa.node import quick_connect
+from lisa.tools import Ls, RemoteCopy
 from lisa.util import ContextMixin, InitializableMixin, subclasses
 from lisa.util.logger import get_logger
 
-from .schema import BuildSchema, FileSchema, SMBBuildSchema
+from .schema import BuildSchema, FileSchema, SMBBuildSchema, TftpBuildSchema
+
+
+def _find_matched_files(
+    sources_path: List[Path],
+    files_map: List[FileSchema],
+) -> Dict[str, FileSchema]:
+    all_files = []
+    match_files: Dict[str, FileSchema] = {}
+    for source_path in sources_path:
+        for root, _, files in os.walk(source_path):
+            for file in files:
+                all_files.append(os.path.join(root, file))
+
+        for file_map in files_map:
+            file_path = rf"{source_path}\{file_map.source}".replace("\\", "\\\\")
+            pattern = re.compile(
+                file_path,
+                re.I | re.M,
+            )
+            for file in all_files:
+                if pattern.match(file):
+                    match_files[file] = file_map
+    return match_files
 
 
 class Build(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixin):
@@ -55,8 +80,9 @@ class SMBBuild(Build):
         ) as conn:
             conn.connect(server_name)
 
-            for file, file_map in self._find_matched_files(
-                sources_path, files_map
+            for file, file_map in _find_matched_files(
+                sources_path,
+                files_map,
             ).items():
                 with open(file, "rb") as f:
                     if file_map.destination:
@@ -79,23 +105,43 @@ class SMBBuild(Build):
                         )
                 self._log.debug(f"copy file {file} to {share_name}\\{file_name}")
 
-    def _find_matched_files(
-        self, sources_path: List[Path], files_map: List[FileSchema]
-    ) -> Dict[str, FileSchema]:
-        all_files = []
-        match_files: Dict[str, FileSchema] = {}
-        for source_path in sources_path:
-            for root, _, files in os.walk(source_path):
-                for file in files:
-                    all_files.append(os.path.join(root, file))
 
-            for file_map in files_map:
-                file_path = rf"{source_path}\{file_map.source}".replace("\\", "\\\\")
-                pattern = re.compile(
-                    file_path,
-                    re.I | re.M,
-                )
-                for file in all_files:
-                    if pattern.match(file):
-                        match_files[file] = file_map
-        return match_files
+class TftpBuild(Build):
+    def __init__(self, runbook: TftpBuildSchema) -> None:
+        super().__init__(runbook)
+        self.pxe_runbook: TftpBuildSchema = self.runbook
+
+    @classmethod
+    def type_name(cls) -> str:
+        return "tftp"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return TftpBuildSchema
+
+    def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
+        assert self.pxe_runbook.connection, "The build server is not specified"
+
+        build_server = quick_connect(
+            self.pxe_runbook.connection,
+            logger_name="build_server",
+        )
+
+        ls = build_server.tools[Ls]
+        rc = build_server.tools[RemoteCopy]
+        self._log.debug(f"Copying files to: {build_server}")
+        for file, file_map in _find_matched_files(
+            sources_path,
+            files_map,
+        ).items():
+            if file_map.destination:
+                file_map_path = PurePath(file_map.destination)
+                if ls.is_file(file_map_path):
+                    file_destination = file_map_path.parent
+                else:
+                    file_destination = file_map_path
+            else:
+                file_destination = PurePath(file).parent
+
+            rc.copy_to_remote(PurePath(file), file_destination)
+        self._log.debug(f"Copied files to: {build_server}")
--- a/lisa/sut_orchestrator/baremetal/platform_.py
+++ b/lisa/sut_orchestrator/baremetal/platform_.py
@@ -7,12 +7,12 @@ from typing import Any, List, Optional, Type
 from lisa import RemoteNode, feature, schema, search_space
 from lisa.environment import Environment
 from lisa.platform_ import Platform
-from lisa.util import fields_to_dict
 from lisa.util.logger import Logger
 from lisa.util.shell import try_connect
 from lisa.util.subclasses import Factory
 
 from .. import BAREMETAL
+from .bootconfig import BootConfig
 from .build import Build
 from .cluster.cluster import Cluster
 from .context import get_build_context, get_node_context
@@ -52,6 +52,7 @@ class BareMetalPlatform(Platform):
         self.key_loader_factory = Factory[KeyLoader](KeyLoader)
         self.source_factory = Factory[Source](Source)
         self.build_factory = Factory[Build](Build)
+        self.boot_config_factory = Factory[BootConfig](BootConfig)
         # currently only support one cluster
         assert self._baremetal_runbook.cluster, "no cluster is specified in the runbook"
         self._cluster_runbook = self._baremetal_runbook.cluster[0]
@@ -66,7 +67,49 @@ class BareMetalPlatform(Platform):
         return self._configure_node_capabilities(environment, log, client_capabilities)
 
     def _deploy_environment(self, environment: Environment, log: Logger) -> None:
-        # copy build (shared, check if it's copied)
+        # process the cluster elements from runbook
+        self._predeploy_environment(environment, log)
+
+        # deploy cluster
+        self.cluster.deploy(environment)
+
+        if self._cluster_runbook.ready_checker:
+            ready_checker = self.ready_checker_factory.create_by_runbook(
+                self._cluster_runbook.ready_checker
+            )
+
+        for index, node in enumerate(environment.nodes.list()):
+            node_context = get_node_context(node)
+
+            # ready checker
+            if ready_checker:
+                ready_checker.is_ready(node)
+
+            # get ip address
+            if self._cluster_runbook.ip_getter:
+                ip_getter = self.ip_getter_factory.create_by_runbook(
+                    self._cluster_runbook.ip_getter
+                )
+                node_context.connection.address = ip_getter.get_ip()
+
+            assert isinstance(node, RemoteNode), f"actual: {type(node)}"
+            node.name = f"node_{index}"
+            try_connect(node_context.connection)
+
+        self._log.debug(f"deploy environment {environment.name} successfully")
+
+    def _copy(self, build_schema: BuildSchema, sources_path: List[Path]) -> None:
+        if sources_path:
+            build = self.build_factory.create_by_runbook(build_schema)
+            build.copy(
+                sources_path=sources_path,
+                files_map=build_schema.files,
+            )
+        else:
+            self._log.debug("no copied source path specified, skip copy")
+
+    def _predeploy_environment(self, environment: Environment, log: Logger) -> None:
+        # download source (shared, check if it's copied)
         if self._baremetal_runbook.source:
             if not self.local_artifacts_path:
                 source = self.source_factory.create_by_runbook(
@@ -99,10 +142,16 @@ class BareMetalPlatform(Platform):
                 self._log.debug("build is already copied, skip copy")
             else:
                 assert self.local_artifacts_path, "no build source is specified"
-                self.copy(
+                self._copy(
                     self.cluster.runbook.build, sources_path=self.local_artifacts_path
                 )
                 build_context.is_copied = True
+
+        if self.cluster.runbook.boot_config:
+            boot_config = self.boot_config_factory.create_by_runbook(
+                self.cluster.runbook.boot_config
+            )
+            boot_config.config()
 
         if self.cluster.runbook.key_loader:
             key_loader = self.key_loader_factory.create_by_runbook(
@@ -141,50 +190,6 @@ class BareMetalPlatform(Platform):
 
             node_context.connection = connection_info
             index = index + 1
-
-        # deploy cluster
-        self.cluster.deploy(environment)
-
-        if self._cluster_runbook.ready_checker:
-            ready_checker = self.ready_checker_factory.create_by_runbook(
-                self._cluster_runbook.ready_checker
-            )
-
-        for index, node in enumerate(environment.nodes.list()):
-            node_context = get_node_context(node)
-
-            # ready checker
-            if ready_checker:
-                ready_checker.is_ready(node)
-
-            # get ip address
-            if self._cluster_runbook.ip_getter:
-                ip_getter = self.ip_getter_factory.create_by_runbook(
-                    self._cluster_runbook.ip_getter
-                )
-                node_context.connection.address = ip_getter.get_ip()
-
-            assert isinstance(node, RemoteNode), f"actual: {type(node)}"
-            node.name = f"node_{index}"
-            node.set_connection_info(
-                **fields_to_dict(
-                    node_context.connection,
-                    ["address", "port", "username", "password", "private_key_file"],
-                ),
-            )
-            try_connect(connection_info)
-
-        self._log.debug(f"deploy environment {environment.name} successfully")
-
-    def copy(self, build_schema: BuildSchema, sources_path: List[Path]) -> None:
-        if sources_path:
-            build = self.build_factory.create_by_runbook(build_schema)
-            build.copy(
-                sources_path=sources_path,
-                files_map=build_schema.files,
-            )
-        else:
-            self._log.debug("no copied source path specified, skip copy")
 
     def _configure_node_capabilities(
         self,
--- a/lisa/sut_orchestrator/baremetal/readychecker.py
+++ b/lisa/sut_orchestrator/baremetal/readychecker.py
@@ -10,13 +10,7 @@ from dataclasses_json import dataclass_json
 
 from lisa import schema
 from lisa.node import Node, RemoteNode
-from lisa.util import (
-    InitializableMixin,
-    LisaException,
-    check_till_timeout,
-    fields_to_dict,
-    subclasses,
-)
+from lisa.util import InitializableMixin, LisaException, check_till_timeout, subclasses
 from lisa.util.logger import get_logger
 from lisa.util.shell import try_connect
 
@@ -115,12 +109,19 @@ class SshChecker(ReadyChecker):
         context = get_node_context(node)
         remote_node = cast(RemoteNode, node)
         remote_node.set_connection_info(
-            **fields_to_dict(
-                context.connection,
-                ["address", "port", "username", "password", "private_key_file"],
+            address=context.connection.address,
+            public_port=context.connection.port,
+            username=context.connection.username,
+            password=cast(
+                str,
+                context.connection.password,
+            ),
+            private_key_file=cast(
+                str,
+                context.connection.private_key_file,
             ),
         )
-        self._log.debug("try to connect to the client")
+        self._log.debug(f"try to connect to client: {node}")
         try_connect(context.connection, ssh_timeout=self.ssh_runbook.timeout)
         self._log.debug("client has been connected successfully")
         return True
--- a/lisa/sut_orchestrator/baremetal/schema.py
+++ b/lisa/sut_orchestrator/baremetal/schema.py
@@ -75,12 +75,19 @@ class KeyLoaderSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
 
 @dataclass_json()
 @dataclass
+class BootConfigSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
+    type: str = field(default="boot_config", metadata=field_metadata(required=True))
+
+
+@dataclass_json()
+@dataclass
 class ClusterSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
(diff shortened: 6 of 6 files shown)


lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -134,6 +141,21 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
...
+class TftpBuildSchema(BuildSchema):
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    share: str = ""
+    server_name: str = ""
+
+    def __post_init__(self, *args: Any, **kwargs: Any) -> None:
+        if self.connection:
+            add_secret(self.connection.username, PATTERN_HEADTAIL)
```
squirrelsc: username is not a secret.

lisa/sut_orchestrator/baremetal/readychecker.py
```diff
@@ -115,12 +109,19 @@ def is_ready(self, node: Node) -> bool:
         context = get_node_context(node)
         remote_node = cast(RemoteNode, node)
         remote_node.set_connection_info(
-            **fields_to_dict(
-                context.connection,
-                ["address", "port", "username", "password", "private_key_file"],
+            address=context.connection.address,
```
squirrelsc, line 112: Why change this part?

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -66,7 +67,49 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
...
+        if sources_path:
+            build = self.build_factory.create_by_runbook(build_schema)
+            build.copy(
+                sources_path=sources_path,
+                files_map=build_schema.files,
+            )
+        else:
+            self._log.debug("no copied source path specified, skip copy")
+
+    def _process_cluster_elements(self, environment: Environment, log: Logger) -> None:
```
squirrelsc: Call it like `_predeploy_environment`. It should be clearer than "process".

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -24,6 +26,28 @@ def __init__(self, runbook: BuildSchema) -> None:
     def type_schema(cls) -> Type[schema.TypedSchema]:
         return BuildSchema
 
+    @classmethod
+    def find_matched_files(
```
squirrelsc: Move it to a private function in this file.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +104,68 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+        return "tftp"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return TftpBuildSchema
+
+    def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
+        assert self.pxe_runbook.connection, "The build server is not specified"
+
+        server_address = self.pxe_runbook.connection.address
```
squirrelsc: Create a lisa remotenode, and use the remote_copy tool to copy.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +104,68 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+                    "failed to copy file {file} to {file_name}",
+                    exc_info=ex,
                 )
-                for file in all_files:
-                    if pattern.match(file):
-                        match_files[file] = file_map
-        return match_files
+
+        try:
+            sftp.close()
```
squirrelsc: As above comment, it should use remote_copy tool. But the close pattern should like below to avoid leak resource.
```python
try:
      try:
            transport = paramiko.Transport((server_address, server_port))
            transport.connect(
                username=server_username,
                password=server_password,
            )
        except Exception as ex:
            self._log.exception(
                "failed to connect to {server_address}:{server_port}",
                exc_info=ex,
            )
    ...
final:
    try:
        if sftp:
            sftp.close()
        if transport:
            transport.close()
    except Exception as ex:
        self._log.exception(
            "failed to close the connection",
            exc_info=ex,
        )
```

lisa/sut_orchestrator/baremetal/readychecker.py
paulli2017: set_connection_info() has both port and public_port parameter with default value at 22, where public_port value is used to connect with the node during later stage. fields_to_dict(context.connection,  ["address", "port", "username", "password", "private_key_file"]) only update the port value, instead of public_port value, which cause connection issue to the node that has access port value other than 22.

lisa/sut_orchestrator/baremetal/platform_.py
paulli2017: the change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, please review it again, thanks

lisa/sut_orchestrator/baremetal/schema.py
paulli2017: you mean no need of username update?

lisa/sut_orchestrator/baremetal/readychecker.py
squirrelsc: I think you can set both public_port and port in runbook. The public_port/port is used to allow different ports for LISA/node and node/node communication.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -9,10 +9,35 @@
...
+from lisa.node import quick_connect
+from lisa.tools import Ls, RemoteCopy
 from lisa.util import ContextMixin, InitializableMixin, subclasses
 from lisa.util.logger import get_logger
 
-from .schema import BuildSchema, FileSchema, SMBBuildSchema
+from .schema import BuildSchema, FileSchema, SMBBuildSchema, TftpBuildSchema
+
+
+def find_matched_files(
```
squirrelsc: private method should be start with `_`

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,54 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+            sources_path,
+            files_map,
+        ).items():
+            if file_map.destination:
+                if ls.is_file(Path(file_map.destination)):
+                    file_destination = os.path.dirname(Path(file_map.destination))
+                else:
+                    file_destination = file_map.destination
+            else:
+                file_destination = file.rsplit("\\")[-1]
```
squirrelsc: convert the path to a PurePath, and then split. So it works for both windows and linux path.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,170 @@
...
+        output_dhcp_info = self.pxe_server.execute(
+            cmd="cat /etc/dhcp/dhcpd.conf",
+            expected_exit_code_failure_message=(
+                f"failed to obtain {node_name}'s configuration information\n"
+            ),
+            expected_exit_code=0,
+            sudo=False,
+        ).stdout
+
+        board_info = output_dhcp_info[output_dhcp_info.index(host_ref) :]
```
squirrelsc: Use regexp, and provide examples above.

lisa/sut_orchestrator/baremetal/readychecker.py
paulli2017: I tried to set both public_port and port to be same value other than 22, however, from my debug, context.connection structure contains port value only, fields_to_dict(context.connection, ["address", "port", "username", "password", "private_key_file"] pass this port value to set_connection_info() function, but the default public_port parameter of set_connection_info() is still 22. Unfortunately, set_connection_info() use public_port value instead of port value as the connection port from its implementation. If set_connection_info() is called, its input parameter of public_port needs to be updated in order to connect the node successfully.

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,54 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+            sources_path,
+            files_map,
+        ).items():
+            if file_map.destination:
+                if ls.is_file(PurePath(file_map.destination)):
+                    file_destination = os.path.dirname(PurePath(file_map.destination))
+                else:
+                    file_destination = file_map.destination
+            else:
+                file_destination = os.path.dirname(PurePath(file))
```
squirrelsc: 1. You don't need to convert between PurePath and string forth and back. Just keep a PurePath, and pass it into the `copy_to_remote`.
2. For remote path, it needs to call `Node.get_pure_path` to get pure path.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+        # node 10 is actually based on its physical address:
+        # 04:27:28:15:3f:f0 as you can see from the above,
+        # and it is called 01-04-27-28-15-3f-f0 under bootup
+        # directory. Below is to find the node's physical
+        # address string 04:27:28:15:3f:f0 from output_dhcp_info
+        # and then obtain the configuration file name for
+        # later bootup menu change.
+        ethern_part = r"\s\{[\r\n]?\s+hardware ethernet\s+"
+        host_part = re.escape(host_ref) + ethern_part
+        node_str = host_part + r"([0-9a-f:]+)"
```
squirrelsc: One line is clearer, and you can capture both host and address, then find it in Python code. And you can test pattern, examples online like https://regex101.com/.
```
re.compile(r"^host\s+(?P<device>.+)\s+\{\n\s+hardware ethernet\s(?P<address>[0-9a-f:]+);$")
```

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+        node_pattern = re.compile(node_str, re.M)
+        node_address = node_pattern.search(output_dhcp_info)
```
squirrelsc: Use LISA internal wrappers for regexp to unify the patterns.

lisa/sut_orchestrator/baremetal/readychecker.py
squirrelsc: @LiliDeng why there is to set connection info? How about call node.set_connection_info below?

https://github.com/paulli2017/lisa/blob/709497ef75f8178ab1d730946958f9be1c5ef85c/lisa/sut_orchestrator/baremetal/platform_.py#L191-L192

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+
+class PxeBoot(BootConfig):
+    def __init__(
+        self,
+        runbook: PxeBootSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.pxe_runbook: PxeBootSchema = self.runbook
+        self._log = get_logger("pxe_boot", self.__class__.__name__)
+        self._boot_label = "label lisav3"
```
squirrelsc: `label lisa baremetal` includes more useful information.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+@dataclass_json()
+@dataclass
+class PxeBootSchema(BootConfigSchema):
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    node_setup: str = field(default="", metadata=field_metadata(required=True))
+    node_number: int = field(
```
squirrelsc: merge `node_setup` and `node_number` to same field `host_name`. It looks they only used once within the same string.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.boot_config_runbook: BootConfigSchema = self.runbook
+        self._log = get_logger("boot_config", self.__class__.__name__)
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return BootConfigSchema
+
+    def set_boot_config(self) -> None:
```
squirrelsc: `config`

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+
+    def _connect_to_pxe_server(self) -> None:
+        assert self.pxe_runbook.connection, "connection is required for pxe_server"
+        self.pxe_runbook.connection.name = "pxe_server"
+        self.pxe_server = quick_connect(
+            self.pxe_runbook.connection, logger_name="pxe_server"
+        )
+
+    def _get_dhcp_info(self, node_name: str, host_ref: str) -> str:
```
squirrelsc: This method output the pxe config path. So the file name should be like `_get_pxe_config_path`.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return PxeBootSchema
+
+    def set_boot_config(self) -> None:
+        try:
+            self._connect_to_pxe_server()
+        except Exception as ex:
+            self._log.debug(
```
squirrelsc: It should fail, instead of output debug messaage.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+        self._log.debug(
+            "Added boot entry for LISA at "
+            f"{pxe_boot_config_path}, "
+            f"pointing to {boot_entry}"
+        )
+
+    def _connect_to_pxe_server(self) -> None:
+        assert self.pxe_runbook.connection, "connection is required for pxe_server"
+        self.pxe_runbook.connection.name = "pxe_server"
+        self.pxe_server = quick_connect(
```
squirrelsc: The server is `dhcp_server`, there is no concept of "pxe server". PXE is a protocol on server side, and supported by DHCP/TFTP servers.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+    def __init__(
+        self,
+        runbook: PxeBootSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.pxe_runbook: PxeBootSchema = self.runbook
+        self._log = get_logger("pxe_boot", self.__class__.__name__)
+        self._boot_label = "label lisav3"
+        self._boot_dir = "/var/lib/tftpboot/"
+        self._boot_config = self._boot_dir + "pxelinux.cfg/"
```
squirrelsc: It looks `_boot_dir` and `_boot_config` are used once each. Can it use one variable?

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,178 @@
...
+        except Exception as ex:
+            self._log.debug(
+                f"failed to connect {self.pxe_runbook.connection}",
+                exc_info=ex,
+            )
+
+        node_name = f"{self.pxe_runbook.node_setup}{self.pxe_runbook.node_number}"
+        host_ref = f"host {node_name}"
+        pxe_boot_config_path = self._get_dhcp_info(node_name, host_ref)
+        if not pxe_boot_config_path:
```
squirrelsc, line 84: It's confusing on the reason of the path is empty. Raise exception in the `_get_dhcp_info`, when incorrect things are found.

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,162 @@
...
+        # to be modified. The configuration file's name for
+        # node 10 is actually based on its physical address:
+        # 04:27:28:15:3f:f0 as you can see from the above,
+        # and it is called 01-04-27-28-15-3f-f0 under bootup
+        # directory. Below is to find the node's physical
+        # address string 04:27:28:15:3f:f0 from output_dhcp_info
+        # and then obtain the configuration file name for
+        # later bootup menu change.
+        pattern = host_ref + r"\s+\{\r?\n\s+hardware ethernet\s+(?P<mac>[0-9a-f:]+);"
+        node_pattern = re.compile(pattern, re.M)
```
squirrelsc, line 154: use LISA internal method `find_group_in_lines`. The warpper is simpler.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,162 @@
...
+        try:
+            self._connect_to_dhcp_server()
+        except Exception as ex:
+            self._log.exception(
+                f"failed to connect {self.pxe_runbook.connection}",
+                exc_info=ex,
+            )
+
+        node_name = f"{self.pxe_runbook.host_name}"
+        host_ref = f"host {node_name}"
```
squirrelsc: This line should be moved into `_get_pxe_config_path`, because it's used in the method only.

lisa/sut_orchestrator/baremetal/bootconfig.py
squirrelsc: use LISA internal method `find_group_in_lines`. The wrapper is simpler.

lisa/sut_orchestrator/baremetal/bootconfig.py
squirrelsc: remove `3` in the name.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,162 @@
...
+
+@dataclass_json()
+@dataclass
+class PxeBootSchema(BootConfigSchema):
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    host_name: str = field(default="", metadata=field_metadata(required=True))
+    image_source: str = field(default="", metadata=field_metadata(required=True))
+    kernel_append: Optional[str] = field(default="")
```
squirrelsc: What's the value is used for? Is it just a string name?

lisa/base_tools/sed.py
```diff
@@ -47,18 +47,29 @@ def append(
...
         text = text.replace('"', '\\"')
-        result = self.run(
-            f"-i.bak '$a{text}' {file}",
-            force_run=True,
-            no_error_log=True,
-            no_info_log=True,
-            sudo=sudo,
-            shell=True,
-        )
+        if match_line:
```
squirrelsc, line 55: the `if/else` should include only the different part.

```python
       if match_line:
            pattern = f"{match_line}a{text}"
        else:
            pattern = f"$a{text}"
        result = self.run(
            f"-i.bak '{pattern}' {file}",
            force_run=True,
            no_error_log=True,
            no_info_log=True,
            sudo=sudo,
            shell=True,
        )
```

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,54 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+
+            try:
+                rc.copy_to_remote(PurePath(file), PurePath(file_destination))
+            except Exception as ex:
```
squirrelsc: Be careufly when catching and eating exceptions. File copy is a critical step, if the file is not copied, code should fail here. So users know something wrong, and fix it. If the exception is ate here, the error is hard to find in next steps.

lisa/sut_orchestrator/baremetal/platform_.py
```diff
@@ -66,7 +67,49 @@ def _prepare_environment(self, environment: Environment, log: Logger) -> bool:
...
+                )
+                node_context.connection.address = ip_getter.get_ip()
+
+            assert isinstance(node, RemoteNode), f"actual: {type(node)}"
+            node.name = f"node_{index}"
+            try_connect(node_context.connection)
+
+        self._log.debug(f"deploy environment {environment.name} successfully")
+
+    def copy(self, build_schema: BuildSchema, sources_path: List[Path]) -> None:
```
squirrelsc: move public methods above private methods.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,162 @@
...
+
+class PxeBoot(BootConfig):
+    def __init__(
+        self,
+        runbook: PxeBootSchema,
+    ) -> None:
+        super().__init__(runbook=runbook)
+        self.pxe_runbook: PxeBootSchema = self.runbook
+        self._log = get_logger("pxe_boot", self.__class__.__name__)
+        self._boot_label = "label lisa3 baremetal"
```
squirrelsc: `_boot_label` is used only in one method, and it's not changed. Use a const in this file.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,162 @@
...
+    @classmethod
+    def type_name(cls) -> str:
+        return "pxe_boot"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return PxeBootSchema
+
+    def config(self) -> None:
+        try:
```
squirrelsc: Not eat the exception, let it raise out.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,162 @@
...
+
+        node_name = f"{self.pxe_runbook.host_name}"
+        host_ref = f"host {node_name}"
+        pxe_boot_config_path = self._get_pxe_config_path(node_name, host_ref)
+
+        boot_image = self.pxe_runbook.image_source.replace(
+            self._boot_dir,
+            "",
+        ).strip("/")
+        boot_entry = f"kernel {boot_image}"
```
squirrelsc, line 91: Does the `boot_entry` need to be some information, or just file name? If it contains different information like kernel version, it should be a separated field.

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done. thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done. thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: Done

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: It is used as kernel flag for bootup. just remove it for now.

lisa/base_tools/sed.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: Done

lisa/sut_orchestrator/baremetal/platform_.py
paulli2017: Done

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: Done

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: Done

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: It points to the kernel image path for bootup. Based on the requirement, it is the relative path of image

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,54 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+            )
+
+        ls = build_server.tools[Ls]
+        rc = build_server.tools[RemoteCopy]
+        for file, file_map in _find_matched_files(
+            sources_path,
+            files_map,
+        ).items():
+            if file_map.destination:
+                if ls.is_file(PurePath(file_map.destination)):
```
squirrelsc: The `PurePath` is called 3 times for `PurePath(file_map.destination)`. Please use a variable.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,167 @@
...
+        try:
+            self._connect_to_dhcp_server()
+        except Exception as ex:
+            self._log.exception(
+                f"failed to connect {self.pxe_runbook.connection}",
+                exc_info=ex,
+            )
+
+        node_name = f"{self.pxe_runbook.host_name}"
+        pxe_boot_config_path = self._get_pxe_config_path(node_name)
```
squirrelsc, line 83: It needs to consider the file not exist in a new environment.

lisa/sut_orchestrator/baremetal/bootconfig.py
squirrelsc: If so, you can keep it, but name it like `kernel_boot_params`. It should be a name from Linux community, please use it.

lisa/sut_orchestrator/baremetal/build.py
squirrelsc: `log.exception` doesn't raise an exception. It needs to call like `raise LisaException(...` to rewrite the exception message and raise it. Please check other try/catch in the new code, make sure exception is not ate by unrecoverable reasons.

lisa/sut_orchestrator/baremetal/bootconfig.py
squirrelsc: Check another comment, `log.exception` also eats the exception.

lisa/sut_orchestrator/baremetal/bootconfig.py
squirrelsc: Where is the "requirement" from? Why doesn't it support absolute paths?

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: _get_pxe_config_path() is sure to return a valid file, else an exception is raised in its implementation if empty file is returned, so I removed the file check here. Anyway, I add file check here again

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: I didn't try the absolute path and just follow the existing menu format. 
Here is the boot menu for current setting

label lisa baremetal
    kernel users/paull/fitImage-gp-bm.bin

label lisav2
    kernel users/lis-test/20240819-221158-881/july-soc-gp.bin

label STOS
    kernel "users/v-sherielgin/Image_rsa.img"

label selgin-STOS-test
    kernel "users/v-sherielgin/GP_StableImages/fitImage-initramfs-ev-stable-ovl-2008-GA-gp-stos.bin"

label lisav2
    kernel users/lis-test/20240819-221158-881/july-soc-gp.bin

label STOS
    kernel "users/v-sherielgin/Image_rsa.img"

label selgin-STOS-test
    kernel "users/v-sherielgin/GP_StableImages/fitImage-initramfs-ev-stable-ovl-2008-GA-gp-stos.bin"

squirrelsc: Please don't share internal info here. Absolute path is more flexible and stable. It doesn't need the logic to find relative path.

lisa/sut_orchestrator/baremetal/bootconfig.py
squirrelsc: If the file doesn't exist, it needs to create a new file.

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: Done

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -134,6 +141,20 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
...
 
 
+@dataclass_json()
+@dataclass
+class TftpBuildSchema(BuildSchema):
+    connection: Optional[schema.RemoteNode] = field(
+        default=None, metadata=field_metadata(required=True)
+    )
+    share: str = ""
+    server_name: str = ""
```
squirrelsc: Both share and server_name are unused, should be removed.

lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -134,6 +141,20 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
...
+
+    def __post_init__(self, *args: Any, **kwargs: Any) -> None:
+        if self.connection:
```
squirrelsc: `if self.connection and self.connection.password:`

The auth may be key based.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,49 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+
+    def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
+        assert self.pxe_runbook.connection, "The build server is not specified"
+
+        try:
+            build_server = quick_connect(
+                self.pxe_runbook.connection,
+                logger_name="build_server",
+            )
+        except Exception:
```
squirrelsc: I feel the try/except is not needed, and the new exception hide details.. Just let internal exception raise out.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,49 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+                if ls.is_file(file_map_path):
+                    file_destination = file_map_path.parent
+                else:
+                    file_destination = file_map_path
+            else:
+                file_destination = PurePath(file).parent
+
+            try:
+                rc.copy_to_remote(PurePath(file), file_destination)
+            except Exception:
```
squirrelsc: Similar as above, the try/except is not needed.

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,49 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+                raise LisaException(f"Failed to copy {file} to {file_destination}")
+
+        self._log.debug(f"Build files are copied to build_server: {build_server}")
```
squirrelsc: Rather than put a log at the end, put it to the beginning is better. So when executing long operation, users know which operation is running.

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: I tried to use absolutely path and it failed due to error of missing boot image file, which means the relative path is needed for the boot image file.

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/schema.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/schema.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: Same as above. thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: Added both start and done for debug message.

lisa/sut_orchestrator/baremetal/bootconfig.py
```diff
@@ -0,0 +1,166 @@
...
+            self.pxe_runbook.connection, logger_name="dhcp_server"
+        )
+
+        node_name = f"{self.pxe_runbook.host_name}"
+        pxe_boot_config_path = self._get_pxe_config_path(node_name)
+        if not pxe_boot_config_path:
+            with open(pxe_boot_config_path, "w") as f:
+                f.write("timeout 150\n\nmenu title selections\n\n")
+
+        boot_image = self.pxe_runbook.image_source.replace(
```
squirrelsc: Use Path.relative_to to get relative path.

lisa/sut_orchestrator/baremetal/bootconfig.py
squirrelsc: ok, we have to use relative path.

lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -81,6 +87,7 @@ class ClusterSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
     ready_checker: Optional[ReadyCheckerSchema] = None
     ip_getter: Optional[IpGetterSchema] = None
     key_loader: Optional[KeyLoaderSchema] = None
+    boot_config: Optional[BuildSchema] = None
```
squirrelsc: Should it be BootConfigSchema?

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,43 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+                file_map_path = PurePath(file_map.destination)
+                if ls.is_file(file_map_path):
+                    file_destination = file_map_path.parent
+                else:
+                    file_destination = file_map_path
+            else:
+                file_destination = PurePath(file).parent
+
+            rc.copy_to_remote(PurePath(file), file_destination)
+        self._log.debug(f"Copying files to: {build_server} is done")
```
squirrelsc: `f"Copied files to: {build_server}")`

lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -79,23 +105,43 @@ def copy(self, sources_path: List[Path], files_map: List[FileSchema]) -> None:
...
+        assert self.pxe_runbook.connection, "The build server is not specified"
+
+        build_server = quick_connect(
+            self.pxe_runbook.connection,
+            logger_name="build_server",
+        )
+
+        ls = build_server.tools[Ls]
+        rc = build_server.tools[RemoteCopy]
+        self._log.debug(f"Started to copy files to: {build_server}")
```
squirrelsc: `f"Copying files to: {build_server}"`

lisa/sut_orchestrator/baremetal/bootconfig.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/schema.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/build.py
paulli2017: The change is done, thanks

LiliDeng (approved):

PR #3424, opened 2024-09-18
force_run wget

kamalca:
If wget loads a cached version, you will get an error: 

```
AssertionError: [Error : cat: wget_temp.log: No such file or directory Get unexpected exit code on cmd ['sudo', 'sh', '-c', 'cat wget_temp.log']] Expected <[0]> to contain item <1>, but did not.
```


kamalca:
Will instead use solution in PR #3436

PR #3426, opened 2024-09-18
DpdkTestpmd: Installation path cleanup

mcgov:
The DPDK installer path has grown to be unmaintainable, this PR series cleans it up using some classes instead of just a bunch of if/else trees. The PR looks scary; but, it is just moving code to the object-oriented replacements. This means when we get to `Tool.install()` we can just call `dpdktestpmd.installer.run_installation()` 

My plan is to re-use the SourceInstaller and PackageManagerInstaller logic for rdma-core, though this series only implents the classes for DPDK. 

For the DPDK and rdma-core installation we have to cover multiple cases:
-install from pkg manager
-install from source
   - install from git
   - install from tar
- (implicit) use already installed version

And we cover these cases for multiple targets:
- debian
   - ubuntu 18.04 special casing
- suse
- fedora

and now need to cover all of these scenarios for both DPDK and rdma-core.

Changes:
- Adds a generic class 'Installer' class to dpdk/common.py
- Adds child class `SourceInstall` and `PackageManagerInstall` to dpdk/common.py
- Adds `DpdkSourceInstall` and `DpdkPackageManagerInstall` classes to dpdk/dpdktestpmd.py
- Adds `DpdkGitInstall` and `DpdkTarInstall` implementations of the generic DpdkSourceInstall class to dpdk/dpdktestpmd.py

This required adding some other logic:
- a Meson tool
- Updating the python class to allow --user pip installation
- Ddding the -f flag to the Ln tool.

### DIFF
--- a/lisa/tools/meson.py
+++ b/lisa/tools/meson.py
@@ -1,0 +1,56 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from pathlib import PurePath
+from typing import cast
+
+from semver import VersionInfo
+
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+
+from .ln import Ln
+from .python import Pip
+from .whoami import Whoami
+
+
+class Meson(Tool):
+    @property
+    def command(self) -> str:
+        return "meson"
+
+    def _check_exists(self) -> bool:
+        result = self.node.execute("meson --version", shell=True)
+        return result.exit_code == 0 and VersionInfo.parse(result.stdout) >= "0.52.0"
+
+    @property
+    def can_install(self) -> bool:
+        return self.node.is_posix
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        # use pip to make sure we install a recent version
+        if (not posix_os.package_exists("meson")) or posix_os.get_package_information(
+            "meson", use_cached=False
+        ) < "0.52.0":
+            username = self.node.tools[Whoami].get_username()
+            self.node.tools[Pip].install_packages("meson", install_to_user=True)
+            # environment variables won't expand even when using shell=True :\
+            self.node.tools[Ln].create_link(
+                f"/home/{username}/.local/bin/meson", "/usr/bin/meson", force=True
+            )
+
+        return self._check_exists()
+
+    def setup(self, args: str, cwd: PurePath, build_dir: str = "build") -> PurePath:
+        self.run(
+            f"{args} {build_dir}",
+            force_run=True,
+            shell=True,
+            cwd=cwd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                f"Could not configure {str(cwd)} with meson using args {args}"
+            ),
+        )
+        return cwd.joinpath(build_dir)
--- a/lisa/tools/ninja.py
+++ b/lisa/tools/ninja.py
@@ -1,0 +1,63 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+
+from typing import List, Type
+
+from lisa.base_tools import Mv
+from lisa.executable import Tool
+from lisa.operating_system import Posix
+from lisa.tools.gcc import Gcc
+from lisa.tools.git import Git
+from lisa.tools.python import Python
+
+
+class Ninja(Tool):
+    _ninja_url = "https://github.com/ninja-build/ninja/"
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Gcc, Python]
+
+    @property
+    def command(self) -> str:
+        return "ninja"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        node = self.node
+        # entirely arbitrary kernel version cutoff
+        # to seperate older distros (pre-2020) from newer ones
+        # that package recent versions of ninja.
+        if (
+            isinstance(self.node.os, Posix)
+            and self.node.os.get_kernel_information().version >= "5.15.0"
+        ):
+            self.node.os.install_packages("ninja-build")
+            return self._check_exists()
+        # otherwise, install a recent version from source
+        git_tool = node.tools[Git]
+        node.tools[Gcc].install_cpp_compiler()
+        ninja_path = git_tool.clone(
+            self._ninja_url,
+            cwd=node.working_path,
+        )
+        node.execute(
+            "./configure.py --bootstrap",
+            cwd=node.get_pure_path(f"{str(ninja_path)}"),
+            expected_exit_code=0,
+            expected_exit_code_failure_message=(
+                "Ninja: install failed to run ./configure.py --bootstrap"
+            ),
+        )
+        node.tools[Mv].move(
+            f"{ninja_path}/ninja",
+            "/usr/bin/ninja",
+            overwrite=True,
+            sudo=True,
+        )
+
+        return self._check_exists()
--- a/lisa/tools/python.py
+++ b/lisa/tools/python.py
@@ -33,6 +33,15 @@ class Python(Tool):
 
         return self._check_exists()
 
+    def get_python_path(self, sudo: bool = False) -> str:
+        return self.run(
+            '-c "import sys; print(' + "':'" + '.join([x for x in sys.path if x]))"',
+            expected_exit_code=0,
+            expected_exit_code_failure_message=("Could not fetch python sys.path!"),
+            shell=True,
+            sudo=sudo,
+        ).stdout
+
 
 class Pip(Tool):
     _no_permission_pattern = re.compile(r"Permission denied", re.M)
@@ -60,9 +69,14 @@ class Pip(Tool):
         self.node.os.install_packages(package_name)
         return self._check_exists()
 
-    def install_packages(self, packages_name: str, install_path: str = "") -> None:
+    def install_packages(
+        self, packages_name: str, install_path: str = "", install_to_user: bool = False
+    ) -> None:
         node = self.node
-        cmd_line = f"install -q {packages_name}"
+        if not install_to_user:
+            cmd_line = f"install -q {packages_name}"
+        else:
+            cmd_line = f"install --user -q {packages_name}"
 
         envs = {}
         if install_path != "":
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -54,8 +54,10 @@ class DependencyInstaller:
         self, node: Node, extra_args: Union[List[str], None]
     ) -> None:
         os = node.os
-        if not isinstance(os, Posix):
-            return
+        assert isinstance(os, Posix), (
+            "DependencyInstaller is not compatible with this OS: "
+            f"{os.information.vendor} {os.information.release}"
+        )
         # find the match for an OS, install the packages.
         # stop on list end or if exclusive_match parameter is true.
         for requirement in self.requirements:
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -41,7 +41,7 @@ class OsPackageDependencies:
     ) -> None:
         self.matcher = matcher
         self.packages = packages
-        self.exclusive_match = stop_on_match
+        self.stop_on_match = stop_on_match
 
 
 class DependencyInstaller:
@@ -61,7 +61,7 @@ class DependencyInstaller:
         for requirement in self.requirements:
             if requirement.matcher(os) and requirement.packages:
                 os.install_packages(requirement.packages, extra_args=extra_args)
-                if requirement.exclusive_match:
+                if requirement.stop_on_match:
                     return
         # NOTE: It is up to the caller to raise an exception on an invalid OS
 
@@ -129,7 +129,7 @@ class TarDownloader(Downloader):
                 "Source path is not a .tar[.gz|.bz2] file. "
                 f"Tar url was set to: {self._tar_url} "
             )
-        ).is_true
+        ).is_true()
         if self._is_remote_tarball:
             tarfile = node.tools[Wget].get(
                 self._tar_url,
@@ -184,7 +184,7 @@ class Installer:
             self._node.log.debug("No downloader assigned to installer.")
 
     # do the build and installation
-    def _run_build(self) -> None:
+    def _install(self) -> None:
         self._download_assets()
 
     # remove an installation
@@ -213,7 +213,7 @@ class Installer:
         if self._should_install():
             self._uninstall()
             self._install_dependencies()
-            self._run_build()
+            self._install()
 
     def __init__(
         self,
@@ -261,7 +261,7 @@ class PackageManagerInstall(Installer):
         return True
 
     # installing dependencies is the installation in this case, so just return
-    def _run_build(self) -> None:
+    def _install(self) -> None:
         return
 
 
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -93,13 +93,13 @@ class GitDownloader(Downloader):
         # NOTE: fail on exists is set to True.
         # The expectation is that the parent Installer class should
         # remove any lingering installations
-        self._source_path = self._node.tools[Git].clone(
+        self._asset_path = self._node.tools[Git].clone(
             self._git_repo,
             cwd=self._node.get_working_path(),
             ref=self._git_ref,
             fail_on_exists=True,
         )
-        return self._source_path
+        return self._asset_path
 
 
 # parent class for tarball source installations
@@ -147,7 +147,7 @@ class TarDownloader(Downloader):
                 node_path=remote_path,
             )
         # create tarfile dest dir
-        self._source_path = work_path.joinpath(
+        self._asset_path = work_path.joinpath(
             self.tar_filename[: -(len(tarfile_suffix))]
         )
         # unpack into the dest dir
@@ -157,7 +157,7 @@ class TarDownloader(Downloader):
             dest_dir=str(work_path),
             gzip=True,
         )
-        return self._source_path
+        return self._asset_path
 
 
 class Installer:
@@ -177,15 +177,15 @@ class Installer:
         raise NotImplementedError(f"_check_if_installed {self._err_msg}")
 
     # setup the installation (install Ninja, Meson, etc)
-    def _setup_installation(self) -> None:
+    def _download_assets(self) -> None:
         if self._downloader:
             self._asset_path = self._downloader.download()
         else:
             self._node.log.debug("No downloader assigned to installer.")
 
     # do the build and installation
-    def _run_installation(self) -> None:
-        raise NotImplementedError(f"_run_installation {self._err_msg}")
+    def _run_build(self) -> None:
+        self._download_assets()
 
     # remove an installation
     def _clean_previous_installation(self) -> None:
@@ -213,8 +213,7 @@ class Installer:
         if self._should_install():
             self._clean_previous_installation()
             self._install_dependencies()
-            self._setup_installation()
-            self._run_installation()
+            self._run_build()
 
     def __init__(
         self,
@@ -230,7 +229,6 @@ class Installer:
         self._os: Posix = self._node.os
         self._package_manager_extra_args: List[str] = []
         self._os_dependencies = os_dependencies
-        self._source_path = self._node.get_working_path()
         self._downloader = downloader
 
 
@@ -263,7 +261,7 @@ class PackageManagerInstall(Installer):
         return True
 
     # installing dependencies is the installation in this case, so just return
-    def _run_installation(self) -> None:
+    def _run_build(self) -> None:
         return
 
 
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -37,11 +37,11 @@ class OsPackageDependencies:
         self,
         matcher: Callable[[Posix], bool],
         packages: Optional[Sequence[Union[str, Tool, Type[Tool]]]] = None,
-        exclusive_match: bool = False,
+        stop_on_match: bool = False,
     ) -> None:
         self.matcher = matcher
         self.packages = packages
-        self.exclusive_match = exclusive_match
+        self.exclusive_match = stop_on_match
 
 
 class DependencyInstaller:
@@ -188,7 +188,7 @@ class Installer:
         self._download_assets()
 
     # remove an installation
-    def _clean_previous_installation(self) -> None:
+    def _uninstall(self) -> None:
         raise NotImplementedError(f"_clean_previous_installation {self._err_msg}")
 
     # install the dependencies
@@ -211,7 +211,7 @@ class Installer:
     def do_installation(self, required_version: Optional[VersionInfo] = None) -> None:
         self._setup_node()
         if self._should_install():
-            self._clean_previous_installation()
+            self._uninstall()
             self._install_dependencies()
             self._run_build()
 
@@ -238,7 +238,7 @@ class PackageManagerInstall(Installer):
         super().__init__(node, os_dependencies)
 
     # uninstall from the package manager
-    def _clean_previous_installation(self) -> None:
+    def _uninstall(self) -> None:
         if not (isinstance(self._os, Posix) and self._check_if_installed()):
             return
         if self._os_dependencies is not None:
@@ -373,3 +373,10 @@ def is_url_for_git_repo(url: str) -> bool:
         or any([x in path_check.parts for x in ["git", "_git"]])
     )
     return scheme == "git" or check_for_git_https
+
+
+def unsupported_os_thrower(os: Posix) -> bool:
+    raise UnsupportedDistroException(
+        os,
+        message=("Installer did not define dependencies for this os."),
+    )
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -2,12 +2,17 @@
 # Licensed under the MIT license.
 
 from datetime import datetime
-from typing import Any, Dict
+from pathlib import PurePath
+from typing import Any, Callable, Dict, List, Optional, Sequence, Type, Union
 
 from assertpy import assert_that
+from semver import VersionInfo
+from urllib3.util.url import parse_url
 
 from lisa import Node
-from lisa.operating_system import Debian, Oracle, Redhat, Suse, Ubuntu
+from lisa.executable import Tool
+from lisa.operating_system import Debian, Fedora, Oracle, Posix, Redhat, Suse, Ubuntu
+from lisa.tools import Git, Tar, Wget
 from lisa.util import UnsupportedDistroException
 
 DPDK_STABLE_GIT_REPO = "https://dpdk.org/git/dpdk-stable"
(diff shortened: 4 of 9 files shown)


lisa/tools/python.py
```diff
@@ -60,9 +60,14 @@ def _install(self) -> bool:
         self.node.os.install_packages(package_name)
         return self._check_exists()
 
-    def install_packages(self, packages_name: str, install_path: str = "") -> None:
+    def install_packages(
+        self, packages_name: str, install_path: str = "", user: bool = False
```
squirrelsc: `user` to `install_to_user`

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -570,199 +873,31 @@ def _install_upstream_rdma_core_for_mana(self) -> None:
...
-        # otherwise, install from package manager, git, or tar
-
-        self._install_dependencies()
-
-        # if this is mana VM, we need an upstream rdma-core package (for now)
+        # ensure no older dependency is installed
+        if not isinstance(node.os, (Debian, Fedora, Suse)):
+            return False
+        if isinstance(node.os, Ubuntu) and node.os.information.codename == "bionic":
+            node.os._update_packages()
```
squirrelsc: Don't call private methods out of the class.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,266 @@
...
+
+    def _run_installation(self) -> None:
+        # define to handle any special special cases
+        self._setup_node()
+        # and install everything
+        self._install_dependencies()
+
+
+# base class for source installs
+class SourceInstall(Installer):
```
squirrelsc: The sourceinstall class has no special implementation for source installation, it looks not needed.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,266 @@
...
+            for os_package_check in self._os_dependencies.requirements:
+                if os_package_check.check_node_os(self._os):
+                    for pkg in os_package_check.packages:
+                        if not self._os.package_exists(pkg):
+                            return False
+        return True
+
+    def _run_installation(self) -> None:
+        # define to handle any special special cases
+        self._setup_node()
```
squirrelsc: It's called in `do_installation`, why call it again in `_run_installation`, which is called by `do_installation` too?

lisa/tools/meson.py
```diff
@@ -0,0 +1,57 @@
...
+from .whoami import Whoami
+
+
+class Meson(Tool):
+    @property
+    def command(self) -> str:
+        return "meson"
+
+    @property
+    def exists(self) -> bool:
```
squirrelsc: rewrite _check_exists is better to reuse logic.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,266 @@
...
+    def __init__(
+        self,
+        os_matcher: Callable[[Posix], bool],
+        packages: Sequence[Union[str, Tool, Type[Tool]]],
+    ) -> None:
+        self.check_node_os = os_matcher
+        self.packages = packages
+
+
+class InstallDependencies:
```
squirrelsc: define it as a function.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,266 @@
...
+                    for pkg in os_package_check.packages:
+                        if not self._os.package_exists(pkg):
+                            return False
+        return True
+
+    def _run_installation(self) -> None:
+        # define to handle any special special cases
+        self._setup_node()
+        # and install everything
+        self._install_dependencies()
```
squirrelsc: It's not easy to understand, why dependencies to be installed after `_run_installation`. What depends on those dependencies?

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -2,47 +2,310 @@
...
+        git_repo: str,
+        git_ref: str,
+    ) -> None:
+        super().__init__(node, os_dependencies, git_repo, git_ref)
+
+    def _configure_installation(self) -> None:
+        if not self.git_ref:
+            git = self._node.tools[Git]
+            self.git_ref = git.get_tag(
+                self._source_path, filter_=r"^v.*"  # starts w 'v'
```
squirrelsc: It's the only deifference with the `DpdkSourceInstall`. The name of classes are confusing.

microsoft/testsuites/dpdk/common.py
mcgov: To be honest, I wrote this one first! I'll fix it up since it probably doesn't even need to be defined now.

microsoft/testsuites/dpdk/common.py
mcgov: I don't see an advantage of defining this as a function, I'll rename it to DependencyInstaller to make the purpose more clear.

microsoft/testsuites/dpdk/common.py
mcgov: Fixing this class up, I re-wrote the generic code at one point but didn't go back and fix this class to use the generic _run_installation. Will fix

microsoft/testsuites/dpdk/common.py
mcgov: I'm going to leave it, even though it doesn't do much.
- It sets _checkout_path
- It serves as a common class for checking isinstance(installer, SourceInstall)

I'm going to re-use this code for rdma-core as well, so having common parent classes is going to be useful.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: We don't need to do this git ref filtering for other projects, it's project specific.
>It's the only deifference with the DpdkSourceInstall. 

Exactly. I only need to overload one or two functions per class, that's expected. There is a lot of shared code between the builds.

microsoft/testsuites/dpdk/common.py
squirrelsc: If a class host only one function, and no data used, it should be a function. Some languages like C#, Jave are purely object oriented, so they cannot use function. But Python uses mixed functions and oo. Avoid overusing classes, so the remaining classes are easier to be understood.

microsoft/testsuites/dpdk/common.py
mcgov: fixed, thanks for taking a look!

microsoft/testsuites/dpdk/common.py
mcgov: I appreciate all of that, I agree with all of what you've said. But:
The os match -> package list _is_ data. I think it's cleaner to encapsulate these dependencies so I can pass an instance of this class to the constructor for the Dpdk___Installer later. This gives the reader all the info they need to know about what's happening.

I can make a couple of DependencyInstaller instances for source and package manager installs. In our __init__ path I just call `self._installer=DpdkGitInstaller(dependencies=dpdk_source_dependencies)`
or  `self._installer=DpdkTarInstaller(dependencies=dpdk_source_dependencies)`
or  `self._installer=DpdkPackageManagerInstaller(dependencies=dpdk_package_manager_dependencies)`
Afterwards, in the install function I just call:

`self._installer.run_installation()`

If you look at the code we're replacing, this is a much more organized way to handle things. If we need to replace this later, I will go back and fix it. For now, this is how I want to handle it.

mcgov: Also remember: this code is going to be used for both dpdk and rdma-core eventually. Keeping that organized and readable is going to be important, I'm doing this intentionally to make that future PR easier to read 😂

I am not exaggerating, I wrote the DPDK 32bit test and looked at the PR and said to myself "Chi is never going to approve this"

So I went back and starting fixing the installation path 😜

mcgov:
Making one more change to make Ninja a tool since it's used by two builds and annoying to install on some systems.

Edit: Change complete! Also fixed a bug when calling `sudo ninja install` on some systems by enforcing a consistent pythonpath

mcgov:
Fixed a rebase issue, should be good to go

microsoft/testsuites/dpdk/common.py
squirrelsc: 1. It's working path, not checkout path. The "checkout" is for GitInstall, but it's a "source" path for TarInstall. So it still presents a "working" path. Also the logic is simple, just save the node working path. It doesn't need to have a base class for one line assignment.
2. You can think about design ahead, but not do too much. When you implement rdma, you can refactor at that moment.

mcgov: Ok, I'll spend the time to remove this code and then add it again later when it's needed.

squirrelsc:
I read more on the code and the problem try to solve. It looks a valid scenario to a couple of tools, like some open source. There are some thoughts and draft design.

### Thoughts

* Two things here in different stages. If we split them into the two stages, the inherits chain is clear.
  * Download. The download ways include by wget, git.
  * Install. The install ways include copy, package, and build.
* Reinstall. There may be different reason to reinstall the tool. For example, upgrade version, or build new temp version (xdp). It needs to define reinstall strategy in each tool.
* Dependencies. Different distro has different dependent packages, or even extra commands to install correctly.
* The composition of Downloader and Installer are implemented in the corresponding tool level. It can be detected by passed in parameters, os versions, etc.
* It doesn't need to define methods to implement by subclass and subclass of subclass separately. The subclasses can override the same method in superclass, and call the method of superclass in the overridden version. It provides enough flexibility for most cases.
* Don't inherits from two different superclasses. It's easy to cause issue. The workaround is either composite them, or use Mixin pattern to minimize conflicts.

### Design

```python

class Downloader:
  # base class of download. Due to different parameters, so use a class instead of a method.
  def __init__(self, **kwargs: Any) -> None:
    ...
  def download(self) -> path.PurePath:
    # download code including with wget or git
    raise NotImplementError()

class GitDownloader(Downloader):
  def __init__(self, repo: str, ref: str, **kwargs: Any) -> None:
    ...
  def download(self) -> path.PurePath:
    # imp git clone and checkout, if needed
    ...
class FileDownloader(Downloader):
  def __init__(self, url: str, **kwargs: Any) -> None:
    ...
  def download(self) -> path.PurePath:
    # imp wget download and return root path.
    ...
class TarFileDownloader(FileDownloader):
  def download(self) -> path.PurePath:
    path = super().download()
    # extra tar file and return the path
    ...


class Installer:
  def __init__(self, downloader: Downloader, **kwargs: Any) -> None:
    # downloader is initialized in corresponding tools by os version, passed in parameters etc.
    # kwargs can be passed through to subclasses
    ...
  def install(self) -> None:
    # implemented here to include reinstall for some reasons, like version mismatches.
    # below is the skeleton code, more error handling may be added.
    if self.need_install():
        self.downloader.download()
        self.uninstall()
        self._install()
  def uninstall(self) -> None:
    raise NotImplementedError()
  def need_install(self) -> bool:
    raise NotImplementedError()
  def _install(self) -> None:
    raise NotImplementedError()

class BinaryInstaller(Installer):
  # install by cp
   ...
class PackageInstaller(Installer):
  # install by packages
   ...
class SourceInstaller(Installer):
  # install by building binary.
   ...
```

mcgov:
> There are some thoughts and draft design.

Thanks @squirrelsc. I like the mix-in idea! I was trying to implement something like that but didn't know the name of it.

I'll implement that in the next PR.

microsoft/testsuites/dpdk/common.py
mcgov: ok fixed

lisa/tools/meson.py
```diff
@@ -0,0 +1,56 @@
...
+    @property
+    def can_install(self) -> bool:
+        return self.node.is_posix
+
+    def _install(self) -> bool:
+        posix_os: Posix = cast(Posix, self.node.os)
+        # use pip to make sure we install a recent version
+        if (not posix_os.package_exists("meson")) or posix_os.get_package_information(
+            "meson", use_cached=False
+        ) < "0.52.0":
```
LiliDeng, line 35: will the comparison work?

lisa/tools/ninja.py
```diff
@@ -0,0 +1,59 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        node = self.node
+        if (
+            isinstance(self.node.os, Posix)
+            and self.node.os.get_kernel_information().version >= "5.15.0"
```
LiliDeng, line 33: ditto as above comments

LiliDeng, line 37: could you please add comments for the kernel version?

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -2,47 +2,316 @@
...
+    def get_installed_version(self) -> VersionInfo:
+        return self._os.get_package_information("dpdk", use_cached=False)
+
+    def _check_if_installed(self) -> bool:
+        return self._os.package_exists("dpdk")
+
+
+# implement SourceInstall for DPDK
+class DpdkSourceInstall(Installer):
+    _dpdk_devname_repo = "https://www.github.com/mcgov/devname.git"
```
LiliDeng: please add comments for the reason using personal repo

squirrelsc: we need to avoid personal repos. You can add the c file in LISA.

mcgov: I'm not maintaining two copies of the project, I'll update the link when it's checked into the DPDK project.

squirrelsc: What means "link"? It looks a core c file, please either include the c file in LISA repo, or create a Microsoft repo (not under personal name). I prefer to maintain the file in LISA. You can still use it in other places. Because the extra repo will be hold by our team too.

mcgov: `_dpdk_devname_repo = "https://www.github.com/mcgov/devname.git" ` <- the link

Sure. For now, I'll just remove it since it's not in use yet and I'm going to change how it's implemented when we start to use it.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+
+class Downloader:
+    def __init__(self, node: Node) -> None:
+        self._node = node
+
+    def download(self) -> PurePath:
+        raise NotImplementedError("Downloader not implemented.")
+
+
+class GitDownload(Downloader):
```
squirrelsc: `GitDownloader` so it's consistent with the base class.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+# This should help us cover the various installation cases in a nice way,
+# and allow us to only re-implement the bits we need for each project.
+class OsPackageDependencies:
+    # A class to reduce the isinstance() trees that are
+    # sprinkled everywhere.
+    # Caller provides a function to match an OS and
+    # the packages to install on that OS.
+    def __init__(
+        self,
+        os_matcher: Callable[[Posix], bool],
```
squirrelsc: `matcher` is better, because it may check package version, not only os version.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+        packages: Sequence[Union[str, Tool, Type[Tool]]],
+    ) -> None:
+        self.check_node_os = os_matcher
```
squirrelsc: use a consistent name `matcher`.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+    # provide a list of OsPackageDependencies for a project
+    def __init__(self, requirements: List[OsPackageDependencies]) -> None:
+        self.requirements = requirements
+
+    # evaluate the list of package dependencies,
+    def install_required_packages(
+        self, node: Node, extra_args: Union[List[str], None]
+    ) -> None:
+        os = node.os
+        if not isinstance(os, Posix):
```
squirrelsc: This is not needed.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+            return
+        # find the first match for an OS, install the packages.
+        # Most specific should go first, most generic should go last.
+        for os_package_check in self.requirements:
```
squirrelsc: call the var `requirement`, to avoid unnecessary different concepts.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+            if os_package_check.check_node_os(os):
+                os.install_packages(os_package_check.packages, extra_args=extra_args)
+                return
```
squirrelsc: It doesn't need to return, just check all of them. There maybe some reason to split packages in different ways. If you want to speed up, add one more field like `stop_match`. If it's True, then return.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+        raise UnsupportedDistroException(
```
squirrelsc: It doesn't need to raise Exception. Add a `match` and raise Exception there.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+            os,
+            message=("Installer did not define dependencies for this os."),
+        )
+
+
+class Downloader:
+    def __init__(self, node: Node) -> None:
```
squirrelsc, line 72: define like `def __init__(self, node: Node, **kwargs: Any) -> None:`. So, it's able to dynamic initialize for subclasses.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+    git_ref: str = ""
+
+    def __init__(
+        self,
+        node: Node,
+        git_repo: str,
+        git_ref: str,
+    ) -> None:
+        super().__init__(node)
+        self.git_repo = git_repo
```
squirrelsc: Define like `self._git_repo = git_repo`. All variables should be private by default, unless it's really needed from other classes.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+        # NOTE: fail on exists is set to True.
+        # The expectation is that the parent Installer class should
+        # remove any lingering installations
+        self._source_path = self._node.tools[Git].clone(
+            self.git_repo,
+            cwd=self._node.get_working_path(),
+            ref=self.git_ref,
+            fail_on_exists=True,
+        )
+        self._branch_handler()
```
squirrelsc: It doesn't need to define another method. When it needs extra operation, just overwrites the download method in the subclass like below. It gives more flexibility. 
```python
def download(self) -> PurePath:
    # other operations before download.
    super().download()
    # other operations after download.
```

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+        for suffix in [".tar.gz", ".tar.bz2", ".tar"]:
+            if self.tar_filename.endswith(suffix):
+                unpacked_filename = self.tar_filename[: -len(suffix)]
+                self._source_path = work_path.joinpath(unpacked_filename)
+                break
+        assert_that(str(self._source_path)).described_as(
+            "Source path was not set by tar installer, "
+            "likely an unexpected file extension for "
+            f"the tar file: {self.tar_filename}"
+        ).is_not_equal_to(str(work_path))
```
squirrelsc: The logic is not easy to understand. It should assert there is an tar file, and unpack it. It shouldn't do above steps silently, and then check if the path have tar or not.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+
+    # run the defined setup and installation steps.
+    def do_installation(self, required_version: Optional[VersionInfo] = None) -> None:
+        self._setup_node()
+        if self._check_if_installed() and required_version is not None:
+            if self.get_installed_version() == required_version:
+                return
+            else:
+                self._clean_previous_installation()
+        self._install_dependencies()
```
squirrelsc: Don't split it to multiple methods, try to overwrite methods of base classes.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,261 @@
...
+            )
+
+    # define how to check the installed version
+    def get_installed_version(self) -> VersionInfo:
+        raise NotImplementedError(f"get_installed_version {self._err_msg}")
+
+    # run the defined setup and installation steps.
+    def do_installation(self, required_version: Optional[VersionInfo] = None) -> None:
+        self._setup_node()
+        if self._check_if_installed() and required_version is not None:
```
squirrelsc: Please check my design to support more flexible logic on check if install or not.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -470,14 +729,47 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
             self._dpdk_repo_path_name = "dpdk"
+            self._expected_install_path = "/usr/local/bin"
+
             work_path = self.node.get_working_path_with_required_space(5)
             self.current_work_path = self.node.get_pure_path(work_path)
             self.dpdk_path = self.node.get_pure_path(work_path).joinpath(
                 self._dpdk_repo_path_name
             )
-        self._determine_network_hardware()
+            if is_url_for_git_repo(self._dpdk_source):
```
squirrelsc, line 771: Below is clearer.

```python
            if is_url_for_git_repo(self._dpdk_source):
                downloader=DpdkGitDownload(
                        node=self.node,
                        git_repo=self._dpdk_source,
                        git_ref=self._dpdk_branch,
                    )
            elif is_url_for_tarball(self._dpdk_source):
                downloader=TarDownload(node=self.node, tar_url=self._dpdk_source)
            else:
                fail(
                    "URL provided for dpdk source did not validate as "
                    f"a tarball or git repo. Found {self._dpdk_source} "
                    " Expected https://___/___.git or /path/to/tar.tar[.gz] or "
                    "https://__/__.tar[.gz]"
                )
            self.installer = DpdkSourceInstall(
                node=self.node,
                os_dependencies=DPDK_SOURCE_INSTALL_PACKAGES,
                downloader=downloader,
            )
```

microsoft/testsuites/dpdk/common.py
mcgov: I'll fix.

microsoft/testsuites/dpdk/common.py
mcgov: I'm fine with fixing it later to suit your preference but I'm not doing that for this commit. The Installer is not used outside of DpdkTestpmd, so it's fine to merge as-is now. I'm going to touch this again when I implement the rdma-core Installer, so I'll change it in that PR.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Sure, fixed.

microsoft/testsuites/dpdk/common.py
mcgov: Ok

microsoft/testsuites/dpdk/common.py
mcgov: I'm not making more changes for a use case that isn't in-use for the PR. I have dependencies where I want to match for ubuntu 18.04 only, any other ubuntu, and then any other debian distro. It's cleaner to have these structured this way. If we ever need it the other way, we'll make that change later.

microsoft/testsuites/dpdk/common.py
mcgov: It is for the type checker.

microsoft/testsuites/dpdk/common.py
mcgov: fixed

mcgov:
edit: testing looks good

lisa/tools/ninja.py
```diff
@@ -0,0 +1,63 @@
...
+        node = self.node
+        # entirely arbitrary kernel version cutoff
+        # to seperate older distros (pre-2020) from newer ones
+        # that package recent versions of ninja.
+        if (
+            isinstance(self.node.os, Posix)
+            and self.node.os.get_kernel_information().version >= "5.15.0"
+        ):
+            self.node.os.install_packages("ninja-build")
+            return True
```
squirrelsc: `return self._check_exists()`

microsoft/testsuites/dpdk/common.py
squirrelsc: use assert, make it fail on unexpected os.

microsoft/testsuites/dpdk/common.py
squirrelsc: We cannot change it later, once it's used by others. Because it's a breaking change on code behaviors.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -564,7 +564,11 @@ def _parallel_cleanup(node: Node) -> None:
         if not interface.is_enabled_sriov():
             interface.switch_sriov(enable=True, wait=False, reset_connections=True)
             # cleanup temporary hugepage and driver changes
-        node.reboot()
+        try:
+            node.reboot(time_out=60)
+        except LisaException:
+            node.log.debug("Timeout during cleanup reboot. Marking node for deletion.")
+            node.mark_dirty()
```
squirrelsc, line 571: Should the test case fail, if reboot failed? The exception is caught, so the test case won't be failed.

mcgov: I care about the DPDK test results. If the test passed and there's a timeout in the cleanup path I think it's better to just mark the node dirty and move on. If there was a crash won't LISA still post the serial console log to the result?

lisa/tools/ninja.py
mcgov: fixed

microsoft/testsuites/dpdk/common.py
mcgov: Ah, neat idea. Fixed

microsoft/testsuites/dpdk/common.py
mcgov: I mean more that if someone else is going to use this and they have a _requirement_ where they need the class to check every OS and not return, we can change it then. I don't see the need to change this _in anticipation_ of someone else using it.

squirrelsc: The dependency pattern is used in a couple of tools. Yes, we need to have a method to go through a list of requirements. The method is shared too.

mcgov: Ok, added a parameter to stop processing early, by default it keeps processing now

microsoft/testsuites/dpdk/common.py
mcgov: Ah ok, annotating elsewhere fixes the type check nag.

microsoft/testsuites/dpdk/common.py
mcgov: I think we talked offline about this one, we'll stick with the strict arg list for now

mcgov:
@squirrelsc could I get one more look before Lili signs on 🥺🙏

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: This change is the result of multiple instances of me going to check a failed test result and seeing the test was fine, LISA just wasn't able to reconnect after rebooting during cleanup for whatetever reason. I just don't think a teardown thing should block the test result, it should result in the node being garbage collected before starting another test though.

microsoft/testsuites/dpdk/common.py
mcgov: fixed

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Resolving since this was removed.

microsoft/testsuites/dpdk/common.py
mcgov: follow up: already using this in a new PR 😂

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -20,30 +19,305 @@
...
+class DpdkSourceInstall(Installer):
+    _sample_applications = [
+        "l3fwd",
+        "multi_process/client_server_mp/mp_server",
+        "multi_process/client_server_mp/mp_client",
+    ]
+
+    def _setup_installation(self) -> None:
+        super()._setup_installation()
+        self._source_path = self._asset_path
```
squirrelsc: If the `_source_path` is the same as `_asset_path`, does it need an extra `_source_path` variable?

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,260 @@
...
+            self._asset_path = self._downloader.download()
+        else:
+            self._node.log.debug("No downloader assigned to installer.")
+
+    # do the build and installation
+    def _run_installation(self) -> None:
+        raise NotImplementedError(f"_run_installation {self._err_msg}")
+
+    # remove an installation
+    def _clean_previous_installation(self) -> None:
```
squirrelsc: `uninstall` is clear, and can be used for uninstall after test also.

microsoft/testsuites/dpdk/common.py
squirrelsc: It's not changed to assert yet.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,260 @@
...
+
+    # evaluate the list of package dependencies,
+    def install_required_packages(
+        self, node: Node, extra_args: Union[List[str], None]
+    ) -> None:
+        os = node.os
+        if not isinstance(os, Posix):
+            return
+        # find the match for an OS, install the packages.
+        # stop on list end or if exclusive_match parameter is true.
```
squirrelsc, line 62: as the comments, call the variable `stop_proceed` or similar. The `exclusive_match` sounds not ordered, but actually those rules are ordered, and handle one by one.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,260 @@
...
+        for requirement in self.requirements:
+            if requirement.matcher(os) and requirement.packages:
+                os.install_packages(requirement.packages, extra_args=extra_args)
+                if requirement.exclusive_match:
+                    return
+        # NOTE: It is up to the caller to raise an exception on an invalid OS
```
squirrelsc, line 68: You can move the definition of `invalid_os_thrower` in this file, and call it like `unsupported_os_thrower`.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: only if you want the variable name to make sense. I'll change

microsoft/testsuites/dpdk/common.py
squirrelsc: @mcgov , please change it to assert.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,249 @@
...
+            required_version is None and required_version > self.get_installed_version()
+        )
+
+    # run the defined setup and installation steps.
+    def do_installation(self, required_version: Optional[VersionInfo] = None) -> None:
+        self._setup_node()
+        if self._should_install():
+            self._uninstall()
+            self._install_dependencies()
+            self._run_build()
```
squirrelsc: How about call it `_install`? so it's consistent with `_uninstall`. The `do_installation` can be called like `install`. The public/private methods is a pattern in LISA, like `initialize`/`_initialize`. The public method is used to be called externally, and the private method is implemented by subclasses.

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,249 @@
...
+    # the packages to install on that OS.
+    def __init__(
+        self,
+        matcher: Callable[[Posix], bool],
+        packages: Optional[Sequence[Union[str, Tool, Type[Tool]]]] = None,
+        stop_on_match: bool = False,
+    ) -> None:
+        self.matcher = matcher
+        self.packages = packages
+        self.exclusive_match = stop_on_match
```
squirrelsc: Use the same name `stop_on_match`.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -20,30 +19,292 @@
...
+DPDK_PACKAGE_MANAGER_PACKAGES = DependencyInstaller(
+    requirements=[
+        OsPackageDependencies(
+            matcher=lambda x: isinstance(x, Debian),
+            packages=["dpdk", "dpdk-dev"],
+            stop_on_match=True,
+        ),
+        OsPackageDependencies(
+            matcher=lambda x: isinstance(x, Suse)
+            and float(x.information.release) == 15.5,
```
squirrelsc, line 63: how about `>=`?

mcgov: I didn't add this check, so I'm not touching it.

microsoft/testsuites/dpdk/common.py
mcgov: Ok

microsoft/testsuites/dpdk/common.py
mcgov: fixed

mcgov:
@LiliDeng can you give a final review once Chi resolves his comments? Please feel free to merge once you approve.

microsoft/testsuites/dpdk/common.py
squirrelsc: @mcgov, please address this one...

microsoft/testsuites/dpdk/common.py
```diff
@@ -17,6 +22,249 @@
...
+            if self._tar_url.endswith(suffix):
+                is_tarball = True
+                tarfile_suffix = suffix
+                break
+        assert_that(is_tarball).described_as(
+            (
+                "Source path is not a .tar[.gz|.bz2] file. "
+                f"Tar url was set to: {self._tar_url} "
+            )
+        ).is_true
```
squirrelsc: It should be `is_true()`. `is_true` is the function, but not called. It won't raise exception, if the condition doesn't meet.

mcgov: Surprised that passed the linter checks! Thanks.

mcgov: fixed, resolving

microsoft/testsuites/dpdk/common.py
mcgov: fixed

microsoft/testsuites/dpdk/common.py
```diff
@@ -54,8 +54,7 @@ def install_required_packages(
         self, node: Node, extra_args: Union[List[str], None]
     ) -> None:
         os = node.os
-        if not isinstance(os, Posix):
-            return
+        assert isinstance(os, Posix)
```
squirrelsc: Add a failure message, so it can show in the test results error message.

mcgov: added

mcgov:
@squirrelsc 
last call on comments, anything else?

microsoft/testsuites/dpdk/common.py
squirrelsc: Please add the actual type like other type assertion, so the error message contains what actual OS it is.

squirrelsc:
> @squirrelsc last call on comments, anything else?

No others except one more polish. Please wait Lili for merging.

@LiliDeng LGTM.

microsoft/testsuites/dpdk/common.py
mcgov: added

mcgov:
@squirrelsc Lili is OOF for the week.

squirrelsc:
> @squirrelsc Lili is OOF for the week.

Please share test results on major versions of Ubuntu, SUSE, RedHat, I can merge code, if no regression.

squirrelsc (approved):

squirrelsc:
Checked test results, no failure happens on major distro with 3 typical dpdk test cases.

PR #3432, opened 2024-09-20
NVMe and resource disc testcase fixes

SRIKKANTH:
NVMe and resource disc testcase fixes

1. Added new testcase "Nvme.verify_nvme_function_unpartitioned".
2. NVMe test case fixes for disc controller type NVMe.
3. "Nvme.verify_nvme_fstrim" - use fstrim testfile size based on available space.
4. Resource disc testcases fixes for VMs with Local NVMEs.
5. Added below new methods:
    a. check_resource_disk_mounted
    b. get_resource_disks
    c. get_resource_disk_type
    d. get_disk_type
    e. _remove_nvme_os_disk
    f. _get_os_disk_nvme_device
    g. _verify_nvme_function


SRIKKANTH:
This PR has a dependency on https://github.com/microsoft/lisa/pull/3420 

Please merge https://github.com/microsoft/lisa/pull/3420 before accepting this PR so that lspci changes are part of different PR. and not clubbed with NVMe changes

lisa/features/nvme.py
```diff
@@ -78,7 +93,17 @@ def get_namespaces(self) -> List[str]:
         return namespaces
 
     def get_namespaces_from_cli(self) -> List[str]:
-        return self._node.tools[Nvmecli].get_namespaces()
+        namespaces_list = self._node.tools[Nvmecli].get_namespaces()
+        node_disk = self._node.features[Disk]
```
squirrelsc: The logic is repeated multiple times, it should be a new method like `_remove_nvme_os_disk`.

microsoft/testsuites/core/storage.py
```diff
@@ -229,27 +240,40 @@ def verify_swap(self, node: RemoteNode) -> None:
...
-        # verify that resource disk is mounted
-        # function returns successfully if disk matching mount point is present
-        node.features[Disk].get_partition_with_mount_point(resource_disk_mount_point)
+        node_disc = node.features[Disk]
+        resource_disks = node_disc.get_resource_disks()
+        if not resource_disks:
+            raise LisaException("Resource disk not found")
+        if "nvme" in resource_disks[0]:
+            raise SkippedException(
+                f"Resource disk type is NVMe and the VM has {len(resource_disks)} NVMe disks"  # noqa: E501
```
squirrelsc: The skip message doesn't explain why nvme disk should be skipped.

lisa/features/disks.py
```diff
@@ -71,6 +71,9 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
     def get_resource_disk_mount_point(self) -> str:
         raise NotImplementedError
 
+    def get_resource_disks(self) -> List[str]:
```
squirrelsc, line 75: The default implementation should return an empty list, because "resource disk" is a concept on cloud or Azure, not all platforms. All test cases, which used the resource disk, should handle the empty list.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1897,6 +1897,20 @@ def remove_data_disk(self, names: Optional[List[str]] = None) -> None:
...
+    def get_resource_disks(self) -> List[str]:
+        resource_disk_mount_point = self.get_resource_disk_mount_point()
+        resource_disks = []
+        try:
+            resourcedisk = self._node.features[Disk].get_partition_with_mount_point(
+                resource_disk_mount_point
+            )
+            if resourcedisk:
+                resource_disks = [resourcedisk.name]
+        except AssertionError:
```
squirrelsc: Is there a way not check by the error? The error can be happened by other reasons, the eating code may hide other problems.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1115,32 +1115,51 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
-
-        # verify 'WARNING: THIS IS A TEMPORARY DISK' contained in
-        # DATALOSS_WARNING_README.txt file.
-        read_text = node.tools[Cat].read(file_path, force_run=True, sudo=True)
-        assert_that(
-            read_text,
-            f"'WARNING: THIS IS A TEMPORARY DISK' should be present in {file_path}",
-        ).contains("WARNING: THIS IS A TEMPORARY DISK")
+        node_disc = node.features[Disk]
+        resource_disks = node_disc.get_resource_disks()
```
squirrelsc: Add a method in `Disk` feature, called like `get_disk_controller_type`, and it accept mount point as a parameter. Please also reduce duplicate code with existing `get_os_disk_controller_type`.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1115,32 +1115,51 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
+        if not resource_disks:
+            raise LisaException("Resource disk not found")
+        if "nvme" in resource_disks[0]:
```
squirrelsc: As above comments, use the type enum, instead of checking by string.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1115,32 +1115,51 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
+            raise SkippedException(
+                f"Resource disk type is NVMe and the VM has {len(resource_disks)} NVMe disks"  # noqa: E501
```
squirrelsc: This error message is unclear, why the tests should be skipped, if the resource disk is nvme. Please check if it's valid to skip for nvme, if so, update the description.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1115,32 +1115,51 @@ def verify_no_pre_exist_users(self, node: Node) -> None:
...
+            # since it uses UFS file system
+            if not isinstance(node.os, BSD):
+                fold_path = f"{resource_disk_mount_point}/lost+found"
+                folder_exists = node.tools[Ls].path_exists(fold_path, sudo=True)
+                assert_that(folder_exists, f"{fold_path} should be present").is_true()
+
+            # Verify lost+found folder exists
+            # Skip this step for BSD as it does not have lost+found folder
+            # since it uses UFS file system
+            if not isinstance(node.os, BSD):
```
squirrelsc: Is it miscopied from above lines/ They are looks the same as above parts.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1159,14 +1178,29 @@ def verify_resource_disk_readme_file(self, node: RemoteNode) -> None:
...
-            # or interrupt
-            assert_that(
-                partition.fstype,
-                "Resource disk file system type should not equal to ntfs",
-            ).is_not_equal_to("ntfs")
+        node_disc = node.features[Disk]
+        resource_disks = node_disc.get_resource_disks()
+        if not resource_disks:
+            raise LisaException("Resource disk not found")
+        if "nvme" in resource_disks[0]:
```
squirrelsc: Wrap the repeat logic to reuse.

lisa/features/nvme.py
SRIKKANTH: Added a new method "_remove_nvme_os_disk()"

microsoft/testsuites/core/storage.py
SRIKKANTH: Modified the exception messages for better understanding.

lisa/features/disks.py
SRIKKANTH: Fixed as suggested

lisa/sut_orchestrator/azure/features.py
SRIKKANTH: Modified the logic to detect the issue at lower level to avoid this.

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Fixed it.

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Introduced new constants for NVMe and SCSI resource disc types.

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Modified the exception messages for better understanding.

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Removed the duplicate. Thanks for pointing it out.

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Added a new method "get_resource_disk_type" for this. Which raises exception if no resource disk is available.

lisa/features/disks.py
```diff
@@ -71,6 +72,12 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+        return []
+
+    def get_resource_disk_type(self) -> str:
+        return ""
```
squirrelsc: It should be RESOURCE_DISK_TYPE_SCSI by default. The empty string is confusing. Actually, an enumeration is better to prevent misuse of a type.

microsoft/testsuites/core/azure_image_standard.py
squirrelsc: Please refine `get_os_disk_controller_type` to use the same logic. You can see the `get_os_disk_controller_type` contains logic for FreeBSD too. If you implement another `get_resource_disk_type`. The FreeBSD needs to fix the new failure again.

microsoft/testsuites/core/azure_image_standard.py
squirrelsc: New constants are right, but enum is needed too.

lisa/features/disks.py
SRIKKANTH: Fixed

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Fixed as suggested

microsoft/testsuites/core/azure_image_standard.py
SRIKKANTH: Added new method "get_disc_type" to re-use in both get_os_disk_controller_type and get_resource_disk_type

squirrelsc: Please use `disk` instead of `disc` to keep consistent.

SRIKKANTH: Fixed

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3433, opened 2024-09-23
Fix extension in 'updating' status issue

LiliDeng:
Recently there are some extensions failed for below reason.
operations.done() is True, but provisioning_state status is `Updating`, I tried to add loop in wait_operation, but found the status is always `Updating`, so to get the extension to retrieve the provisioning_state status, verified the fix worked well.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -5,6 +5,7 @@ import copy
 import json
 import re
 import string
+import time
 from dataclasses import dataclass, field
 from functools import partial
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -2853,15 +2857,42 @@ def create_or_update(
...
             vm_name=self._vm_name,
             vm_extension_name=name,
             extension_parameters=extension_parameters,
         )
-        result = wait_operation(operation, timeout)
 
-        return result
+        interval = 10
+        timer = create_timer()
+        while timeout > timer.elapsed(False):
```
squirrelsc: use `>=` to prevent some corner case.

squirrelsc (approved):

PR #3435, opened 2024-09-23
add scripts_execution_transformer

LiliDeng:


lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+@dataclass_json()
+@dataclass
+class ScriptsExecutionTransformerSchema(DeploymentTransformerSchema):
+    dependent_packages: List[str] = field(default_factory=list)
+    scripts: List[ScriptEntry] = field(default_factory=list)
+
+
+class ScriptExecutionResult:
```
squirrelsc: It's similar with the ExecutionResult. Does it need?

lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,87 @@
...
+    output: str = ""
+    exit_code: Optional[int] = -1
+
+
+class ScriptsExecutionTransformer(DeploymentTransformer):
```
squirrelsc: `ScriptTransformer` has enough information.

lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,87 @@
...
+from lisa.transformers.deployment_transformer import (
+    DeploymentTransformer,
+    DeploymentTransformerSchema,
+)
+
+
+@dataclass_json()
+@dataclass
+class ScriptEntry:
+    script: str = ""
```
squirrelsc: The script should be defined as an enum, we only invoke what we supported. It's for security and compatibility. For new, we have bash only. In future, maybe python, sh...

squirrelsc: give a default value.

lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,87 @@
...
+        runbook: ScriptsExecutionTransformerSchema = self.runbook
+        os = self._node.os
+        if isinstance(os, Posix):
+            os.install_packages(runbook.dependent_packages)
+
+        results: Dict[str, Any] = {}
+        for item in runbook.scripts:
+            args = ""
+            if item.args:
+                args = " ".join(item.args)
```
squirrelsc: A simple join may not work in some strings. Maybe we just regard args as a string. So the user can modify the escape chars, and make it working.

lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,87 @@
...
+        return ScriptsExecutionTransformerSchema
+
+    @property
+    def _output_names(self) -> List[str]:
+        return [self.__results_name]
+
+    def _internal_run(self) -> Dict[str, Any]:
+        runbook: ScriptsExecutionTransformerSchema = self.runbook
+        os = self._node.os
+        if isinstance(os, Posix):
```
squirrelsc: Does it need to filter by Posix? The dependencies should be able to use by all oses.

lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,87 @@
...
+            os.install_packages(runbook.dependent_packages)
+
+        results: Dict[str, Any] = {}
+        for item in runbook.scripts:
+            args = ""
+            if item.args:
+                args = " ".join(item.args)
+            command = f"{item.interpreter} {item.script} {args}"
+            execution_result = self._node.execute(command, sudo=True, shell=True)
```
squirrelsc: Does it need to support set the expected exit code?

lisa/transformers/scripts_execution_transformer.py
LiliDeng: the same name has existed

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

lisa/transformers/scripts_execution_transformer.py
squirrelsc: Ok... How about ScriptFileTransformer?

lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,95 @@
...
+    BASH = "bash"
+
+
+@dataclass_json()
+@dataclass
+class ScriptEntry:
+    script: str = ""
+    interpreter: ScriptInterpreter = ScriptInterpreter.BASH
+    args: Optional[str] = None
+
```
squirrelsc: Add one more field `expected_exit_code`. The default value is `0`. It can be set to `None`, and ignore the exit code.

lisa/transformers/scripts_execution_transformer.py
```diff
@@ -0,0 +1,95 @@
...
+
+@dataclass_json()
+@dataclass
+class ScriptsExecutionTransformerSchema(DeploymentTransformerSchema):
+    dependent_packages: List[str] = field(default_factory=list)
+    scripts: List[ScriptEntry] = field(default_factory=list)
+    reboot: bool = field(default=False)
+    exit_on_error: bool = field(default=False)
```
squirrelsc: See above comments, with the `expected_exit_code`, this field is not needed.

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

lisa/transformers/scripts_execution_transformer.py
LiliDeng: updated

squirrelsc (approved):

PR #3437, opened 2024-09-24
Libvirt: Enable security profile feature

gamora12:
- Add the security profile feature to handle CVM guest type for nested case and pass the security profile to the lisa platform. 
- Add launch security configuration to the cloud-hypervisor platform with type "sev" that will generate the xml to launch guest CVM.
- Change cloud-hypervisor schema in libvirt, use kernel parameter to determine local or remote kernel file.

USAGE:

```
requirement:
      core_count: 1
      memory_mb: 1000
      features:
        items:
          - type: Security_Profile
            security_profile: "cvm"
      cloud-hypervisor:
        disk_img: ""
        disk_img_format: ""
        kernel: 
          path: ""
          is_remote_path: True
        firmware: ""
        cloud_init:
          extra_user_data: ""
        disk_img_resize_gib: 25     
```


lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -113,9 +114,21 @@ def _create_node_domain_xml(
 
         os_type = ET.SubElement(os, "type")
         os_type.text = "hvm"
-
         os_kernel = ET.SubElement(os, "kernel")
-        os_kernel.text = node_context.firmware_path
+        if node_context.guest_vm_type == "ConfidentialVM":
+            os_kernel.text = "/usr/share/cloud-hypervisor/cvm/linux.bin"
```
smit-gardhariya: Can we make it runbook parameter. If nothing set from runbook, we can use this as default path.

lisa/util/constants.py
```diff
@@ -116,6 +116,8 @@
 SECURITY_PROFILE_CVM = "cvm"
 SECURITY_PROFILE_STATELESS = "stateless"
 
+GUEST_VM_TYPE_STANDARD = "STANDARD"
```
squirrelsc: Can it reuse above security profile items?

squirrelsc: If it's used in the libvirt platform only, it doesn't need to define in this file.

lisa/sut_orchestrator/libvirt/features.py
```diff
@@ -0,0 +1,57 @@
...
+        super()._initialize(*args, **kwargs)
+
+    @classmethod
+    def settings_type(cls) -> Type[schema.FeatureSettings]:
+        return SecurityProfileSettings
+
+    @classmethod
+    def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
+        environment = cast(Environment, kwargs.get("environment"))
+        security_profile = [kwargs.get("settings")]
```
squirrelsc: It doesn't need be an array.

lisa/sut_orchestrator/libvirt/features.py
```diff
@@ -0,0 +1,57 @@
...
+from lisa.sut_orchestrator.libvirt.context import get_node_context
+
+
+@dataclass_json()
+@dataclass()
+class SecurityProfileSettings(features.SecurityProfileSettings):
+    def __hash__(self) -> int:
+        return hash(self._get_key())
+
+    def _get_key(self) -> str:
```
squirrelsc, line 18: It doesn't need to rewrite this method.

gamora12: We will still need to rewrite the _get_key method since it's different from the base class method. And __hash__ method still needs to be rewritten otherwise we're getting this error- unhashable type: 'SecurityProfileSettings'

lisa/sut_orchestrator/libvirt/features.py
```diff
@@ -0,0 +1,57 @@
...
+        security_profile = [kwargs.get("settings")]
+        for node in environment.nodes._list:
+            if security_profile:
+                settings = security_profile[0]
+                assert isinstance(settings, SecurityProfileSettings)
+                assert isinstance(settings.security_profile, SecurityProfileType)
+                node_context = get_node_context(node)
+                node_context.guest_vm_type = cls._security_profile_mapping[
+                    settings.security_profile
+                ]
```
pupacha:
```suggestion
        if security_profile:
            settings = security_profile[0]
            for node in environment.nodes._list:
                assert isinstance(settings, SecurityProfileSettings)
                assert isinstance(settings.security_profile, SecurityProfileType)
                node_context = get_node_context(node)
                node_context.guest_vm_type = cls._security_profile_mapping[
                    settings.security_profile
                ]
```

lisa/util/constants.py
pupacha: We can remove this Variables. I dont see it being used anywhere.

pupacha (commented):
Good work. Just minor comments.

lisa/sut_orchestrator/libvirt/features.py
gamora12: Thanks, I'll follow this change.

lisa/sut_orchestrator/libvirt/ch_platform.py
pupacha: @smit-gardhariya @gamora12  I'd suggest we have single parameter for firmware and kernel (since at the end they both are passed as `--kernel` parameter to cloud-hypervisor).

We can remove existing `firmware` field from `CloudHypervisorNodeSchema` and add a new param `kernel`. Something like following:
```
kernel:
  path: ''abc/xyz'
  is_remote: True
```

gamora12: Sure, we can do that.

lisa/sut_orchestrator/libvirt/context.py
```diff
@@ -48,6 +48,7 @@ class NodeContext:
     vm_name: str = ""
     firmware_source_path: str = ""
     firmware_path: str = ""
+    guest_vm_type: str = ""
```
anirudhrb: Please make this an enum type instead of string. That way, it is well defined what the possible values for `guest_vm_type` are.

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -133,10 +133,17 @@ def __post_init__(self) -> None:
...
+@dataclass
+class KernelSchema:
+    path: str = ""
+    is_remote_path: bool = False
+
+
 @dataclass_json()
 @dataclass
 class CloudHypervisorNodeSchema(BaseLibvirtNodeSchema):
     # Local path to the cloud-hypervisor firmware.
```
anirudhrb, line 147: Please update this comment too.

lisa/sut_orchestrator/libvirt/context.py
gamora12: The ```guest_vm_type``` will take value from security profile mapping which has a list of possible types of guest vm and returns a string literal of guest vm type. So, there is an indirect enforcement of possible values.
Please refer line no.  54 in features.py. If you think it's required, we can do it.

anirudhrb: The security profile mapping could use the enum as well instead of string literals.

Let's make it a direct enforcement of possible values by defining an enum. That way someone reading the code can easily figure out the possible values by just looking at the type of the `guest_vm_type` field. Otherwise, they would need to do some digging around to find the relevant code in features.py...

gamora12: Sure, will update the security profile mapping accordingly.

gamora12:
> @gamora12 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

lisa/sut_orchestrator/libvirt/context.py
```diff
@@ -43,11 +44,16 @@ class DevicePassthroughContext:
     managed: str = ""
 
 
+@dataclass
+class GuestVmType(Enum):
+    Standard = ""
```
anirudhrb: Why is this an empty string? Can't we have "Standard" as the string here?

lisa/sut_orchestrator/libvirt/features.py
```diff
@@ -0,0 +1,56 @@
...
+
+    @classmethod
+    def settings_type(cls) -> Type[schema.FeatureSettings]:
+        return SecurityProfileSettings
+
+    @classmethod
+    def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
+        environment = cast(Environment, kwargs.get("environment"))
+        settings = kwargs.get("settings")
+        if settings:
```
anirudhrb: nit: you could reduce nesting with:

```py
if not settings:
    return
```

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -133,10 +133,15 @@ def __post_init__(self) -> None:
...
+
 @dataclass_json()
 @dataclass
 class CloudHypervisorNodeSchema(BaseLibvirtNodeSchema):
-    # Local path to the cloud-hypervisor firmware.
-    # Can be obatained from:
-    # https://github.com/cloud-hypervisor/rust-hypervisor-firmware
-    firmware: str = ""
+    # Local or remote path to the cloud-hypervisor kernel.
+    kernel: Optional[KernelSchema] = None
```
anirudhrb, line 152: This is a backwards incompatible schema change. I'm fine with it but we need to make sure to update our runbooks as soon as the new version of LISA lands.

Another option is to support both fields for now and remove `firmware` in a later release. This gets rid of the urgency to update runbooks with new LISA release.

cc: @pupacha @smit-gardhariya

anirudhrb (commented):
Mostly LGTM. Just some minor suggestions.

lisa/sut_orchestrator/libvirt/features.py
gamora12: Sure.

lisa/sut_orchestrator/libvirt/context.py
gamora12: Yes, we can have "Standard" as string.

lisa/sut_orchestrator/libvirt/schema.py
gamora12: Yes, it's a breaking change, we can have both `firmware` and `kernel` in runbook schema for now.

pupacha: @gamora12 @anirudhrb  Since the schema change is not so complex, we can update our runbooks with the new LISA release.

anirudhrb: Just discussed offline with @pupacha. Summarizing it here:

It is not about the schema change being complex or simple. It is about the urgency it creates to update the runbooks immediately after new release. It could happen that we might miss it and all the pipelines will start failing. I guess this is what we have been doing in the past. However, it is a good practice in general to make backwards compatible changes and eventually deprecate the old interface. So, let's start practicing that starting with this change.

gamora12: Sure, we should continue this practice. We can remove the firmware parameter in the next phase.

lisa/sut_orchestrator/libvirt/ch_platform.py
```diff
@@ -57,13 +62,25 @@ def _configure_node(
...
+            if self.host_node.is_remote:
+                if node_runbook.kernel.is_remote_path:
+                    node_context.kernel_path = node_runbook.kernel.path
+                else:
+                    node_context.kernel_source_path = node_runbook.kernel.path
+                    node_context.kernel_path = os.path.join(
+                        self.vm_disks_dir, os.path.basename(node_runbook.kernel.path)
+                    )
+            else:
+                node_context.kernel_path = node_runbook.kernel.path
```
pupacha:
```suggestion
            if self.host_node.is_remote and not node_runbook.kernel.is_remote_path:
                node_context.kernel_source_path = node_runbook.kernel.path
                node_context.kernel_path = os.path.join(
                    self.vm_disks_dir, os.path.basename(node_runbook.kernel.path)
                )
            else:
                node_context.kernel_path = node_runbook.kernel.path
```

lisa/sut_orchestrator/libvirt/schema.py
```diff
@@ -133,10 +133,16 @@ def __post_init__(self) -> None:
...
+
 @dataclass_json()
 @dataclass
 class CloudHypervisorNodeSchema(BaseLibvirtNodeSchema):
-    # Local path to the cloud-hypervisor firmware.
-    # Can be obatained from:
-    # https://github.com/cloud-hypervisor/rust-hypervisor-firmware
+    # Local or remote path to the cloud-hypervisor kernel.
     firmware: str = ""
+    kernel: Optional[KernelSchema] = None
```
anirudhrb:
```suggestion
    # DEPRECATED: use the 'kernel' field instead.
    # Local path to the cloud-hypervisor firmware.
    # Can be obtained from:
    # https://github.com/cloud-hypervisor/rust-hypervisor-firmware
    firmware: str = ""
    # Local or remote path to the cloud-hypervisor kernel.
    kernel: Optional[KernelSchema] = None
```

anirudhrb (approved):
LGTM!

lisa/sut_orchestrator/libvirt/schema.py
gamora12: Thanks, this gives a better understanding of what we're doing.

lisa/sut_orchestrator/libvirt/ch_platform.py
gamora12: Thanks for the suggestion, I'll make the change and push the update shortly.

pupacha (approved):
LGTM!

anirudhrb (approved):

LiliDeng (approved):

PR #3438, opened 2024-09-24
Remove SecurityProfileType.Standard from CVM images

kamalca:
When calculating image capabilities, do not include `Standard` for images supporting CVM. 
Currently LISA will default to a standard SKU for CVM images, but this is not really a valid testing scenario.

For example, all automated testing of CVM Ubuntu 24.04 has been on non-CVM SKUs
### canonical ubuntu-24_04-lts cvm
| VMSize| TestCount|
|--------|--------|
| Standard_DS3_v2	|  116 |
| Standard_DS2_v2	|  55 |
| Standard_D2s_v3	|  21 |
| Standard_D2s_v5	|  13 | 
| Standard_E2as_v5	| 10 |
| Standard_D4s_v3	| 4 |

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -378,16 +378,19 @@ class AzureImageSchema(schema.ImageSchema):
         self, raw_features: Dict[str, Any], log: Logger
     ) -> None:
         security_profile = raw_features.get("SecurityType")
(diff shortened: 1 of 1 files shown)


kamalca:
Note: I also had to remove `SecureBoot` because many standard SKUs also support secure boot, but that is also not the intended use case for CVM images.

LiliDeng:
What is the expectation when use cvm + default security type?

LiliDeng (approved):

PR #3443, opened 2024-09-27
add azure credential schema for different authentication

paxue:
add azure credential schema for different authentication method: certificate, assert, client secret


squirrelsc:
@LiliDeng LGTM, we have done review internally.

LiliDeng (approved):

PR #3446, opened 2024-09-27
doc: add lisa case training data.

squirrelsc:
It generate a jsonl file for llm training purpose. The input is test case descriptions of a test file, output is the test case file content.

### DIFF
--- a/docs/tools/finetune_gen.py
+++ b/docs/tools/finetune_gen.py
@@ -1,0 +1,84 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""
+This script generates fine-tune data for AI models.
+"""
+import ast
+import json
+import os
+from pathlib import Path
+from typing import Dict
+
+from .doc_generator import TESTS, ClassVisitor, FuncVisitor, extract_metadata, load_path
+
+base_path = Path(__file__).parent
+file_path = (base_path / "../_build/html/finetune_cases.jsonl").resolve()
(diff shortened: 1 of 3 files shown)


docs/tools/finetune_gen.py
```diff
@@ -0,0 +1,90 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+"""
+Update test_spec.rst with test metadata
```
LiliDeng: update comments

docs/tools/finetune_gen.py
```diff
@@ -0,0 +1,90 @@
...
+from pathlib import Path
+from typing import Dict
+
+from .doc_generator import TESTS, ClassVisitor, FuncVisitor, extract_metadata, load_path
+
+base_path = Path(__file__).parent
+file_path = (base_path / "../_build/html/finetune_cases.jsonl").resolve()
+system_prompt = "You are a helpful assistant to write Python code for linux validation."
+user_prompt = (
+    "Please bases on the test case description to write a LISA test case.\n\n\n"
```
LiliDeng: `Please write a LISA test case based on the test case description.`

docs/tools/finetune_gen.py
squirrelsc: fixed

docs/tools/finetune_gen.py
squirrelsc: fixed

LiliDeng (approved):

PR #3448, opened 2024-09-28
Use mq-deadline for perf nvme tests

kanchansenlaskar:
Changing the scheduler to this does not show workqueue lockups, even at higher qdepths. These scheduling algorithms, along with other queue parameters are there to allow the developers/users to tune them, based on the devices and system load that is that is going to be there on the system. higher system load on a few CPUs can lead to workqueue lockups when hrtimers and interrupts start taking longer time.

This change is tested with the following two tests in oracle:
perf_nvme
perf_nvme_io_uring

### DIFF
--- a/microsoft/testsuites/performance/nvmeperf.py
+++ b/microsoft/testsuites/performance/nvmeperf.py
@@ -11,6 +11,7 @@ from lisa import (
 )
 from lisa.features import Nvme, NvmeSettings
 from lisa.messages import DiskSetupType, DiskType
+from lisa.operating_system import Oracle
 from lisa.testsuite import TestResult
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/performance/nvmeperf.py
```diff
@@ -77,6 +78,19 @@ def _perf_nvme(
                 node.get_pure_path(f"/sys/block/{disk_name}/queue/rq_affinity"),
                 sudo=True,
             )
+            if isinstance(node.os, Oracle):
+                # Changing the scheduler to this does not show workqueue lockups,
```
squirrelsc, line 82: Please also explain why it applies to Oracle only.

PR #3449, opened 2024-09-30
Allow File share creation with Access key with default false

shekharsorot:
Currently Azure files Does not supports MSI based authentication. To allow AD based authentication will require additional work such as setting up ADDS ( linux only ) and performing file share + node domain join.
As such, the documented and supported approach is to mount SMB with storage account name / access key as user/pass k:p.
This PR adds two params to the definition "create_file_share" within features.py to allow creation of file shares with access key enabled. This feature is default to false and will require specifying in extensions such as xfstesting within Microsoft test cases.

Note: this PR is required to support XFStesting SMB tests.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -3254,7 +3254,7 @@ class AzureFileShare(AzureFeatureMixin, Feature):
             resource_group_name=resource_group_name,
             location=location,
             log=self._log,
-            allow_shared_key_access=allow_shared_key_access
+            allow_shared_key_access=allow_shared_key_access,
         )
 
         for share_name in file_share_names:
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -2718,6 +2718,7 @@ def create_share(self) -> None:
             sku="Premium_LRS",
             kind="FileStorage",
             enable_https_traffic_only=False,
+
```
LiliDeng: remove blank line

shekharsorot: blank line at 2721 removed.

LiliDeng (dismissed):

LiliDeng (approved):

PR #3450, opened 2024-09-30
Support kdump in ready platform which missing serial console feature

LiliDeng:


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -390,76 +471,38 @@ def _check_kdump_result(
...
+                node=node,
+                log_path=log_path,
+                log=log,
+                kdump=kdump,
+                timer=timer,
+                serial_console=serial_console,
+            )
+            if dump_file_found:
+                return
+        self._get_console_log(
```
squirrelsc: How the ready platform supported? It looks still raising exceptions for ready platform. How about write a mock SerialConsole support for ready platform? It do nothing for save serial log, but if some test cases need to ready the console log, raise an exception.

LiliDeng: updated

squirrelsc (approved):

PR #3453, opened 2024-09-30
Pass custom timeout

kanchansenlaskar:
This is the change to incorporate a custom timeout for the test executions.

### DIFF
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -1487,6 +1487,9 @@ class TestCase(BaseTestCaseFilter):
             field_function=fields.Int, validate=validate.Range(min=0)
         ),
     )
+    # each test case will have this timeout
+    # if it is 0 it will use default timeout from test case
+    timeout: int = 0
     # each case with this rule will be run in a new environment.
     use_new_environment: bool = False
     # Once it's set, failed test result will be rewrite to success
--- a/lisa/testselector.py
+++ b/lisa/testselector.py
@@ -126,6 +126,13 @@ def _apply_settings(
         applied_case_data.use_new_environment or case_runbook.use_new_environment
     )
 
+    # check for positive value just to be clearer
+    applied_case_data.timeout = (
+        max(case_runbook.timeout, applied_case_data.timeout)
+        if (case_runbook.timeout and case_runbook.timeout > 0)
+        else applied_case_data.timeout
+    )
+
     # use default value from selector
     applied_case_data.select_action = action
 
--- a/lisa/testsuite.py
+++ b/lisa/testsuite.py
@@ -522,6 +522,7 @@ class TestCaseRuntimeData:
         self.select_action: str = ""
         self.times: int = 1
         self.retry: int = 0
+        self.timeout: int = metadata.timeout
         self.use_new_environment: bool = metadata.use_new_environment
         self.ignore_failure: bool = False
         self.environment_name: str = ""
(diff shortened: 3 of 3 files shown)


lisa/testselector.py
```diff
@@ -125,6 +125,7 @@ def _apply_settings(
     applied_case_data.use_new_environment = (
         applied_case_data.use_new_environment or case_runbook.use_new_environment
     )
+    applied_case_data.timeout = case_runbook.timeout or applied_case_data.timeout
```
squirrelsc, line 128: When is this change needed? If it needs to extend the timeout, it should take max of both. If it wants to run test cases faster, it should take min of them. They are conflict goals.

For the `or` condition, it takes case_runbook higher priority, for some designed long running test cases, they will mostly failed. It's not right to overwrite case settings always.

lisa/testsuite.py
```diff
@@ -630,7 +631,10 @@ def start(
                 constants.RUN_LOCAL_LOG_PATH
             ).as_posix()
             case_result.set_status(TestStatus.RUNNING, "")
-            case_timeout = case_result.runtime_data.metadata.timeout
+            case_timeout = (
```
squirrelsc, line 636: When is this change needed? If it needs to extend the timeout, it should take max of both. If it wants to run test cases faster, it should take min of them. They are conflict goals.

For the `or` condition, it takes case_runbook higher priority, for some designed long running test cases, they will mostly failed. It's not right to overwrite case settings always.

lisa/variable.py
```diff
@@ -371,6 +371,7 @@ def _replace_variables(data: Any, variables: Dict[str, VariableEntry]) -> Any:
                             f"cannot find variable '{variable_name[2:-1]}', "
                             "make sure its value filled in runbook, "
                             "command line or environment variables."
+                            f"data: {data}"
```
squirrelsc: When is it needed? I'm worry about if it prints too early to leak secrets in the log.

lisa/testsuite.py
kanchansenlaskar: Changed it, it will take the max of both.
However, it will consider the one in metadata if the one passed from outside is None

lisa/testselector.py
squirrelsc: It needs to use max value too.

lisa/schema.py
```diff
@@ -1487,6 +1487,7 @@ class TestCase(BaseTestCaseFilter):
             field_function=fields.Int, validate=validate.Range(min=0)
         ),
     )
+    timeout: int = 3600
```
squirrelsc: It needs to be optional, so if some test case is shorter timeout, it should be respected.

kanchansenlaskar: made it optional

lisa/testselector.py
kanchansenlaskar: done

lisa/schema.py
```diff
@@ -1487,6 +1487,8 @@ class TestCase(BaseTestCaseFilter):
             field_function=fields.Int, validate=validate.Range(min=0)
         ),
     )
+    # each test case will have this timeout, default is 3600 seconds
+    timeout: Optional[int] = 3600
```
squirrelsc: It's not optional by default, please set the default value to `None`. So the test case default timeout are respected.

lisa/schema.py
```diff
@@ -1487,6 +1487,9 @@ class TestCase(BaseTestCaseFilter):
             field_function=fields.Int, validate=validate.Range(min=0)
         ),
     )
+    # each test case will have this timeout
+    # if it iss 0 it will use default timeout from test case
```
squirrelsc: typo: `iss` to `is`

lisa/schema.py
kanchansenlaskar: I have set the default value to 0 to maintain the consistency with the other places in other repo. Please do let me know if you feel otherwise.

squirrelsc:
@LiliDeng LGTM. @kanchansenlaskar, please wait Lili having a regression test.

LiliDeng (approved):

PR #3454, opened 2024-10-01
Regex to fix mount issue

r-dailey:
The zfs freebsd image uses a different format for the mounts and was causing a mount test to fail. I updated the regex filter to handle this additional formatting.

### DIFF
--- a/lisa/tools/mount.py
+++ b/lisa/tools/mount.py
@@ -72,8 +72,9 @@ class Mount(Tool):
     )
 
     # /dev/da1p1 on /mnt/resource (ufs, local, soft-updates)
+    # zroot/ROOT/default on / (zfs, local, nfsv4acls)
     _partition_info_regex_bsd = re.compile(
-        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+"
+        r"\s*(?:\/dev\/|zroot\/)(?P<name>.*)\s+on\s+(?P<mount_point>.*?)\s+"
         r"(\((?P<type>.*),(?P<options>.*)\))"
     )
 


lisa/tools/mount.py
```diff
@@ -73,7 +73,7 @@ class Mount(Tool):
 
     # /dev/da1p1 on /mnt/resource (ufs, local, soft-updates)
     _partition_info_regex_bsd = re.compile(
-        r"\s*/dev/(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+"
+        r"\s*(?:\/dev\/|zroot\/)(?P<name>.*)\s+on\s+(?P<mount_point>.*)\s+"
```
squirrelsc: Provide new examples above, and make sure there is no regression on existing patterns.

r-dailey: Tested the changes on all the freebsd image versions there is no regression and I have added an example below the existing one

squirrelsc: I tested above examples on https://regex101.com/. It looks the new case doesn't return the mount point. 

```suggestion
        r"\s*(?:\/dev\/|zroot\/)(?P<name>.*)\s+on\s+(?P<mount_point>.*?)\s+"
```

r-dailey: I also tested it on regex101.com and it works. The regex does check for a newline as the end of the capture for the mount point so if you just copy and paste the examples from the code into regex101.com without a trailing newline it will cause some weird behavior with the second example. Would it be helpful to add a newline character to the end of the examples so that it is more explicit?

squirrelsc (approved):

PR #3455, opened 2024-10-01
ip_service_tag

kamalca:
New security standard will require all IPs to be tagged.

Please note this IP tagging system is different from general resource tagging.

### DIFF
--- a/lisa/sut_orchestrator/azure/arm_template.bicep
+++ b/lisa/sut_orchestrator/azure/arm_template.bicep
@@ -46,6 +46,9 @@ param data_disks array
 @description('whether to use ultra disk')
 param is_ultradisk bool = false
 
+@description('IP Service Tags')
+param ip_service_tags object
+
 var vnet_id = virtual_network_name_resource.id
 var node_count = length(nodes)
 var availability_set_name_value = 'lisa-availabilitySet'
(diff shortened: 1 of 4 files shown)


lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -46,6 +46,9 @@ param data_disks array
 @description('whether to use ultra disk')
 param is_ultradisk bool = false
 
+@description('IP Service Tag')
+param ip_service_tag string
```
squirrelsc: Kameron, is the service tag a public Azure concept? If it's not, please implement it by internal extensions.

kamalca: @squirrelsc 
Service Tags are a public Azure concept, but I believe the use of 'FirstPartyUsage' is not.

I see two options:
1. Generalize the implementation to service tags generally and the runbook can pass in something like "FirstPartyUsage:\SyntheticLoad" (key value pair)
2.  Implement the full thing as an internal extension

I slightly lean towards 1. since it would allow others to make use of the functionality.

However, I am not sure whether this is a real use case and it may only be needlessly complicating the tools.

For reference, the [public documentation](https://learn.microsoft.com/en-us/azure/templates/microsoft.network/publicipaddresses?pivots=deployment-language-arm-template#iptag-1) gives 'FirstPartyUsage' as an example value, but I had to add an AFEC feature flag to be able to use it on our subscriptions. Also the name itself suggests it is not for public use.

Thoughts?

squirrelsc: Please pass in a key/value pair.

kamalca: Changed

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3456, opened 2024-10-02
Standardize CpuArchitecture enum

mcgov:
Preparing for the DPDK 32bit test by cleaning up how we handle cpu arch names.

- Replaces adhoc handling of arch enums using a shared CpuArchitecture Enum.
- Adds i386 to CpuArchitecture enum
- get_architecture now returns CpuArchitecture instead of a bare string.

### DIFF
--- a/lisa/tools/lscpu.py
+++ b/lisa/tools/lscpu.py
@@ -57,7 +57,6 @@ ArchitectureNames = {
     "aarch64": CpuArchitecture.ARM64,
     "amd64": CpuArchitecture.X64,
     "arm64": CpuArchitecture.ARM64,
-    "i386": CpuArchitecture.I386,
 }
 
 
--- a/lisa/tools/lscpu.py
+++ b/lisa/tools/lscpu.py
@@ -8,7 +8,7 @@ from typing import Any, List, Optional, Tuple, Type
 from assertpy import assert_that
 
 from lisa.executable import Tool
-from lisa.operating_system import FreeBSD, Posix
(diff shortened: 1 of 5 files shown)


lisa/tools/lscpu.py
```diff
@@ -52,8 +52,13 @@ def __repr__(self) -> str:
...
 
 
-ARCH_X86_64 = "x86_64"
-ARCH_AARCH64 = "aarch64"
+ArchitectureNames = {
+    "x86_64": CpuArchitecture.X64,
+    "aarch64": CpuArchitecture.ARM64,
+    "amd64": CpuArchitecture.X64,
+    "arm64": CpuArchitecture.ARM64,
+    "i386": CpuArchitecture.I386,
```
squirrelsc: Should it map to X64? If it's not found in existing scenarios, it doesn't need to be added.

mcgov: I'm using it in my next PR that's based on this branch, it doesn't hurt to add it while I'm touching this code anyway.
This is my effort to break up reviews into smaller chunks 😁

mcgov: removed for now

squirrelsc: If it will be used, it's ok to add it. But if it's just for possible backward compatibility, it doesn't need. It's a long list for old architectures.

squirrelsc:
@LiliDeng LGTM

mcgov:
@squirrelsc Lili is on vacation for the rest of the week. I'll share some test results.

squirrelsc (approved):

PR #3457, opened 2024-10-02
RdmaInstaller: refactor rdma-core dependency and source install

mcgov:
refactors how the DPDK tests handle the rdma-core dependency by implementing the Installer class for it.
Now allows arbitrary source builds via tar or git as well as handling the MANA special case of building a recent rdma-core version.

### DIFF
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -1,7 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-from datetime import datetime
 from pathlib import PurePath
 from typing import Any, Callable, Dict, List, Optional, Sequence, Type, Union
 
@@ -275,43 +274,30 @@ def force_dpdk_default_source(variables: Dict[str, Any]) -> None:
         variables["dpdk_source"] = DPDK_STABLE_GIT_REPO
 
 
-# rough check for ubuntu supported versions.
-# assumes:
-# - canonical convention of YEAR.MONTH for major versions
-# - canoical release cycle of EVEN_YEAR.04 for lts versions.
-# - 4 year support cycle. 6 year for ESM
-# get the age of the distro, if negative or 0, release is new.
-# if > 6, distro is out of support
-def _get_ubuntu_distro_age(distro: Ubuntu) -> int:
-    version_info = distro.information.version
-    # check release is within esm window
-    year_string = str(datetime.today().year)
-    assert_that(len(year_string)).described_as(
-        "Package bug: The year received from datetime module is an "
-        "unexpected size. This indicates a broken package or incorrect "
-        "date in this computer."
-    ).is_greater_than_or_equal_to(4)
-    # TODO: handle the century rollover edge case in 2099
-    current_year = int(year_string[-2:])
-    release_year = int(version_info.major)
-    # 23-18 == 5
-    # long term support and extended security updates for ~6 years
-    return current_year - release_year
+_UBUNTU_LTS_VERSIONS = ["24.4.0", "22.4.0", "20.4.0", "18.4.0"]
 
 
+# see https://ubuntu.com/about/release-cycle
 def is_ubuntu_latest_or_prerelease(distro: Ubuntu) -> bool:
-    distro_age = _get_ubuntu_distro_age(distro)
-    return distro_age <= 2
+    return bool(distro.information.version >= max(_UBUNTU_LTS_VERSIONS))
 
 
+# see https://ubuntu.com/about/release-cycle
 def is_ubuntu_lts_version(distro: Ubuntu) -> bool:
-    # asserts if not ubuntu OS object
-    version_info = distro.information.version
-    distro_age = _get_ubuntu_distro_age(distro)
-    is_even_year = (version_info.major % 2) == 0
-    is_april_release = version_info.minor == 4
-    is_within_support_window = distro_age <= 6
-    return is_even_year and is_april_release and is_within_support_window
+    major = str(distro.information.version.major)
+    minor = str(distro.information.version.minor)
+    # check for major+minor version match
+    return any(
+        [
+            major == x.split(".", maxsplit=1)[0] and minor == x.split(".")[1]
+            for x in _UBUNTU_LTS_VERSIONS
+        ]
+    )
+
+
+# check if it's a lts release outside the initial 2 year lts window
+def ubuntu_needs_backports(os: Ubuntu) -> bool:
+    return not is_ubuntu_latest_or_prerelease(os) and is_ubuntu_lts_version(os)
 
 
 def check_dpdk_support(node: Node) -> None:
@@ -391,16 +377,12 @@ def get_debian_backport_repo_args(os: Debian) -> List[str]:
     # ex: 'bionic-backports' or 'buster-backports'
     # these backport repos are available for the older OS's
     # and include backported fixes which need to be opted into.
-    # So filter for recent OS's and
-    # add the backports repo if it's available.
+    # So filter out recent OS's and
+    # add the backports repo for older ones, if it should be available.
     if not isinstance(os, Debian):
         return []
-    # workaround for recent Ubuntu distros which may have a
-    # backport repo defined but will raise an error if you
-    # try to use it explicitly for our packages.
-    if isinstance(os, Ubuntu) and (
-        is_ubuntu_latest_or_prerelease(os) or not is_ubuntu_lts_version(os)
-    ):
+    # don't enable backport args for releases which don't need/have them.
+    if isinstance(os, Ubuntu) and not ubuntu_needs_backports(os):
         return []
     repos = os.get_repositories()
     backport_repo = f"{os.information.codename}-backports"
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -250,6 +250,8 @@ class PackageManagerInstall(Installer):
             for os_package_check in self._os_dependencies.requirements:
                 if os_package_check.matcher(self._os) and os_package_check.packages:
                     self._os.uninstall_packages(os_package_check.packages)
+                    if os_package_check.stop_on_match:
+                        break
 
     # verify packages on the node have been installed by
     # the package manager
@@ -263,6 +265,8 @@ class PackageManagerInstall(Installer):
                     for pkg in os_package_check.packages:
                         if not self._os.package_exists(pkg):
                             return False
+                    if os_package_check.stop_on_match:
+                        break
         return True
 
     # installing dependencies is the installation in this case, so just return
--- a/microsoft/testsuites/dpdk/common.py
+++ b/microsoft/testsuites/dpdk/common.py
@@ -97,13 +97,13 @@ class GitDownloader(Downloader):
         # NOTE: fail on exists is set to True.
         # The expectation is that the parent Installer class should
         # remove any lingering installations
-        self._asset_path = self._node.tools[Git].clone(
+        self.asset_path = self._node.tools[Git].clone(
             self._git_repo,
             cwd=self._node.get_working_path(),
             ref=self._git_ref,
             fail_on_exists=True,
         )
-        return self._asset_path
+        return self.asset_path
 
 
 # parent class for tarball source installations
@@ -151,7 +151,7 @@ class TarDownloader(Downloader):
                 node_path=remote_path,
             )
         # create tarfile dest dir
-        self._asset_path = work_path.joinpath(
+        self.asset_path = work_path.joinpath(
             self.tar_filename[: -(len(tarfile_suffix))]
         )
         # unpack into the dest dir
@@ -161,7 +161,7 @@ class TarDownloader(Downloader):
             dest_dir=str(work_path),
             gzip=True,
         )
-        return self._asset_path
+        return self.asset_path
 
 
 class Installer:
@@ -183,7 +183,7 @@ class Installer:
     # setup the installation (install Ninja, Meson, etc)
     def _download_assets(self) -> None:
         if self._downloader:
-            self._asset_path = self._downloader.download()
+            self.asset_path = self._downloader.download()
         else:
             self._node.log.debug("No downloader assigned to installer.")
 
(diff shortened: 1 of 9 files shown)


microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -53,6 +50,14 @@
 # declare package dependencies for package manager DPDK installation
 DPDK_PACKAGE_MANAGER_PACKAGES = DependencyInstaller(
     requirements=[
+        # install linux-modules-extra-azure if it's available for mana_ib
+        # bionic will never have mana/mana_ib, so skip in that case
+        OsPackageDependencies(
+            matcher=lambda x: isinstance(x, Debian)
```
squirrelsc, line 57: Because it checks `bionic`, does it apply to Ubuntu only? Check distro version is easier read than code name.

mcgov: Hmm.. I added it only for Ubuntu and have tested it with Ubuntu and older debians already. I don't think Debian has the linux-modules-extra-azure package but it's probably safe to enable it for Ubuntu *.*.*

mcgov: Ok! Tests show kernel version works best. I'm keeping it 🙂

squirrelsc: You don't need to use "kernel" version. The codename is the same as "distro" version. But the combination of "Debian" and "Ubuntu codename/distro version" doesn't make sense.

microsoft/testsuites/dpdk/common.py
```diff
@@ -382,3 +385,40 @@ def unsupported_os_thrower(os: Posix) -> bool:
         os,
         message=("Installer did not define dependencies for this os."),
     )
+
+
+def get_debian_backport_repo_args(os: Debian) -> List[str]:
+    if not isinstance(os, Debian):
+        return []
+    if isinstance(os, Ubuntu) and (
+        is_ubuntu_latest_or_prerelease(os) or not is_ubuntu_lts_version(os)
```
squirrelsc: `is_ubuntu_latest_or_prerelease` is not stable, we should avoid using the concept. Please use a method to check the distro version like 20.04, 22.04.

microsoft/testsuites/dpdk/common.py
```diff
@@ -382,3 +385,40 @@ def unsupported_os_thrower(os: Posix) -> bool:
...
+    backport_repo = f"{os.information.codename}-backports"
+    if any([backport_repo in repo.name for repo in repos]):
+        return [f"-t {backport_repo}"]
+    return []
+
+
+def update_old_distros_to_latest_kernel(
+    node: Node, use_backports: bool = False
+) -> None:
+    assert isinstance(node.os, (Debian, Fedora, Suse))
```
squirrelsc: output the actual type.

microsoft/testsuites/dpdk/common.py
```diff
@@ -382,3 +385,40 @@ def unsupported_os_thrower(os: Posix) -> bool:
...
+    if (
+        isinstance(node.os, Debian)
+        and node.os.get_kernel_information().version < "6.5.0"
+    ):
+        package = "linux-azure"
```
squirrelsc, line 408: Does it work for Ubuntu only?

microsoft/testsuites/dpdk/common.py
```diff
@@ -382,3 +385,40 @@ def unsupported_os_thrower(os: Posix) -> bool:
...
+        return []
+    repos = os.get_repositories()
+    backport_repo = f"{os.information.codename}-backports"
+    if any([backport_repo in repo.name for repo in repos]):
+        return [f"-t {backport_repo}"]
+    return []
+
+
+def update_old_distros_to_latest_kernel(
+    node: Node, use_backports: bool = False
```
squirrelsc: `use_backports` is not used.

microsoft/testsuites/dpdk/common.py
```diff
@@ -382,3 +385,40 @@ def unsupported_os_thrower(os: Posix) -> bool:
...
+    ):
+        return []
+    repos = os.get_repositories()
+    backport_repo = f"{os.information.codename}-backports"
+    if any([backport_repo in repo.name for repo in repos]):
+        return [f"-t {backport_repo}"]
+    return []
+
+
+def update_old_distros_to_latest_kernel(
```
squirrelsc: When is the method needed? I think it's also unstable. It doesn't compatible with newer kernel versions. You can rename it to `update_kernel_from_repo`. But it's hard to check if the kernel is up-to-date or not. You can use a tool, and save the updated information into the tool. So the updates happen only once on a node.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: Right, I had used codename just because I was testing bionic, buster, focal and jammy. But the issue is really the kernel version, so it's bad to rely on codename or even distro version.
I should update the comment so it's not confusing.

microsoft/testsuites/dpdk/common.py
mcgov: > But it's hard to check if the kernel is up-to-date or not

It is not when you are using the package manager, this is the whole point of package managers. I'll rename the function so the scope is more clear.

microsoft/testsuites/dpdk/common.py
mcgov: Compromise: I will check the versions in that function.

microsoft/testsuites/dpdk/common.py
mcgov: No. This works for all DPDK suite supported distros.

microsoft/testsuites/dpdk/common.py
mcgov: I think I see what you mean in a general sense, but I'm not attempting to solve the problem of kernel updates happening before the test just yet. That will come in another patch later in the semester.

mcgov: Updated the function and added comments. This is just for dpdk/common.py so isn't for general use. mana_ib was added in 6.2 and backported to 5.15, so for many old images you must update to latest to get the driver. This ends up lining up nicely with images in general. The older ones need you to pull latest updates before starting the test. To install the latest DPDK packages, you need to update the base kernel first and then update the rest of the packages later. So this function updates the kernel package first in initialize_node_resources before we start installing dependencies. It ends up being much cleaner than trying to do this inline in the tools.

microsoft/testsuites/dpdk/common.py
squirrelsc: I don't understand why the latest os are not supported. Is it a temp period, or should be always true? Please add comments in the code.

microsoft/testsuites/dpdk/common.py
```diff
@@ -382,3 +366,45 @@ def unsupported_os_thrower(os: Posix) -> bool:
...
+
+
+# NOTE: mana_ib was added in 6.2 and backported to 5.15
+# this ends up lining up with kernels that need to be updated before
+# starting our DPDK tests. This function is not meant for general use
+# outside of the DPDK suite.
+def update_kernel_from_repo(node: Node) -> None:
+    assert isinstance(node.os, (Debian, Fedora, Suse)), (
+        f"DPDK test does not support OS type: "
+        f"{node.os.information.vendor} {node.os.information.release}"
```
squirrelsc: Just use `type(node.os)` only. If it's Windows, there is no `information` object.

microsoft/testsuites/dpdk/common.py
mcgov: I think you're misunderstanding the core concept. The 'backports' args are only needed for old OS's (specifically debian distros). So this function decides whether to add an extra argument to `apt _____` to use the backports repository over the regular one. This is to get more recent versions of packages that are backported to work with the older OS. It's an optional flag but we want it because there are bug fixes in there.

This isn't deciding whether to support an OS or not, it's just deciding whether to add this extra backports repo argument to apt.

mcgov: The repo isn't available for latest or non-lts verisons of debian distros, also. So we don't include it in those cases. They don't need backported stuff because they're already using latest.

mcgov: I think I also included that filter for latest because occasionally there _would_ be a jammy-backports repo but it would raise an error if you tried to use it explicitly. I'll comment this function more, it's not clear enough currently.

mcgov: Commented. I don't want to mess with this more than I need to, it was already checked in and is needed for bionic/buster. We're just moving the function in this PR.

squirrelsc: The "latest" is not a stable concept. I understand like 20.04/22.04 needs backports to support dpdk. But it doesn't mean in 2026, the 24.04 needs backports. It may make the test cases suddenly fail after 2026/1/1.

So, the logic should be like `if version <= 22.04, then install new kernels`.

mcgov: Well... wait. I switched this code to _not_ use a date comparison because you commented that latest wasn't a stable concept. Can I switch it back? Previously the function looked at whether the OS was over two years old (since that's the release schedule Canonical uses)

squirrelsc: I just check the method name, haven't check content of the method. That's great to see the date check was removed! How about use a name like `ubuntu_need_backports`? The "latest" is confusing, because 24.04 is not "latest" at 2026, but it doesn't need the backports.

mcgov: I kind of want to revert this to what is was before. It's unrelated to the refactor. I get that we always can improve the code but... this was already reviewed and approved. Let's change it in it's own PR if we're going to change it.

mcgov: > The "latest" is not a stable concept. I understand like 20.04/22.04 needs backports to support dpdk. But it doesn't mean in 2026, the 24.04 needs backports. It may make the test cases suddenly fail after 2026/1/1.
> 
> So, the logic should be like `if version <= 22.04, then install new kernels`.

No, this _isn't_ right. There are unsupported interrum releases of ubuntu where we don't want to try to use the backport repo. 21.11 or 21.04 don't have backport repos because they're not supported past the release of 22.04.

This is why this is written this way, checking if it's LTS or latest. This also _isn't_ the function where the kernel is updated. It is just to get this specific backport argument.

mcgov: undoing my previous change. The check was fine and working before

squirrelsc: > No, this _isn't_ right. There are unsupported interrum releases of ubuntu where we don't want to try to use the backport repo. 21.11 or 21.04 don't have backport repos because they're not supported past the release of 22.04.

I'm trying to provide logic about LTS versions, you can still have other logic to check if it's LTS or not. But for LTS versions, it shouldn't rely on actual years. It's highly like to be failed in 2026.

mcgov: > I just check the method name, haven't check content of the method. That's great to see the date check was removed! How about use a name like `ubuntu_need_backports`? The "latest" is confusing, because 24.04 is not "latest" at 2026, but it doesn't need the backports.

Ah ok this makes a lot of sense. Will change the name and comment things.

I need to disagree about using the date though. That check is exactly to get around having a hard-coded 'latest' that we have to update every two years. Until Canonical changes their release schedule and decides to release an LTS version on an odd year, or move to a rolling release or something, the function will continue to work as intended. We need to enable the repo for releases which are LTS versions that are older than two years (the initial support window). This function does that.

see https://ubuntu.com/about/release-cycle

It's been like this as far back as I can find and there aren't plans to change it as far as I can tell.

I've restructured the code and added more thorough comments to explain things. Let's finish this review up?

squirrelsc: Let me try to discuss from a question. Today, we know the 24.04 doesn't need backports to work well. How can we know it needs backports, and there are backports for it at 2026/1/1?

mcgov: fixed

squirrelsc:
@LiliDeng LGTM

mcgov:
One line bugfix, changing package name for package manager version check to the generic 'rdma-core' instead of 'libibuverbs'. Source install version check is the same.

LiliDeng:
@mcgov found regression during testing, provide the details offline

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -299,7 +304,7 @@ def download(self) -> PurePath:
             self._git_ref = git.get_tag(
                 self.asset_path, filter_=r"^v.*"  # starts w 'v'
             )
-            git.checkout(self._git_ref, cwd=self.asset_path)
+            git.checkout(self._git_ref, cwd=self.asset_path, force_run=False)
```
squirrelsc: What's the reason to set it to false? Git operations are not to get information, it should always run to avoid confusing.

mcgov: Yeah you're right, this will break if you check out a branch, check out another branch, then try to check out the first branch again.

I'm trying to work around an issue with the Git tool. The Git tool creates a branch based on the *run id* if you don't give it a branch name, so will throw if you try to checkout any branches twice on a node without providing some branch name. You can't just pass the branch name as 'branch name' since it uses -b to create a new branch.

Maybe it makes more send to add an option to not try to create a temporary branch?

squirrelsc: You can specify the `checkout_branch` to avoid generate a local name for branch.

mcgov: This is the problematic code:
```
        result = self.run(
            f"checkout {ref}",
            force_run=True,
            cwd=cwd,
            no_info_log=True,
            no_error_log=True,
        )

        result = self.run(
            f"checkout -b {checkout_branch}",
            force_run=True,
            cwd=cwd,
            no_info_log=True,
            no_error_log=True,
        )
```

mcgov: Oh duh I can just delete the temp branch if it exists before creating it again.

mcgov: fixed

LiliDeng (approved):

PR #3460, opened 2024-10-03
parse_literal transformer

kamalca:
Can convert string inputs into python literals, including ints, floats, dicts, lists, sets, etc.

It functions similarly to the existing to_list transformer, but it is more flexible in its applications.

### DIFF
--- a/lisa/transformers/parse_literal.py
+++ b/lisa/transformers/parse_literal.py
@@ -1,0 +1,48 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from ast import literal_eval
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import schema
+from lisa.transformer import Transformer
+
+PARSE_LITERAL = "parse_literal"
+
+
+@dataclass_json
+@dataclass
+class ParseLiteralTransformerSchema(schema.Transformer):
+    # items to expand from string to list
+    items: Dict[str, str] = field(default_factory=dict)
+
+
+class ParseLiteralTransformer(Transformer):
+    """
+    This transformer transfer string to a literal
+    (list, set, dictionary, int, float, etc.)
+    """
+
+    @classmethod
+    def type_name(cls) -> str:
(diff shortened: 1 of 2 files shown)


kamalca:
Motivated by #3455 This transformer would allow a dictionary of tags to be passed in such as "{'FirstPartyUsage': '/SyntheticLoad'}" then be converted into a python literal before being passed into the runbook.

squirrelsc:
For #3455, how about use like `ip_service_tag_name` and `ip_service_tag_value`? It's easier to use than a transformer. Or it can be a dict in runbook itself. The value shouldn't be dynamic, so write them in runbook should be good enough.

lisa/transformers/parse_literal.py
```diff
@@ -0,0 +1,48 @@
...
+
+from lisa import schema
+from lisa.transformer import Transformer
+
+PARSE_LITERAL = "parse_literal"
+
+
+@dataclass_json
+@dataclass
+class ToListTransformerSchema(schema.Transformer):
```
squirrelsc: The name shouldn't be "ToList".

lisa/transformers/parse_literal.py
```diff
@@ -0,0 +1,48 @@
...
+from ast import literal_eval
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Type
+
+from dataclasses_json import dataclass_json
+
+from lisa import schema
+from lisa.transformer import Transformer
+
+PARSE_LITERAL = "parse_literal"
```
squirrelsc, line 13: `literal_eval` may be more consistent. How about update existing `eval` transformer to use `literal_evel`?

kamalca:
> For #3455, how about use like `ip_service_tag_name` and `ip_service_tag_value`? It's easier to use than a transformer. Or it can be a dict in runbook itself. The value shouldn't be dynamic, so write them in runbook should be good enough.

Technically there can be any number of ip tags. Passing in two variables is an option, but it would limit the use cases to a single ip tag.

I also believe that the functionality of converting a string to a literal has many uses beyond my current motivation.

kamalca:
Closing in favor of #3478

PR #3463, opened 2024-10-03
check platforms to prevent exception by mismatched feature types on different platforms

squirrelsc:
See details on commits.


LiliDeng:
for some cases, it is unexpected to be skipped.
```
AzureImageStandard.verify_default_targetpw: SKIPPED  capability cannot support some of requirements, requirement: 'allowed:True,items:[ready,azure]'capability: 'allowed:True,items:[azure]', 
```

LiliDeng (approved):

PR #3464, opened 2024-10-07
Add support for Private endpoint to SMB file share

shekharsorot:
This PR ports some code from class NFS to class AzureFileShare.
This enables creation of private endpoint, DNS zone and records which is necessary for scenarios where policy does not allows access to SMB share over public endpoint.

note: private endpoint must be created after shares are created. There is no workaround at the moment to do otherwise via API at the moment.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -172,9 +172,9 @@ class StartStop(AzureFeatureMixin, features.StartStop):
         node_info = self._node.connection_info
         node_info[constants.ENVIRONMENTS_NODES_REMOTE_PUBLIC_ADDRESS] = public_ip
         node_info[constants.ENVIRONMENTS_NODES_REMOTE_ADDRESS] = private_ip
-        node_info[constants.ENVIRONMENTS_NODES_REMOTE_USE_PUBLIC_ADDRESS] = (
-            platform._azure_runbook.use_public_address
-        )
+        node_info[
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -3273,6 +3276,55 @@ def create_file_share(
                 resource_group_name=resource_group_name,
                 log=self._log,
             )
+
+        # Create file private endpoint, always after all shares are created
+        if enable_private_endpoint is True:
```
squirrelsc: `if enable_private_endpoint:`

shekharsorot: updated code

lisa/sut_orchestrator/azure/features.py
```diff
@@ -3263,6 +3264,8 @@ def create_file_share(
             allow_shared_key_access=allow_shared_key_access,
         )
 
+        # The SMB endpoint https://<share>.file.core.windows.net
+        # will auto point to <share>.privatelink.file.core.windows.net
```
squirrelsc: The added comments is always true, or only when `enable_private_endpoint` is True? If it happens when `enable_private_endpoint` is true, move the code comment to right place.

shekharsorot: updated comment to ensure clarity on condition and sequence

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3469, opened 2024-10-09
neseted: create L2 image during runtime

LiliDeng:


microsoft/testsuites/nested/common.py
```diff
@@ -27,6 +31,54 @@
 )
 
 
+def _create_cloud_init_iso(
```
squirrelsc, line 34: Is this method used by qemu only? If so, does it need to enable for Azure tests?

LiliDeng: it can be used by others I think, btw, there is a similar implement in libvirt platform. https://github.com/microsoft/lisa/blob/a75eda8fc5ef7d16e969d4a9b45f657cbd9261da/lisa/sut_orchestrator/libvirt/platform.py#L929

squirrelsc (approved):

PR #3473, opened 2024-10-14
Fix script file transformer and openssl command issues 

LiliDeng:


microsoft/testsuites/nested/common.py
```diff
@@ -38,7 +38,19 @@ def _create_cloud_init_iso(
     password: str,
     host_name: str = "l2vm",
 ) -> str:
-    cmd_result = host.execute(f"openssl passwd -6 {password}", sudo=True, shell=True)
+    cmd_result = host.execute(
+        f"openssl passwd -6 {password}",
+        sudo=True,
+        shell=True,
+    )
+    if cmd_result.exit_code != 0:
```
squirrelsc, line 62: Add comments above to explain the expected errors.

squirrelsc: Please give an example error message. So if there are other non-zero exit code, others know the `if` branch is not designed for the new situation.

LiliDeng: updated

lisa/tools/qemu.py
```diff
@@ -76,7 +80,7 @@ def create_vm(
             # e.g. QEMU emulator version 1.5.3 (qemu-kvm-1.5.3-175.el7_9.6)
             # on centos 7.9
             try_pcid_flag = self.node.execute(
-                f"{self._qemu_command} -cpu host,pcid=no", sudo=True
+                f"timeout 20 {self._qemu_command} -cpu host,pcid=no", sudo=True
```
squirrelsc, line 85: Why add time out here? Add comments above.

LiliDeng: added

squirrelsc (approved):
Please don't add more changes into this PR...

LiliDeng:
> Please don't add more changes into this PR...

sure, I have tested the last commit, and it is related change, so I appended it here.

PR #2370, opened 2022-11-07
[improvement] add restriction for cvm and tvm

LiliDeng:

PR #3474, opened 2024-10-14
Add CVM v6 SKUs

kamalca:
CVM v6 SKUs need to be added to the list of CVM supporting SKUs.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -2299,25 +2299,12 @@ class SecurityProfile(AzureFeatureMixin, features.SecurityProfile):
                 and raw_capabilities.get("TrustedLaunchDisabled", "False") == "False"
             ):
                 capabilities.append(SecurityProfileType.SecureBoot)
-        # https://learn.microsoft.com/en-us/azure/confidential-computing/confidential-vm-overview # noqa: E501
-        if cvm_value and resource_sku.family.casefold() in [
-            "standarddcasv5family",
-            "standarddcadsv5family",
-            "standardecasv5family",
-            "standardecadsv5family",
-            "standarddcev5family",
-            "standarddcedv5family",
-            "standardecev5family",
-            "standardecedv5family",
-        ]:
+
+        if cvm_value and cvm_value.casefold() == "snp":
             capabilities.append(SecurityProfileType.CVM)
 
-        if cvm_value == "TDX" and resource_sku.family.casefold() in [
-            "standarddcev5family",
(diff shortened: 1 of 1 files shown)


kamalca:
@squirrelsc @LiliDeng I don't like this method of hard coding SKU families. I don't understand the original justification for it. Is there a better solution?

kamalca:
I see this commit https://github.com/microsoft/lisa/commit/313cefd01a18cfa295190cd6a1ee4fa4661a3106 is where the check for CVM SKU families is added, but #2370 provides no justification for why we need to check this. We should trust the capability tagging on the SKUs and we should avoid creating more work for ourselves needing to manually update SKU family lists.

LiliDeng:
Do you know the Capabilities names for CVM and Stateless? if yes, we can update coed to reply on them.

LiliDeng:
> I see this commit [313cefd](https://github.com/microsoft/lisa/commit/313cefd01a18cfa295190cd6a1ee4fa4661a3106) is where the check for CVM SKU families is added, but #2370 provides no justification for why we need to check this. We should trust the capability tagging on the SKUs and we should avoid creating more work for ourselves needing to manually update SKU family lists.

Do you know the Capabilities names for CVM and Stateless? if yes, we can update coed to reply on them.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2300,24 +2300,10 @@ def create_setting(
...
-            "standarddcasv5family",
-            "standarddcadsv5family",
-            "standardecasv5family",
-            "standardecadsv5family",
-            "standarddcev5family",
-            "standarddcedv5family",
-            "standardecev5family",
-            "standardecedv5family",
-        ]:
+        if cvm_value.casefold() == "snp":
```
squirrelsc: Update comments above.

kamalca: Do you just want me to remove the link to CVM overview?

squirrelsc: The original link was the source of vm size families. It looks the content was moved, so the link should be removed too.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

LiliDeng:
@kamalca could you please make sure you have tested this change? Thanks.

kamalca:
> @kamalca could you please make sure you have tested this change? Thanks.

I tested SNP and TDX deployment. I am currently running all of my CVM v6 validation with this change and have seen no issues.

PR #3475, opened 2024-10-16
Fix kdump test for Mariner

0xba1a:
- Update correct config file when crash dump needs to be stored in ephemeral disk
- Add more nr_cpus when the test machine has more than 56 cores (more than on xeon socket)
- Add an option not to kill a process when it timeouts
- Add a provision to print additional information before hitting panic

### DIFF
--- a/lisa/tools/kdump.py
+++ b/lisa/tools/kdump.py
@@ -619,7 +619,7 @@ class KdumpCBLMariner(KdumpBase):
             sudo=True,
         )
         # Set mariner_2_initrd_use_suffix. Otherwise it will replace
-        # the original initrd file which will cause a boot-loop
+        # the original initrd file which will cause a reboot-loop
         sed.substitute(
             match_lines="mariner_2_initrd_use_suffix",
             regexp="#mariner_2_initrd_use_suffix",
@@ -669,7 +669,6 @@ class KdumpCBLMariner(KdumpBase):
         If the system memory size is bigger than 1T, the default size of /var/crash
         may not be enough to store the dump file, need to change the dump path
         """
-        kdump_conf = "/etc/kdump.conf"
         self.node.execute(
             f"mkdir -p {dump_path}",
             expected_exit_code=0,
@@ -679,6 +678,7 @@ class KdumpCBLMariner(KdumpBase):
         )
         self.dump_path = dump_path
         # Change dump path in kdump conf
+        kdump_conf = "/etc/kdump.conf"
         sed = self.node.tools[Sed]
         sed.substitute(
             match_lines="^path",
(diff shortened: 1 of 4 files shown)


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -318,6 +327,7 @@ def _kdump_test(self, node: Node, log_path: Path, log: Logger) -> None:
                 shell=True,
                 sudo=True,
                 timeout=10,
+                kill_on_timeout=False
```
squirrelsc: Use the timeout tool, it can kill the process after timeout.

lisa/tools/kdump.py
```diff
@@ -451,6 +454,10 @@ def check_crashkernel_loaded(self, crashkernel_memory: str) -> None:
         # Check if memory is reserved for crash kernel
         self._check_crashkernel_memory_reserved()
 
+    def print_additional_info_before_panic(self) -> None:
```
squirrelsc: It can be a private method of Mariner, and always print it out. There is no shared logic, so a method on super class is not needed.

microsoft/testsuites/kdump/kdumpcrash.py
0xba1a: The `execute` function tries to kill the process if the process times-out. In this case the process that gets timed-out is `echo c > /proc/sysrq-trigger`. Trying to kill it makes no sense. So, eventually it goes for a 10-minute unnecessary wait (the timeout of `kill` function is around 10 minute).

With this `kill_on_timeout=False`, `execute` will not try to kill the process after it times-out.

lisa/tools/kdump.py
0xba1a: I'm sorry that I couldn't understand this comment.

I want `print_additional_info_before_panic` to be called just before the panic is induced. The panic is induced in the object of `node` not the object of `kdump`. So, calling a private function of `kdump` may not be possible in the call-chain. So, if I remove the method from super class, it may throw exception when other distros are being tested.

Please let me know if my understanding is wrong.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: LISA has a tool called `Timeout`. You can use it with your commands, so the kill behavior can be controlled in this tool, and it doesn't need to add another argument in the `execute` method.

lisa/tools/kdump.py
squirrelsc: I double checked the current implementation of kdump, it looks need refactor to make it clearer. So far, your way may be the best way to implement it. Please use the method name like `capture_info`.

microsoft/testsuites/kdump/kdumpcrash.py
0xba1a: I've updated to use `execute_async` and removed `kill_on_timeout` parameter itself.

lisa/tools/kdump.py
0xba1a: Updated the name of the function

squirrelsc:
@LiliDeng LGTM, @0xba1a please fix check errors.

0xba1a:
> @LiliDeng LGTM, @0xba1a please fix check errors.

Fixed the errors. Can you please check now?

LiliDeng (approved):

PR #3477, opened 2024-10-17
Raise exception when getting timeout for running cmd to detect OS

lubaihua33:
For some images, like sciencelogicinc1622565452194, middleware publisher's images, running some basic commands will be time out. This is due to the image not allow to run commands or in bad state. 

In _get_detect_string, there are more than 10 commands to detect the OS. If every command gets time out (10 min), the case might last longer than 1 hour. 

The case should be go to the end immediately if the system can't run any commands.

In this PR, add timeout=30 for every command in _get_detect_string, and raise an exception when gets timeout to avoid more timeout commands.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -59,6 +59,17 @@ if TYPE_CHECKING:
 _get_init_logger = partial(get_logger, name="os")
 
 
+def get_matched_os_name(cmd_result: ExecutableResult, pattern: Pattern[str]) -> str:
+    # Check if the command is timedout. If it is, the system might be in a bad state
+    # Then raise an exception to avoid more timedout commands.
+    if cmd_result.is_timeout:
+        raise LisaTimeoutException(
+            f"Command timed out: {cmd_result.cmd}. "
+            "Please check if the system allows to run command from remote."
+        )
+    return get_matched_str(cmd_result.stdout, pattern)
+
+
 class CpuArchitecture(str, Enum):
     X64 = "x86_64"
     ARM64 = "aarch64"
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -215,53 +215,96 @@ def capture_system_information(self, saved_path: Path) -> None:
     @classmethod
     def _get_detect_string(cls, node: Any) -> Iterable[str]:
         typed_node: Node = node
-        cmd_result = typed_node.execute(cmd="lsb_release -d", no_error_log=True)
-        yield get_matched_str(cmd_result.stdout, cls.__lsb_release_pattern)
+        cmd_result = typed_node.execute(
+            cmd="lsb_release -d", no_error_log=True, timeout=30
```
squirrelsc: I'm general ok with the change, but the timeout can be set to 60 seconds for some extremely conditions. The key thing is to raise the exception, if first timeout met.

lisa/operating_system.py
```diff
@@ -215,53 +215,96 @@ def capture_system_information(self, saved_path: Path) -> None:
...
+        )
+        yield cls._get_matched_str_from_cmd_result(
```
squirrelsc: Use it as a function, not in the class. Because the function is not related to the class. Try shorter name like `get_matched_os_name`?

lisa/operating_system.py

lisa/operating_system.py
lubaihua33: I wanted to add function in the same file lisa\util\__init__.py as get_matched_str. But it has a circle import due to adding ExecutableResult. Then I added get_matched_os_name in operating_system.py.

squirrelsc: You can add the method in this file, but not in this class.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3479, opened 2024-10-17
Marked Swap test case to be skipped

r-dailey:
Verify_swap test case needs to be skipped as waagent does not currently endorse support for freebsd and while it works for most things they do not guarantee all features work correctly or that they will support them on freebsd.

### DIFF
--- a/microsoft/testsuites/core/storage.py
+++ b/microsoft/testsuites/core/storage.py
@@ -197,7 +197,11 @@ class Storage(TestSuite):
         3. Verify that truth value in step 1 and step 2 match.
         """,
         priority=1,
-        requirement=simple_requirement(supported_platform_type=[AZURE]),
+        requirement=simple_requirement(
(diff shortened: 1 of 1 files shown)


LiliDeng:
could you please double check, swap case pass in thefreebsdfoundation freebsd-14_0 14_0-release-amd64-gen2-zfs 14.0.0

r-dailey:
> could you please double check, swap case pass in thefreebsdfoundation freebsd-14_0 14_0-release-amd64-gen2-zfs 14.0.0

Per the documentation here: https://learn.microsoft.com/en-us/azure/virtual-machines/extensions/agent-linux#:~:text=The%20Agent%20works,able%20to%20help. FreeBSD might work but is not an endorsed distro so I think this test that relies on Waagent should be skipped. If you think that we should keep it and file an issue on waagent instead please let me know.

LiliDeng (approved):

PR #3481, opened 2024-10-18
Dpdk: add package info to test results

mcgov:
Swap to adding dpdk/rdma/nic info to dpdk test results for both functional and performance tests.

This makes triaging issues much quicker and enables rich coverage and test status reporting, it's not possible to graph (dpdk_version x hardware_type) or (rdma_version x dpdk_version x hardware type) without this annotation.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -836,7 +836,7 @@ class Debian(Linux):
         if not match:
             raise LisaException(
                 f"Could not parse version info: {version_str} "
-                "for package {package_name}"
+                f"for package {package_name}"
             )
         self._node.log.debug(f"Attempting to parse version string: {version_str}")
         version_info = self._get_version_info_from_named_regex_match(
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -532,7 +532,7 @@ class Posix(OperatingSystem, BaseClassMixin):
     def _get_version_info_from_named_regex_match(
         self, package_name: str, named_matches: Match[str]
     ) -> VersionInfo:
-        essential_matches = ["major", "minor", "build"]
+        essential_matches = ["major", "minor"]
 
         # verify all essential keys are in our match dict
         assert_that(
@@ -547,7 +547,6 @@ class Posix(OperatingSystem, BaseClassMixin):
             patch_match = "0"
         major_match = named_matches.group("major")
         minor_match = named_matches.group("minor")
-        build_match = named_matches.group("build")
         major, minor, patch = map(
             int,
             [major_match, minor_match, patch_match],
@@ -752,8 +751,8 @@ class Debian(Linux):
     _debian_version_splitter_regex = re.compile(
         r"([0-9]+:)?"  # some examples have a mystery number followed by a ':' (git)
         r"(?P<major>[0-9]+)\."  # major
-        r"(?P<minor>[0-9]+)[\-\.]"  # minor
-        r"(?P<patch>[0-9]+)"  # patch
+        r"(?P<minor>[0-9]+)"  # minor
+        r"([\-\.](?P<patch>[0-9]+))?"  # patch
         r"(?:-)?(?P<build>[a-zA-Z0-9-_\.~+]+)"  # build
         # '-' is added after minor and made optional before build
         # due to the formats like 23.11-1build3
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -750,6 +750,10 @@ class Debian(Linux):
         r"Package: ([a-zA-Z0-9:_\-\.]+)\r?\n"  # package name group
         r"Version: ([a-zA-Z0-9:_\-\.~+]+)\r?\n"  # version number group
     )
+    # ex: 3.10
+    # ex: 3.10.5-git
+    # ex: 3.10-5git3
+    # ex: 3.10.1-build3
     _debian_version_splitter_regex = re.compile(
         r"([0-9]+:)?"  # some examples have a mystery number followed by a ':' (git)
         r"(?P<major>[0-9]+)\."  # major
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -554,10 +554,9 @@ class Posix(OperatingSystem, BaseClassMixin):
         build_match = named_matches.group("build")
         log_message = (
             f"Found {package_name} version "
-            f"{major_match}.{minor_match}.{patch_match}"
+            f"major:{major_match} minor:{minor_match} "
+            f"patch:{patch_match} build:{build_match}"
         )
-        if build_match:
-            log_message += f"-{build_match}"
         self._node.log.debug(log_message)
         return VersionInfo(major, minor, patch, build=build_match)
 
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -1979,9 +1979,9 @@ class Suse(Linux):
     _suse_version_splitter_regex = re.compile(
         r"([0-9]+:)?"  # some examples have a mystery number followed by a ':' (git)
         r"(?P<major>[0-9]+)\."  # major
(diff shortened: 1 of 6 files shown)


lisa/operating_system.py
```diff
@@ -532,7 +532,7 @@ def _get_package_information(self, package_name: str) -> VersionInfo:
     def _get_version_info_from_named_regex_match(
         self, package_name: str, named_matches: Match[str]
     ) -> VersionInfo:
-        essential_matches = ["major", "minor", "build"]
+        essential_matches = ["major", "minor"]
```
squirrelsc, line 535: Why remove the build part? It may affect all other distros.

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -71,7 +71,7 @@
         ),
         OsPackageDependencies(
             matcher=lambda x: isinstance(x, Suse)
-            and parse_version(x.information.release) == "15.5.0",
+            and bool(parse_version(x.information.release) == "15.5.0"),
```
squirrelsc, line 74: It doesn't need the `bool` conversion.

lisa/operating_system.py
mcgov: It's not removing it, it's making it optional since sometimes it's missing, similar to patch. If you check the processing function below: build can be None. It was supposed to be optional, so this is more of a bug fix. I've run into package versions like '4.10' so making patch and build optional is needed.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: image

mcgov: parse_version can apparently return None; casting to bool is still safe since None == "15.5.0" will return False (in the event the release info is garbage).

I'm pretty sure that situation would break code elsewhere before we hit this check, so this is really just to satisfy the type checker.

lisa/operating_system.py
```diff
@@ -547,16 +547,18 @@ def _get_version_info_from_named_regex_match(
...
             [major_match, minor_match, patch_match],
         )
         build_match = named_matches.group("build")
-        self._node.log.debug(
+        log_message = (
             f"Found {package_name} version "
-            f"{major_match}.{minor_match}.{patch_match}-{build_match}"
+            f"{major_match}.{minor_match}.{patch_match}"
         )
+        if build_match:
```
squirrelsc: The log message can be clearer like:

f"Found {package_name} version "
            f"major: {major_match}, minor: {minor_match}, patch: {patch_match}, build: {build_match}"

mcgov: Nice, I agree! Thanks

lisa/operating_system.py
```diff
@@ -752,9 +753,9 @@ class Debian(Linux):
     _debian_version_splitter_regex = re.compile(
         r"([0-9]+:)?"  # some examples have a mystery number followed by a ':' (git)
         r"(?P<major>[0-9]+)\."  # major
-        r"(?P<minor>[0-9]+)[\-\.]"  # minor
-        r"(?P<patch>[0-9]+)"  # patch
-        r"(?:-)?(?P<build>[a-zA-Z0-9-_\.~+]+)"  # build
+        r"(?P<minor>[0-9]+)"  # minor
+        r"([\-\.](?P<patch>[0-9]+))?"  # patch
+        r"((?:-)?(?P<build>[a-zA-Z0-9-_\.~+]+))?"  # build
         # '-' is added after minor and made optional before build
```
squirrelsc, line 763: Add new examples above

mcgov: Added! Thanks

microsoft/testsuites/dpdk/dpdksuite.py
```diff
@@ -362,10 +386,18 @@ def verify_dpdk_multiprocess(
         ),
     )
     def verify_dpdk_sriov_rescind_failover_receiver(
-        self, environment: Environment, log: Logger, variables: Dict[str, Any]
+        self,
+        environment: Environment,
+        log: Logger,
+        variables: Dict[str, Any],
+        result: TestResult,
```
squirrelsc: result is not used in this method.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -1085,4 +1099,80 @@ def create_l3fwd_rules_files(
...
+        map(str, [NicType.CX3, NicType.CX4, NicType.CX5, NicType.MANA])
+    )
+    found_nic_types = ",".join(map(str, [x.device_id for x in devices]))
+    node.log.debug(
+        "Unknown NIC hardware was detected during DPDK test case. "
+        f"Expected one of: {known_nic_types}. Found {found_nic_types}. "
+    )
+    # this is just a function for annotating a result, so don't assert
+    # if there's
+    return ""
```
squirrelsc: It can returns the `found_nic_types`. it's easier to troubleshoot.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -1085,4 +1099,80 @@ def create_l3fwd_rules_files(
...
+def annotate_dpdk_test_result(
+    test_kit: DpdkTestResources, test_result: TestResult, log: Logger
+) -> None:
+    dpdk_version = None
+    rdma_version = None
+    nic_hw = None
+    try:
+        dpdk_version = test_kit.testpmd.get_dpdk_version()
+    except AssertionError:
+        test_kit.node.log.debug("Could not fetch DPDK version info.")
```
squirrelsc: Log the error info, so it's easier for troubleshooting.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -1085,4 +1099,80 @@ def create_l3fwd_rules_files(
...
+    try:
+        rdma_version = test_kit.rdma_core.get_installed_version()
+    except AssertionError:
+        test_kit.node.log.debug("Could not fetch RDMA version info.")
+    try:
+        nic_hw = get_node_nic_short_name(test_kit.node)
+    except AssertionError:
+        test_kit.node.log.debug("Could not fetch NIC short name.")
+    if dpdk_version:
```
squirrelsc: Those lines can be in `try` blocks, the code is clearer.

microsoft/testsuites/dpdk/dpdkutil.py
```diff
@@ -1085,4 +1099,80 @@ def create_l3fwd_rules_files(
...
+    if rdma_version:
+        test_result.information["rdma_version"] = str(rdma_version)
+    if nic_hw:
+        test_result.information["nic_hw"] = nic_hw
+
+
+def annotate_dpdk_send_receive_test_result(
+    test_kit: DpdkTestResources, environment: Environment, log: Logger
+) -> None:
+    test_result = environment.source_test_result
```
squirrelsc: Don't use the `source_test_result`, it may be incorrect, when an env is reused. Use the test result which is passed from test methods.

microsoft/testsuites/dpdk/dpdksuite.py
mcgov: Nice catch, thank you!

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: I love this idea

microsoft/testsuites/dpdk/dpdkutil.py
mcgov: Agreed!

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3484, opened 2024-10-21
sut_orchestrator: hyperv: don't automatically add disks

anirudhrb:
The create_vm() method of the HyperV tool, automatically finds offline disks and adds them to the VM. This behavior is not required for hyperv platform. So, introduce a parameter to control it and modify hyperv platform to disable that behavior.

### DIFF
--- a/lisa/tools/hyperv.py
+++ b/lisa/tools/hyperv.py
@@ -72,6 +72,7 @@ class HyperV(Tool):
         generation: int = 1,
         cores: int = 2,
         memory: int = 2048,
+        attach_offline_disks: bool = True,
         com_ports: Optional[Dict[int, str]] = None,
         secure_boot: bool = True,
         stop_existing_vm: bool = True,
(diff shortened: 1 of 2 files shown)


pupacha (dismissed):

smit-gardhariya (dismissed):

gamora12 (approved):

lisa/tools/hyperv.py
```diff
@@ -72,6 +72,7 @@ def create_vm(
         generation: int = 1,
         cores: int = 2,
         memory: int = 2048,
+        add_disks: bool = True,
```
LiliDeng: from the description, the default value should be false?

lisa/tools/hyperv.py
```diff
@@ -119,20 +120,21 @@ def create_vm(
...
+            disk_info = powershell.run_cmdlet(
+                "(Get-Disk | Where-Object {$_.OperationalStatus -eq 'offline'}).Number",
                 force_run=True,
             )
+            matched = re.findall(r"\d+", disk_info)
+            disk_numbers = [int(x) for x in matched]
+            for disk_number in disk_numbers:
+                self._run_hyperv_cmdlet(
+                    "Add-VMHardDiskDrive",
+                    f"-VMName {name} -DiskNumber {disk_number} -ControllerType 'SCSI'",
```
LiliDeng, line 134: is the ControllerType always SCSI?

lisa/tools/hyperv.py
```diff
@@ -119,20 +120,21 @@ def create_vm(
...
-        disk_numbers = [int(x) for x in matched]
-        for disk_number in disk_numbers:
-            self._run_hyperv_cmdlet(
-                "Add-VMHardDiskDrive",
-                f"-VMName {name} -DiskNumber {disk_number} -ControllerType 'SCSI'",
-                extra_args=extra_args,
+        # add disks if requested
+        if add_disks:
+            disk_info = powershell.run_cmdlet(
+                "(Get-Disk | Where-Object {$_.OperationalStatus -eq 'offline'}).Number",
```
squirrelsc, line 126: How is the number gotten? Is it possible to set to number = 0 to avoid add disks?

lisa/tools/hyperv.py
anirudhrb: The default value should be True in order to not break the existing callers of this function.

lisa/tools/hyperv.py
anirudhrb: This is existing code. I just moved it inside the `if` block.

lisa/tools/hyperv.py
anirudhrb: This PowerShell command lists all offline disks and gets the count. `Number` is that count.

In our server, we have some extra disks lying around but we don't want to add them to the VMs. At least, not automatically.

lisa/tools/hyperv.py
squirrelsc: ok, the parameter name could be like `attach_offline_disks`.

anirudhrb: Renamed.

pupacha (approved):

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3489, opened 2024-10-28
DPDK: Add 32bit test

mcgov:
Add a test which compiles DPDK as a 32bit application before run. Implements basic build and send_receive tests.

Required a little fiddling with meson and pkgconfig to ensure the 32bit libraries are used when needed.

### DIFF
--- a/lisa/tools/meson.py
+++ b/lisa/tools/meson.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 from pathlib import PurePath
-from typing import cast
+from typing import Dict, Optional, cast
 
 from semver import VersionInfo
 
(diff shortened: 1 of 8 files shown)


mcgov:
@squirrelsc  I'm getting a ton of mypy errors which are due to the old version we're using :\ 
Any objections to upgrading to a slightly more recent version?

squirrelsc:
> @squirrelsc I'm getting a ton of mypy errors which are due to the old version we're using :\ Any objections to upgrading to a slightly more recent version?

We don't have a plan to upgrade mypy yet, it's not a trivial and safe work.

mcgov:
> > @squirrelsc I'm getting a ton of mypy errors which are due to the old version we're using :\ Any objections to upgrading to a slightly more recent version?
> 
> We don't have a plan to upgrade mypy yet, it's not a trivial and safe work.

Definitely not trivial, tons of errors specifically in the aws and azure sut_orchestrator code 😵‍💫

mcgov:
@squirrelsc I've worked out the mypy issues. Should be all good now.

squirrelsc:
@LiliDeng LGTM

PR #3492, opened 2024-10-30
mdatp check: add mdatp health invocation to check for onboarding

mcgov:
After talking with the defender team, I'm adding a second check which uses `mdatp health` to check for the current onboarding state. This avoids the problem of relying entirely on the well known file path to find the onboarding info.

### DIFF
--- a/microsoft/testsuites/mdatp/check-mdatp.sh
+++ b/microsoft/testsuites/mdatp/check-mdatp.sh
@@ -113,5 +113,20 @@ if [ -f "$MDATP_OPT_DIR/mdatp_onboard.json" ]; then
     EXIT_CODE=$EXIT_ONBOARD_INFO_FOUND
 fi
 
+# special log line if mdatp installed and reports it is onboarded
+MDATP_ORG_ID=$(\
+    sudo command -v mdatp \
+    && sudo mdatp health \
+    | grep --fixed-strings 'org_id:' \
+    | cut -f 2 -d ':' \
+    | tr -d '[:blank:][:punct:]' \
+)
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/mdatp/check-mdatp.sh
```diff
@@ -113,5 +113,20 @@ if [ -f "$MDATP_OPT_DIR/mdatp_onboard.json" ]; then
     EXIT_CODE=$EXIT_ONBOARD_INFO_FOUND
 fi
 
+# special log line if mdatp installed and reports it is onboarded
+MDATP_ORG_ID=$(\
+    sudo command -v mdatp \
```
LiliDeng: the script is executed under sudo, no need sudo here.
I got error when run this command, but actually `command` exists
```
sudo command -v mdatp \
    && sudo mdatp health \
    | grep --fixed-strings 'org_id:' \
    | cut -f 2 -d ':' \
    | tr -d '[:blank:][:punct:]' \
sudo: command: command not found
```

mcgov: What distro was this on?

mcgov: I had used sudo because we also recommend this script for use by end users who hit this problem. If they want to see what went wrong and reproduce the failure, they can run it locally. 

It doesn't explain the command -v thing though. Let me know the distro and I'll investigate.

LiliDeng: ubuntu 2204

LiliDeng (approved):

PR #3493, opened 2024-11-01
Marked Lun based testcases to be skipped as FreeBSD does not use LUN

r-dailey:
LUN is not a feature on FreeBSD so this testcase that is focused on using LUN is not applicable and needs to be skipped. Tested on all the current images and it is being skipped correctly.


LiliDeng:
@r-dailey we can use an alternative way to detect lun numbers on FreeBSD

PR #3494, opened 2024-11-01
add logic to accept only good arm64 images

feng-j678:
[x] add logic to include only support arm64 images
[x] block debian debian-10 10-backports 0.20230917.1506

### DIFF
--- a/microsoft/testsuites/vm_extensions/linux_patch_extension.py
+++ b/microsoft/testsuites/vm_extensions/linux_patch_extension.py
@@ -15,7 +15,15 @@ from lisa import (
     simple_requirement,
 )
 from lisa.base_tools.service import Service
-from lisa.operating_system import BSD, SLES, CBLMariner, CentOs, Debian, Oracle, Ubuntu
+from lisa.operating_system import (
+    BSD,
+    SLES,
+    CBLMariner,
+    CentOs,
+    Debian,
+    Oracle,
+    Ubuntu,
+)
 from lisa.sut_orchestrator import AZURE
 from lisa.sut_orchestrator.azure.common import (
     get_compute_client,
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/linux_patch_extension.py
```diff
@@ -27,21 +27,51 @@
...
-def _verify_unsupported_images(node: Node) -> None:
+def _verify_supported_arm64_images(
+    node: Node, log: Logger, full_version: str, arch: str
+) -> None:
+    # LPE current supported images for arm64
+    supported_versions_arm64 = {
+        # major minor gen
+        CentOs: ["7-9 2"],
+        Oracle: ["8-10 2", "9-4 2"],
+        Ubuntu: ["20-4 2"],
```
squirrelsc, line 38: Please try use VersionInfo like "20.4.2", it they can be compared directly.

microsoft/testsuites/vm_extensions/linux_patch_extension.py
```diff
@@ -51,6 +81,14 @@ def _verify_unsupported_images(node: Node) -> None:
                 _unsupported_image_exception_msg(node)
 
 
+def _get_os_full_version(node: Node) -> str:
```
squirrelsc: The version is used to compare directly, so it doesn't need to be converted to a string. You can use it like `node.os.information.version == "11.7.1"` or `node.os.information.version >= "11.7.1"`.

feng-j678: I've tried this approach it's not working well, images are failing and i dislike how it doesn't pick up the gen info and it add many decimals. for example this image openlogic centos 7_9-arm64 7.9.2024020802, the full version is 7.9.2009, so  i revert it back to use _get_os_full_version, i like this approach better because its easy to follow/understand the code:

image

squirrelsc (approved):

PR #3496, opened 2024-11-05
Update junit.py

shekharsorot:
Context : for cases such as XFSTest, iPerf and others where each test case will have multiple nested cases controlled directly by the tool being called, we needed a way to cleanly pass subtests with messages, stacks and other information to Junit XML for parsing and possible injection into databases,

Added 3 new parameters.
    include_passed_messages:  ( Bool ) Include message and stacktrace in Junit XML output. Defaults to False
    include_failed_messages: ( Bool ) Include message and stacktrace in Junit XML output. Defaults to True
    include_skipped_messages: ( Bool ) Include message and stacktrace in Junit XML output. Defaults to True

Enhancements : Stacktrace now exists as an XML subproperty, making XML much cleaner.
If "message" and "stacktrace" properties are empty  in the TestResultMessage object, its relevant subXML objects will not be added, makes XML parsing & viewing cleaner.

note: Ideally this change should not break existing Junit parsing scripts, but its expected that end users are using properly traversing XML tree to ensure they do not end up with an exeception when say..."message" attribute is missing because there was actually no message passed by the TestResultMessage object.

how to call ? Example code from runbook--

notifier:
  - type: html
  - type: env_stats
  - type: junit 
    name: junit.lisa.xml 
    include_subtest: true 
    include_passed_messages: false 
    include_failed_messages: true 
    include_skipped_messages: true

### DIFF
--- a/lisa/notifiers/junit.py
+++ b/lisa/notifiers/junit.py
@@ -312,8 +312,7 @@ class JUnit(Notifier):
             failure = ET.SubElement(testcase, "failure")
             if runbook.include_failed_messages:
                 failure.attrib["message"] = message.message
-                # failure.text = message.stacktrace
-                if str(message.stacktrace) != "":
+                if message.stacktrace:
                     failure.attrib["stacktrace"] = str(message.stacktrace)
             testsuite_info.failed_count += 1
 
--- a/lisa/notifiers/junit.py
+++ b/lisa/notifiers/junit.py
@@ -30,10 +30,6 @@ class JUnitSchema(schema.Notifier):
     include_subtest: bool = True
     # show passed case 'message' and 'stacktraces'
     include_passed_messages: bool = False
-    # show 'message' for failed cases and 'stacktrace'
-    include_failed_messages: bool = True
-    # show 'message' for skipped cases
-    include_skipped_messages: bool = True
 
 
 class _TestSuiteInfo:
@@ -310,26 +306,24 @@ class JUnit(Notifier):
 
         if message.status == TestStatus.FAILED:
             failure = ET.SubElement(testcase, "failure")
-            if runbook.include_failed_messages:
-                failure.attrib["message"] = message.message
-                if message.stacktrace:
-                    failure.attrib["stacktrace"] = str(message.stacktrace)
+            failure.attrib["message"] = message.message
+            failure.text = message.stacktrace
             testsuite_info.failed_count += 1
 
         elif message.status in [TestStatus.SKIPPED, TestStatus.ATTEMPTED]:
             skipped = ET.SubElement(testcase, "skipped")
-            if runbook.include_skipped_messages:
-                skipped.attrib["message"] = message.message
-                if str(message.stacktrace) != "":
-                    skipped.attrib["stacktrace"] = str(message.stacktrace)
+            skipped.attrib["message"] = message.message
+            if message.stacktrace:
+                skipped.text = str(message.stacktrace)
         # Only add XML sub element if include_passed_messages is True in runbook
-        # By default, its assumed subtest is a pass
+        # By default, its assumed subtest is a pass as per Junit schema
         elif message.status == TestStatus.PASSED and runbook.include_passed_messages:
             passed = ET.SubElement(testcase, "passed")
             passed.attrib["message"] = message.message
             # passed.text = message.stacktrace
-            if str(message.stacktrace) != "":
-                passed.attrib["stacktrace"] = str(message.stacktrace)
+            if message.stacktrace:
+                passed.text = str(message.stacktrace)
+
         testsuite_info.test_count += 1
 
         # Write out current results to file.
--- a/lisa/notifiers/junit.py
+++ b/lisa/notifiers/junit.py
@@ -28,6 +28,12 @@ class JUnitSchema(schema.Notifier):
     path: str = "lisa.junit.xml"
     # respect the original behavior, include subtest by default
     include_subtest: bool = True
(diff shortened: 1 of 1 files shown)


lisa/notifiers/junit.py
```diff
@@ -302,18 +310,27 @@ def _add_test_case_result(
...
         if message.status == TestStatus.FAILED:
             failure = ET.SubElement(testcase, "failure")
-            failure.attrib["message"] = message.message
-            failure.text = message.stacktrace
-
+            if runbook.include_failed_messages:
+                failure.attrib["message"] = message.message
+                # failure.text = message.stacktrace
+                if str(message.stacktrace) != "":
+                    failure.attrib["stacktrace"] = str(message.stacktrace)
```
squirrelsc: Is this defined by junit schema?

lisa/notifiers/junit.py
```diff
@@ -28,6 +28,12 @@ class JUnitSchema(schema.Notifier):
     path: str = "lisa.junit.xml"
     # respect the original behavior, include subtest by default
     include_subtest: bool = True
+    # show passed case 'message' and 'stacktraces'
+    include_passed_messages: bool = False
+    # show 'message' for failed cases and 'stacktrace'
+    include_failed_messages: bool = True
```
squirrelsc: When should it be set to False? If no scenario, it can be included always.

lisa/notifiers/junit.py
```diff
@@ -28,6 +28,12 @@ class JUnitSchema(schema.Notifier):
     path: str = "lisa.junit.xml"
     # respect the original behavior, include subtest by default
     include_subtest: bool = True
+    # show passed case 'message' and 'stacktraces'
+    include_passed_messages: bool = False
```
squirrelsc, line 32: It doesn't need to be False by default, so the logic is more consistent for different types.

lisa/notifiers/junit.py
shekharsorot: This is compliant with Jnuit schema.  
Its ideally expected to have stacktrace even for successful cases. in our case, we include the  <case>.dmesg output from results folder here
For XFSTesting,  passed messages will include run time and stacktrace will include diff of test data + dmesg dump

for skipped cases, message will include actual reason for test being skipped. Stack trace can result if the case was skipped due to missing dependencies or conditions.

lisa/notifiers/junit.py
shekharsorot: Default behavior is to include messages and stacktraces for Failed tests/subcases.
Ideally we want messages for skipped tests.
Passed tests/cases should be optional and extension can bundle additional information in stacktrace

lisa/notifiers/junit.py
shekharsorot: I marked it false to ensure current output is not modified in any way. This is because i expect users to be parsing this output in their implementation of LISA. 
Marking passed and skipped flag as false by default allows us to maintain some level of compat with existing implementation while allowing customization for new implementations.

lisa/notifiers/junit.py
squirrelsc: I don't see the scenarios, which want set failed/skippped to false, and set passed to true. So the three flags are not needed. The logic always include stackstrace for failed and skipped, but never for passed.

lisa/notifiers/junit.py
```diff
@@ -302,18 +310,27 @@ def _add_test_case_result(
 
         if message.status == TestStatus.FAILED:
             failure = ET.SubElement(testcase, "failure")
-            failure.attrib["message"] = message.message
-            failure.text = message.stacktrace
-
+            if runbook.include_failed_messages:
+                failure.attrib["message"] = message.message
+                # failure.text = message.stacktrace
```
squirrelsc: Remove the commented line.

lisa/notifiers/junit.py
```diff
@@ -302,18 +310,27 @@ def _add_test_case_result(
...
+                if str(message.stacktrace) != "":
```
squirrelsc: `if message.stacktrace:` is enough for None, "".

lisa/notifiers/junit.py
```diff
@@ -302,18 +310,27 @@ def _add_test_case_result(
...
+                    failure.attrib["stacktrace"] = str(message.stacktrace)
             testsuite_info.failed_count += 1
```
squirrelsc, line 311: Why does it need force convert to string?

shekharsorot: For some weird reason, linting is complaining about typecasting. I forced a string typecast just to be sure.

lisa/notifiers/junit.py
shekharsorot: updated

lisa/notifiers/junit.py
shekharsorot: updated

lisa/notifiers/junit.py
shekharsorot: understood. I will modify code to only implement logic for successful test cases.

lisa/notifiers/junit.py
squirrelsc: If no scenario needs the stack trace for passed test cases, it doesn't need the flag.

shekharsorot:
Closing PR, will implement this as a private Patch instead as the changes are unique for XFStest instead of larger suite of existing test cases

squirrelsc:
@shekharsorot It's close to complete, please try to contribute to upstream to avoid extra cost on maintaining backports in future.

PR #3499, opened 2024-11-07
Addition of "Diff" to LISA tools

shekharsorot:
Adding simple functionality for "diff" command in Linux based test runners. Diff allows for comparison of two files or directories.
Tool can be called via class "Diff"
Currently only "comparefiles" function is implemented,
TODO: comparedirectory is pending

### DIFF
--- a/lisa/tools/__init__.py
+++ b/lisa/tools/__init__.py
@@ -27,6 +27,7 @@ from .curl import Curl
 from .date import Date
 from .df import Df
 from .dhclient import Dhclient
+from .diff import Diff
 from .dmesg import Dmesg
 from .dnsmasq import Dnsmasq
 from .docker import Docker
@@ -140,6 +141,7 @@ __all__ = [
     "Date",
     "Df",
     "Dhclient",
+    "Diff",
     "Dmesg",
     "Dnsmasq",
     "Docker",
--- a/lisa/tools/diff.py
+++ b/lisa/tools/diff.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 from pathlib import PurePath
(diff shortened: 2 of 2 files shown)


lisa/tools/__init__.py
```diff
@@ -1,262 +1,264 @@
...
-    "YumConfigManager",
-    "Vdsotest",
-    "VcRunner",
-    "VcTargetInfo",
-    "VirtualClientTool",
-    "Who",
-    "Whoami",
-    "Wsl",
-]
+# Copyright (c) Microsoft Corporation.
```
squirrelsc, line 1: What's change the whole file? Please double check it.

lisa/tools/diff.py
```diff
@@ -0,0 +1,73 @@
...
+    def comparefiles(
+        self,
+        src: PurePath,
+        dest: PurePath,
+        cwd: Optional[PurePath] = None,
+        sudo: bool = False,
+        timeout: int = 600,
+    ) -> str:
+        cmd = f"{self.node.get_str_path(src)} {self.node.get_str_path(dest)}"
+        result = self.run(
```
squirrelsc, line 61: What's the output looks like? How the result will be used?

lisa/tools/diff.py
```diff
@@ -0,0 +1,73 @@
...
+    @property
+    def command(self) -> str:
+        return "diff"
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _check_exists(self) -> bool:
+        return True
```
squirrelsc, line 50: Don't overwrite this method, let LISA check if it exists in default paths.

shekharsorot:
Due to a checksum issue, _init_.py is being treated as new file. Will re-raise PR with correction

PR #3500, opened 2024-11-08
added filter for benign mellanox warning on freebsd

r-dailey:
Added a filter for a benign warning on freebsd to resolve an issue with the verify_boot_error_warnings test case.


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -243,6 +243,11 @@ class AzureImageStandard(TestSuite):
         ),
         # pam_unix,pam_faillock
         re.compile(r"^(.*pam_unix,pam_faillock.*)$", re.M),
+        # benign warning from mellanox on freebsd
```
squirrelsc: typo, and add some details, why it can be ignored.

r-dailey: Ok I will update the comment.

squirrelsc (approved):

PR #3501, opened 2024-11-08
Add tool "diff" for Linux and Unix

shekharsorot:
Diff is used in Linux/Unix to compare two files, two directories or two streams.

In our case, Diff is used in XFStest since we need to compare *.full and *.out.full files for failed subtests to know what has went wrong. 

TODO : Currently this tool is only for Linux and Unix, Windows alternative is FC which needs to be implemented.
Currently only "comparefiles" is implemented. as we come across requirements to compare directories and streams, we will add support for them.

            Args:
                src (PurePath): The source file or directory.
                dest (PurePath): The destination file or directory.
                cwd (Optional[PurePath]): The current working directory for the command.
                Defaults to None.
                sudo (bool): Whether to run the command with sudo. Defaults to False.
                timeout (int): The timeout for the command in seconds. Defaults to 600.
            Returns:
                str: The output of the diff command.

            Raises:
                AssertionError: If the exit code of the diff command is not 0 or 1.

# 0 means no changes, 1 means changes detected. 2 means general failure.

Edit: Due to a checksum issue, Github was treating _init_.py as new file. This has been fixed with a new commit and a new PR is being raised.

### DIFF
--- a/lisa/tools/__init__.py
+++ b/lisa/tools/__init__.py
@@ -27,6 +27,7 @@ from .curl import Curl
 from .date import Date
 from .df import Df
 from .dhclient import Dhclient
+from .diff import Diff
 from .dmesg import Dmesg
 from .dnsmasq import Dnsmasq
 from .docker import Docker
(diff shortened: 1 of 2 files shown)


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3502, opened 2024-11-09
Availability: Skip bad configurations

kamalca:
Mismatches in Availability Set requirements and capabilities may be due to test case requirements, so we can skip the test instead of failing.

### DIFF
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -2495,7 +2495,8 @@ class Availability(AzureFeatureMixin, features.Availability):
                 "is not supported in the current configuration. "
                 "Please select one of "
                 f"{[type.value for type in settings.availability_type.items]}. "
-                "Or consider changing the disk type or location."
+                "The supported availability types is affected by disk type, "
+                "location, and test case requirements."
             )
 
     @classmethod
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -2477,11 +2477,12 @@ class Availability(AzureFeatureMixin, features.Availability):
                     for zone in params.availability_zones
                     if zone in settings.availability_zones
                 ]
-                assert params.availability_zones, (
-                    "Invalid zones provided. "
-                    "This SKU in this location supports zones: "
-                    f"{settings.availability_zones}. "
-                )
+                if not params.availability_zones:
+                    raise SkippedException(
+                        "Invalid zones provided. "
+                        "This SKU in this location supports zones: "
+                        f"{settings.availability_zones}. "
+                    )
             elif settings.availability_zones:
                 params.availability_zones = [settings.availability_zones.items[0]]
 
@@ -2489,15 +2490,16 @@ class Availability(AzureFeatureMixin, features.Availability):
                 type.value for type in AvailabilityType
             ], ("Not a valid Availability Type: " f"{params.availability_type}")
 
-            assert (
+            if not (
                 AvailabilityType(params.availability_type) in settings.availability_type
-            ), (
-                f"Availability Type "
-                f"'{params.availability_type}' "
-                "is not supported in the current configuration. Please select one of "
-                f"{[type.value for type in settings.availability_type.items]}. "
-                "Or consider changing the disk type or location."
-            )
+            ):
+                raise SkippedException(
+                    f"Availability Type "
+                    f"'{params.availability_type}' "
+                    "is not supported in the current configuration. Please select one of "
+                    f"{[type.value for type in settings.availability_type.items]}. "
+                    "Or consider changing the disk type or location."
+                )
 
         # If the availability_type is still set to Default, then
         # resolve the default without considering capabilities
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/features.py
```diff
@@ -2517,9 +2519,10 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
             if "platformUpdateDomainCount" not in params.availability_set_properties:
                 params.availability_set_properties["platformUpdateDomainCount"] = 1
         elif params.availability_type == AvailabilityType.AvailabilityZone:
-            assert (
-                params.availability_zones
-            ), "Availability Zone is selected, but no zone was provided."
+            if not params.availability_zones:
+                raise SkippedException(
+                    "Availability Zone is selected, but no zone was provided."
```
squirrelsc: Please refine the messages clearer. Where the zone is selected, and where it's not provided?

lisa/sut_orchestrator/azure/features.py
```diff
@@ -2477,27 +2477,29 @@ def on_before_deployment(cls, *args: Any, **kwargs: Any) -> None:
...
-                "Or consider changing the disk type or location."
-            )
+            ):
+                raise SkippedException(
+                    f"Availability Type "
+                    f"'{params.availability_type}' "
+                    "is not supported in the current configuration. Please select one of "
+                    f"{[type.value for type in settings.availability_type.items]}. "
+                    "Or consider changing the disk type or location."
+                )
```
adityagesh: Do you think we should add this statement "consider changing the disk type or location.", in error log only if is_ultra_disk is true?
I feel it would be misleading otherwise.

kamalca: I don't think it is an issue. The only case you hit this error message without having ultra disk enabled is if you explicitly state zone redundancy in the runbook. Zone redundancy is not really used outside of ultra disk deployment.

Did you have a specific scenario in mind where this message would be exceedingly misleading?

lisa/sut_orchestrator/azure/features.py
kamalca: Fixed

kamalca:
@squirrelsc Any suggestions on how to resolve `'Availability.on_before_deployment' is too complex` flake error?

I considered making a common function, called something like `assert_or_skip(condition: Any, message: str)` that would work like an assert statement but raise a SkippedException instead of an assert error. This would remove a level of logic branching and resolve the error.

What are your thoughts? There are complex interactions between ultradisk, availability types, and maximize capability, so it is not easy to break this logic into small pieces.

lisa/sut_orchestrator/azure/features.py
squirrelsc: How about the source of first half sentence? where the zone is selected from?

kamalca: Technically it could have been from a test case requirement but very likely it was also from the runbook.

With these error messages, it is hard to think of every single combination of (runbook, test case, vm capabilities) × (disk type, maximize capability, availability type, tip session, zones)

I wrote this logic to verify that all the setting look correct and have appropriate defaults, agnostic of where the requirements are coming from. Because most of the requirement merging happens before this logic in `on_before_deployment`

lisa/sut_orchestrator/azure/features.py
squirrelsc: For the error message, it's read by human. It doesn't need to be accurate or an exact result. So, you can try to provide most or all possible reasons to help troubleshooting.

lisa/sut_orchestrator/azure/features.py
kamalca: I have changed the error message to include explicit troubleshooting steps.

squirrelsc:
> @squirrelsc Any suggestions on how to resolve `'Availability.on_before_deployment' is too complex` flake error?
> 
> I considered making a common function, called something like `assert_or_skip(condition: Any, message: str)` that would work like an assert statement but raise a SkippedException instead of an assert error. This would remove a level of logic branching and resolve the error.
> 
> What are your thoughts? There are complex interactions between ultradisk, availability types, and maximize capability, so it is not easy to break this logic into small pieces.

I thought I replied, but lost. Instead of the `assert_or_skip`, I suggest splitting a meaningful sub-method for this method. The sub-method may be called like `_check_configuration_conflict`, and move all related content here. The `assert_and_skip` is easy to misuse, because most skip should happen in test case level. Also, there are many `assert` methods, like `is_true`, `is_empty`, and so on. It's not worth to wrap all of them.

squirrelsc:
@LiliDeng LGTM

lisa/sut_orchestrator/azure/features.py
adityagesh: I plan to add AvailabilityType.NoRedudancy as a test case requirement for Hibernation.
If a Runbook has case Hibernation, but has availability_type set to AvailabilitySet, it would raise this SkippedException, but would have nothing to do with the disk

LiliDeng: @kamalca @adityagesh let know your discussion results, then I can proceed this PR, thanks.

kamalca: I have made the language more clear. @LiliDeng you can proceed, thank you.

LiliDeng (approved):

PR #3503, opened 2024-11-10
Fixed VM Snapshot Extension test failure - Expected <10> to be less than <10>, but was not

vipunj:
Test VmSnapsotLinuxBVTExtension.verify_vmsnapshot_extension was failing due AssertionError: [Restore point creation failed.] Expected <10> to be less than <10>, but was not.
For this, have removed redundant count increment from the func _verify_vmsnapshot_extension and now it is passing

### DIFF
--- a/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
+++ b/microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
@@ -265,7 +265,6 @@ class VmSnapsotLinuxBVTExtension(TestSuite):
                 else:
                     raise e
             time.sleep(1)
-            count = count + 1
         assert_that(count, "Restore point creation failed.").is_less_than(10)
 
     def _find_extension_dir(self, node: Node) -> str:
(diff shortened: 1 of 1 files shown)


arisettisanjana:
the error is because of repeated retries that were happening because of failures, and since it exceeded 10 retries we see that error.

Now we will have to check why were the failures occurring/happening

LiliDeng (dismissed):

LiliDeng:
@arisettisanjana please fix the CI check error

microsoft/testsuites/vm_extensions/vmsnapshot_extension.py
```diff
@@ -265,7 +265,7 @@ def _verify_vmsnapshot_extension(
                 else:
                     raise e
             time.sleep(1)
-            count = count + 1
+            count +=1
```
LiliDeng: count += 1

LiliDeng (approved):

PR #3504, opened 2024-11-11
 Fix GPU CUDA Installation in Ubuntu

adityagesh:
Update keyring file version
Fetch cuda-drivers package version automatically and install.

### DIFF
--- a/lisa/features/gpu.py
+++ b/lisa/features/gpu.py
@@ -16,7 +16,12 @@ from lisa.feature import Feature
 from lisa.operating_system import CBLMariner, CpuArchitecture, Oracle, Redhat, Ubuntu
 from lisa.tools import Lspci, Lsvmbus, NvidiaSmi
 from lisa.tools.lspci import PciDevice
-from lisa.util import LisaException, SkippedException, constants
+from lisa.util import (
+    LisaException,
+    MissingPackagesException,
+    SkippedException,
+    constants,
+)
 
 FEATURE_NAME_GPU = "Gpu"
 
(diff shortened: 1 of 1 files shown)


adityagesh:
Tested using Ubuntu 2004, 2204, 2404

lisa/features/gpu.py
```diff
@@ -257,16 +265,18 @@ def _install_cuda_driver(self, version: str) -> None:
...
-                package_version = cuda_package_versions.get(release, "545")
-                self._node.os.install_packages(f"cuda-drivers-{package_version}")
+                result = self._node.execute(
+                    f"apt search {cuda_package_name}", sudo=True
+                )
+                available_versions = cuda_drivers_package_pattern.findall(result.stdout)
+
+            if available_versions:
+                # Sort versions and select the highest one
+                highest_version = max(available_versions)
```
kanchansenlaskar, line 278: for now this will work fine, but with double digits lexicographical max may not work the way we want, perhaps you can check this?

kanchansenlaskar: ignore this, this should be fine

lisa/features/gpu.py
```diff
@@ -207,6 +212,9 @@ def _install_cuda_driver(self, version: str) -> None:
             self._node.os.install_packages(install_packages, signed=False)
 
         elif isinstance(self._node.os, Ubuntu):
+            cuda_package_name = "cuda-drivers"
+            cuda_drivers_package_pattern = re.compile(r"cuda-drivers-(\d+)")
```
kanchansenlaskar: better to match the whole string (or the whole version part at least), for more robust logic

lisa/features/gpu.py
adityagesh: The packages does not use semantic versioning. 
The examples of packages are "cuda-drivers-550", "cuda-drivers-555", "cuda-drivers-560" and so on

https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/x86_64/

lisa/features/gpu.py
adityagesh: Updated, testing.

adityagesh: tested

LiliDeng (approved):

PR #3505, opened 2024-11-11
DPDK: Fix source for tarball

mcgov:
Fix the way the TarDownloader handles dowloading files and identifying their filenames. This required fixing a bug in the Tar tool where Linux would not infer the filename in the default case like the Windows version does. Previously, providing a directory and not a filename would result in the tool identifying an entire directory of files as the filename _when the Wget tool fetches a cached result_.

This PR fixes that issue by allowing the tool to infer the filename, this fixes the ability to use the Wget tool without force_run when using a default filename.

### DIFF
--- a/lisa/base_tools/wget.py
+++ b/lisa/base_tools/wget.py
@@ -26,7 +26,7 @@ class Wget(Tool):
 
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._url_file_cache: Dict[str, str] = dict()
-        return super()._initialize(*args, **kwargs)
+        super()._initialize(*args, **kwargs)
 
     @property
     def can_install(self) -> bool:
--- a/lisa/base_tools/wget.py
+++ b/lisa/base_tools/wget.py
@@ -1,5 +1,5 @@
 import re
-from typing import TYPE_CHECKING, Optional, Tuple, Type
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Type
 from urllib.parse import urlparse
 
 from retry import retry
@@ -24,6 +24,10 @@ class Wget(Tool):
     def command(self) -> str:
         return "wget"
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._url_file_cache: Dict[str, str] = dict()
+        return super()._initialize(*args, **kwargs)
+
     @property
     def can_install(self) -> bool:
         return True
@@ -45,7 +49,18 @@ class Wget(Tool):
         force_run: bool = False,
         timeout: int = 600,
     ) -> str:
+        cached_filename = self._url_file_cache.get(url, None)
+        if cached_filename:
+            if force_run:
+                del self._url_file_cache[url]
+            else:
+                return cached_filename
+
         is_valid_url(url)
+
+        if not filename:
+            filename = urlparse(url).path.split("/")[-1]
+            self._log.debug(f"filename is not provided, use {filename} from url.")
 
         file_path, download_path = self._ensure_download_path(file_path, filename)
 
@@ -84,6 +99,7 @@ class Wget(Tool):
                     f" stdout: {command_result.stdout}"
                     f" templog: {temp_log}"
                 )
+            self.node.tools[Rm].remove_file(log_file, sudo=sudo)
         else:
             download_file_path = download_path
 
@@ -91,7 +107,7 @@ class Wget(Tool):
             raise LisaTimeoutException(
                 f"wget command is timed out after {timeout} seconds."
             )
-        actual_file_path = self.node.execute(
+        ls_result = self.node.execute(
             f"ls {download_file_path}",
             shell=True,
             sudo=sudo,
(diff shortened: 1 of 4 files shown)


lisa/base_tools/wget.py
```diff
@@ -47,6 +47,10 @@ def get(
     ) -> str:
         is_valid_url(url)
 
+        if not filename:
```
squirrelsc, line 61: When is it needed? The `wget` should give a name by default. The web downloader obtains the name from a specific header firstly. If it doesn't specified, then use the last part of url. But below logic doesn't consider the header and it includes all parts after last `/`, and it may include the query string, which shouldn't be a part of the file name.

The following logic parses the file name from download log. Please check what's wrong without force_run, and propose a fix.

mcgov: The issue is that Wget.get does not cache results the same way that Tool.run does. Maybe we can add caching. Just so you're aware, that same filename default is already present in the Windows version of the tool. I had hoped that since it was already present it wouldn't be an issue to add it to Linux too. We can fix both.

The bug is this:

Without force_run, Wget.get will:
1. check if the directory exists. (but not check the filename arg)
2. remove the directory if overwrite is present.
1. download the file with Wget.run(... force_run=force_run)
3. check the log file for the filename if the log file exists
4. deletes the log file.
5. uses LS to try to find the filename if the log file isn't present. It uses the filename and path arugments in this case. But both are optional.
6. returns the filename

All of those steps happen every time, even if force_run is False. If force_run is False, the log file is not available since it's deleted. So if filename is not present and filepath *is* present (the default case fills it in as the working path), the tool will return the entire contents of the directory as the filename. So LISA ends up trying to use .../working_path/dpdk\trdma-core\t as a tar file.

mcgov: > But below logic doesn't consider the header and it includes all parts after last `/`, and it may include the query string, which shouldn't be a part of the file name.

This also isn't entirely accurate. urlparse seperates the path and query. An example would be urlparse(url).path will return just `/path/to/file.tar.gz` from `https://[place]/path/to/file.tar.gz?query_string=query`

squirrelsc: I see, please cache a map with (url, returned file name). If force_run is false, check the dict firstly.

mcgov: Thanks! Added

mcgov:
Note: testing in progress.

lisa/base_tools/wget.py
```diff
@@ -24,6 +24,10 @@ class Wget(Tool):
     def command(self) -> str:
         return "wget"
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self.__filename_result_cache: Dict[str, str] = dict()
```
squirrelsc: 1. This variable should be used by subclasses, so one underscore is right.
2. Call it like `_url_file_map`. It explains the variable names.

lisa/base_tools/wget.py
```diff
@@ -45,8 +49,26 @@ def get(
         force_run: bool = False,
         timeout: int = 600,
     ) -> str:
+        if not force_run:
```
squirrelsc: The logic could be simpler.

```python
cached_filename = self.__filename_result_cache.get(url, None)
if cached_filename:
    if force_run:
        del self.__filename_result_cache[url]
    else:
        return cached_filename

File: lisa/tools/tar.py [hunk 4]
```diff
@@ -48,6 +49,21 @@ def extract(
         if strip_components:
             # optionally strip N top level components from a tar file
             tar_cmd += f" --strip-components={strip_components}"
+
+        if skip_old_files:
```
squirrelsc: How about remove existing extracted files always, or always overwrites? The skip-old-files is hard to use with other tools right.

lisa/base_tools/wget.py
mcgov: I would use 'map' for static key-value associations, this is a dynamic dict where we save old results and flush them occasionally... It's a classic cache, right?

lisa/base_tools/wget.py
mcgov: This is good stuff 🙂 I'll fix

lisa/tools/tar.py
mcgov: No, we specifically don't want to overwrite the old files here. 

We _have_ to assume that changes have been applied, we will wipe out patches that have been applied to source files in the output directory otherwise. We can't just delete the output directory and start over every time, since this will potentially wipe out a build directory.

squirrelsc: ok, it makes sense. But the name `skip_old_files` is not accurate. Please use `skip_exist_files` for the method argument.

lisa/base_tools/wget.py
squirrelsc: It's ok to call it like `_url_file_cache`. It includes both key/value names.

lisa/tools/tar.py
mcgov: I feel strongly that options should have the same name as the flag, as long as we're just doing a passthrough like this. It allows people to look up the man page for the flag without having to check the LISA tool's source to see what the option resolves to.

squirrelsc: It depends on the design principle. If the goal is to be same as original commands, you are right. But LISA abstracts the commands for better understanding. In this case, you explained the term "old" in the code, but I read the original document to understand it finally. If it's called "exist", it would be clearer. When the tool is implemented by BSD, Windows and other OS, it also makes sense.

lisa/base_tools/wget.py
```diff
@@ -24,6 +24,10 @@ class Wget(Tool):
     def command(self) -> str:
         return "wget"
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        self._url_file_cache: Dict[str, str] = dict()
+        return super()._initialize(*args, **kwargs)
```
squirrelsc: It doesn't need `return`, because the function signature return value is `None`. It means nothing to return.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3506, opened 2024-11-12
Add b4 Tool

adityagesh:
Add b4 tool to download upstream patches.
Add am method to git in order to apply this patch.

The usage of these will be raised as a separate PR. A new modifier will be added to kernel_installer to modify code using b4.

### DIFF
--- a/lisa/tools/git.py
+++ b/lisa/tools/git.py
@@ -191,14 +191,25 @@ class Git(Tool):
         cwd: pathlib.PurePath,
         patches: pathlib.PurePath,
     ) -> None:
-        result = self.run(
-            f"apply {patches}",
-            shell=True,
-            cwd=cwd,
-            force_run=True,
-            no_info_log=True,
-            no_error_log=True,
-        )
+        file_extension = patches.suffix
+        if file_extension == ".mbx":
+            result = self.run(
+                f"am {patches}",
+                shell=True,
+                cwd=cwd,
+                force_run=True,
+                no_info_log=True,
+                no_error_log=True,
+            )
+        else:
+            result = self.run(
+                f"apply {patches}",
+                shell=True,
+                cwd=cwd,
+                force_run=True,
+                no_info_log=True,
+                no_error_log=True,
+            )
         result.assert_exit_code(message=f"failed on applying patches. {result.stdout}")
 
     def list_tags(self, cwd: pathlib.PurePath) -> List[str]:
--- a/lisa/tools/__init__.py
+++ b/lisa/tools/__init__.py
@@ -16,6 +16,7 @@ from lisa.base_tools import (
 )
 
 from .aria import Aria
+from .b4 import B4
 from .blkid import Blkid
 from .bzip2 import Bzip2
 from .cargo import Cargo
@@ -130,6 +131,7 @@ from .wsl import Wsl
 __all__ = [
     "AptAddRepository",
     "Aria",
+    "B4",
     "Blkid",
     "Bzip2",
     "Cargo",
--- a/lisa/tools/b4.py
+++ b/lisa/tools/b4.py
@@ -1,0 +1,62 @@
+import pathlib
+import re
+from typing import List, Type
+
+from lisa.executable import Tool
+from lisa.operating_system import Debian
+from lisa.tools.git import Git
+from lisa.tools.python import Pip
+from lisa.util import LisaException, find_group_in_lines
+
+
+class B4(Tool):
+    # Output log is of the form
+    # git am /mnt/code/linux/v2_20241029_xxx_offers.mbx
+    _output_file_pattern = re.compile(
+        r"^.*git.*/(?P<filename>[\w-]+\.mbx).*$", re.MULTILINE
+    )
+
+    @property
+    def command(self) -> str:
+        return "b4"
+
+    @property
+    def dependencies(self) -> List[Type[Tool]]:
+        return [Git]
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, Debian):
+            self.node.os.install_packages("b4")
+        installed = self._check_exists()
+        if not installed:
+            pip = self.node.tools[Pip]
+            pip.install_packages("b4", install_to_user=True)
+        return self._check_exists()
+
+    def apply(
+        self, message_id: str, cwd: pathlib.PurePath, sudo: bool = False
+    ) -> pathlib.PurePath:
+        """
+        Download the patch using the message id and apply it to the git repository.
+        """
+        result = self.run(
+            f"am -o '{cwd}' '{message_id}'",
+            force_run=True,
+            expected_exit_code=0,
+            sudo=sudo,
+        )
+        filename = find_group_in_lines(
+            lines=result.stdout, pattern=self._output_file_pattern, single_line=False
+        ).get("filename")
+        if not filename:
+            raise LisaException("Failed to get filename from b4 am output")
+        filepath = pathlib.PurePath(cwd, filename)
+
+        git = self.node.tools[Git]
+        git.apply(cwd=cwd, patches=filepath)
+
+        return filepath
--- a/lisa/transformers/kernel_source_installer.py
+++ b/lisa/transformers/kernel_source_installer.py
@@ -11,7 +11,7 @@ from lisa import schema
 from lisa.base_tools import Mv
 from lisa.node import Node
 from lisa.operating_system import CBLMariner, CpuArchitecture, Redhat, Ubuntu
-from lisa.tools import Cp, Echo, Git, Make, Sed, Uname
+from lisa.tools import B4, Cp, Echo, Git, Make, Sed, Uname
 from lisa.tools.gcc import Gcc
 from lisa.tools.lscpu import Lscpu
 from lisa.util import (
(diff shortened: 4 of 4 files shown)


lisa/tools/git.py
```diff
@@ -201,6 +201,17 @@ def apply(
         )
         result.assert_exit_code(message=f"failed on applying patches. {result.stdout}")
 
+    def am(self, cwd: pathlib.PurePath, patch: pathlib.PurePath) -> None:
+        result = self.run(
+            f"am {patch}",
```
LiliDeng: does git am depend on b4 package?

adityagesh: The output of b4 is a .mbx file. You can apply this patch to the repo using git am "filename.mbx"

As per description, git am can be used to apply any email related patch, whereas git apply is for the .patch files

lisa/tools/git.py
```diff
@@ -201,6 +201,17 @@ def apply(
         )
         result.assert_exit_code(message=f"failed on applying patches. {result.stdout}")
 
+    def am(self, cwd: pathlib.PurePath, patch: pathlib.PurePath) -> None:
```
kanchansenlaskar: I see you put arg name as "patch" unlike git apply where it has "patches". Does it not work with multiple patches/mboxes? Can you test once if not tested for the tools/library sake?

squirrelsc: Change the name to `apply_mail`. But if it's possible, you can check the file ext, and implement it in `apply` method.

lisa/tools/git.py
LiliDeng: I didn't see the place which uses b4 tool

adityagesh: I will be raising that as a separate PR, added that in the description.
I have created a Modifier in Kernel Install transformer that uses b4 to apply patches

adityagesh: I can add those changes to this PR, if you prefer having it in the same PR

lisa/tools/git.py
adityagesh: Interesting, I did not notice that.
Though the parameter is named patches, it accepts a Purepath and not list of Purepath.

I will verify the behavior and fix it along with Chi's comment below

kanchansenlaskar: By multiple I meant something like a pattern.
Ref: https://github.com/microsoft/lisa/blob/5c8d0258a6123094311e1bb7e9d64357fee99ee3/lisa/transformers/kernel_source_installer.py#L75

adityagesh: This is different, it is part of the modifier

lisa/tools/git.py
```diff
@@ -191,14 +191,25 @@ def apply(
...
-            shell=True,
-            cwd=cwd,
-            force_run=True,
-            no_info_log=True,
-            no_error_log=True,
-        )
+        file_extension = patches.suffix
+        if file_extension == ".mbx":
+            result = self.run(
+                f"am {patches}",
```
squirrelsc, line 197: The am is supported by b4 tool already. Does it need to support in git also?

lisa/tools/git.py
```diff
@@ -191,14 +191,25 @@ def apply(
...
-        result = self.run(
-            f"apply {patches}",
-            shell=True,
-            cwd=cwd,
-            force_run=True,
-            no_info_log=True,
-            no_error_log=True,
-        )
+        file_extension = patches.suffix
+        if file_extension == ".mbx":
```
squirrelsc, line 195: I just understood the scenario. Because the `.mbx` is always generated by b4, and the call sequence is always b4.am and git.apply. So the b4 can be called in git, or call git in b4, so it's easier to use.

adityagesh: How about something like this?
```
class Git(Tool):
  def apply_patch(self,....):
      # handles both apply and am
      
class B4(Tool):
  def am(self,....):
    self.run("am {patches}", ....)
    
  def apply(self,....):
      git = node.tool[Git]
      self.am()
      git.apply()

```

squirrelsc: 1. The `apply_patch` and `apply` are confusing in the Git tool. 
2. If call B4 in git, so the API can handle different types of patch files automatically. But if call git in B4, it still needs to know the type of patch files, and then call either git.apply or b4.apply.

adityagesh: The tool b4 has dependency on Git while pulling patches (taking Message_id as the parameter), so I don't think we can add b4 to Git.

Currently in Git.apply, it takes the path as parameter, which is basically the operation after we download the patch using b4.
B4 after downloading the patch should always call git apply to apply the patch.

The b4 apply that I showed above is : download patch using b4 -> apply downloaded patch using git

squirrelsc: 1. I saw the code mark Git as a dependency. It's not necessary, because it's always used after Git tool is called. 
2. The dependencies between b4 and git, are depended by how we manage them. The goal is to provide easy to use interface. If we provide B4 supports in git, the user doesn't need to take care the ext name, just call git to fit different file types. But if the B4 call git, it needs to know ext name firstly.

adityagesh: Are you recommending adding something like below in Git?
```
download_and_apply_patch(self, Message_Id):
 path = b4.am(Message_id)
 git.apply_patch(path)
```




> I saw the code mark Git as a dependency. It's not necessary, because it's always used after Git tool is called.

If we have the tool B4, it is possible someone uses it directly before invoking git. b4 will fail is git is not installed and not set as a dependency

squirrelsc: It's like below. There is no conflict to block direct use b4.

```python
    def apply(
        self,
        cwd: pathlib.PurePath,
        patches: pathlib.PurePath,
    ) -> None:
        file_extension = patches.suffix
        if file_extension == ".<b4's ext>"
            from .b4 import B4
            b4 = self.node.tools[B4]
            b4.am(...
            self.run(...
        else:
            self.run(...
```

adityagesh:
```
  def apply(
        self,
        cwd: pathlib.PurePath,
        patches: pathlib.PurePath,
    ) -> None:
        file_extension = patches.suffix
        if file_extension == ".<b4's ext>"
            from .b4 import B4
            b4 = self.node.tools[B4]
            b4.am(...
            self.run(...
        else:
            self.run(...
```

This logic is actually incorrect. The b4 tool accepts Message_id, not the filepath. 
B4 downloads the patchfile using the message id and then we need to parse the log to get the filepath, which is the then applied using git.


An example is 
 b4 am -o "/home/aditya/tmp/" "<20240726225910.1912537-4-romank@linux.microsoft.com>"
 
 Output
 
 ```
Looking up https://lore.kernel.org/r/20240726225910.1912537-4-romank%40linux.microsoft.com
Grabbing thread from lore.kernel.org/all/20240726225910.1912537-4-romank%40linux.microsoft.com/t.mbox.gz
Analyzing 53 messages in the thread
Checking attestation on all messages, may take a moment...
---
  ✓ [PATCH v3 1/7] arm64: hyperv: Use SMC to detect hypervisor presence
    + Acked-by: Wei Liu <wei.liu@kernel.org>
  ✓ [PATCH v3 2/7] Drivers: hv: Enable VTL mode for arm64
    + Acked-by: Wei Liu <wei.liu@kernel.org>
  ✓ [PATCH v3 3/7] Drivers: hv: Provide arch-neutral implementation of get_vtl()
  ✓ [PATCH v3 4/7] arm64: hyperv: Boot in a Virtual Trust Level
  ✓ [PATCH v3 5/7] dt-bindings: bus: Add Hyper-V VMBus cache coherency and IRQs
  ✓ [PATCH v3 6/7] Drivers: hv: vmbus: Get the IRQ number from DT
  ✓ [PATCH v3 7/7] PCI: hv: Get vPCI MSI IRQ domain from DT
  ---
  ✓ Signed: DKIM/linux.microsoft.com
  ---
  NOTE: install patatt for end-to-end signature verification
---
Total patches: 7
---
Cover: /home/aditya/tmp/v3_20240726_romank_arm64_hyperv_support_virtual_trust_level_boot.cover
 Link: https://lore.kernel.org/r/20240726225910.1912537-1-romank@linux.microsoft.com
 Base: 831bcbcead6668ebf20b64fdb27518f1362ace3a
       git checkout -b v3_20240726_romank_linux_microsoft_com 831bcbcead6668ebf20b64fdb27518f1362ace3a
       git am /home/aditya/tmp/v3_20240726_romank_arm64_hyperv_support_virtual_trust_level_boot.mbx
```

Comment (squirrelsc) [reply lisa/tools/git.py:195]: Thank you for explanation. I didn't notice the parameter of b4 is not a file path. How about like below?

```python
class B4(Tool):
    def apply(self, message_id, ...):
        result = self.run(
            f"am -o '{output_dir}' '{message_id}'",
            force_run=True,
            expected_exit_code=0,
            sudo=sudo,
        )
        filename = find_group_in_lines(
            lines=result.stdout, pattern=self._output_file_pattern, single_line=False
        ).get("filename")
        if not filename:
            raise LisaException("Failed to get filename from b4 am output")

      git = node.tool[Git]
      git.apply(...
```

adityagesh: @squirrelsc , this is updated

squirrelsc:
@LiliDeng LGTM

LiliDeng (dismissed):

adityagesh:
Rebased

LiliDeng (approved):

PR #3511, opened 2024-11-13
Add basic tests for mdatp

atbagga:
- Adding new tests for mdatp which check for onboarding and basic eicar_detection test
- Renaming the existing test which checks for clean state of the image as mdatp_clean

This change is based of the original PR - https://github.com/microsoft/lisa/pull/3113


microsoft/testsuites/mdatp/mdatp_basic.py
```diff
@@ -0,0 +1,135 @@
+# Copyright (c) Microsoft Corporation.
```
squirrelsc, line 1: Please put them under the vm_extensions folder, so we can manage vm extension cases easier. You can have a subfolder there.

microsoft/testsuites/mdatp/mdatp_basic.py
```diff
@@ -0,0 +1,135 @@
...
+        1. Installation test
+        2. Onboarding test
+        3. Health test
+        4. EICAR detection test
+    """,
+)
+class MDETest(TestSuite):
+    def before_case(self, log: Logger, **kwargs: Any) -> None:
+        variables = kwargs["variables"]
+        self.onboarding_script_sas_uri = variables.get("onboarding_script_sas_uri", "")
```
squirrelsc, line 48: Can it support non-sas uri? The extra requirement makes the test case skipped in most pipelines.

microsoft/testsuites/mdatp/mdatp_tools.py
```diff
@@ -0,0 +1,93 @@
...
+                url="https://raw.githubusercontent.com/microsoft/mdatp-xplat/"
+                "master/linux/installation/mde_installer.sh",
+                filename="mde_installer.sh",
+            )
+            self.mde_installer = download_path
+            self.node.tools[Chmod].update_folder(self.mde_installer, "777", sudo=True)
+        return True
+
+    def _install(self) -> bool:
+        if not self.get_mde_installer():
```
squirrelsc, line 34: It looks the `get_mde_installer` return true always. The code logic won't be hit.

microsoft/testsuites/mdatp/mdatp_basic.py
```diff
@@ -0,0 +1,135 @@
...
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            min_core_count=2, min_memory_mb=1024, unsupported_os=[BSD]
+        ),
+    )
+    def verify_mde(self, node: Node, log: Logger, result: TestResult) -> None:
+        # Invoking tools first time, intalls the tool.
+        try:
+            output = node.tools[Mdatp]._check_exists()
```
squirrelsc, line 64: 1. Don't call the private method like `_check_exists`, call `exists`.
2. But when you refer a tool like this, the check/install happens automatically. So, the code can be like `_ = node.tools[Mdatp]`, and capture the exception if there is.

microsoft/testsuites/mdatp/mdatp_tools.py
```diff
@@ -0,0 +1,93 @@
...
+        self._log.info("Installing MDE")
+        result1 = self.node.execute(
+            f"{self.mde_installer} --install", shell=True, sudo=True
+        )
+        self._log.info(result1)
+
+        return self._check_exists()
+
+    def onboard(self, onboarding_script_sas_uri: str) -> bool:
+        if not self._check_exists():
```
squirrelsc, line 48: It doesn't need to check, when you have an instance of the tool, it means the check exists already passed.

microsoft/testsuites/mdatp/mdatp_tools.py
```diff
@@ -0,0 +1,93 @@
...
+        if json_out:
+            arg += " --output json"
+        result = self.run(
+            arg,
+            sudo=sudo,
+            shell=True,
+            force_run=True,
+        )
+
+        result.assert_exit_code(include_output=True)
```
squirrelsc, line 91: Add error message for assertion, so the test case can bring more information, when it's failed.

LiliDeng:
@atbagga feel free to open one when it is ready, thanks.

PR #3515, opened 2024-11-15
avoid fetching information unnecessarily when rerun env

LiliDeng:


lisa/testsuite.py
```diff
@@ -290,7 +290,18 @@ def _send_result_message(self, stacktrace: Optional[str] = None) -> None:
             # force refresh information, when test result status is changed. The
             # refreshed information is not used so far. But in case it's needed
             # in future, keep it up to date.
-            self._environment_information = self.environment.get_information()
+            # avoid fetching information unnecessarily when rerun env
+            force_run = True
```
squirrelsc, line 294: The information maybe changed by install new kernel or other situation. So it needs to force run every time.

lisa/testsuite.py
```diff
@@ -290,7 +290,18 @@ def _send_result_message(self, stacktrace: Optional[str] = None) -> None:
...
+            if (
```
squirrelsc, line 295: The `if` block needs more explanation to understand what's the problem is fixed. so it won't be confused in future.

kamalca:
@LiliDeng 
Can you help me understand what state the environment is in when it is returning incomplete info from `environemnt.get_info()`?

Why is it not possible to fix `environemnt.get_info()` such that it will always return correct and complete information without needing to rely on cache.

I am fine with relying on cache to save time, but we should not rely on cache for correctness. We should aim for `environemnt.get_info(force_run=True)` to always return correct and complete information if possible.

PR #3520, opened 2024-11-20
Check and uninstall extensions before installing it

LiliDeng:


microsoft/testsuites/vm_extensions/runtime_extensions/custom_script.py
```diff
@@ -38,10 +39,22 @@ def _create_and_verify_extension_run(
     assert_exception: Any = None,
 ) -> None:
     extension = node.features[AzureExtension]
+    extension_name = "CustomScript"
```
squirrelsc, line 42: Why does it need to uninstall? The `create_or_update` can update the existing extension.

LiliDeng: the env can be reused, the second installation return result is different from the first installation result, which cause the assert error.

squirrelsc (approved):

PR #3521, opened 2024-11-20
Stabilize MPI tests for Azure Linux

sandeepkarambelkar:


microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -286,6 +286,9 @@ def verify_intel_mpi(self, environment: Environment, log: Logger) -> None:
         client_ssh.enable_public_key(server_ssh.generate_key_pairs())
         server_ssh.add_known_host(client_ip)
         client_ssh.add_known_host(server_ip)
+        sudo=False
+        if isinstance(server_node.os, CBLMariner):
```
squirrelsc, line 290: Please add comments above to explain why Mariner needs sudo.

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -360,10 +366,13 @@ def verify_open_mpi(self, environment: Environment, log: Logger) -> None:
             raise SkippedException(err)
 
         run_in_parallel([server_ib.install_open_mpi, client_ib.install_open_mpi])
-
         server_node.execute("ldconfig", sudo=True)
         client_node.execute("ldconfig", sudo=True)
 
+        # Only for mariner, we need to build intel benchmarking tools 
+        # as they are not included in our packages
+        server_ib.install_intel_mpi_benchmarking_tool()
```
squirrelsc, line 374: Please also add `if` block here to skip other distros. So the logic here is easier to know at the test case level. The `if` block in the `install_intel_mpi_benchmarking_tool` is a safe guard.

microsoft/testsuites/hpc/infinibandsuite.py
```diff
@@ -571,6 +580,12 @@ def verify_mvapich_mpi(self, environment: Environment, log: Logger) -> None:
             raise SkippedException(err)
 
         run_in_parallel([server_ib.install_mvapich_mpi, client_ib.install_mvapich_mpi])
+        test_names = ["IMB-MPI1", "IMB-RMA", "IMB-NBC"]
```
squirrelsc, line 583: Add `if` block for Mariner.

sandeepkarambelkar: This is moved from line no 596 to reuse for mariner. I will add if block after this as this is required for other tests also.

microsoft/testsuites/hpc/infinibandsuite.py
sandeepkarambelkar: Sure.

PR #3522, opened 2024-11-20
Add support for mariner-3.0 in dom0 kernel installer

smit-gardhariya:
dom0 image based on mariner 3.0 has different
parameter that need to be changed while installing kernel under different file. Add support for it under kernel installer transformer.

### DIFF
--- a/lisa/transformers/dom0_kernel_installer.py
+++ b/lisa/transformers/dom0_kernel_installer.py
@@ -9,6 +9,7 @@ from dataclasses_json import dataclass_json
 
 from lisa import schema
 from lisa.node import Node
+from lisa.operating_system import CBLMariner
 from lisa.tools import Cp, Echo, Ln, Ls, Sed, Tar, Uname
 from lisa.util import field_metadata
 
@@ -80,6 +81,8 @@ class BinaryInstaller(BaseInstaller):
         uname = node.tools[Uname]
         current_kernel = uname.get_linux_information().kernel_version_raw
 
+        mariner_version = int(node.os.information.version.major)
+
         # Kernel absolute path: /home/user/vmlinuz-5.15.57.1+
         # Naming convention : vmlinuz-<version>
         new_kernel = os.path.basename(kernel_image_path).split("-")[1].strip()
@@ -124,10 +127,19 @@ class BinaryInstaller(BaseInstaller):
                 node.get_pure_path(f"/boot/initrd.img-{new_kernel}"),
             )
         else:
+            # Mariner 3.0 initrd
+            target = f"/boot/initramfs-{current_kernel}.img"
+            link = f"/boot/initramfs-{new_kernel}.img"
+
+            if isinstance(node.os, CBLMariner) and mariner_version == 2:
+                # Mariner 2.0 initrd
+                target = f"/boot/initrd.img-{current_kernel}"
+                link = f"/boot/initrd.img-{new_kernel}"
+
             ln = node.tools[Ln]
             ln.create_link(
-                target=f"/boot/initrd.img-{current_kernel}",
-                link=f"/boot/initrd.img-{new_kernel}",
+                target=target,
+                link=link,
             )
 
         if kernel_config_path:
@@ -148,6 +160,7 @@ class BinaryInstaller(BaseInstaller):
             node,
             current_kernel,
             new_kernel,
+            mariner_version,
         )
 
         return new_kernel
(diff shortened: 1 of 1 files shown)


lisa/transformers/dom0_kernel_installer.py
```diff
@@ -124,10 +127,19 @@ def install(self) -> str:
                 node.get_pure_path(f"/boot/initrd.img-{new_kernel}"),
             )
         else:
+            # Mariner 2.0 initrd
+            target = f"/boot/initrd.img-{current_kernel}"
+            link = f"/boot/initrd.img-{new_kernel}"
+
+            if distro == "Microsoft Azure Linux 3.0":
```
squirrelsc: Check it by `isinstance(node.os, Mariner)` and the `node.os.information.version`. Not use the string.

smit-gardhariya: Sure Chi

lisa/transformers/dom0_kernel_installer.py
```diff
@@ -80,6 +81,9 @@ def install(self) -> str:
         uname = node.tools[Uname]
         current_kernel = uname.get_linux_information().kernel_version_raw
 
+        mariner_version = node.os.information.version
+        print(f"mariner_version: {mariner_version}")
```
squirrelsc: Don't use print in code. The version number should be in debug information already, it doesn't need to be printed again.

smit-gardhariya: It was just temporary print. I removed it in last push.

lisa/transformers/dom0_kernel_installer.py
smit-gardhariya: Done

lisa/transformers/dom0_kernel_installer.py
```diff
@@ -124,10 +127,19 @@ def install(self) -> str:
...
             )
         else:
+            # Mariner 2.0 initrd
+            target = f"/boot/initrd.img-{current_kernel}"
+            link = f"/boot/initrd.img-{new_kernel}"
+
+            if isinstance(node.os, CBLMariner) and mariner_version == 3:
+                # Mariner 3.0 initrd
+                target = f"/boot/initramfs-{current_kernel}.img"
+                link = f"/boot/initramfs-{new_kernel}.img"
```
pupacha: Can we make 3.0 's values as default? Most probably, next versions of mariner (4, 5 etc) would have same file paths as 3.0 and it would avoid us code changes.

```suggestion
            # Mariner 3.0 initrd
            target = f"/boot/initramfs-{current_kernel}.img"
            link = f"/boot/initramfs-{new_kernel}.img"

            if isinstance(node.os, CBLMariner) and mariner_version == 2:
                # Mariner 2.0 initrd
                target = f"/boot/initrd.img-{current_kernel}"
                link = f"/boot/initrd.img-{new_kernel}"
```

lisa/transformers/dom0_kernel_installer.py
```diff
@@ -211,22 +226,35 @@ def _update_mariner_config(
...
+    vmlinuz_regexp = f"vmlinuz-{current_kernel}"
+    vmlinuz_replacement = f"vmlinuz-{new_kernel}"
+    initrd_regexp = f"mariner_initrd_mshv=initrd.img-{current_kernel}"
+    initrd_replacement = f"mariner_initrd_mshv=initrd.img-{new_kernel}"
+
+    if isinstance(node.os, CBLMariner) and mariner_version == 3:
+        # Change param for Dom0 3.0 kernel installation
+        mariner_config = "/boot/grub2/grub.cfg"
+        initrd_regexp = f"initramfs-{current_kernel}.img"
+        initrd_replacement = f"initramfs-{new_kernel}.img"
```
pupacha: Same comment here.

lisa/transformers/dom0_kernel_installer.py
smit-gardhariya: You are right. If newer mariner version supposed to use the same grub file, it would make sense to use >=3 as default case. Let me check on this.

smit-gardhariya: Done please check

lisa/transformers/dom0_kernel_installer.py
smit-gardhariya: Done please check

pupacha (approved):

LiliDeng (approved):

PR #3523, opened 2024-11-21
CVM: Skip test cases not supporting TDX

kamalca:
Attestation report test cases only support AMD SEV-SNP so they should skip on other CPU types.

Also do not skip cpu_offline_channel_add test case as it is supported on all CPU types.

### DIFF
--- a/microsoft/testsuites/cvm/cvm_attestation.py
+++ b/microsoft/testsuites/cvm/cvm_attestation.py
@@ -16,7 +16,8 @@ from lisa.features.security_profile import CvmEnabled
 from lisa.operating_system import Ubuntu
 from lisa.sut_orchestrator import AZURE
 from lisa.testsuite import TestResult, simple_requirement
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/cvm/cvm_attestation.py
```diff
@@ -61,6 +62,10 @@ def verify_azure_cvm_attestation_report(
         result: TestResult,
         variables: Dict[str, Any],
     ) -> None:
+        if node.tools[Lscpu].get_cpu_type() != CpuType.AMD:
```
trungams: maybe move this condition check to the `before_case` function above, that's where the current environment check is happening

kamalca: Good catch :) Changed

trungams (approved):

LiliDeng (approved):

PR #3525, opened 2024-11-22
add templates for sig/vhd for AITL partners

katmnsn:
These templates are necessary for our AITL partners to more easily understand how to interact with the AITL service for vhd, shared image gallery and marketplace image jobs, and the different tiers of tests available to them.

### DIFF
--- a/microsoft/utils/aitl/tier0-sig.json
+++ b/microsoft/utils/aitl/tier0-sig.json
@@ -1,0 +1,25 @@
+{
+    "location": "westus3",
+    "properties": {
+        "jobTemplateInstance": {
+            "templateTags": [],
+            "selections": [
+                {
+                    "casePriority": [
+                        0
+                    ]
+                }
+            ],
+            "region": [],
+            "vmSize": []
+        },
+        "image": {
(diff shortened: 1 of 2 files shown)


katmnsn:
@microsoft-github-policy-service agree [company="Microsoft"]

katmnsn:
@microsoft-github-policy-service agree

microsoft/utils/aitl/tier0-sig.json
```diff
@@ -0,0 +1,25 @@
+{
```
squirrelsc, line 1: tier0-sig and tier0-vhd are enough. Because those templates cannot be used directly, so it doesn't worth to provide multiple files for convenient.

katmnsn: Makes sense, will get rid of others and just explain how to use the different tiers in docs

squirrelsc:
Please merge commits into one to get the commit history clear, and then it's ready to publish. It doesn't need to create in a draft stage by default, because the checks cannot be triggered in draft phase.

squirrelsc (approved):

PR #3526, opened 2024-11-22
xfstests: fix umask for mariner

realsdx:
CBL-Mariner has a different umask (`027`) setting than other operating systems, which prevents any new directories created by `other` users non-executable. This scenario is incompatible with xfstests. This PR changes the umask setting for Mariner to `0022` for the xfstests.

Also, the already created `/root` directory has similar permissions. This PR changes the directories to `/mnt` to avoid the same.

### DIFF
--- a/microsoft/testsuites/xfstests/xfstesting.py
+++ b/microsoft/testsuites/xfstests/xfstesting.py
@@ -18,7 +18,7 @@ from lisa import (
     simple_requirement,
 )
 from lisa.features import Disk, Nvme
-from lisa.operating_system import BSD, Oracle, Redhat, Windows
+from lisa.operating_system import BSD, Oracle, Redhat, Windows, CBLMariner
 from lisa.sut_orchestrator import AZURE
 from lisa.sut_orchestrator.azure.features import AzureFileShare
 from lisa.sut_orchestrator.azure.platform_ import AzurePlatform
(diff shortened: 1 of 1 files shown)


LiliDeng (approved):

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -27,8 +27,8 @@
 from lisa.util import BadEnvironmentStateException, generate_random_chars
 from microsoft.testsuites.xfstests.xfstests import Xfstests
 
-_scratch_folder = "/root/scratch"
-_test_folder = "/root/test"
+_scratch_folder = "/mnt/scratch"
```
squirrelsc, line 30: I'm not sure the folder exists or not, if the VM size doesn't have a resource disk.

realsdx: `/mnt` should be present in all standard linux distros, as it's part of the standard linux file hierarchy.  If VM doesn't have a resource disk there would be no mounts I think, but the directory will exist. For BSD this might be different, which BSD distros we are supporting? I can check for the specific BSD distro's default file hierarchy.

squirrelsc: Thank you, please have a check on FreeBSD.

PR #3529, opened 2024-11-23
CVM: add attestation report test for Azure Linux using snpguest tool

trungams:
- Extend cargo tool to support optional build parameters
- Add support for [snpguest](https://github.com/virtee/snpguest)
- Extend Azure CVM attestation report test case for Azure Linux 3.0. The attestation workflow done in the test case is described in [snpguest's README](https://github.com/virtee/snpguest?tab=readme-ov-file#regular-attestation-flowchart)


lisa/tools/cargo.py
```diff
@@ -117,6 +117,7 @@ def __install_dependencies(self) -> None:
 
     def build(
         self,
+        options: str = "",
```
squirrelsc: The options is flexible, but not easy to reuse logic. Please define specified arguments as much as possible.

microsoft/testsuites/cvm/cvm_attestation.py
```diff
@@ -13,14 +15,15 @@
     features,
 )
 from lisa.features.security_profile import CvmEnabled
-from lisa.operating_system import Ubuntu
+from lisa.operating_system import CBLMariner, Ubuntu
```
manuelh-dev, line 15: We should rename CBLMariner to AzureLinux. Please feel free to resolve this comment, just wanted to bring this to attention (feel free to volunteer to change in a separate PR or create an issue or notify some other maintainer :)

microsoft/testsuites/cvm/cvm_attestation_tool.py
```diff
@@ -121,6 +122,151 @@ def _save_attestation_report(self, output: str, log_path: Path) -> None:
             f.write(output)
 
 
+class SnpGuest(Tool):
```
manuelh-dev, line 125: Instead of introducing a new class with new name, can we streamline this better? In a way, this class could as well be called "AzureCVMAttestationTests". Is there a way to implement this in a single class (or rename classes with more clear names) or in a pattern where we don't need to add the `if isinstance(node.os, Ubuntu):` condition?

We can e.g. pass node.os to the runc_cvm_attestation function as parameter

microsoft/testsuites/cvm/cvm_attestation.py
```diff
@@ -61,11 +65,14 @@ def verify_azure_cvm_attestation_report(
...
-            log_path,
-        )
+        if isinstance(node.os, Ubuntu):
+            node.tools[AzureCVMAttestationTests].run_cvm_attestation(
+                result,
+                environment,
+                log_path,
+            )
+        elif isinstance(node.os, CBLMariner):
+            assert_that(node.tools[SnpGuest].run_cvm_attestation()).is_true()
```
manuelh-dev: Do we not want/need to use similar parameters to the existing test?

microsoft/testsuites/cvm/cvm_attestation.py
squirrelsc: Please provide a separated PR to rename it.

microsoft/testsuites/cvm/cvm_attestation_tool.py
trungams: This class represents the [snpguest](https://github.com/virtee/snpguest) tool and should at least be separate from the tool above. The distro check is to install specific packages since package names may be different across distros. It's generally recommended to do that in Lisa I think

microsoft/testsuites/cvm/cvm_attestation.py
trungams: imo since this is the actual test case definition. Assertion should be done here rather than in the tool

squirrelsc: If all cases need a success result, you can assert in the tool. And please always add `describe_as` to provide an error message in case it happens. So it's easy to know the error in the test results, instead of digging into the code.

microsoft/testsuites/cvm/cvm_attestation_tool.py
manuelh-dev: But: If we do that check, it sounds like this makes the tool/test specific for Azure Linux, similar to as the previous tool is specific to Ubuntu.

Is there any desire and better way to make this more explicit for both tools?

squirrelsc: If there are many places need to use this tool, it needs to create a virtual tool like `Service` in LISA. It wraps different implementation and provide an unified interface.

microsoft/testsuites/cvm/cvm_attestation.py
trungams: Yup. Each function in the tool already has an assert for exit code = 0, and an error message in case of failure. The assert here is more or less to make it clear what we're looking for.

squirrelsc: 1. This line can be `node.tools[SnpGuest].run_cvm_attestation())`. It should be enough, if the method have enough assertion already.
2. In case it needs to check the return code, add `described_as` to provide error message for this `assert_that`. Please have a try to understand the behaviors of assert_that.

trungams: Updated

squirrelsc:
@LiliDeng LGTM

LiliDeng (dismissed):

LiliDeng:
@trungams please fix the conflict.

trungams:
@LiliDeng done

LiliDeng (approved):

PR #3531, opened 2024-11-25
RunCommandv1: uninstall extension before installing it

LiliDeng:


microsoft/testsuites/vm_extensions/runtime_extensions/run_commandv1.py
```diff
@@ -37,10 +38,22 @@ def _create_and_verify_extension_run(
     assert_exception: Optional[Any] = None,
 ) -> None:
     extension = node.features[AzureExtension]
+    extension_name = "RunCommandv1"
+    try:
+        # Delete VM Extension if already present
+        extension.delete(extension_name)
```
squirrelsc: Update the method to add a flag like `ignore_not_found`. So the logic can be shared.

squirrelsc (approved):

PR #3532, opened 2024-11-26
Hibernation Test case fixes

adityagesh:
PR includes 3 commits
1. Reboot after running hibernation-setup-tool on RHEL
2. Skip Hibernation on Mariner. It is not supported.
3. Change use of `uptime -s` to `last` command

### DIFF
--- a/microsoft/testsuites/power/common.py
+++ b/microsoft/testsuites/power/common.py
@@ -10,7 +10,7 @@ from lisa import Environment, Logger, Node, RemoteNode, features
 from lisa.base_tools.cat import Cat
 from lisa.features import StartStop
 from lisa.features.startstop import VMStatus
-from lisa.operating_system import CBLMariner, Redhat, Suse, Ubuntu
+from lisa.operating_system import Debian, Redhat, Ubuntu
 from lisa.tools import (
     Dmesg,
     Fio,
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/power/common.py
```diff
@@ -72,9 +72,20 @@ def verify_hibernation(
 
     # only set up hibernation setup tool for the first time
     hibernation_setup_tool.start()
-    uptime = node.tools[Uptime]
+    # This is a temporary workaround for a bug observed in Redhat Distros
+    # where the VM is not able to hibernate immediately after installing
+    # the hibernation-setup tool.
+    # A sleep(100) also works, but we are unsure of the exact time required.
+    # So it is safer to reboot the VM.
+    if isinstance(node.os, Redhat):
```
LiliDeng: if only happen in redhat, you can use type(node.os) == Redhat

adityagesh: done

adityagesh: done

LiliDeng (approved):

microsoft/testsuites/power/common.py
```diff
@@ -38,10 +37,10 @@ def is_distro_supported(node: Node) -> None:
             f"version {node.os.information.version}"
         )
 
-    if (
-        (isinstance(node.os, Redhat) and node.os.information.version < "8.3.0")
-        or (isinstance(node.os, Ubuntu) and node.os.information.version < "18.4.0")
-        or (isinstance(node.os, Suse) and node.os.information.version < "15.3.0")
+    if not (
+        (type(node.os) == Ubuntu and node.os.information.version >= "18.04.0")
```
LiliDeng, line 41: please fix this regression `(type(node.os) == Ubuntu and node.os.information.version >= "18.4.0")`

LiliDeng: fixed in #3533

PR #3535, opened 2024-12-02
Add more description for the connection failure

lubaihua33:
I have checked the failure triage. All the added description just match ".*" of existing failure pattern. So it won't have any impact to the current failure pattern.

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -269,7 +269,11 @@ class SshShell(InitializableMixin):
             stdout = try_connect(self.connection_info, sock=sock)
         except Exception as identifier:
             raise LisaException(
-                f"failed to connect SSH "
+                "failed to connect SSH port of the VM. It might be due to one of the "
+                "following reasons: 1. Port 22 is not open. 2. The VM denies other "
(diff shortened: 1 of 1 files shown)


lisa/util/shell.py
```diff
@@ -269,7 +269,11 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
             stdout = try_connect(self.connection_info, sock=sock)
         except Exception as identifier:
             raise LisaException(
-                f"failed to connect SSH "
+                f"failed to connect SSH port of the VM. It might be due to one of the "
```
LiliDeng: "failed to connect SSH port of the VM. It might be due to one of the "

squirrelsc: > "failed to connect SSH port of the VM. It might be due to one of the "

What's changed from the original text?

LiliDeng: > > "failed to connect SSH port of the VM. It might be due to one of the "
> 
> What's changed from the original text?

remove the `f`

lubaihua33: Done

LiliDeng (approved):

PR #3536, opened 2024-12-02
change dockerfile

LiliDeng:


squirrelsc (approved):

Dockerfile
```diff
@@ -1,17 +1,32 @@
 FROM mcr.microsoft.com/cbl-mariner/base/core:2.0
```
squirrelsc, line 1: Is the 3.0 ready to use?

LiliDeng: not yet, I have confirmed with their team. Also try to search here https://mcr.microsoft.com/en-us/catalog?search=mariner&type=partial&page=1.

PR #3538, opened 2024-12-02
Boot Partition can be mounted to /efi

kamalca:
For Azure Linux 3.0 we find the boot partition mounted on /efi instead of /boot/efi

This is following [UAPI's boot loader specs](https://uapi-group.org/specifications/specs/boot_loader_specification/#mount-points)

>  If both XBOOTLDR and the ESP are present, the ESP should be mounted to `/efi/`

> Mounting the ESP to `/boot/efi/`, as was traditionally done, is not recommended.

### DIFF
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -8,7 +8,7 @@ from typing import Any, Dict, List, Optional, Type, Union
 from lisa import schema
 from lisa.feature import Feature
 from lisa.operating_system import BSD
-from lisa.tools import Mount
+from lisa.tools import Ls, Mount
 from lisa.tools.mount import PartitionInfo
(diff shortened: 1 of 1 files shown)


kamalca:
This is actually not an adequate solution.

Here is another quote from the spec:
> Mounting the two partitions via autofs is recommended because the simple VFAT file system has weak data integrity properties and should remain unmounted whenever possible.

This means the VFAT partition can and likely will be dismounted, leaving the boot partition unmounted. We will need a better way to determine the boot partition than looking for a mount point.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3543, opened 2024-12-03
Remove the check of connection to azsecmond in auoms log

lubaihua33:
In some scenarios, the line "Output(azsecmond): Connected" isn't in auoms log. Confirmed with the AzSecpack team, they didn't see any impact when there is no successfull connection to azsecmond in auoms log". Therefore, we have temporarily removed this check until the AzSecPack team provides final confirmation.

### DIFF
--- a/microsoft/testsuites/vm_extensions/azsecpack.py
+++ b/microsoft/testsuites/vm_extensions/azsecpack.py
@@ -325,17 +325,23 @@ class AzSecPack(TestSuite):
         log.info("Azsecmond connection to mdsd is successful")
 
         output = journalctl.logs_for_unit("auoms", sudo=True)
+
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/vm_extensions/azsecpack.py
```diff
@@ -325,17 +325,23 @@ def _check_journalctl_logs(self, node: Node, log: Logger) -> None:
         log.info("Azsecmond connection to mdsd is successful")
 
         output = journalctl.logs_for_unit("auoms", sudo=True)
+
+        # Previously, we alo checked the string "Output(azsecmond): Connected".
```
LiliDeng: alo => also

LiliDeng (approved):

PR #3544, opened 2024-12-05
skip case when no libvirt module installed

LiliDeng:


microsoft/testsuites/cvm/cvm_attestation.py
```diff
@@ -122,6 +121,10 @@ def verify_nested_cvm_attestation_report(
         result: TestResult,
         variables: Dict[str, Any],
     ) -> None:
+        try:
+            from lisa.sut_orchestrator.libvirt.context import get_node_context
+        except ModuleNotFoundError as e:
+            raise SkippedException(f"Skipping test: {str(e)}")
```
squirrelsc: it doesn't need to add the try/except, just move the import here.

squirrelsc (approved):

PR #3546, opened 2024-12-06
Improve VhdTransformer with CVM

kamalca:
Have the VhdTransformer support CVM by exporting the VMGS along with the VHD.

Then restrict the Security Profile of the VhdSchema based on the presence of a VMGS file.

Together, these changes allow the VhdTransformer to take a VM and test the exported VHD without knowing whether the image is CVM or not. The presence of the VMGS file during disk export provides the necessary information.

### DIFF
--- a/lisa/sut_orchestrator/azure/transformers.py
+++ b/lisa/sut_orchestrator/azure/transformers.py
@@ -116,6 +116,7 @@ class VhdTransformer(Transformer):
     """
 
     __url_name = "url"
+    __vmgs_url_name = "vmgs_url"
 
     @classmethod
     def type_name(cls) -> str:
@@ -127,7 +128,7 @@ class VhdTransformer(Transformer):
 
     @property
     def _output_names(self) -> List[str]:
-        return [self.__url_name]
+        return [self.__url_name, self.__vmgs_url_name]
 
     def _internal_run(self) -> Dict[str, Any]:
         runbook: VhdTransformerSchema = self.runbook
@@ -157,11 +158,11 @@ class VhdTransformer(Transformer):
 
         virtual_machine = get_vm(platform, node)
 
-        vhd_location = self._export_vhd(platform, virtual_machine)
+        outputs = self._export_vhd(platform, virtual_machine)
 
         self._restore_vm(platform, virtual_machine, node)
 
-        return {self.__url_name: vhd_location}
+        return outputs
 
     def _prepare_virtual_machine(self, node: RemoteNode) -> None:
         runbook: VhdTransformerSchema = self.runbook
@@ -178,7 +179,9 @@ class VhdTransformer(Transformer):
         startstop = node.features[StartStop]
         startstop.stop()
 
-    def _export_vhd(self, platform: AzurePlatform, virtual_machine: Any) -> str:
+    def _export_vhd(
+        self, platform: AzurePlatform, virtual_machine: Any
+    ) -> Dict[str, str]:
         runbook: VhdTransformerSchema = self.runbook
         compute_client = get_compute_client(platform)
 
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -53,15 +53,18 @@
...
-        f"{DEFAULT_VHD_SUFFIX}_{file_name_part}.vhd"
+def _generate_vhd_path(
+    container_client: Any, file_name_part: str = ""
+) -> Tuple[str, str]:
+    path_prefix = (
+        f"{get_date_str()}/{get_datetime_path()}_{DEFAULT_VHD_SUFFIX}_{file_name_part}"
     )
-    blobs = container_client.list_blobs(name_starts_with=path)
+    vhd_path = PurePosixPath(f"{path_prefix}.vhd")
+    vmgs_path = PurePosixPath(f"{path_prefix}_vmgs.vhd")
```
squirrelsc: For vmgs file, is the vmgs an ext name or still be the vhd ext name?

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -53,15 +53,18 @@
 
 
 @retry(tries=10, jitter=(1, 2))
-def _generate_vhd_path(container_client: Any, file_name_part: str = "") -> str:
-    path = PurePosixPath(
-        f"{get_date_str()}/{get_datetime_path()}_"
-        f"{DEFAULT_VHD_SUFFIX}_{file_name_part}.vhd"
+def _generate_vhd_path(
```
squirrelsc: Let it return one path like before, and use replace to generate vmgs path like when it's used. You can still check both path inside of the method.

lisa/sut_orchestrator/azure/transformers.py
kamalca: VMGS still has .vhd extension

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -178,22 +182,38 @@ def _prepare_virtual_machine(self, node: RemoteNode) -> None:
...
         runbook: VhdTransformerSchema = self.runbook
         compute_client = get_compute_client(platform)
 
         # generate sas url from os disk, so it can be copied.
         self._log.debug("generating sas url...")
         location = virtual_machine.location
         os_disk_name = virtual_machine.storage_profile.os_disk.name
+
+        is_cvm = (
+            virtual_machine.storage_profile.os_disk.managed_disk.security_profile
```
squirrelsc, line 194: Is it accurate? Why security file is not None equals CVM?

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -219,16 +239,35 @@ def _export_vhd(self, platform: AzurePlatform, virtual_machine: Any) -> str:
...
+            wait_copy_blob(vmgs_blob_client, vmgs_url_path, self._log)
         else:
-            path = _generate_vhd_path(container_client, runbook.file_name_part)
-        vhd_path = f"{container_client.url}/{path}"
-        blob_client = container_client.get_blob_client(path)
-        blob_client.start_copy_from_url(sas_url, metadata=None, incremental_copy=False)
+            vmgs_url_path = ""
 
-        wait_copy_blob(blob_client, vhd_path, self._log)
+        wait_copy_blob(vhd_blob_client, vhd_url_path, self._log)
```
squirrelsc, line 265: Add comments to explain the reason to wait here, instead of earlier.

lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -178,22 +182,38 @@ def _prepare_virtual_machine(self, node: RemoteNode) -> None:
         startstop = node.features[StartStop]
         startstop.stop()
 
-    def _export_vhd(self, platform: AzurePlatform, virtual_machine: Any) -> str:
+    def _export_vhd(
+        self, platform: AzurePlatform, virtual_machine: Any
+    ) -> Tuple[str, str]:
```
squirrelsc: return a dict with name, the Tuple will be out of maintain. In future, it may supports data disk vhds.

lisa/sut_orchestrator/azure/transformers.py
kamalca: Maybe it would be more clear if the variable name was `has_vmgs`. That is what I am actually checking here. But yes, it is accurate that only a CVM will have a security profile on the OS Disk.

lisa/sut_orchestrator/azure/transformers.py
kamalca: It allows the copy of VMGS and VHD to happen in parallel. I will add a comment.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3550, opened 2024-12-09
Added ip_power control for BareMetal platform

paulli2017:


squirrelsc:
Please rebase on latest main to avoid conflict.

lisa/sut_orchestrator/baremetal/ip_power.py
```diff
@@ -0,0 +1,54 @@
...
+import requests
+
+from lisa import schema
+from lisa.util import ContextMixin, InitializableMixin, subclasses
+from lisa.util.logger import get_logger
+
+from .schema import IPPowerSchema
+
+
+class IPPower(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixin):
```
squirrelsc: Follow SerialConsole's design, implement the logic in the feature, and call it in cluster, when it's needed.

lisa/sut_orchestrator/baremetal/schema.py
```diff
@@ -73,6 +73,18 @@ class KeyLoaderSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
     type: str = field(default="build", metadata=field_metadata(required=True))
 
 
+@dataclass_json()
+@dataclass
+class IPPowerSchema(schema.TypedSchema, schema.ExtendableSchemaMixin):
+    type: str = field(default="Ip9285", metadata=field_metadata(required=True))
+    hostname: str = ""
```
squirrelsc: `host`

lisa/sut_orchestrator/baremetal/ip_power.py
```diff
@@ -0,0 +1,54 @@
...
+        raise NotImplementedError()
+
+    def off(self, port: int) -> None:
+        raise NotImplementedError()
+
+
+class Ip9285(IPPower):
+    def __init__(self, runbook: IPPowerSchema) -> None:
+        super().__init__(runbook)
+        self._request_cmd = (
```
squirrelsc: What does the `P6` mean? If it's part of the `port`, move it to on/off, so it's clearer.

lisa/sut_orchestrator/baremetal/ip_power.py
```diff
@@ -0,0 +1,54 @@
...
+class IPPower(subclasses.BaseClassWithRunbookMixin, ContextMixin, InitializableMixin):
+    def __init__(self, runbook: IPPowerSchema) -> None:
+        super().__init__(runbook=runbook)
+        self._log = get_logger("cluster", self.__class__.__name__)
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return IPPowerSchema
+
+    def on(self, port: int) -> None:
```
squirrelsc: Change the type of port to `str`, because maybe other type uses a name instead of an int.

lisa/sut_orchestrator/baremetal/ip_power.py
```diff
@@ -0,0 +1,54 @@
...
+    def type_name(cls) -> str:
+        return "Ip9285"
+
+    @classmethod
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return IPPowerSchema
+
+    def on(self, port: int) -> None:
+        request_on = f"{self._request_cmd}{port}=1"
+        requests.get(request_on)
```
squirrelsc: Follow the internal example to check the status, make sure it's reliable.

squirrelsc:
If you merge from main, please use `rebase`.

squirrelsc:
Use `git rebase -i` and `git add/reset -p` to merge commits to make each commit meaningful.

lisa/sut_orchestrator/baremetal/ip_power.py
squirrelsc: Implement it like StartStop feature, not another class.

lisa/sut_orchestrator/baremetal/ip_power.py
```diff
@@ -0,0 +1,68 @@
...
+    def type_schema(cls) -> Type[schema.TypedSchema]:
+        return IPPowerSchema
+
+    def on(self, port: str) -> None:
+        request_on = f"{self._request_cmd}{port}=1"
+        self._set_ip_power(request_on)
+
+    def off(self, port: str) -> None:
+        request_off = f"{self._request_cmd}{port}=0"
+        self._set_ip_power(request_off)
```
squirrelsc: How long to wait, to make sure the turn off stable?

lisa/sut_orchestrator/baremetal/ip_power.py
paulli2017: The change is done, thanks for your suggestions.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -103,6 +107,56 @@ def _connect(self) -> None:
...
+
+        self._request_cmd = (
+            f"http://{ip_power_runbook.host}/set.cmd?"
+            f"user={ip_power_runbook.username}+pass="
+            f"{ip_power_runbook.password}+cmd="
+            f"setpower+P6{ip_power_runbook.ctl_port}"
+        )
+
+    def _set_ip_power(self, power_cmd: str) -> None:
+        try:
```
squirrelsc: If there is no specific reason, don't overwrite the existing exception. The raw call trace is lost by doing this. Refer to below pattern.

```python
        response = requests.get(power_cmd, timeout=REQUEST_TIMEOUT)
        response.raise_for_status()
        self._log.debug(f"Command {power_cmd} done in set_ip_power")
```

lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -103,6 +107,56 @@ def _connect(self) -> None:
...
+        else:
+            self._log.debug(f"Command {power_cmd} done in set_ip_power")
+
+    def _stop(
+        self,
+        wait: bool = True,
+        state: features.StopState = features.StopState.Shutdown,
+    ) -> None:
+        request_off = f"{self._request_cmd}=0"
+        self._set_ip_power(request_off)
```
squirrelsc, line 145: As mentioned before, please check how long is safe to make sure the power off is totally done.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -126,11 +180,21 @@ def get_serial_console(self) -> Type[SerialConsole]:
                 f"is not supported."
             )
 
+    def get_start_stop(self) -> Type[StartStop]:
+        assert self.runbook.start_stop, "start_stop is not defined"
+        if self.runbook.start_stop.type == "Ip9285":
+            return Ip9285StartStop
+        else:
+            raise NotImplementedError(
+                f"start_stop type {self.runbook.start_stop.type} " f"is not supported."
```
squirrelsc:
```suggestion
                f"start_stop type {self.runbook.start_stop.type} is not supported."
```

lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -126,11 +180,21 @@ def get_serial_console(self) -> Type[SerialConsole]:
...
+            )
+
     def deploy(self, environment: Environment) -> Any:
         # connect to serial console
         for node in environment.nodes.list():
             # start serial console to save all log
             _ = node.features[features.SerialConsole]
+            _ = node.features[features.StartStop]
```
squirrelsc: It doesn't need to follow above SerialConsole line. The SerialConsole must collect log from beginning, so I wrote previous line. The StartStop feature doesn't need to do the same.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: From my local tests, it seems that 1s is good enough to make sure the power off is totally done, and here I used 3s (request_timeout) that should guarantee the power off is executed successfully. Thanks for your suggestion.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: The change is done, thanks a lot for your suggestions

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: The change is done. thanks

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/cluster/pxe.py
squirrelsc: Good to know, please add code comments to explain the 1s sleep. So if it's found unstable, people knows where to fix it.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -100,7 +105,52 @@ def _connect(self) -> None:
...
+        self._log.debug("Command {power_cmd} done in set_ip_power")
+
+    def _stop(
+        self,
+        wait: bool = True,
+        state: features.StopState = features.StopState.Shutdown,
+    ) -> None:
+        request_off = f"{self._request_cmd}=0"
+        self._set_ip_power(request_off)
+        sleep(REQUEST_TIMEOUT)
```
squirrelsc: It's not for "timeout" call it like `POWER_DOWN_DELAY`

lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -2,11 +2,14 @@
 # Licensed under the MIT license.
 
 from pathlib import Path
```
squirrelsc, line 4: Use `git rebase -i` and `git add/reset -p` to merge commits to make each commit meaningful.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: Comments are added, thanks

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: The change is done, thanks

lisa/sut_orchestrator/baremetal/cluster/pxe.py
squirrelsc: There are 16 commits, and from the names, it doesn't clear what's done inside. For each time reviews, please merge the commits, which address review comments. The commits are used to track changes for different purpose. It's ok to have multiple commits in a PR, and it's useful to understand the whole PR step by step. But it shouldn't be used for intermediate code review changes.

image

lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -100,7 +106,56 @@ def _connect(self) -> None:
...
+    def _stop(
+        self,
+        wait: bool = True,
+        state: features.StopState = features.StopState.Shutdown,
+    ) -> None:
+        request_off = f"{self._request_cmd}=0"
+        self._set_ip_power(request_off)
+        # To make sure power-off is fully settled down, it is recommended
+        # to wait a short time before power-on is triggered. Local tests showed
+        # 1s is actually good enough to get a stable power-off status, here
```
squirrelsc, line 148: request timeout is not used here, so the comment shouldn't mention it to avoid confusing.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: The change is done, thanks

squirrelsc (approved):

PR #3551, opened 2024-12-10
Enforce stricter target criteria for DPDK 

mcgov:
Enforcing supported targets to be CentOS, RedHat, Debian, Ubuntu, SUSE. Reducing ARM64 scope to Ubuntu only.
Add function enforcing PMD requirements for netvsc/failsafe/mana etc.


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3552, opened 2024-12-11
VhdTransformer: use vhd_path

kamalca:
use vhd_path instead of custom_blob_name because the custom blob name is not always provided.


kamalca:
Corrects a mistake in #3546

squirrelsc (approved):

PR #3553, opened 2024-12-12
Transformer Logs: Save to separate file

adityagesh:
Set transformer property 'enable_file_logging' to True to save the transformer logs to separate file

### DIFF
--- a/docs/run_test/runbook.rst
+++ b/docs/run_test/runbook.rst
@@ -524,6 +524,13 @@ its value will be overwritten by the transformer. For example,
 ``to_list_image`` to ``image``. The original variable name must exist in
 the output variables of the transformer.
 
+enable_file_logging
+^^^^^^^^^^^^^^^^^^^
+
+type: bool, optional, default is False.
+
+When set to True, the transformer will log the output to a file.
+
 .. _combinator:
 
 combinator
--- a/lisa/schema.py
+++ b/lisa/schema.py
@@ -187,6 +187,8 @@ class Transformer(TypedSchema, ExtendableSchemaMixin):
     rename: Dict[str, str] = field(default_factory=dict)
     # enable this transformer or not, only enabled transformers run actually.
     enabled: bool = True
+    # write logs to file. set to false by default
+    enable_file_logging: bool = False
     # decide when the transformer run. The init means run at very beginning
     # phase, which is before the combinator. The expanded means run after
     # combinator expanded variables.
--- a/lisa/transformer.py
+++ b/lisa/transformer.py
@@ -3,14 +3,23 @@
 
 import copy
 import functools
+from logging import FileHandler
+from pathlib import Path
 from typing import Any, Dict, List, Optional, Set
 
 from lisa import schema
 from lisa.environment import Environment
 from lisa.node import Node
 from lisa.parameter_parser.runbook import RunbookBuilder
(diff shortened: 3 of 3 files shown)


squirrelsc:
Is this PR to save transformer logs in a separated file, or the transformer logs are not in the whole log file?

adityagesh:
> Is this PR to save transformer logs in a separated file, or the transformer logs are not in the whole log file?

It is to save to a separate file. Corrected the title.

squirrelsc:
> > Is this PR to save transformer logs in a separated file, or the transformer logs are not in the whole log file?
> 
> It is to save to a separate file. Corrected the title.

Why does it need to save to separated files? Which scenario does it need? The current env/test split log is easy to understand, but transformer is a LISA concept. It would be confusing to have separated log.

adityagesh:
> > > Is this PR to save transformer logs in a separated file, or the transformer logs are not in the whole log file?
> > 
> > 
> > It is to save to a separate file. Corrected the title.
> 
> Why does it need to save to separated files? Which scenario does it need? The current env/test split log is easy to understand, but transformer is a LISA concept. It would be confusing to have separated log.

The primary motivation for this PR was the usage in kernel installer. I thought it makes sense to write the kernel installer log to a separate file. Alternatively, we can update the kernel installer transformer to write logs to a file, but I felt having a generic logging solution is better, in case others have similar use cases for different transformer

squirrelsc:
> 
> The primary motivation for this PR was the usage in kernel installer. I thought it makes sense to write the kernel installer log to a separate file. Alternatively, we can update the kernel installer transformer to write logs to a file, but I felt having a generic logging solution is better, in case others have similar use cases for different transformer

Thank you for telling where the separated log file is used. But I cannot understand why it's needed. In the most scenarios, the kernel installer transformer runs early stage with single thread. It should be easy view in the whole log.

adityagesh:
> > The primary motivation for this PR was the usage in kernel installer. I thought it makes sense to write the kernel installer log to a separate file. Alternatively, we can update the kernel installer transformer to write logs to a file, but I felt having a generic logging solution is better, in case others have similar use cases for different transformer
> 
> Thank you for telling where the separated log file is used. But I cannot understand why it's needed. In the most scenarios, the kernel installer transformer runs early stage with single thread. It should be easy view in the whole log.

When we want to process the log (eg: using AI) or upload the log (eg: upload the log of kernel build), it is difficult to read from the full lisa log

squirrelsc:
> > > The primary motivation for this PR was the usage in kernel installer. I thought it makes sense to write the kernel installer log to a separate file. Alternatively, we can update the kernel installer transformer to write logs to a file, but I felt having a generic logging solution is better, in case others have similar use cases for different transformer
> > 
> > 
> > Thank you for telling where the separated log file is used. But I cannot understand why it's needed. In the most scenarios, the kernel installer transformer runs early stage with single thread. It should be easy view in the whole log.
> 
> When we want to process the log (eg: using AI) or upload the log (eg: upload the log of kernel build), it is difficult to read from the full lisa log

For the 1 purpose: AI. Please write a filter to reduce log in the AI code, instead of a separated log in LISA. It's not flexible, if your AI requirement is changed.
For the 2 purpose: upload kernel build. Please dump a raw kernel build log without LISA prefix. It should a raw log like building in a bash.

PR #3555, opened 2024-12-13
Support placement type for disks

sharsonia:
Currently, Ephemeral OS Disk placement value is hardcoded in LISA. This PR enables LISA to fetch the Ephemeral OS disk placement type value from the SKU capability in Azure VMs.
To allow this, DiskPlacementType setting and supported values are added to the AzureDiskOptionSettings.
LISA user can also pass the EphemeralDiskPlacementType from Runbook to deploy VM with specific supported placement type value.

For Ephemeral OS Disk placement types available, refer
 https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks-faq

### DIFF
--- a/lisa/sut_orchestrator/azure/arm_template.bicep
+++ b/lisa/sut_orchestrator/azure/arm_template.bicep
@@ -98,7 +98,7 @@ func getEphemeralOSImage(node object) object => {
   name: '${node.name}-osDisk'
   diffDiskSettings: {
       option: 'local'
-      placement: 'CacheDisk'
+      placement: node.ephemeral_disk_placement_type
   }
   caching: 'ReadOnly'
   createOption: 'FromImage'
@@ -333,7 +333,7 @@ resource nodes_data_disks 'Microsoft.Compute/disks@2022-03-02' = [
   }
 ]
 
-resource nodes_vms 'Microsoft.Compute/virtualMachines@2022-08-01' = [for i in range(0, node_count): {
+resource nodes_vms 'Microsoft.Compute/virtualMachines@2024-03-01' = [for i in range(0, node_count): {
   name: nodes[i].name
   location: nodes[i].location
   tags: combined_vm_tags
--- a/lisa/sut_orchestrator/azure/features.py
+++ b/lisa/sut_orchestrator/azure/features.py
@@ -98,6 +98,7 @@ from .common import (
     AzureCapability,
     AzureImageSchema,
     AzureNodeSchema,
+    DiskPlacementType,
     check_or_create_storage_account,
     create_update_private_dns_zone_groups,
     create_update_private_endpoints,
@@ -113,6 +114,7 @@ from .common import (
     delete_virtual_network_links,
     find_by_name,
     get_compute_client,
+    get_disk_placement_priority,
     get_network_client,
     get_node_context,
     get_or_create_file_share,
@@ -1308,6 +1310,31 @@ _disk_size_performance_map: Dict[schema.DiskType, List[Tuple[int, int, int]]] = 
 @dataclass()
 class AzureDiskOptionSettings(schema.DiskOptionSettings):
     has_resource_disk: Optional[bool] = None
+    ephemeral_disk_placement_type: Optional[
+        Union[search_space.SetSpace[DiskPlacementType], DiskPlacementType]
+    ] = field(  # type:ignore
+        default_factory=partial(
+            search_space.SetSpace,
+            items=[
+                DiskPlacementType.NONE,
+                DiskPlacementType.RESOURCE,
+                DiskPlacementType.CACHE,
+                DiskPlacementType.NVME,
+            ],
+        ),
+        metadata=field_metadata(
+            decoder=partial(
+                search_space.decode_nullable_set_space,
+                base_type=DiskPlacementType,
+                default_values=[
+                    DiskPlacementType.NONE,
+                    DiskPlacementType.RESOURCE,
+                    DiskPlacementType.CACHE,
+                    DiskPlacementType.NVME,
+                ],
+            )
+        ),
+    )
 
     def __hash__(self) -> int:
         return super().__hash__()
@@ -1316,7 +1343,11 @@ class AzureDiskOptionSettings(schema.DiskOptionSettings):
         return self.__repr__()
 
     def __repr__(self) -> str:
-        return f"has_resource_disk: {self.has_resource_disk},{super().__repr__()}"
+        return (
+            f"has_resource_disk: {self.has_resource_disk},"
+            f"ephemeral_disk_placement_type: {self.ephemeral_disk_placement_type},"
+            f"{super().__repr__()}"
+        )
 
     def __eq__(self, o: object) -> bool:
         if not super().__eq__(o):
@@ -1375,6 +1406,13 @@ class AzureDiskOptionSettings(schema.DiskOptionSettings):
             "has_resource_disk",
         )
         result.merge(
+            search_space.check_setspace(
+                self.ephemeral_disk_placement_type,
+                capability.ephemeral_disk_placement_type,
+            ),
+            "ephemeral_disk_placement_type",
+        )
+        result.merge(
             search_space.check_countspace(
                 self.max_data_disk_count, capability.max_data_disk_count
             ),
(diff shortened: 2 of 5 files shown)


lisa/schema.py
```diff
@@ -446,11 +446,22 @@ class ResourceDiskType(str, Enum):
     NVME = constants.STORAGE_INTERFACE_TYPE_NVME
 
 
+class EphemeralDiskPlacementType(str, Enum):
+    Resource = constants.DISK_PLACEMENT_TYPE_RESOURCE
```
squirrelsc: Please provide source url of types.

sharsonia: > Please provide source url of types.

Added in the constants.

squirrelsc:
> Currently, Ephemeral OS Disk placement value is hardcoded in LISA. This PR enables LISA to fetch the Ephemeral OS disk placement type value from the SKU capability in Azure VMs. To allow this, EphemeralDiskPlacementType setting and supported values are added to the DiskOptionSettings in the schema. LISA user can also pass the EphemeralDiskPlacementType from Runbook to deploy VM with specific supported placement type value.
> 
> For Ephemeral OS Disk placement types available, refer https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks-faq

How is the placement type used for? If it's set in runbook only, it doesn't need to be in the feature level.

lisa/features/disks.py
```diff
@@ -54,6 +54,9 @@ def get_all_disks(self) -> List[str]:
     def get_hardware_disk_controller_type(self) -> schema.DiskControllerType:
         raise NotImplementedError
 
+    def get_ephemeral_disk_placement_type(self) -> schema.EphemeralDiskPlacementType:
```
squirrelsc: It looks an Azure only feature, please add it in the inherited Azure disk feature.

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -98,7 +98,7 @@ func getEphemeralOSImage(node object) object => {
   name: '${node.name}-osDisk'
   diffDiskSettings: {
       option: 'local'
-      placement: 'CacheDisk'
+      placement: node.ephemeral_disk_placement_type
```
squirrelsc, line 101: Please check with Lili, how to apply bicep changes. It needs to compile and commit arm template changes together.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1447,6 +1447,32 @@ def _call_requirement_method(
                 self.os_disk_size, capability.os_disk_size
             )
 
+        if value.os_disk_type == schema.DiskType.Ephemeral:
```
squirrelsc: Add comments above to explain why the logic only run for ephemeral os disk.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1665,6 +1691,11 @@ def get_hardware_disk_controller_type(self) -> Any:
         vm = get_vm(azure_platform, self._node)
         return vm.storage_profile.disk_controller_type
 
+    def get_ephemeral_disk_placement_type(self) -> Any:
```
squirrelsc: The method name is not accurate, it should mention it's for os disk.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1794,14 +1817,40 @@ def _resource_sku_to_capability(  # noqa: C901
...
+            # refer
+            # https://learn.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks-faq
+            resource_disk_bytes = azure_raw_capabilities.get("MaxResourceVolumeMB", 0)
             cached_disk_bytes = azure_raw_capabilities.get("CachedDiskBytes", 0)
-            cached_disk_bytes_gb = int(int(cached_disk_bytes) / 1024 / 1024 / 1024)
-            if cached_disk_bytes_gb >= 30:
-                node_space.disk.os_disk_type.add(schema.DiskType.Ephemeral)
-                node_space.disk.os_disk_size = cached_disk_bytes_gb
+            nvme_disk_bytes = azure_raw_capabilities.get("NvmeDiskSizeInMiB", 0)
+            if nvme_disk_bytes:
```
squirrelsc, line 1879: Does it mean Nvme disk size is higher than normal disk size? Is it possible the two settings co-exists? If they cannot be co-exists, it needs to check and assert.

squirrelsc, line 1879: Why the original `if cached_disk_bytes_gb >= 30:` is removed?

lisa/sut_orchestrator/azure/features.py
sharsonia: > The method name is not accurate, it should mention it's for os disk.

Hmm actually in Azure, it looks like disk placement is configured only for Ephemeral disk type. 
See here - https://learn.microsoft.com/en-us/powershell/module/az.compute/set-azvmssstorageprofile?view=azps-13.0.0
But if you think, the functional name should be generic, I can change it. let me know your preference after reading through above link.

lisa/sut_orchestrator/azure/platform_.py
sharsonia: > Does it mean Nvme disk size is higher than normal disk size? Is it possible the two settings co-exists? If they cannot be co-exists, it needs to check and assert.

NVMe disk size is higher and preferred if it is present. I know that "CacheDisk" and "ResourceDisk" can co-exist. As of now, I haven't seen "NVMeDisk" co-exist with any other, but I am not ruling it out

For example - 
```
(Get-AzComputeResourceSku -Location EastUS2euap | ? {$_.Name -like "Standard_D2s_v3"}).Capabilities

Name                                                              Value
----                                                                   -----
MaxResourceVolumeMB                                         16384
OSVhdSizeMB                                                         1047552
vCPUs                                                                      2
MemoryPreservingMaintenanceSupported            True
HyperVGenerations                                                 V1,V2
SupportedEphemeralOSDiskPlacements                ResourceDisk,CacheDisk
CachedDiskBytes                                                     53687091200
```

lisa/sut_orchestrator/azure/features.py
squirrelsc: But the L1697 called for `os_disk`.
```vm.storage_profile.**os_disk**.diff_disk_settings.placement```

File: lisa/sut_orchestrator/azure/arm_template.json [hunk 7]
```diff
@@ -0,0 +1,881 @@
+{
```
squirrelsc: Remove this file.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -1475,6 +1509,34 @@ def _call_requirement_method(
             schema.disk_controller_type_priority,
         )
 
+        # refer
+        # https://learn.microsoft.com/en-us/powershell/module/az.compute/set-azvmssstorageprofile?view=azps-13.0.0               # noqa: E501
```
squirrelsc: It's good to have the urls, please summarize to avoid people must visit links to know details. In case the links are expired, nothing can help understand it.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1433,6 +1436,18 @@ def _create_node_arm_parameters(
             )
         arm_parameters.disk_controller_type = capability.disk.disk_controller_type.value
 
+        assert isinstance(
+            features.AzureDiskOptionSettings(
+                capability.disk
+            ).ephemeral_disk_placement_type,
+            DiskPlacementType,
+        )
```
squirrelsc, line 1479: follow other `isinstance` assertions to print actual types, in case it fails.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2754,6 +2796,14 @@ def _set_disk_features(
             isinstance(node_space.disk.os_disk_type, search_space.SetSpace)
             and node_space.disk.os_disk_type.isunique(schema.DiskType.Ephemeral)
         ):
+            node_space.disk = features.AzureDiskOptionSettings(node_space.disk)
```
squirrelsc: Why create a new disk setting here?

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3556, opened 2024-12-13
Marked all sockperf tests to skip

r-dailey:
Since sockperf compilation is not possible without changes to the sockperf tool or the compiler these tests are to be skipped until sockperf can be compiled natively.


squirrelsc (dismissed):

squirrelsc:
Please add more your valuable investigation, including which c++ compiler is used by sockperf, why it cannot support FreeBSD, how it supposedly to be fixed, and so on.

squirrelsc (approved):

PR #3559, opened 2024-12-17
Use new environment for 'verify_gpu_rescind_validation' test case.

SRIKKANTH:
Most of the failures of 'verify_gpu_rescind_validation' are not reproducing when executed this test case in the new environment.


LiliDeng:
@SRIKKANTH can we do cleanup work to recover the environment, then it reduces the impact for the following case?

squirrelsc:
> Most of the failures of 'verify_gpu_rescind_validation' are not reproducing when executed this test case in the new environment.

What's error with existing environment?

PR #3573, opened 2024-12-30
Windows support for Service tool added.

SRIKKANTH:
This pull request adds Windows support for the Service tool.

### DIFF
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -244,6 +244,9 @@ class WindowsService(Tool):
             if "Cannot find any service with service name" in str(identifier):
                 self._log.debug(f"service '{name}' does not exist")
                 return
+            elif "cannot be stopped due to the following error" in str(identifier):
+                self._log.debug(f"service '{name}' cannot be stopped")
+                return
             raise identifier
         self.wait_for_service_start(name)
 
@@ -257,6 +260,9 @@ class WindowsService(Tool):
             if "Cannot find any service with service name" in str(identifier):
                 self._log.debug(f"service '{name}' does not exist")
                 return
+            elif "cannot be stopped due to the following error" in str(identifier):
+                self._log.debug(f"service '{name}' cannot be stopped")
+                return
             raise identifier
         self.wait_for_service_stop(name)
 
@@ -266,15 +272,15 @@ class WindowsService(Tool):
     def wait_for_service_stop(self, name: str) -> None:
         self._wait_for_service(name, WindowsServiceStatus.STOPPED)
 
-    def _check_exists(self) -> bool:
-        return True
-
     def check_service_exists(self, name: str) -> bool:
         if (
             self._get_status(name, fail_on_error=False)
             == WindowsServiceStatus.NOT_FOUND
         ):
             return False
+        return True
+
+    def _check_exists(self) -> bool:
         return True
 
     def _get_status(
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -241,9 +241,7 @@ class WindowsService(Tool):
                 force_run=True,
             )
         except LisaException as identifier:
-            if "Cannot find any service with service name" in str(identifier):
-                self._log.debug(f"service '{name}' does not exist")
-            elif "cannot be stopped due to the following error" in str(identifier):
+            if "cannot be stopped due to the following error" in str(identifier):
                 self._log.debug(f"service '{name}' cannot be stopped")
             raise identifier
         self.wait_for_service_start(name)
@@ -255,9 +253,7 @@ class WindowsService(Tool):
                 force_run=True,
             )
         except LisaException as identifier:
-            if "Cannot find any service with service name" in str(identifier):
-                self._log.debug(f"service '{name}' does not exist")
-            elif "cannot be stopped due to the following error" in str(identifier):
+            if "cannot be stopped due to the following error" in str(identifier):
                 self._log.debug(f"service '{name}' cannot be stopped")
             raise identifier
         self.wait_for_service_stop(name)
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -235,27 +235,17 @@ class WindowsService(Tool):
         pass
 
     def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
-        try:
-            self.node.tools[PowerShell].run_cmdlet(
-                f"Restart-service {name}",
-                force_run=True,
-            )
-        except LisaException as identifier:
-            if "cannot be stopped due to the following error" in str(identifier):
-                self._log.debug(f"service '{name}' cannot be stopped")
-            raise identifier
+        self.node.tools[PowerShell].run_cmdlet(
+            f"Restart-service {name}",
+            force_run=True,
+        )
         self.wait_for_service_start(name)
 
     def stop_service(self, name: str) -> None:
-        try:
-            self.node.tools[PowerShell].run_cmdlet(
-                f"Stop-Service {name} -Force",
-                force_run=True,
-            )
-        except LisaException as identifier:
-            if "cannot be stopped due to the following error" in str(identifier):
-                self._log.debug(f"service '{name}' cannot be stopped")
-            raise identifier
+        self.node.tools[PowerShell].run_cmdlet(
+            f"Stop-Service {name} -Force",
+            force_run=True,
+        )
         self.wait_for_service_stop(name)
 
     def wait_for_service_start(self, name: str) -> None:
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -243,10 +243,8 @@ class WindowsService(Tool):
         except LisaException as identifier:
             if "Cannot find any service with service name" in str(identifier):
                 self._log.debug(f"service '{name}' does not exist")
-                return
             elif "cannot be stopped due to the following error" in str(identifier):
                 self._log.debug(f"service '{name}' cannot be stopped")
-                return
             raise identifier
         self.wait_for_service_start(name)
 
@@ -259,10 +257,8 @@ class WindowsService(Tool):
         except LisaException as identifier:
             if "Cannot find any service with service name" in str(identifier):
                 self._log.debug(f"service '{name}' does not exist")
-                return
             elif "cannot be stopped due to the following error" in str(identifier):
                 self._log.debug(f"service '{name}' cannot be stopped")
-                return
             raise identifier
         self.wait_for_service_stop(name)
 
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -219,6 +219,7 @@ class WindowsServiceStatus(int, Enum):
     START_PENDING = 2
     STOP_PENDING = 3
     STOPPED = 1
+    NOT_FOUND = 0
 
 
 class WindowsService(Tool):
@@ -258,29 +259,30 @@ class WindowsService(Tool):
     def _check_exists(self) -> bool:
         return True
 
-    def _check_service_exists(self, name: str) -> bool:
-        try:
-            self._get_status(name)
-            return True
-        except LisaException:
+    def check_service_exists(self, name: str) -> bool:
+        if (
+            self._get_status(name, fail_on_error=False)
+            == WindowsServiceStatus.NOT_FOUND
+        ):
             return False
+        return True
 
-    def _check_service_running(self, name: str) -> bool:
-        return self._get_status(name) == WindowsServiceStatus.RUNNING
-
-    def _get_status(self, name: str = "") -> WindowsServiceStatus:
-        service_status = self.node.tools[PowerShell].run_cmdlet(
-            f"Get-Service {name}",
-            force_run=True,
-            output_json=True,
-            fail_on_error=False,
-        )
-        if not service_status:
-            raise LisaException(f"service '{name}' does not exist")
-        return WindowsServiceStatus(int(service_status["Status"]))
-
-    def _is_service_inactive(self, name: str) -> bool:
-        return self._get_status(name) == WindowsServiceStatus.STOPPED
+    def _get_status(
+        self, name: str = "", fail_on_error: bool = True
+    ) -> WindowsServiceStatus:
+        try:
+            service_status = self.node.tools[PowerShell].run_cmdlet(
+                f"Get-Service {name}",
+                force_run=True,
+                output_json=True,
+            )
+            return WindowsServiceStatus(int(service_status["Status"]))
+        except LisaException as identifier:
+            if "Cannot find any service with service name" in str(identifier):
+                if fail_on_error:
+                    raise LisaException(f"service '{name}' does not exist")
+                return WindowsServiceStatus.NOT_FOUND
+            raise identifier
 
     def _wait_for_service(self, name: str, status: WindowsServiceStatus) -> None:
         timeout = 60
@@ -294,5 +296,6 @@ class WindowsService(Tool):
 
         if timeout < timer.elapsed():
             raise LisaException(
-                f"service '{name}' still in '{current_service_status}' state after '{timeout}' seconds"  # noqa: E501
+                f"service '{name}' still in '{current_service_status}' state"
+                f"after '{timeout}' seconds"
             )
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -1,11 +1,16 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 import re
-from typing import Type
+from enum import Enum
+from time import sleep
+from typing import Optional, Type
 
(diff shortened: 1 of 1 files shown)


lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
     def is_service_running(self, name: str) -> bool:
         return self._internal_tool._check_service_running(name)  # type: ignore
 
+    def wait_for_service_start(self, name: str) -> None:
```
squirrelsc: It doesn't need to be a common method in the base class.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+        return False
+
+    @property
+    def command(self) -> str:
+        return self._command
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
```
squirrelsc: Move public methods above private methods.

lisa/schema.py
```diff
@@ -431,6 +431,16 @@ class DiskType(str, Enum):
 ]
 
 
+class WindowsServiceStatus(int, Enum):
```
squirrelsc: Move it along in to the service class.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+            )
+            if schema.WindowsServiceStatus.RUNNING == schema.WindowsServiceStatus(
+                service_status["Status"]
+            ):
+                return
+
+            self._log.debug(
+                f"service '{name}' is not ready yet, retrying... after 3 seconds"
+            )
+            sleep(3)
```
squirrelsc: Change the interval to 0.5 second.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+
+    def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
+        self._powershell.run_cmdlet(
+            f"Restart-service {name}",
+            force_run=True,
+        )
+        self.wait_for_service_start(name)
+
+    def wait_for_service_start(self, name: str) -> None:
+        for _ in range(10):
```
squirrelsc: Create a builtin Timer by `create_timer` to count the timeout by seconds, instead of count by times.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+                force_run=True,
+                output_json=True,
+                fail_on_error=False,
+            )
+            if schema.WindowsServiceStatus.RUNNING == schema.WindowsServiceStatus(
+                service_status["Status"]
+            ):
+                return
+
+            self._log.debug(
```
squirrelsc: Output before the waiting once, not output during waiting.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+                f"service '{name}' is not ready yet, retrying... after 3 seconds"
+            )
+            sleep(3)
+
+        raise LisaException(f"service '{name}' failed to start")
```
squirrelsc: Give more details, like wait 30 seconds timeout, it's still in NNN status.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+
+    def _get_status(self, name: str = "") -> schema.WindowsServiceStatus:
+        service_status = self._powershell.run_cmdlet(
+            f"Get-Service {name}",
+            force_run=True,
+            output_json=True,
+            fail_on_error=False,
+        )
+        if not service_status:
```
squirrelsc: Use above fail_on_error, instead of custom error.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+    def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
+        self._powershell.run_cmdlet(
+            f"Restart-service {name}",
+            force_run=True,
+        )
+        self.wait_for_service_start(name)
+
+    def wait_for_service_start(self, name: str) -> None:
+        for _ in range(10):
+            service_status = self._powershell.run_cmdlet(
```
squirrelsc: Use the `_get_status`, not to implement it again.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+            f"Get-Service {name}",
+            force_run=True,
+            output_json=True,
+            fail_on_error=False,
+        )
+        if not service_status:
+            raise LisaException(f"service '{name}' does not exist")
+        return schema.WindowsServiceStatus(service_status["Status"])
+
+    def stop_service(self, name: str) -> None:
```
squirrelsc: Wait on stop too.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+        self._powershell.run_cmdlet(
+            f"Stop-Service {name} -Force",
+            force_run=True,
+            output_json=True,
+            fail_on_error=False,
+        )
+        assert_that(self._get_status(name)).described_as(
+            f"Failed to stop service {name}"
+        ).is_not_equal_to(schema.WindowsServiceStatus.RUNNING)
```
squirrelsc: It doesn't make sense, why stop service expect the service status is running?

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+
+    def enable_service(self, name: str) -> None:
+        raise NotImplementedError()
```
squirrelsc: use `pass` instead of raise exception.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+
+    def check_service_status(self, name: str) -> bool:
```
squirrelsc, line 143: Implement the internal `_check_service_running`, instead of replacing the public method.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
...
+        return self._get_status(name) == schema.WindowsServiceStatus.RUNNING
+
+    def check_service_exists(self, name: str) -> bool:
+        try:
+            self._get_status(name)
+            return True
+        except LisaException:
+            return False
+
+    def is_service_inactive(self, name: str) -> bool:
```
squirrelsc, line 153: Implement the private method, not public methods.

lisa/base_tools/service.py
```diff
@@ -58,6 +68,95 @@ def is_service_inactive(self, name: str) -> bool:
     def is_service_running(self, name: str) -> bool:
         return self._internal_tool._check_service_running(name)  # type: ignore
 
+    def wait_for_service_start(self, name: str) -> None:
+        raise NotImplementedError()
+
+
+class WindowsService(Tool):
```
squirrelsc: Move Windows class to the end like other tools.

lisa/schema.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Fixed

SRIKKANTH:
This PR has dependency on https://github.com/microsoft/lisa/pull/3574

lisa/base_tools/service.py
```diff
@@ -200,3 +209,90 @@ def _check_service_running(self, name: str) -> bool:
...
+        self._log.debug(f"waiting for service '{name}' to be in '{status}' state")
+        while timeout > timer.elapsed(False):
+            current_service_status = self._get_status(name)
+            if status == current_service_status:
+                return
+            sleep(0.5)
+
+        if timeout < timer.elapsed():
+            raise LisaException(
+                f"service '{name}' still in '{current_service_status}' state after '{timeout}' seconds"  # noqa: E501
```
squirrelsc: Change it to multiple lines, not to overuse E501.
```python
f"service '{name}' still in '{current_service_status}' state "
f"after '{timeout}' seconds"
```

lisa/base_tools/service.py
```diff
@@ -200,3 +209,90 @@ def _check_service_running(self, name: str) -> bool:
...
+            force_run=True,
+            output_json=True,
+            fail_on_error=False,
+        )
+        if not service_status:
+            raise LisaException(f"service '{name}' does not exist")
+        return WindowsServiceStatus(int(service_status["Status"]))
+
+    def _is_service_inactive(self, name: str) -> bool:
+        return self._get_status(name) == WindowsServiceStatus.STOPPED
```
squirrelsc: Should it use the same result as `_check_service_running`? If the service is in other statuses, what's expected?

lisa/base_tools/service.py
squirrelsc: it looks not fixed. Set above `fail_on_error=True`, let the above line failed. If the error message is not clear, catch it and add more information.

lisa/base_tools/service.py
```diff
@@ -200,3 +209,90 @@ def _check_service_running(self, name: str) -> bool:
...
+
+    def _check_exists(self) -> bool:
+        return True
+
+    def _check_service_exists(self, name: str) -> bool:
+        try:
+            self._get_status(name)
+            return True
+        except LisaException:
+            return False
```
squirrelsc: 1. When checking the service exist or not, use more accurate checking, like parse output, instead of getting status, and judge non-exists on any exception.
2. Not overuse try/except on normal status.

lisa/base_tools/service.py
```diff
@@ -200,3 +209,90 @@ def _check_service_running(self, name: str) -> bool:
...
+    def stop_service(self, name: str) -> None:
+        self.node.tools[PowerShell].run_cmdlet(
+            f"Stop-Service {name} -Force",
+            force_run=True,
+            output_json=True,
+            fail_on_error=False,
+        )
+        self.wait_for_service_stop(name)
+
+    def wait_for_service_start(self, name: str) -> None:
```
squirrelsc, line 251: If the method is not used outside of the class, remove them.

lisa/base_tools/service.py
SRIKKANTH: Fixed as suggested

lisa/base_tools/service.py
SRIKKANTH: Fixed

lisa/base_tools/service.py
SRIKKANTH: Removed this method and few others which I am not going to use in my next patch. I will add them I started using them.

lisa/base_tools/service.py
squirrelsc: Please follow the suggestion `fail_on_error=True`, so it raise exceptions if any error met.

lisa/base_tools/service.py
```diff
@@ -200,3 +209,93 @@ def _check_service_running(self, name: str) -> bool:
...
+            return WindowsServiceStatus(int(service_status["Status"]))
+        except LisaException as identifier:
+            if "Cannot find any service with service name" in str(identifier):
+                if fail_on_error:
+                    raise LisaException(f"service '{name}' does not exist")
+                return WindowsServiceStatus.NOT_FOUND
+            raise identifier
+
+    def _wait_for_service(self, name: str, status: WindowsServiceStatus) -> None:
+        timeout = 60
```
squirrelsc: Move the timeout to a parameter. 30 seconds should be enough.

lisa/base_tools/service.py
```diff
@@ -200,3 +209,93 @@ def _check_service_running(self, name: str) -> bool:
...
+    def _get_status(
+        self, name: str = "", fail_on_error: bool = True
+    ) -> WindowsServiceStatus:
+        try:
+            service_status = self.node.tools[PowerShell].run_cmdlet(
+                f"Get-Service {name}",
+                force_run=True,
+                output_json=True,
+            )
+            return WindowsServiceStatus(int(service_status["Status"]))
```
squirrelsc: Move this line to the end of the method, so it's easy to read.

SRIKKANTH: Fixed as suggested

lisa/base_tools/service.py
SRIKKANTH: Fixed as suggested

lisa/base_tools/service.py
```diff
@@ -235,19 +235,29 @@ def enable_service(self, name: str) -> None:
...
-            output_json=True,
-            fail_on_error=False,
-        )
+        try:
+            self.node.tools[PowerShell].run_cmdlet(
+                f"Stop-Service {name} -Force",
+                force_run=True,
+            )
+        except LisaException as identifier:
+            if "Cannot find any service with service name" in str(identifier):
```
squirrelsc: With this error message, when a non-exist service will be stop or restarted in a normal state? if the service doesn't exist, it should fail earlier, instead of stop/start it without errors.

In case it's really needed, please call check_service_exists before stop/start, and raise all errors. The code is clean and easy to read. But before apply like this, please explain when a non-exist service is started or stopped.

SRIKKANTH: Adding an extra call to check_service_exists will add latency. Since restart/stop can capture it as part of the reason for failure. I think we can keep this message. 

I added extra messages to capture reasons for not stopping a service. One of the reasons for not able to stop is dependent services.

squirrelsc: The up-layer logic should make sure start/stop are not called, if the service doesn't exist, instead of eating the exception here. It may hide the problem, and fails on other steps and confusing. So it doesn't need to eat any exception here, just let the code raise as is.

When the service is used, if the service doesn't exist, the test case level should fix it or skip the test case, instead of continuing to start/stop service or any other logic.

SRIKKANTH: Makes sense. Removed the service not found scenario from the stop, restart methods.

lisa/base_tools/service.py
```diff
@@ -200,3 +209,102 @@ def _check_service_running(self, name: str) -> bool:
...
+    def enable_service(self, name: str) -> None:
+        pass
+
+    def restart_service(self, name: str, ignore_exit_code: int = 0) -> None:
+        try:
+            self.node.tools[PowerShell].run_cmdlet(
+                f"Restart-service {name}",
+                force_run=True,
+            )
+        except LisaException as identifier:
```
squirrelsc: The try/except can be removed totally. Just let the code raise the exception, instead of wrap it. Same to stop_service.

SRIKKANTH: Removed.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
Please fix CI check error @SRIKKANTH

LiliDeng:
@SRIKKANTH please rebase latest main, it will fix the CI check error.

SRIKKANTH:
> @SRIKKANTH please rebase latest main, it will fix the CI check error.

Fixed all CI check errors. please complete the PR @LiliDeng

LiliDeng (approved):

PR #3574, opened 2024-12-31
powershell - ouput as json

SRIKKANTH:
This patch enables the `run_cmdlet` method in `lisa/tools/powershell.py` to output the result as JSON when the `output_json` parameter is set to `True`.

Getting Powershell output as JSON is beneficial because it allows for easier parsing and processing of the output data.

### DIFF
--- a/lisa/tools/ls.py
+++ b/lisa/tools/ls.py
@@ -88,7 +88,7 @@ class WindowsLs(Ls):
             sudo=sudo,
         )
 
-        return output.strip() == "True"
+        return bool(output.strip() == "True")
 
     def path_exists(self, path: str, sudo: bool = False) -> bool:
         output = self.node.tools[PowerShell].run_cmdlet(
@@ -96,7 +96,7 @@ class WindowsLs(Ls):
             force_run=True,
             sudo=sudo,
         )
-        return output.strip() == "True"
+        return bool(output.strip() == "True")
 
     def list(self, path: str, sudo: bool = False) -> List[str]:
         command = f'Get-ChildItem -Path "{path}" | Select-Object -ExpandProperty Name'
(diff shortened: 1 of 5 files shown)


lisa/tools/powershell.py
```diff
@@ -48,14 +48,17 @@ def run_cmdlet_async(
...
+        output_json: bool = False,
         force_run: bool = False,
         sudo: bool = False,
         fail_on_error: bool = True,
         timeout: int = 600,
         # Powershell error log is the xml format, it needs extra decoding. But
         # for long running script, it needs to look real time results.
         no_debug_log: bool = True,
-    ) -> str:
+    ) -> Any:
```
squirrelsc, line 59: Don't use Any type. Use Union to specify str and json.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@SRIKKANTH please fix the CI check error.

SRIKKANTH:
> @SRIKKANTH please fix the CI check error.

All issues are fixed. @LiliDeng

LiliDeng (approved):

lisa/tools/ls.py
```diff
@@ -106,7 +106,7 @@ def list(self, path: str, sudo: bool = False) -> List[str]:
             sudo=sudo,
         )
         if output:
-            return output.split()
+            return List(output.split())
```
pupacha, line 109: @SRIKKANTH  What is the reason for doing `List()` here ? It actually is not instantiate-able. It throws following error.

 ```
 File "XYZ/lisa/lisa/tools/ls.py", line 109, in list
    return List(output.split())
  File "XYZ/Python/3.10.15/x64/lib/python3.10/typing.py", line 955, in __call__
    raise TypeError(f"Type {self._name} cannot be instantiated; "
TypeError: Type List cannot be instantiated; use list() instead

```

pupacha: ok, I see this is fixed in main now.

PR #3577, opened 2025-01-01
Added setup commands for Chrony test on FreeBSD

r-dailey:
Added commands to enable chrony on freebsd into the chrony test case. Tested and passed on all 4 images.

### DIFF
--- a/microsoft/testsuites/core/timesync.py
+++ b/microsoft/testsuites/core/timesync.py
@@ -24,8 +24,10 @@ from lisa import (
 from lisa.operating_system import BSD, CpuArchitecture, Redhat, Suse, Windows
 from lisa.tools import (
     Cat,
+    Chmod,
     Chrony,
     Dmesg,
+    Echo,
     Hwclock,
     Ls,
     Lscpu,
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/timesync.py
```diff
@@ -414,6 +414,11 @@ def verify_timesync_ntp(self, node: Node) -> None:
         priority=2,
     )
     def verify_timesync_chrony(self, node: Node) -> None:
+        # On FreeBSD images, Chrony is not enabled by default after installation.
+        # To enabled it the line chronyd_enable=YES must be added to /etc/rc.conf.
+        if isinstance(node.os, BSD):
+            node.execute("chmod +x /etc/rc.conf", sudo=True)
+            node.execute("bash -c 'echo chronyd_enable=YES >> /etc/rc.conf'", sudo=True)
```
LiliDeng: use chmod and echo tool

r-dailey: Ok will do!

LiliDeng (approved):

PR #3578, opened 2025-01-02
HyperV improvements for DHCP and default switch

SRIKKANTH:
Added below new functionality:
    1. get_default_internal_switch
    2. configure_dhcp 

These changes improve the HyperV tool by adding new functionality to get the default internal switch and configure DHCP to set HyperV environment in an Azure Windows Server Host. 

Reference doc:
https://techcommunity.microsoft.com/blog/itopstalkblog/how-to-setup-nested-virtualization-for-azure-vmvhd/1115338

### DIFF
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -68,7 +68,7 @@ class Service(Tool):
         return self._internal_tool._check_service_running(name)  # type: ignore
 
     def wait_for_service_start(self, name: str) -> None:
-        pass
+        raise NotImplementedError("'wait_for_service_start' is not implemented")
 
 
 class ServiceInternal(Tool):
--- a/lisa/base_tools/service.py
+++ b/lisa/base_tools/service.py
@@ -67,6 +67,9 @@ class Service(Tool):
     def is_service_running(self, name: str) -> bool:
         return self._internal_tool._check_service_running(name)  # type: ignore
 
+    def wait_for_service_start(self, name: str) -> None:
+        pass
+
 
 class ServiceInternal(Tool):
     @property
--- a/lisa/tools/hyperv.py
+++ b/lisa/tools/hyperv.py
@@ -4,14 +4,17 @@
 import re
 import time
 from dataclasses import dataclass, field
+from enum import Enum
 from typing import Dict, Optional
 
 from assertpy import assert_that
 from dataclasses_json import config, dataclass_json
 
+from lisa.base_tools import Service
 from lisa.executable import Tool
 from lisa.operating_system import Windows
 from lisa.tools.powershell import PowerShell
+from lisa.tools.windows_feature import WindowsFeatureManagement
 from lisa.util import LisaException
 from lisa.util.process import Process
 
@@ -20,6 +23,11 @@ from lisa.util.process import Process
 @dataclass
 class VMSwitch:
     name: str = field(metadata=config(field_name="Name"))
+
+
+class HypervSwitchType(Enum):
+    INTERNAL = "Internal"
+    EXTERNAL = "External"
 
 
 class HyperV(Tool):
@@ -201,15 +209,21 @@ class HyperV(Tool):
             force_run=True,
         )
 
-    def get_default_external_switch(self) -> Optional[VMSwitch]:
+    def get_default_switch(
+        self, switch_type: HypervSwitchType = HypervSwitchType.EXTERNAL
+    ) -> Optional[VMSwitch]:
+        if switch_type not in (HypervSwitchType.INTERNAL, HypervSwitchType.EXTERNAL):
+            raise LisaException(f"Unknown switch type {switch_type}")
+
+        # get default switch of type `switch_type` from hyperv
         switch_json = self.node.tools[PowerShell].run_cmdlet(
-            'Get-VMSwitch | Where-Object {$_.SwitchType -eq "External"} '
+            f'Get-VMSwitch | Where-Object {{$_.SwitchType -eq "{switch_type}"}}'
             "| Select -First 1 | select Name | ConvertTo-Json",
             force_run=True,
         )
 
         if not switch_json:
-            return None
+            raise LisaException(f"Could not find default switch of type {switch_type}")
 
         return VMSwitch.from_json(switch_json)  # type: ignore
 
@@ -228,13 +242,13 @@ class HyperV(Tool):
                 force_run=True,
             )
 
-    def create_switch(self, name: str) -> None:
+    def create_switch(self, name: str, switch_type: str = "Internal") -> None:
         # remove switch if it exists
         self.delete_switch(name)
 
         # create a new switch
         self.node.tools[PowerShell].run_cmdlet(
-            f"New-VMSwitch -Name {name} -SwitchType Internal",
+            f"New-VMSwitch -Name {name} -SwitchType {switch_type}",
             force_run=True,
         )
 
@@ -253,16 +267,17 @@ class HyperV(Tool):
         )
         return bool(output.strip() != "")
 
-    def delete_nat(self, name: str) -> None:
+    def delete_nat(self, name: str, fail_on_error: bool = True) -> None:
         if self.exists_nat(name):
             self.node.tools[PowerShell].run_cmdlet(
                 f"Remove-NetNat -Name {name} -Confirm:$false",
                 force_run=True,
+                fail_on_error=fail_on_error,
             )
 
     def create_nat(self, name: str, ip_range: str) -> None:
         # delete NAT if it exists
-        self.delete_nat(name)
+        self.delete_nat(name, fail_on_error=False)
 
         # create a new NAT
         self.node.tools[PowerShell].run_cmdlet(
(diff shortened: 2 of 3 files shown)


SRIKKANTH:
This PR has a dependency on https://github.com/microsoft/lisa/pull/3573

lisa/tools/hyperv.py
```diff
@@ -206,6 +208,20 @@ def get_default_external_switch(self) -> Optional[VMSwitch]:
             'Get-VMSwitch | Where-Object {$_.SwitchType -eq "External"} '
             "| Select -First 1 | select Name | ConvertTo-Json",
             force_run=True,
+            fail_on_error=False,
+        )
+
+        if not switch_json:
```
squirrelsc: Above code may fail on other reasons. Catch the exception, and check the error output. Only return None, if it doesn't exist.

lisa/tools/hyperv.py
```diff
@@ -206,6 +208,20 @@ def get_default_external_switch(self) -> Optional[VMSwitch]:
...
+            return None
+
+        return VMSwitch.from_json(switch_json)  # type: ignore
+
+    def get_default_internal_switch(self) -> Optional[VMSwitch]:
```
squirrelsc: It doesn't need to create another method with similar functionality. Add an argument to set returned types.

lisa/tools/hyperv.py
```diff
@@ -221,20 +237,21 @@ def exists_switch(self, name: str) -> bool:
...
                 f"Remove-VMSwitch -Name {name} -Force",
                 force_run=True,
+                fail_on_error=fail_on_error,
             )
 
-    def create_switch(self, name: str) -> None:
+    def create_switch(self, name: str, switch_type: str = "Internal") -> None:
         # remove switch if it exists
-        self.delete_switch(name)
+        self.delete_switch(name, fail_on_error=False)
```
squirrelsc: What's the error found when deleting a switch? It looks the delete_switch checked if it exists, it shouldn't have other errors, which can be ignored.

lisa/tools/hyperv.py
```diff
@@ -221,20 +237,21 @@ def exists_switch(self, name: str) -> bool:
...
+    def delete_switch(self, name: str, fail_on_error: bool = True) -> None:
         if self.exists_switch(name):
             self.node.tools[PowerShell].run_cmdlet(
                 f"Remove-VMSwitch -Name {name} -Force",
                 force_run=True,
+                fail_on_error=fail_on_error,
             )
 
-    def create_switch(self, name: str) -> None:
+    def create_switch(self, name: str, switch_type: str = "Internal") -> None:
```
squirrelsc, line 245: Use an enum for Internal/External, I saw it's used multiple places.

lisa/tools/hyperv.py
```diff
@@ -381,30 +399,59 @@ def create_virtual_disk(self, name: str, pool_name: str, columns: int = 2) -> No
             force_run=True,
         )
 
-    def _check_exists(self) -> bool:
```
squirrelsc, line 384: Where is this method?

lisa/tools/hyperv.py
```diff
@@ -381,30 +399,59 @@ def create_virtual_disk(self, name: str, pool_name: str, columns: int = 2) -> No
...
+            force_run=True,
+        )
+
+        # Restart the DHCP server to apply the changes
+        service.restart_service("dhcpserver")
 
     def _install(self) -> bool:
         assert isinstance(self.node.os, Windows)
 
+        service: Service = self.node.tools[Service]
```
squirrelsc: Move this line to when it's used. So when `_check_exists` returns, it doesn't need to initialize the Service tool.

lisa/tools/hyperv.py
```diff
@@ -381,30 +399,59 @@ def create_virtual_disk(self, name: str, pool_name: str, columns: int = 2) -> No
...
-            return True
-        except LisaException as e:
-            self._log.debug(f"Hyper-V not installed: {e}")
-            return False
+    def configure_dhcp(self, dhcp_scope_name: str = "DHCPInternalNAT") -> None:
+        powershell = self.node.tools[PowerShell]
+        service: Service = self.node.tools[Service]
+
+        # Install DHCP server
+        self.node.tools[WindowsFeatureManagement].install_feature("DHCP")
```
squirrelsc, line 410: When is it needed? Hyper-V can assign IP addresses to VMs, it doens't need the DHCP server. When using the DHCP server, it's need to be careful, because it may conflict with lab DHCP.

SRIKKANTH: In Lab, nested VMs will get IPs from LAN DHCP using external Hyper-V Switch.
In Azure Windows Hyper-V hosts, We need to setup our own DHCP server with Internal Switch.

This will be used only in "HyperVPreparationTransformer" not in common Hyper-V environment setup.

SRIKKANTH: In Lab, nested VMs will get IPs from LAN DHCP using external Hyper-V Switch.
In Azure Windows Hyper-V hosts, We need to setup our own DHCP server with Internal Switch.

This will be used only in "HyperVPreparationTransformer" not in common Hyper-V environment setup.

lisa/tools/hyperv.py

lisa/tools/hyperv.py
SRIKKANTH: Moved to the bottom

lisa/tools/hyperv.py
SRIKKANTH: Fixed

lisa/tools/hyperv.py

lisa/tools/hyperv.py

lisa/tools/hyperv.py
SRIKKANTH: "Get-VMSwitch | Where-Object..." command never throws error only returns nothing if it didn't find any matching switch.

Changed the logic to raise exception if it didn't find any switch

lisa/tools/hyperv.py
```diff
@@ -221,20 +235,21 @@ def exists_switch(self, name: str) -> bool:
         )
         return bool(output.strip() != "")
 
-    def delete_switch(self, name: str) -> None:
+    def delete_switch(self, name: str, fail_on_error: bool = True) -> None:
```
squirrelsc: remove the unused `fail_on_error` flag.

lisa/tools/hyperv.py
squirrelsc: It's not fixed, it should be moved above `service.wait_for_service_start("vmms")`

lisa/tools/hyperv.py
squirrelsc: > In Azure Windows Hyper-V hosts, We need to setup our own DHCP server with Internal Switch.

Can you include a reference of official public document on how it works in the code comment? I need to take a look, and the code needs to be documented by comments also. My understanding is that the *internal* network is managed by HyperV, including IP address assignment. And the *external* switch doesn't work, if no extra IP address is assigned in the subnet on Azure.

lisa/tools/hyperv.py
SRIKKANTH: Its already moved to after "self.node.reboot()" not sure why its still showing up for you like this. please refresh once and check it again.

lisa/tools/hyperv.py
SRIKKANTH: I added the reference doc to the PR description. Repasting it here as well.

https://techcommunity.microsoft.com/blog/itopstalkblog/how-to-setup-nested-virtualization-for-azure-vmvhd/1115338

" My understanding is that the internal network is managed by HyperV, including IP address assignment." - This may be true on Windows desktop versions but I found DHCP config needed on Win server 2025.

lisa/tools/hyperv.py

lisa/tools/hyperv.py
```diff
@@ -381,30 +396,59 @@ def create_virtual_disk(self, name: str, pool_name: str, columns: int = 2) -> No
...
-            force_run=True,
-        )
+        self.node.tools[WindowsFeatureManagement].install_feature("Hyper-V")
 
         # reboot node
         self.node.reboot()
 
+        service: Service = self.node.tools[Service]
+        # wait for Hyper-V services to start
+        self.node.tools[Service].wait_for_service_start("vmms")
```
squirrelsc:
```suggestion
        service.wait_for_service_start("vmms")
```

lisa/tools/hyperv.py
squirrelsc: > " My understanding is that the internal network is managed by HyperV, including IP address assignment." - This may be true on Windows desktop versions but I found DHCP config needed on Win server 2025.

We have Windows Servers in lab, although they are not 2025, but I don't think it would be a regression to require DHCP. Did you try on Azure without DHCP server installed? The only benefit of the DHCP server is to specify the IP range, but it's not in our current test scope.


Add the link in the code comments instead of PR description. Please read the LISA documents about where to write documents. PR description is used to describe the change history, and the code comment is used to explain the current behavior.

SRIKKANTH: "Did you try on Azure without DHCP server installed?" I tried. L2 VMs are not getting IPs until I enable DHCP as mentioned the doc.

I added details about where this method should be used and shouldn't with a link to doc.

lisa/tools/hyperv.py
```diff
@@ -381,30 +396,66 @@ def create_virtual_disk(self, name: str, pool_name: str, columns: int = 2) -> No
...
-            self._log.debug("Hyper-V is installed")
-            return True
-        except LisaException as e:
-            self._log.debug(f"Hyper-V not installed: {e}")
-            return False
+    # This method is specifically for Azure Windows Server edition VMs
+    # with internal NAT switch.
+    # Do not use this method in Hyper-V hosts on LAN or other networks where DHCP
+    # server is already available.
+    # This method will configure the DHCP server on the Hyper-V host to provide IP
```
squirrelsc: Remove this line, it's confusing.

lisa/tools/hyperv.py
```diff
@@ -381,30 +396,66 @@ def create_virtual_disk(self, name: str, pool_name: str, columns: int = 2) -> No
...
+    # addresses to VMs.
+    # Reference doc:
+    # https://techcommunity.microsoft.com/blog/itopstalkblog/how-to-setup-nested-virtualization-for-azure-vmvhd/1115338  # noqa
+    def configure_dhcp(self, dhcp_scope_name: str = "DHCPInternalNAT") -> None:
```
squirrelsc: I don't see this method is called in this PR, how will it be called?

lisa/tools/hyperv.py
SRIKKANTH: Removed these lines:
# This method will configure the DHCP server on the Hyper-V host to provide IP
# addresses to VMs."

lisa/tools/hyperv.py
SRIKKANTH: It will be used in "HyperVPreparationTransformer" transformer. I will raise the PR after this one.

lisa/tools/hyperv.py
```diff
@@ -381,30 +396,64 @@ def create_virtual_disk(self, name: str, pool_name: str, columns: int = 2) -> No
...
-        except LisaException as e:
-            self._log.debug(f"Hyper-V not installed: {e}")
-            return False
+    # This method is specifically for Azure Windows Server edition VMs
+    # with internal NAT switch.
+    # Do not use this method in Hyper-V hosts on LAN or other networks where DHCP
+    # server is already available.
+    # Reference doc:
+    # https://techcommunity.microsoft.com/blog/itopstalkblog/how-to-setup-nested-virtualization-for-azure-vmvhd/1115338  # noqa
+    def configure_dhcp(self, dhcp_scope_name: str = "DHCPInternalNAT") -> None:
```
squirrelsc: `enable_internal_dhcp`

lisa/tools/hyperv.py
squirrelsc: Why is it reverted?

lisa/base_tools/service.py
```diff
@@ -67,6 +67,9 @@ def is_service_inactive(self, name: str) -> bool:
     def is_service_running(self, name: str) -> bool:
         return self._internal_tool._check_service_running(name)  # type: ignore
 
+    def wait_for_service_start(self, name: str) -> None:
```
squirrelsc, line 70: Why add it back? It should be called inside of the service, not a public method to control in test case level.

lisa/base_tools/service.py
```diff
@@ -67,6 +67,9 @@ def is_service_inactive(self, name: str) -> bool:
...
+        pass
```
squirrelsc: Raise Not implement error, so we know the code runs into an unknown path.

SRIKKANTH: Fixed

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3580, opened 2025-01-03
auth azure: support access token

LiliDeng:
managed identity doesn't work with windows container directly for this [reason](https://learn.microsoft.com/en-us/azure/container-instances/container-instances-managed-identity#managed-identity-on-windows-containers)

find a way to auth azure with access token

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+import base64
 import hashlib
 import json
 import os
@@ -26,6 +27,7 @@ from typing import (
 
 import requests
 from assertpy import assert_that
+from azure.core.credentials import AccessToken, TokenCredential
 from azure.core.exceptions import ResourceExistsError
 from azure.keyvault.certificates import (
     CertificateClient,
@@ -1634,12 +1636,14 @@ def generate_user_delegation_sas_token(
     connection_string: Optional[str] = None,
     writable: bool = False,
     expired_hours: int = 1,
+    platform: Optional["AzurePlatform"] = None,
 ) -> Any:
     blob_service_client = get_blob_service_client(
         cloud=cloud,
         credential=credential,
         account_name=account_name,
         connection_string=connection_string,
+        platform=platform,
     )
     start_time = datetime.now(timezone.utc)
     expiry_time = start_time + timedelta(hours=expired_hours)
@@ -1664,6 +1668,7 @@ def get_blob_service_client(
     credential: Optional[Any] = None,
     account_name: Optional[str] = None,
     connection_string: Optional[str] = None,
+    platform: Optional["AzurePlatform"] = None,
 ) -> BlobServiceClient:
     """
     Create a Azure Storage container if it does not exist.
@@ -1677,7 +1682,10 @@ def get_blob_service_client(
         assert (
             account_name
         ), "account_name is required, if connection_string is not set."
-
+        if platform and platform._azure_runbook.azure_storage_access_token:
+            credential = get_static_access_token(
+                platform._azure_runbook.azure_storage_access_token
+            )
         blob_service_client = BlobServiceClient(
             f"https://{account_name}.blob.{cloud.suffixes.storage_endpoint}",
             credential,
@@ -1692,15 +1700,21 @@ def get_or_create_storage_container(
     account_name: Optional[str] = None,
     connection_string: Optional[str] = None,
     allow_create: bool = True,
+    platform: Optional["AzurePlatform"] = None,
 ) -> ContainerClient:
     """
     Create a Azure Storage container if it does not exist.
     """
+    if platform and platform._azure_runbook.azure_storage_access_token:
+        credential = get_static_access_token(
+            platform._azure_runbook.azure_storage_access_token
+        )
     blob_service_client = get_blob_service_client(
         cloud=cloud,
         credential=credential,
         account_name=account_name,
         connection_string=connection_string,
+        platform=platform,
     )
     container_client = blob_service_client.get_container_client(container_name)
     if not container_client.exists():
@@ -1836,6 +1850,7 @@ def copy_vhd_to_storage(
         cloud=platform.cloud,
         account_name=storage_name,
         container_name=SAS_COPIED_CONTAINER_NAME,
+        platform=platform,
     )
     full_vhd_path = f"{container_client.url}/{dst_vhd_name}"
 
@@ -1882,6 +1897,7 @@ def copy_vhd_to_storage(
                     cloud=platform.cloud,
                     account_name=storage_name,
                     writable=True,
+                    platform=platform,
                 )
                 dst_vhd_sas_url = f"{full_vhd_path}?{sas_token}"
                 log.info(f"copying vhd by azcopy {dst_vhd_name}")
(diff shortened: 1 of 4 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -246,6 +248,8 @@ class AzurePlatformSchema:
         ),
     )
     service_principal_key: str = field(default="")
+    access_token: str = field(default="")
+    token_expiry_hours: int = 24
```
squirrelsc: `access_token_expiry_hours`

LiliDeng:
During testing, found this token is working for managing azure resources, I am consulting azure python sdk how to handle this.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -936,10 +966,18 @@ def _initialize_credential(self) -> None:
...
+            if azure_runbook.access_token:
+                os.environ["AZURE_ACCESS_TOKEN"] = azure_runbook.access_token
 
-            credential = DefaultAzureCredential(
-                authority=self.cloud.endpoints.active_directory,
-            )
+            if "AZURE_ACCESS_TOKEN" in os.environ:
+                credential = StaticAccessTokenCredential(
+                    os.environ["AZURE_ACCESS_TOKEN"],
+                    int(time.time()) + 3600 * azure_runbook.access_token_expiry_hours,
```
squirrelsc: Maybe try to parse the expiration time from the token itself. So, the parameter is not needed.

squirrelsc:
> During testing, found this token is working for managing azure resources, I am consulting azure python sdk how to handle this.

What does it mean? Do you mean it fails in some scenarios, so you need have more tests or confirmation before merging? If so, please set it to Draft.

LiliDeng:
> > During testing, found this token is working for managing azure resources, I am consulting azure python sdk how to handle this.
> 
> What does it mean? Do you mean it fails in some scenarios, so you need have more tests or confirmation before merging? If so, please set it to Draft.

Never mind, this one works.

lisa/sut_orchestrator/azure/platform_.py
LiliDeng: Fixed

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -400,21 +406,44 @@ def cloud(self, value: Optional[CloudSchema]) -> None:
             self.cloud_raw = value.to_dict()  # type: ignore
 
 
+class StaticAccessTokenCredential(TokenCredential):
+    def __init__(self, token: str, expires_on: int) -> None:
```
squirrelsc: The code can be simplified by chanage to `def __init__(self, token: str) -> None:`, and then parse the expiration time in the method. When it's used, parse expiration date in the token itself.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2696,6 +2702,41 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+        # The second part of the JWT is the payload
+        payload = token.split(".")[1]
+        # Add padding to ensure Base64 decoding works properly
+        padded_payload = payload + "=" * (4 - len(payload) % 4)
+        # Decode the Base64 URL-safe encoded payload
+        decoded_payload = base64.urlsafe_b64decode(padded_payload)
+        # Convert the payload into a dictionary
+        return json.loads(decoded_payload)
+
+    def __init__(self, token: str) -> None:
```
squirrelsc, line 2719: Usually, the `__init__` (construct method) is the first method of a class.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2696,6 +2702,41 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
             raise LisaException(f"Data disk type {disk_type} is unsupported.")
 
 
+class StaticAccessTokenCredential(TokenCredential):
+    def decode_jwt(self, token: str) -> Any:
```
squirrelsc: 1. it should be a private method, and move to the bottom of the class.
2. The all purpose of the logic is to extract `exp`, so rename it like `get_exp`, unless other fields needed.
3. The `token` was saved to the class, before call this method, so it doesn't need to pass in the token from the method call.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1696,6 +1698,10 @@ def get_or_create_storage_container(
     """
     Create a Azure Storage container if it does not exist.
     """
+    if "AZURE_STORAGE_ACCESS_TOKEN" in os.environ:
```
squirrelsc: minimized the access to the env variable. For here, the logic can be like below. the method is used to wrap the env check.
`credential = get_static_access_token() or credential`

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2696,6 +2702,41 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+
+    def __init__(self, token: str) -> None:
+        """
+        Initialize StaticAccessTokenCredential with the provided token.
+
+        :param token: The Azure access token as a string.
+        """
+        self._token = token
+        # Parse the token to get the expiration timestamp
+        decoded_token = self.decode_jwt(os.environ["AZURE_ACCESS_TOKEN"])
```
squirrelsc: The token was passed in, why access the env variable?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1696,6 +1698,7 @@ def get_or_create_storage_container(
     """
     Create a Azure Storage container if it does not exist.
     """
+    credential = get_static_access_token("AZURE_STORAGE_ACCESS_TOKEN") or credential
```
squirrelsc: Why the env name variable is different? Is it defined by the staroge class or us?

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2696,6 +2699,48 @@ def get_size(disk_type: schema.DiskType, data_disk_iops: int = 1) -> int:
...
+
+    def _get_exp(self) -> Any:
+        # The second part of the JWT is the payload
+        payload = self._token.split(".")[1]
+        # Add padding to ensure Base64 decoding works properly
+        padded_payload = payload + "=" * (4 - len(payload) % 4)
+        # Decode the Base64 URL-safe encoded payload
+        decoded_payload = base64.urlsafe_b64decode(padded_payload)
+        # Convert the payload into a dictionary
+        return json.loads(decoded_payload)
```
squirrelsc: `return json.loads(decoded_payload).get("exp")`
It's not only to change the method name, but also to encapsulate the logic.

lisa/sut_orchestrator/azure/common.py
LiliDeng: it is defined by us, the tokens need to be generated based on different scopes.

squirrelsc: 1. Is it possible to have one token with all scopes like *?
2. Define schema to accept tokens, instead of env vars. All LISA vars should be defined in runbook.

Let me know, if it's hard to fulfill above.

LiliDeng: 1. I have confirmed with the maintainer of python sdk, he said we have to use different tokens for different scopes.
2. Use different variables to accept the tokens?

squirrelsc: > 2. Use different variables to accept the tokens?

As I said, define in schema for them, not directly use env vars. After you defined schema, you can set values by LISA vars, and then assign LISA vars by env vars.

Like you defined `access_token` in PlatformSchema. Please define similar tokens for other Azure scopes.

squirrelsc (dismissed):

squirrelsc (approved):

PR #3581, opened 2025-01-03
DPDK: use multiple mempools, assert hugepage count is > 0

mcgov:
Fufilling a request to excersize code where DPDK uses more than the default amount of memory pools. Adds an option to testpmd to allocate multiple sizes. Includes a commit for the hugepages tool to assert the requested count is > 0.


mcgov:
Reworking some of this, will create a new PR after changing some things and testing.

mcgov:
That was fast! Ready for eyes

PR #3583, opened 2025-01-06
ch_tests: dynamically determine data disk path

pupacha:
The test VM undergoes a reboot during preparation for CH community tests, causing the device path for the data disk to potentially change. So, data disk device path is now determined dynamically just before launching the CH tests, eliminating the need for 'datadisk_name' test input variable.

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -22,11 +22,12 @@ from lisa.tools import (
     Echo,
     Git,
     Ls,
+    Lsblk,
     Mkdir,
     Modprobe,
     Whoami,
 )
-from lisa.util import find_groups_in_lines
+from lisa.util import LisaException, find_groups_in_lines
 
 
 @dataclass
@@ -62,7 +63,6 @@ class CloudHypervisorTests(Tool):
 
     # Block perf related env var
     use_datadisk = ""
-    datadisk_name = ""
     disable_datadisk_cache = ""
     block_size_kb = ""
 
@@ -164,6 +164,9 @@ class CloudHypervisorTests(Tool):
         skip: Optional[List[str]] = None,
         subtest_timeout: Optional[int] = None,
     ) -> None:
+        if self.use_datadisk:
+            self._set_data_disk()
+
         if ref:
             self.node.tools[Git].checkout(ref, self.repo_root)
 
@@ -264,8 +267,6 @@ class CloudHypervisorTests(Tool):
 
             if self.use_datadisk:
                 self.env_vars["USE_DATADISK"] = self.use_datadisk
-            if self.datadisk_name:
-                self.env_vars["DATADISK_NAME"] = self.datadisk_name
             if self.disable_datadisk_cache:
                 self.env_vars["DISABLE_DATADISK_CACHING"] = self.disable_datadisk_cache
             if self.block_size_kb:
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -496,6 +497,23 @@ def _configure_vdpa_devices(self, node: Node) -> None:
...
+        disks = self.node.tools[Lsblk].get_disks()
+        # get the first unmounted disk (data disk)
+        for disk in disks:
+            if disk.is_mounted:
+                continue
+            if disk.name.startswith("sd"):
+                datadisk_name = disk.device_name
+                break
+        # running lsblk once again, just for human readable logs
+        self.node.execute("lsblk", shell=True)
```
LiliDeng: can lsblk tool be used here?

pupacha: `lsblk` is already used to get the devices list.
I'm doing `node.execute("lsblk")` additionally for debugging purposes. This will print human readable table of the devices and its information.

squirrelsc: What the output of `self.node.tools[Lsblk].get_disks()` looks like? If it's not readable, you can call like lsblk.run() without parameters.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -496,6 +497,23 @@ def _configure_vdpa_devices(self, node: Node) -> None:
...
+        if not datadisk_name:
+            raise LisaException("No data disk found")
```
squirrelsc: Specify which name is looking from lsblk, like no disk name starts with `sd`, and give the returned names from lsblk.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -496,6 +497,23 @@ def _configure_vdpa_devices(self, node: Node) -> None:
...
+        self._log.debug(f"Using data disk: {datadisk_name}")
+        self.env_vars["DATADISK_NAME"] = datadisk_name
```
squirrelsc, line 516: Why set it to env variables? If it's used in the class only, return the name. Not use env or class level variables.

pupacha: This env variable is need by the ch community test scripts that are run in the CH LISA tests.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
pupacha: We are not looking for specially device just starting with `sd`. Instead, we are looking for a data disk which is unmounted.
I feel this exception message is good enough for debugging the failures.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
pupacha: `self.node.tools[Lsblk].get_disks()` gives a json output. While simple `lsblk` command output is more readable. I changed it to `lsblk.run()`

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
squirrelsc: The error message is for people, who run tests but never read the code. So, it needs to include as much information as possible to reduce their troubleshooting cost.

pupacha: Updated the message to `No unmounted data disk (/dev/sdX) found`

smit-gardhariya (approved):

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3587, opened 2025-01-07
Add TestCategory and MachineName to JUnitReport

snipex111:


snipex111:
Hi @LiliDeng, this branch is used for local testing, and that master-se-1 was for pipelines. so please approve this also. thanks.

LiliDeng (approved):

PR #3590, opened 2025-01-08
Collect namespace ids for NVMe devices

SRIKKANTH:
NVME namespace ids are unique for each disk under any NVME controller.
These are useful in detecting the LUN IDs of the remote azure disk disks.

### DIFF
--- a/lisa/tools/nvmecli.py
+++ b/lisa/tools/nvmecli.py
@@ -179,13 +179,14 @@ class BSDNvmecli(Nvmecli):
     def can_install(self) -> bool:
         return False
 
-    def get_namespaces(self) -> List[str]:
+    def get_namespaces(self, force_run: bool = False) -> List[str]:
         output = self.run(
             "devlist",
             shell=True,
             sudo=True,
             expected_exit_code=0,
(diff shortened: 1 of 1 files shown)


lisa/tools/nvmecli.py
```diff
@@ -96,15 +97,74 @@ def support_device_format(self, device_name: str) -> bool:
         cmd_result.assert_exit_code()
         return self.__format_device_support in cmd_result.stdout
 
-    def get_namespaces(self) -> List[str]:
+    def get_namespaces(self, force_run: bool = False) -> List[str]:
```
kanchansenlaskar, line 100: Where are we calling this with force_run value as True, given that "run" function also has force_run defaulted to False only? Do you intend to call it as part of a separate PR?

SRIKKANTH: Yes. This is for serial disk testcases which are failing on dct =nvme with index out of error.

That pr will use these methods.

kanchansenlaskar: can you share the link please?

kanchansenlaskar (approved):

LiliDeng (approved):

PR #3591, opened 2025-01-08
Johnf/add supported images list

feng-j678:
[x] add lpe supported images list according to public and internal doc https://learn.microsoft.com/en-us/azure/virtual-machines/automatic-vm-guest-patching


microsoft/testsuites/vm_extensions/linux_patch_extension.py
```diff
@@ -48,28 +48,108 @@ def _verify_supported_arm64_images(node: Node, log: Logger, full_version: Any) -
...
+            "8.10.2",
+            "9.0.1",
+            "9.0.2",
+            "9.1.1",
+            "9.1.2",
+            "9.4.1",
+            "9.4.2",
+        ],
+        Redhat: [
+            "7.2.1",
```
squirrelsc: You can use a range and operator, instead of list every version.

feng-j678: i use the range() function to shorten, is that what you meant by it?

squirrelsc: Partially. LISA version comparison supports `>` and `<` too, so you can use like `versionA >= "8.0.0" and versionA < "10.0.0"` for all 8.x.x and 9.x.x.

Use `parse_version` in LISA utils like some other cases, then `versionA = parse_version("x.x.x")`.

feng-j678: i see that parse_version in lisa, but i prefer the current implementation because its easier to follow the logic

feng-j678: done

squirrelsc:
@LiliDeng LGTM

microsoft/testsuites/vm_extensions/linux_patch_extension.py
```diff
@@ -48,29 +48,63 @@ def _verify_supported_arm64_images(node: Node, log: Logger, full_version: Any) -
...
+        or (
+            full_version >= parse_version("9.4.1")
+            and full_version <= parse_version("9.4.2")
+        )
+    ):
+        log.debug(f"This is a supported image: {full_version}")
+        return
+
+    # check for other supported image versions
+    for distro in lpe_supported_images_versions:
```
LiliDeng: for distro, version_list in lpe_supported_images_versions.items()

LiliDeng (approved):

PR #3593, opened 2025-01-08
Added logging the input content for process by default

paulli2017:
Added logging the input content for process by default, which will help analyze the log.

### DIFF
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -344,13 +344,16 @@ class Process:
                             "Running commands with sudo requires user's password,"
                             " but no password is provided."
                         )
-                    self.input(f"{self._shell.connection_info.password}\n")
+                    self.input(f"{self._shell.connection_info.password}\n", False)
                     self._log.debug("The user's password is input")
                     break
 
-    def input(self, content: str) -> None:
+    def input(self, content: str, is_log_input: bool = True) -> None:
         assert self._process, "The process object is None, the process may end."
-        self._log.debug(f"Inputting {len(content)} chars to process.")
+        if is_log_input:
+            self._log.debug(f"input content: {content}")
+        else:
+            self._log.debug(f"Inputting {len(content)} chars to process.")
         self._process.stdin_write(content)
 
     def wait_result(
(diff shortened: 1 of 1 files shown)


lisa/util/process.py
```diff
@@ -344,13 +344,15 @@ def check_and_input_password(self) -> None:
...
                         )
-                    self.input(f"{self._shell.connection_info.password}\n")
+                    self.input(f"{self._shell.connection_info.password}\n", False)
                     self._log.debug("The user's password is input")
                     break
 
-    def input(self, content: str) -> None:
+    def input(self, content: str, is_log_input: bool = True) -> None:
         assert self._process, "The process object is None, the process may end."
         self._log.debug(f"Inputting {len(content)} chars to process.")
```
squirrelsc: This line should output only no raw input.

paulli2017: The change is updated, please check it again, thanks

squirrelsc:
@LiliDeng LGTM

lisa/util/process.py
```diff
@@ -344,13 +344,16 @@ def check_and_input_password(self) -> None:
...
                     break
 
-    def input(self, content: str) -> None:
+    def input(self, content: str, is_log_input: bool = True) -> None:
         assert self._process, "The process object is None, the process may end."
-        self._log.debug(f"Inputting {len(content)} chars to process.")
+        if is_log_input:
+            self._log.debug(f"input content: {content}")
+        else:
+            self._log.debug(f"Inputting {len(content)} chars to process.")
```
LiliDeng, line 356: Is the content of this if and else statement reversed?

squirrelsc: It's not, but please advise how to make the name clearer. The name means "is log input?" if it's true, log the input, if not just log the length of input.

paulli2017: the changed behavior is to log the input content by default and otherwise the input content length. The only exception for now is password (the change for the call of self.input() in check_and_input_password() function)

LiliDeng (approved):

PR #3596, opened 2025-01-09
Write test log for deployment failures

adityagesh:
At present, test logs are not generated for deployment failures. This change creates case log for deployment failure with the failure reason.

### DIFF
--- a/lisa/platform_.py
+++ b/lisa/platform_.py
@@ -177,7 +177,7 @@ class Platform(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
     def deploy_environment(self, environment: Environment) -> None:
         platform_runbook = cast(schema.Platform, self.runbook)
 
-        log = get_logger(f"deploy[{environment.name}]", parent=self._log)
+        log = get_logger("deploy", parent=environment.log)
         log.info(f"deploying environment: {environment.name}")
         timer = create_timer()
         environment.platform = self
--- a/lisa/runners/lisa_runner.py
+++ b/lisa/runners/lisa_runner.py
@@ -291,6 +291,8 @@ class LisaRunner(BaseRunner):
         self, environment: Environment, test_results: List[TestResult]
     ) -> None:
         try:
+            test_result = test_results[0]
+            test_result.subscribe_log(environment.log)
             try:
                 # Attempt to deploy the environment
                 self.platform.deploy_environment(environment)
@@ -315,10 +317,12 @@ class LisaRunner(BaseRunner):
                 # Final attempt failed; handle the failure
                 self._attach_failed_environment_to_result(
                     environment=environment,
-                    result=test_results[0],
+                    result=test_result,
                     exception=identifier,
                 )
                 self._delete_environment_task(environment=environment, test_results=[])
+        finally:
+            test_result.unsubscribe_log(environment.log)
 
     def _initialize_environment_task(
         self, environment: Environment, test_results: List[TestResult]
@@ -616,7 +620,10 @@ class LisaRunner(BaseRunner):
         # so deployment failure can be tracked.
         environment.platform = self.platform
         result.environment = environment
-        result.handle_exception(exception=exception, log=self._log, phase="deployment")
+
+        result.handle_exception(
+            exception=exception, log=environment.log, phase="deployment"
+        )
         self._log.info(
             f"'{environment.name}' attached to test case "
             f"'{result.runtime_data.metadata.full_name}({result.id_})': "
--- a/lisa/testsuite.py
+++ b/lisa/testsuite.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import copy
+import logging
 import traceback
 from dataclasses import dataclass, field
 from functools import wraps
@@ -85,12 +86,16 @@ class TestResult:
     log_file: str = ""
     stacktrace: Optional[str] = None
     retried_times: int = 0
+    _log_file_handler: Optional[logging.FileHandler] = None
+    _case_log_path: Optional[Path] = None
 
     def __post_init__(self, *args: Any, **kwargs: Any) -> None:
         self._send_result_message()
         self._timer: Timer
 
         self._environment_information: Dict[str, Any] = {}
+        # parent_log = get_logger("suite", self.runtime_data.metadata.suite.name)
+        self.log = get_logger("case", self.name)
 
     @property
     def is_queued(self) -> bool:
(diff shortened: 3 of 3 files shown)


lisa/runners/lisa_runner.py
```diff
@@ -616,12 +617,15 @@ def _attach_failed_environment_to_result(
...
         result.environment = environment
-        result.handle_exception(exception=exception, log=self._log, phase="deployment")
+
+        result.handle_exception(exception=exception, log=result.log, phase="deployment")
         self._log.info(
             f"'{environment.name}' attached to test case "
             f"'{result.runtime_data.metadata.full_name}({result.id_})': "
             f"{exception}"
         )
+        result.save_nodes_serial_console_logs()
```
adityagesh: @squirrelsc  , this doesn't work as environment.nodes is empty list

Shall I make _save_console_log_and_check_panic as a public method of Platform and then invoke platform.save_console_log(path=result.log_path)

https://github.com/microsoft/lisa/blob/4b8bca855ea5463585e7e08a73bed1bee753c238/lisa/sut_orchestrator/azure/platform_.py#L654

squirrelsc: `_attach_failed_environment_to_result` is too late to attach log. What I mean is to attach log when this method is called in deploy/prepare methods. It can attach to the environment log, not the node log. env should be created already, and it includes all node logs.

lisa/platform_.py
```diff
@@ -177,7 +177,7 @@ def prepare_environment(self, environment: Environment) -> Environment:
     def deploy_environment(self, environment: Environment) -> None:
         platform_runbook = cast(schema.Platform, self.runbook)
 
-        log = get_logger(f"deploy[{environment.name}]", parent=self._log)
+        log = get_logger("", parent=environment.log)
```
adityagesh: @squirrelsc is this fine?

squirrelsc: It doesn't need to change the logger prefix. The original prefix can help find log from different sinks.

lisa/runners/lisa_runner.py
```diff
@@ -318,6 +321,7 @@ def _deploy_environment_task(
                     result=test_results[0],
                     exception=identifier,
                 )
+                remove_handler(log_file_handler, environment.log)
```
squirrelsc: It should be in final, not except.

lisa/runners/lisa_runner.py
```diff
@@ -616,12 +620,14 @@ def _attach_failed_environment_to_result(
...
         result.environment = environment
-        result.handle_exception(exception=exception, log=self._log, phase="deployment")
+
+        result.handle_exception(exception=exception, log=result.log, phase="deployment")
         self._log.info(
             f"'{environment.name}' attached to test case "
             f"'{result.runtime_data.metadata.full_name}({result.id_})': "
             f"{exception}"
         )
+        result.close_log_file_handler()
```
squirrelsc: Don't remove handler here, it should be paired with add.

lisa/testsuite.py
```diff
@@ -91,6 +93,10 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
         self._timer: Timer
 
         self._environment_information: Dict[str, Any] = {}
+        parent_log = get_logger("suite", self.runtime_data.metadata.suite.name)
+        self.log = get_logger("case", self.name, parent=parent_log)
+        self.case_log_path = self.__create_case_log_path()
```
squirrelsc: The log path shouldn't be created at this memont, because the timestamp of the path indicates when the case started.

lisa/testsuite.py
```diff
@@ -265,6 +271,32 @@ def check_platform(
 
         return result
 
+    def get_log_file_handler(self) -> logging.FileHandler:
```
squirrelsc: Have a paired methods called, `subscribe_log` and `unsubscribe_log`, and pass in the logger to be subscribed.

lisa/testsuite.py
```diff
@@ -91,6 +93,10 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
         self._timer: Timer
 
         self._environment_information: Dict[str, Any] = {}
+        parent_log = get_logger("suite", self.runtime_data.metadata.suite.name)
```
squirrelsc: The parent_log is not useful, you can just use the "case" prefix.

lisa/platform_.py
adityagesh: With my change, the log is 
`lisa.env[generated_0]. creating deployment`

Without removing "deploy..."
`lisa.env[generated_0].deploy[generated_0] creating deployment`

With the original platform logger (self._log)
`lisa.[azure].deploy[generated_0] creating deployment`

squirrelsc: `lisa.env[generated_0].deploy creating deployment` is better.

lisa/runners/lisa_runner.py
```diff
@@ -616,12 +620,15 @@ def _attach_failed_environment_to_result(
         # so deployment failure can be tracked.
         environment.platform = self.platform
         result.environment = environment
+
+        result.subscribe_log(self._log)
```
squirrelsc: Does it need to subscribe runner here? The below line `result.handle_exception` should have enough log. The runner is called in multi threads, and may bring logs from other cases.

lisa/platform_.py
squirrelsc: `lisa.env[generated_0].deploy` is better.

lisa/testsuite.py
```diff
@@ -265,6 +270,44 @@ def check_platform(
...
+            ).as_posix()
+        return self._log_file_handler
+
+    def close_log_file_handler(self) -> None:
+        if self._log_file_handler:
+            remove_handler(self._log_file_handler, self.log)
+            self._log_file_handler.close()
+            self._log_file_handler = None
+
+    def save_nodes_serial_console_logs(self) -> None:
```
squirrelsc: `capture_serial_console_log`

lisa/runners/lisa_runner.py
adityagesh: That seems to be sufficient for failed deployment.
In case of successful deployment, the deployment logs wouldn't show up in test log. That should be fine, right?

lisa/testsuite.py
adityagesh: fixed

lisa/testsuite.py
adityagesh: Changed, now it is created when path is read for the first time. Is that fine?

lisa/runners/lisa_runner.py
squirrelsc: subscribe here is too late, it won't capture the log generated when deploying. In another word, it includes nothing except the exception is logged out again.

adityagesh: Are you recommending to subscribe and unsubscribe in _deploy_environment_task itself rather than _attach_failed_environment_to_result ?

squirrelsc: Yes, subscribe in deploy and prepare, not in the attach failed.

adityagesh: I am trying to find which test result to use in _prepare_environment

squirrelsc: You can manually create an exception in prepare, so it can be tested easier.

adityagesh: As discussed offline, prepare has a different way of selecting test result. In this PR we focus on improving Deployment logging alone

adityagesh:
@squirrelsc Can we merge this? or do you have more suggestions?

lisa/runners/lisa_runner.py
squirrelsc: The subscribe should include deploy step as well. If only subscribe here, it looks only the exception is included, no other logs during deploying.

adityagesh: Sorry, I didn't realize that I did not push this change. 
Pushed now

squirrelsc:
@LiliDeng LGTM

LiliDeng:
I found when run against this branch, it costs 30 mins - 60 mins longer than using main branch. @adityagesh could you please check this?

LiliDeng (approved):

PR #3597, opened 2025-01-09
Enabled wait_output() to search from latest log buffer

paulli2017:
Enabled wait_output() to search from latest log buffer, which will help understand more of running behavior from buffer content.

### DIFF
--- a/lisa/util/process.py
+++ b/lisa/util/process.py
@@ -214,6 +214,7 @@ class Process:
 
         # add a string stream handler to the logger
         self.log_buffer = io.StringIO()
+        self.log_buffer_offset = 0
 
     @_retry_spawn
     def start(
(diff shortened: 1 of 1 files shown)


lisa/util/process.py
```diff
@@ -499,7 +501,10 @@ def wait_output(
             self._stderr_writer.flush()
 
             # check if buffer contains the keyword
-            if keyword in self.log_buffer.getvalue():
+            find_pos = self.log_buffer_offset if delta_only else 0
+            index = self.log_buffer.getvalue().find(keyword, find_pos)
+            self.log_buffer_offset = len(self.log_buffer.getvalue())
```
squirrelsc: Not update offset here, it may break keyword to two parts, and missed finding. Only update the offset when exiting the method.

paulli2017: The change is done, thanks for pointing out this issue.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3598, opened 2025-01-13
DPDK: install path fixes for meson and Ubuntu 24.04

mcgov:
Fixing some bugs which were noted after testing on Ubuntu 24.04:

- 'meson' is occasionally packaged as 'meson' but it is being standardized as 'python3-meson' in future releases. This PR fixes the installation path for Meson to account for an prefer the packaged version if it's available and the correct version.
- linux-modules-extra-azure requires a restart, dependencies occasionally do. So add a 'needs_restart' parameter to the dependency installer to allow rebooting before continuing to install dependencies.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -2315,6 +2315,35 @@ class Suse(Linux):
                 "There are no enabled repositories defined in this image.",
             )
 
+    def _uninstall_packages(
+        self,
+        packages: List[str],
+        signed: bool = True,
+        timeout: int = 600,
+        extra_args: Optional[List[str]] = None,
+    ) -> None:
+        add_args = self._process_extra_package_args(extra_args)
+        command = f"zypper --non-interactive {add_args}"
+        if not signed:
+            command += " --no-gpg-checks "
+        command += f" rm {' '.join(packages)}"
+        self.wait_running_process("zypper")
+        install_result = self._node.execute(
+            command, shell=True, sudo=True, timeout=timeout
+        )
+        if install_result.exit_code in (1, 100):
+            raise LisaException(
+                f"Failed to install {packages}. exit_code: {install_result.exit_code}, "
+                f"stderr: {install_result.stderr}"
+            )
+        elif install_result.exit_code == 0:
+            self._log.debug(f"{packages} is/are installed successfully.")
+        else:
+            self._log.debug(
+                f"{packages} is/are installed."
+                " A system reboot or package manager restart might be required."
+            )
+
     def _install_packages(
         self,
         packages: List[str],
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -2331,17 +2331,13 @@ class Suse(Linux):
         install_result = self._node.execute(
             command, shell=True, sudo=True, timeout=timeout
         )
-        if install_result.exit_code in (1, 100):
-            raise LisaException(
-                f"Failed to install {packages}. exit_code: {install_result.exit_code}, "
-                f"stderr: {install_result.stderr}"
-            )
-        elif install_result.exit_code == 0:
-            self._log.debug(f"{packages} is/are installed successfully.")
+
+        if install_result.exit_code == 0:
+            self._log.debug(f"{packages} is/are removed successfully.")
         else:
-            self._log.debug(
-                f"{packages} is/are installed."
-                " A system reboot or package manager restart might be required."
+            raise LisaException(
+                f"Failed to remove {packages}. exit_code: {install_result.exit_code}, "
+                f"stderr: {install_result.stderr}"
             )
 
     def _install_packages(
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -2341,13 +2341,14 @@ class Suse(Linux):
         command = f"zypper --non-interactive {add_args}"
         if not signed:
             command += " --no-gpg-checks "
-        command += f" rm {' '.join(packages)}"
+        remove_packages = " ".join(packages)
+        command += f" rm {remove_packages}"
         self.wait_running_process("zypper")
         install_result = self._node.execute(
             command, shell=True, sudo=True, timeout=timeout
         )
         assert_that(install_result.exit_code).described_as(
-            f"Failed to remove {packages}. "
+            f"Failed to remove {remove_packages}. "
             f"exit_code: {install_result.exit_code}, "
             f"stderr: {install_result.stderr}"
         ).is_equal_to(0)
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -418,6 +418,7 @@ class Posix(OperatingSystem, BaseClassMixin):
         self,
         package: Union[str, Tool, Type[Tool]],
         assert_existance: Union[bool, None] = None,
+        minimum_version: Optional[VersionInfo] = None,
     ) -> bool:
         """
         Query if a package/tool is installed on the node.
@@ -433,7 +434,13 @@ class Posix(OperatingSystem, BaseClassMixin):
                 f"Package {package} installation status is unexpected."
             ).is_equal_to(assert_existance)
 
-        return exists
+        if exists and minimum_version:
+            return (
+                self.get_package_information(package_name=package_name)
+                >= minimum_version
+            )
+        else:
+            return exists
 
     def is_package_in_repo(self, package: Union[str, Tool, Type[Tool]]) -> bool:
         """
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -2346,14 +2346,12 @@ class Suse(Linux):
         install_result = self._node.execute(
             command, shell=True, sudo=True, timeout=timeout
         )
-
-        if install_result.exit_code == 0:
-            self._log.debug(f"{packages} is/are removed successfully.")
-        else:
-            raise LisaException(
-                f"Failed to remove {packages}. exit_code: {install_result.exit_code}, "
-                f"stderr: {install_result.stderr}"
-            )
+        assert_that(install_result.exit_code).described_as(
+            f"Failed to remove {packages}. "
+            f"exit_code: {install_result.exit_code}, "
+            f"stderr: {install_result.stderr}"
+        ).is_equal_to(0)
+        self._log.debug(f"{packages} is/are removed successfully.")
 
     def _install_packages(
         self,
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -434,11 +434,19 @@ class Posix(OperatingSystem, BaseClassMixin):
                 f"Package {package} installation status is unexpected."
             ).is_equal_to(assert_existance)
 
-        if exists and minimum_version:
-            return (
+        if exists and minimum_version is not None:
+            acceptable = bool(
                 self.get_package_information(package_name=package_name)
                 >= minimum_version
             )
+            if not acceptable:
+                self._log.debug(
+                    (
+                        f"Minimum version of package {package_name} "
+                        f"did not meet minimum of {str(minimum_version)}."
+                    )
+                )
+            return acceptable
         else:
             return exists
 
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -434,21 +434,16 @@ class Posix(OperatingSystem, BaseClassMixin):
                 f"Package {package} installation status is unexpected."
             ).is_equal_to(assert_existance)
 
-        if exists and minimum_version is not None:
-            acceptable = bool(
-                self.get_package_information(package_name=package_name)
-                >= minimum_version
-            )
-            if not acceptable:
-                self._log.debug(
-                    (
-                        f"Minimum version of package {package_name} "
-                        f"did not meet minimum of {str(minimum_version)}."
-                    )
-                )
-            return acceptable
-        else:
+        self._log.debug(f"package '{package}' exists: {exists}")
+        if minimum_version is None or not exists:
             return exists
+
+        actual_version = self.get_package_information(package_name=package_name)
+        self._log.debug(
+            f"package '{package}' expected min version: "
+            f"{str(minimum_version)}, actual version: {str(actual_version)}"
+        )
+        return actual_version >= minimum_version
 
     def is_package_in_repo(self, package: Union[str, Tool, Type[Tool]]) -> bool:
         """
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -418,7 +418,6 @@ class Posix(OperatingSystem, BaseClassMixin):
         self,
         package: Union[str, Tool, Type[Tool]],
         assert_existance: Union[bool, None] = None,
-        minimum_version: Optional[VersionInfo] = None,
     ) -> bool:
         """
         Query if a package/tool is installed on the node.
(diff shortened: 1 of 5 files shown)


lisa/operating_system.py
```diff
@@ -2094,6 +2094,35 @@ def _initialize_package_installation(self) -> None:
...
+        )
+        if install_result.exit_code in (1, 100):
+            raise LisaException(
+                f"Failed to install {packages}. exit_code: {install_result.exit_code}, "
+                f"stderr: {install_result.stderr}"
+            )
+        elif install_result.exit_code == 0:
+            self._log.debug(f"{packages} is/are installed successfully.")
+        else:
+            self._log.debug(
```
squirrelsc: Should the reboot be a specified exit code? Why are only 1/100 failed codes?

mcgov: I am following the convention of the install_packages code for SUSE. I'm assuming it was based on https://manpages.opensuse.org/Tumbleweed/zypper/zypper.8.en.html

The others seem to be info error codes

squirrelsc: I read the doc, I feel only 0 means success. For example, 2 means command parameter errors, 3 means invalid args ...

mcgov: sure we don't need this. I was following the convention but let's just fail on not 0

microsoft/testsuites/dpdk/dpdktestpmd.py
```diff
@@ -226,7 +227,8 @@ def _setup_node(self) -> None:
         # which breaks when another tool checks for it's existence before building...
         # like cmake, meson, make, autoconf, etc.
         self._node.tools[Ninja].install()
-        self._node.tools[Pip].install_packages("pyelftools")
+        if not isinstance(self._os, Debian):
+            self._node.tools[Pip].install_packages("pyelftools")
```
squirrelsc, line 238: Will Debian pass without this package?

lisa/tools/meson.py
```diff
@@ -30,9 +30,25 @@ def can_install(self) -> bool:
...
         # use pip to make sure we install a recent version
-        if (not posix_os.package_exists("meson")) or posix_os.get_package_information(
-            "meson", use_cached=False
-        ) < "0.52.0":
+
+        package_installed = ""
+        package_available = ""
+        for pkg in ["meson", "python3-meson"]:
+            if (
+                posix_os.package_exists(pkg)
```
squirrelsc: Move the implementation into check_exists, so other place can reuse it.

lisa/tools/meson.py
```diff
@@ -30,9 +30,25 @@ def can_install(self) -> bool:
...
+                and posix_os.get_package_information(pkg, use_cached=False) >= "0.52.0"
+            ):
+                package_installed = pkg
```
squirrelsc, line 54: The logic is unclear to me. Does it mean either mason or python3-meson installed, it means installed? Please add code comments to explain the logic, if it's true.

lisa/operating_system.py
```diff
@@ -2094,6 +2094,31 @@ def _initialize_package_installation(self) -> None:
...
+        command = f"zypper --non-interactive {add_args}"
+        if not signed:
+            command += " --no-gpg-checks "
+        command += f" rm {' '.join(packages)}"
+        self.wait_running_process("zypper")
+        install_result = self._node.execute(
+            command, shell=True, sudo=True, timeout=timeout
+        )
+
+        if install_result.exit_code == 0:
```
squirrelsc: Assert the result to get it simpler.

lisa/tools/meson.py
mcgov: It's packaged as 'meson' on older systems and 'python3-meson' on newer ones, since it's actually just a python package. I'll comment it better.

microsoft/testsuites/dpdk/dpdktestpmd.py
mcgov: This is a good comment, the install order of pyelftools turns out to be important. It's listed as a dependency in a few places, but this is kind of confusing. I'll make this more consistent so we're not relying on which tool runs first and possibly trying to install it multiple times.

lisa/tools/meson.py
mcgov: Version checking?

squirrelsc: I mean the package existing check, it should be done in `_check_exists`

lisa/operating_system.py
```diff
@@ -404,13 +404,24 @@ def uninstall_packages(
...
+    ) -> bool:
         """
         Query if a package/tool is installed on the node.
         Return Value - bool
         """
         package_name = self.__resolve_package_name(package)
-        return self._package_exists(package_name)
+        exists = self._package_exists(package_name)
+        if exists and minimum_version:
+            return (
```
squirrelsc: It needs to print debug message, if min version is not met. It saves time to troubleshooting the reason.

lisa/tools/meson.py
mcgov: Ok, this is an case of the word 'exists' being overloaded in our code base. OperatingSystem.package_exists is a bad name IMO. That function is actually a check if a specific _apt/zypper/dnf package_ *is installed*.

Tool._check_exists is supposed to see if _the binary is available_ (and in the path, typically). For Meson we also check the version of the binary since it can be installed but may be an old version. Tool._check_exists should not care whether the binary was installed from the package manager, pip, by hand, or wherever. It looks at the binary in the path.

So I disagree here. It's not appropriate to check OS packages in Tool._check_exists.

LiliDeng:
I still saw issues when run test against canonical ubuntu-24_04-lts server 24.04.202411030

```
2025-01-23T03:45:57.2367877Z 2025-01-23 03:45:57.233[5488][DEBUG] lisa.env[generated_0].node[0].cmd[2463] cmd: ['pip3', 'install', '-q', 'pyelftools'], cwd: None, shell: False, sudo: False, nohup: False, posix: True, remote: True, encoding: utf-8
2025-01-23T03:45:57.5747063Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout error: externally-managed-environment
2025-01-23T03:45:57.5752932Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout � This environment is externally managed
2025-01-23T03:45:57.5757649Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout \u2570\u2500> To install Python packages system-wide, try apt install
2025-01-23T03:45:57.5762505Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     python3-xyz, where xyz is the package you are trying to
2025-01-23T03:45:57.5765328Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     install.
2025-01-23T03:45:57.5771737Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     If you wish to install a non-Debian-packaged Python package,
2025-01-23T03:45:57.5778365Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     create a virtual environment using python3 -m venv path/to/venv.
2025-01-23T03:45:57.5783775Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
2025-01-23T03:45:57.5788002Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     sure you have python3-full installed.
2025-01-23T03:45:57.5794100Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     If you wish to install a non-Debian packaged Python application,
2025-01-23T03:45:57.5799329Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     it may be easiest to use pipx install xyz, which will manage a
2025-01-23T03:45:57.5804530Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     virtual environment for you. Make sure you have pipx installed.
2025-01-23T03:45:57.5810086Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout     See /usr/share/doc/python3.12/README.venv for more information.
2025-01-23T03:45:57.5822285Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
2025-01-23T03:45:57.5825567Z 2025-01-23 03:45:57.573[5492][DEBUG] lisa.env[generated_0].node[0].cmd[2463].stdout hint: See PEP 668 for the detailed specification.
2025-01-23T03:45:57.6262525Z 2025-01-23 03:45:57.625[5488][DEBUG] lisa.env[generated_0].node[0].cmd[2463] execution time: 0.389 sec, exit code: 1
```

lisa/operating_system.py
```diff
@@ -404,13 +404,32 @@ def uninstall_packages(
...
+        minimum_version: Optional[VersionInfo] = None,
+    ) -> bool:
         """
         Query if a package/tool is installed on the node.
         Return Value - bool
         """
         package_name = self.__resolve_package_name(package)
-        return self._package_exists(package_name)
+        exists = self._package_exists(package_name)
+        if exists and minimum_version is not None:
```
squirrelsc: Make it clearer like
```python
self._log.debug(f"package '{package}' exists: {exists}")
if minimum_version is None or not exists:
    return exists

actual_version = self.get_package_information(package_name=package_name)
self._log.debug(f"package '{package}' expected min version: {minimum_version}, actual version: {actual_veresion}")
return actual_version >= minimum_version
```

lisa/operating_system.py
```diff
@@ -2094,6 +2113,30 @@ def _initialize_package_installation(self) -> None:
...
+        command = f"zypper --non-interactive {add_args}"
+        if not signed:
+            command += " --no-gpg-checks "
+        remove_packages = " ".join(packages)
+        command += f" rm {remove_packages}"
+        self.wait_running_process("zypper")
+        install_result = self._node.execute(
+            command, shell=True, sudo=True, timeout=timeout
+        )
+        assert_that(install_result.exit_code).described_as(
```
squirrelsc: install_result.assert_exist_code implemented the similar error message, please reuse it.

lisa/operating_system.py
```diff
@@ -2094,6 +2113,30 @@ def _initialize_package_installation(self) -> None:
...
+        extra_args: Optional[List[str]] = None,
+    ) -> None:
+        add_args = self._process_extra_package_args(extra_args)
+        command = f"zypper --non-interactive {add_args}"
+        if not signed:
+            command += " --no-gpg-checks "
+        remove_packages = " ".join(packages)
+        command += f" rm {remove_packages}"
+        self.wait_running_process("zypper")
+        install_result = self._node.execute(
```
squirrelsc: `install_result` -> `uninstall_result`

lisa/tools/meson.py
```diff
@@ -15,13 +15,18 @@
 
 
 class Meson(Tool):
+    _minimum_version = VersionInfo(major=0, minor=52, patch=0)
```
squirrelsc: use `lisa.util.parse_version`.

lisa/tools/meson.py
```diff
@@ -30,12 +35,39 @@ def can_install(self) -> bool:
...
+        # since it's actually just a python package.
+        # So check for both
+        for pkg in [
+            "python3-meson",
+            "meson",
+        ]:
+            if posix_os.package_exists(pkg, minimum_version=self._minimum_version):
+                package_installed = pkg
+                break
+            elif posix_os.is_package_in_repo(pkg):
```
squirrelsc: `elif` to `if` is the same, but it means the two `if` blocks are actually independent.

mcgov:
> I still saw issues when run test against canonical ubuntu-24_04-lts server 24.04.202411030
@LiliDeng Correct! Fixed this with the latest two commits. Apologies.

lisa/tools/meson.py
mcgov: Can you be more specific with your request from me?

squirrelsc:
```suggestion
            if posix_os.is_package_in_repo(pkg):
```

mcgov: Right. So making _just_ that change will introduce an unintended side effect of updating the package before we check the version of the currently installed package. Let me see if I can make the decisions more clear in the code. I was _intentionally_ not installing an updated version from the package manager if one was already installed because it was convenient for my meson use cases... But really it would be better to write for the general case; even though it will be a little uglier and use another round trip or two.

mcgov: It wasn't so bad to rework! Testing the more commented, reworked version now

mcgov:
rebased to current main

lisa/tools/meson.py
squirrelsc: Use LISA's internal method to support more flexible version formats while preserving the original string.

mcgov: Nice, thanks :)

lisa/tools/meson.py
```diff
@@ -4,24 +4,28 @@
...
 from lisa.operating_system import Posix
+from lisa.util import LisaVersionInfo
 
 from .ln import Ln
 from .python import Pip
 from .whoami import Whoami
 
 
 class Meson(Tool):
+    _minimum_version = LisaVersionInfo.parse("0.52.0")
```
squirrelsc: Please use parse_version method in lisa.util.

lisa/tools/meson.py
```diff
@@ -4,24 +4,28 @@
...
     @property
     def command(self) -> str:
         return "meson"
 
     def _check_exists(self) -> bool:
         result = self.node.execute("meson --version", shell=True)
-        return result.exit_code == 0 and VersionInfo.parse(result.stdout) >= "0.52.0"
+        return (
+            result.exit_code == 0
+            and LisaVersionInfo.parse(result.stdout) >= self._minimum_version
```
squirrelsc: `parse_version`

lisa/tools/meson.py
```diff
@@ -30,24 +34,75 @@ def can_install(self) -> bool:
...
+
+        # otherwise, install the available package from the repo
+        if package_available:
+            posix_os.install_packages(package_available)
+            # and update the cached version info
+            posix_os.get_package_information(pkg, use_cached=False)
+            package_installed = package_available
+
+        # check the version, return if it's good, remove if not
+        if package_installed:
```
squirrelsc, line 69: This logic should be merged with previous `if package_installed`. It's like: 

```
if package_installed:
    if version is ok:
        return check exists
    else:
        uninstall

if package_available:
    install (assume the version is ok.)
    return check exists

```

mcgov: We can't assume the version is ok, but I think I can merge some of the logic. I'll ping you when it's updated and tested

squirrelsc: It doesn't assume the version is valid, so my logic uninstalls the package if the version is invalid. **else: uninstall**

mcgov: @squirrelsc updated, now we just install if there's an update then check the version once (and uninstall if it doesn't meet the requirements)

squirrelsc: It seems to be missing a login step to reinstall the correct version. If v2 is expected but v1 is installed, can it uninstall v1 and install v2?

mcgov: Installing a (later) version from the package manager will implicitly update the version to a later version. You don't need to call apt remove before upgrading a package. If one is available, it will be a later version.
The 'installed but correct version' case is already covered before we get to _install(). The tool will call _check_exists.
If none of that works, we do the less desireable workaround of messing with pip, and the package manager version is uninstalled first in that case.

squirrelsc:
@LiliDeng LGTM

squirrelsc (approved):

PR #3600, opened 2025-01-15
hyperv: expand root partition

gamora12:
The vm root partition (mounted at /) is not getting expanded to full disk size with dom0 3.0 vhd. With 2.0 vhd, this issue is not seen, resize vhd works as expected.

- Implement a function to expand root partition to full disk size on hyperv platform.
- After the vm starts, the root partition is expanded.

### DIFF
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+import re
 from functools import partial
 from pathlib import PurePath
 from typing import Any, List, Optional, Type, cast
@@ -8,7 +9,7 @@ from lisa import feature, schema, search_space
 from lisa.environment import Environment
 from lisa.node import RemoteNode
 from lisa.platform_ import Platform
-from lisa.tools import Cp, HyperV, Mkdir, PowerShell
+from lisa.tools import Cp, HyperV, Mkdir, Mount, PowerShell
 from lisa.util import LisaException, constants
 from lisa.util.logger import Logger, get_logger
 from lisa.util.parallel import run_in_parallel
@@ -314,6 +315,9 @@ class HypervPlatform(Platform):
             node.set_connection_info(
                 address=ip_addr, username=username, password=password
             )
+            # In some cases, we observe that resize vhd resizes the entire disk
+            # but fails to expand the partition size.
+            self._expand_root_partition(node)
 
     def _resize_vhd_if_needed(
         self, vhd_path: PurePath, node_runbook: HypervNodeSchema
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -326,6 +328,23 @@ def _resize_vhd_if_needed(
...
+        # Grow the partition and resize the filesystem
+        cmd_result = node.execute(
+            f"growpart /dev/{partition} {root_part_num}", sudo=True
+        )
+        # In case, the partition is already expanded to full disk size, the
+        # command will print "NOCHANGE: partition 2 is size <size>. it cannot
+        # be grown". In this case, it returns exit code 1 which we can ignore
+        if cmd_result.exit_code != 0 and "NOCHANGE" not in cmd_result.stdout:
+            raise LisaException(f"Failed to grow partition: {cmd_result.stdout}")
+        node.execute(f"resize2fs {device_name}", sudo=True)
```
smit-gardhariya: Use expected_exit_code to ensure if command get executed successfully

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -315,6 +315,8 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
                 address=ip_addr, username=username, password=password
             )
 
+            self._expand_root_partition(node)
```
squirrelsc, line 320: How to make sure it's only apply to hyperv specific case? What happened, if the size of root partition is enough?

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -326,6 +328,23 @@ def _resize_vhd_if_needed(
                 f"-SizeBytes {node_runbook.osdisk_size_in_gb * 1024 * 1024 * 1024}"
             )
 
+    def _expand_root_partition(self, node: RemoteNode) -> None:
+        # Get the root partition info
+        root_partition = node.tools[Mount].get_partition_info("/")[0]
```
squirrelsc, line 338: Use regexp (lisa builtin patterns) and provide examples above.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -326,6 +328,23 @@ def _resize_vhd_if_needed(
...
+        device_name = root_partition.name
+        partition = root_partition.disk
+        root_part_num = device_name[-1]
```
squirrelsc: Provide example of the name above.

lisa/sut_orchestrator/hyperv/platform_.py
anirudhrb: @smit-gardhariya There is a case where we want to treat a non-zero error code as success. Please read the comment.

anirudhrb: Ah, your comment is for the resize2fs command. Never mind.

gamora12: @smit-gardhariya alright

lisa/sut_orchestrator/hyperv/platform_.py
gamora12: This will be applicable to all the nodes created on hyperv platform. In case, the root partition is already expanded, it would print something like- `NOCHANGE: partition 2 is size 1073217503. it cannot be grown`. This is what we're checking in the condition below- `if cmd_result.exit_code != 0 and "NOCHANGE" not in cmd_result.stdout:`

anirudhrb, line 320: Add a comment here explaining why we need this even though we're doing `_resize_vhd_if_needed()` above.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -326,6 +328,23 @@ def _resize_vhd_if_needed(
...
+        # Grow the partition and resize the filesystem
+        cmd_result = node.execute(
+            f"growpart /dev/{partition} {root_part_num}", sudo=True
+        )
+        # In case, the partition is already expanded to full disk size, the
```
anirudhrb:
```suggestion

        # In case the partition is already expanded to full disk size, the
```

lisa/sut_orchestrator/hyperv/platform_.py
gamora12: We're not using regex here. The function `get_partition_info` in the mount tool finds the root partition info, it takes the mount point ("/") as input. The output is a list of name, disk, mountpoint, etc.

lisa/sut_orchestrator/hyperv/platform_.py
squirrelsc: It looks a regexp is more accurate, in case the number has two or more digits.

lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -326,6 +328,28 @@ def _resize_vhd_if_needed(
...
+        # command will print "NOCHANGE: partition 2 is size <size>. it cannot
+        # be grown". In this case, it returns exit code 1 which we can ignore
+        if cmd_result.exit_code != 0 and "NOCHANGE" not in cmd_result.stdout:
+            raise LisaException(f"Failed to grow partition: {cmd_result.stdout}")
+        node.execute(f"resize2fs {device_name}", sudo=True, expected_exit_code=0)
```
squirrelsc, line 356: Execute only if it can be grown. Reduce the number of running commands, even it do nothing.

gamora12: okay

lisa/sut_orchestrator/hyperv/platform_.py
gamora12: Yes, it's better to use regexp for this, I'll update this.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@anirudhrb please fix CI check error.

LiliDeng (approved):

PR #3602, opened 2025-01-16
Changed to allow regular user to run kselftest

paulli2017:
Changed to allow regular user to run kselftest


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3603, opened 2025-01-16
Security Profile Information

kamalca:
The test case results need a way to differentiate between standard, secure boot, and CVM; as well as between disk encryption and no disk encryption.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -53,7 +53,12 @@ from retry import retry
 
 from lisa import feature, schema, search_space
 from lisa.environment import Environment
-from lisa.features import Disk
+from lisa.features import (
+    Disk,
+    SecurityProfile,
+    SecurityProfileSettings,
+    SecurityProfileType,
+)
 from lisa.features.availability import AvailabilityType
 from lisa.node import Node, RemoteNode, local
 from lisa.platform_ import Platform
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -716,6 +720,21 @@ def _get_node_information(self, node: Node) -> Dict[str, str]:
             node.log.debug("detecting vm generation...")
             information[KEY_VM_GENERATION] = node.tools[VmGeneration].get_generation()
             node.log.debug(f"vm generation: {information[KEY_VM_GENERATION]}")
+
+            if node.capability and node.capability.features:
+                for feat in node.capability.features:
```
squirrelsc: Why does it check the capability of node, not check by node.features.is_supported? The capability is what's supported by the node, but it's not the runtime value.

kamalca: @squirrelsc 
Since this logic only runs when `node.is_connected` the security profile capability should be resolved to the runtime value. I used some of the same logic from `SecurityProfile.on_before_deployment`.

Right now, `SecurityProfile.enabled` returns True and `SecurityProfie.create_settings` always returns a value. But I think your point is in cases where SecurityProfile is disabled or no settings are returned?

I can try to make this logic more robust, but **can you help me understand the scenario in which you expect it will break?**

kamalca: @squirrelsc I believe I have fixed this issue. Please see latest changes :)

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3609, opened 2025-01-22
Hyper-V: Auto select default switch

SRIKKANTH:


lisa/tools/hyperv.py
```diff
@@ -209,23 +211,26 @@ def enable_device_passthrough(self, name: str, mmio_mb: int = 5120) -> None:
...
+                )
+                if switch_json:
+                    self._default_switch = VMSwitch()
+                    self._default_switch.name = switch_json["Name"]
+                    self._default_switch.type = switch_type
+                    break
 
-        return VMSwitch.from_json(switch_json)  # type: ignore
+            if self._default_switch is None:
+                raise LisaException("Could not find any default switch")
```
squirrelsc: The error message should be like `couldn't find any internal or external switch.`.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3611, opened 2025-01-23
Enable deploying Azure Storage account File shares with user specified storage quota.

shekharsorot:
Changes done: Addition of "quota" int param to get_or_create_file_share() in commons.py and create_share() in features.py

Why? : Current code deploys a default 100GB volume when provisioning SMB/NFS shares using Azure Storage account. This may ( and in current scenario) cause tests such as XFSTest to fail for specific test cases ( FSStress, Zero fill perf, etc ).

Default Behavior: When called, get_or_create_file_share() will continue to deploy 100B shares unless param 'quota' is passed as an integer value corresponding to GiB of volume we want to provision. Existing extensions using this function will continue to behave as is without disruption.

Things to consider when using: Code currently does not check if quota > 100 is supplied as param. This will cause REST API to return with malformed header error for Storage_V2 and Premium types. Also, Azure Python API documentation states that quota is specified in Bytes, which is incorrect. All values passed should be in GiB ( 1000 MiB -> 1 GiB). This is done since we would eventually like to support PV2 billing model.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1986,7 +1986,7 @@ def get_or_create_file_share(
     resource_group_name: str,
     log: Logger,
     protocols: str = "SMB",
-    quota: int = 100,
+    quota_in_gb: int = 100,
 ) -> str:
     """
     Create an Azure Storage file share if it does not exist.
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1984,9 +1986,10 @@ def get_or_create_file_share(
     resource_group_name: str,
     log: Logger,
     protocols: str = "SMB",
+    quota: int = 100,
```
squirrelsc: `quota_in_gb`

shekharsorot: updated

LiliDeng:
@shekharsorot please fix the CI check error.

LiliDeng (approved):

PR #3613, opened 2025-01-23
Improvements to "HyperVPreparationTransformer"

SRIKKANTH:
Changed the transformer to use recently added methods to HyperV tool instead of running commands.

### DIFF
--- a/lisa/transformers/hyperv_preparation.py
+++ b/lisa/transformers/hyperv_preparation.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict, List, Type
 
 from lisa import schema
-from lisa.tools import PowerShell, HyperV
+from lisa.tools import HyperV
 from lisa.transformers.deployment_transformer import (
     DeploymentTransformer,
     DeploymentTransformerSchema,
--- a/lisa/transformers/hyperv_preparation.py
+++ b/lisa/transformers/hyperv_preparation.py
@@ -1,7 +1,7 @@
 from typing import Any, Dict, List, Type
 
 from lisa import schema
-from lisa.tools import PowerShell
+from lisa.tools import PowerShell, HyperV
 from lisa.transformers.deployment_transformer import (
     DeploymentTransformer,
     DeploymentTransformerSchema,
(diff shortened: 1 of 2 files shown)


lisa/transformers/hyperv_preparation.py
```diff
@@ -28,23 +28,16 @@ def _output_names(self) -> List[str]:
...
-        )
-        powershell.run_cmdlet(
-            "New-NetNat -Name 'InternalNAT' -InternalIPInterfaceAddressPrefix '192.168.0.0/24'",  # noqa: E501
-            force_run=True,
-        )
-        powershell.run_cmdlet(
-            'New-NetIPAddress -IPAddress 192.168.0.1 -InterfaceIndex (Get-NetAdapter | Where-Object { $_.Name -like "*InternalNAT)" } | Select-Object -ExpandProperty ifIndex) -PrefixLength 24',  # noqa: E501
-            force_run=True,
-        )
+        switch_name = "InternalNAT"
```
LiliDeng, line 31: use HypervSwitchType.INTERNAL?

SRIKKANTH: Its just a name @LiliDeng it can be anything. HypervSwitchType.INTERNAL == 'Internal' not "InternalNAT"

LiliDeng: we can set InternalNAT as default value of name in hv.create_switch

squirrelsc: It's ok to have any name, but the logic needs to check if there is an internal switch already. If there is, don't add a new one.

SRIKKANTH: hv.create_switch already handles deletion of existing switch with same name and re-create a new one.

squirrelsc: It should avoid to delete/create. If a switch exists, it should be reused. It can avoid unstable problems on add/remove switches. And sometimes, the lab may need create a special switch, it shouldn't be deleted every time.

SRIKKANTH: Got it. Let me check with Virt team who does this testing in Lab environments. I only test Azure Windows HyperV hosts where each host is created and destroyed after test completion.

squirrelsc: You can use Azure to test it too, just keep environment and run again.

SRIKKANTH: Improved as suggested @squirrelsc

LiliDeng (approved):

PR #3614, opened 2025-01-24
Check all log locations for if serial console enabled

katmnsn:
There is a potential bug in the logic of this test, where only some log locations are checked for the target substring.  I manually ran the commands on an image which failed this test and discovered that the substring could be detected in the journalctl log.  The test now checks all possible locations for the substring to ensure that it will be detected if present.

### DIFF
--- a/microsoft/testsuites/core/azure_image_standard.py
+++ b/microsoft/testsuites/core/azure_image_standard.py
@@ -798,52 +798,29 @@ class AzureImageStandard(TestSuite):
 
     @TestCaseMetadata(
         description="""
-        This test will check the serial console is enabled from kernel command line
-         in dmesg.
+        This test will check the serial console is enabled from kernel command line.
 
         Steps:
-        1. Get the kernel command line from /var/log/messages or
-            /var/log/syslog output.
+        1. Get the kernel command line from /var/log/messages, /var/log/syslog,
+            dmesg, or journalctl output.
         2. Check expected setting from kernel command line.
             2.1. Expected to see 'console=ttyAMA0' for aarch64.
             2.2. Expected to see 'console=ttyS0' for x86_64.
+            2.3. Expected to see 'uart0: console (115200,n,8,1)' for FreeBSD.
         """,
         priority=1,
         requirement=simple_requirement(supported_platform_type=[AZURE, READY]),
     )
     def verify_serial_console_is_enabled(self, node: Node) -> None:
-        console_device = {
-            CpuArchitecture.X64: "ttyS0",
-            CpuArchitecture.ARM64: "ttyAMA0",
-        }
         if isinstance(node.os, CBLMariner):
             if node.os.information.version < "2.0.0":
                 raise SkippedException(
                     "CBLMariner 1.0 has a known 'wont fix' issue with this test"
                 )
-        if isinstance(node.os, CBLMariner) or (
-            isinstance(node.os, Ubuntu) and node.os.information.version >= "22.10.0"
-        ):
-            log_output = node.tools[Dmesg].get_output()
-            log_file = "dmesg"
-        else:
-            cat = node.tools[Cat]
-            if node.shell.exists(node.get_pure_path("/var/log/messages")):
-                log_file = "/var/log/messages"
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -798,68 +798,69 @@ def verify_repository_installed(self, node: Node) -> None:  # noqa: C901
...
+            ]
+            journalctl_tool = node.tools[Journalctl]
+            for log_file, tool in log_sources:
+                if node.shell.exists(node.get_pure_path(log_file)):
+                    current_log_output = tool.read(log_file, force_run=True, sudo=True)
+                    if current_log_output:
+                        if any(find_patterns_in_lines(current_log_output, patterns)):
+                            pattern_found = True
+                            break
+            if not pattern_found and journalctl_tool._check_exists():
```
squirrelsc: use `journalctl_tool`.exists. any method starts with `_` is a private method, shouldn't be used by other classes.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -798,68 +798,69 @@ def verify_repository_installed(self, node: Node) -> None:  # noqa: C901
...
+                if current_log_output:
+                    if any(find_patterns_in_lines(current_log_output, patterns)):
+                        pattern_found = True
+        if not pattern_found:
             raise LisaException(
                 "Fail to find console enabled line "
                 f"'console [{current_console_device}] enabled' "
                 "or 'uart0: console (115200,n,8,1)' "
-                f"from {log_file} output",
+                "from any log output"
```
squirrelsc: List names also. The error message is important, when the failure happens.

microsoft/testsuites/core/azure_image_standard.py
katmnsn: good catch, fixed

microsoft/testsuites/core/azure_image_standard.py
katmnsn: added a way to print the log file that was checked for each code path, if the test fails

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -798,68 +798,74 @@ def verify_repository_installed(self, node: Node) -> None:  # noqa: C901
...
+            logs_checked.append(f"{dmesg_tool.command}")
+            if any(find_patterns_in_lines(log_output, patterns)):
+                pattern_found = True
+        else:
+            log_sources = [
+                ("/var/log/syslog", node.tools[Cat]),
+                ("/var/log/messages", node.tools[Cat]),
+            ]
+            journalctl_tool = node.tools[Journalctl]
+            for log_file, tool in log_sources:
```
squirrelsc, line 851: add some comment to explain what's checked, and the strategy to cover different sources.

squirrelsc:
Make commits meaningful.

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -798,68 +798,79 @@ def verify_repository_installed(self, node: Node) -> None:  # noqa: C901
...
 
         Steps:
-        1. Get the kernel command line from /var/log/messages or
-            /var/log/syslog output.
+        1. Get the kernel command line from /var/log/messages, /var/log/syslog,
+            dmesg, or journalctl output.
         2. Check expected setting from kernel command line.
             2.1. Expected to see 'console=ttyAMA0' for aarch64.
             2.2. Expected to see 'console=ttyS0' for x86_64.
+            2.3. Expected to see 'uart0: console 115200,n,8,1 for FreeBSD.
```
LiliDeng: `'uart0: console (115200,n,8,1)'`

katmnsn: good catch, fixed that

microsoft/testsuites/core/azure_image_standard.py
katmnsn: I added the comments, let me know if they're enough or more needed

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3615, opened 2025-01-25
[Hyper-v platform] Enabling port forwarding when switch type is Internal

SRIKKANTH:
Virtual machines with an internal switch type cannot be accessed from outside the host. To enable external access, you need to use NAT and port mapping.

### DIFF
--- a/lisa/tools/hyperv.py
+++ b/lisa/tools/hyperv.py
@@ -559,4 +559,4 @@ class HyperV(Tool):
         if port in self._assigned_nat_ports:
             self._assigned_nat_ports.remove(port)
         else:
-            print(f"Port {port} was not assigned.")
+            self._log.debug(f"Port {port} was not assigned.")
--- a/lisa/tools/hyperv.py
+++ b/lisa/tools/hyperv.py
@@ -5,7 +5,7 @@ import re
 import time
 from dataclasses import dataclass
 from enum import Enum
-from typing import Dict, Optional
+from typing import Any, Dict, Optional, Set
 
 from assertpy import assert_that
 from dataclasses_json import dataclass_json
@@ -32,9 +32,18 @@ class VMSwitch:
 
 
 class HyperV(Tool):
+    # Internal NAT network configuration
+    INTERNAL_NAT_ROUTER = "192.168.5.1"
+    INTERNAL_NAT_SUBNET = "192.168.5.0/24"
+    INTERNAL_NAT_DHCP_IP_START = "192.168.5.50"
+    INTERNAL_NAT_DHCP_IP_END = "192.168.5.100"
+    # Azure DNS server
+    AZURE_DNS_SERVER = "168.63.129.16"
     # 192.168.5.12
     IP_REGEX = r"\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}"
     _default_switch: Optional[VMSwitch] = None
+    _external_forwarding_port_start = 50000
+    _assigned_nat_ports: Set[int] = set()
 
     @property
     def command(self) -> str:
@@ -57,6 +66,10 @@ class HyperV(Tool):
         # check if vm is present
         if not self.exists_vm(name):
             return None
+
+        # delete port mapping for internal IP address of the VM
+        internal_ip = self.get_ip_address(name)
+        self.delete_nat_mapping(internal_ip=internal_ip)
 
         # stop and delete vm
         self.stop_vm(name=name)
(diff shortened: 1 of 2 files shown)


lisa/tools/hyperv.py
```diff
@@ -293,6 +293,37 @@ def create_nat(self, name: str, ip_range: str) -> None:
             force_run=True,
         )
 
+    def add_nat_mapping(
+        self, nat_name: str, internal_ip: str, external_port: int
+    ) -> None:
+        # delete existing NAT mapping
+        self.delete_nat_mapping(external_port)
```
squirrelsc, line 300: maintain the assigned external ports in the hyperV tool.
1. It can still set the start port number, but the default value is 50000.
2. In the hyperV tool, it saves assigned ports.
3. When it needs to assign a new port, start from the loweest unassigned port number.

lisa/tools/hyperv.py
```diff
@@ -293,6 +293,37 @@ def create_nat(self, name: str, ip_range: str) -> None:
             force_run=True,
         )
 
+    def add_nat_mapping(
```
squirrelsc: maintain the assigned external ports in the hyperV tool.

1. It can still set the start port number, but the default value is 50000.
1. In the hyperV tool, it saves assigned ports.
1. When it needs to assign a new port, start from the loweest unassigned port number.
1. This method returns the assigned port number.

lisa/tools/hyperv.py
```diff
@@ -432,13 +463,13 @@ def enable_internal_dhcp(self, dhcp_scope_name: str = "DHCPInternalNAT") -> None
 
         # Configure the DHCP server to use the internal NAT network
         powershell.run_cmdlet(
-            f'Add-DhcpServerV4Scope -Name "{dhcp_scope_name}" -StartRange 192.168.0.50 -EndRange 192.168.0.100 -SubnetMask 255.255.255.0',  # noqa: E501
+            f'Add-DhcpServerV4Scope -Name "{dhcp_scope_name}" -StartRange 192.168.5.50 -EndRange 192.168.5.100 -SubnetMask 255.255.255.0',  # noqa: E501
```
squirrelsc: Why change the ip range?

SRIKKANTH: To use the same subnet as other parts of the code like 'setup_nat_networking()'

lisa/tools/hyperv.py
SRIKKANTH: Modified as suggested.

lisa/tools/hyperv.py
squirrelsc: Please define consts to reuse.

lisa/tools/hyperv.py
```diff
@@ -293,6 +295,40 @@ def create_nat(self, name: str, ip_range: str) -> None:
             force_run=True,
         )
 
+    def add_nat_mapping(self, nat_name: str, internal_ip: str) -> int:
+        external_port = self._get_next_available_nat_port()
+
+        # delete existing NAT mapping
+        self.delete_nat_mapping(external_port)
```
squirrelsc: The unused ports should be deleted when VM is deleted too. Here should have a check if the port mapping exists, direct fails. It can help identifying potential leaking issues. If we found it's normal scenario, and then add this kind of prevention deletion.

SRIKKANTH: Added port mapping deletion at VM deletion and added a check at creation to abort if port already used.

lisa/tools/hyperv.py
SRIKKANTH: Modified as suggested.

lisa/tools/hyperv.py
```diff
@@ -468,13 +472,13 @@ def enable_internal_dhcp(self, dhcp_scope_name: str = "DHCPInternalNAT") -> None
 
         # Configure the DHCP server to use the internal NAT network
         powershell.run_cmdlet(
-            f'Add-DhcpServerV4Scope -Name "{dhcp_scope_name}" -StartRange 192.168.5.50 -EndRange 192.168.5.100 -SubnetMask 255.255.255.0',  # noqa: E501
+            f'Add-DhcpServerV4Scope -Name "{dhcp_scope_name}" -StartRange {self.internal_nat_dhcp_ip_start} -EndRange {self.internal_nat_dhcp_ip_end} -SubnetMask 255.255.255.0',  # noqa: E501
```
squirrelsc: Remove `# noqa: E501`. Only url needs it for copying/opening easily. Otherwise, break the string can improve readability.

SRIKKANTH: Removed # noqa: E501 from one more place along with this one.

squirrelsc:
@LiliDeng LGTM

lisa/tools/hyperv.py
```diff
@@ -482,3 +544,19 @@ def _run_hyperv_cmdlet(
...
+            port += 1
+        self._assigned_nat_ports.add(port)  # Assign this port
+        return port
+
+    def _release_nat_port(self, port: int) -> None:
+        # Release a port if used
+        if port in self._assigned_nat_ports:
+            self._assigned_nat_ports.remove(port)
+        else:
+            print(f"Port {port} was not assigned.")
```
LiliDeng: use log here

LiliDeng (approved):

PR #3617, opened 2025-01-27
New Notifier to check the DMesg Logs for OOPS Exceptions

umfranci:
At the moment when a test case is executed, the Pass/Fail status is specific to the tasks executed as part of the test but the OOPS exceptions are not checked. Hence creating a notifier that can run a regex pattern against the DMesg logs and capture the contextual log lines around the found OOPS. 

These findings will then be saved in the artifacts generated against the tests which can further be utilized for creating any kind of notifications/alerts.

### DIFF
--- a/lisa/mixin_modules.py
+++ b/lisa/mixin_modules.py
@@ -11,6 +11,7 @@ import lisa.combinators.csv_combinator  # noqa: F401
 import lisa.combinators.git_bisect_combinator  # noqa: F401
 import lisa.combinators.grid_combinator  # noqa: F401
 import lisa.notifiers.console  # noqa: F401
+import lisa.notifiers.dmesg_oops  # noqa: F401
 import lisa.notifiers.env_stats  # noqa: F401
 import lisa.notifiers.file  # noqa: F401
 import lisa.notifiers.html  # noqa: F401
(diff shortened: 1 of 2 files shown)


squirrelsc:
The notifier isn't a right place for panic checks, and the dmesg tool handles more scenarios. The notifier cannot fail the test case, so the results is hard to associate with the failed cause. 

1. If the purpose wants to capture kernel panic or call trace in some test cases, you can always call `Dmesg.check_kernel_errors` in the test case.
2. If it wants to capture errors in general, you can check what LISA have now, and may apply an extra check switch, so any kernel error happens, fail the just ran test case and mark the env bad.

umfranci:
> The notifier isn't a right place for panic checks, and the dmesg tool handles more scenarios. The notifier cannot fail the test case, so the results is hard to associate with the failed cause.
> 
> 1. If the purpose wants to capture kernel panic or call trace in some test cases, you can always call `Dmesg.check_kernel_errors` in the test case.
> 2. If it wants to capture errors in general, you can check what LISA have now, and may apply an extra check switch, so any kernel error happens, fail the just ran test case and mark the env bad.

@squirrelsc Thanks for the comment. The purpose of the Notifier is not to fail the test cases in case of an OOPS/Kernel Error but to simply save the detections in a file, which can then be later utilized for notification purposes. There are many tests which do not check the Dmesg logs for issues since the purpose of the test case might not necessarily need it - though it was still important to check Dmesg logs for each of the tests without actually failing them.

PR #3620, opened 2025-01-30
libvirt: Use 16/32 hex long host-data based on libvirt version

smit-gardhariya:
Use token_hex with 16/32 to generate hex string
for host data based on libvirt version.

Also use hex format in attestation tool to compare as libvirt convert host data in ASCII hex string. Based on node context, either decode or use plain string to compare as libvirt expect base64 encoded string for >=10.5.0 version


anirudhrb (approved):

praveen-pk:
LGTM!

squirrelsc (approved):

PR #3622, opened 2025-01-31
Refine doc for transformer output

LiliDeng:


docs/run_test/transformers.rst
```diff
@@ -179,6 +179,7 @@ Usage
 
   transformer:
     - type: azure_deploy
+      name: custom_name
```
squirrelsc: It should be introduced in a general section, not for one of transformer.

squirrelsc (approved):

PR #3624, opened 2025-02-03
CVM: Add boot/reboot tests

### LINKED ISSUE #3625: Posix.get_package_information fails to parse package version if it does not follow semver
The current implementation of `Posix.get_package_information` uses the package manager for a distro to query for a package information, then uses some regexes to force parse the package version into semver. This logic fails if a package does not follow semantic versioning for their versioning scheme. Example for Azure Linux (which uses rpm):

```python3
>>> import re
>>> r = re.compile(
           r"(?P<package_name>[a-zA-Z0-9\-_]+)-"
           r"(?P<major>[0-9]+)\."
           r"(?P<minor>[0-9]+)\.?"
           r"(?P<patch>[0-9]+)?"
           r"(?P<build>-[a-zA-Z0-9-_\.]+)?"
       )
>>> print(r.search("systemd-255-20.azl3.x86_64"))
None
```

Suggestion: query the version directly using a distro's package tool/manager (dpkg/apt, rpm, etc), then return the raw version string as part of the result. If there is any logic to check versioning of a package, that logic can be implemented for each distro. Here is an example of how one can query a package version for debian and rpm based systems:
- Debian: `dpkg-query -f '${Version}' -W <pkg>`
- RPM: rpm --queryformat '%{VERSION}-%{RELEASE}' -q <pkg>

I demoed this change in commit https://github.com/microsoft/lisa/pull/3624/commits/365ce3550d803d427542ae81e9c661857ca17f29, adding a new function `_query_package` since I don't want to change existing code yet. Please let me know your thoughts.

trungams:
- Tools:
  - Add tpm2 tool to support reading PCR values from the TPM
  - Add bootctl tool
- Tests: add boot tests for Azure Linux CVM:
  - verify_encrypted_root_partition: sanity check that the root partition on an Azure Linux CVM is encrypted when deployed with `DiskWithVMGuestState` encryption setting.
  - verify_boot_success_after_component_upgrade: check that a CVM can reboot after a boot component is upgraded. 
- Core: include raw version string in Posix.get_package_information return value in case the package does not follow semantic versioning (fixes #3625)

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -39,6 +39,7 @@ from lisa.util import (
     BaseClassMixin,
     LisaException,
     LisaTimeoutException,
+    LisaVersionInfo,
     MissingPackagesException,
     ReleaseEndOfLifeException,
     RepoNotExistException,
@@ -152,7 +153,7 @@ class OperatingSystem:
         self._is_posix = is_posix
         self._log = get_logger(name="os", parent=self._node.log)
         self._information: Optional[OsInformation] = None
-        self._packages: Dict[str, VersionInfo] = dict()
+        self._packages: Dict[str, LisaVersionInfo] = dict()
 
     @classmethod
     def create(cls, node: "Node") -> Any:
@@ -494,7 +495,7 @@ class Posix(OperatingSystem, BaseClassMixin):
 
     def get_package_information(
         self, package_name: str, use_cached: bool = True
-    ) -> VersionInfo:
+    ) -> LisaVersionInfo:
         found = self._packages.get(package_name, None)
         if found and use_cached:
             return found
@@ -549,12 +550,24 @@ class Posix(OperatingSystem, BaseClassMixin):
         # sub os can override it, but it's optional
         pass
 
-    def _get_package_information(self, package_name: str) -> VersionInfo:
+    def _get_package_information(self, package_name: str) -> LisaVersionInfo:
         raise NotImplementedError()
 
-    def _get_version_info_from_named_regex_match(
-        self, package_name: str, named_matches: Match[str]
-    ) -> VersionInfo:
+    def _get_version_info_from_regex(
+        self, version_str: str, regex: Pattern[str]
+    ) -> LisaVersionInfo:
+        self._node.log.debug(f"Attempting to parse version string ({version_str})")
+        matches = regex.search(version_str)
+        if not matches:
+            self._node.log.debug(
+                f"Could not find regex match for ({version_str}) using regex"
+            )
+            return LisaVersionInfo(version_str, 0)
+        return self._get_version_info_from_named_matches(version_str, matches)
+
+    def _get_version_info_from_named_matches(
+        self, version_str: str, named_matches: Match[str]
+    ) -> LisaVersionInfo:
         essential_matches = ["major", "minor"]
 
         # verify all essential keys are in our match dict
@@ -576,16 +589,16 @@ class Posix(OperatingSystem, BaseClassMixin):
         )
         build_match = named_matches.group("build")
         log_message = (
-            f"Found {package_name} version "
+            f"Found version from ({version_str}) "
             f"major:{major_match} minor:{minor_match} "
             f"patch:{patch_match} build:{build_match}"
         )
         self._node.log.debug(log_message)
-        return VersionInfo(major, minor, patch, build=build_match)
+        return LisaVersionInfo(version_str, major, minor, patch, build=build_match)
 
     def _cache_and_return_version_info(
-        self, package_name: str, info: VersionInfo
-    ) -> VersionInfo:
+        self, package_name: str, info: LisaVersionInfo
+    ) -> LisaVersionInfo:
         self._packages[package_name] = info
         return info
 
@@ -838,32 +851,17 @@ class Debian(Linux):
                 error_lines.append(line)
         return error_lines
 
-    def _get_package_information(self, package_name: str) -> VersionInfo:
-        # run update of package info
-        apt_info = self._node.execute(
-            f"apt show {package_name}",
+    def _get_package_information(self, package_name: str) -> LisaVersionInfo:
+        query_result = self._node.execute(
+            f"dpkg-query -f '${{Version}}' -W {package_name}",
             expected_exit_code=0,
             expected_exit_code_failure_message=(
                 f"Could not find package information for package {package_name}"
             ),
         )
-        match = self._debian_package_information_regex.search(apt_info.stdout)
-        if not match:
-            raise LisaException(
-                "Package information parsing could not find regex match "
-                f" for {package_name} using regex "
-                f"{self._debian_package_information_regex.pattern}"
-            )
-        version_str = match.group(2)
-        match = self._debian_version_splitter_regex.search(version_str)
-        if not match:
-            raise LisaException(
-                f"Could not parse version info: {version_str} "
-                f"for package {package_name}"
-            )
-        self._node.log.debug(f"Attempting to parse version string: {version_str}")
-        version_info = self._get_version_info_from_named_regex_match(
-            package_name, match
+        version_str = query_result.stdout.strip()
+        version_info = self._get_version_info_from_regex(
+            version_str, self._debian_version_splitter_regex
         )
         return self._cache_and_return_version_info(package_name, version_info)
 
@@ -1537,23 +1535,17 @@ class RPMDistro(Linux):
     def clean_package_cache(self) -> None:
         self._node.execute(f"{self._dnf_tool()} clean all", sudo=True, shell=True)
 
-    def _get_package_information(self, package_name: str) -> VersionInfo:
-        rpm_info = self._node.execute(
-            f"rpm -q {package_name}",
+    def _get_package_information(self, package_name: str) -> LisaVersionInfo:
+        query_result = self._node.execute(
+            f"rpm --queryformat '%{{VERSION}}-%{{RELEASE}}' -q {package_name}",
             expected_exit_code=0,
             expected_exit_code_failure_message=(
                 f"Could not find package information for package {package_name}"
             ),
         )
-        # rpm package should be of format (package_name)-(version)
-        matches = self._rpm_version_splitter_regex.search(rpm_info.stdout)
-        if not matches:
-            raise LisaException(
-                f"Could not parse package version {rpm_info} for {package_name}"
-            )
-        self._node.log.debug(f"Attempting to parse version string: {rpm_info.stdout}")
-        version_info = self._get_version_info_from_named_regex_match(
-            package_name, matches
+        version_str = query_result.stdout.strip()
+        version_info = self._get_version_info_from_regex(
+            version_str, self._rpm_version_splitter_regex
         )
         return self._cache_and_return_version_info(package_name, version_info)
 
@@ -2139,33 +2131,17 @@ class Suse(Linux):
         result = self._node.execute(command, sudo=True, shell=True)
         return 0 == result.exit_code
 
-    def _get_package_information(self, package_name: str) -> VersionInfo:
-        # run update of package info
-        zypper_info = self._node.execute(
-            f"zypper info {package_name}",
+    def _get_package_information(self, package_name: str) -> LisaVersionInfo:
+        query_result = self._node.execute(
+            f"rpm --queryformat '%{{VERSION}}-%{{RELEASE}}' -q {package_name}",
             expected_exit_code=0,
             expected_exit_code_failure_message=(
                 f"Could not find package information for package {package_name}"
             ),
         )
-        output = self._ansi_escape.sub("", zypper_info.stdout)
-        match = self._suse_package_information_regex.search(output)
-        if not match:
-            raise LisaException(
-                "Package information parsing could not find regex match "
-                f" for {package_name} using regex "
-                f"{self._suse_package_information_regex.pattern}"
-            )
-        version_str = match.group("package_version")
-        match = self._suse_version_splitter_regex.search(version_str)
-        if not match:
-            raise LisaException(
-                f"Could not parse version info: {version_str} "
-                f"for package {package_name}"
-            )
-        self._node.log.debug(f"Attempting to parse version string: {version_str}")
-        version_info = self._get_version_info_from_named_regex_match(
-            package_name, match
+        version_str = query_result.stdout.strip()
+        version_info = self._get_version_info_from_regex(
+            version_str, self._suse_version_splitter_regex
         )
         return self._cache_and_return_version_info(package_name, version_info)
 
--- a/lisa/tools/bootctl.py
+++ b/lisa/tools/bootctl.py
@@ -1,0 +1,46 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from lisa.executable import Tool
+from lisa.operating_system import CBLMariner
+from lisa.util import UnsupportedDistroException
+
+
+class BootCtl(Tool):
+    @property
+    def command(self) -> str:
+        return "bootctl"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, CBLMariner):
+            self.node.os.install_packages("systemd-udev")
+        else:
+            raise UnsupportedDistroException(
+                self.node.os,
+                f"tool {self.command} can't be installed in {self.node.os.name}",
+            )
+        return self._check_exists()
+
+    def get_esp_path(self) -> str:
+        return self._get_cmd_output("--print-esp-path")
+
+    def get_boot_path(self) -> str:
+        return self._get_cmd_output("--print-boot-path")
+
+    def get_root_device(self) -> str:
+        return self._get_cmd_output("--print-root-device")
+
+    def _get_cmd_output(self, cmd: str) -> str:
+        cmd_result = self.run(
+            cmd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="failed to get ESP path",
+            shell=True,
+            sudo=True,
+            force_run=True,
+        )
+        return cmd_result.stdout
--- a/lisa/tools/tpm2.py
+++ b/lisa/tools/tpm2.py
@@ -1,0 +1,97 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import re
+from typing import Dict, List, Match, Optional, Sequence, Union
+
+from lisa.executable import Tool
+from lisa.operating_system import CBLMariner
+from lisa.util import UnsupportedDistroException
+
+
+class Tpm2(Tool):
+    """tpm2 provides a toolset based on tpm2-tss to interact with TPM 2.0 devices"""
+
+    # example:
+    #   sha256:
+    #       4 : 0x0000000000000000000000000000000000000000000000000000000000000003
+    #       7 : 0x0000000000000000000000000000000000000000000000000000000000000003
+    _pcrread_split_hash_and_pcrs_regex = re.compile(
+        r"(?P<hash>sha\d+):\s*(?P<pcrs>(.*\s*)*)"
+    )
+    _pcrread_split_pcr_index_and_value_regex = re.compile(
+        r"(?P<pcr_index>\d+)\s*:\s*(?P<hash_value>0x[a-fA-F0-9]+)"
+    )
+
+    @property
+    def command(self) -> str:
+        return "tpm2"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, CBLMariner):
+            self.node.os.install_packages("tpm2-tools")
+        else:
+            raise UnsupportedDistroException(
+                self.node.os,
+                f"tool {self.command} can't be installed in {self.node.os.name}.",
+            )
+        return self._check_exists()
+
+    def pcrread(
+        self, pcrs: Optional[Union[int, Sequence[int]]] = None
+    ) -> Dict[int, str]:
+        """List PCR values
+
+        A Platform Configuration Register (PCR) is a memory location in the TPM that
+        stores cryptographic measurements of a system's state.
+
+        Example command/output
+        $ tpm2 pcrread sha256:4,7
+          sha256:
+            4 : 0x0000000000000000000000000000000000000000000000000000000000000003
+            7 : 0x0000000000000000000000000000000000000000000000000000000000000003
+        """
+        alg = "sha256"
+        pcrs = self._get_pcr_list(pcrs)
+        if len(pcrs) == 0:
+            pcrs_arg = "all"
+        else:
+            pcrs_arg = ",".join(map(str, pcrs))
+        cmd = f"pcrread {alg}:{pcrs_arg}"
+        cmd_result = self.run(
+            cmd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="failed to read PCR values",
+            shell=True,
+            sudo=True,
+            force_run=True,
+        )
+        output = cmd_result.stdout
+
+        m = self._pcrread_split_hash_and_pcrs_regex.search(output)
+        assert isinstance(m, Match)
+        hash_alg = m.group("hash")
+        pcrs_info = m.group("pcrs")
+
+        assert (
+            hash_alg == alg
+        ), "pcrread output does not contain the requested algorithm"
+
+        result = dict()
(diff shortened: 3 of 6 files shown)


trungams:
@LiliDeng @squirrelsc - This PR is ready to review. Could you help me review it when you have time? Thank you.

The `query_package` addition - if you are not certain about it, I can split it into another PR.

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,176 @@
...
+        security_profile_settings = cast(
+            SecurityProfileSettings, node.features[SecurityProfile].get_settings()
+        )
+        if not security_profile_settings.encrypt_disk:
+            raise SkippedException("This test requires disk encryption to be enabled")
+        lsblk = node.tools[Lsblk]
+        disks = lsblk.get_disks(force_run=True)
+        partitions: List[PartitionInfo] = next(
+            (d.partitions for d in disks if d.name == "sda"), []
+        )
```
LiliDeng, line 71: please use below code to get root partition, since fixed label name may be changed when VM reboot
```
node_disk = node.features[Disk]
node_disk.get_os_boot_partition()
```

LiliDeng:
could you provide the full cvm mariner image name? like this format: microsoftcblmariner azure-linux-3 azure-linux-3-gen2 latest

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,176 @@
...
+        requirement=simple_requirement(
+            supported_features=[CvmEnabled()],
+            supported_platform_type=[AZURE],
+        ),
+    )
+    def verify_encrypted_root_partition(
+        self,
+        log: Logger,
+        node: RemoteNode,
+        environment: Environment,
```
squirrelsc: It doesn't need to list all of arguments. If it's not used, please remove.

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,176 @@
...
+        boot_components_new_versions = dict()
+        for pkg in boot_components:
+            pkg_info = posix_os.query_package(pkg)
+            boot_components_new_versions[pkg] = pkg_info.version_str
+
+        # Reboot
+        # - Make sure VM boots up again
+        reboot_tool = node.tools[Reboot]
+        reboot_tool.reboot_and_check_panic(log_path)
+        is_ready, tcp_error_code = wait_tcp_port_ready(
```
squirrelsc: It doesn't need to wait for ready in the test case. When you call next command, LISA will do it.

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,176 @@
...
+        assert_that(root_partition, "Cannot locate root partition").is_not_none()
+        assert isinstance(root_partition, PartitionInfo)
+        assert_that(root_partition.fstype).is_equal_to("crypto_LUKS")
+
+    @TestCaseMetadata(
+        description="""This test case verifies that a CVM can still boot if any boot
+        component is upgraded.
+
+        Steps:
+        1. On first boot, check current PCR values for PCR4 and PCR7
```
squirrelsc, line 82: Add requirements on new environment, if it's needed.

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,176 @@
...
+
+        pcrs_after_reboot = node.tools[Tpm2].pcrread(pcrs=[4, 7])
+        boot_component_changed = any(
+            boot_components_versions[pkg] != boot_components_new_versions[pkg]
+            for pkg in boot_components
+        )
+
+        # - PCR4 should change if any of the boot components is upgraded
+        # - PCR7 may change if a signing cert is changed
+        if boot_component_changed:
```
squirrelsc, line 133: Is it possible to make sure all paths tested in the same test case?

lisa/operating_system.py
```diff
@@ -119,6 +119,12 @@ class KernelInformation:
     version_parts: List[str]
 
 
+@dataclass
+class PackageInformation:
```
squirrelsc: Please modify existing `get_package_information`, make it compatible with any version format.

1. Create a class called `LisaVersionInfo`, which inherits from `VersionInfo`, so all existing cases should continue to pass.
2. Provide the `raw_version` field, so it can pass with non-semver.
3. Rewrite `parse_version` to return the new class.

lisa/tools/tpm2.py
```diff
@@ -0,0 +1,67 @@
...
+        if isinstance(self.node.os, CBLMariner):
+            self.node.os.install_packages("tpm2-tools")
+        else:
+            raise LisaException(
+                f"tool {self.command} can't be installed in distro {self.node.os.name}."
+            )
+        return self._check_exists()
+
+    def pcrread(
+        self, alg: str = "sha256", pcrs: Optional[Union[int, Sequence[int]]] = None
```
squirrelsc: Remove the unused argument `alg`.

trungams: It is used here
```python3
cmd = f"pcrread {alg}:{pcrs_arg}"
```

squirrelsc: I mean it's not assigned value when calling the method. So it should be removed, until it's really used when calling the method.

trungams: ah okay. Will do

microsoft/testsuites/cvm/cvm_boot.py
trungams: I have defined requirement for this test case below (CvmEnabled, AZURE). Does this suffice?

microsoft/testsuites/cvm/cvm_boot.py
trungams: Do you mean the `wait_tcp_port_ready` call can be removed? I was mostly following the pattern in smoke_test and a boot test defined in microsoft/testsuites/core/boot.py

microsoft/testsuites/cvm/cvm_boot.py
trungams: I think this logic already covers what we want to test for CVM booting, at least for Azure Linux. If you're referring to merging the 2 test cases defined in this file, I think it's possible. If that's preferred then I'll update my code

microsoft/testsuites/cvm/cvm_boot.py
trungams: I want to get the root partition rather than boot partition, but following the suggestion, I added a call to `bootctl` to get the root device path more reliably.

trungams:
@LiliDeng The image I used to test this change is `MicrosoftCBLMariner azure-linux-3 azure-linux-3-cvm latest`

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,164 @@
...
+        requirement=simple_requirement(
+            supported_features=[CvmEnabled()],
+            supported_platform_type=[AZURE],
+        ),
+    )
+    def verify_encrypted_root_partition(self, node: RemoteNode) -> None:
+        security_profile_settings = cast(
+            SecurityProfileSettings, node.features[SecurityProfile].get_settings()
+        )
+        if not security_profile_settings.encrypt_disk:
```
trungams, line 63: @squirrelsc @LiliDeng Ideally I want to be able to detect this setting earlier and specify it in the `requirement` above. I tried adding a function to lisa/features/security_profile.py as below:

```python3
DiskEncryptionEnabled = partial(
    SecurityProfileSettings,
    encrypt_disk=True
)
```

and then use it similarly to how `CvmEnabled` is used. But it doesn't skip this test case if `encrypt_disk` is set to false in the runbook. Rather, it turns on that value for the running VM. That's why I had to resort to VM runtime check like this. Do you know how I can implement that kind of check to add to `simple_requirement`?

lisa/operating_system.py
trungams: Done. Also refactored some related functions a bit. I'll need some help to validate that the change doesn't break any existing logic.

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,163 @@
...
+
+@TestSuiteMetadata(
+    area="cvm",
+    category="functional",
+    description="""This test suite covers some common scenarios related to
+    CVM boot on Azure.
+    """,
+)
+class CVMBootTestSuite(TestSuite):
+    def before_case(self, log: Logger, **kwargs: Any) -> None:
```
squirrelsc, line 40: `log` is not used, it doesn't need.

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: The test logic of smoke_test is different with most cases, because it doesn't need SSH port 22 working. The cvm test case must have the port 22, so it can follow other cases on handling reboot.

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: Because the steps writes "on **first** boot". If some test coverage needs a fresh new environment, it needs a flag to make sure the test case runs first on a new deployed environment.

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: For the `if` logic, only one path (`if` or `else`) can be covered each time. From your logic, both `if/else` has assertion to validate something. Only one of them can be covered in a test pass, so another path is failed to cover.

lisa/util/__init__.py
```diff
@@ -418,6 +418,17 @@ def enable(self) -> None:
         self._switch(True)
 
 
+class LisaVersionInfo(VersionInfo):
+    def __init__(self, raw_version: str, *args: Any, **kwargs: Any):
```
squirrelsc: `raw_version` -> `version_str`

lisa/util/__init__.py
```diff
@@ -418,6 +418,17 @@ def enable(self) -> None:
         self._switch(True)
 
 
+class LisaVersionInfo(VersionInfo):
```
squirrelsc, line 421: Please either split the version change into a separate commit or PR. If you want to split by commits, make all commits meaningful in the PR.

lisa/operating_system.py
```diff
@@ -549,12 +550,22 @@ def _initialize_package_installation(self) -> None:
...
         pass
 
-    def _get_package_information(self, package_name: str) -> VersionInfo:
+    def _get_package_information(self, package_name: str) -> LisaVersionInfo:
         raise NotImplementedError()
 
-    def _get_version_info_from_named_regex_match(
-        self, package_name: str, named_matches: Match[str]
-    ) -> VersionInfo:
+    def _get_version_info_from_regex(
```
squirrelsc, line 556: What's the new method is for? The `package_name` should be the same meaning for `raw_version`, and the `package_name` is not used.

lisa/operating_system.py
```diff
@@ -765,14 +776,6 @@ class Debian(Linux):
         r"\s+(?P<metadata>.*)\s*"
     )
 
-    """ Package: dpdk
```
squirrelsc: Why remove this logic?

lisa/operating_system.py
```diff
@@ -1487,7 +1475,6 @@ class RPMDistro(Linux):
 
     # ex: dpdk-20.11-3.el8.x86_64 or dpdk-18.11.8-1.el7_8.x86_64
     _rpm_version_splitter_regex = re.compile(
-        r"(?P<package_name>[a-zA-Z0-9\-_]+)-"
```
squirrelsc: It may bring regressions.

lisa/operating_system.py
```diff
@@ -549,12 +550,22 @@ def _initialize_package_installation(self) -> None:
...
 
-    def _get_version_info_from_named_regex_match(
-        self, package_name: str, named_matches: Match[str]
-    ) -> VersionInfo:
+    def _get_version_info_from_regex(
+        self, package_name: str, raw_version: str, regex: Pattern[str]
+    ) -> LisaVersionInfo:
+        matches = regex.search(raw_version)
+        if not matches:
+            self._node.log.warning("")
```
squirrelsc: Don't use `warning` log in LISA, it's not recommended.

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: I didn't think about this scenario. When is it needed? If a vm size supports CVM, and the test case need CVM enabled, it shouldn't be controlled by runbook. If you need to verify the CVM image on a non-CVM scenario, it needs another test case.

microsoft/testsuites/cvm/cvm_boot.py
trungams: pylint will complain without that parameter, so I just keep it there like other places

lisa/operating_system.py
trungams: Just trying to extract similar code to a separate function, each os' `_get_package_information` was using similar logic, with only the regex being different

lisa/operating_system.py
trungams: The new rpm query being used will no longer return package name in the result, but I agree that refactor can get a bit risky. Do you think it's possible to trigger test cases that are affected by this change to catch errors?

microsoft/testsuites/cvm/cvm_boot.py
trungams: I mean CVM can be deployed with different os disk encryption settings:
- VMGuestStateOnly: root partition is not encrypted, this equals encrypt_disk: false in runbook and is the default setting if not specified.
- DiskWithVMGuestState: root partition is encrypted, this equals encrypt_disk: true in runbook.

In my test, I want to skip this test case if encrypt_disk is false. I tried to define a new function `DiskEncryptionEnabled` as described above and pass it to `simple_requirement.supported_features` but it didn't work as I expected. Rather, it just turns on disk encryption for the deployed VM even though it has been explicitly set to false in my runbook.

microsoft/testsuites/cvm/cvm_boot.py
trungams: I see. We plan to run this test case against an Azure Linux CVM image on a per-release basis, so it's expected that we only hit one branch for each test run. From our side that is acceptable.

squirrelsc: Do you know which branch supposedly hit? If so, the `if` block can be removed to get more determinate results.

trungams: I mean it will vary for each release. If we bump versions of any boot component package for a new release, the `boot_component_changed = True` case will run, and I expect this case to be run more frequently. The assertion in the other case is more like a sanity check and is _**very**_ unlikely to fail. I can remove that but I'd prefer to leave it there for completeness

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: I'm not sure fully understand it. Please describe in below format, don't mix with other settings.

* What's the settings
* Expected behavior
* Actual behavior
* Why you think the expected behavior is right

microsoft/testsuites/cvm/cvm_boot.py
trungams: I see. It looks like a new feature to add to Lisa. I was only thinking of controlling it inside a runbook via `keep_environment = false`. Do you have any suggestions on how I should approach it?

lisa/operating_system.py
trungams: With the new dpkg query being used, we can get the version string directly, so this regex is no longer needed to extract the version string

microsoft/testsuites/cvm/cvm_boot.py
trungams: The scenario: I want to skip a test against CVM image if the CVM is not deployed with disk encryption enabled.

In [lisa/features/security_profile.py](https://github.com/microsoft/lisa/blob/main/lisa/features/security_profile.py). I add the following function: 
```python3
DiskEncryptionEnabled = partial(
    SecurityProfileSettings,
    encrypt_disk=True
)
```

- What's the settings:
  - Test case: `requirement=simple_requirement(supported_features[CvmEnabled(), DiskEncryptionEnabled()])`
  - Runbook: `encrypt_disk: false`
- Expected behaviour: Since `encrypt_disk` is set to false in my runbook. I expect the CVM to be deployed without disk encryption, and the test case will then be skipped.
- Actual behaviour: The CVM was deployed **with** disk encryption, and the test case still ran.
- Why you think the expected behaviour is right: That's my understanding of `requirement` in Lisa - it should not alter a VM's deployment settings. A VM deployment settings should match what is specified in the runbook (or take default values if nothing is specified)

lisa/util/__init__.py
trungams: Updated the commits

microsoft/testsuites/cvm/cvm_boot.py
LiliDeng: I think it should be this one 

https://github.com/microsoft/lisa/blob/main/microsoft/testsuites/core/azure_image_standard.py#L887

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: I see. Please fix it in another PR, because this one is already big enough.

It caused by the encrypt_disk is a bool, not a SetSpace. And the current logic is to set it to true, if true and false meet.

* The definition should be SetSpace([True, False]). https://github.com/trungams/lisa/blob/83004c9fc5dcd09cafaa7a343c714dca13927e06/lisa/features/security_profile.py#L61
* The logic should be merge SetSpace. After that, this test case should be skipped, because it selects nothing. https://github.com/trungams/lisa/blob/83004c9fc5dcd09cafaa7a343c714dca13927e06/lisa/features/security_profile.py#L80

lisa/operating_system.py
squirrelsc: This it needs to be compatible with different distro versions from AzureLinux 2.0 to other RPM based Linux. So it needs to keep for old versions.

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+
+        # - PCR4 should change if any of the boot components is upgraded
+        # - PCR7 may change if a signing cert is changed
+        if boot_component_changed:
+            assert_that(
+                pcrs_after_reboot[4],
+                "PCR4 value is still the same even though a boot component changed",
+            ).is_not_equal_to(pcrs_before_reboot[4])
+            assert_warn(
```
squirrelsc: Why is it a warning? The warning means it's not important and should be an info level bug. If it's important, make the test case fail. The middle area is not a good practice.

lisa/tools/tpm2.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, CBLMariner):
+            self.node.os.install_packages("tpm2-tools")
+        else:
+            raise LisaException(
```
squirrelsc: Raise unsupported distro exception.

lisa/tools/tpm2.py
```diff
@@ -0,0 +1,68 @@
...
+    @property
+    def command(self) -> str:
+        return "tpm2"
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, CBLMariner):
```
squirrelsc, line 35: Does it include Mariner 2.0 also? If not, check the versions too.

lisa/tools/tpm2.py
```diff
@@ -0,0 +1,68 @@
...
+            self.node.os.install_packages("tpm2-tools")
+        else:
+            raise LisaException(
+                f"tool {self.command} can't be installed in distro {self.node.os.name}."
+            )
+        return self._check_exists()
+
+    def pcrread(
+        self, pcrs: Optional[Union[int, Sequence[int]]] = None
```
squirrelsc, line 45: Add comments above what's pcrs, and what's the expected types.

lisa/tools/tpm2.py
```diff
@@ -0,0 +1,68 @@
...
+            pcrs_arg = "all"
+        else:
+            pcrs_arg = ",".join(map(str, pcrs))
+        cmd = f"pcrread {alg}:{pcrs_arg}"
+        cmd_result = self.run(
+            cmd,
+            expected_exit_code=0,
+            expected_exit_code_failure_message="failed to read PCR values",
+            shell=True,
+            sudo=True,
```
squirrelsc, line 70: If sudo is not necessary, don't use sudo by default.

lisa/tools/tpm2.py
```diff
@@ -0,0 +1,68 @@
...
+        output = cmd_result.stdout
+
+        lines = [line.strip() for line in output.splitlines()]
+        # first line of output will have the format "<alg-name>:", e.g "sha256:"
+        assert (
+            lines[0][:-1] == alg
+        ), "pcrread output does not contain the requested algorithm"
+
+        result = dict()
+        for line in lines[1:]:
```
squirrelsc: Use regular expressions to extra values, it's easier to read and maintain. Also give examples above.

lisa/tools/bootctl.py
```diff
@@ -0,0 +1,45 @@
...
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    def _install(self) -> bool:
+        if isinstance(self.node.os, CBLMariner):
+            self.node.os.install_packages("systemd-udev")
+        else:
+            raise LisaException(
```
squirrelsc: Raise unsupported distro exception.

lisa/tools/tpm2.py
trungams: We need sudo to interact with the TPM device

lisa/tools/tpm2.py
trungams: This tool is available for Mariner 2.0

microsoft/testsuites/cvm/cvm_boot.py
trungams: updated to only use `log.info`

lisa/tools/tpm2.py
trungams: I added a docstring to give an example input/output for this command. This is one of the cases where I think using a regex would be more complicated than just using python's string manipulation

lisa/operating_system.py
trungams: I will keep the regex patterns unmodified for this PR and do the refactor in a separate PR. Since the semver regex (*_version_splitter_regex) were the reason the parsing logic failed in the first place, so I still think they need to be updated. The dpkg and rpm queries I introduced should be consistent across distros so they are safe to use.

microsoft/testsuites/cvm/cvm_boot.py
trungams: Thank you for the explanation. I will address it after this PR

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: Please change it to assert_that or info level log.

lisa/tools/tpm2.py
squirrelsc: Regexp improves readability of the code. The doc string is not close enough to the parsing logic, so people need to spend extra time to read and analyze the code. If it's examples/Regexp like other places in LISA, it's more straightforward.

microsoft/testsuites/cvm/cvm_boot.py
trungams: I forgot to remove old code. Corrected

lisa/tools/tpm2.py
trungams: Added regex

lisa/tools/tpm2.py
```diff
@@ -0,0 +1,93 @@
...
+
+from lisa.executable import Tool
+from lisa.operating_system import CBLMariner
+from lisa.util import UnsupportedDistroException
+
+
+class Tpm2(Tool):
+    """tpm2 provides a toolset based on tpm2-tss to interact with TPM 2.0 devices"""
+
+    _pcrread_split_hash_and_pcrs_regex = re.compile(
```
squirrelsc, line 19: Add examples above regexps

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3626, opened 2025-02-03
[Hyper-v platform] Use data disk as LISA working directory on Hyper-v host

SRIKKANTH:
The default lisa working path on Hyper-v hosts is under 'C:/' which is not ideal for large data.
This change initializes an attached disk, creates a partition, formats it, and mounts it to the lisa working directory.
This change expects an un-initialized (RAW) disk to be available on Hyper-V.

### DIFF
--- a/lisa/tools/hyperv.py
+++ b/lisa/tools/hyperv.py
@@ -445,6 +445,71 @@ class HyperV(Tool):
         # Restart the DHCP server to apply the changes
         service.restart_service("dhcpserver")
 
+    # The default lisa working path is under 'C:/' which is not ideal for large data.
+    # This method initializes an attached disk, creates a partition, formats it,
+    # and mounts it to the lisa working directory
+    # This method expects an un-initialized (RAW) disk to be available on Hyper-V.
+    def configure_lisa_working_dir(self) -> None:
+        powershell = self.node.tools[PowerShell]
+
+        # Get the disk that is not initialized
+        disk_number = powershell.run_cmdlet(
+            "(Get-Disk | Where-Object PartitionStyle -Eq 'RAW').Number",
+            force_run=True,
+        )
+        # If no un-initialized disk is found, log a warning and return
+        if not disk_number:
+            self._log.warning("No un-initialized disk found.")
+            self._log.warning("Skipping lisa working dir setup on additional disk.")
+            return
+        # Initialize the disk
+        powershell.run_cmdlet(
+            f"Initialize-Disk -Number {disk_number}",
+            force_run=True,
+        )
+        # Create a new partition
+        powershell.run_cmdlet(
+            f"New-Partition -DiskNumber {disk_number}"
+            " -UseMaximumSize -AssignDriveLetter",
+            force_run=True,
+        )
+        # Get the drive letter
+        drive_letter = powershell.run_cmdlet(
+            f"(Get-Partition -DiskNumber {disk_number}"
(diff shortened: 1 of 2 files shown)


lisa/transformers/hyperv_preparation.py
```diff
@@ -33,6 +33,9 @@ def _internal_run(self) -> Dict[str, Any]:
         # Enable Hyper-V
         hv = self._node.tools[HyperV]
 
+        # Configure lisa working path to use free disk.
+        hv.use_raw_disk_for_lisa_working_dir()
```
squirrelsc, line 37: It shouldn't be the default way. What happened, if it's running on Azure?

lisa/tools/hyperv.py
```diff
@@ -445,6 +445,71 @@ def enable_internal_dhcp(self, dhcp_scope_name: str = "DHCPInternalNAT") -> None
         # Restart the DHCP server to apply the changes
         service.restart_service("dhcpserver")
 
+    # The default lisa working path is under 'C:/' which is not ideal for large data.
```
squirrelsc, line 448: Use a configurable path, instead of use the LISA working path. The big files should be tracked separately.

lisa/transformers/hyperv_preparation.py
SRIKKANTH: This is written and tested on azure. My use case is also Azure only

squirrelsc: I see. Please make it optional. Only if give a path in config, then create the data disk.

SRIKKANTH: Also use_raw_disk_for_lisa_working_dir() will skip if there is not raw disk available on the Host.

lisa/tools/hyperv.py
SRIKKANTH: These big files are vhds used as OS and data disks. They are stored by default under relevant test log folder. Its easier if we leave them that way so that its easier to map them to the test runs and clean them up once the tests are finished.

squirrelsc: You can use LISA working path as the default path, but it should be configurable for lab environments.

lisa/transformers/hyperv_preparation.py
squirrelsc: It's not flexible enough. Please create configurations in the runbook. For example,
```
data_disk: # a list.
    - source: # new/exist/raw
      path: # for new or exist. It must be a file name for exist. It can be file name or path for new. If it ends with vhd, it means a file name, otherwise a path. If it's not set, use lisa working path for new.
```

lisa/tools/hyperv.py
```diff
@@ -445,6 +445,76 @@ def enable_internal_dhcp(self, dhcp_scope_name: str = "DHCPInternalNAT") -> None
...
+
+        # Get the first disk that is not initialized
+        raw_disks = powershell.run_cmdlet(
+            "Get-Disk | Where-Object PartitionStyle -Eq 'RAW'",
+            force_run=True,
+            output_json=True,
+        )
+        # If no un-initialized disk is found, log a warning and return
+        if not raw_disks:
+            self._log.warning("No un-initialized disk found.")
```
squirrelsc, line 463: 1. Don't print warning, either print info or raise exception.
2. The following log line should be merged into one line, so people always know they are logged together.

PR #3628, opened 2025-02-04
Added log support for ready_checker of Pxe cluster

paulli2017:
Added log support for ready_checker of Pxe cluster

### DIFF
--- a/lisa/sut_orchestrator/baremetal/readychecker.py
+++ b/lisa/sut_orchestrator/baremetal/readychecker.py
@@ -3,7 +3,7 @@
 
 import os
 from dataclasses import dataclass
-from typing import Type, cast
+from typing import Any, Type, cast
 
 import requests
 from dataclasses_json import dataclass_json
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/baremetal/cluster/pxe.py
```diff
@@ -38,6 +38,12 @@ def write(self, data: str) -> None:
         self._process.input(f"{data}\n")
 
     def _get_console_log(self, saved_path: Optional[Path]) -> bytes:
+        self._process.wait_output(
```
squirrelsc: Add comments to explain the wait `""` logic.

lisa/sut_orchestrator/baremetal/readychecker.py
```diff
@@ -3,7 +3,8 @@
 
 import os
 from dataclasses import dataclass
-from typing import Type, cast
+from logging import Logger
```
squirrelsc: Use two commits for the two different changes. So each commit has a simple purpose.

lisa/sut_orchestrator/baremetal/cluster/pxe.py
paulli2017: Change will be in another PR. thanks for your suggestions.

lisa/sut_orchestrator/baremetal/readychecker.py
paulli2017: The change is done, thanks

squirrelsc: It can be two "commits" in the same PR, not two "PRs". But both ways are ok.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3637, opened 2025-02-07
Creating a plugin for DMesg log analysis post TC Execution

umfranci:
- The purpose of this Draft PR is to create a code for checking the DMesg logs for any exceptions once the test case execution is completed.
- The execution will be done based on a switch - the default value of which would be false - hence avoiding any unnecessary failures/noise.
- This can be later on extended to also check panic patterns or increase the scope of the checks executed after the test case is completed.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -24,7 +24,7 @@ from lisa.feature import Features
 from lisa.nic import Nics, NicsBSD
 from lisa.operating_system import OperatingSystem
 from lisa.secret import add_secret
-from lisa.tools import Chmod, Df, Echo, Lsblk, Mkfs, Mount, Reboot, Uname, Wsl
+from lisa.tools import Chmod, Df, Dmesg, Echo, Lsblk, Mkfs, Mount, Reboot, Uname, Wsl
 from lisa.tools.mkfs import FileSystem
 from lisa.util import (
     ContextMixin,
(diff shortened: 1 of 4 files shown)


lisa/node.py
```diff
@@ -1186,6 +1192,9 @@ class NodeHookSpec:
     def get_node_information(self, node: Node) -> Dict[str, str]:
         ...
 
+    @hookspec
+    def check_dmesg_oops(self, node: Node) -> bool:
```
squirrelsc, line 1196: eturn the found error logs, so the errors can be in test case error message directly.

lisa/node.py
```diff
@@ -501,6 +502,11 @@ def get_information(self) -> Dict[str, str]:
 
         return final_information
 
+    def exec_check_dmesg_oops(self) -> bool:
+        return plugin_manager.hook.check_dmesg_oops(
```
squirrelsc, line 506: 1. Group errors from different plugins, if it returns non-empty.
2. Raise failure exception and let the upper layer to deal with it.

lisa/node.py
```diff
@@ -1186,6 +1192,9 @@ class NodeHookSpec:
     def get_node_information(self, node: Node) -> Dict[str, str]:
         ...
 
+    @hookspec
```
squirrelsc, line 1195: If it's only hooked in LISA upstream code, it doesn't need to use the hookspec. Just implement as a normal method.

lisa/testsuite.py
```diff
@@ -752,6 +752,16 @@ def start(
                     log=case_log,
                 )
 
+            nodes = environment.nodes
```
squirrelsc, line 755: Move into `__run_case` after test passed, and let the `handle_exception` proceed the raised exception.

lisa/schema.py
```diff
@@ -1444,6 +1444,8 @@ class Platform(TypedSchema, ExtendableSchemaMixin):
     capture_azure_information: bool = False
     # capture boot time info or not
     capture_boot_time: bool = False
+    # to check if dmesg logs need to be analyzed after each test case
+    check_dmesg_after_case: bool = False
```
squirrelsc, line 1448: assert_kernel_error

PR #3641, opened 2025-02-12
add_azure_core_repo: use bionic when the repo doesn't exist for newer distro version

LiliDeng:


lisa/operating_system.py
```diff
@@ -1345,6 +1345,10 @@ def add_azure_core_repo(
         if not code_name:
             code_name = self.information.codename
         repo_url = "http://packages.microsoft.com/repos/azurecore/"
+        # temporary workaround for the issue that azurecore repo is not available
+        # for ubuntu 2404 and 2410
+        if code_name in ["noble", "oracular"]:
```
squirrelsc: It should be maintained as an allowed list, so it doesn't need to add every half an year. like `if code_name not in ["bionic", ...]:`

squirrelsc (approved):

PR #3642, opened 2025-02-12
fix nested test cases issues

LiliDeng:


lisa/tools/qemu.py
```diff
@@ -121,13 +121,19 @@ def create_vm(
...
+            )
 
         # kill any existing qemu process if stop_existing_vm is True
         if stop_existing_vm:
             self.delete_vm()
 
         cmd = self._configure_qemu_command_for_cpu(cmd)
 
+        # allow qemu to use 64 GB memory
+        cmd += "-m 64000 "
```
squirrelsc: Is it another fix or fix the out of space issue? If the host doesn't have enough memory, will it cause errors?

LiliDeng: I will raise another PR for it.

squirrelsc (approved):

PR #3643, opened 2025-02-12
SharedImageGalleryTransformer: Take marketplace features

kamalca:
If a SIG is created from a marketplace image, you can provide the source image instead of specifying all of the features, hyper-v gen, and architecture individually.

### DIFF
--- a/lisa/sut_orchestrator/azure/transformers.py
+++ b/lisa/sut_orchestrator/azure/transformers.py
@@ -27,6 +27,7 @@ from lisa.util import (
 
 from .common import (
     AZURE_SHARED_RG_NAME,
+    AzureNodeSchema,
     check_blob_exist,
     check_or_create_gallery,
     check_or_create_gallery_image,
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/transformers.py
```diff
@@ -643,7 +660,7 @@ def _internal_run(self) -> Dict[str, Any]:
             runbook.gallery_image_osstate,
             runbook.gallery_image_hyperv_generation,
             runbook.gallery_image_architecture,
-            runbook.gallery_image_securitytype,
+            features,
```
squirrelsc, line 663: The property bag style parameters are not validated by schema, it's hard to find out supported values.

kamalca: It is not passed in through the schema, it is collected automatically from the marketplace image. I believe all feature values for a marketplace image are supported for a SIG image.

PR #3644, opened 2025-02-13
Update to Network Channel count calculation in test case 'verify_vmbus_devices_channels'

umfranci:
At present the network channel count calculation in the test case 'verify_vmbus_devices_channels' is based on the number of vCPUs and has a cap value of 8 - which might get updated, like in the following patch: [[PATCH v4] net: netvsc: Update default VMBus channels](https://lore.kernel.org/linux-hyperv/1724735791-22815-1-git-send-email-ernis@linux.microsoft.com/)

This patch is currently available in Upstream but not all downstream distros. Hence it is better to get the expected count using ethtool which would be dependent on the files updated in the patch above and help avoid discrepancies while testing different scenarios.

### DIFF
--- a/microsoft/testsuites/core/lsvmbus.py
+++ b/microsoft/testsuites/core/lsvmbus.py
@@ -149,7 +149,7 @@ class LsVmBus(TestSuite):
             if vmbus_device.name == "Synthetic SCSI Controller":
                 assert_that(vmbus_device.channel_vp_map).is_length(
                     expected_scsi_channel_count
-                )        
+                )
 
     @TestCaseMetadata(
         description="""
--- a/microsoft/testsuites/core/lsvmbus.py
+++ b/microsoft/testsuites/core/lsvmbus.py
@@ -107,7 +107,7 @@ class LsVmBus(TestSuite):
             channels created and associated.
             2.1 Check expected channel count of each netvsc matches that obtained from the ethtool.
                 2.1.1 Get expected channel count of netvsc from ethtool.
-                https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/tree/drivers/net/hyperv/netvsc_drv.c#n2001
+                https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/tree/drivers/net/hyperv/netvsc_drv.c#n2001 # noqa: E501
             2.2 Check expected channel count of each storvsc SCSI device is min (num of
                  vcpu/4, 64).
                 2.2.1 Caculate channel count of each storvsc SCSI device.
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/core/lsvmbus.py
```diff
@@ -134,11 +130,13 @@ def verify_vmbus_devices_channels(self, node: Node) -> None:
...
         core_count = lscpu_tool.get_core_count()
-        # Each netvsc device should have "the_number_of_vCPUs" channel(s)
-        #  with a cap value of 8.
-        expected_network_channel_count = min(core_count, 8)
-        # Each storvsc SCSI device should have "the_number_of_vCPUs / 4" channel(s)
-        #  with a cap value of 64.
+
+        # 2.1 Get expected channel count of each netvsc is min (num of vcpu, 8).
+        expected_network_channel_count = (
+            node.tools[Ethtool].get_device_channels_info("eth0", True)
```
LiliDeng, line 136: it is possible that case will change the combined channel count

LiliDeng: I think it is the actual value, not the expected value.

squirrelsc: > I think it is the actual value, not the expected value.

Agree, the value should be calculated, instead of getting from the system. This test case is to verify if the count of channel is the same as designed.

umfranci: Thanks for the review team. If the logic for channel calculation is updated and the said patch is available only in certain distros - it would cause the test to fail in certain cases and to succeed in others, thus creating a discrepancy. The current patch is available in SLES and Upstream but not in other distros. How can we alter the test cases to handle different logics for different situations?
Also, to avoid such a situation, wouldn't it be helpful to use a tool which refers the relevant channel calculation code at the Kernel level?

squirrelsc: Can it be differentiated by kernel version?

umfranci: We can do that for a short-term solution but as the patch is backported, we would need to update the kernel version for the test case.

squirrelsc: > We can do that for a short-term solution but as the patch is backported, we would need to update the kernel version for the test case.

If the patch is backported, all newer version can use the new algorithm. So, the test cases can work with both old and newer versions.

umfranci: Thanks for the inputs @squirrelsc . Started an internal thread on this to get inputs from the Kernel team as well.

PR #3646, opened 2025-02-15
fix xdp test issues

LiliDeng:


microsoft/testsuites/xdp/common.py
```diff
@@ -80,7 +80,11 @@ def set_hugepage(node: Node) -> None:
         huge_page_disks = _huge_page_disks.items()
     for point, options in huge_page_disks:
         mount.mount(
-            name="nodev", point=point, fs_type=FileSystem.hugetlbfs, options=options
+            name="nodev",
+            point=point,
+            fs_type=FileSystem.hugetlbfs,
+            options=options,
+            ignore_error=True,
```
squirrelsc: If error happens, what's the expected behavior? Add in the code comments.

squirrelsc (dismissed):

PR #3651, opened 2025-02-19
skip tests requiring package/repo mods for slmicro class

katmnsn:
Skipped for now due to transactional update feature: [docs here](https://documentation.suse.com/smart/systems-management/html/SLE-Micro-5.5-admin/index.html#how-transactional-update-works)

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -42,6 +42,8 @@ from lisa.util import (
     MissingPackagesException,
     ReleaseEndOfLifeException,
     RepoNotExistException,
+    SkippedException,
+    UnsupportedDistroException,
     filter_ansi_escape,
     get_matched_str,
     parse_version,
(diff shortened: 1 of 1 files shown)


lisa/operating_system.py
```diff
@@ -2177,10 +2179,75 @@ def name_pattern(cls) -> Pattern[str]:
 
 
 class SlMicro(Suse):
+    # all tests requiring packages/repos skipped due to transactional-update feature
```
squirrelsc: 1. When adding comments on class level, use docstring like below.
2. The comments isn't clear enough to me, if I don't have context on "transactional-update". Maybe we can just mentioned SlMicro repos are not supported in LISA.

```python
"""
 ...
"""
```

lisa/operating_system.py
```diff
@@ -2177,10 +2179,75 @@ def name_pattern(cls) -> Pattern[str]:
...
     @classmethod
     def name_pattern(cls) -> Pattern[str]:
         return re.compile(r"^sl-micro$", re.IGNORECASE)
 
+    def add_azure_core_repo(
+        self, repo_name: Optional[AzureCoreRepo] = None, code_name: Optional[str] = None
+    ) -> None:
+        raise SkippedException(UnsupportedDistroException(self._node.os))
```
squirrelsc: For all skipped exception, add one more message to explain the reason, why SlMicro test cases are skipped as unsupported os.

lisa/operating_system.py
```diff
@@ -2179,15 +2179,20 @@ def name_pattern(cls) -> Pattern[str]:
...
     def name_pattern(cls) -> Pattern[str]:
         return re.compile(r"^sl-micro$", re.IGNORECASE)
 
     def add_azure_core_repo(
         self, repo_name: Optional[AzureCoreRepo] = None, code_name: Optional[str] = None
     ) -> None:
-        raise SkippedException(UnsupportedDistroException(self._node.os))
+        raise SkippedException(
+            UnsupportedDistroException(self._node.os),
+            f"Installing the azure core repo not supported for {self._node.os}"
```
squirrelsc: "Not implemented" is clearer than "not supported". The os information should be printed by unsupported distro exception already, so you can just bring additional information. Please check the final error messages, try best to make them clear and simple.

katmnsn: Here's how they look now: 
 
 XdpFunctional.verify_xdp_basic: SKIPPED  skipped: Unsupported system: 'SUSE Linux Micro 6.1'. Getting package list for this distro not implemented in LISA.

squirrelsc (dismissed):

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3674, opened 2025-02-20
Checking DMesg logs for Panics/Errors post test case execution

umfranci:
- The purpose of this PR is to add code for checking the DMesg logs for any panics/exceptions once the test case execution is completed.
- The execution will be done based on a switch (_assert_kernel_error_) - the default value of which would be false - hence avoiding any unnecessary failures/noise.
- Thus, we aim to check both kernel errors and panics and fail the test in case either is found.

### DIFF
--- a/lisa/node.py
+++ b/lisa/node.py
@@ -24,7 +24,7 @@ from lisa.feature import Features
 from lisa.nic import Nics, NicsBSD
 from lisa.operating_system import OperatingSystem
 from lisa.secret import add_secret
-from lisa.tools import Chmod, Df, Echo, Lsblk, Mkfs, Mount, Reboot, Uname, Wsl
+from lisa.tools import Chmod, Df, Dmesg, Echo, Lsblk, Mkfs, Mount, Reboot, Uname, Wsl
 from lisa.tools.mkfs import FileSystem
 from lisa.util import (
     ContextMixin,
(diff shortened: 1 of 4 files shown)


lisa/schema.py
```diff
@@ -1444,6 +1444,8 @@ class Platform(TypedSchema, ExtendableSchemaMixin):
     capture_azure_information: bool = False
     # capture boot time info or not
     capture_boot_time: bool = False
+    # to check if dmesg logs need to be analyzed after each test case
+    assert_kernel_error: bool = False
```
squirrelsc: `assert_kernel_error_after_test`

lisa/testsuite.py
```diff
@@ -857,6 +857,11 @@ def __run_case(
                 test_kwargs=test_kwargs,
             )
             case_result.set_status(TestStatus.PASSED, "")
+            if case_result.environment is not None:
```
squirrelsc, line 859: It should be above the set_status to passed.

lisa/node.py
```diff
@@ -501,6 +502,11 @@ def get_information(self) -> Dict[str, str]:
 
         return final_information
 
+    def exec_check_dmesg_oops(self) -> None:
```
squirrelsc: `check_kernel_error`

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3677, opened 2025-02-21
Fix regression for query command changes in get_package_information and remove unused variables

LiliDeng:
querying package command changed, the return value missed the package info, now the version can't be got since the pattern mismatch. So the regex needs to be updated, removing package from the pattern can fix the issue.

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -780,14 +780,6 @@ class Debian(Linux):
         r"\s+(?P<metadata>.*)\s*"
     )
 
-    """ Package: dpdk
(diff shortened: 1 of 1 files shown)


squirrelsc:
What's the regression, how it's fixed?

LiliDeng:
> What's the regression, how it's fixed?
querying package command changed, the return value missed the package info, now the version can't be got since the pattern mismatch. So the regex needs to be updated, removing package from the pattern can fix the issue.

squirrelsc (approved):

Issue #3681, opened 2025-02-24
Support for using combinators in Lisa


Ankita13-code:
For the testing of Azure Linux CVM images, currently we have 6 CVM test cases in LISA – 
-	verify_isolation_config
-	verify_lsvmbus
-	verify_azure_cvm_attestation_report (currently implemented only for SEV-SNP)
-	verify_encrypted_root_partition
-	verify_boot_success_after_component_upgrade

We want to execute all these tests with the following combinations - 
-	Intel TDX - with disk encryption
-	Intel TDX- without disk encryption
-	AMD SEV-SNP - with disk encryption
-	AMD SNP - without disk encryption

For doing so we are using a batch combinator in our runbook. However, at the end of test runs, the JUnit file produced does not honor the combinator effect. Therefore, we see only one combination result in the Junit file for each unique test case name while the others are ignored. 

How can we use the batch combinator concept for running our CVM tests and for the JUnit file to reflect all the possible combinations?

squirrelsc:
Thanks for reporting. Please check if #3682 fixed it.

trungams:
@squirrelsc I can see multiple results for the same test case now (example below)

```xml
<testsuite name="AzureCVMAttestationTestSuite" timestamp="2025-02-24T21:10:24" tests="2" failures="0" errors="0">
    <testcase name="verify_azure_cvm_attestation_report (lisa_0_1)" classname="AzureCVMAttestationTestSuite" time="405.972" />
    <testcase name="verify_azure_cvm_attestation_report (lisa_1_1)" classname="AzureCVMAttestationTestSuite" time="430.000" />
  </testsuite>
```

Is there any way to tell which configs have been used for each test id from the junit file? e.g `lisa_0_*` = SEV-SNP, `lisa_1_*` = TDX. Do the IDs correspond to the order the settings are specified in a runbook if we use batch combinator?

squirrelsc:
You can search the id in the log, the raw test results should have enough information. The junit results is for "unit test", the information is very limited.

Ankita13-code:
Thanks @squirrelsc. This helps us get all the results in the junit file.

squirrelsc:
Thanks for confirming, I will close this issue.

PR #3682, opened 2025-02-24
junit: add unique id in names.

squirrelsc:
In some junit file consumers, the test name is
 considered as the key. If it's duplicated, it
 shows only one. This change use the test case id
 to make test name unique.


squirrelsc:
fix #3681

LiliDeng (approved):

PR #3684, opened 2025-02-26
Swap testcase priority

kamalca:
Prioritize environment reuse over priority by P0-P5. This is important to avoid concurrent environments exceeding concurrency setting in LISA. It resolves issues with cluster capacity and helps with quota management.

### DIFF
--- a/lisa/runners/lisa_runner.py
+++ b/lisa/runners/lisa_runner.py
@@ -116,60 +116,54 @@ class LisaRunner(BaseRunner):
         # Loop environments instead of test results, because it needs to reuse
         # environment as much as possible.
         if available_results and available_environments:
-            for priority in range(6):
-                can_run_results = self._get_results_by_priority(
-                    available_results, priority
-                )
-                if not can_run_results:
+            # it means there are test cases and environment, so it needs to
+            # schedule task.
+            for environment in available_environments:
+                if environment.is_in_use:
+                    # skip in used environments
                     continue
 
-                # it means there are test cases and environment, so it needs to
-                # schedule task.
-                for environment in available_environments:
-                    if environment.is_in_use:
-                        # skip in used environments
+                # Try to pick the designated test result from the current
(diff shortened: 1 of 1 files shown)


lisa/runners/lisa_runner.py
```diff
@@ -116,18 +116,18 @@ def fetch_task(self) -> Optional[Task[None]]:
...
+                    # skip in used environments
                     continue
 
-                # it means there are test cases and environment, so it needs to
-                # schedule task.
-                for environment in available_environments:
-                    if environment.is_in_use:
-                        # skip in used environments
+                for priority in range(6):
+                    can_run_results = self._get_results_by_priority(
```
squirrelsc: We can totally remove the `by_priority` concept, because it's not valuable as a second level loop. After it's removed, the logic is simpler.

kamalca: I think it is still important for priority order to be followed as much as possible, but I believe this is achieved by sorting the results beforehand, correct?

squirrelsc:
@LiliDeng LGTM

lisa/runners/lisa_runner.py
```diff
@@ -116,60 +116,54 @@ def fetch_task(self) -> Optional[Task[None]]:
         # Loop environments instead of test results, because it needs to reuse
         # environment as much as possible.
         if available_results and available_environments:
-            for priority in range(6):
-                can_run_results = self._get_results_by_priority(
```
squirrelsc, line 120: The method _get_results_by_priority can be removed.

lisa/runners/lisa_runner.py
squirrelsc: Yes, the order of results can achieve it too.

squirrelsc:
@LiliDeng LGTM. It maybe hard to validate. Please just keep an eye, any weird behavior or incorrect ending case status (like queued or assigned).

LiliDeng (approved):

PR #3685, opened 2025-02-27
Skip storage testcases if nvme cmd doesn't have 'NameSpace' in output.

SRIKKANTH:
Older versions of nvme-cli do not have the 'NameSpace' field in the output.
This change will make the storage testcases to skip if NameSpace field is not available instead of failing.

### DIFF
--- a/lisa/tools/nvmecli.py
+++ b/lisa/tools/nvmecli.py
@@ -7,7 +7,7 @@ from typing import Any, Dict, List, Optional, Type, cast
 from lisa.executable import Tool
 from lisa.operating_system import Posix
 from lisa.tools import Git, Make
-from lisa.util import find_patterns_in_lines
+from lisa.util import LisaException, find_patterns_in_lines
 from lisa.util.process import ExecutableResult
 
 
@@ -107,7 +107,20 @@ class Nvmecli(Tool):
         return namespaces_cli
 
     def get_devices(self, force_run: bool = False) -> Any:
-        nvme_list = self.run("list -o json", shell=True, sudo=True, force_run=force_run)
+        # get nvme devices information ignoring stderror
+        nvme_list = self.run(
+            "list -o json 2>/dev/null",
+            shell=True,
+            sudo=True,
+            force_run=force_run,
+            no_error_log=True,
+        )
+        # NVMe list command returns empty string when no NVMe devices are found.
+        if not nvme_list.stdout:
+            raise LisaException(
+                "No NVMe devices found. "
+                "The 'nvme list' command returned an empty string."
+            )
         nvme_devices = json.loads(nvme_list.stdout)
         return nvme_devices["Devices"]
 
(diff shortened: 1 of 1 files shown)


lisa/tools/nvmecli.py
```diff
@@ -161,6 +161,13 @@ def get_disks(self, force_run: bool = False) -> List[str]:
 
     def get_namespace_ids(self, force_run: bool = False) -> List[Dict[str, int]]:
         nvme_devices = self.get_devices(force_run=force_run)
+        # Older versions of nvme-cli do not have the NameSpace key in the output
+        # skip the test if NameSpace key is not available
+        if not nvme_devices or "NameSpace" not in nvme_devices[0]:
+            raise SkippedException(
+                "'NameSpace' key is not available in 'nvme -list -o json' output"
```
squirrelsc: The message is not clear enough. It says what's verified, but not explain why the test case is skipped from the test purpose. Why is the test skipped, when the NameSpace is not available? Even the Namespace doesn't exist, it's still a nvme device.

SRIKKANTH: Older images like CentOS don't have latest nvme cli which is missing support for "Namespace" field in the output.

This is needed only for serial hot add storage testcases (with disk controller == NVMe) but not for actual NVMe testcases.
Parallel storage testcases and other tests are not impacted by this.

The "Namespace" field is used to get the LUN numbers of the remote disks which is used by all serial hot add disk tests. Without this the tests will not work on VMs with NVMe disk controller.  

Modified the exception message saying the reason and added a debug message

squirrelsc: Thanks for fixing, it's better, and can be improved too.

1. It needs to raise skipped exceptions in tool level, because it doesn't know how the tools will be used. Some test cases maybe skipped unexpected. Please raise general exception here, and catch/rethrow it to skipped in test cases.
2. The message should tell the details like in the review comments. Some information is important to help understanding why it's skipped like: _The "Namespace" field is used to get the LUN numbers of the remote disks which is used by all serial hot add disk tests. Without this the tests will not work on VMs with NVMe disk controller._

squirrelsc:
@LiliDeng LGTM

SRIKKANTH:
@LiliDeng  please accept this PR.

lisa/tools/nvmecli.py
```diff
@@ -159,8 +173,22 @@ def get_disks(self, force_run: bool = False) -> List[str]:
...
+        self, log: Logger, force_run: bool = False
+    ) -> List[Dict[str, int]]:
         nvme_devices = self.get_devices(force_run=force_run)
+        # Older versions of nvme-cli do not have the NameSpace key in the output
+        # skip the test if NameSpace key is not available
+        if not nvme_devices or "NameSpace" not in nvme_devices[0]:
+            log.debug(
+                "'NameSpace' key is not available in 'nvme -list -o json' "
+                "output to get namespace ids."
+            )
```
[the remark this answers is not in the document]
SRIKKANTH: Fixed

LiliDeng (approved):

PR #3686, opened 2025-02-27
adds another cvm tag

katmnsn:
This adds another way to detect CVM generalized image, created using cvm capture: https://learn.microsoft.com/en-us/azure/confidential-computing/create-confidential-vm-from-compute-gallery#create-a-confidential-vm-type-image-using-confidential-vm-capture


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3688, opened 2025-03-04
add azure linux-specific fips tests

tobiasb-ms:
Add Azure Linux-specific tests around FIPS compliance. Specifically, this change adds:

`verify_fips_is_enabled_azl`: Verifies that an AZL system is in FIPS mode.
`verify_fips_is_disabled_azl`: Verifies that an AZL system is not in FIPS mode.
`verify_fips_enable_azl`: Verifies that an AZL system can be put into FIPS mode.
`verify_fips_disable_azl`: Verifies that an AZL system can be taken out of FIPS mode.

It also removes the AZL code from `verify_fips_enable`.

Due to the nature of these tests, they only make sense to run if we believe a machine is already in the correct mode. For example, `verify_fips_is_enabled_azl` would make sense on the marketplace image `MicrosoftCBLMariner:azure-linux-3:azure-linux-3-fips:latest` but not on `MicrosoftCBLMariner:azure-linux-3:azure-linux-3:latest`.

To accomplish this, we look at two things (see function `ensure_fips_expectations`).
First, we see if a variable `testing_fips_image` was provided to the test case. If so, and if it is `yes` or `no`, we respect that.
Second, we try to hit the azure metadata endpoint to get the image sku name. If that is a FIPS sku, we consider it to be a FIPS machine.
If neither exist, we skip the test.

### DIFF
--- a/lisa/base_tools/sed.py
+++ b/lisa/base_tools/sed.py
@@ -92,6 +92,7 @@ class Sed(Tool):
             sudo=sudo,
             shell=True,
         )
+        result.assert_exit_code(message=result.stdout)
 
 
 class SedBSD(Sed):
--- a/lisa/base_tools/sed.py
+++ b/lisa/base_tools/sed.py
@@ -79,6 +79,33 @@ class Sed(Tool):
         )
         result.assert_exit_code(message=result.stdout)
 
+    def raw(self, expression: str, file: str, sudo: bool = False) -> None:
+        """
+        Apply a raw sed expression to a file, as opposed to separating the expression into a pattern and replacement.
+        This allows for more advanced `sed` usage.
+        Args:
+            expression (str): The sed expression to apply.
+            file (str): The path to the file to modify.
+            sudo (bool, optional): Whether to execute the command with sudo. Defaults to False.
+        Returns:
+            None
+        """
+        self._log.debug(f"Applying raw sed expression '{expression}' to file '{file}' with sudo={sudo}")
+
+        # Escape special characters in the expression.
+        expression = expression.replace('"', r"\"").replace("$", r"\$")
+        self._log.debug(f"Escaped expression: '{expression}'")
+
+        cmd = f'-i.bak "{expression}" {file}'
+        result = self.run(
+            cmd,
+            force_run=True,
+            no_error_log=True,
+            no_info_log=True,
+            sudo=sudo,
+            shell=True,
+        )
+        result.assert_exit_code(message=result.stdout)
 
 class SedBSD(Sed):
     @property
--- a/lisa/base_tools/sed.py
+++ b/lisa/base_tools/sed.py
@@ -79,19 +79,10 @@ class Sed(Tool):
         )
         result.assert_exit_code(message=result.stdout)
 
-    def raw(self, expression: str, file: str, sudo: bool = False) -> None:
-        """
-        Apply a raw sed expression to a file, as opposed to separating the expression into a pattern and replacement.
-        This allows for more advanced `sed` usage.
-        Args:
-            expression (str): The sed expression to apply.
-            file (str): The path to the file to modify.
-            sudo (bool, optional): Whether to execute the command with sudo. Defaults to False.
-        Returns:
-            None
-        """
-        # Escape special characters in the expression.
-        expression = expression.replace('"', r"\"").replace("$", r"\$")
+    def delete_line_substring(
+        self, match_line: str, regex_to_delete: str, file: PurePath, sudo: bool = False
+    ) -> None:
+        expression = f"/{match_line}/s/{regex_to_delete}//g"
         cmd = f'-i.bak "{expression}" {file}'
         result = self.run(
             cmd,
@@ -101,7 +92,7 @@ class Sed(Tool):
             sudo=sudo,
             shell=True,
         )
-        result.assert_exit_code(message=result.stdout)
+
 
 class SedBSD(Sed):
     @property
--- a/lisa/base_tools/sed.py
+++ b/lisa/base_tools/sed.py
@@ -90,12 +90,8 @@ class Sed(Tool):
         Returns:
             None
         """
-        self._log.debug(f"Applying raw sed expression '{expression}' to file '{file}' with sudo={sudo}")
-
         # Escape special characters in the expression.
         expression = expression.replace('"', r"\"").replace("$", r"\$")
-        self._log.debug(f"Escaped expression: '{expression}'")
-
         cmd = f'-i.bak "{expression}" {file}'
         result = self.run(
             cmd,
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -42,6 +42,17 @@ class Disk(Feature):
         else:
             return None
 
+    def get_partition_with_mount_point_or_raise(
+        self, mount_point: str
+    ) -> PartitionInfo:
+        """
+        Get partition with mount point or raise an exception if not found.
+        """
+        partition = self.get_partition_with_mount_point(mount_point)
+        if partition is None:
+            raise LisaException(f"Partition with mount point {mount_point} not found.")
+        return partition
+
     def check_resource_disk_mounted(self) -> bool:
         return False
 
--- a/lisa/features/disks.py
+++ b/lisa/features/disks.py
@@ -42,17 +42,6 @@ class Disk(Feature):
         else:
             return None
 
-    def get_partition_with_mount_point_or_raise(
-        self, mount_point: str
-    ) -> PartitionInfo:
-        """
-        Get partition with mount point or raise an exception if not found.
-        """
-        partition = self.get_partition_with_mount_point(mount_point)
-        if partition is None:
-            raise LisaException(f"Partition with mount point {mount_point} not found.")
-        return partition
-
     def check_resource_disk_mounted(self) -> bool:
         return False
 
--- a/lisa/tools/fips.py
+++ b/lisa/tools/fips.py
@@ -88,9 +88,6 @@ class Fips(Tool):
             boot_disk_part_uuid, boot_and_root_same
         )
 
-        # Update the grub configuration.
-        self.node.tools[GrubConfig].apply()
-
     def disable_fips(self) -> None:
         # dracut-fips provides FIPS support in the bootloader.
         self.node.os.uninstall_packages("dracut-fips")
@@ -103,9 +100,6 @@ class Fips(Tool):
         # Set the boot flag to the kernel command line.
         (boot_and_root_same, _) = self._get_boot_uuid()
         self.node.tools[GrubConfig].unset_boot_uuid(boot_and_root_same)
-
-        # Update the grub configuration.
-        self.node.tools[GrubConfig].apply()
 
     def _assert_kernel_fips_mode(self, expected: bool) -> None:
         """
--- a/lisa/tools/fips.py
+++ b/lisa/tools/fips.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 import re
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, cast
 
 from assertpy import assert_that
 
@@ -40,6 +40,10 @@ class Fips(Tool):
             os=node.os, message="FIPS tool only supported on CBLMariner 2.0 and 3.0."
         )
 
+    def __init__(self, node: "Node", *args: Any, **kwargs: Any) -> None:
+        super().__init__(node=node, args=args, **kwargs)
+        self._azl_os = cast(CBLMariner, node.os)
+
     @property
     def command(self) -> str:
         """
@@ -69,7 +73,7 @@ class Fips(Tool):
         FIPS is enabled or disabled on the system.
         """
         # We rely on the dracut-fips package for bootloader FIPS support.
-        self.node.os.package_exists("dracut-fips", assert_existance=expect_fips_mode)
+        self._azl_os.package_exists("dracut-fips", assert_existance=expect_fips_mode)
 
         # Kernel needs to be in the correct FIPS mode.
         assert_that(self.is_kernel_fips_mode()).described_as(
@@ -100,7 +104,7 @@ class Fips(Tool):
 
     def _enable_fips(self) -> None:
         # dracut-fips provides FIPS support in the bootloader.
-        self.node.os.install_packages("dracut-fips")
+        self._azl_os.install_packages("dracut-fips")
 
         # Set the fips flag to the kernel command line.
         self.node.tools[GrubConfig].set_kernel_cmdline_arg("fips", "1")
@@ -113,7 +117,7 @@ class Fips(Tool):
 
     def _disable_fips(self) -> None:
         # dracut-fips provides FIPS support in the bootloader.
-        self.node.os.uninstall_packages("dracut-fips")
+        self._azl_os.uninstall_packages("dracut-fips")
 
         # Set the fips flag to the kernel command line.
         self.node.tools[GrubConfig].set_kernel_cmdline_arg("fips", "0")
@@ -136,7 +140,9 @@ class Fips(Tool):
 
         disk = self.node.features[Disk]
         root_disk_partition = disk.get_partition_with_mount_point("/")
+        assert_that(root_disk_partition).is_not_none()
         boot_disk_partition = disk.get_partition_with_mount_point("/boot")
+        assert_that(boot_disk_partition).is_not_none()
 
         # If the boot and root devices are different, get the boot uuid it.
         if boot_disk_partition.name == root_disk_partition.name:
@@ -205,10 +211,10 @@ class AzlV3Fips(Fips):
         # when FIPS is disabled.
         if expect_fips_mode:
             for package in self._SYMCRYPT_PACKAGES:
-                self.node.os.package_exists(package, assert_existance=True)
+                self._azl_os.package_exists(package, assert_existance=True)
 
     def _enable_fips(self) -> None:
         super()._enable_fips()
 
         # AZL3 requres the SymCrypt provider for FIPS-compliant openssl.
-        self.node.os.install_packages(self._SYMCRYPT_PACKAGES)
+        self._azl_os.install_packages(self._SYMCRYPT_PACKAGES)
--- a/lisa/tools/fips.py
+++ b/lisa/tools/fips.py
@@ -10,7 +10,7 @@ from lisa.executable import Tool
 from lisa.operating_system import CBLMariner
 from lisa.tools import Blkid, Cat
 from .grub_config import GrubConfig
-from lisa.util import UnsupportedDistroException, get_matched_str
+from lisa.util import UnsupportedDistroException, get_matched_str, str_to_bool
 
 if TYPE_CHECKING:
     from lisa.node import Node
@@ -71,9 +71,33 @@ class Fips(Tool):
         self.node.os.package_exists("dracut-fips", assert_existance=expect_fips_mode)
 
         # Kernel needs to be in the correct FIPS mode.
-        self._assert_kernel_fips_mode(expect_fips_mode)
+        assert_that(self.is_kernel_fips_mode()).described_as(
+            "kernel fips mode"
+        ).is_equal_to(expect_fips_mode)
 
-    def enable_fips(self) -> None:
+    def set_fips_mode(self, fips_mode: bool) -> None:
+        """
+        Set the FIPS mode on the system.
+        This method should be called to enable or disable FIPS mode.
+        """
+        if fips_mode:
+            self._enable_fips()
+        else:
+            self._disable_fips()
+
+    def is_kernel_fips_mode(self) -> bool:
+        """
+        Get the current FIPS mode.
+        This method checks if the dracut-fips package is installed and if the
+        kernel FIPS mode is enabled.
+        """
+        fips_enabled = self.node.tools[Cat].read(
+            "/proc/sys/crypto/fips_enabled", force_run=True
+        )
+
+        return str_to_bool(fips_enabled)
+
+    def _enable_fips(self) -> None:
         # dracut-fips provides FIPS support in the bootloader.
         self.node.os.install_packages("dracut-fips")
 
@@ -86,7 +110,7 @@ class Fips(Tool):
         boot_arg_value = f"UUID={boot_disk_part_uuid}" if boot_disk_part_uuid else ""
         self.node.tools[GrubConfig].set_kernel_cmdline_arg("boot", boot_arg_value)
 
-    def disable_fips(self) -> None:
+    def _disable_fips(self) -> None:
         # dracut-fips provides FIPS support in the bootloader.
         self.node.os.uninstall_packages("dracut-fips")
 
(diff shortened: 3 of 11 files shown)


lisa/base_tools/sed.py
```diff
@@ -79,6 +79,33 @@ def delete_lines(self, pattern: str, file: PurePath, sudo: bool = False) -> None
...
+        Apply a raw sed expression to a file, as opposed to separating the expression into a pattern and replacement.
+        This allows for more advanced `sed` usage.
+        Args:
+            expression (str): The sed expression to apply.
+            file (str): The path to the file to modify.
+            sudo (bool, optional): Whether to execute the command with sudo. Defaults to False.
+        Returns:
+            None
+        """
+        self._log.debug(f"Applying raw sed expression '{expression}' to file '{file}' with sudo={sudo}")
```
squirrelsc: It doesn't need to log command line and "sudo" info, because LISA log the command line, and additional parameters automatically.

microsoft/testsuites/security/fips.py
```diff
@@ -1,14 +1,357 @@
...
+def assert_package_not_exists(os: OperatingSystem, package: str) -> None:
+    """
+    Assert that a package/tool is not installed on the node.
+    """
+    assert_that(os.package_exists(package)).described_as(
+        f"Package {package} is installed."
+    ).is_false()
+
+
+class AzlFips(ABC):
```
squirrelsc: It can be implemented by a tool. We have some virtual tools, they are not actually commands, but easy to be abtracted for reusing. In the tool, the `create` method can replace create_instance. This class can be called as "Fips", and derives a "AzLinuxFips/AzLinuxV2Fips,AzLinuxV3Fips" classes as needed.

microsoft/testsuites/security/fips.py
```diff
@@ -19,9 +362,111 @@
...
+        # Skip the test if the image is already not FIPS enabled.
+        ensure_fips_expectations(log, node, variables, should_be_fips=True)
+
+        # Disable FIPS on the system and make sure it is worked.
+        azl_fips = AzlFips.create_instance(log, node)
+        azl_fips.disable_fips()
+        node.reboot()
+        azl_fips.assert_fips_disabled()
+
+        log.debug("Successfully disabled FIPS.")
```
squirrelsc: It can be info level to output a successful information.

microsoft/testsuites/security/fips.py
```diff
@@ -1,14 +1,357 @@
...
+    FIPS is enabled or disabled, and to enable or disable FIPS.
+    '''
+    # In AZL2, the command openssl md5 should fail with error messages like:
+    #   Error setting digest
+    #   131590634539840:error:060800C8:digital envelope routines:EVP_DigestInit_ex:disabled for FIPS:crypto/evp/digest.c:135: # noqa: E501
+    _md5_expected_failure_pattern = re.compile(
+        "Error setting digest\r\n.*EVP_DigestInit_ex:disabled for FIPS.*", re.M
+    )
+
+    def assert_fips_enabled(self) -> None:
```
squirrelsc: It can provide additional parameters to merge with `assert_fips_disabled`, like `assert_status(self, is_enabled)`

microsoft/testsuites/security/fips.py
LiliDeng: Restore it to its original status, since the environment may be reused.

microsoft/testsuites/security/fips.py
tobiasb-ms: I considered doing that, the code for assert_status would be several
```
if is_enabled:
   # whatever thing for enabled
else:
   # whatever disabled thing
It felt better to have them as totally separate functions.
```
Do you have a strong preference here?

squirrelsc: I prefer it, because 50%+ code are same or similar. If convert true/false to 1/0, most code is the same. So sharing logic is always better.

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+    @classmethod
+    def create(cls, node: "Node", *args: Any, **kwargs: Any) -> Tool:
+        """
+        Factory method to create an instance of the Grub tool.
+        """
+        if isinstance(node.os, CBLMariner):
+            if node.os.information.release == "2.0":
+                return GrubAzl2(node, args, kwargs)
+            if node.os.information.release == "3.0":
+                return GrubAzl3(node, args, kwargs)
```
squirrelsc, line 42:
```python
            if node.os.information.release == "2.0":
                return GrubAzl2(node, args, kwargs)
            elif node.os.information.release == "3.0":
                return GrubAzl3(node, args, kwargs)
            else:
                raise UnsupportedDistroException...
```

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+) -> None:
+    """
+    Assert that a package/tool is installed on the node.
+    """
+    assert_that(os.package_exists(package)).described_as(
+        f"Package {package} is installed."
+    ).is_equal_to(should_exist)
+
+
+class Grub(Tool, ABC):
```
squirrelsc: Move Grub to a separated tool file. It's not a part of fips tool.

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+            self.remove_kernel_cmdline_arg(r"boot")
+        else:
+            self.set_kernel_cmdline_arg(f"boot=UUID={uuid}")
+
+    @abstractmethod
+    def unset_boot_uuid(self, same_as_root: bool) -> None:
+        """
+        Unset the boot UUID.
+        """
+
```
squirrelsc: We don't use `abstractmethod` in LISA, because sometimes it has some problems on reflection. So, use normal methods and raise NotImplementedError inside.

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+    @abstractmethod
+    def set_kernel_cmdline_arg(self, arg: str) -> None:
+        """
+        Append the specified kernel command line argument to the grub configuration.
+        """
+
+    @abstractmethod
+    def apply(self) -> None:
```
squirrelsc: It doesn't need to be a separated method, because azl2 doesn't support it, and it's always called after set/unset. So it's just an private method in azl3.

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+    @property
+    def command(self) -> str:
+        return self._command
+
+    @property
+    def can_install(self) -> bool:
+        return True
+
+    @abstractmethod
+    def set_fips_mode(self, fips_mode: bool) -> None:
```
squirrelsc: This method should be in the fips class.

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+from lisa.operating_system import CBLMariner, OperatingSystem
+from lisa.tools import Blkid, Cat, Sed
+from lisa.util import UnsupportedDistroException, get_matched_str
+
+if TYPE_CHECKING:
+    from lisa.operating_system import Posix
+    from lisa.node import Node
+
+
+def assert_package_existance(
```
squirrelsc: you can extend the `os.package_exists` method to support it for all places. It may be like `def package_exists(self, package: Union[str, Tool, Type[Tool]], raise_error: bool = False) -> bool:

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+    os: OperatingSystem, package: str, should_exist: bool
+) -> None:
+    """
+    Assert that a package/tool is installed on the node.
+    """
+    assert_that(os.package_exists(package)).described_as(
+        f"Package {package} is installed."
```
squirrelsc: Should the message be like "pakcage {package} should be installed, but not found."?

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+        if same_as_root:
+            self.remove_kernel_cmdline_arg(r"boot")
+        else:
+            self.set_kernel_cmdline_arg(f"boot=UUID={uuid}")
+
+    @abstractmethod
+    def unset_boot_uuid(self, same_as_root: bool) -> None:
+        """
+        Unset the boot UUID.
+        """
```
squirrelsc: Why the `set_bool_uuid` have the default implementation, but `unset_boot_uuid` cannot? You can write a default method, and subclass can overwrite it, if the logic is different. So the code looks more consistent.

lisa/tools/fips.py
```diff
@@ -0,0 +1,379 @@
...
+        Set the boot UUID to the specified value.
+        """
+        if same_as_root:
+            self.remove_kernel_cmdline_arg(r"boot")
+        else:
+            self.set_kernel_cmdline_arg(f"boot=UUID={uuid}")
+
+    @abstractmethod
+    def unset_boot_uuid(self, same_as_root: bool) -> None:
```
squirrelsc, line 79: What does `same_as_root` mean? Is it a concept in grub? If not, it needs a clearer name.

microsoft/testsuites/security/fips.py
```diff
@@ -18,10 +27,106 @@
...
+    @TestCaseMetadata(
+        description="""
+            Ensures that an AZL machine is fips enabled.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            supported_os=[CBLMariner],
+        ),
+    )
+    def verify_fips_is_enabled_azl(
```
squirrelsc: For test case names, how about like `verify_azl_fips_is_enabled`? Move azl to the front, so it's easier to identify azl cases.

microsoft/testsuites/security/fips.py
```diff
@@ -32,61 +137,84 @@ class Fips(TestSuite):
         4. Verify that FIPS was enabled properly
         """,
         priority=3,
+        requirement=simple_requirement(
+            unsupported_os=[CBLMariner],
```
squirrelsc: Why azl is unsupported?

squirrelsc: if it's by design, I prefer to have the same `verify_fips_enable` for all distros. Azl can go with a separated path.

microsoft/testsuites/security/fips.py
```diff
@@ -32,61 +137,84 @@ class Fips(TestSuite):
...
+            variables (Dict[str, Any]): A dictionary of variables containing the
+                                    'testing_fips_image' key.
+        Raises:
+            SkippedException: If the FIPS image expectation does not match the actual image SKU.
+        """
+        log.debug(f"ensure_fips_expectations: should_be_fips is '{should_be_fips}'")
+        log.debug(f"ensure_fips_expectations: variables is '{variables}'")
+
+        # First, try to deduce the FIPS image type from the variables dictionary.
+        fips_image_map = {"yes": True, "no": False}
```
squirrelsc: If use "true/false" as input, it should be more straightforward.

lisa/tools/fips.py
tobiasb-ms: I'm not convinced it should, but happy to discuss.

This line asserts that the package is either installed or is not in stalled, depending on the `should_exist` parameter. So it's not always the case that the package was installed but isn't -- it could also be the case that the package is installed but shouldn't be.

My understanding of `described_as` is that it's more of a description of the test being performed than the final state. In this case, if `should_exist` is `True`, but the package is not installed, we get an assertion message like this:
```
AssertionError: [Package dracut-fips is installed.] Expected <False> to be equal to <True>, but was not.
```

Meaning: "We're asserting whether or not dracut-fips is installed. The answer is False, but we expected True."

If you'd prefer, I'll add logic to have a more friendly string like you suggested, but I wanted to provide context around the overall usage of this function and how the assertion message ends up being formatted.

squirrelsc: How about "Package {package} installation status is unexpected"?

tobiasb-ms: That's great wording.

lisa/tools/fips.py
```diff
@@ -0,0 +1,213 @@
...
+        self.node.os.package_exists("dracut-fips", assert_existance=expect_fips_mode)
+
+        # Kernel needs to be in the correct FIPS mode.
+        self._assert_kernel_fips_mode(expect_fips_mode)
+
+    def enable_fips(self) -> None:
+        # dracut-fips provides FIPS support in the bootloader.
+        self.node.os.install_packages("dracut-fips")
+
+        from lisa.tools import GrubConfig
```
tobiasb-ms: @squirrelsc -- I found that when I had this at the top of the file (the normal place), `python` complained of a circular dependency. I couldn't immediately see it though. Any ideas?

microsoft/testsuites/security/fips.py
tobiasb-ms: Do we have common code that can parse `true/false` as a bool?

squirrelsc: No yet, can you create one in util/__init__.py like this? https://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python

lisa/tools/fips.py
squirrelsc: tool reference tool, it needs to use like `from .grub import GrubConfig`

lisa/tools/fips.py
tobiasb-ms: `Fips` has the methods `enable_fips` and `disable_fips`. These handle the overall, multi-part process of enabling/disabling fips. One of those steps is to set or clear the `fips` kernel flag. That process is different depending on how we configure `grub`, which is encapsulated in the `GrubConfig.set_fips_mode` class and its subclasses (Note: I just changed it from `Grub` to `GrubConfig`).

Perhaps changing this method's name would help clarify that? Maybe `set_fips_kernel_flag`?

lisa/tools/fips.py
tobiasb-ms: `set_boot_uuid` is the same across versions of azl, so putting that code in the base class makes sense.

`unset_boot_uuid` is different -- and different enough that having a base class implementation isn't helpful. And it doesn't make sense to me to arbitrarily decide that one version of azl's implementation is the "default", which is what would go in the base class.

For me, it's more clear to have the base class not implement this method.

microsoft/testsuites/security/fips.py
```diff
@@ -18,10 +27,106 @@
...
+    @TestCaseMetadata(
+        description="""
+            Ensures that an AZL machine is fips enabled.
+        """,
+        priority=1,
+        requirement=simple_requirement(
+            supported_os=[CBLMariner],
+        ),
+    )
+    def verify_azl_fips_is_enabled(
```
squirrelsc: I think this test case can be merged with `verify_azl_fips_is_disabled`. When only one of test cases runs, it doesn't need to have two cases.

The logic can check consistent between variable/metadata with the actual fips mode. Let me know, if there are other situations.

microsoft/testsuites/security/fips.py
```diff
@@ -18,10 +27,106 @@
...
+        azl_fips = node.tools[Fips]
+        azl_fips.enable_fips()
+        node.reboot()
+        azl_fips.assert_fips_mode(True)
+
+        log.info("Successfully enabled FIPS.")
+
+        # Re-disable FIPS to make sure the test can be run multiple times.
+        azl_fips.disable_fips()
+        node.reboot()
```
squirrelsc: Recover the fips status, because we don't know the order of test cases. If the test case run before verifying initial state case, the initial state case will be failed. If it's hard to guarantee reset, you can mark the environment is dirty at end of this test case. So this env won't be reused.

microsoft/testsuites/security/fips.py
```diff
@@ -18,10 +27,106 @@
...
+            1. Enable FIPS on the AZL machine
+            2. Restart the machine
+            3. Verify that FIPS was enabled properly
+        """,
+        priority=2,
+        requirement=simple_requirement(
+            supported_os=[CBLMariner],
+        ),
+    )
+    def verify_azl_fips_enable(
```
squirrelsc: It's similar with another pair of test cases. This case should be able to merge with `verify_azl_fips_disable`. We don't prefer a fine granularity of test cases, especially it makes more skipped test cases.

lisa/tools/grub_config.py
```diff
@@ -0,0 +1,172 @@
...
+        posix_os: Posix = self.node.os  # type: ignore
+        posix_os.install_packages(self._package)
+        return self._check_exists()
+
+
+class GrubConfigAzl2(GrubConfig):
+    def __init__(self, node: "Node", *args: Any, **kwargs: Any) -> None:
+        super().__init__("grubby", "grubby", node, *args, **kwargs)
+
+    def set_fips_mode(self, fips_mode: bool) -> None:
```
squirrelsc, line 97: The GrubConfig tool should provide functionality itself, and not provide high level functionality. So the `set_fips_mode` should be moved to fips level.

lisa/tools/fips.py
squirrelsc: My point is the "fips" related methods shouldn't be in the GrubConfig class. The GrubConfig tool provides its own functionality, and not provide high level functionality. So, the set_fips_mode should be moved to fips level.

tobiasb-ms: But the exact way to set the `fips` flag is fundamentally different between AZL2 and AZL3, at the `grub` level, not at the higher-level `Fips` level.

At the `Fips` level, you want to either turn that specific kernel flag on or off.

At the grub level, there are different ways to do that depending on how use/configure grub on the specific platform.

Moving this function into `Fips` means those classes need to the internals of grub our grub configuration, which I don't want.

microsoft/testsuites/security/fips.py
tobiasb-ms: Wouldn't it be better, when looking at test results, to immediately see from the test name that enabling or disabling fips specifically didn't work? Rather than something generic like `verify_azl_fips_enable_disable` failed and needing to dig into the specific assert to know which part failed? As a consumer of the test output, I much prefer that.

squirrelsc: The name could be like `verify_azl_fips_enablement`. Some partners will look at each skipped test cases, so small number of skipped tests will reduce the overhead a little bit. LISA test results always bring the error message, it's the right place to understand failure reasons. A shorter test case name is also better for documentation and display.

One more reason, it's important, but I haven't thought about. The test cases run on Azl only, but the distro version checks happen runtime. So, one test case can reduce one deployment.

lisa/tools/fips.py
squirrelsc: It seems like azl2/3 have different way on removing args in grub config. It's a different behavior from grub config, not for fips only. The remove_kernel_cmdline_arg is well wrapped the behavior.

Another way to simplify it to accept "fips=0" in azl3, is to add `same_as_root` for `remove_kernel_cmdline_arg`. So it can be used in azl3 to decide set "key=" or remove "key".

lisa/tools/fips.py
```diff
@@ -9,6 +9,7 @@
 from lisa.executable import Tool
 from lisa.operating_system import CBLMariner
 from lisa.tools import Blkid, Cat
+from .grub_config import GrubConfig
```
squirrelsc, line 14: Please follow dev env setup in vscode, the isort can fix the order automatically when saving a file.

lisa/util/__init__.py
```diff
@@ -908,3 +908,21 @@ def check_panic(content: str, stage: str, log: "Logger") -> None:
...
         raise KernelPanicException(stage, panics)
+
+
+def str_to_bool(value: str) -> Union[bool, None]:
+    """
+    Convert a string to a boolean value.
+    returns True for "true", False for "false", and None for any other value.
+    Allows for casing and leading/trailing whitespace.
+    """
+    str_to_bool_map = {
```
tobiasb-ms, line 920: @squirrelsc -- I made this only support `true` and `false` (allowing for casing and leading/trailing whitespace). Happy to expand that to things like `0/1`, `yes/no` or other things you think would be good.

Also, I didn't go looking for other places that could use this -- kind of a needle in a haystack.

squirrelsc: It would be great, if it follows the original implementation in Python lib, which supports 0/1, yes/no. And it should also support bool. For example, `str_to_bool(True)` should return `True`. In case it's used for safe conversion.

lisa/tools/fips.py
tobiasb-ms: I took a step back and considered what I was really accomplishing with the differences between AZL2 and AZL3 grub configs. A fundamental issue is that "remove kernel parameter" had fundamentally different semantics between the two, which is not good. I was also trying very hard to maintain kind of an arbitrary design goal I had. By relaxing that design goal, I was able to come up with a clean solution that gets rid of the "remove" function in favor of "set", as that's really what clients want to do (at least in my use cases).

Refactored accordingly.

lisa/tools/fips.py
tobiasb-ms: I'm fixing "bigger issues" right now, then I'll make sure everything is formatted correctly. Thanks for the reminder.

lisa/util/__init__.py
tobiasb-ms: Added support for `bool` and for `0/1` and `yes/no`.

lisa/util/__init__.py
```diff
@@ -919,8 +919,18 @@ def str_to_bool(value: str) -> Union[bool, None]:
     str_to_bool_map = {
         "true": True,
         "false": False,
+        "yes": True,
+        "no": False,
+        "1": True,
+        "0": False,
```
squirrelsc, line 926: It looks `0`, and `1` integers are not added.

tobiasb-ms: Oh, we wanted `ints` too? Sorry for the misunderstanding.

That brings up a couple interesting questions.

Near as I can tell, simply casting an int to a bool returns `True` for non-zero and `False` for zero. Do we want to support those semantics? Basically... If they give us an `int`, just return `bool(int)`?

And that brings up semantics of how we handle a `string` that can be converted to an `int`. Right now, we support `"0"` and `"1"`. Should instead convert it to an `int` (if possible) and then just use `bool(int)`?

squirrelsc: I hope this method is smart to handle string and basic types, because the type depends on how it's written or the source. For example, if a value is passed from env variable or command line, they are always string. But if they are from yaml, it may be typed or string.

tobiasb-ms: I also notice that I accidentally allow `value` to be an `int` there -- I'll fix that regardless of what we decide.

microsoft/testsuites/security/fips.py
tobiasb-ms: I combined these.

microsoft/testsuites/security/fips.py
tobiasb-ms: combined

microsoft/testsuites/security/fips.py
tobiasb-ms: I combined these.

lisa/util/__init__.py
tobiasb-ms: renamed to `to_bool` and added support for `int`. It now supports reasonable strings, `int` and `bool`.

squirrelsc:
@tobiasb-ms please continue to fix errors in the PR checks.
@LiliDeng LGTM

tobiasb-ms:
> @tobiasb-ms please continue to fix errors in the PR checks.
All checks now pass.

lisa/features/disks.py
```diff
@@ -42,6 +42,17 @@ def get_partition_with_mount_point(
         else:
             return None
 
+    def get_partition_with_mount_point_or_raise(
```
squirrelsc: Please add an argument like `raise_error` to `get_partition_with_mount_point` instead of a new method.

tobiasb-ms: I thought about that, but it doesn't actually solve the problem I was addressing, which is having a guaranteed return type of `PartitionInfo` instead of `Union[PartitionInfo, None]`. This makes `mypy` checks happy.

If you'd really rather not have the extra function, I'll just cast it on the calling side but it defeats the purpose of the change.

squirrelsc: use `assert root_disk_partition, ....`, to fix mypy error.

tobiasb-ms: There we go. I had tried `assert_that`, but I guess using the built in `assert` is the key. Thanks.

LiliDeng (approved):

PR #3690, opened 2025-03-04
dom0_kernel_installer: generate initramfs and use PARTUUID

pupacha:
Using older initramfs does not work anymore as there are some kernel module dependencies. Hence, generate a new one. Also, replace UUID with PARTUUID since initramfs fails to understand the UUID.

While at it, update validate() method to check if node's os is Mariner.

### DIFF
--- a/lisa/tools/lsblk.py
+++ b/lisa/tools/lsblk.py
@@ -35,6 +35,8 @@ class PartitionInfo(object):
     total_blocks: int = 0
     percentage_blocks_used: int = 0
     fstype: str = ""
+    uuid: str = ""
+    part_uuid: str = ""
 
     @property
     def is_mounted(self) -> bool:
@@ -59,6 +61,8 @@ class PartitionInfo(object):
         total_blocks: int = 0,
         percentage_blocks_used: int = 0,
         fstype: str = "",
+        uuid: str = "",
+        part_uuid: str = "",
     ):
         self.name = name
         self.mountpoint = mountpoint
@@ -69,6 +73,8 @@ class PartitionInfo(object):
         self.total_blocks = total_blocks
         self.percentage_blocks_used = percentage_blocks_used
         self.fstype = fstype
+        self.uuid = uuid
+        self.part_uuid = part_uuid
 
 
 @dataclass
@@ -79,6 +85,7 @@ class DiskInfo(object):
     type: str = ""
     fstype: str = ""
     partitions: List[PartitionInfo] = field(default_factory=list)
+    uuid: str = ""
 
     @property
     def is_os_disk(self) -> bool:
@@ -116,6 +123,7 @@ class DiskInfo(object):
         dev_type: str = "",
         fstype: str = "",
         partitions: Optional[List[PartitionInfo]] = None,
+        uuid: str = "",
     ):
         self.name = name
         self.mountpoint = mountpoint
@@ -123,6 +131,7 @@ class DiskInfo(object):
         self.type = dev_type
         self.fstype = fstype
         self.partitions = partitions if partitions else []
+        self.uuid = uuid
 
 
 class Lsblk(Tool):
@@ -175,6 +184,7 @@ class Lsblk(Tool):
                 size=int(lsblk_entry["size"]),
                 dev_type=lsblk_entry["type"],
                 fstype=lsblk_entry["fstype"],
+                uuid=lsblk_entry["uuid"],
             )
 
             for child in lsblk_entry.get("children", []):
@@ -185,6 +195,8 @@ class Lsblk(Tool):
                         size=int(child["size"]),
                         dev_type=child["type"],
                         fstype=child["fstype"],
+                        uuid=child["uuid"],
+                        part_uuid=child["partuuid"],
                     )
                 )
             # add disk to list of disks
@@ -222,6 +234,8 @@ class Lsblk(Tool):
                     dev_type=lsblk_entry["type"],
                     mountpoint=lsblk_entry["mountpoint"],
                     fstype=lsblk_entry["fstype"],
+                    uuid=lsblk_entry["uuid"],
+                    part_uuid=lsblk_entry["partuuid"],
                 )
             )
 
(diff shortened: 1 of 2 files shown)


lisa/transformers/dom0_kernel_installer.py
```diff
@@ -128,19 +128,20 @@ def install(self) -> str:
             )
         else:
             # Mariner 3.0 initrd
-            target = f"/boot/initramfs-{current_kernel}.img"
-            link = f"/boot/initramfs-{new_kernel}.img"
+            initramfs = f"/boot/initramfs-{new_kernel}.img"
+            dracut_cmd = f"dracut --force {initramfs} {new_kernel}"
+            node.execute(dracut_cmd, sudo=True, shell=True)
```
smit-gardhariya: This will run dracut in case of 2.0 as well. Do we really need it?

lisa/transformers/dom0_kernel_installer.py
```diff
@@ -258,3 +262,25 @@ def _update_mariner_config(
...
+    ).stdout.strip()
+    partuuid = node.execute(
+        "lsblk -o MOUNTPOINT,PARTUUID | grep '/ ' | awk '{print $2}'",
+        sudo=True,
+        shell=True,
+        expected_exit_code=0,
+    ).stdout.strip()
+
+    # initramfs can only understand PARTUUID
+    sed.substitute(
```
smit-gardhariya, line 273: This also will run for 2.0 as well. Do we need that?

pupacha: This is good to have for 2.0 as well, since PARTUUID is used in 2.0's config.

lisa/transformers/dom0_kernel_installer.py
pupacha: Thanks for the catch. Update it to run only for 3.0

lisa/transformers/dom0_kernel_installer.py
```diff
@@ -258,3 +265,25 @@ def _update_mariner_config(
...
         sudo=True,
     )
+
+    uuid = node.execute(
+        "lsblk -o MOUNTPOINT,UUID | grep '/ ' | awk '{print $2}'",
+        sudo=True,
+        shell=True,
+        expected_exit_code=0,
+    ).stdout.strip()
+    partuuid = node.execute(
```
anirudhrb: Can lsblk tool do this?

lisa/transformers/dom0_kernel_installer.py
```diff
@@ -67,8 +67,11 @@ def _output_names(self) -> List[str]:
         return []
 
     def validate(self) -> None:
-        # nothing to validate before source installer started.
-        ...
+        if not isinstance(self._node.os, CBLMariner):
+            raise UnsupportedDistroException(
```
squirrelsc, line 71: Is it a regression, or only work with Azure Linux before?

pupacha: This kernel installer only works with Azure Linux.

lisa/transformers/dom0_kernel_installer.py
pupacha: No, lsblk tool does not fetch UUID and PARTUUID.

anirudhrb (approved):

lisa/transformers/dom0_kernel_installer.py
squirrelsc: Can you change lsblk to fetch the IDs?

pupacha: Changed lsblk tool fetch UUID and PARTUUID

lisa/tools/lsblk.py
```diff
@@ -274,21 +289,37 @@ def get_disks(self, force_run: bool = False) -> List[DiskInfo]:
...
+                    return partition if find_partition else disk
+
+        return None
 
-        raise LisaException(f"Could not find disk with mountpoint {mountpoint}")
+    def find_disk_by_mountpoint(
+        self, mountpoint: str, force_run: bool = False
+    ) -> DiskInfo:
+        result = self._find_by_mountpoint(mountpoint, force_run, find_partition=False)
+        if result is None or isinstance(result, PartitionInfo):
```
squirrelsc: `or isinstance(result, PartitionInfo)` is not needed. You can rely on above logic to make sure it returns the right type.

lisa/tools/lsblk.py
```diff
@@ -274,21 +289,37 @@ def get_disks(self, force_run: bool = False) -> List[DiskInfo]:
...
-            # check if any of the partitions is mounted and mountpoint matches
             for partition in disk.partitions:
                 if partition.mountpoint == mountpoint:
-                    return disk
+                    return partition if find_partition else disk
+
+        return None
 
-        raise LisaException(f"Could not find disk with mountpoint {mountpoint}")
+    def find_disk_by_mountpoint(
```
squirrelsc: It doesn't need to modify existing method, and just add a new method.

```python
def find_partition_by_mountpoint(
    self, mountpoint: str, force_run: bool = False
) -> PartitionInfo:
    result = self.find_disk_by_mountpoint(mountpoint, force_run)
    
    for partition in disk.partitions:
            return partition

    return Exception...
```

pupacha: Sure. Made this change.

lisa/tools/lsblk.py
pupacha: This will not be applicable after taking your suggested approach for `find_partition_by_mountpoint`

lisa/tools/lsblk.py
```diff
@@ -290,6 +305,21 @@ def find_disk_by_mountpoint(
 
         raise LisaException(f"Could not find disk with mountpoint {mountpoint}")
 
+    def find_partition_by_mountpoint(
+        self, mountpoint: str, force_run: bool = False
+    ) -> PartitionInfo:
+        disk = self.find_disk_by_mountpoint(mountpoint, force_run=force_run)
+
+        if disk is None:
```
squirrelsc: It doesn't need, because find_disk_by_mountpoint raises exception already.

pupacha: My bad. removed it.

squirrelsc:
@LiliDeng LGTM

smit-gardhariya (approved):

LiliDeng (approved):

PR #3691, opened 2025-03-04
ch_tests_tool: support using pmem for ch block tests

pupacha:
change kernel command line to create a pmem device which can be used for running block perf tests.

### DIFF
--- a/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
+++ b/microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
@@ -12,9 +12,10 @@ from lisa import Node
 from lisa.executable import Tool
 from lisa.features import SerialConsole
 from lisa.messages import TestStatus, send_sub_test_result_message
-from lisa.operating_system import CBLMariner
+from lisa.operating_system import CBLMariner, Ubuntu
 from lisa.testsuite import TestResult
 from lisa.tools import (
+    Cat,
     Chmod,
     Chown,
     Dmesg,
@@ -25,9 +26,10 @@ from lisa.tools import (
     Lsblk,
     Mkdir,
     Modprobe,
+    Sed,
     Whoami,
 )
-from lisa.util import LisaException, find_groups_in_lines
+from lisa.util import LisaException, UnsupportedDistroException, find_groups_in_lines
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -497,7 +508,53 @@ def _configure_vdpa_devices(self, node: Node) -> None:
             node.tools[Chown].change_owner(file=PurePath(device_path), user=user)
             node.tools[Chmod].chmod(path=device_path, permission=permission, sudo=True)
 
-    def _set_data_disk(self) -> None:
+    def _get_pmem_for_block_tests(self) -> str:
```
smit-gardhariya, line 522: Will this work for 2.0 as well?

pupacha: yes, this is unrelated to Mariner 2.0/3.0 versions.

smit-gardhariya: Agree. What i meant was, in 2.0 we used some config file. I was asking if grub2-mkconfig works for 2.0

For 3.0, we use grub only anyway

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -497,7 +508,53 @@ def _configure_vdpa_devices(self, node: Node) -> None:
...
+        """
+        Following is the last usable entry in e829 table and is safest to use for
```
anirudhrb:
```suggestion
        Following is the last usable entry in e820 table and is safest to use for
```

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
```diff
@@ -497,7 +508,53 @@ def _configure_vdpa_devices(self, node: Node) -> None:
...
+        pmem since its most likely to be free. 0x0000001000000000 is 64G.
+
+        [    0.000000] BIOS-e820: [mem 0x0000001000000000-0x00000040ffffffff] usable
+
+        """
+        memmap_str = "memmap=16G!64G"
```
anirudhrb: It would be better to make this configurable from runbook. So that we can easily run these tests on different servers.

pupacha: Made it configurable.

microsoft/testsuites/cloud_hypervisor/ch_tests_tool.py
pupacha: Made changes for 2.0

anirudhrb (approved):

smit-gardhariya (approved):

pupacha:
@squirrelsc Please review this as well.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3692, opened 2025-03-04
Update to the list of suites executed as part of LTP Lite Test Case

umfranci:
- The list of tests that are available in the latest tag of LTP have been updated and 'fsx' is no longer available.
- Making LTP Run Timeout value configurable.

### DIFF
--- a/microsoft/testsuites/ltp/ltpsuite.py
+++ b/microsoft/testsuites/ltp/ltpsuite.py
@@ -29,7 +29,18 @@ from microsoft.testsuites.ltp.ltp import Ltp
 )
 class LtpTestsuite(TestSuite):
     _TIME_OUT = 21000
-    LTP_LITE_TESTS = ["math", "ipc", "mm", "sched", "pty", "fs", "uevent", "syscalls", "hyperthreading", "irq"]
+    LTP_LITE_TESTS = [
+        "math",
+        "ipc",
+        "mm",
+        "sched",
+        "pty",
+        "fs",
+        "uevent",
+        "syscalls",
+        "hyperthreading",
+        "irq",
+    ]
     LTP_REQUIRED_DISK_SIZE_IN_GB = 2
 
     @TestCaseMetadata(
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -28,8 +28,19 @@
...
+        "ipc",
+        "mm",
+        "sched",
+        "pty",
+        "fs",
+        "uevent",
+        "syscalls",
+        "hyperthreading",
+        "irq",
+    ]
```
squirrelsc: How long the new tests taken? The list is used for "lite" run, so it doesn't need to be a completed list. The "ltp_test" is a case visible parameter, you can add more in the runbook.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -28,8 +28,19 @@
     """,
 )
 class LtpTestsuite(TestSuite):
-    _TIME_OUT = 18000
-    LTP_LITE_TESTS = ["math", "fsx", "ipc", "mm", "sched", "pty", "fs"]
+    _TIME_OUT = 21000
```
squirrelsc: I'm not sure if LTP is stable enough now. Previously, it's timed out, even the test is dead. So extending the time may cause the whole pipeline running longer to wait it timeout.
Please change timeout to a parameter of ltp key process, and set default value to 18000 and make it configurable. This value can be extended, because it's not configurable.

umfranci: Thanks for the comments @squirrelsc. With the increased list, the test case was indeed timing out a few times. Hence reverted to the previous list with fsx removed. Also, keeping the TestSuite Timeout at 18000 but made LTP Run Timeout configurable.

microsoft/testsuites/ltp/ltpsuite.py
umfranci: Made relevant changes in latest commit.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3693, opened 2025-03-05
Xfstests tool & test update to support Azure file Share - SMB testing

shekharsorot:
Xfstest is a file system and share testing tool that is maintained by kernel.org and is often used for performance, stability, feature operability and regression testing.

This LISA tool feature is an update to the existing Xfstests.py within Microsoft test cases and has been refactored to be able to run SMB as well as NFS tests in addition of existing set of Managed disks ( XFS, EXT3/4 and BTRFS ) as well as NVME

It's an implementation of "https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git", although users can specify custom repository and forks as well with their own test cases.

It's intended to use this via custom test suites to test SMB and NFS client functionality with various SMB and NFS servers (in addition to AZFiles ) currently in use across cloud and on-premises environments (You can call XFStest to create config file for existing shares, specify custom flags in addition to existing flags and call run_test() to trigger testing // reporting.

Note: detailed docstrings are included in tool class and respective method definitions.

Non-Breaking but Major changes:

TIMEOUT changes to 6 hours to allow additional time for managed disk tests to run.
SMB // CIFS test is hardcoded for specific cases as these are well known tests for Azure Files - SMB Premium // Standard
XFStests will now always pull the 'master' branch which is the last known stable release.
"verify_azure_file_share" test case priority has been reduced back to P3 as the test takes < 60 min to complete for CIFS
Xfstests will by default generate dmesg output per case. This can be found in your runtime / log / <Test Project > /testss / xfstests directory for analysis. 
SubtestMessages will now contain diff of input vs output file for failed cases in the additional information field. Can be used for automated reporting when test environment is not preserved.
Upon Failure of any sub test, HTML will output only the failed subtests.
In case of success ( including skipped  and expunged ), HTML report will dump the list of passed test cases along with test options.

Usage: 
import using "from lisa.testsuite import TestResult"

Instantiate & install using "xfstests = node.tools[Xfstests]"

to create a config:
xfstest.set_local_config(
            scratch_dev="/dev/sdb",
            scratch_mnt="/mnt/scratch",
            test_dev="/dev/sdc",
            test_folder="/mnt/test",
            file_system="xfs",
            test_section="xfs-custom",
            mount_opts="noatime",
            testfs_mount_opts="noatime",
            additional_parameters={"TEST_DEV2": "/dev/sdd"},
            overwrite_config=True
            )
Note: You can create multiple test sections using this method and call run_test() in sequence.
Please note guidance in "https://github.com/kdave/xfstests/blob/master/README.config-sections"

To exclude global list of tests
xfstests.set_excluded_tests(excluded_tests)

To run tests
xfstests.run_test(
            test_section=test_type,
            log_path=log_path,
            result=result,
            data_disk=data_disk,
            test_cases=test_cases,
            timeout=self.TIME_OUT - 30,
        )

Tool instance will create and send sub-test messages to log as well as HTML & Junit.
On failure, it will display the failed sub-test as well as XFStest message to allow for troubleshooting.
On success, it will display list of passed // skipped sub-tests via PassedException.
Note: Expunged cases are treated as skipped
        
Supported OS: 
redhat, suse , fedora, mariner, debian


squirrelsc:
Can you improve the current version directly?

shekharsorot:
> Can you improve the current version directly?

I will work to update the XFStesting suite in Microsoft folder. It will take some time due to long run durations and validations for each test case. I will modify xfstesting.py to use the new Xfstests tool.

squirrelsc (commented):
> I will work to update the XFStesting suite in Microsoft folder. It will take some time due to long run durations and validations for each test case. I will modify xfstesting.py to use the new Xfstests tool. 

There is a xfstests.py file already, can you directly fix there?

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,90 @@ def _prepare_data_disk(
...
+            _test_folder: fs_url_dict[file_share_name],
+            _scratch_folder: fs_url_dict[scratch_name],
+        }
+        azure_file_share.create_fileshare_folders(test_folders_share_dict)
+    # else:
+    # NFS yet to be implemented
+    return fs_url_dict
+
+
+# DEPRECATED !!!!
```
squirrelsc: Remove it directly. The code history will have it.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,90 @@ def _prepare_data_disk(
...
-        node.execute(f"mkdir {folder_name}", sudo=True)
-        echo.write_to_file(
-            f"{share} {folder_name} cifs {fstab_info}",
-            node.get_pure_path("/etc/fstab"),
-            sudo=True,
-            append=True,
+    environment: Environment,
+    file_share_name: str,
+    scratch_name: str,
+    azure_file_share: Any,
```
squirrelsc: The type is: `Union[AzureFileShare, Nfs]`

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,90 @@ def _prepare_data_disk(
...
+    storage account.
+    Returns: Dict[str, str] - A dictionary containing the file share names
+    and their respective URLs.
+    """
+    if isinstance(azure_file_share, AzureFileShare):
+        file_share_protocol = "SMB"
+    elif isinstance(azure_file_share, Nfs):
+        file_share_protocol = "NFS"
+    else:
+        raise LisaException("Unsupported file share protocol")
```
squirrelsc:
```python
raise LisaException(f"Unsupported file share type: {type(azure_file_share)}")
```

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,90 @@ def _prepare_data_disk(
...
+            allow_shared_key_access=allow_shared_key_access,
+            enable_private_endpoint=enable_private_endpoint,
+            quota_in_gb=file_share_quota_in_gb,
         )
+        test_folders_share_dict = {
+            _test_folder: fs_url_dict[file_share_name],
+            _scratch_folder: fs_url_dict[scratch_name],
+        }
+        azure_file_share.create_fileshare_folders(test_folders_share_dict)
+    # else:
```
squirrelsc: Raise an exception, until it's implemented.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,90 @@ def _prepare_data_disk(
...
+            file_share_names=[file_share_name, scratch_name],
+            environment=environment,
+            sku=storage_account_sku,
+            kind=storage_account_kind,
+            allow_shared_key_access=allow_shared_key_access,
+            enable_private_endpoint=enable_private_endpoint,
+            quota_in_gb=file_share_quota_in_gb,
         )
+        test_folders_share_dict = {
+            _test_folder: fs_url_dict[file_share_name],
```
squirrelsc: Define a class, instead of use the dict. It's easy to misuse the dict.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,90 @@ def _prepare_data_disk(
...
+# def _get_smb_version(node: Node) -> str:
+#     if node.tools[KernelConfig].is_enabled("CONFIG_CIFS_SMB311"):
+#         version = "3.1.1"
+#     else:
+#         version = "3.0"
+#     return version
 
 
-def _prepare_azure_file_share(
+def _deploy_azure_file_share(
```
squirrelsc, line 112: Can we remove this method, direct call to azure file share feature?

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -126,8 +228,9 @@ class Xfstesting(TestSuite):
     # generic/738 case might cause hang more than 4 hours on old kernel
     # TODO: will figure out the detailed reason of every excluded case.
     # exclude generic/680 for security reason.
+    # include generic/211 for testing
     excluded_tests = (
```
squirrelsc: I see you defined a new list, can we use it?

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -126,8 +228,9 @@ class Xfstesting(TestSuite):
     # generic/738 case might cause hang more than 4 hours on old kernel
     # TODO: will figure out the detailed reason of every excluded case.
     # exclude generic/680 for security reason.
+    # include generic/211 for testing
```
squirrelsc: It doesn't explain in the code, when you remove something, it's confusing.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -504,11 +609,17 @@ def verify_btrfs_nvme_datadisk(self, log_path: Path, result: TestResult) -> None
         ),
         timeout=TIME_OUT,
         use_new_environment=True,
-        priority=5,
+        priority=3,
     )
     def verify_azure_file_share(
         self, log: Logger, log_path: Path, result: TestResult
     ) -> None:
+        """
```
squirrelsc: Write it in the description of test case metadata, so it can be generated in the LISA document.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -520,40 +631,41 @@ def verify_azure_file_share(
...
-                xfstests,
-                result,
-                test_dev=fs_url_dict[file_share_name],
-                scratch_dev=fs_url_dict[scratch_name],
-                excluded_tests=self.excluded_tests,
-                mount_opts=mount_opts,
-            )
-        finally:
-            # clean up resources after testing.
-            azure_file_share.delete_azure_fileshare([file_share_name, scratch_name])
```
squirrelsc, line 556: Keep the deleted `finally` part, it needs to recycle the resources and clean up for the VM.

microsoft/testsuites/xfstests/xfstesting.py
shekharsorot: I envisioned this as an encapsulation of SMB as well as NFS. It just makes the test suite code a bit cleaner as it will be a common code base for two test cases.

microsoft/testsuites/xfstests/xfstesting.py
shekharsorot: This list is a carry forward from the original xfstesting test suite and really did not work for SMB as we don't have 100% implementation of SMB RFC.
BTRFS/xxx , XFS/xxx does not really works for shared files and just adds to clutter in the report.

That is the reason for us to create seperate exception list for SMB // NFS to reduce the runtime to < 60 min. Plus, as we add more features to File Share server side, we will revisit and update the exception list and test case list on regular basis.

On that note, in order to reduce runtime for disk tests, we can use the same approach as shared files to create unique exception list for XFS, BTRFS and NVME to known test cases that will not work.

microsoft/testsuites/xfstests/xfstesting.py
shekharsorot: These comments are all carry over from previous testsuite. We need some time to validate if the issues relevant to the mentioned cases have been fixed or now.
Which is why they were left as is.

microsoft/testsuites/xfstests/xfstesting.py
shekharsorot: done

microsoft/testsuites/xfstests/xfstesting.py
shekharsorot: This will need changes to Nfs and AzureFileShare class in features.py
This is a carry over code from previous iteration of the xfstesting.py, hence no changes.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +108,63 @@ def _prepare_data_disk(
...
+    file_share_protocol: str = "SMB",
+    file_share_quota_in_gb: int = 500,
+) -> Dict[str, str]:
+    """
+    About: This method will provision azure file shares on a new // existing
+    storage account.
+    Returns: Dict[str, str] - A dictionary containing the file share names
+    and their respective URLs.
+    """
+    if isinstance(azure_file_share, AzureFileShare):
```
squirrelsc, line 129: It looks confusing by passed in `file_share_protocol`, and overwrite it here. It means the `file_share_protocol` is not a required variable.

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: Follow previous code, pass in the names as a list like `[file_share_name, scratch_name]`, and then return a dict with their name/values. It saves code, and not hardcode them.

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: I mean remove this line below. You don't need to tell something added in an "excluded" list.
`    # include generic/211 for testing`

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -491,11 +569,14 @@ def verify_btrfs_nvme_datadisk(self, log_path: Path, result: TestResult) -> None
...
-
-        Downgrading priority from 3 to 5. The file share relies on the
-         storage account key, which we cannot use currently.
-        Will change it back once file share works with MSI.
+        azure file share.
+        The case will provision storage account with private endpoint
+        and use access key // ntlmv2 for authentication.
+        This will be changed to MSI in the near future
+        To modify the test case parameters:
+        Update the mount options via '_default_smb_mount'
```
squirrelsc: Don't mention internal variables in the description. The description is published separated without code. To improve readability, you can give a brief introduction on the options, include/excluded tests. If people are really interested on details, they can look at the code.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -156,28 +190,72 @@ def dependencies(self) -> List[Type[Tool]]:
 
     def run_test(
         self,
-        test_type: str,
+        # test_type: str,
```
squirrelsc: delete it directly

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -156,28 +190,72 @@ def dependencies(self) -> List[Type[Tool]]:
...
+            test_section="generic/quick",
+            data_disk="/dev/sdb",
+            test_cases="generic/001 generic/002",
+            timeout=14400,
         )
+        """
+        # if Test group is specified, and exists in local.config, run tests.
+        if test_section:
+            self.run_async(
+                f"-s {test_section} -E exclude.txt {test_cases} > xfstest.log 2>&1",
```
squirrelsc: the command line is not much different with below one, so just write like,

```python
cmd = ""
if test_section:
   cmd += f"-s {test_section} "
cmd += f"-g generic/quick -E exclude.txt {test_cases} > xfstest.log 2>&1"

self.run_async(cmd, ...
```

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -156,28 +190,72 @@ def dependencies(self) -> List[Type[Tool]]:
 
     def run_test(
         self,
-        test_type: str,
+        # test_type: str,
         log_path: Path,
-        result: TestResult,
+        result: "TestResult",
+        test_section: str,
```
squirrelsc: give it a default value `test_section: str = "generic"`. So it doesn't need separated handling in following code.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -374,56 +548,106 @@ def create_send_subtest_msg(
...
-                self.create_send_subtest_msg(result, raw_message, test_type, data_disk)
+                self.create_send_subtest_msg(
+                    test_result=result,
+                    raw_message=raw_message,
+                    test_section=test_section,
+                    data_disk=data_disk,
+                )
 
             if not self.node.shell.exists(results_path):
                 self._log.error(
```
squirrelsc: Remove the error line, if it needs to raise an exception. The error line is not needed.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -435,36 +659,54 @@ def check_test_results(
                     self._log.debug(
                         f"All pass in xfstests, total pass case count is {pass_count}."
                     )
-                    return
+                    # Xperimental : Passedexception with message/content of XFSTestLog.
+                    # Intent is to display XFSReport on HTML pager.
+                    # TODO: Fix this not displaying output
+                    raise PassedException(
```
squirrelsc: Don't raise passed exception. The test case should pass by nature.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -435,36 +659,54 @@ def check_test_results(
...
-                for fail_case in fail_cases.split():
-                    fail_info += find_patterns_in_lines(
-                        raw_message, [re.compile(f".*{fail_case}.*$", re.MULTILINE)]
-                    )[0][0]
-                fail_cases_list = fail_cases.split()
-                raise LisaException(
-                    f"Fail {fail_count} cases of total {total_count}, fail cases"
-                    f" {fail_cases}, details {fail_info}, please investigate."
-                )
+                if fail_match:
```
squirrelsc, line 671: If use `if`, it doesn't need to assert in the `if` block.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -435,36 +659,54 @@ def check_test_results(
...
+                    assert fail_match
```
squirrelsc: If use if, it doesn't need to assert in the if block.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -435,36 +659,54 @@ def check_test_results(
...
     ) -> None:
-        if "generic" == test_type:
-            test_type = "xfs"
+        """
+        About:This method is intended to be called by check_test_results method only.
+        This method will copy the output of XFSTest results to the Log folder of host
+        calling LISA. Files copied are xfsresult.log, check.log and all failed cases
+        files if they exist.
+        """
+        # if "generic" == test_section:
```
squirrelsc: Remove it directly.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +108,63 @@ def _prepare_data_disk(
         node.execute(f"mkdir {mount_point}", sudo=True)
 
 
-def _get_smb_version(node: Node) -> str:
-    if node.tools[KernelConfig].is_enabled("CONFIG_CIFS_SMB311"):
-        version = "3.1.1"
-    else:
-        version = "3.0"
-    return version
+# DEPRECATED !!!
```
squirrelsc: Remove it directly

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +108,63 @@ def _prepare_data_disk(
...
+# This does not works on newer kernels.
+# Pls see:
+# https://lists.samba.org/archive/samba-technical/2018-June/128806.html
+# def _get_smb_version(node: Node) -> str:
```
squirrelsc: Remove it directly.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -479,15 +721,163 @@ def save_xfstests_log(
...
+        xfstests_path = self.get_xfstests_path()
+        test_class = case.split("/")[0]
+        test_id = case.split("/")[1]
+        result_path = xfstests_path / f"results/{test_section}/{test_class}"
+        return_message: str = ""
+        # <TODO> this needs to be fixed as it's spilling over to console output.
+        if self.node.tools[Ls].path_exists(str(result_path), sudo=True):
+            # Note. This will dump a lot of output on debug console screen.
+            # Only un-comment for debugging.
+            # self._log.debug(
```
squirrelsc: Don't use comment to switch code. Use a test case variable to control it.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -479,15 +721,163 @@ def save_xfstests_log(
...
+        """
+        # Use the cat tool to read the file content
+        if not Path(file_path).exists():
+            self._log.debug(f"{file_path} doesn't exist.")
+            return ""
+        cat_tool = self.node.tools[Cat]
+        file_content = cat_tool.run(file_path, force_run=True)
+        return str(file_content.stdout)
+
+    def create_xfstest_stack_info(
```
squirrelsc, line 792: Will it be called by every result? I'm worry about the performance, if so.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -479,15 +721,163 @@ def save_xfstests_log(
...
+                    str(full_out), sudo=True
+                ) and self.node.tools[Ls].path_exists(str(fail_out), sudo=True):
+                    diff_result = self.node.tools[Diff].comparefiles(
+                        src=full_out,
+                        dest=fail_out,
+                    )
+                # else if full_out is null, return the fail_out file content.
+                # In some test cases, full_out is not generated due to permissions
+                # or other issues. However a fail file will always exists in such cases.
+                elif self.node.tools[Ls].path_exists(str(fail_out), sudo=True):
```
LiliDeng: There is a case I saw there is no fail out but full out, then below exception was thrown out.
```
  File "C:\app\lsg-lisa\lisa\microsoft\testsuites\xfstests\xfstests.py", line 869, in create_xfstest_stack_info
    return_message = f"DIFF: {diff_result}\n\nDMESG: {dmesg_result}"
                              ^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'diff_result' where it is not associated with a value
```

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -17,18 +17,71 @@
...
+    "generic/257 generic/258 generic/286 generic/306 generic/308 generic/310 "
+    "generic/313 generic/315 generic/339 generic/340 generic/344 generic/345 "
+    "generic/346 generic/354 generic/360 generic/391 generic/393 generic/394 "
+    "generic/406 generic/412 generic/422 generic/428 generic/432 generic/433 "
+    "generic/437 generic/443 generic/450 generic/451 generic/452 generic/460 "
+    "generic/464 generic/465 generic/469 generic/524 generic/528 generic/538 "
+    "generic/565 generic/567 generic/568 generic/586 generic/590 generic/591 "
+    "generic/598 generic/599 generic/604 generic/609 generic/615 generic/632 "
+    "generic/634 generic/635 generic/637 generic/638 generic/639 "
+)
```
LiliDeng, line 79: Maintaining only exclude tests should be enough?

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -17,18 +17,71 @@
...
+# Section : Global options
 _scratch_folder = "/mnt/scratch"
 _test_folder = "/mnt/test"
+_xfstests_repp = "https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git"
+_xfstests_branch = "master"
```
LiliDeng: remove unused variables

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -18,20 +18,39 @@
...
+    installed (default) from https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git
+    Mirrored daily from kernel.org repository.
+    For details, refer to https://github.com/kdave/xfstests/blob/master/README
+    """
+
+    # This is the default repo and branch for xfstests.
+    # Override this via _install method if needed.
     repo = "https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git"
-    branch = "v2024.02.09"
+    branch = "master"
```
LiliDeng, line 44: master branch may be unstable

microsoft/testsuites/xfstests/xfstests.py
shekharsorot: We have a requirement from internal team members that they would like to view the passed cases results in HTML when running lisa locally. Its easier and quick to identify potential issues. 
Currently this is the only way we have found to trigger a message output to html provider and not causing the test case to be marked failed or skipped.
We used this from provisioning.py where this is being used by the test case '_smoke_test' to display output on report.

microsoft/testsuites/xfstests/xfstests.py
shekharsorot: After checking with Lili, it seems we cannot use the -d flag here to output the ls. I will remove it as it adds significant verbose output to the logs and slows down the diff generation after xfstests run.

microsoft/testsuites/xfstests/xfstests.py
squirrelsc: Do you mean the names of xfstests in the html report? How about the junit report? It's an xml, and can include all subtests. The PassedException can bring message, but it's used for partial passed.

microsoft/testsuites/xfstests/xfstests.py
shekharsorot: Master branch in this case is actually stable as its updated once every 3 months.
The branches "for-next' and 'patches-in-queue' are actually unstable as they are updated weekly and daily

microsoft/testsuites/xfstests/xfstests.py
shekharsorot: This is quite fast and is only called after all tests have run.
Takes about a min for an XFSrun of about 600 subtests. We can parameterize this if needed. Please suggest.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,51 @@ def _prepare_data_disk(
...
+    file_share_protocol: str = "SMB",
+    file_share_quota_in_gb: int = 500,
+) -> Dict[str, str]:
+    """
+    About: This method will provision azure file shares on a new // existing
+    storage account.
+    Returns: Dict[str, str] - A dictionary containing the file share names
+    and their respective URLs.
+    """
+    # if isinstance(azure_file_share, AzureFileShare):
```
squirrelsc: Remove the commented code.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,51 @@ def _prepare_data_disk(
...
+            quota_in_gb=file_share_quota_in_gb,
         )
+        test_folders_share_dict = {
+            _test_folder: fs_url_dict[file_share_name],
+            _scratch_folder: fs_url_dict[scratch_name],
+        }
+        azure_file_share.create_fileshare_folders(test_folders_share_dict)
+    elif isinstance(azure_file_share, Nfs):
+        # NFS yet to be implemented
+        raise LisaException("Skipping NFS deployment. Pending implementation.")
```
squirrelsc: Missed else, if neither smb nor nfs.

microsoft/testsuites/xfstests/xfstesting.py
shekharsorot: complete exclusion list is much longer than actual test case and exclusion.
the SMB exclusion list covers tests that will 100% fail with AZfiles today, while smb_testcases containts list of updating tests that we know work for GA features.

microsoft/testsuites/xfstests/xfstesting.py
shekharsorot: removed

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,46 @@ def _prepare_data_disk(
...
-    node.execute("cp -f /etc/fstab /etc/fstab_cifs", sudo=True)
-    for folder_name, share in test_folders_share_dict.items():
-        node.execute(f"mkdir {folder_name}", sudo=True)
-        echo.write_to_file(
-            f"{share} {folder_name} cifs {fstab_info}",
-            node.get_pure_path("/etc/fstab"),
-            sudo=True,
-            append=True,
+    environment: Environment,
+    file_share_name: str,
```
squirrelsc: Use a list, so the method can support any number of shares.
```suggestion
    names: List[str],
```

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,46 @@ def _prepare_data_disk(
...
+    scratch_name: str,
+    azure_file_share: Union[AzureFileShare, Nfs],
+    allow_shared_key_access: bool = True,
+    enable_private_endpoint: bool = True,
+    storage_account_sku: str = "Standard_LRS",
+    storage_account_kind: str = "StorageV2",
+    file_share_protocol: str = "SMB",
```
squirrelsc: Remove it, because it's not used in the method.

microsoft/testsuites/xfstests/xfstests.py
squirrelsc: Another place needs to be updated too. Please check all related places for each code comment.

microsoft/testsuites/xfstests/xfstests.py
squirrelsc: 1 min for 600 tests are acceptable. It looks extending less than 5% of running time.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -656,9 +656,9 @@ def _execute_xfstests(
         # test_group
         # a test group for XFS will fail for a config for ext or btrfs
         test_group: str = ""
-        if not test_type:
-            test_group = f"{file_system.name}/{test_type}"
-        if test_type == "generic":
+        if not test_type or test_type == file_system.name:
+            test_group = f"{file_system.name}/quick"
+        elif test_type == "generic":
```
squirrelsc: The default value of test_type is `generic/quick`, and the test_type is given in your own code. So, it shouldn't be set to `generic`.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -18,20 +18,34 @@
     Suse,
     Ubuntu,
 )
-from lisa.testsuite import TestResult
-from lisa.tools import Cat, Chmod, Echo, Git, Make, Pgrep, Rm, Sed
+
+if TYPE_CHECKING:
+    from lisa.testsuite import TestResult
```
LiliDeng: have you tried to revert it back since you didn't change the location of this file. see whether we still see cycle import issue

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -299,49 +411,109 @@ def get_xfstests_path(self) -> PurePath:
...
+                Empty signifies disk target
+            testfs_mount_opts (str): (O)The test filesystem mount options to be used for
+                testing.Defaults to mount_opts
+            additional_parameters (dict): (O)Additional parameters (dict) to be used for
+                testing
+            overwrite_config (bool): (O)If True, the existing local.config file will be
+                overwritten
+        Example:
+        xfstest.set_local_config(
+            scratch_dev="/dev/sdb",
```
LiliDeng, line 458: please add comments for the datadisk label name

microsoft/testsuites/xfstests/xfstests.py
LiliDeng: @shekharsorot this change is not needed, I have tried.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -504,7 +559,7 @@ def verify_btrfs_nvme_datadisk(self, log_path: Path, result: TestResult) -> None
         ),
         timeout=TIME_OUT,
         use_new_environment=True,
-        priority=5,
+        priority=3,
```
LiliDeng: no need change this, till it can be pass

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: Please keep the finally part.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -156,14 +182,70 @@ def dependencies(self) -> List[Type[Tool]]:
...
-        result: TestResult,
+        result: "TestResult",
+        test_section: str,
+        test_group: str = "generic/quick",
         data_disk: str = "",
+        test_cases: str = "",
         timeout: int = 14400,
     ) -> None:
+        """About: This method runs XFSTest on a given node with the specified
+        test group and test cases.If test_section is not specified , test is
```
squirrelsc: The code logic is changed, the test_section is a required argument. So it's impossible to be not specified. Please update the comments to make it consistent.

microsoft/testsuites/xfstests/xfstesting.py
squirrelsc: Please take this comment to simplify the code, and make it more generic.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -479,15 +713,208 @@ def save_xfstests_log(
...
+        test_id = case.split("/")[1]
+        result_path = xfstests_path / f"results/{test_section}/{test_class}"
+        # <TODO> this needs to be fixed as it's spilling over to console output.
+        # ls_tool = self.node.tools[Ls]
+        cat_tool = self.node.tools[Cat]
+        result = ""
+        if not self.node.shell.exists(result_path):
+            self._log.debug(f"No files found in path {result_path}")
+            # non terminating error !!!
+            result = f"No files found in path {result_path}"
```
squirrelsc, line 852: Why isn't a terminating failure? If there is no result, it means xfstest has no result found. maybe some errors on setup progress, so the case should be failed.

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -55,39 +106,47 @@ def _prepare_data_disk(
...
-    node.execute("cp -f /etc/fstab /etc/fstab_cifs", sudo=True)
-    for folder_name, share in test_folders_share_dict.items():
-        node.execute(f"mkdir {folder_name}", sudo=True)
-        echo.write_to_file(
-            f"{share} {folder_name} cifs {fstab_info}",
-            node.get_pure_path("/etc/fstab"),
-            sudo=True,
-            append=True,
+    environment: Environment,
+    names: List[str],
```
squirrelsc: Sorry, I just found actually it needs key/value pairs. So, 

1. pass in the pairs to the method like `names: dict[str, str]`.
2. And then pass into `azure_file_share.create_file_share` like `file_share_names=names.values()`
3. Set test_folders_share_dict.
   ```python
   test_folders_share_dict : Dict[str, str] = {}
   for key, value in names:
        test_folders_share_dict[key] = fs_url_dict[value]
   ```

microsoft/testsuites/xfstests/xfstesting.py
```diff
@@ -575,6 +574,13 @@ def verify_azure_file_share(
                 node.os, "current distro not enable cifs module."
             )
         xfstests = self._install_xfstests(node)
+        # These local variables are needed to track resource retention
+        # on demand / test failure.
+        # This is to ensure that the storage account is not deleted
+        # if the test fails and the keep_environment is set to "always" or "failed".
+
+        keep_environment = environment.platform.runbook.keep_environment
```
squirrelsc: Please use a separated test variable to control it, instead of the keep_env value. As I said, it doesn't work as you expected, and it breaks the code structure. For example, if you run on a pre-created VM, the keep_env may be `no`, so the share will be deleted too.

microsoft/testsuites/xfstests/xfstests.py
squirrelsc: @shekharsorot please check this comment.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3696, opened 2025-03-06
hyperv: cleanup stale vm before raising exception

pupacha:
If start_vm() fails due to unassigned IP address, an exception is thrown but it results in a stale VM on the HyperV host. This is inefficient as it continues to consume cpu, memory and storage resource of the host. Hence, cleanup the VM before raising the exception.

### DIFF
--- a/lisa/sut_orchestrator/hyperv/platform_.py
+++ b/lisa/sut_orchestrator/hyperv/platform_.py
@@ -306,8 +306,14 @@ class HypervPlatform(Platform):
                 hv=hv,
                 vm_name=vm_name,
             )
-            # Start the VM
-            hv.start_vm(name=vm_name, extra_args=extra_args)
+
+            try:
+                # Start the VM
+                hv.start_vm(name=vm_name, extra_args=extra_args)
+            except Exception as ex:
+                # avoid stale VMs; delete the VM before raising exception
+                self._delete_node(node_context, wait_delete=False, log=log)
+                raise ex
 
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/hyperv/platform_.py
```diff
@@ -367,34 +373,37 @@ def _expand_root_partition(self, node: RemoteNode) -> None:
...
+            f"{vm_name} serial log process exited with {result.exit_code}. "
+            f"stdout: {result.stdout}"
+        )
+
+    def _delete_nodes(self, environment: Environment, log: Logger) -> None:
         run_in_parallel(
             [
                 partial(
-                    _delete_node,
+                    self._delete_node,
```
anirudhrb, line 403: Did you test this? Now that you made it a member function, I believe it requires one additional argument: `self`.

pupacha: I thought the same too. To my surprise, this works only without `self` parameter. It throws error saying too many arguments when `self` is additionally passed. This could be because we are passing the function already as `self.<method>`

squirrelsc: This change should work.

pupacha: Yes, it does work!

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

anirudhrb (approved):

PR #3699, opened 2025-03-09
Local source path change for non-zip artifact

paulli2017:
Local source path change for non-zip artifact


lisa/sut_orchestrator/baremetal/source.py
```diff
@@ -38,7 +38,7 @@ def _extract_artifacts(
                     source_path = _extract(artifact_path)
                     artifact_local_path.append(Path(source_path))
                 else:
-                    artifact_local_path.append(artifact_path)
+                    artifact_local_path.append(artifact_path.parent)
```
squirrelsc, line 41: Please explain why make this change in the PR.

PR #3700, opened 2025-03-10
deployment: support ipv6

LiliDeng:


lisa/util/shell.py
```diff
@@ -39,6 +39,18 @@
 _spawn_initialization_error_pattern = re.compile(
     r"(Failed to parse line \'b[\'\"](?P<linux_profile_error>.*?)[\'\"]\' as integer)"
 )
+_ip_v6_pattern = re.compile(
```
squirrelsc: Please try the standard lib, but parse IPv4 firstly.

```python
import ipaddress

def is_valid_ipv6(address):
    try:
        ipaddress.IPv6Address(address)
        return True
    except ipaddress.AddressValueError:
        return False
```

squirrelsc:
How is the setting used? It looks just supported from underlying.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -2215,33 +2217,41 @@ def get_vm(platform: "AzurePlatform", node: Node) -> Any:
...
     for network_interface in vm.network_profile.network_interfaces:
         assert isinstance(
             network_interface.id, str
         ), f"actual: {type(network_interface.id)}"
         nic_name = get_matched_str(network_interface.id, NIC_NAME_PATTERN)
         nic = network_client.network_interfaces.get(resource_group_name, nic_name)
         if nic.primary:
-            if not nic.ip_configurations[0].public_ip_address:
+            if use_ipv6:
+                nic_index = 1
```
squirrelsc, line 2245: Is it stable? please check if any stable way to detect ipv6 nic.

lisa/util/shell.py
```diff
@@ -101,6 +104,14 @@ def wait_tcp_port_ready(
     return is_ready, result
 
 
+def _is_valid_ipv6(address: str) -> bool:
+    try:
+        ipaddress.IPv6Address(address)
```
squirrelsc: Check IPv4, so it won't raise exception on most scenarios. Raise exceptions is much higher cost the if/else.

lisa/schema.py
```diff
@@ -1315,6 +1315,7 @@ class RemoteNode(Node):
     username: str = constants.DEFAULT_USER_NAME
     password: str = ""
     private_key_file: str = ""
+    use_ipv6: bool = False
```
squirrelsc, line 1318: I'm thinking if it's confusing, when set it to true for other platforms like ready? Does ipv6 work directly for ready, if set the IP address to ipv6?

LiliDeng: I tried to set ipv6 for ready platform, it worked with changes in shell.py

lisa/sut_orchestrator/azure/common.py
LiliDeng: Currently, IPv6 must be used together with IPv4. In LISA testing, when multiple NICs are used, we assign different subnets. Therefore, if a NIC has two IP configurations, the secondary one is typically IPv6. Additionally, I also check the IP version in this section.

squirrelsc (dismissed):

squirrelsc (approved):

PR #3704, opened 2025-03-10
junit: Add properties to the JUnit.xml file

Ankita13-code:
- In case of CVM test cases, we need the information about `vmsize` and `disk_encryption` status in the JUnit.xml file for testing it on different combinations.
- This PR adds these values to the JUnit.xml file in the form of system properties by extracting it from the `information` in the `TestResultMessageBase` class. 
- In future, this method can also be used to add more properties to the Junit file if required

### DIFF
--- a/lisa/notifiers/junit.py
+++ b/lisa/notifiers/junit.py
@@ -300,6 +300,18 @@ class JUnit(Notifier):
         testcase.attrib["classname"] = class_name
         testcase.attrib["time"] = self._get_elapsed_str(elapsed)
 
+        # Add a standard system-properties element for all test cases
+        if hasattr(message, 'information') and isinstance(message.information, dict):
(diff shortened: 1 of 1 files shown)


lisa/notifiers/junit.py
```diff
@@ -300,6 +300,18 @@ def _add_test_case_result(
         testcase.attrib["classname"] = class_name
         testcase.attrib["time"] = self._get_elapsed_str(elapsed)
 
+        # Add a standard system-properties element for all test cases
+        if hasattr(message, "information") and isinstance(message.information, dict):
```
squirrelsc, line 304: Please make sure it works within Azure DevOps test results. A few teams tried to add it, but found it either not working or breaking the test results. Please make sure no regression on ADO test results.

PR #3706, opened 2025-03-11
NIC passthrough perf test

smit-gardhariya:
Add support to create libvirt device pool from BDF list
Add NIC perf test for passthrough scenario

### DIFF
--- a/lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
+++ b/lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
@@ -186,23 +186,77 @@ class LibvirtDevicePool(BaseDevicePool):
             vendor_id=vendor_id,
             device_id=device_id,
         )
-        primary_nic_iommu = self.get_primary_nic_id()
-        for item in device_list:
-            device = DeviceAddressSchema()
-            domain, bus, slot, fn = self._parse_pci_address_str(addr=item.slot)
-            device.domain = domain
-            device.bus = bus
-            device.slot = slot
-            device.function = fn
-            iommu_group = self._get_device_iommu_group(device)
-            is_vfio_pci = self._is_driver_vfio_pci(device)
+        bdf_list = [i.slot for i in device_list]
+        self._create_pool(pool_type, bdf_list)
 
-            if not is_vfio_pci and iommu_group not in primary_nic_iommu:
+    def create_device_pool_from_pci_addresses(
+        self,
+        pool_type: HostDevicePoolType,
+        pci_addr_list: List[str],
+    ) -> None:
+        self.available_host_devices[pool_type] = {}
+        for bdf in pci_addr_list:
+            domain, bus, slot, fn = self._parse_pci_address_str(bdf)
+            device = self._get_pci_address_instance(domain, bus, slot, fn)
+            iommu_group = self._get_device_iommu_group(device)
+
+            # Get all the devices of that iommu group
+            iommu_path = f"/sys/kernel/iommu_groups/{iommu_group}/devices"
+            bdf_list = [i.strip() for i in self.host_node.tools[Ls].list(iommu_path)]
+            bdf_list.append(bdf.strip())  # append the given device in list
+
+            self._create_pool(pool_type, bdf_list)
+
+    def _create_pool(
+        self,
+        pool_type: HostDevicePoolType,
+        bdf_list: List[str],
+    ) -> None:
+        iommu_grp_of_used_devices = []
+        primary_nic_iommu = self.get_primary_nic_id()
+        for bdf in bdf_list:
+            domain, bus, slot, fn = self._parse_pci_address_str(bdf)
+            dev = self._get_pci_address_instance(domain, bus, slot, fn)
+            is_vfio_pci = self._is_driver_vfio_pci(dev)
+            iommu_group = self._get_device_iommu_group(dev)
+
+            if iommu_group in iommu_grp_of_used_devices:
+                # No need to add this device in pool as one of the devices for this
+                # iommu group is in use
+                continue
+
+            if is_vfio_pci:
+                # Do not consider any device for pool if any device of same iommu group
+                # is already assigned
+                pool = self.available_host_devices.get(pool_type, {})
+                pool.pop(iommu_group, [])
+                self.available_host_devices[pool_type] = pool
+                iommu_grp_of_used_devices.append(iommu_group)
+            elif (
+                iommu_group not in primary_nic_iommu
+                and iommu_group not in iommu_grp_of_used_devices
+            ):
                 pool = self.available_host_devices.get(pool_type, {})
                 devices = pool.get(iommu_group, [])
-                devices.append(device)
+                if dev not in devices:
+                    devices.append(dev)
                 pool[iommu_group] = devices
                 self.available_host_devices[pool_type] = pool
+
+    def _get_pci_address_instance(
+        self,
+        domain: str,
+        bus: str,
+        slot: str,
+        fn: str,
+    ) -> DeviceAddressSchema:
+        device = DeviceAddressSchema()
+        device.domain = domain
+        device.bus = bus
+        device.slot = slot
+        device.function = fn
+
+        return device
 
     def _add_device_passthrough_xml(
         self,
--- a/lisa/sut_orchestrator/util/device_pool.py
+++ b/lisa/sut_orchestrator/util/device_pool.py
@@ -1,6 +1,10 @@
-from typing import Any, List, Optional
+from typing import Any, List, Optional, cast
 
-from lisa.sut_orchestrator.util.schema import HostDevicePoolSchema, HostDevicePoolType
+from lisa.sut_orchestrator.util.schema import (
+    HostDevicePoolSchema,
+    HostDevicePoolType,
+    VendorDeviceIdIdentifier,
+)
 from lisa.util import LisaException
 
 
@@ -13,6 +17,13 @@ class BaseDevicePool:
         pool_type: HostDevicePoolType,
         vendor_id: str,
         device_id: str,
+    ) -> None:
+        raise NotImplementedError()
+
+    def create_device_pool_from_pci_addresses(
+        self,
+        pool_type: HostDevicePoolType,
+        pci_addr_list: List[str],
     ) -> None:
         raise NotImplementedError()
 
@@ -44,22 +55,40 @@ class BaseDevicePool:
                         f"Pool type '{pool_type}' is not supported by platform"
                     )
             for config in device_configs:
-                vendor_device_list = config.devices
-                if len(vendor_device_list) > 1:
-                    raise LisaException(
-                        "Device Pool does not support more than one "
-                        "vendor/device id list for given pool type"
+                devices = config.devices
+                if isinstance(devices, list) and all(
+                    isinstance(d, VendorDeviceIdIdentifier) for d in devices
+                ):
+                    if len(devices) > 1:
+                        raise LisaException(
+                            "Device Pool does not support more than one "
+                            "vendor/device id list for given pool type"
+                        )
+
+                    vendor_device_id = devices[0]
+                    assert vendor_device_id.vendor_id.strip()
+                    vendor_id = vendor_device_id.vendor_id.strip()
+
+                    assert vendor_device_id.device_id.strip()
+                    device_id = vendor_device_id.device_id.strip()
+
+                    self.create_device_pool(
+                        pool_type=config.type,
+                        vendor_id=vendor_id,
+                        device_id=device_id,
                     )
+                elif isinstance(devices, dict):
+                    bdf_list = devices.get("pci_bdf", [])
+                    assert bdf_list, "Key not found: 'pci_bdf'"
+                    pci_addr_list: List[str] = cast(List[str], bdf_list)
 
-                vendor_device_id = vendor_device_list[0]
-                assert vendor_device_id.vendor_id.strip()
-                vendor_id = vendor_device_id.vendor_id.strip()
-
-                assert vendor_device_id.device_id.strip()
-                device_id = vendor_device_id.device_id.strip()
-
-                self.create_device_pool(
-                    pool_type=config.type,
-                    vendor_id=vendor_id,
-                    device_id=device_id,
-                )
+                    # Create pool from the list of PCI addresses
+                    self.create_device_pool_from_pci_addresses(
+                        pool_type=config.type,
+                        pci_addr_list=pci_addr_list,
+                    )
+                else:
+                    raise LisaException(
+                        f"Unknown device identifier of type: {type(devices)}"
+                        f", value: {devices}"
+                    )
--- a/lisa/sut_orchestrator/util/schema.py
+++ b/lisa/sut_orchestrator/util/schema.py
@@ -1,6 +1,6 @@
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import List
+from typing import List, Union, cast
 
 from dataclasses_json import dataclass_json
 
@@ -12,9 +12,16 @@ class HostDevicePoolType(Enum):
 
 @dataclass_json()
 @dataclass
-class DeviceIdentifier:
+class VendorDeviceIdIdentifier:
     vendor_id: str = ""
     device_id: str = ""
+
+
+@dataclass_json()
+@dataclass
+class PciAddressIdentifier:
+    # list of bdf like 0000:3b:00.0 - <domain>:<bus>:<slot>.<fn>
+    pci_bdf: List[str] = field(default_factory=list)
 
 
 # Configuration options for device-passthrough for the VM.
@@ -22,7 +29,9 @@ class DeviceIdentifier:
 @dataclass
 class HostDevicePoolSchema:
     type: HostDevicePoolType = HostDevicePoolType.PCI_NIC
-    devices: List[DeviceIdentifier] = field(default_factory=list)
+    devices: Union[List[VendorDeviceIdIdentifier], PciAddressIdentifier] = field(
+        default_factory=lambda: cast(List[VendorDeviceIdIdentifier], [])
+    )
 
 
 @dataclass_json()
--- a/lisa/tools/iperf3.py
+++ b/lisa/tools/iperf3.py
@@ -119,6 +119,7 @@ class Iperf3(Tool):
         use_json_format: bool = False,
         one_connection_only: bool = False,
         daemon: bool = True,
+        interface_ip: str = "",
     ) -> Process:
         # -s: run iperf3 as server mode
         # -D: run iperf3 as a daemon
@@ -135,6 +136,8 @@ class Iperf3(Tool):
             cmd += f" -f {report_unit} "
         if port:
             cmd += f" -p {port} "
+        if interface_ip:
+            cmd += f" -B {interface_ip}"
         process = self.node.execute_async(
             f"{self.command} {cmd}", shell=True, sudo=True
         )
--- a/lisa/tools/netperf.py
+++ b/lisa/tools/netperf.py
@@ -31,10 +31,17 @@ class Netperf(Tool):
     def dependencies(self) -> List[Type[Tool]]:
         return [Gcc, Git, Make, Texinfo]
 
-    def run_as_server(self, port: int = 30000, daemon: bool = True) -> None:
+    def run_as_server(
+        self,
+        port: int = 30000,
+        daemon: bool = True,
+        interface_ip: str = "",
+    ) -> None:
         cmd = f"netserver -p {port} "
         if not daemon:
             cmd += " -D "
+        if interface_ip:
+            cmd += f" -L {interface_ip}"
         self.node.execute(
             cmd,
             sudo=True,
@@ -50,9 +57,11 @@ class Netperf(Tool):
         test_name: str = "TCP_RR",
         seconds: int = 150,
         time_unit: int = 1,
+        interface_ip: str = "",
         send_recv_offset: str = "THROUGHPUT, THROUGHPUT_UNITS, MIN_LATENCY, MAX_LATENCY, MEAN_LATENCY, REQUEST_SIZE, RESPONSE_SIZE, STDDEV_LATENCY",  # noqa: E501
     ) -> Process:
         # -H: Specify the target machine and/or local ip and family
+        # -L: Specify the IP for client interface to be used
         # -p: Specify netserver port number and/or local port
         # -t: Specify test to perform
         # -n: Set the number of processors for CPU util
@@ -61,8 +70,11 @@ class Netperf(Tool):
         #     initial guess for units per second. A negative value for time will make
         #     heavy use of the system's timestamping functionality
         # -O: Set the remote send,recv buffer offset
-        cmd = (
-            f"-H {server_ip} -p {port} -t {test_name} -n {core_count} -l {seconds}"
+        cmd: str = ""
+        if interface_ip:
+            cmd += f" -L {interface_ip}"
+        cmd += (
+            f" -H {server_ip} -p {port} -t {test_name} -n {core_count} -l {seconds}"
             f" -D {time_unit} -- -O '{send_recv_offset}'"
         )
         process = self.node.execute_async(f"{self.command} {cmd}", sudo=True)
--- a/microsoft/testsuites/performance/common.py
+++ b/microsoft/testsuites/performance/common.py
@@ -207,6 +207,7 @@ def perf_tcp_pps(
     test_type: str,
     server: Optional[RemoteNode] = None,
     client: Optional[RemoteNode] = None,
+    use_internal_address: bool = False,
 ) -> None:
     # Either server and client are set explicitly or we use the first two nodes
     # from the environment. We never combine the two options. We need to specify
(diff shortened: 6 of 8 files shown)


microsoft/testsuites/performance/networkperf.py
```diff
@@ -424,6 +435,568 @@ def perf_sockperf_latency_tcp_synthetic_busy_poll(self, result: TestResult) -> N
             set_busy_poll=True,
         )
 
+    # Network device passthrough tests between host and guest
```
squirrelsc: If no code is reused, move it to a separated file for cloud_hypervisor, like `networkperf_ch.py`.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -424,6 +435,568 @@ def perf_sockperf_latency_tcp_synthetic_busy_poll(self, result: TestResult) -> N
...
+        passthrough_nic_ip = interface_ip.get("INTERFACE_IP", "")
+        assert passthrough_nic_ip, "Can not find interface IP"
+
+        test_node = cast(RemoteNode, node)
+        test_node.internal_address = passthrough_nic_ip
+
+        return test_node, interface_name
+
+    def _get_host_as_server(self, variables: Dict[str, Any]) -> RemoteNode:
+        ip = variables.get("pa_ip", "")
```
squirrelsc: What means "pa" here? it should provide clear names, so it's easy to use. For host operations, it should be wrap as features, so it doesn't need to handle connection from test cases.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -424,6 +435,568 @@ def perf_sockperf_latency_tcp_synthetic_busy_poll(self, result: TestResult) -> N
...
+            server_nic_name=server_nic_name,
+            client_nic_name=client_nic_name,
+            udp_mode=True,
+        )
+
+    def _configure_passthrough_nic_for_node(
+        self,
+        node: Node,
+    ) -> Tuple[RemoteNode, str]:
+        ctx = get_node_context(node)
```
squirrelsc: Avoid to use context in test cases, use features to operate nodes, so it's easy to support multiple platforms, and decouple cases and platforms.
So, use the `NetworkInterface` feature, add a method like `enable_passthrough`.

microsoft/testsuites/performance/networkperf.py
```diff
@@ -424,6 +435,568 @@ def perf_sockperf_latency_tcp_synthetic_busy_poll(self, result: TestResult) -> N
...
+        if not ctx.passthrough_devices:
+            raise SkippedException("No passthrough devices found for node")
+
+        # Configure the nw interface on guest
+        node.execute(
+            cmd="dhclient",
```
squirrelsc: Use the dhclient tool

microsoft/testsuites/performance/networkperf.py
smit-gardhariya: @squirrelsc Agree.

This is the host machine used to measure perf test between VM on one physical server and another physical server NIC. I changed now it to `baremetal_host_` now.

microsoft/testsuites/performance/networkperf.py
smit-gardhariya: @squirrelsc Done

microsoft/testsuites/performance/networkperf.py
smit-gardhariya: @squirrelsc I will name it as `networkperf_passthrough.py` test. As of now we need run it for CH only. But this may extend for QEMU platform too and testcase scenarios remain same.

lisa/tools/netperf.py
```diff
@@ -61,8 +70,11 @@ def run_as_client_async(
...
         # -O: Set the remote send,recv buffer offset
-        cmd = (
-            f"-H {server_ip} -p {port} -t {test_name} -n {core_count} -l {seconds}"
+        cmd: str = ""
+        if interface_ip:
+            cmd += f" -L {interface_ip}"
+        cmd += (
+            f" -H {server_ip} -p {port} -t {test_name} -n {core_count} -l {seconds}"
             f" -D {time_unit} -- -O '{send_recv_offset}'"
         )
```
pupacha, line 79: The new options can be added at the end directly.
```
        cmd = (
          f" -H {server_ip} -p {port} -t {test_name} -n {core_count} -l {seconds}"
          f" -D {time_unit} -- -O '{send_recv_offset}'"
        )
        if interface_ip:
            cmd += f" -L {interface_ip}"

   ```

File: lisa/tools/iperf3.py [hunk 6]
```diff
@@ -135,6 +136,8 @@ def run_as_server_async(
             cmd += f" -f {report_unit} "
         if port:
             cmd += f" -p {port} "
+        if interface_ip:
+            cmd += f" -B {interface_ip}"
```
pupacha, line 140: Could you explain what does `-B` exactly do ? 
Something like `server_ip` is passed from the client while running the iperf tool. What does the new `interface_ip` additional do here ?

microsoft/testsuites/performance/common.py
```diff
@@ -207,6 +207,7 @@ def perf_tcp_pps(
     test_type: str,
     server: Optional[RemoteNode] = None,
     client: Optional[RemoteNode] = None,
+    run_with_internal_address: bool = False,
```
pupacha: `use_internal_address` would be a better name.

microsoft/testsuites/performance/common.py
```diff
@@ -226,6 +227,14 @@ def perf_tcp_pps(
         [lambda: client.tools[Netperf], lambda: server.tools[Netperf]]  # type: ignore
     )
 
+    server_interface_ip: str = ""
+    client_interface_ip: str = ""
+    if run_with_internal_address:
+        server_interface_ip = server.internal_address
+        client_interface_ip = client.internal_address
+        assert server_interface_ip, "Server Node: internal address is not set"
+        assert client_interface_ip, "Client Node: internal address is not set"
```
pupacha: assertion should be done before assignment.
```
        assert server.internal_address, "Server Node: internal address is not set"
        assert client.internal_address, "Client Node: internal address is not set"
        server_interface_ip = server.internal_address
        client_interface_ip = client.internal_address       
```

microsoft/testsuites/performance/common.py
```diff
@@ -446,17 +460,33 @@ def perf_iperf(
...
     udp_mode: bool = False,
+    server: Optional[RemoteNode] = None,
+    client: Optional[RemoteNode] = None,
+    run_with_internal_address: bool = False,
 ) -> None:
-    environment = test_result.environment
-    assert environment, "fail to get environment from testresult"
+    if server is not None or client is not None:
+        assert server is not None, "server need to be specified, if client is set"
+        assert client is not None, "client need to be specified, if server is set"
```
pupacha, line 469: `if server or client` is sufficient. `is not None` is redundant.
Same for the assert statements.

microsoft/testsuites/performance/common.py
```diff
@@ -446,17 +460,33 @@ def perf_iperf(
...
+        assert environment, "fail to get environment from testresult"
+        # set server and client from environment, if not set explicitly
+        client = cast(RemoteNode, environment.nodes[0])
+        server = cast(RemoteNode, environment.nodes[1])
 
-    client = cast(RemoteNode, environment.nodes[0])
-    server = cast(RemoteNode, environment.nodes[1])
     client_iperf3, server_iperf3 = run_in_parallel(
-        [lambda: client.tools[Iperf3], lambda: server.tools[Iperf3]]
+        [lambda: client.tools[Iperf3], lambda: server.tools[Iperf3]]  # type: ignore
```
pupacha, line 478: Why is `type: ignore` needed now ?

microsoft/testsuites/performance/common.py
```diff
@@ -446,17 +460,33 @@ def perf_iperf(
...
     )
     test_case_name = inspect.stack()[1][3]
     iperf3_messages_list: List[Any] = []
+    server_interface_ip = ""
+    client_interface_ip = ""
+    if run_with_internal_address:
+        server_interface_ip = server.internal_address
+        client_interface_ip = client.internal_address
+        assert server_interface_ip, "Server Node: internal address is not set"
+        assert client_interface_ip, "Client Node: internal address is not set"
```
pupacha, line 488: same comment as before - assertions should be before assignment.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+
+@TestSuiteMetadata(
+    area="network passthrough",
+    category="performance",
+    description="""
+    This test suite is to validate linux network performance
+    for various NIC passthrough scenarios.
+    """,
+)
+class NetworkPerformace(TestSuite):
```
pupacha, line 54: There is already a class with same name in `networkperf.py` file. Rename this class to probably `PassthroughNetworkPerformance`

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+
+        test_node = cast(RemoteNode, node)
+        test_node.internal_address = passthrough_nic_ip
+
+        return test_node, interface_name
+
+    def _get_host_as_server(self, variables: Dict[str, Any]) -> RemoteNode:
+        ip = variables.get("baremetal_host_ip", "")
+        username = variables.get("baremetal_host_username", "")
+        passwd = variables.get("baremetal_host_password", "")
```
pupacha, line 570: The variables names should be more specific to the testcase but generic enough (dont mention the host type i.e., baremetal).
May be `nested_passthrough_perf_tests_host_*` or if you can come up with something better.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+
+        if not (ip and username and passwd):
+            raise SkippedException("Server-Node details are not provided")
+
+        server = RemoteNode(
+            runbook=schema.Node(name="baremetal-host"),
+            index=-1,
+            logger_name="baremetal-host",
+            parent_logger=get_logger("baremetal-host-platform"),
```
pupacha, line 579: Same comment here. We should get rid of `baremetal` prefix.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+        """,
+        priority=3,
+        timeout=TIMEOUT,
+        requirement=simple_requirement(
+            min_count=2,
+            supported_platform_type=[CLOUD_HYPERVISOR],
+            unsupported_os=[Windows],
+        ),
+    )
+    def perf_tcp_iperf_passthrough_two_guest(self, result: TestResult) -> None:
```
pupacha, line 289: How about `perf_tcp_iperf_passthrough_guest_guest` ?
This would be similar to the other tests for example, `perf_tcp_iperf_passthrough_host_guest`, where the server type is followed by client type.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+            public_address=ip,
+            public_port=22,
+            username=username,
+            password=passwd,
+        )
+        return server
+
+    def _get_host_nic_name(self, node: RemoteNode) -> str:
+        ip = node.connection_info[constants.ENVIRONMENTS_NODES_REMOTE_ADDRESS]
+        command = "ip route show"
```
pupacha, line 592: Use `Ip` tool that is already available in lisa instead.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+        interface_name: str = ""
+        for match in matches:
+            match_ip = match.get("ip", "").strip()
+            if match_ip == ip:
+                interface_name = match.get("interface", "")
+                break
+        assert interface_name, f"Can't get {ip} interface name from: {routes}"
+        return interface_name
+
+    def after_case(self, log: Logger, **kwargs: Any) -> None:
```
pupacha, line 620: This whole method is similar to the one in `networkperf.py`.
You can create a util function in `common.py` and change both the `after_case` methods to use it.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+            pattern=pattern,
+            lines=device_path,
+        )
+        interface_name = interface_name_raw.get("INTERFACE_NAME", "")
+        assert interface_name, "Can not find interface name"
+
+        # Get the interface ip
+        err_msg = f"Failed to get interface details for: {interface_name}"
+        interface_details = node.execute(
+            cmd=f"ip addr show {interface_name}",
```
pupacha, line 548: Use `Ip` tool here as well.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+        interface_ip = find_group_in_lines(
+            lines=interface_details,
+            pattern=ip_regex,
+            single_line=False,
+        )
+        passthrough_nic_ip = interface_ip.get("INTERFACE_IP", "")
+        assert passthrough_nic_ip, "Can not find interface IP"
+
+        test_node = cast(RemoteNode, node)
+        test_node.internal_address = passthrough_nic_ip
```
pupacha, line 563: My understanding is `internal_address` is a field to store any internal network's IP. For example, on an Azure node, it is IP associated with Vnet and for libvirt platforms, it is IP of virtual net setup that libvirt already has.

A testcase should not change node's `internal_address`. This attribute ownership has to be just with platform.

@squirrelsc , please correct me if my understanding is wrong.

microsoft/testsuites/performance/networkperf.py
smit-gardhariya: Sure, Will make this change

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+    def perf_udp_iperf_passthrough_host_guest(
+        self,
+        node: Node,
+        result: TestResult,
+        variables: Dict[str, Any],
+    ) -> None:
+        server = self._get_host_as_server(variables)
+        client, _ = self._configure_passthrough_nic_for_node(node)
+
+        # Reboot the nodes. Libvirt sometime re-use the nodes.
```
pupacha, line 117: What do you mean `Libvirt sometime re-use the node` ?

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+        # Try to run the test on fresh state of the nodes
+        client.reboot()
+        server.reboot(time_out=1200)
```
pupacha, line 120: There is no change happening to host, from the time the test started, so reboot is not needed here.

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+        description="""
+        This test case uses ntttcp to test tcp network throughput.
+        We will run nttcp server of physical host and client on VM
+        """,
+        priority=3,
+        timeout=TIMEOUT,
+        requirement=node_requirement(
+            node=schema.NodeSpace(
+                node_count=1,
+                memory_mb=search_space.IntRange(min=8192),
```
pupacha, line 214: Why is this a requirement ?

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -44,22 +55,40 @@ def configure_device_passthrough_pool(
...
+                    device_id = vendor_device_id.device_id.strip()
 
-                assert vendor_device_id.device_id.strip()
-                device_id = vendor_device_id.device_id.strip()
+                    self.create_device_pool(
+                        pool_type=config.type,
+                        vendor_id=vendor_id,
+                        device_id=device_id,
+                    )
+                elif isinstance(devices, dict):
```
pupacha, line 80: Should this be `PciAddressIdentifier` instead of `dict` ?

pupacha (changes requested):

microsoft/testsuites/performance/networkperf.py
squirrelsc: It looks like test between host and guest, not other passthrough scenarios. How about networkperf_guest?

smit-gardhariya: @squirrelsc 

Guest VM will have passthrough interface. Host is physical server where we will run server daemon of required tool and client would run with Guest VM with passthrough NIC.

lisa/tools/netperf.py
smit-gardhariya: @pupacha This wont work. Refer nested arg `-- -O '{send_recv_offset}'`

We have to pass it before `-- -O '{send_recv_offset}'`

lisa/tools/iperf3.py
smit-gardhariya: @purna It will run the command and use dedicated interface passed with `-B`

smit-gardhariya: One server might have multiple interface and we want to run with dedicated passthrough NIC IP

microsoft/testsuites/performance/common.py
smit-gardhariya: @pupacha  Sure will make that change.

microsoft/testsuites/performance/common.py
smit-gardhariya: There is default value set to None if we dont pass any. I will make this change

microsoft/testsuites/performance/common.py
smit-gardhariya: @pupacha Agree but it was causing issue with flak8 or mypy validation if i remember correctly. So i put it this way.

microsoft/testsuites/performance/networkperf_passthrough.py
smit-gardhariya: This is internal address of node only. In passthrough testing, node will have 2 interfaces configured. One is virtio-net and other is with vfio. Any case as of now won't use `internal_address `attribute for passthrough.

For Nating usecases just like we have for libvirt platform, we can use it but node is getting initialized there having host detail similar for `internal_address`

microsoft/testsuites/performance/networkperf_passthrough.py
smit-gardhariya: Rebooting the node will create fresh state conditions. I have seen that is requirements are similar, node is getting re-used.

microsoft/testsuites/performance/networkperf_passthrough.py
smit-gardhariya: I have seen some port being occupied issue where we run test with libvirt platform one by one. So, rebooting it will create fresh state of VM.

microsoft/testsuites/performance/networkperf_passthrough.py
smit-gardhariya: I have seen some OOM while running tests. So keeping enough memory for the node

lisa/sut_orchestrator/util/device_pool.py
smit-gardhariya: It will be dict when it runbook loads it. PciAddressIdentifier is getting treated as dict.

microsoft/testsuites/performance/common.py
smit-gardhariya: Done

microsoft/testsuites/performance/common.py
smit-gardhariya: Done

microsoft/testsuites/performance/networkperf_passthrough.py
```diff
@@ -0,0 +1,645 @@
...
+        passthrough_nic_ip = interface_ip.get("INTERFACE_IP", "")
+        assert passthrough_nic_ip, "Can not find interface IP"
+
+        test_node = cast(RemoteNode, node)
+        test_node.internal_address = passthrough_nic_ip
+
+        return test_node, interface_name
+
+    def _get_host_as_server(self, variables: Dict[str, Any]) -> RemoteNode:
+        ip = variables.get("baremetal_host_ip", "")
```
squirrelsc, line 568: It doesn't need to be baremetal, and the name should specify where the variables used. How about `networkperf_peer_xxx`?

microsoft/testsuites/performance/common.py
pupacha: Could you post the error message here?

lisa/sut_orchestrator/util/device_pool.py
pupacha: If thats the case, why use ` isinstance(d, VendorDeviceIdIdentifier)` in line 60 ?

microsoft/testsuites/performance/common.py
pupacha: Please make this change globally in all files.

microsoft/testsuites/performance/networkperf_passthrough.py
pupacha: why would server ports be occupied?
If the tests are doing this, it should be cleanup proper in the after_case section.

microsoft/testsuites/performance/networkperf_passthrough.py
pupacha: I dont think I fully understand it.

TestCase should not change the `internal_address` or any attribute of the node. It should only be done by platform.
There could be other entities like Notifier, Transformer, Feature that could use `internal_address`. I agree that its not used anywhere as of now, but we shouldn't change it.

May be we should figure out another way of storing IP assigned with the passthrough-ed NIC. How about having `PassthroughNIC` as a feature and platform will create the feature based on the passthrough config from runbook.

pupacha: @anirudhrb @gamora12 Any thoughts on this ?

anirudhrb: The feature approach is cleaner. If it works, we should go with that.

gamora12: Agreed, we can have a feature for this.

squirrelsc: It looks like a tool more than a feature. The tool returns passthrough IP, and you can use it for testing. The features are used for platform operations, for example, add a data disk by Azure. The passthrough IP can be got from the VM itself, a tool is enough.

pupacha: Since, the device assignment to VM is already taken care by the platform (with device pool), the only part left is to fetch the IP etc which are all node operations. I agree, `Tool` is more applicable here.

lisa/sut_orchestrator/util/schema.py
```diff
@@ -12,17 +12,26 @@ class HostDevicePoolType(Enum):
...
+class VendorDeviceIdIdentifier:
     vendor_id: str = ""
     device_id: str = ""
 
 
+@dataclass_json()
+@dataclass
+class PciAddressIdentifier:
+    # list of bdf like 0000:3b:00.0 - <domain>:<bus>:<slot>.<fn>
+    pci_bdf: List[str] = field(default_factory=list)
```
anirudhrb, line 24: Identifier should have one BDF. Same as how the VendorDeviceIdIdentifer has one set of (vendor_id, device_id).

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
```diff
@@ -186,24 +186,78 @@ def create_device_pool(
...
+            ):
                 pool = self.available_host_devices.get(pool_type, {})
                 devices = pool.get(iommu_group, [])
-                devices.append(device)
+                if dev not in devices:
+                    devices.append(dev)
                 pool[iommu_group] = devices
                 self.available_host_devices[pool_type] = pool
 
+    def _get_pci_address_instance(
```
anirudhrb, line 246: This should be a constructor in `DeviceAddressSchema`.

Also, `DeviceAddressSchema` can be renamed to just `DeviceAddress`

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -44,22 +55,40 @@ def configure_device_passthrough_pool(
...
-                        "vendor/device id list for given pool type"
-                    )
+                devices = config.devices
+                if isinstance(devices, list) and all(
+                    isinstance(d, VendorDeviceIdIdentifier) for d in devices
+                ):
+                    if len(devices) > 1:
+                        raise LisaException(
+                            "Device Pool does not support more than one "
+                            "vendor/device id list for given pool type"
```
anirudhrb, line 65: Do we support one set of vendor id + device id or one list?

anirudhrb (changes requested):

lisa/sut_orchestrator/util/device_pool.py
```diff
@@ -44,22 +55,40 @@ def configure_device_passthrough_pool(
                         f"Pool type '{pool_type}' is not supported by platform"
                     )
             for config in device_configs:
-                vendor_device_list = config.devices
-                if len(vendor_device_list) > 1:
-                    raise LisaException(
-                        "Device Pool does not support more than one "
-                        "vendor/device id list for given pool type"
-                    )
+                devices = config.devices
```
gamora12, line 58: You can check for the existence of `device_configs` early and return immediately if it doesn't exist to avoid nesting.

gamora12, line 58: You can check for the existence of `device_configs` early and return immediately if it doesn't exist to avoid nesting.

lisa/sut_orchestrator/libvirt/libvirt_device_pool.py
```diff
@@ -186,24 +186,78 @@ def create_device_pool(
...
+    def _create_pool(
+        self,
+        pool_type: HostDevicePoolType,
+        bdf_list: List[str],
+    ) -> None:
+        iommu_grp_of_used_devices = []
+        primary_nic_iommu = self.get_primary_nic_id()
+        for bdf in bdf_list:
+            domain, bus, slot, fn = self._parse_pci_address_str(bdf)
+            dev = self._get_pci_address_instance(domain, bus, slot, fn)
```
gamora12, line 219: Suggestion: Rename `dev` to `device` for uniform naming scheme.

PR #3707, opened 2025-03-11
get security_profile and security_profile when encounter provision issue

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -710,7 +710,14 @@ def _get_node_information(self, node: Node) -> Dict[str, str]:
                     information[key] = value
             except Exception as identifier:
                 node.log.exception(f"error on get {key}.", exc_info=identifier)
-
+        platform_data = platform_runbook.requirement
```
squirrelsc:
```suggestion
        platform_requirement = platform_runbook.requirement
```

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -710,7 +710,14 @@ def _get_node_information(self, node: Node) -> Dict[str, str]:
...
+        if platform_data:
+            information["security_profile"] = self._get_item_fields(
```
squirrelsc: Can it be accessed like features, or feature settings, instead of dict?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -730,7 +737,7 @@ def _get_node_information(self, node: Node) -> Dict[str, str]:
                     "security_profile"
                 ] = security_profile.security_profile.value
                 if security_profile.security_profile == SecurityProfileType.CVM:
-                    information["encrypt_disk"] = security_profile.encrypt_disk
+                    information["security_profile"] = security_profile.encrypt_disk
```
squirrelsc: Does it needs to know where the security_profile from? The log also have those information, why does it need to add into information?

LiliDeng: lisa customer wants to get these info in junit report, any suggestions?

squirrelsc: How about we add an extra switch like `collect_settings` on platform. Once it's enabled, add all features's settings into information.

LiliDeng: we already use capture_vm_information to control this logic, though default value for it is True.

squirrelsc: We should had a better name. For this one, it can be `capture_features_settings`. Once it's enabled, save all feature settings into the information of test results.

lisa/schema.py
```diff
@@ -1450,6 +1450,8 @@ class Platform(TypedSchema, ExtendableSchemaMixin):
     # capture kernel config info or not
     capture_kernel_config_information: bool = False
     capture_vm_information: bool = True
+    # capture features settings
+    capture_features_settings: bool = True
```
squirrelsc, line 1454: Set it to false by default.

lisa/schema.py
```diff
@@ -1450,6 +1450,8 @@ class Platform(TypedSchema, ExtendableSchemaMixin):
     # capture kernel config info or not
     capture_kernel_config_information: bool = False
     capture_vm_information: bool = True
+    # capture features settings
```
squirrelsc, line 1453: This comment is useless. You can explain where the feature settings are captured to.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -699,6 +699,34 @@ def _save_console_log_and_check_panic(
             if check_serial_console is True:
                 check_panic(log_response_content.decode("utf-8"), "provision", log)
 
+    def _get_features_settings(self, node: Node) -> Dict[str, str]:
+        platform_runbook = cast(schema.Platform, self.runbook)
+        information: Dict[str, Any] = {}
+        if platform_runbook.capture_features_settings is False:
+            return information
+        if node.is_connected and node.is_posix:
```
squirrelsc, line 707: Why does it need the two conditions? I think feature settings can be got for any condition.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -699,6 +699,34 @@ def _save_console_log_and_check_panic(
...
+            security_profile = node.features[SecurityProfile].get_settings()
```
squirrelsc, line 708: How about capture all feature settings, instead of only security profile?

squirrelsc:
Sorry, I have another idea. How about we add a method in `TestResult`, called `add_environment_information`? The test case can add environment information as needed. So, the same information won't be added for all test cases in the same run.

LiliDeng:
> How about we add a method in `TestResult`, called `add_environment_information`? The test case can add environment information as needed. So, the same information won't be added for all test cases in the same run.

Is below one similar to what your idea? How do we control which case is enabled or which one is not enabled?

```
    @property
    def environment_information(self) -> Dict[str, Any]:
        # It's used by subtests.
        if not self._environment_information and self.environment:
            self._environment_information = self.environment.get_information(
                force_run=False
            )

        return self._environment_information
```

squirrelsc:
> environment_information(self) -> Dict

It's not perfect, but we can go with this way.

PR #3708, opened 2025-03-12
Added properties to fdisk to prevent install issue.

r-dailey:
To handle the issue that lists a failure to install fdisk on freebsd I have added the required properties to the bsdfdisk class that will prevent this. The affected tests will still fail due to other issues with storage management and a lack of tooling for setting up raid. That is an issue I am currently working on but it shouldn't prevent this fix from going in.


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3711, opened 2025-03-14
adding BSDTaskSet tool

r-dailey:
created a BSDTaskSet tool to prevent task set issues on arm64 vms. The test case now fails with the interrupt inspector failure that the other images face.

### DIFF
--- a/lisa/tools/taskset.py
+++ b/lisa/tools/taskset.py
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
+from typing import Optional, Type
+
 from lisa.executable import Tool
 from lisa.util.process import Process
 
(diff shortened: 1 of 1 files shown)


squirrelsc:
@LiliDeng LGTM

lisa/tools/taskset.py
```diff
@@ -10,9 +12,22 @@ class TaskSet(Tool):
...
         return self.run_async(f"-c {cpu_id} yes > /dev/null", shell=True)
+
+
+class BSDTaskSet(TaskSet):
+    @property
+    def command(self) -> str:
+        return "cpuset"
+
+    def run_on_specific_cpu(self, cpu_id: int) -> Process:
+        return self.run_async(f"-l {cpu_id} yes > /dev/null", shell=True)
```
LiliDeng, line 33: if this method is the same, we can reuse above one?

r-dailey: The method I wrote uses a different flag from the above one. I could rewrite it so that the above method checks the OS and dynamically changes the flag but I think this is a cleaner and more easily readable format.

LiliDeng (approved):

PR #3713, opened 2025-03-14
Improve kselftest test support

paulli2017:
Improve kselftest test support to cover some SOC platform which supports only kselftest tar file

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest-suite.py
+++ b/microsoft/testsuites/kselftest/kselftest-suite.py
@@ -48,11 +48,14 @@ class KselftestTestsuite(TestSuite):
         result: TestResult,
     ) -> None:
         file_path = variables.get("kselftest_file_path", "")
+        working_path = variables.get("kselftest_working_path", "")
+        run_as_root = variables.get("kselftest_run_as_root", False)
         try:
             kselftest: Kselftest = node.tools.get(
                 Kselftest,
-                kselftest_file_path=file_path,
+                working_path=working_path,
+                file_path=file_path,
             )
-            kselftest.run_all(result, log_path, self._KSELF_TIMEOUT)
+            kselftest.run_all(result, log_path, self._KSELF_TIMEOUT, run_as_root)
         except UnsupportedDistroException as identifier:
             raise SkippedException(identifier)
--- a/microsoft/testsuites/kselftest/kselftest.py
+++ b/microsoft/testsuites/kselftest/kselftest.py
@@ -87,38 +87,49 @@ class Kselftest(Tool):
         return True
 
     def _check_exists(self) -> bool:
-        return (
-            len(self.node.tools[Ls].list(str(self._kself_installed_dir), sudo=True)) > 0
-        )
+        return len(self.node.tools[Ls].list(str(self._installed_path), sudo=True)) > 0
 
     def __init__(
-        self, node: Node, kselftest_file_path: str, *args: Any, **kwargs: Any
+        self,
+        node: Node,
+        working_path: str,
+        file_path: str,
+        *args: Any,
+        **kwargs: Any,
     ) -> None:
         super().__init__(node, *args, **kwargs)
 
         # tar file path specified in yml
-        self._tar_file_path = kselftest_file_path
+        self._working_path = working_path
+        self._tar_file_path = file_path
+        kselftest_packages = "kselftest-packages"
+        if self._working_path:
+            package_path = self.node.get_pure_path(
+                self._working_path,
+            ) / (kselftest_packages)
+        else:
+            package_path = self.get_tool_path(use_global=True) / kselftest_packages
+
+        self._installed_path = package_path
+
         if self._tar_file_path:
-            self._remote_tar_path = self.get_tool_path(
-                use_global=True
-            ) / os.path.basename(self._tar_file_path)
+            self._remote_tar_path = package_path / os.path.basename(self._tar_file_path)
 
-        # command to run kselftests
-        self._kself_installed_dir = (
-            self.get_tool_path(use_global=True) / "kselftest-packages"
-        )
-
-        self._command = self._kself_installed_dir / "run_kselftest.sh"
+        self._command = self._installed_path / "run_kselftest.sh"
 
     # install common dependencies
     def _install(self) -> bool:
-        if not (
+        is_support = False
+        if (
             (
                 isinstance(self.node.os, Ubuntu)
                 and self.node.os.information.version >= "18.4.0"
             )
             or isinstance(self.node.os, CBLMariner)
+            or (self._tar_file_path and self._working_path)
         ):
+            is_support = True
+        if not is_support:
             raise UnsupportedDistroException(
                 self.node.os, "kselftests in LISA does not support this os"
             )
@@ -128,12 +139,12 @@ class Kselftest(Tool):
             mkdir.create_directory(self._remote_tar_path.parent.as_posix())
             self.node.shell.copy(PurePath(self._tar_file_path), self._remote_tar_path)
             self.node.tools[Tar].extract(
-                str(self._remote_tar_path), str(self._kself_installed_dir)
+                str(self._remote_tar_path), str(self._installed_path)
             )
             self._log.debug(f"Extracted tar from path {self._remote_tar_path}!")
         else:
             mkdir = self.node.tools[Mkdir]
-            mkdir.create_directory(self._kself_installed_dir.as_posix())
+            mkdir.create_directory(self._installed_path.as_posix())
             if isinstance(self.node.os, Ubuntu):
                 arch = self.node.os.get_kernel_information().hardware_platform
                 if arch == "aarch64":
@@ -183,7 +194,7 @@ class Kselftest(Tool):
 
             # build and install kselftests
             self.node.execute(
-                cmd=f"./kselftest_install.sh {self._kself_installed_dir}",
+                cmd=f"./kselftest_install.sh {self._installed_path}",
                 shell=True,
                 cwd=PurePosixPath(kernel_path, "tools/testing/selftests"),
                 sudo=True,
@@ -193,7 +204,7 @@ class Kselftest(Tool):
             # change permissions of kselftest-packages directory
             # to run test as non root user.
             chmod = self.node.tools[Chmod]
-            chmod.update_folder(self._kself_installed_dir.as_posix(), "777", sudo=True)
+            chmod.update_folder(self._installed_path.as_posix(), "777", sudo=True)
 
         return self._check_exists()
 
@@ -210,6 +221,8 @@ class Kselftest(Tool):
         # get username
         username = self.node.tools[Whoami].get_username()
         result_directory = f"/home/{username}"
+        if self._working_path:
+            result_directory = self._working_path
         if os.path.exists(result_directory) is False:
             mkdir = self.node.tools[Mkdir]
             mkdir.create_directory(result_directory)
--- a/docs/docker_linux.rst
+++ b/docs/docker_linux.rst
@@ -31,6 +31,21 @@ After installing Docker, you can start the Docker service with the following com
 
    sudo systemctl start docker
 
+Managing Docker Permissions
+--------------------
+
+Add the current user to the "docker" group to run Docker without sudo.
+
+.. code:: bash
+
+   sudo usermod -aG docker $USER
+
+Apply the group change immediately wihout logging out and logging back in.
+
+.. code:: bash
+
+   newgrp docker
+
 Launch LISA container
 -----------------------
 
@@ -38,4 +53,4 @@ Use below command to launch the LISA container.
 
 .. code:: bash
 
-   sudo docker run --rm -i mcr.microsoft.com/lisa/runtime:latest lisa -r ./examples/runbook/hello_world.yml
+   docker run --rm -i mcr.microsoft.com/lisa/runtime:latest lisa -r ./examples/runbook/hello_world.yml
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -6,8 +6,8 @@ Currently we support installation on Linux & Windows.
 .. toctree::
     :maxdepth: 2
 
-    Use docker images on Linux. <docker_linux>
-    Use docker images on Windows. <docker_windows>
+    Use docker images on Linux <docker_linux>
+    Use docker images on Windows <docker_windows>
     Installation on Linux <installation_linux>
     Installation on Windows <installation_windows>
     Updating LISA <update>
--- a/docs/quick_start.rst
+++ b/docs/quick_start.rst
@@ -22,6 +22,8 @@ Installation
 
 Follow the installation guide below to install **LISA** on your OS.
 
+-  :doc:`Use docker image on Linux <docker_linux>`
+-  :doc:`Use docker image on Windows <docker_windows>`
 -  :doc:`Installation on Linux <installation_linux>`
 -  :doc:`Installation on Windows <installation_windows>`
 
--- a/docs/run_test/platform.rst
+++ b/docs/run_test/platform.rst
@@ -139,11 +139,13 @@ deployment.
    platform:
    - type: azure
       ...
+      admin_private_key_file: "<path of private key file>"
       azure:
          virtual_network_resource_group: $(vnet_resource_group)
          virtual_network_name: $(vnet_name)
          subnet_prefix: $(subnet_name)
          use_public_address: "<true or false>"
+         use_ipv6: "<true or false>"
       requirement:
          ...
          ignored_capability:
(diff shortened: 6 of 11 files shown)


microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -98,26 +108,44 @@ def __init__(
...
+        if (
+            (isinstance(self.node.os, Posix) and not isinstance(self.node.os, Linux))
+            or (
+                isinstance(self.node.os, Debian)
+                and not isinstance(self.node.os, Ubuntu)
+            )
+            or (
                 isinstance(self.node.os, Ubuntu)
-                and self.node.os.information.version >= "18.4.0"
+                and self.node.os.information.version < "18.4.0"
```
squirrelsc: Why change Ubuntu version from `>=` to `<`?

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -98,26 +108,44 @@ def __init__(
...
+            ) / (kselftest_packages)
 
         self._command = self._kself_installed_dir / "run_kselftest.sh"
 
     # install common dependencies
     def _install(self) -> bool:
-        if not (
-            (
+        if (
+            (isinstance(self.node.os, Posix) and not isinstance(self.node.os, Linux))
```
squirrelsc: The conditions are complicated, change it to one by one like,
```python
if isinstance(...):
    is_supported = True
elif isinstance(...):
    is_supported = True

if not is_supported:
    raise raise UnsupportedDistroException(
                self.node.os, ...
```

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -98,26 +108,44 @@ def __init__(
...
+        kselftest_packages = "kselftest-packages"
         if self._tar_file_path:
+            self._remote_tar_path = self.node.get_pure_path(
+                kselftest_install_path,
+            ) / os.path.basename(self._tar_file_path)
+
+            self._kself_installed_dir = self.node.get_pure_path(
+                kselftest_install_path,
+            ) / (kselftest_packages)
+        else:
```
squirrelsc, line 110: Kselftest supports runtime build currently. With this change, is the runtime build broken?

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -98,26 +108,44 @@ def __init__(
...
             self._remote_tar_path = self.get_tool_path(
-                use_global=True
+                use_global=True,
             ) / os.path.basename(self._tar_file_path)
```
squirrelsc: There is a but, the `self._tar_file_path` doesn't exists from above `if self._tar_file_path`, but it's used here.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -202,14 +230,12 @@ def run_all(
         test_result: TestResult,
         log_path: str,
         timeout: int = 5000,
-        run_test_as_root: bool = False,
```
squirrelsc: Don't remove existing functionality, it's used in other scenarios.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -202,14 +230,12 @@ def run_all(
...
     ) -> List[KselftestResult]:
         # Executing kselftest as root may cause
         # VM to hang
 
-        # get username
-        username = self.node.tools[Whoami].get_username()
```
squirrelsc, line 211: Use tool to expand the $HOME, so the $HOME folder works for both users and root.

microsoft/testsuites/kselftest/kselftest.py
paulli2017: the condition used to be not >=, now change to be positive, another way of expression

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks for suggestion

microsoft/testsuites/kselftest/kselftest.py
paulli2017: no, else situation is the previous path

microsoft/testsuites/kselftest/kselftest.py
paulli2017: actually, this change is done based on the suggestion as the old PR that's not merged. I can restore it back if that's the case

microsoft/testsuites/kselftest/kselftest.py
paulli2017: if self._tar_file_path is "", does os.path.basename(self._tar_file_path) return empty too and it should be ok? This is the actually old path and it seems that there is no report issue here based on pipeline run results before.

microsoft/testsuites/kselftest/kselftest.py
paulli2017: right now, the result directory is moved to /tmp directory instead of $HOME/{username} as soc system doesn't allow write on that directory,

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -98,27 +108,48 @@ def __init__(
...
+        elif (
+            isinstance(self.node.os, Ubuntu)
+            and self.node.os.information.version < "18.4.0"
+        ):
+            is_support = False
+        elif isinstance(self.node.os, RPMDistro) and not isinstance(
+            self.node.os, CBLMariner
         ):
+            is_support = False
+        elif isinstance(self.node.os, (Suse, NixOS, OtherLinux)):
```
squirrelsc: Please have the whole list as `is_supported = True`. Because we may add new distro, new versions in future. An allowed list is easier to manage than a denied list.

squirrelsc: You can have other conditions than only distro version. For example, if there is a kselftest binary, it's always supported.

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: I cannot get `else situation is the previous path` from the code. The original code didn't have an else path.

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: > if self._tar_file_path is "", does os.path.basename(self._tar_file_path) return empty too and it should be ok? This is the actually old path and it seems that there is no report issue here based on pipeline run results before.

It reduces readability on code like `os.path.basename("")`, and unnecessary. From my search, the `self._tar_file_path` is not used in the original code, when it's empty.

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: Please avoid creating new PR, because it lost context. What's the reason we want to hard code sudo?

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: Every user should have a $HOME folder. It's either `/home/{username}` or `/root` for root user. $HOME is variable, and use `node.expand_env_path` to get the expanded path. Please have a test.

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks for your suggestion.

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -92,33 +92,53 @@ def _check_exists(self) -> bool:
...
+            if self._kselftest_working_dir:
+                self._remote_tar_path = self.node.get_pure_path(
+                    self._kselftest_working_dir,
+                ) / os.path.basename(self._tar_file_path)
+            else:
+                self._remote_tar_path = self.get_tool_path(
+                    use_global=True
+                ) / os.path.basename(self._tar_file_path)
+
+        if self._kselftest_working_dir:
```
squirrelsc, line 118:
```python
if not self._kselftest_working_dir:
    self._kselftest_working_dir = self.get_tool_path(use_global=True)

self._kself_installed_dir = self.node.get_pure_path(
                self._kselftest_working_dir,
            ) / (kselftest_packages)
```

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -92,33 +92,53 @@ def _check_exists(self) -> bool:
...
         if self._tar_file_path:
-            self._remote_tar_path = self.get_tool_path(
-                use_global=True
-            ) / os.path.basename(self._tar_file_path)
-
-        # command to run kselftests
-        self._kself_installed_dir = (
-            self.get_tool_path(use_global=True) / "kselftest-packages"
-        )
+            if self._kselftest_working_dir:
```
squirrelsc:
```python
if kselftest_working_path:
    self._kselftest_working_dir = self.node.get_pure_path(self._kselftest_working_dir)
else:
    self._kselftest_working_dir = self.get_tool_path(use_global=True)

self._kself_installed_dir = self._kselftest_working_dir / (kselftest_packages)

if self._tar_file_path:
    self._remote_tar_path = self._kselftest_working_dir / os.path.basename(self._tar_file_path)
```

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -92,33 +92,53 @@ def _check_exists(self) -> bool:
...
+        node: Node,
+        kselftest_working_path: str,
+        kselftest_file_path: str,
+        *args: Any,
+        **kwargs: Any,
     ) -> None:
         super().__init__(node, *args, **kwargs)
 
         # tar file path specified in yml
+        self._kselftest_working_dir = kselftest_working_path
```
squirrelsc: `self._kselftest_working_dir` should be like `self._kselftest_working_path`. Don't mix use of path and dir.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -48,9 +48,11 @@ def verify_kselftest(
         result: TestResult,
     ) -> None:
         file_path = variables.get("kselftest_file_path", "")
+        working_dir = variables.get("kselftest_working_path", "")
```
squirrelsc: `working_dir` to `working_path`. consistent on names.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -92,33 +92,50 @@ def _check_exists(self) -> bool:
         )
 
     def __init__(
-        self, node: Node, kselftest_file_path: str, *args: Any, **kwargs: Any
+        self,
+        node: Node,
+        kselftest_working_path: str,
```
squirrelsc: `working_path` and `file_path`. It's in the kselftest class already, it doesn't need to have the name again.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -92,33 +92,50 @@ def _check_exists(self) -> bool:
...
+        kselftest_file_path: str,
+        *args: Any,
+        **kwargs: Any,
     ) -> None:
         super().__init__(node, *args, **kwargs)
 
         # tar file path specified in yml
+        self._kselftest_working_path = kselftest_working_path
```
squirrelsc: Same as above, self._working_path

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -92,33 +92,50 @@ def _check_exists(self) -> bool:
...
         self._tar_file_path = kselftest_file_path
-        if self._tar_file_path:
-            self._remote_tar_path = self.get_tool_path(
-                use_global=True
-            ) / os.path.basename(self._tar_file_path)
+        kselftest_packages = "kselftest-packages"
+        if self._kselftest_working_path:
+            self._kselftest_working_dir = self.node.get_pure_path(
```
squirrelsc: 1. self._kselftest_working_dir is used in this method only, it doesn't need to be a class level variable.
1. the name is confusing, call it like package_path.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -92,33 +92,50 @@ def _check_exists(self) -> bool:
...
+        else:
+            self._kselftest_working_dir = (
+                self.get_tool_path(use_global=True) / kselftest_packages
+            )
 
-        # command to run kselftests
-        self._kself_installed_dir = (
-            self.get_tool_path(use_global=True) / "kselftest-packages"
-        )
+        self._kself_installed_dir = self._kselftest_working_dir / (kselftest_packages)
```
squirrelsc: self._installed_path

microsoft/testsuites/kselftest/kselftest-suite.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: done, thanks

squirrelsc: _installed_path, not `dir`.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -218,7 +231,7 @@ def run_all(
         result_file = f"{result_directory}/{result_file_name}"
         self.run(
             f" 2>&1 | tee {result_file}",
-            sudo=run_test_as_root,
+            sudo=run_test_as_root if not self._working_path else True,
```
squirrelsc: Why make this change? It's not easy to understand.

paulli2017: For soc system, it has to be root to run the test.

squirrelsc: Use a test variable to control it. The working_path is not designed to be always non sudo. They are separated settings.

microsoft/testsuites/kselftest/kselftest.py
paulli2017: done, thanks

microsoft/testsuites/kselftest/kselftest.py
paulli2017: The change is done, thanks

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -48,10 +48,14 @@ def verify_kselftest(
         result: TestResult,
     ) -> None:
         file_path = variables.get("kselftest_file_path", "")
+        working_path = variables.get("kselftest_working_path", "")
+        test_user = variables.get("kselftest_user", "")
```
squirrelsc: kselftest_user -> kselftest_run_as_root, and set the default to True.

```suggestion
        run_as_root = variables.get("kselftest_run_as_root", True)
```

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -48,10 +48,14 @@ def verify_kselftest(
...
+        run_as_root = variables.get("kselftest_run_as_root", "")
         try:
             kselftest: Kselftest = node.tools.get(
                 Kselftest,
-                kselftest_file_path=file_path,
+                working_path=working_path,
+                file_path=file_path,
+                run_as_root=run_as_root,
             )
             kselftest.run_all(result, log_path, self._KSELF_TIMEOUT)
```
squirrelsc: Sorry, I didn't check the logic. But you can use the run_as_root here. So the variable doesn't need to pass into the kselftest class.

microsoft/testsuites/kselftest/kselftest-suite.py
paulli2017: based on the original code, the default should be False, right?

squirrelsc: Yes, follow existing logic. I just saw it in your tests it's true. So I thought the default value is true.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -87,38 +87,50 @@ def can_install(self) -> bool:
...
-        )
+        return len(self.node.tools[Ls].list(str(self._installed_path), sudo=True)) > 0
 
     def __init__(
-        self, node: Node, kselftest_file_path: str, *args: Any, **kwargs: Any
+        self,
+        node: Node,
+        working_path: str,
+        file_path: str,
+        run_as_root: str,
```
squirrelsc: remove it

paulli2017: done, thanks

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3719, opened 2025-03-19
UpgradeInstaller: Retry package installation

kamalca:
VM extensions (or other causes) can result in the package manager lock being busy. Retrying the operation allows opportunities for other package manager operations to finish so the lock can be acquired.

### DIFF
--- a/lisa/transformers/upgrade_packages.py
+++ b/lisa/transformers/upgrade_packages.py
@@ -188,6 +188,7 @@ class UnattendedUpgradeInstaller(UpgradeInstaller):
         )
         upgradable_before = [package.strip() for package in result.stdout.split("\n")]
 
+        node.os.wait_running_package_process()
         node.execute(
             "unattended-upgrade -d -v",
             sudo=True,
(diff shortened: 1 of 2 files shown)


squirrelsc:
> VM extensions (or other causes) can result in the package manager lock being busy. Retrying the operation allows opportunities for other package manager operations to finish so the lock can be acquired.

Debian has wait_running_package_process, it should solve the package installation problem. Do you find similar problems on other distros?

kamalca:
> Debian has wait_running_package_process, it should solve the package installation problem.

Thank you for the tip. I have updated my changes to use this function.

> Do you find similar problems on other distros?

Package validation is only on Debian-based images at this time. When we expand to other distros I will monitor for any installation issues.

lisa/transformers/upgrade_packages.py
```diff
@@ -280,6 +282,7 @@ def _update_packages(self) -> List[str]:
         )
         upgradable_before = [package.strip() for package in result.stdout.split("\n")]
 
+        node.os.wait_running_package_process()
```
squirrelsc: Add it in the os class.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3720, opened 2025-03-20
Add ability to pass a SKU capability file along with VM SKUs

sharsonia:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -2098,6 +2098,47 @@ def _process_marketplace_image_plan(
...
 
+    def _generate_sku_capability(
+        self,
+        vm_size: str,
+        location: str,
+        cap_file: str,
+    ) -> AzureCapability:
+        # some vm size cannot be queried from API, and the capability
+        # may be queried from capability files through hooks.
+        capability_dict = plugin_manager.hook.azure_add_sku_capability(
```
kamalca, line 2109: Is this supposed to match the hookspec `azure_update_sku_capability`? Or is this a different hook?

squirrelsc: I guess they should be the same.

PR #3721, opened 2025-03-20
Redhat GPU Suite: Update prerequiste package

adityagesh:
Remove kernel version from packages "kernel-devel" and "kernel-headers"


adityagesh:
Closing for now as end to end flow cannot be verified due to a bug in nvidia cuda driver package installation.

PR #3723, opened 2025-03-20
Added file type support for build copy() implementation

paulli2017:
Added file type support for build copy() implementation

### DIFF
--- a/lisa/sut_orchestrator/baremetal/build.py
+++ b/lisa/sut_orchestrator/baremetal/build.py
@@ -24,12 +24,20 @@ def _find_matched_files(
     all_files = []
     match_files: Dict[str, FileSchema] = {}
     for source_path in sources_path:
-        for root, _, files in os.walk(source_path):
-            for file in files:
-                all_files.append(os.path.join(root, file))
+        is_source_dir = os.path.isdir(source_path)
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/baremetal/build.py
```diff
@@ -24,12 +24,20 @@ def _find_matched_files(
     all_files = []
     match_files: Dict[str, FileSchema] = {}
     for source_path in sources_path:
-        for root, _, files in os.walk(source_path):
-            for file in files:
-                all_files.append(os.path.join(root, file))
+        is_source_dir = os.path.isdir(source_path)
+        if not is_source_dir:
```
squirrelsc: Use positive check, if possible.

```python
        if is_source_dir:
            for root, _, files in os.walk(source_path):
                for file in files:
                    all_files.append(os.path.join(root, file))
        else:
            all_files.append(os.path.abspath(source_path))
```

paulli2017: The change is done, thanks

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3727, opened 2025-03-21
Install Azure CLI on a Linux release image in a Docker container for Azure authentication.

LiliDeng:


Dockerfile.linux
```diff
@@ -19,6 +19,7 @@ RUN tdnf update -y && \
         curl \
         wget \
         tar \
+        azure-cli \
```
squirrelsc, line 22: Is it in the windows docker too? If not, it should be added too.

squirrelsc (approved):

PR #3728, opened 2025-03-21
VC test result improvements

squirrelsc:
Enable perf dump json for VC messages, and fill in perf message fields for VC messages.

### DIFF
--- a/lisa/tools/virtualclient.py
+++ b/lisa/tools/virtualclient.py
@@ -251,7 +251,6 @@ class VirtualClientTool(Tool):
     def send_metrics_message(self, metrics_file: Path) -> None:
         with open(metrics_file, mode="r", newline="") as csvfile:
             reader = csv.DictReader(csvfile)
-            metric_messages: List[VCMetrics] = []
             for row in reader:
                 metric = VCMetrics(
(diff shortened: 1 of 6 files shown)


lisa/tools/virtualclient.py
```diff
@@ -248,40 +248,53 @@ def download_raw_data_file(self) -> Path:
...
-                    operating_system_platform=row["OperatingSystemPlatform"],
-                    operation_id=row["OperationId"],
-                    operation_parent_id=row["OperationParentId"],
-                    app_host=row["AppHost"],
-                    app_name=row["AppName"],
-                    app_version=row["AppVersion"],
-                    app_telemetry_version=row["AppTelemetryVersion"],
-                    tags=row["Tags"],
+                fields = {
+                    "time": self._parse_timestamp(row["Timestamp"]),
```
LiliDeng, line 261: time is duplicated with time_stamp

squirrelsc: I thought to keep vc fields. But I will remove vc fields, if they are defined by us already.

LiliDeng (approved):

PR #3729, opened 2025-03-22
move env vars out to AzureDefaultCredential in credential.py

paxue:
In credential.py, Move AZURE_CLIENT_ID, AZURE_TENANT_ID to AzureDefaultCredential

### DIFF
--- a/lisa/sut_orchestrator/azure/credential.py
+++ b/lisa/sut_orchestrator/azure/credential.py
@@ -113,13 +113,17 @@ class AzureCredential(subclasses.BaseClassWithRunbookMixin):
         if runbook.tenant_id:
             self._tenant_id = runbook.tenant_id
             self._log.debug(f"Use defined tenant id: {self._tenant_id}")
-            os.environ["AZURE_TENANT_ID"] = self._tenant_id
         if runbook.client_id:
             self._client_id = runbook.client_id
             self._log.debug(f"Use defined client id: {self._client_id}")
-            os.environ["AZURE_CLIENT_ID"] = self._client_id
 
         self._allow_all_tenants = runbook.allow_all_tenants
+
+    def _set_auth_env_variables(self) -> None:
+        if self._tenant_id:
+            os.environ["AZURE_TENANT_ID"] = self._tenant_id
+        if self._client_id:
+            os.environ["AZURE_CLIENT_ID"] = self._client_id
 
     def __hash__(self) -> int:
         return hash(self._get_key())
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/credential.py
```diff
@@ -146,6 +144,19 @@ def type_name(cls) -> str:
     def type_schema(cls) -> Type[schema.TypedSchema]:
         return AzureCredentialSchema
 
+    def __init__(
```
squirrelsc, line 153: Please move it to a method, and call it in AzureDefaultCredential, ClientSecretCredential and the CertificationCredential. Because they depend on env variables.

lisa/sut_orchestrator/azure/credential.py
```diff
@@ -113,14 +113,18 @@ def __init__(
...
             self._log.debug(f"Use defined tenant id: {self._tenant_id}")
-            os.environ["AZURE_TENANT_ID"] = self._tenant_id
         if runbook.client_id:
             self._client_id = runbook.client_id
             self._log.debug(f"Use defined client id: {self._client_id}")
-            os.environ["AZURE_CLIENT_ID"] = self._client_id
 
         self._allow_all_tenants = runbook.allow_all_tenants
 
+    def _set_envvar_tenand_client_id(self) -> None:
```
squirrelsc: `_set_auth_env_variables`

lisa/sut_orchestrator/azure/credential.py
```diff
@@ -146,6 +150,16 @@ def type_name(cls) -> str:
...
 
+    def __init__(
+        self,
+        runbook: AzureCredentialSchema,
+        logger: Logger,
+        cloud: Cloud = AZURE_PUBLIC_CLOUD,
+    ) -> None:
+        super().__init__(runbook, logger=logger, cloud=cloud)
+        self._set_envvar_tenand_client_id()
+        self._credential_type = AzureCredentialType.DefaultAzureCredential
```
squirrelsc: Does this line need? It should be set by LISA already, or it's not useful.

paxue: For here it's not needed, can be removed. I added it just for the easy reading and keep consistent as other credential type.

squirrelsc: Please help removing others, it looks the base class handled it already.

paxue: Sure.

lisa/sut_orchestrator/azure/credential.py
```diff
@@ -176,6 +189,14 @@ def type_name(cls) -> str:
     def type_schema(cls) -> Type[schema.TypedSchema]:
         return AzureCredentialSchema
 
+    def __init__(
```
squirrelsc: It's not needed, because no modification on base __init__

paxue: true, removed

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3730, opened 2025-03-24
add doc for using different authentication methods on azure

LiliDeng:


docs/run_test/azure_auth.rst
```diff
@@ -0,0 +1,164 @@
...
+.. code:: yaml
+
+   platform:
+     - type: azure
+       azure:
+         credential:
+           type: certificate
+           tenant_id: <tenant id> # Required
+           client_id: <client id> # Required
+           certificate_file: <path to certificate file> # Required
```
squirrelsc: Some fields are inconsistent with the actual value. Please double check them.

docs/run_test/azure_auth.rst
```diff
@@ -0,0 +1,164 @@
...
+.. code:: yaml
+
+   platform:
+     - type: azure
+       azure:
+         credential:
+           type: workloadidentity
+           client_id: <client id> # Required
+           tenant_id: <tenant id> # Required
+           allow_all_tenants: false | true # Optional. Default is `false`.
```
squirrelsc: it's in the base schema, not only for workload identity.

docs/run_test/azure_auth.rst
```diff
@@ -0,0 +1,164 @@
...
+
+Example:
+.. code:: yaml
+
+   platform:
+     - type: azure
+       azure:
+         credential:
+           type: assertion
+           tenant_id: <tenant id> # Required
```
squirrelsc, line 83: It still has the client_id, because the client_id is in base class.

LiliDeng: client id is not used in this authentication way

docs/run_test/azure_auth.rst
```diff
@@ -0,0 +1,151 @@
...
+           client_id: <client id>  # Optional
+           tenant_id: <tenant id>  # Optional
+           allow_all_tenants: false | true  # Optional. Default is `false`.
+
+* **type**: `default` indicates default credential authentication.
+* **client_id**: (Optional) Needed when there are multiple managed identities associated with the running machine.
+* **tenant_id**: (Optional) Needed when you have multiple tenants.
+* **allow_all_tenants**: (Optional) Specifies whether to allow cross-tenant authorization. Default is `false`.
+
+**Note:** The `default` authentication type defines a base schema, which is inherited by all other authentication methods. Fields such as `allow_all_tenants` are applicable to all authentication types.
```
squirrelsc: default is not the base class. Default is to try all auth method as best effort. The default is not related with other auth methods. For `allow_all_tenants`, it needs to list in other classes, if a class uses the field.

squirrelsc (approved):

PR #3732, opened 2025-03-25
Adding a Kself Lite Test Case

umfranci:
Currently the 'verify_kselftest' Test Case executes all the tests and test suites which are a part of the Linux KSelf tests.

Although we wanted to create a lighter version, where there would be a predefined list of the most critical test suites. Also, users would be able to pass the test suites they want to execute and the specific test cases that need to be skipped.

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest-suite.py
+++ b/microsoft/testsuites/kselftest/kselftest-suite.py
@@ -18,6 +18,17 @@ class KselftestTestsuite(TestSuite):
     # timeout below is in seconds and set to 2 hours.
     _CASE_TIME_OUT = 7200
     _KSELF_TIMEOUT = 6700
+    # the default list of tests executed as part of kself lite
+    _KSELF_LITE_TESTS = [
+        "bpf",
+        "core",
+        "futex",
+        "ipc",
+        "kvm",
+        "mm",
+        "net",
+        "timers",
+    ]
 
     @TestCaseMetadata(
         description="""
@@ -57,5 +68,68 @@ class KselftestTestsuite(TestSuite):
                 file_path=file_path,
             )
             kselftest.run_all(result, log_path, self._KSELF_TIMEOUT, run_as_root)
+        except UnsupportedDistroException as identifier:
+            raise SkippedException(identifier)
+
+    @TestCaseMetadata(
+        description="""
+        This test case will run kself lite tests - a subset of kselftests.
+        Cases:
+        1. When a tarball is specified in .yml file, extract the tar and run kselftests.
+        Example:
+        - name: kselftest_file_path
+          value: <path_to_kselftests.tar.xz>
+          is_case_visible: true
+        2. When a tarball is not specified in .yml file, clone Mariner kernel,
+        copy current config to .config, build kselftests and generate a tar.
+        For both cases, verify that the kselftest tool extracts the tar, runs the script
+        run_kselftest.sh and redirects test results to a file kselftest-results.txt.
+        3. For this lite version, the list of collections to run and the list
+        of tests to skip can be specified in the .yml file.
+        Example:
+        - name: kself_skip_tests
+          value: <comma_separated_list_of_tests_to_skip - format: collection:test_name>
+          is_case_visible: true
+        - name: kself_test_collection
+          value: <comma_separated_list_of_collections_to_run - format: collection_name>
+          is_case_visible: true
+        """,
+        priority=3,
+        timeout=_CASE_TIME_OUT,
+        requirement=simple_requirement(
+            min_core_count=16,
+        ),
+    )
+    def verify_kself_lite(
+        self,
+        node: Node,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        file_path = variables.get("kselftest_file_path", "")
+        skip_tests = variables.get("kself_skip_tests", "")
+        test_collection = variables.get("kself_test_collection", "")
+        # get comma separated list of tests
+        if test_collection:
+            test_collection_list = test_collection.split(",")
+        else:
+            test_collection_list = self._KSELF_LITE_TESTS
+        if skip_tests:
+            skip_tests_list = skip_tests.split(",")
+        else:
+            skip_tests_list = []
+        try:
+            kselftest: Kselftest = node.tools.get(
+                Kselftest,
+                kselftest_file_path=file_path,
+            )
+            kselftest.run_all(
+                test_result=result,
+                log_path=log_path,
+                timeout=self._KSELF_TIMEOUT,
+                run_collections=test_collection_list,
+                skip_tests=skip_tests_list,
+            )
         except UnsupportedDistroException as identifier:
             raise SkippedException(identifier)
--- a/microsoft/testsuites/kselftest/kselftest-suite.py
+++ b/microsoft/testsuites/kselftest/kselftest-suite.py
@@ -71,28 +71,38 @@ class KselftestTestsuite(TestSuite):
 
     @TestCaseMetadata(
         description="""
-        This test case will run kself lite tests - a subset of kselftests.
-        A predefined list of tests is provided in _KSELF_LITE_TESTS,
-        though a custom value can be passed to the test case as well.
-        Cases:
-        1. When a tarball is specified in .yml file, extract the tar and run kselftests.
-        Example:
-        - name: kselftest_file_path
-          value: <path_to_kselftests.tar.xz>
-          is_case_visible: true
-        2. When a tarball is not specified in .yml file, clone Mariner kernel,
-        copy current config to .config, build kselftests and generate a tar.
-        For both cases, verify that the kselftest tool extracts the tar, runs the script
-        run_kselftest.sh and redirects test results to a file kselftest-results.txt.
-        3. For this lite version, the list of collections to run and the list
-        of tests to skip can be specified in the .yml file.
-        Example:
-        - name: kself_skip_tests
-          value: <comma_separated_list_of_tests_to_skip - format: collection:test_name>
-          is_case_visible: true
-        - name: kself_test_collection
-          value: <comma_separated_list_of_collections_to_run - format: collection_name>
-          is_case_visible: true
+        This test case will run a lighter version of kselftests, focusing on specific test suites
+        and skipping less critical and noisy tests. The default list of tests to run
+        is defined in the `_KSELF_LITE_TESTS` list, which includes collections such as "bpf",
+        "core", "futex", "ipc", "mm", "net", "timers", and "x86". These tests were
+        selected to cover critical kernel functionalities, such as memory management, inter-process
+        communication, and synchronization primitives, while reducing execution time and resource usage.
+
+        Purpose:
+        This "lite" version is designed for scenarios where running the full kselftest suite is
+        not feasible, such as environments with limited resources or during iterative development
+        where faster feedback is required. It ensures that critical kernel features are tested
+        without the overhead of running the entire suite. Also, ensuring that the user can avoid tests/
+        suites that are known to fail or are not relevant to their use case.
+        
+        Customization:
+        Users can customize the test by specifying the `kself_test_collection` and `kself_skip_tests`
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -56,3 +68,66 @@ def verify_kselftest(
...
+        if skip_tests:
+            skip_tests_list = skip_tests.split(",")
+        else:
+            skip_tests_list = []
+        try:
+            kselftest: Kselftest = node.tools.get(
+                Kselftest,
+                kselftest_file_path=file_path,
+            )
+            kselftest.run_all(
```
LiliDeng: can we move the common code to a method, and it can be used by both verify_kself_lite and verify_kself?

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -56,3 +68,66 @@ def verify_kselftest(
             kselftest.run_all(result, log_path, self._KSELF_TIMEOUT)
         except UnsupportedDistroException as identifier:
             raise SkippedException(identifier)
+
+    @TestCaseMetadata(
+        description="""
+        This test case will run kself lite tests - a subset of kselftests.
```
squirrelsc: Specify the subset, and the scenario. If it's unclear, it's hard to know the value of this test case.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -56,3 +68,66 @@ def verify_kselftest(
...
+          value: <comma_separated_list_of_collections_to_run - format: collection_name>
+          is_case_visible: true
+        """,
+        priority=3,
+        timeout=_CASE_TIME_OUT,
+        requirement=simple_requirement(
+            min_core_count=16,
+        ),
+    )
+    def verify_kself_lite(
```
squirrelsc: use full name for kselftest: `verify_kselftest_lite`

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -216,13 +217,51 @@ def run_all(
...
+                    if any(
+                        collection == test.split(":")[0].split("/")[0]
+                        for collection in run_collections
+                    )
+                ]
+
+            # Exclude tests based on skip_tests
+            tests_to_run = [test for test in filtered_tests if test not in skip_tests]
+
+            for test in tests_to_run:
```
squirrelsc: Can it run in one command, instead of one by one? It can reuse more code.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -56,3 +68,66 @@ def verify_kselftest(
...
+        self,
+        node: Node,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        file_path = variables.get("kselftest_file_path", "")
+        skip_tests = variables.get("kself_skip_tests", "")
+        test_collection = variables.get("kself_test_collection", "")
```
LiliDeng: by default, other users won't pass these variables, I tried this scenario, this case timeout.

microsoft/testsuites/kselftest/kselftest-suite.py
umfranci: Created a common function _run_kselftest which is invoked by both the tests.

microsoft/testsuites/kselftest/kselftest-suite.py
umfranci: Added more lines to the description - the default list of tests is provided in the _KSELF_LITE_TESTS List and an example is also provided as to what values can be passed from the runbook.

microsoft/testsuites/kselftest/kselftest.py
umfranci: We expect that the number of tests would be quite high even if a few important test suites are selected for execution. If we try to create a single command, it might result in a huge single line command and cause unforeseen issues. Hence we went ahead with the approach of single case execution using a loop.

microsoft/testsuites/kselftest/kselftest-suite.py
squirrelsc: It's not enough, because the document is provided separately than the code. So it needs to roughly describe what test cases are included, and the purpose of the scenario.

microsoft/testsuites/kselftest/kselftest.py
squirrelsc: Run in one command should be faster. It's ok to have a long command line. Check if it's possible to read test case list from a file.

microsoft/testsuites/kselftest/kselftest-suite.py
umfranci: Sure @squirrelsc - have added description on the tests that are included by default and explained the scenario. Please let me know if anything specific needs to be added.

microsoft/testsuites/kselftest/kselftest-suite.py
umfranci: Done.

microsoft/testsuites/kselftest/kselftest.py
umfranci: The option to read tests from a file is not available for the Kself command. Though since we have the list available in a variable, updated the test to form the required single command and execute the same.

microsoft/testsuites/kselftest/kselftest-suite.py
umfranci: @LiliDeng tested both Kself and Kself-Lite test cases with and without passing these parameters and in both the cases the tests execute without issues.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -216,13 +217,52 @@ def run_all(
...
-        )
+
+        # Initialize run_tests and skip_tests from kwargs if provided
+        run_collections = kwargs.get("run_collections", [])
+        skip_tests = kwargs.get("skip_tests", [])
+
+        if run_collections or skip_tests:
+            # List all available tests
+            list_result = self.run(" -l", shell=True)
+            list_result.assert_exit_code()
```
squirrelsc: provide an error message.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -216,13 +217,52 @@ def run_all(
...
+
+            # Filter tests based on run_collections if it exists
+            if not run_collections:
+                filtered_tests = all_tests
+            else:
+                filtered_tests = [
+                    test
+                    for test in all_tests
+                    if any(
+                        collection == test.split(":")[0].split("/")[0]
```
squirrelsc: provide examples above, and use regexp.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -216,13 +217,52 @@ def run_all(
...
+        skip_tests = kwargs.get("skip_tests", [])
+
+        if run_collections or skip_tests:
+            # List all available tests
+            list_result = self.run(" -l", shell=True)
+            list_result.assert_exit_code()
+            all_tests = list_result.stdout.splitlines()
+
+            # Filter tests based on run_collections if it exists
+            if not run_collections:
```
squirrelsc: switch the order of `if/else`, so the `if run_collection` is the first one. It's more natural on the order with positive conditions.

microsoft/testsuites/kselftest/kselftest.py
umfranci: Added an error in the assertion

microsoft/testsuites/kselftest/kselftest.py
umfranci: Updated to use regex and provided an example of how the filter works

microsoft/testsuites/kselftest/kselftest.py
umfranci: Updated the code and switched if-else order

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@umfranci please resolve the conflicts.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -47,15 +59,111 @@ def verify_kselftest(
...
     ) -> None:
-        file_path = variables.get("kselftest_file_path", "")
-        working_path = variables.get("kselftest_working_path", "")
-        run_as_root = variables.get("kselftest_run_as_root", False)
         try:
-            kselftest: Kselftest = node.tools.get(
-                Kselftest,
-                working_path=working_path,
-                file_path=file_path,
+            self._run_kselftest(
```
LiliDeng, line 63: Please check this error

```
 KselftestTestsuite.verify_kselftest: FAILED   failed. TypeError: Kselftest.__init__() missing 2 required positional arguments: 'working_path' and 'file_path'
2025-04-01 08:22:31.118[1856][INFO] lisa.RootRunner test result summary
```

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -47,15 +59,111 @@ def verify_kselftest(
...
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        test_collection_list = (
+            variables.get("kself_test_collection", "").split(",")
+            if variables.get("kself_test_collection", "")
+            else self._KSELF_LITE_TESTS
+        )
+        skip_tests_list = (
+            variables.get("kself_skip_tests", "").split(",")
```
squirrelsc: change to `kselftest_skip_tests`.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -47,15 +59,111 @@ def verify_kselftest(
...
+    )
+    def verify_kselftest_lite(
+        self,
+        node: Node,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        test_collection_list = (
+            variables.get("kself_test_collection", "").split(",")
```
squirrelsc: `kself_test_collection` -> `kselftest_include_test_collections`

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -47,15 +59,117 @@ def verify_kselftest(
...
+        list of tests to run is defined in the `_KSELF_LITE_TESTS` list, which includes
+        collections such as "bpf", "core", "futex", "ipc", "mm", "net", "timers", and
+        "x86". These tests were selected to cover critical kernel functionalities, such
+        as networking, timer management, and architecture-specific features, while
+        reducing execution time and resource usage.
+
+        Purpose:
+        This "lite" version is designed for scenarios where running the full kselftest
+        suite is not feasible or not required. It ensures that
+        critical kernel features are tested without the overhead of running the entire
```
squirrelsc, line 85: Sorry for commenting this late, but it's worth to get test case simpler.
Is the "critical kernel features" defined somewhere? If not, please add variables supports on the existing test case, and no need to add a new test case. If both `kselftest_include_test_collections` and `kselftest_skip_tests` are not provided, it runs all cases. The `kselftest_include_test_collections` can be configured at runtime, so it doesn't need to check in the list.

umfranci: Hi @squirrelsc, we had a discussion internally to decide which collections are supposed to be critical for testing the kernel functionality in Kself. Hence we created this pre-defined list while also providing users the option to send the list of collections they want to run. 

In case we try to add the functionality to the existing test case, there would not be any pre-defined list and it would always be the user-list that would be run for the lite case. I can make the required changes if we want to follow that approach.

squirrelsc: Thanks for sharing the context.

Recently, another team needs to customize test cases, and it's not the same as the list in the PR. So, I'm wondering if there is a common concept of "critical" tests. Also, it seems no scenario needs to run both full and lite test cases. So, one test case makes more sense to me. When you define the list within the pipeline, you have more flexibility to update the list.

PR #3736, opened 2025-04-01
junit: add DeploymentMessage in the notifier to pass the azure deployment parameters

LiliDeng:


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -602,7 +603,11 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
                 location, deployment_parameters = self._create_deployment_parameters(
                     resource_group_name, environment, log
                 )
-
+                deployment_parameters_message = DeploymentMessage()
```
squirrelsc, line 606: Can it reuse the environment message?

LiliDeng: we can get the value from environment.runbook.nodes_requirement[0].features.items, can I pass environment.runbook as the message?

PR #3737, opened 2025-04-01
azure: handle PendingCopyOperation error during vhd copy

pupacha:
When a single vhd url is passed to concurrent LISA runs, we land in a situation where Azure throws following error to one of LISA runs.

ResourceExistsError: There is currently a pending copy operation.

This is because multiple LISA runs try to copy the vhd to a same destination. While there is already a lock to avoid such situation within one LISA run, this could still happen with multiple LISA runs. Hence, catch the exception and gracefully wait until copy operation is completed.

### DIFF
--- a/lisa/sut_orchestrator/azure/common.py
+++ b/lisa/sut_orchestrator/azure/common.py
@@ -1934,9 +1934,26 @@ def copy_vhd_to_storage(
                     log=log,
                 )
             else:
-                blob_client.start_copy_from_url(
-                    src_vhd_sas_url, metadata=None, incremental_copy=False
-                )
+                # When multiple LISA runs use the same VHD URL concurrently, Azure may
+                # throw the following error during start_copy_from_url:
+                #
+                # ResourceExistsError: There is currently a pending copy operation.
+                #
+                # This happens because multiple runs attempt to copy the VHD to the same
+                # destination. Although a lock prevents this within a single LISA run,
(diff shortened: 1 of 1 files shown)


lisa/sut_orchestrator/azure/common.py
```diff
@@ -1934,9 +1934,17 @@ def copy_vhd_to_storage(
...
             else:
-                blob_client.start_copy_from_url(
-                    src_vhd_sas_url, metadata=None, incremental_copy=False
-                )
+                try:
+                    blob_client.start_copy_from_url(
+                        src_vhd_sas_url, metadata=None, incremental_copy=False
+                    )
+                except ResourceExistsError as e:
+                    if "PendingCopyOperation" in str(e):
```
squirrelsc, line 1951: When it happens, does the `wait_copy_blob` still work?

pupacha: I explained the scenario in the PR description. Let me know, if you need more details.

squirrelsc: Please add it in the code comment too. The PR is for change, and the code comment is to explain the behavior, and it exists with the code logic together.

pupacha: Added comments

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3747, opened 2025-04-03
Add case verify_python_version

lubaihua33:


microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1298,3 +1304,47 @@ def verify_waagent_version(self, node: Node) -> None:
...
+        minimum_version = Version("3.8.0")
+        python_commands = ["python3 --version", "python --version"]
+        python_version_output = None
+
+        for command in python_commands:
+            result = node.execute(command, sudo=True, shell=True)
+            if result.exit_code == 0:
+                python_version_output = result.stdout.strip()
+                break
+        if not python_version_output:
```
LiliDeng, line 1332: is python needed because of waagent? if so, this case may not be applicable to ready and hyperv?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1298,3 +1304,47 @@ def verify_waagent_version(self, node: Node) -> None:
...
+        priority=1,
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY, HYPERV]),
+    )
+    def verify_python_version(self, node: Node) -> None:
+        minimum_version = Version("3.8.0")
+        python_commands = ["python3 --version", "python --version"]
+        python_version_output = None
+
+        for command in python_commands:
+            result = node.execute(command, sudo=True, shell=True)
```
LiliDeng: Is that possible both python2 and python3 were installed?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1298,3 +1304,47 @@ def verify_waagent_version(self, node: Node) -> None:
...
+        Steps:
+        1. Retrieve the Python version.
+        2. Check if the version is lower than the minimum supported version.
+           The minimum supported version can be found at:
+           https://devguide.python.org/versions/
+        3. Fail the test if the version is lower than the minimum supported version,
+           otherwise pass.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_platform_type=[AZURE, READY, HYPERV]),
```
squirrelsc: How about just leave Azure, because it's an Azure standard?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1298,3 +1304,47 @@ def verify_waagent_version(self, node: Node) -> None:
...
+    )
+    def verify_python_version(self, node: Node) -> None:
+        minimum_version = Version("3.8.0")
+        python_commands = ["python3 --version", "python --version"]
```
squirrelsc: It should always support python3, if it's Linux. So it doesn't need to check `python`.

microsoft/testsuites/core/azure_image_standard.py
squirrelsc: sudo is not necessary.

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: No, it's not becasue of waagent.

lubaihua33: https://learn.microsoft.com/en-us/azure/security/fundamentals/azure-marketplace-images
The case is for some security check.

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: Fixed

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: Fixed

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: We can just use python3, then after our AzCertify internal test, we can see if we need add python2 or python commands.

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: Yes, we can just leave Azure.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3748, opened 2025-04-05
Added new locations for dhclient.conf

r-dailey:
Added two new locations to check for dhclient.conf to resolve an issue where it was failing to be found and causing a verify_dhcp_client_timeout to fail on ~65 images.

### DIFF
--- a/lisa/tools/dhclient.py
+++ b/lisa/tools/dhclient.py
@@ -60,7 +60,6 @@ class Dhclient(Tool):
             config_path = next(
                 (path for path in paths_to_check if ls.path_exists(path, sudo=True)), ""
             )
-            self._log.info(f"dhclient config paths: {paths_to_check}")
             if not config_path:
                 raise LisaException(f"Configuration file for {self._command} not found")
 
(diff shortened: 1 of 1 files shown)


lisa/tools/dhclient.py
```diff
@@ -52,13 +52,15 @@ def get_timeout(self) -> int:
...
+                f"/usr/lib/dracut/modules.d/40network/{self._command}.conf",
+                f"/usr/lib/dracut/modules.d/35network-legacy/{self._command}.conf",
             ]
 
             ls = self.node.tools[Ls]
             config_path = next(
                 (path for path in paths_to_check if ls.path_exists(path, sudo=True)), ""
             )
-
+            self._log.info(f"dhclient config paths: {paths_to_check}")
```
LiliDeng: no need this log, ls will print the log

r-dailey: Oh thank you I forgot to remove this. It is gone now.

LiliDeng (approved):

PR #3750, opened 2025-04-08
enable_vm_nat option

LiliDeng:


lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -52,6 +52,9 @@ param ip_service_tags object
 @description('whether to use ipv6')
 param use_ipv6 bool = false
 
+@description('whether to disable network outbound access')
+param disable_outbound_access bool = false
```
squirrelsc: Set the default value to true, or not provide the default value.

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1208,6 +1208,7 @@ class AzureArmParameter:
     subnet_prefix: str = AZURE_SUBNET_PREFIX
     is_ultradisk: bool = False
     use_ipv6: bool = False
+    disable_outbound_access: bool = False
```
squirrelsc: maybe it should be called `enable_vm_nat`?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -296,6 +296,7 @@ class AzurePlatformSchema:
     use_public_address: bool = field(default=True)
     use_ipv6: bool = field(default=False)
     ip_service_tags: Optional[Dict[str, str]] = field(default=None)
+    disable_outbound_access: bool = field(default=False)
```
squirrelsc: Add comments above, when it's needed, and why it's disabled by default.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -350,6 +351,7 @@ def __post_init__(self, *args: Any, **kwargs: Any) -> None:
                 "subnet_prefix",
                 "use_public_address",
                 "use_ipv6",
+                "default_outbound_access",
```
squirrelsc: Why is the name different with other places?

squirrelsc (approved):

PR #3754, opened 2025-04-08
Revert "ssh connection supports ipv6"

glima:
This reverts commit 967e89eab65e06d37aec105231a69f572cb53acf.

No idea how this was merged, sorry. Any connection_info.address can be a domain name, not a resolved IP addr, at that point.

We get a straight

    Traceback (most recent call Last):
    File "<...>\lisa-overlake\lisa\lisa\util\shell.py", line 89, in wait_tcp_port_ready
    result = tcp_socket.connect_ex((address, port))
    socket.gaierror: [Errno 11001] getaddrinfo failed

with that.

### DIFF
--- a/lisa/util/shell.py
+++ b/lisa/util/shell.py
@@ -1,7 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-import ipaddress
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


lisa/util/shell.py
```diff
@@ -104,14 +101,6 @@ def wait_tcp_port_ready(
     return is_ready, result
 
 
-def _is_valid_ipv6(address: str) -> bool:
-    try:
-        ipaddress.IPv4Address(address)
-        return False
-    except ipaddress.AddressValueError:
```
squirrelsc, line 111: @LiliDeng It looks the logic should be changed. We need to check if it's a valid IPv6 here. Only set it to true, when it's a valid IPv6 address.

LiliDeng (approved):

LiliDeng:
Thanks @glima, I will raise a fix for this today.

LiliDeng (commented):
Thanks @glima, I will raise a fix for this today.

LiliDeng (changes requested):
I will raise a fix today.

glima:
> I will raise a fix today.

Thank you, looking forward to it.

LiliDeng:
> > I will raise a fix today.
> 
> Thank you, looking forward to it.

Fixed in https://github.com/microsoft/lisa/pull/3759, let know whether you still encounter the issue, thanks!

glima:
OK, your change makes things better. Closing this one, thanks.

PR #3755, opened 2025-04-09
Rydailey/devicestatistics

r-dailey:
Updated BSD Ethtool and the verify_device_settings test case. Ethtool get_device_statistics() now works on BSD and added new regex filters and conditionals to the test case to work with the BSD statistics format.

### DIFF
--- a/lisa/tools/ethtool.py
+++ b/lisa/tools/ethtool.py
@@ -1128,7 +1128,8 @@ class EthtoolFreebsd(Ethtool):
         if not force_run and device.device_statistics:
             return device.device_statistics
 
-        result = self.run(f"netstat -i | grep {interface}", force_run=True, shell=True)
+        result = self.run(f"ifconfig {interface}", force_run=True, shell=True)
+        self.node.log.info(f"ifconfig {interface} result: {result.stdout}")
         if (result.exit_code != 0) and (
             "Operation not supported" in result.stdout
             or "no stats available" in result.stdout
--- a/lisa/tools/ethtool.py
+++ b/lisa/tools/ethtool.py
@@ -1140,6 +1140,10 @@ class EthtoolFreebsd(Ethtool):
         if not force_run and device.device_statistics:
             return device.device_statistics
 
+        # Example Matches
+        # hn0
+        # mce0
+        # da0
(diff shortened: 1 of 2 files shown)


lisa/tools/ethtool.py
```diff
@@ -1120,3 +1132,37 @@ def get_device_enabled_features(
...
         return device_features
+
+    def get_device_statistics(
+        self, interface: str, force_run: bool = False
+    ) -> DeviceStatistics:
+        device = self._get_or_create_device_setting(interface)
+        if not force_run and device.device_statistics:
+            return device.device_statistics
+
+        match = re.match(r"^(\w+)(\d+)$", interface)
```
LiliDeng, line 1147: give an example here

microsoft/testsuites/network/networksettings.py
```diff
@@ -56,6 +56,29 @@ class NetworkSettings(TestSuite):
...
+    # dev.mce.0.txstat3tc0.cqe_err: 0
+    # dev.mce.0.txstat3tc0.enobuf: 0
+    # dev.mce.0.txstat3tc0.dropped: 0
+    # dev.mce.0.txstat3tc0.defragged: 0
+    # dev.mce.0.txstat3tc0.csum_offload_none: 0
+    # dev.mce.0.txstat3tc0.tso_bytes: 331660
+    # dev.mce.0.txstat3tc0.tso_packets: 125
+    # dev.mce.0.txstat3tc0.bytes: 857284
+    # dev.mce.0.txstat3tc0.packets: 3287
+    _bsd_vf_queue_stats_regex = re.compile(r"[tr]xstat(?P<name>[\d]+)\S*\.packets")
```
LiliDeng, line 62: add comments for the matched lines or just list matched lines.

lisa/tools/ethtool.py
r-dailey: Added an example

microsoft/testsuites/network/networksettings.py
r-dailey: Removed non-matching lines and left only the two example lines.

LiliDeng (approved):

PR #3758, opened 2025-04-09
Remove sudo=True from kselftest, make it a parameter

paxue:
sudo=True is used a lot in kselftest which is not allowed on some restricted platform.
Make it as a parameter so that we can control it.

### DIFF
--- a/lisa/tools/remote_copy.py
+++ b/lisa/tools/remote_copy.py
@@ -32,9 +32,14 @@ class RemoteCopy(Tool):
         src: PurePath,
         dest: PurePath,
         recurse: bool = False,
+        sudo: bool = True,
     ) -> List[PurePath]:
         return self._copy_internal(
-            src=src, dest=dest, recurse=recurse, is_copy_to_local=True
+            src=src,
+            dest=dest,
+            recurse=recurse,
+            is_copy_to_local=True,
+            sudo=sudo,
         )
 
     def copy_to_remote(
@@ -42,9 +47,14 @@ class RemoteCopy(Tool):
         src: PurePath,
         dest: PurePath,
         recurse: bool = False,
+        sudo: bool = True,
     ) -> List[PurePath]:
         return self._copy_internal(
-            src=src, dest=dest, recurse=recurse, is_copy_to_local=False
+            src=src,
+            dest=dest,
+            recurse=recurse,
+            is_copy_to_local=False,
+            sudo=sudo,
         )
 
     @classmethod
@@ -63,9 +73,10 @@ class RemoteCopy(Tool):
         dest: PurePath,
         recurse: bool = False,
         is_copy_to_local: bool = True,
+        sudo: bool = True,
     ) -> List[PurePath]:
         is_file = self._is_file(
-            self._get_source_node(is_copy_to_local=is_copy_to_local), src
+            self._get_source_node(is_copy_to_local=is_copy_to_local), src, sudo
         )
 
         # recurse should be false for files
@@ -91,7 +102,9 @@ class RemoteCopy(Tool):
             )
 
             # copy files to a temp directory with updated permissions
-            tmp_location = self._prepare_tmp_copy(src, recurse=recurse, is_file=is_file)
+            tmp_location = self._prepare_tmp_copy(
+                src, recurse=recurse, is_file=is_file, sudo=sudo
+            )
 
             # copy files from the temp directory and remove the temp directory
             try:
@@ -113,14 +126,15 @@ class RemoteCopy(Tool):
         else:
             return self.node
 
-    def _is_file(self, node: "Node", path: PurePath) -> bool:
-        return node.tools[Ls].is_file(path, sudo=True)
+    def _is_file(self, node: "Node", path: PurePath, sudo: bool = True) -> bool:
+        return node.tools[Ls].is_file(path, sudo=sudo)
 
     def _prepare_tmp_copy(
         self,
         src: PurePath,
         is_file: bool = False,
         recurse: bool = False,
+        sudo: bool = True,
     ) -> PurePath:
         # copy file/folder to a temp location
         tmp_location = PurePath("/tmp")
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -229,22 +244,35 @@ def run_all(
...
         result_file = f"{result_directory}/{result_file_name}"
-        self.run(
-            f" 2>&1 | tee {result_file}",
-            sudo=run_test_as_root,
-            force_run=True,
-            shell=True,
-            timeout=timeout,
-        )
+
+        if self._tar_file_path:
```
paxue, line 230: For this part of change, it's based on the old version of code. I will test the newest merged main and let's see whether the change is needed

lisa/tools/remote_copy.py
```diff
@@ -32,19 +32,29 @@ def copy_to_local(
         src: PurePath,
         dest: PurePath,
         recurse: bool = False,
+        run_as_sudo: bool = True,
```
squirrelsc: name it `sudo` like other places.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -229,22 +244,35 @@ def run_all(
...
+            self.run(
+                f" -s 2>&1 | tee {result_file}",
+                cwd=PurePosixPath(self._installed_path),
```
squirrelsc: You can set this path before call `run`. So it doesn't need to duplcate the `run` for the variable.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -87,18 +89,27 @@ def can_install(self) -> bool:
...
+            )
+            > 0
+        )
 
     def __init__(
         self,
         node: Node,
         working_path: str,
         file_path: str,
+        install_as_sudo: bool = True,
```
squirrelsc: `sudo`

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -139,7 +150,9 @@ def _install(self) -> bool:
             mkdir.create_directory(self._remote_tar_path.parent.as_posix())
             self.node.shell.copy(PurePath(self._tar_file_path), self._remote_tar_path)
             self.node.tools[Tar].extract(
-                str(self._remote_tar_path), str(self._installed_path)
+                str(self._remote_tar_path),
+                str(self._installed_path),
+                sudo=self._install_as_sudo,
```
squirrelsc: Does it need? The original code doesn't use sudo, but it works. It doesn't need to add `sudo` for this command.

lisa/tools/remote_copy.py
paxue: done

microsoft/testsuites/kselftest/kselftest.py
paxue: Done

microsoft/testsuites/kselftest/kselftest.py
paxue: Done

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -55,7 +55,7 @@ def verify_kselftest(
                 Kselftest,
                 working_path=working_path,
                 file_path=file_path,
-                install_as_sudo=run_as_root,
+                sudo=run_as_root,
```
squirrelsc: missed one to change.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -103,13 +103,13 @@ def __init__(
...
         file_path: str,
-        install_as_sudo: bool = True,
+        sudo: bool = True,
         *args: Any,
         **kwargs: Any,
     ) -> None:
         super().__init__(node, *args, **kwargs)
 
-        self._install_as_sudo = install_as_sudo
+        self._install_as_sudo = sudo
```
squirrelsc, line 112: `self._sudo`

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -87,18 +89,27 @@ def can_install(self) -> bool:
...
         node: Node,
         working_path: str,
         file_path: str,
+        sudo: bool = True,
         *args: Any,
         **kwargs: Any,
     ) -> None:
         super().__init__(node, *args, **kwargs)
 
+        self._install_as_sudo = sudo
```
squirrelsc, line 112: Don't use `self` in test cases. It's not recommended.

squirrelsc: `self._sudo`

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@paxue the new change fails on the image `microsoftcblmariner azure-linux-3 azure-linux-3 3.20250102.02` for this reason
```
  KselftestTestsuite.verify_kselftest: FAILED   failed. AssertionError: [
Get unexpected exit code on cmd ['sh', '-c', 'cp /boot/config-6.6.57.1-7.azl3 .config']] Expected <[0]> to contain item <1>, but did not.
```

lisa/tools/remote_copy.py
```diff
@@ -32,19 +32,29 @@ def copy_to_local(
         src: PurePath,
         dest: PurePath,
         recurse: bool = False,
+        sudo: bool = True,
```
squirrelsc, line 35: The default value should be False, so the default behavior is not changed.

microsoft/testsuites/kselftest/kselftest-suite.py
```diff
@@ -49,13 +49,14 @@ def verify_kselftest(
     ) -> None:
         file_path = variables.get("kselftest_file_path", "")
         working_path = variables.get("kselftest_working_path", "")
-        run_as_root = variables.get("kselftest_run_as_root", False)
+        install_as_sudo = variables.get("install_as_sudo", True)
```
squirrelsc: This change should be reverted, the installation doesn't need the configurable sudo.

squirrelsc: Please revert the change, so the behavior is not changed when running kselftests.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3762, opened 2025-04-14
Updating Kself Test Case to support custom collections and tests to be skipped

umfranci:
Currently the 'verify_kselftest' Test Case executes all the tests and test suites which are a part of the Linux KSelf tests.

Although we wanted to add an option where the users can either provide the list of collections that should be executed, or a list of tests that should be skipped - possibly to avoid any noisy or known error-prone tests.

Hence, providing users more flexibility as to what gets executed from the broader Kself Test Collection.

### DIFF
--- a/microsoft/testsuites/kselftest/kselftest.py
+++ b/microsoft/testsuites/kselftest/kselftest.py
@@ -214,6 +214,7 @@ class Kselftest(Tool):
         log_path: str,
         timeout: int = 5000,
         run_test_as_root: bool = False,
+        **kwargs: Any,
     ) -> List[KselftestResult]:
         # Executing kselftest as root may cause
         # VM to hang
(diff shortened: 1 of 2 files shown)


squirrelsc:
Please avoid creating new PRs on the same change. The previous comments lost.

microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -214,6 +214,7 @@ def run_all(
         log_path: str,
         timeout: int = 5000,
         run_test_as_root: bool = False,
+        **kwargs: Any,
```
squirrelsc: Add new parameters explicitly.

umfranci:
> Please avoid creating new PRs on the same change. The previous comments lost.

Sure @squirrelsc, had abandoned the previous PR to avoid the multiple unnecessary commits to be merged into main. Will ensure to continue same PR henceforth.

microsoft/testsuites/kselftest/kselftest.py
umfranci: Updated in latest commit.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3767, opened 2025-04-16
Added kselftest collection and skip test support

paulli2017:


microsoft/testsuites/kselftest/kselftest.py
```diff
@@ -330,3 +338,41 @@ def _parse_exit_val_to_test_status(self, result_list: List[Dict[str, str]]) -> i
         # tests that passed, skipped or timed out
         else:
             return 0
+
+    def _generate_run_tests(
```
squirrelsc, line 348: call it like `_run_tests`, it's not just "generate"

paulli2017: This function here is to create the test cases that need to be run based on the input test_collections and skip_tests parameters, running the tests is the next step self.run() as before. So, this function does only generate or update the test cases list, without it, all test cases will be run by default. Maybe it should be called "_generate_testcases_list" or "_update_testcases_list"?

PR #3773, opened 2025-04-17
Add dst_vhd_sas_url into the secret list to avoid credential leak

lubaihua33:


LiliDeng (approved):

lisa/sut_orchestrator/azure/common.py
```diff
@@ -1926,6 +1926,7 @@ def copy_vhd_to_storage(
                     platform=platform,
                 )
                 dst_vhd_sas_url = f"{full_vhd_path}?{sas_token}"
+                add_secret(dst_vhd_sas_url, PATTERN_URL)
```
squirrelsc, line 1929: The best way to hide it is to pass in the value by a S_LISA_ env name.

PR #3774, opened 2025-04-17
use standard pip and add nsg rule

LiliDeng:


lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -253,9 +256,37 @@ resource virtual_network_name_resource 'Microsoft.Network/virtualNetworks@2024-0
...
+  ]
+}
+
+resource nsg 'Microsoft.Network/networkSecurityGroups@2024-05-01' = {
+  name: '${toLower(virtual_network_name)}-nsg'
+  location: location
+  properties: {
+    securityRules: [
+      {
+        name: 'LisaSSH'
```
squirrelsc: Add ports for KVM test cases too.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -956,6 +960,19 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
             self.credential, self.subscription_id, self.cloud
         )
 
+    @retry(tries=10, delay=0.5)
+    def _get_external_ip_address(self) -> str:
```
squirrelsc: 1. Move it to util class
2. the name is like `get_public_ip`.

lisa/sut_orchestrator/azure/arm_template.bicep
```diff
@@ -253,9 +256,37 @@ resource virtual_network_name_resource 'Microsoft.Network/virtualNetworks@2024-0
...
+        properties: {
+          priority: 100
+          direction: 'Inbound'
+          access: 'Allow'
+          protocol: 'Tcp'
+          sourcePortRange: '*'
+          destinationPortRange: '22'
+          sourceAddressPrefixes: source_address_prefixes
```
squirrelsc, line 284: Does it work for both IP and IP ranges?

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1254,6 +1271,13 @@ def _create_deployment_parameters(
             self._azure_runbook.shared_resource_group_name
         )
         arm_parameters.enable_vm_nat = self._azure_runbook.enable_vm_nat
+        if self._azure_runbook.source_address_prefixes:
```
squirrelsc: The logic can be in `_get_ip_addresses`, so the flow is clearer in "create deployment parameters“.

lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -956,6 +960,14 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
             self.credential, self.subscription_id, self.cloud
         )
 
+    def _get_ip_addresses(self) -> List[str]:
+        if self._azure_runbook.source_address_prefixes:
```
squirrelsc: The logic should be like:

```python
    def _get_ip_addresses(self) -> List[str]:
        if self._cached_ip_address:
            return self._cached_ip_address
        if self._azure_runbook.source_address_prefixes:
            self._cached_ip_address = self._azure_runbook.source_address_prefixes
        else:
            self._cached_ip_address = [get_public_ip()]
        return self._cached_ip_address
```

squirrelsc (dismissed):

PR #3775, opened 2025-04-22
Remove LTP sudo, Change sudo as parameter, Add full test support for LTP

paxue:
Add full test support for LTP
Remove LTP sudo, change as parameter

### DIFF
--- a/microsoft/testsuites/ltp/ltp.py
+++ b/microsoft/testsuites/ltp/ltp.py
@@ -24,6 +24,7 @@ from lisa.tools import (
     Make,
     Mkdir,
     Pidof,
+    RemoteCopy,
     Rm,
     Swap,
     Sysctl,
@@ -51,7 +52,7 @@ class Ltp(Tool):
     _RESULT_LTP_ARCH_REGEX = re.compile(r"Machine Architecture: (.*)\s+")
 
     LTP_DIR_NAME = "ltp"
-    DEFAULT_LTP_TESTS_GIT_TAG = "20230929"
+    DEFAULT_LTP_TESTS_GIT_TAG = "20250130"
     LTP_GIT_URL = "https://github.com/linux-test-project/ltp.git"
     LTP_RESULT_FILE = "ltp-results.log"
     LTP_OUTPUT_FILE = "ltp-output.log"
@@ -80,6 +81,7 @@ class Ltp(Tool):
         node: Node,
         source_file: str,
         binary_file: str,
+        install_path: str,
         *args: Any,
         **kwargs: Any,
     ) -> None:
@@ -88,11 +90,17 @@ class Ltp(Tool):
         self._git_tag = git_tag if git_tag else self.DEFAULT_LTP_TESTS_GIT_TAG
         self._source_file = source_file
         self._binary_file = binary_file
-        self._command_path = "/tmp/ltp"
+
+        if install_path:
+            self._command_path = install_path
+        else:
+            # default binary_file install path is /tmp/ltp
+            self._command_path = "/tmp/ltp" if self._binary_file else "/opt/ltp"
 
     def run_test(
         self,
         test_result: TestResult,
+        # empty ltp_tests will run LTP full test
         ltp_tests: List[str],
         skip_tests: List[str],
         log_path: str,
@@ -104,8 +112,6 @@ class Ltp(Tool):
         output_file = f"{self._command_path}/{self.LTP_OUTPUT_FILE}"
         skip_file = f"{self._command_path}/{self.LTP_SKIP_FILE}"
 
-        # tests cannot be empty
-        assert_that(ltp_tests, "ltp_tests cannot be empty").is_not_empty()
         ls = self.node.tools[Ls]
         rm = self.node.tools[Rm]
 
@@ -128,7 +134,9 @@ class Ltp(Tool):
         parameters = f"-p -q -l {result_file} -o {output_file} "
 
         # add the list of tests to run
-        parameters += f"-f {','.join(ltp_tests)} "
+        # if ltp_tests=[], "-f" not set, ltp will run full test
+        if ltp_tests:
+            parameters += f"-f {','.join(ltp_tests)} "
 
         # some tests require a big unmounted block device
         # to run correctly.
@@ -243,21 +251,13 @@ class Ltp(Tool):
         source_file: str,
         dest_path: str,
     ) -> PurePath:
-        remote_tar_path = self.node.get_pure_path(
-            self._command_path
-        ) / os.path.basename(source_file)
-
-        mkdir = self.node.tools[Mkdir]
-        mkdir.create_directory(remote_tar_path.parent.as_posix())
-        self.node.shell.copy(
-            PurePath(self._source_file),
-            remote_tar_path,
-        )
+        remote_source_folder = self.get_tool_path(use_global=True)
+        remote_tar_path = remote_source_folder / os.path.basename(source_file)
+        copy = self.node.tools[RemoteCopy]
+        copy.copy_to_remote(PurePath(self._source_file), remote_source_folder)
 
         self.node.tools[Tar].extract(
-            str(remote_tar_path),
-            dest_path,
-            sudo=True,
+            str(remote_tar_path), dest_path, strip_components=1
         )
 
         return self.node.get_pure_path(dest_path)
@@ -279,7 +279,7 @@ class Ltp(Tool):
 
         return ltp_path
 
-    def _build_from_source(self) -> None:
+    def _install_dependencies(self) -> None:
         assert isinstance(self.node.os, Posix), f"{self.node.os} is not supported"
 
         # install common dependencies
@@ -376,6 +376,7 @@ class Ltp(Tool):
         else:
             raise LisaException(f"{self.node.os} is not supported")
 
+    def _set_realtime_cgroup(self) -> None:
         # Some CPU time is assigned to set real-time scheduler and it affects
         # all cgroup test cases. The values for rt_period_us(1000000us or 1s)
         # and rt_runtime_us (950000us or 0.95s). This gives 0.05s to be used
@@ -416,9 +417,13 @@ class Ltp(Tool):
             sysctl = self.node.tools[Sysctl]
             sysctl.write("vm.dirty_ratio", "10")
             sysctl.write("vm.dirty_background_ratio", "5")
-            sysctl.run("-p")
+            sysctl.run("-p", sudo=True)
 
-        # find partition to install ltp
+    def _build_from_source(self) -> None:
+        self._install_dependencies()
+        self._set_realtime_cgroup()
+
+        # find partition to build ltp
         build_dir = self.node.find_partition_with_freespace(
             self.BUILD_REQUIRED_DISK_SIZE_IN_GB
         )
@@ -437,15 +442,17 @@ class Ltp(Tool):
         else:
             cur_ltp_path = self._clone_source(self.LTP_GIT_URL, top_src_dir)
 
-        # build ltp in /opt/ltp since this path is used by some
+        # better to install ltp in /opt/ltp since this path is used by some
         # tests, e.g, block_dev test
         make = self.node.tools[Make]
-        self.node.execute("autoreconf -f", cwd=cur_ltp_path, sudo=True)
-        make.make("autotools", cwd=cur_ltp_path, sudo=True)
+        # uploaded release ltp.tar.xz don't need autoreconf
+        if not self._source_file:
+            self.node.execute("autoreconf -f", cwd=cur_ltp_path)
+            make.make("autotools", cwd=cur_ltp_path)
         self.node.execute(
-            f"./configure --prefix={self._command_path}", cwd=cur_ltp_path, sudo=True
+            f"./configure --prefix={self._command_path}", cwd=cur_ltp_path
         )
-        make.make("all", cwd=cur_ltp_path, sudo=True, timeout=self.COMPILE_TIMEOUT)
+        make.make("all", cwd=cur_ltp_path, timeout=self.COMPILE_TIMEOUT)
 
         # Specify SKIP_IDCHECK=1 since we don't want to modify /etc/{group,passwd}
         # on the remote system's sysroot
--- a/microsoft/testsuites/ltp/ltpsuite.py
+++ b/microsoft/testsuites/ltp/ltpsuite.py
@@ -1,7 +1,6 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
 
-
 from logging import Logger
 from typing import Any, Dict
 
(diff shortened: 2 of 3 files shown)


microsoft/testsuites/ltp/ltp.py
```diff
@@ -72,11 +72,12 @@ def dependencies(self) -> List[Type[Tool]]:
     def can_install(self) -> bool:
         return True
 
-    def __init__(self, node: Node, source_file: str, *args: Any, **kwargs: Any) -> None:
-        super().__init__(node, args, kwargs)
+    def __init__(self, node: Node, *args: Any, **kwargs: Any) -> None:
```
squirrelsc: Give the value with a default value, instead of using kwargs.get

microsoft/testsuites/ltp/ltp.py
```diff
@@ -72,11 +72,12 @@ def dependencies(self) -> List[Type[Tool]]:
...
+        super().__init__(node, *args, **kwargs)
         git_tag = kwargs.get("git_tag", "")
         self._git_tag = git_tag if git_tag else self.DEFAULT_LTP_TESTS_GIT_TAG
-        self._source_file = source_file
+        self._source_file = kwargs.get("source_file", "")
+        self._run_as_sudo = kwargs.get("sudo", True)
```
squirrelsc: A similar question like Kselftest. Any difference to run with/out sudo?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -87,6 +88,9 @@ def run_test(
         block_device: Optional[str] = None,
         temp_dir: str = "/tmp/",
         ltp_run_timeout: int = 12000,
+        # run full test if True, else run lite test
+        run_full_test: bool = False,
```
squirrelsc, line 92: If the list is not given, it can run full. So it doesn't need the additional parameter.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -71,6 +71,9 @@ def verify_ltp_lite(
         # If not provided, we will find a disk with enough space
         block_device = variables.get("ltp_block_device", None)
 
+        run_full_ltp_test = variables.get("run_full_ltp_test", False)
```
squirrelsc: Is there any requirement to run all of them? This flag is easy to be conflict with other parameters. Please create another test case called verify_ltp. The only difference is the new test case is to run all, if no list given.

microsoft/testsuites/ltp/ltp.py
paxue: Do you mean add the following lines in __init__()?
   self._source_file = ""
   self._run_as_sudo = True

squirrelsc: it should be like `def __init__(self, node: Node, source_file: str = "", run_as_sudo:bool =True, *args: Any, **kwargs: Any) -> None:`

microsoft/testsuites/ltp/ltp.py
paxue: For using released file.
Kselftest: don't have config, can direct unzip the release file and run.
LTP: does have config step, this step involve lots of sudo issue.

microsoft/testsuites/ltp/ltpsuite.py
paxue: Is there any requirement to run all of them?
[panfeng xue]: yes. it's required from Cooperate team

I am fine with 2 paths: Create another verify_ltp_full() or just using the same test suite verify_ltp_lite.
From my view, their changes are very limited, only one line code change. Depends how you think this question

microsoft/testsuites/ltp/ltp.py
squirrelsc: I mean for LTP, if run without sudo, does it work?

microsoft/testsuites/ltp/ltpsuite.py
squirrelsc: I don't want to add a parameter called `run_full_ltp_test`, it conflicts with existing parameters like `ltp_test` and `ltp_skip_test`

lisa/operating_system.py
```diff
@@ -388,9 +388,10 @@ def install_packages(
         signed: bool = False,
         timeout: int = 1200,
         extra_args: Optional[List[str]] = None,
+        sudo: bool = True,
```
squirrelsc: did you find any package installation command are blocked? Please provide examples here. We can think about if it can be detected in code, instead of depending on configuration.

lisa/tools/pgrep.py
```diff
@@ -42,10 +42,12 @@ def _freebsd_tool(cls) -> Optional[Type[Tool]]:
...
         return False
 
-    def get_processes(self, process_identifier: str) -> List[ProcessInfo]:
+    def get_processes(
+        self, process_identifier: str, sudo: bool = True
+    ) -> List[ProcessInfo]:
         running_process: List[ProcessInfo] = []
         output = self.run(
-            f'-l "{process_identifier}"', sudo=True, force_run=True
+            f'-l "{process_identifier}"', sudo=sudo, force_run=True
```
squirrelsc: The same above, what's the error message, when this command failed with sudo?

lisa/operating_system.py
```diff
@@ -1922,7 +1922,9 @@ def __init__(self, node: Any) -> None:
         self._dnf_tool_name: str
 
     def _initialize_package_installation(self) -> None:
-        self.set_kill_user_processes()
+        self._log.debug(f"--- CBLMariner, type_name: {self._node.type_name()}")
```
squirrelsc: clean up the temp log.

lisa/operating_system.py
```diff
@@ -1922,7 +1922,9 @@ def __init__(self, node: Any) -> None:
...
+        if self._node.type_name() != "local":
```
squirrelsc: 1. what does it need to skip local? if any reason, write it in the code comment.
2. use `isinstance to check types, not use a string`.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -87,14 +93,18 @@ def run_test(
...
     ) -> List[LtpResult]:
-        # tests cannot be empty
-        assert_that(ltp_tests, "ltp_tests cannot be empty").is_not_empty()
+        if not run_full_test:
+            # ltp_tests cannot be empty
+            assert_that(ltp_tests, "ltp_tests cannot be empty").is_not_empty()
```
squirrelsc: How about empty ltp_tests means run all? So it doesn't need the `fun_full_test` variable.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -327,12 +341,15 @@ def _install(self) -> bool:
...
         sysctl.write("vm.dirty_background_ratio", "5")
-        sysctl.run("-p")
+        sysctl.run("-p", sudo=True)
 
         # find partition to install ltp
-        build_dir = self.node.find_partition_with_freespace(
-            self.BUILD_REQUIRED_DISK_SIZE_IN_GB
-        )
+        if self.node.type_name() == "local":
+            build_dir = "/"
```
squirrelsc: Why use root for the local node?

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -105,6 +107,69 @@ def verify_ltp_lite(
...
+        ),
+    )
+    def verify_ltp_full(
+        self,
+        node: Node,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        run_as_sudo = variables.get("run_as_sudo", True)
```
squirrelsc: remove `run_as_sudo`

microsoft/testsuites/ltp/ltp.py
```diff
@@ -72,29 +72,33 @@ def dependencies(self) -> List[Type[Tool]]:
...
+        super().__init__(node, *args, **kwargs)
         git_tag = kwargs.get("git_tag", "")
         self._git_tag = git_tag if git_tag else self.DEFAULT_LTP_TESTS_GIT_TAG
         self._source_file = source_file
 
     def run_test(
         self,
         test_result: TestResult,
-        ltp_tests: List[str],
+        ltp_tests: List[str],  # empty will run full test
```
squirrelsc: Write comments above of code.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -72,29 +72,33 @@ def dependencies(self) -> List[Type[Tool]]:
...
         ltp_run_timeout: int = 12000,
     ) -> List[LtpResult]:
-        # tests cannot be empty
-        assert_that(ltp_tests, "ltp_tests cannot be empty").is_not_empty()
         ls = self.node.tools[Ls]
         rm = self.node.tools[Rm]
 
         # remove skipfile if it exists
-        if ls.path_exists(self.LTP_SKIP_FILE):
+        if ls.path_exists(self.LTP_SKIP_FILE, sudo=True):
```
squirrelsc: Is the sudo needed?

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -83,7 +83,7 @@ def verify_ltp_lite(
         else:
             skip_test_list = []
 
-        if not block_device:
+        if not block_device and node.type_name() != "local":
```
squirrelsc: 1. Why not support local? Add code comments.
2. Use isinstance to check by type, not by a string.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -105,6 +107,66 @@ def verify_ltp_lite(
...
+        ),
+    )
+    def verify_ltp_full(
+        self,
+        node: Node,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        ltp_source_file = variables.get("ltp_source_file", "")
```
squirrelsc: As the new logic, it can share much more logic.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -111,8 +116,10 @@ def run_test(
         # add parameters for the test logging
         parameters = f"-p -q -l {self.LTP_RESULT_PATH} -o {self.LTP_OUTPUT_PATH} "
 
-        # add the list of tests to run
-        parameters += f"-f {','.join(ltp_tests)} "
+        # add the list of tests to run, only when ltp_tests is not empty
+        # if ltp_tests=[], "-f" not set, ltp will run full test
+        if len(ltp_tests) > 0:
```
squirrelsc: `if ltp_tests:` is enough.

microsoft/testsuites/ltp/ltp.py
```diff
@@ -347,15 +356,19 @@ def _install(self) -> bool:
...
                 f"Use downloaded source code tar file: {self._source_file}!"
             )
+            remote_source_folder = self.get_tool_path(use_global=True)
             remote_source_file = self.get_tool_path(use_global=True) / os.path.basename(
                 self._source_file
             )
 
             copy = self.node.tools[RemoteCopy]
-            copy.copy_to_remote(PurePath(self._source_file), remote_source_file)
+            copy.copy_to_remote(
```
squirrelsc: Does it need sudo false?

microsoft/testsuites/ltp/ltp.py
```diff
@@ -347,15 +356,19 @@ def _install(self) -> bool:
...
+                PurePath(self._source_file), remote_source_folder, sudo=False
+            )
             self.node.tools[Tar].extract(
-                str(remote_source_file), top_src_dir, sudo=True
+                str(remote_source_file), top_src_dir, strip_components=1
```
squirrelsc: What's the strip_components for? Why it can pass before?

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -32,6 +32,44 @@ class LtpTestsuite(TestSuite):
...
     LTP_REQUIRED_DISK_SIZE_IN_GB = 2
 
+    def _get_parameter(
+        self,
+        variables: Dict[str, Any],
+    ) -> None:
+        # read parameters from .yml file
+        tests = variables.get("ltp_test", "")
+        skip_tests = variables.get("ltp_skip_test", "")
+        self.ltp_source_file = variables.get("ltp_source_file", "")
```
squirrelsc: Don't use class members in the test suite, it breaks the design. Use a top-level common method, and no output parameters needed.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -19,6 +19,59 @@
...
+    binary_file = variables.get("ltp_binary_file", "")
+    # block device is required for few ltp tests
+    # If not provided, we will find a disk with enough space
+    block_device = variables.get("ltp_block_device", None)
+
+    # get comma separated list of tests
+    if test_str:
+        test_list = test_str.split(",")
+    else:
+        test_list = LTP_LITE_TESTS  # fallback lite tests
```
squirrelsc: This is the only different code between lite and full test cases. All other logic should be shared. It doesn't need to create a test config class. Refer to smoke_test, how to use one method to serve multiple test cases.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -32,57 +31,32 @@ class LtpTestsuite(TestSuite):
...
         self,
         node: Node,
-        log_path: str,
         variables: Dict[str, Any],
+        run_full_test: bool,
+        log_path: str,
         result: TestResult,
     ) -> None:
-        # parse variables
+        test_str = variables.get("ltp_test", "")
```
squirrelsc: This line can be moved to the test case level, like below. It would be cleared on the difference.

```python
# verify_ltp_lite
test_str = variables.get("ltp_test", self.LTP_LITE_TESTS)

# verify_ltp_full
test_str = variables.get("ltp_test", "")
```

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -93,22 +61,92 @@ def verify_ltp_lite(
                     node.tools[Lsblk].find_disk_by_mountpoint(mountpoint).device_name
                 )
 
-        # run ltp lite tests
+        if not test_list:
+            # empty test_list means full test, require long timeout
+            run_timeout = 72000
```
squirrelsc: It won't work, because the test case metadata has another timeout. This value couldn't be longer than that. 72000 means  20 hours, it's too long to finish all cases.

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -93,22 +61,92 @@ def verify_ltp_lite(
...
+        ),
+    )
+    def verify_ltp_lite(
+        self,
+        node: Node,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        test_str = variables.get("ltp_test", self.LTP_LITE_TESTS)
```
squirrelsc, line 66: Change it like below, and change the argument type to string

```python
        test_str = variables.get("ltp_test", self.LTP_LITE_TESTS)
        self._verify_ltp(node, variables, test_str, log_path, result)
```

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -93,22 +61,92 @@ def verify_ltp_lite(
...
+        ),
+    )
+    def verify_ltp_full(
+        self,
+        node: Node,
+        log_path: str,
+        variables: Dict[str, Any],
+        result: TestResult,
+    ) -> None:
+        self._verify_ltp(node, variables, [], log_path, result)
```
squirrelsc: Change it like below, and change the argument type to string

```python
        test_str = variables.get("ltp_test", "")
        self._verify_ltp(node, variables, test_str, log_path, result)
```

microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -29,60 +28,29 @@
...
-            min_core_count=8,
-            disk=schema.DiskOptionSettings(
-                data_disk_count=search_space.IntRange(min=1),
-                data_disk_size=search_space.IntRange(min=12),
-            ),
-            unsupported_os=[BSD, Windows],
-        ),
-    )
-    def verify_ltp_lite(
+    def _verify_ltp(
```
squirrelsc: In a class, move the private method below public methods.

squirrelsc:
@LiliDeng LGTM

LiliDeng:
@paxue could you please rebase the latest main, i saw permission issue on mariner against this PR.

LiliDeng (approved):

PR #3778, opened 2025-04-23
Add tool for resize os partition and use it to expand it

smit-gardhariya:
Extend expand_os_partition() for libvirt
post node creation. Use same function for hyperv/libvirt platforms to grow root partition size.

### DIFF
--- a/lisa/sut_orchestrator/libvirt/platform.py
+++ b/lisa/sut_orchestrator/libvirt/platform.py
@@ -38,6 +38,7 @@ from lisa.tools import (
     Ls,
     Mkdir,
     QemuImg,
+    ResizePartition,
     Sed,
     Service,
     Uname,
@@ -398,6 +399,7 @@ class BaseLibvirtPlatform(Platform, IBaseLibvirtPlatform):
         try:
             self._create_nodes(environment, log)
             self._fill_nodes_metadata(environment, log)
+            self._expand_nodes_os_partition(environment, log)
 
         except Exception as ex:
             assert environment.platform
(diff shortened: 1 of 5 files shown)


lisa/sut_orchestrator/libvirt/platform.py
```diff
@@ -829,6 +830,10 @@ def _fill_nodes_metadata(self, environment: Environment, log: Logger) -> None:
                     expected_exit_code_failure_message="waiting on cloud-init",
                 )
 
+            # In some cases, we observe that resize vhd resizes the entire disk
+            # but fails to expand the partition size.
+            expand_root_partition(node)
```
pupacha: I think its better if this is part of `_deploy_nodes`.
`_fill_nodes_metadata` generally should really change the node in anyway.

smit-gardhariya: Until we set NATing, we cant perform this action. Sure, we can call this function (or write new one) after calling `_fill_nodes_metadata ` under `_deploy_nodes`. we traverse through nodes list already at `_fill_nodes_metadata `. There is no point in keeping it in `_deploy_nodes` and do the same nodes list traverse again.

lisa/sut_orchestrator/util/resize_partition.py
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+import re
+
+from lisa.node import RemoteNode
+from lisa.tools import Mount
+from lisa.util import LisaException
+
+
+def expand_root_partition(node: RemoteNode) -> None:
```
squirrelsc: Can you put it in a tool called `ResizePartition`? And there is another similar method `_expand_os_partition` in `microsoft\testsuites\power\common.py`

smit-gardhariya: Done

squirrelsc:
@LiliDeng LGTM

lisa/sut_orchestrator/libvirt/platform.py
pupacha: Sure, Lets create a new function and call it from `_deploy_nodes`. That would be more clean.

pupacha: @smit-gardhariya  Feel free to resolve this comment after making this change.

LiliDeng (approved):

PR #3780, opened 2025-04-24
Fixes to dhclient renew logic

umfranci:
Following are the fixes/updates as part of this PR:
- The dhclient.py file is using either dhclient or dhcpcd depending on what is available on the system. Although in the renew function, "-r {interface} && dhclient {interface}" was executed, resulting in dhcpcd and dhclient both being called in a single command when dhcpcd was available - resulting in failures. 
- For dhcpcd '-r' is not the right option for removing the lease, instead -k should be used. 
- In case dhcpcd is not already running on an interface, executing the release and assignment in a single command would fail the assignment part as well. Hence, splitting the release and assignment into two steps.
- Added GB200 GPU as a minor update to the list of nvidia-smi devices.


squirrelsc:
@LiliDeng LGTM. @umfranci even the GB200 is a small fix, it's better to have it in a separated PR or you can organize commits by one for GB200, one for dhcp.

LiliDeng (approved):

PR #3789, opened 2025-04-30
Updating Infiniband setup to support ARM64 

umfranci:
- Currently the packages installed as part of the Infiniband setup are for x86. Hence the setup is seen to fail for ARM64. Updating the packages as part of the PR and adding support to check the CPU Architecture before installing arch dependent packages.
- Adding appropriate checks for hpc image where missing.

### DIFF
--- a/lisa/features/infiniband.py
+++ b/lisa/features/infiniband.py
@@ -173,11 +173,7 @@ class Infiniband(Feature):
         return cat.read(f"/sys/class/infiniband/{ib_device_name}/ports/1/pkeys/0")
 
     def setup_rdma(self) -> None:
-        if self._node.tools[Ls].path_exists(
-            "/opt/azurehpc/component_versions.txt"
-        ) or self._node.tools[Ls].path_exists("/opt/azurehpc/component-versions"):
-            self.is_hpc_image = True
-        else:
+        if not self.is_hpc_image:
             self.install_ofed()
 
         node = self._node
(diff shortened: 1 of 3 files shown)


LiliDeng:
Do you know any vm sizes offer rdma and amd64? @umfranci

lisa/features/infiniband.py
```diff
@@ -173,7 +173,11 @@ def get_pkey(self) -> str:
         return cat.read(f"/sys/class/infiniband/{ib_device_name}/ports/1/pkeys/0")
 
     def setup_rdma(self) -> None:
-        if not self.is_hpc_image:
+        if self._node.tools[Ls].path_exists(
```
squirrelsc: How the is_hpc_image is set before? it looks ignored the existing `is_hpc_image` value. This logic is applied to Azure only, so the logic should be Azure as well. Please update the method in the Azure extended feature.

umfranci: The is_hpc_image variable was given a default value of False during initialization and then the function setup_rdma() is called. So, this variable is never updated in the infiniband suite and ofed gets installed each time, even if it is an hpc image.  

Infiniband suite has an independent setup_rdma() where we would need to check if it is an hpc image. Should we still be relaying through the Azure Feature file? Would the component_version file not exist for other platforms except Azure?

squirrelsc: hpc is a Microsoft image, so leave it to Azure feature overwrites make more sense.

umfranci: Sure, thanks @squirrelsc, reverted to the original check in Azure Feature

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3796, opened 2025-05-05
Fixed LTP test on Mariner OS

paulli2017:
Fixed LTP test on Mariner OS

### DIFF
--- a/lisa/tools/pidof.py
+++ b/lisa/tools/pidof.py
@@ -27,7 +27,11 @@ class Pidof(Tool):
         return pids
 
     def wait_processes(
-        self, process_name: str, timeout: int = 600, interval: int = 10
+        self,
+        process_name: str,
+        sudo: bool = False,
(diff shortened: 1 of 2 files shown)


lisa/tools/pidof.py
```diff
@@ -27,7 +27,11 @@ def get_pids(self, process_name: str, sudo: bool = False) -> List[str]:
         return pids
 
     def wait_processes(
-        self, process_name: str, timeout: int = 600, interval: int = 10
+        self,
+        process_name: str,
+        sudo: bool = True,
```
squirrelsc: Keep the original behavior, and minimal the permission. So it should be False.

paulli2017: The change is done, thanks

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3797, opened 2025-05-06
add create_public_address to allow not to create public ip address when use_public_address is False

LiliDeng:


docs/run_test/platform.rst
```diff
@@ -146,6 +146,7 @@ deployment.
          subnet_prefix: $(subnet_name)
          use_public_address: "<true or false>"
          use_ipv6: "<true or false>"
+         generate_public_address: "<true or false>"
```
squirrelsc: 1. `create_public_address`
2. move it closer to `use_public_address`
3. When `create_public_address` is False, it mean `use_public_address` should be False too. So please update the logic, if the `create_public_address` == False, overwrite `use_public_address` to False.

docs/run_test/platform.rst
```diff
@@ -190,6 +191,9 @@ deployment.
   public IP addresses.  False means to connect with the private IP addresses.
   If not provided, the connections will default to using the public IP
   addresses.
+* **generate_public_address**. True means to create a public IP address for the
```
squirrelsc: Please explain, when use_public_address is False, why the create_public_address == True is a valid scenario about the default outbound.

squirrelsc:
Please add commit message to explain the scenario, which is supported by this change.

squirrelsc (dismissed):

lisa/sut_orchestrator/azure/features.py
```diff
@@ -171,7 +171,11 @@ def _start(self, wait: bool = True) -> None:
         platform: AzurePlatform = self._platform  # type: ignore
 
         public_ip, private_ip = get_primary_ip_addresses(
-            platform, self._resource_group_name, get_vm(platform, self._node)
+            platform,
+            self._resource_group_name,
+            get_vm(platform, self._node),
+            platform._azure_runbook.use_ipv6,
```
squirrelsc: They can get from the first argument `platform`, so it doesn't need separated arguments.

squirrelsc (approved):

PR #3802, opened 2025-05-08
ignore_dirty_env: make tests run continuously against ready platform

LiliDeng:


lisa/sut_orchestrator/ready.py
```diff
@@ -35,6 +45,13 @@ def supported_features(cls) -> List[Type[Feature]]:
             features.SerialConsole,
         ]
 
+    def _initialize(self, *args: Any, **kwargs: Any) -> None:
+        runbook: ReadyPlatformSchema = self.runbook.get_extended_runbook(
```
squirrelsc: How about just reset the dirty flag in the ready platform? It looks easier to understand.

lisa/sut_orchestrator/ready.py
```diff
@@ -66,6 +66,7 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
 
     def _delete_environment(self, environment: Environment, log: Logger) -> None:
         # ready platform doesn't support delete environment
+        environment.mark_clean()
```
squirrelsc: Add comments above to explain why need this. And as we discussed, use a switch to control it.

lisa/sut_orchestrator/ready.py
```diff
@@ -66,7 +83,10 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
 
     def _delete_environment(self, environment: Environment, log: Logger) -> None:
         # ready platform doesn't support delete environment
-        pass
+        if self._ready_runbook.ignore_dirty_env:
+            # if ignore_dirty_env is set, it will reset the environment to clean
+            # then test cases can be run on the same environment
+            environment.mark_clean()
```
LiliDeng: this happens before `environment.mark_dirty()`, will find a right place to do this.

lisa/environment.py
```diff
@@ -182,6 +182,7 @@ def __init__(
 
         self._default_node: Optional[Node] = None
         self._is_dirty: bool = False
+        self.ignore_dirty_env: bool = False
```
squirrelsc: It should be a switch in the ready platform schema. If it's True, the switch won't be reset in the ready platform. It doesn't need to change the lisa_runner logic.

lisa/sut_orchestrator/ready.py
squirrelsc: Try to overwrite the `deploy_environment` method like below in the ready platform.
```python
def deploy_environment(self, environment: Environment) -> None:
    #  check and overwrite dirty flag here
    super().deploy_environment(environment)
```

LiliDeng: this also happens before `environment.mark_dirty()`

image

squirrelsc: How about overwrite the delete_environment?

LiliDeng: current implement works

lisa/sut_orchestrator/ready.py
```diff
@@ -14,11 +17,25 @@
 from . import READY
 
 
+@dataclass_json()
+@dataclass
+class ReadyPlatformSchema:
+    ignore_dirty_env: bool = False
```
squirrelsc: I think `reuse_dirty_env` is clearer and set it to True by default. So the dirty env will be reused by default.

lisa/sut_orchestrator/ready.py
```diff
@@ -68,6 +85,11 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
         # ready platform doesn't support delete environment
         pass
 
+    def delete_environment(self, environment: Environment) -> None:
+        # ready platform doesn't support delete environment
```
squirrelsc: This comment is confusing. Please write the logic about why it needs to prevent deleting dirty env in some cases.

lisa/sut_orchestrator/ready.py
```diff
@@ -68,6 +86,12 @@ def _delete_environment(self, environment: Environment, log: Logger) -> None:
         # ready platform doesn't support delete environment
         pass
 
+    def delete_environment(self, environment: Environment) -> None:
+        # This is necessary when running multiple test cases that share the same
+        # environment, as it avoids skipping later cases due to cleanup.
+        if not self._ready_runbook.reuse_dirty_env:
```
squirrelsc: I think we need log a message, if an env is dirty, but not be deleted. So we know an environment is marked as dirty. And the test case may need to be fixed.

LiliDeng: I don't quite understand the meaning of `the test case may need to be fixed.`, I try to write log lines, please double check, thanks.

squirrelsc: We should know, when an env is dirty, it mostly like the previous test case damaged the env. We need to check if the test case should be fixed.

lisa/sut_orchestrator/ready.py
```diff
@@ -14,11 +17,25 @@
 from . import READY
 
 
+@dataclass_json()
+@dataclass
+class ReadyPlatformSchema:
+    # When set to True, the environment will not be deleted even if it becomes dirty.
+    reuse_dirty_env: bool = field(default=False)
```
LiliDeng: I set it as false by default, since when default value is set as true, the run against hello_world.yml which use local node type, will enter into dead loop. I didn't figure out the root cause yet.

squirrelsc: Do you mean, the case failed only the "default" is True, but not failed, if it's "set" to True?

LiliDeng: When the reuse_dirty_env is set as True,  the run will enter into a dead loop against hello_world.yml which uses local node type.

squirrelsc: Please try to set the env status back to prepared, instead of deployed.

lisa/sut_orchestrator/ready.py
```diff
@@ -65,8 +83,13 @@ def _deploy_environment(self, environment: Environment, log: Logger) -> None:
         pass
 
     def _delete_environment(self, environment: Environment, log: Logger) -> None:
-        # ready platform doesn't support delete environment
-        pass
+        if self._ready_runbook.reuse_dirty_env:
+            log.info(
```
squirrelsc: Move it to debug level

lisa/sut_orchestrator/ready.py
LiliDeng: when set it as prepared, some cases failed for this reason
`2025-05-21 15:10:29.319[15904][ERROR] lisa.env[customized_0] case failed
  File "C:\code\lisa-lsg\lsg-lisa\lisa\lisa\runners\lisa_runner.py", line 331, in _initialize_environment_task
    environment.status == EnvironmentStatus.Connected
AssertionError: actual: EnvironmentStatus.Deployed
2025-05-21 15:10:29.321[15904][INFO] lisa.runner[0] 'customized_0' attached to test case 'TimeSync.verify_timedrift_corrected(lisa_0_48)': actual: EnvironmentStatus.Deployed
`

squirrelsc (approved):

PR #3804, opened 2025-05-08
Fixed an issue when ltp_test is specified with empty value

paulli2017:
Fixed an issue when ltp_test is specified with empty value

### DIFF
--- a/microsoft/testsuites/ltp/ltpsuite.py
+++ b/microsoft/testsuites/ltp/ltpsuite.py
@@ -63,7 +63,9 @@ class LtpTestsuite(TestSuite):
         variables: Dict[str, Any],
         result: TestResult,
     ) -> None:
-        test_str = variables.get("ltp_test", self.LTP_LITE_TESTS)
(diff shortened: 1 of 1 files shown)


microsoft/testsuites/ltp/ltpsuite.py
```diff
@@ -95,8 +97,7 @@ def verify_ltp_full(
         variables: Dict[str, Any],
         result: TestResult,
     ) -> None:
-        test_str = variables.get("ltp_test", "")
-        self._verify_ltp(node, variables, test_str, log_path, result)
+        self._verify_ltp(node, variables, "", log_path, result)
```
squirrelsc: Keep this unchanged. This test case should still take parameters.

paulli2017: The change is done, thanks

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3805, opened 2025-05-09
Fixed the regression error caused by support_sudo change

paulli2017:


squirrelsc:
Reviewing offline to understand the scenario.

lisa/node.py
```diff
@@ -134,33 +134,29 @@ def support_sudo(self) -> bool:
...
         process = self._execute("command -v sudo", shell=True, no_info_log=True)
         result = process.wait_result(10)
-        if result.exit_code != 0:
-            self.log.debug("node doesn't support 'sudo', may cause failure later.")
-            return False
+        if result.exit_code == 0:
+            support_sudo = True
 
         # Further test: try running 'ls' with sudo /bin/sh
         process = self._execute("ls", shell=True, sudo=True, no_info_log=True)
```
squirrelsc, line 156: Please call `check_sudo_password_required` before this command, see if it works.

squirrelsc: The `check_sudo_password_required` is in RemoteNode, so it may need to move to here. @lubaihua33 do you remember why the `check_sudo_password_required` is added in RemoteNode, not in the Node?

paulli2017: Added check_sudo_password_required() call helped solved the issue on Overlake. The change is done. thanks.

squirrelsc:
@paxue please continue to work on this change. Thanks.

lisa/node.py
lubaihua33: > The `check_sudo_password_required` is in RemoteNode, so it may need to move to here. @lubaihua33 do you remember why the `check_sudo_password_required` is added in RemoteNode, not in the Node?

Hi @squirrelsc, we found this password issue in RemoteNode scenario. So I just added in RemoteNode.

squirrelsc: ok, it can be moved to node level. Is it right?

lubaihua33: Yes, I think so. Moving to the node level needs some changes. I'm not quite sure about the scenario in this PR. We'd better understand the scenario and see if moving to node level is needed.

lisa/node.py
```diff
@@ -157,8 +156,13 @@ def _check_sudo_available(self) -> bool:
...
         result = process.wait_result(10)
         if result.exit_code != 0:
-            self.log.debug("node doesn't support sudo /bin/sh.")
-            return False
+            self.log.debug(
+                f"_check_sudo_available: 'sudo /bin/sh ls' not succeeded,stderr: {result.stderr}"
```
squirrelsc: It looks this line is not necessary. The command log outputs the raw output already. And when it runs remotely, the stderr is empty. Please read the generated log to make sure no duplicate logs, and the content output is expected.

lisa/node.py
```diff
@@ -157,8 +156,13 @@ def _check_sudo_available(self) -> bool:
...
+            )
+            if "not allowed" in result.stderr:
+                # e.g. OS Security: "user is not allowed to execute '/bin/sh -c"
+                self.log.debug("node doesn't support sudo /bin/sh.")
```
squirrelsc: Here’s a clearer version:
"The command ’sudo /bin/sh -c‘ ls may fail due to SELinux policies that restrict the use of sudo in combination with /bin/sh."

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3809, opened 2025-05-20
Change/fix in the rg deletion condition

kanchansenlaskar:
At present, when there is a deployment failure, the environment gets deleted even if the keep_environment value is set to alway or failed. This PR is to change that and set the below expectations:
keep_environment: always
	Deployment failure - retain environment
	Test failure - retain environment
	Test Pass - retain environment

keep_environment: failed
	Deployment failure - retain environment
	Test failure - retain environment
	Test Pass - delete environment
	
keep_environment: no
	Deployment failure - delete environment
	Test failure - delete environment
	Test Pass - delete environment

### DIFF
--- a/lisa/platform_.py
+++ b/lisa/platform_.py
@@ -181,7 +181,11 @@ class Platform(subclasses.BaseClassWithRunbookMixin, InitializableMixin):
         log.info(f"deploying environment: {environment.name}")
         timer = create_timer()
         environment.platform = self
-        self._deploy_environment(environment, log)
+        try:
+            self._deploy_environment(environment, log)
+        except Exception as identifier:
+            environment.status = EnvironmentStatus.Bad
+            raise identifier
         environment.status = EnvironmentStatus.Deployed
 
         # initialize features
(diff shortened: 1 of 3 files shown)


lisa/platform_.py
```diff
@@ -214,10 +214,14 @@ def delete_environment(self, environment: Environment) -> None:
         # deleting, it should be ignored.
         environment.status = EnvironmentStatus.Deleted
         environment.cleanup()
-        if self.runbook.keep_environment == constants.ENVIRONMENT_KEEP_ALWAYS:
+        if self.runbook.keep_environment in {
+            constants.ENVIRONMENT_KEEP_ALWAYS,
+            constants.ENVIRONMENT_KEEP_FAILED,
```
squirrelsc: How to make sure the env is deleted, when the test passed, but keep_env is `failed`?

kanchansenlaskar: I have pushed a new change.
And tested the below:
keep_environment: always
	Deployment failure - retains environment
	Test failure - retains environment
	Test Pass - retains environment

keep_environment: failed
	Deployment failure - retains environment
	Test failure - retains environment
	Test Pass - deletes environment
	
keep_environment: no
	Deployment failure - deletes environment
	Test failure - deletes environment
	Test Pass - deletes environment

LiliDeng:
@kanchansenlaskar  When a VM deployment fails due to a provisioning timeout, Azure automatically rolls back the deployment by deleting the entire resource group, ignoring the keep_environment setting - unless a Delete lock is in place on the resource group. If your code is intended for this purpose, it won't work as expected in this situation.

squirrelsc:
> @kanchansenlaskar When a VM deployment fails due to a provisioning timeout, Azure automatically rolls back the deployment by deleting the entire resource group, ignoring the keep_environment setting - unless a Delete lock is in place on the resource group. If your code is intended for this purpose, it won't work as expected in this situation.

It's still valid, because the existing code deletes envs, if envs' initialization is failed. Do you know, if there is a way to prevent auto rollback rather than lock the rg?

kanchansenlaskar:
> > @kanchansenlaskar When a VM deployment fails due to a provisioning timeout, Azure automatically rolls back the deployment by deleting the entire resource group, ignoring the keep_environment setting - unless a Delete lock is in place on the resource group. If your code is intended for this purpose, it won't work as expected in this situation.
> 
> It's still valid, because the existing code deletes envs, if envs' initialization is failed. Do you know, if there is a way to prevent auto rollback rather than lock the rg?

Can we go ahead with the current change in this PR and perform the next implementation (of stopping the auto rollback) as part of a separate PR once the solution is decided? I am at present not aware of any way to prevent the auto rollback

lisa/platform_.py
```diff
@@ -231,6 +241,7 @@ def delete_environment(self, environment: Environment) -> None:
             if remote_addresses:
                 log.info(f"node ip addresses: {remote_addresses}")
         else:
+            environment.status = EnvironmentStatus.Deleted
```
squirrelsc, line 248: What's the reason the move this line? It's in both if/else, there is no difference with before.

kanchansenlaskar: it is because we are putting "environment.status == EnvironmentStatus.Bad" for keep_environment "failed" in the if condition

squirrelsc: You change removed L215, and added L225 and L244. They are equal.

lisa/runners/lisa_runner.py
```diff
@@ -470,6 +470,7 @@ def _delete_environment_task(
             # The prepared only environment doesn't need to be deleted.
             # It may cause platform fail to delete non-existing environment.
             environment.status = EnvironmentStatus.Deleted
+            self._log.debug(f"Marking the environment {environment.name} as deleted.")
```
squirrelsc: remove this log, the console notifier will log the information.

lisa/platform_.py
kanchansenlaskar: that is because I am setting the value to EnvironmentStatus.Bad in file lisa/platform_.py L187, and the same value I am looking for in the if condition in L223.
I can use a separate variable to capture the old value (which can be EnvironmentStatus.Bad) and use that in the if condition of L223 and set the env status to deleted like earlier, if that is what you would prefer,

kanchansenlaskar: infact let me make that change.

kanchansenlaskar: kindly check now

squirrelsc: The new logic is clearer. But you can move this line to the end of the if/else logic. It makes more sense to set it to deleted, after the _delete_env called. But it needs to add a try/except for the _delete_env method to catch the errors earlier, and remove the try/except in lisa_runner, when calling the delete_env.

kanchansenlaskar: The reason for marking the environment.status as EnvironmentStatus.Deleted at the very beginning is mentioned in the L217 and L218 comment. Hence I did not change the order. Please let me know your thought on the same.

squirrelsc: Please try below,

1. Move the `environment.status = EnvironmentStatus.Deleted` to end of the `if/else` block.
2. Add try/except on `self._delete_environment(environment, log)`, so even delete failed in subclasses, the status is also updated.

kanchansenlaskar: done

squirrelsc:
@LiliDeng LGTM
@kanchansenlaskar please fix conflicts with main branch.

kanchansenlaskar:
> @LiliDeng LGTM @kanchansenlaskar please fix conflicts with main branch.

I have resolved the conflict

lisa/platform_.py
```diff
@@ -210,29 +214,38 @@ def deploy_environment(self, environment: Environment) -> None:
...
+        try:
+            environment.cleanup()
+            if (self.runbook.keep_environment == constants.ENVIRONMENT_KEEP_ALWAYS) or (
+                self.runbook.keep_environment == constants.ENVIRONMENT_KEEP_FAILED
+                and environment.status == EnvironmentStatus.Bad
+            ):
+                log.info(
+                    f"skipped to delete environment {environment.name}, "
+                    "as on runbook, keep_environment value "
+                    f"is set to {self.runbook.keep_environment}"
```
LiliDeng: `f"is set to {self.runbook.keep_environment} "`

kanchansenlaskar: done

LiliDeng (approved):

PR #3813, opened 2025-05-21
test: Add mshv crash test

smit-gardhariya:
Add testcase to verify crash functionality having crash generated from mshv.

Re-use kdump setup of LISA to run the test and verify if crashdump is generated.

### DIFF
--- a/lisa/tools/kdump.py
+++ b/lisa/tools/kdump.py
@@ -1,29 +1,38 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT license.
+from __future__ import annotations
 
 import math
 import re
-from pathlib import PurePath, PurePosixPath
+from pathlib import Path, PurePath, PurePosixPath
 from time import sleep
-from typing import TYPE_CHECKING, Any, List, Type
+from typing import TYPE_CHECKING, Any, List, Type, cast
 
+from func_timeout import FunctionTimedOut, func_set_timeout  # type: ignore
 from semver import VersionInfo
 
 from lisa.base_tools import Cat, Sed, Service, Wget
 from lisa.executable import Tool
 from lisa.operating_system import CBLMariner, Debian, Oracle, Posix, Redhat, Suse
 from lisa.tools import Find, Gcc
+from lisa.tools.df import Df
+from lisa.tools.dmesg import Dmesg
+from lisa.tools.echo import Echo
+from lisa.tools.free import Free
 from lisa.tools.lsblk import Lsblk
 from lisa.tools.lscpu import Lscpu
 from lisa.tools.make import Make
+from lisa.tools.stat import Stat
 from lisa.tools.sysctl import Sysctl
 from lisa.tools.tar import Tar
 from lisa.util import LisaException, SkippedException, UnsupportedDistroException
+from lisa.util.perf_timer import create_timer
+from lisa.util.shell import try_connect
 
 from .kernel_config import KernelConfig
 
 if TYPE_CHECKING:
-    from lisa.node import Node
+    from lisa.node import Node, RemoteNode
 
 
 class Kexec(Tool):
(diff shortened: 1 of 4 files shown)


microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -56,6 +72,9 @@ class KdumpCrash(TestSuite):
     trigger_kdump_cmd = "echo c > /proc/sysrq-trigger"
     is_auto = False
 
+    mshv_debug_sysfs = "/sys/kernel/debug/mshv/hvdbg"
+    mshv_crash_command = f"echo 0x4856434f5245 > {mshv_debug_sysfs}"
```
LiliDeng: add comments for the meaning of `0x4856434f5245`

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,138 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+        log: Logger,
+        node: Node,
+        variables: Dict[str, Any],
+        log_path: Path,
+    ) -> None:
+        chkbinpath = variables.get("mshv_chk_bin", "")
+        chkloaderpath = variables.get("mshv_chk_loader", "")
+        if not chkbinpath or not chkloaderpath:
+            raise SkippedException(
+                "Requires a path to MSHV binaries to be passed."
```
LiliDeng: `"Requires a path to MSHV binaries to be passed. "`

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: There is no meaning actually. This is expected value by the sysfs to trigger crash in hv bins. It compares if the value is set to this or not, If so, it will trigger the hv crash.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Done

microsoft/testsuites/kdump/kdumpcrash.py
LiliDeng: then any value can be set here?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+
+        chk_loader_dir = "chk_loader"
+        chk_loader_folder = f"/tmp/{chk_loader_dir}"
+        chk_loader_src_path = f"/tmp/{chk_loader_dir}.tar.gz"
+
+        # Copy artifacts on to the node
+        node.shell.copy(PurePath(chkbinpath), PurePath(chk_bin_src_path))
+        node.shell.copy(PurePath(chkloaderpath), PurePath(chk_loader_src_path))
+        node.execute("sudo ls -lrt /tmp/chk_loader")
+        node.execute("sudo ls -lrt /tmp/chk_bin")
```
LiliDeng: Can we use Ls tool here, and specify run as sudo as parameter for method execute?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
         self.is_auto = True
         self._kdump_test(node, log_path, log)
 
+    @TestCaseMetadata(
+        description="""
+        This test case will
+        1. replace FRE bins with CHK bins and reboot the VM
```
LiliDeng: `Replace FRE bins with CHK bins and reboot the VM`
 If possible use full name for FRE and CHK?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+            raise SkippedException(
+                "Requires a path to MSHV binaries to be passed. "
+                "Please set mshv_chk_bin and mshv_chk_loader testcase variable."
+            )
+        if not isinstance(node.os, CBLMariner):
+            raise SkippedException(
+                f"Testcase only support Mariner-MSHV. Found: {node.os}"
+            )
+
+        mshv = node.tools[Ls].path_exists("/dev/mshv", sudo=True)
```
squirrelsc: If all above conditions meet, including the paths specified and it's the Mariner, can the test case just fail without mshv?

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+        if not mshv:
+            raise SkippedException("MSHV not present")
```
squirrelsc: Give more information, how the MSHV is checked.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+
+        log.info(f"chkbinpath: {chkbinpath}")
```
LiliDeng: make it as debug level log?

squirrelsc: 1. Use one line, instead of two lines. It saves people's attentions. 
2. Move them above, after the variables are read.
3. Use the test case variable names, instead of internal names.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+        chk_bin_dir = "chk_bin"
+        chk_bin_folder = f"/tmp/{chk_bin_dir}"
+        chk_bin_src_path = f"/tmp/{chk_bin_dir}.tar.gz"
+
+        chk_loader_dir = "chk_loader"
+        chk_loader_folder = f"/tmp/{chk_loader_dir}"
+        chk_loader_src_path = f"/tmp/{chk_loader_dir}.tar.gz"
+
+        # Copy artifacts on to the node
+        node.shell.copy(PurePath(chkbinpath), PurePath(chk_bin_src_path))
```
squirrelsc: Use RemoteCopy tool.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+            pattern=regex,
+            single_line=False,
+        )
+        chainloader_config = chainloader_grub_data.get("chainloader_cfg", "").strip()
+        err_msg = f"Cannot get chainloader config, got {chainloader_config}"
+        assert chainloader_config, err_msg
+        if hv_debug_option not in chainloader_config:
+            node.tools[Sed].substitute(
+                regexp="MSHV_SEV_SNP=TRUE",
+                replacement=f"MSHV_SEV_SNP=TRUE {hv_debug_option}",
```
LiliDeng: add comments for why need to add hv_debug_option

microsoft/testsuites/kdump/kdumpcrash.py
LiliDeng: explian more context here, why skip when MSHV doesn't exist

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +242,133 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+                file=grub_config_file,
+                sudo=True,
+            )
+
+        node.tools[Reboot].reboot_and_check_panic(log_path)
+
+        # Trigger HV crash and check if dump is generated
+        hvdbg = node.tools[Ls].path_exists(self.mshv_debug_sysfs, sudo=True)
```
squirrelsc: It looks a lot of system things are changed, please mark the env to dirty, so it won't be reused by other test cases.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: No. This is the specific value needed to be set to trigger the crash.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: This was temporary. I removed it.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: FRE / CHK is the term we use for hv builds. FRE is the release build and CHK is used for debug. Some of the things like hv crash will work with CHK build only. There is not full form as such. It is free/check build. I added this in description.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: yes, it is must to present. If /dev/mshv is not present means it is not running with our kernel. In that case, we will not have the sysfs entry present. Only Mariner-MSHV distro of ours will have it.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Our kernel will create the /dev/mshv interface. This check is just to make sure that test will run with Mariner only when /dev/mshv (meaning it is running with our kernel) is present.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Done

smit-gardhariya: done

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Done

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Done. Added as comment before line `hv_debug_option = "LXHVLOADER_DEBUG=TRUE"`

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: I think i already addressed this above in Chi's comment. We will skip because MSHV will be present only with our kernel where sysfs entry is present. This is internal kernel and not upstream one.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: done

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Added description for it now

microsoft/testsuites/kdump/kdumpcrash.py
pupacha: May be a comment explaining that sysfs expects this value to trigger a crash will be helpful.

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +243,147 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+            raise SkippedException(
+                "Requires a path to MSHV binaries to be passed. "
+                "Please set mshv_chk_bin and mshv_chk_loader testcase variable."
+            )
+        if not isinstance(node.os, CBLMariner):
+            raise SkippedException(
+                f"Testcase only support Mariner-MSHV. Found: {node.os}"
+            )
+
+        # Check if /de/mshv is present to make sure node is running with
```
pupacha: nit: `/dev/mshv`

smit-gardhariya: done

pupacha:
While I agree that the testcase is to test and check for kdumpcrash, I feel it would be more aligned for this test to be present in `testsuites/mshv` path. This way, all mshv specific tests are under a single `category / area`.
What do you think? 
@smit-gardhariya @anirudhrb @LiliDeng @squirrelsc

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: done

microsoft/testsuites/kdump/kdumpcrash.py
```diff
@@ -223,6 +245,147 @@ def verify_kdumpcrash_large_memory_auto_size(
...
+        if not hvdbg:
+            raise LisaException(f"sysfs entry not present: {self.mshv_debug_sysfs}")
+        self.trigger_kdump_cmd = self.mshv_crash_command
+        self._kdump_test(
+            node=node,
+            log_path=log_path,
+            log=log,
+        )
+
+        node.mark_dirty()
```
LiliDeng: We already have the logic in after_case, so no need this step. 
https://github.com/microsoft/lisa/blob/1d865b0262c277014e81d9b99fdfb7216beda9ec/microsoft/testsuites/kdump/kdumpcrash.py#L665

smit-gardhariya:
> While I agree that the testcase is to test and check for kdumpcrash, I feel it would be more aligned for this test to be present in `testsuites/mshv` path. This way, all mshv specific tests are under a single `category / area`. What do you think? @smit-gardhariya @anirudhrb @LiliDeng @squirrelsc

Here Idea is to leverage kdump implementation of LISA. It might be related to MSHV the way we crash the system but here we are crashing hv which requires crash kernel configured. Apart from the hv setup and command used to crash, all pieces are similar to guest kernel crash.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Removed it.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: As Lili mentioned in one of her comments, after_case already mark the node dirty so i removed the explicit `node.mark_dirty()`

pupacha:
> Here Idea is to leverage kdump implementation of LISA. It might be related to MSHV the way we crash the system but here we are crashing hv which requires crash kernel configured. Apart from the hv setup and command used to crash, all pieces are similar to guest kernel crash.

Existing methods already present in LISA can be made as utility functions and be re-used in our test. So, it should be fine.

smit-gardhariya:
> > Here Idea is to leverage kdump implementation of LISA. It might be related to MSHV the way we crash the system but here we are crashing hv which requires crash kernel configured. Apart from the hv setup and command used to crash, all pieces are similar to guest kernel crash.
> 
> Existing methods already present in LISA can be made as utility functions and be re-used in our test. So, it should be fine.

Please check now

microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -77,6 +90,155 @@ def verify_mshvlog_is_active(
...
+
+        kdump_util = KdumpUtils(crashdump_timeout=800, kdump_cmd=mshv_crash_command)
+        kdump_util._kdump_test(
+            node=node,
+            log_path=log_path,
+            log=log,
+            is_auto=False,
+        )
+
+        node.mark_dirty()
```
LiliDeng: It might not work if placed here. Perhaps using a try-catch-finally block would help?

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: Sorry, I didn’t express my point clearly earlier. The current logic skips the test case if /dev/mshv is not found. However, since the conditions above assume that MSHV-related files should exist, shouldn't the absence of /dev/mshv cause the test case to fail instead?

Also, the error message is a bit confusing. The node.os value is always "CBLMariner". It never shows as "Mariner-MSHV". Could we clarify that as well?

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: Add the description in the exception message, so when the test case skipped, the reason is clear in the test results.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: Use one line log, instead two lines for two variables.

microsoft/testsuites/kdump/kdumputils.py
```diff
@@ -0,0 +1,307 @@
+# Copyright (c) Microsoft Corporation.
```
squirrelsc: Move it to tools folder.

microsoft/testsuites/kdump/kdumputils.py
```diff
@@ -0,0 +1,307 @@
...
+)
+from lisa.features import Disk, SerialConsole
+from lisa.operating_system import Redhat
+from lisa.tools import Df, Dmesg, Echo, KdumpBase, KernelConfig, Lscpu, Stat
+from lisa.tools.free import Free
+from lisa.util.perf_timer import create_timer
+from lisa.util.shell import try_connect
+
+
+class KdumpUtils:
```
squirrelsc: Change the class name to `Kdump`.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: This test intends to run with our private CBLMariner distro which comes with /dev/mshv. For other distro, this test should skip.

If test run with public CBLMariner distro, which do not have /dev/mshv, test should skip as well. Private mariner build of ours only will have the /dev/mshv available as it depends on the kernel we ship with private build.

I modified the exception msg now.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Modified the exception message.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: Done

microsoft/testsuites/mshv/mshv_root_tests.py
smit-gardhariya: Done. Please check

microsoft/testsuites/kdump/kdumputils.py
smit-gardhariya: Done. Moved it as kdumputils.py under tools

microsoft/testsuites/kdump/kdumputils.py
smit-gardhariya: Done

microsoft/testsuites/kdump/kdumputils.py
smit-gardhariya: @squirrelsc I have not created tool out of it. Should we create tool?

squirrelsc: Yes, please create a tool, don't share tools between different test suites.

microsoft/testsuites/kdump/kdumpcrash.py
squirrelsc: I'm concerned that some bugs might be missed. For example, what happens if a build includes the MSHV kernel but omits this specific file? Is there a way to distinguish between an MSHV build that missed the file and a non-MSHV build?

My thinking is: if you have two separate pipelines for the different build types, or if you can pass in a variable indicating the build type, then you could always raise an exception here. But if there's no reliable way to differentiate the two build types and this step is skipped, then a missing file in an MSHV build might go undetected.

lisa/tools/kdumputils.py
```diff
@@ -0,0 +1,307 @@
+# Copyright (c) Microsoft Corporation.
```
squirrelsc: Add an entry in the lisa/tools/__init__.py like other tools.

microsoft/testsuites/kdump/kdumpcrash.py
smit-gardhariya: When our kernel is up, it will always create /dev/mshv. If it is not present with our distro, that means something is wrong in kernel. We can check kernel logs in such cases. 

Only difference in public CBLMariner and private build is /dev/mshv. I can add some check of dmesg to verify if this is our private build or not, but it is just that, those messages would appear on successful creation of /dev/mshv only so we are checking the path. Similar piece of code is used for cloud-hypervisor perf test of LISA as well.

My thought process was, if we fail the test if /dev/mshv is not present then it will fail on public CBLMariner too which should ideally be skipped. I dont want to fail the test until necessary.

lisa/tools/kdumputils.py
smit-gardhariya: Done. Corresponding Import are changed too.

lisa/tools/kdumputils.py
```diff
@@ -0,0 +1,317 @@
...
```
squirrelsc: I didn't noticed the file name is not changed. It needs to be same to the class name "kdump.py", if it has only the one class.

smit-gardhariya: we already have kdump.py under tools that is why i kept kdumputils as this is not kdump tool. This is some utility functions which helps to run kdump. Let me know if we move back to kdumputils.

squirrelsc: I see. I hadn't noticed that. The name Kdump doesn't seem appropriate in this context. Would it be possible to move this tool into kdump.py and rename it to something more descriptive? The current name "utils" is a bit ambiguous.

smit-gardhariya: Ideally this is not tool. All the utility functions we have under kdumputils.py is to execute KDUMP tool of LISA with some pre/post checks. That is why initially i kept it under testcase. Also we have different kdump tool flavors for different distro where as utility functions only use KdumpBase tool (parent tool).

If we want to integrate this utility functions to tool level then we have to determine which flavor of kdump tool we should use based on distro and that might require again some change in utility functions to cover specific scenarion if required any while performing pre/post checks of kdump aspect.

Scope for this PR is to re-use kdump pre/post checks for kdumpcrash and MSHV testsuites.

squirrelsc: 1. A LISA tool can be virtual—it might consist of a set of scripts or commands rather than a single executable.
1. Code sharing between test suites should be avoided, as it complicates maintenance and can introduce circular dependencies.
1. A LISA tool may rely on different executables depending on the distribution—for example, variations in how firewall or service management is handled.

Please consider a better name to make it clear in the kdump.py, like KdumpCheck or something else?

smit-gardhariya: Hi @squirrelsc

I created KdumpCheck tool which is virtual tool as suggested. It uses KdumpBase to trigger kernel crash and check kdump functionality similar way it did earlier. kdumpcrash/mshv crash will use this KdumpCheck tool to trigger crash now.

All the pre/post checks are part of KdumpCheck tool. Please check and let me know

lisa/tools/kdump.py
```diff
@@ -729,3 +738,300 @@ def capture_info(self) -> None:
...
+
+    @property
+    def command(self) -> str:
+        return ""
+
+    @property
+    def can_install(self) -> bool:
+        return False
+
+    def _check_exists(self) -> bool:
```
squirrelsc: move private methods after public methods.

smit-gardhariya: @squirrelsc Done

microsoft/testsuites/mshv/mshv_root_tests.py
```diff
@@ -77,6 +90,160 @@ def verify_mshvlog_is_active(
...
+                raise LisaException(f"sysfs entry not present: {mshv_debug_sysfs}")
+
+            kdump_util = node.tools[KdumpCheck]
+            kdump_util.kdump_test(
+                log_path=log_path,
+                trigger_kdump_cmd=mshv_crash_command,
+                is_auto=False,
+            )
+        except Exception as e:
+            raise LisaException(f"verify_mshv_crash failed: {e}")
```
squirrelsc: The new exception error doesn't provide additional information, so the entire except block can be removed.

smit-gardhariya: @squirrelsc Removed it in latest push.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3828, opened 2025-05-28
Add verify_os_architecture_is_64bit test case

lubaihua33:


LiliDeng (dismissed):

lisa/base_tools/uname.py
```diff
@@ -80,6 +80,11 @@ def get_linux_information(
 
         return result
 
+    def get_machine_architecture(self, force_run: bool = False) -> str:
```
squirrelsc: Can it reuse the arch enum in lscpu?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1365,6 +1366,29 @@ def verify_openssl_version(self, node: Node) -> None:
...
+        Steps:
+        1. Retrieve the OS architecture using the Uname tool.
+        2. Verify that the architecture is either x86_64 (AMD64) or aarch64 (ARM64).
+        3. Fail the test if the architecture is not 64-bit.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_platform_type=[AZURE]),
+    )
+    def verify_os_architecture_is_64bit(self, node: Node) -> None:
+        uname_tool = node.tools[Uname]
```
squirrelsc, line 1382: Why choose uname, is it the most widely supported?

microsoft/testsuites/core/azure_image_standard.py
```diff
@@ -1365,6 +1366,29 @@ def verify_openssl_version(self, node: Node) -> None:
...
+
+        Steps:
+        1. Retrieve the OS architecture using the Uname tool.
+        2. Verify that the architecture is either x86_64 (AMD64) or aarch64 (ARM64).
+        3. Fail the test if the architecture is not 64-bit.
+        """,
+        priority=1,
+        requirement=simple_requirement(supported_platform_type=[AZURE]),
+    )
+    def verify_os_architecture_is_64bit(self, node: Node) -> None:
```
squirrelsc: How about a shorter name like `verify_azure_64bit_os`?

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: `uname -m` is a very simple command to get the machine architecture of a system and it is most widely supported like "ls, cp, mv, cat" which are all from `coreutils` package.

The output of `uname -m` is just `x86_64` or `aarch64`.

lisa/base_tools/uname.py
lubaihua33: Yes, we can reuse lscpu.get_architecture(). 
However, I don't think this is more widely supported than "uname -m". And the assert message in get_architecture is not clear because it doesn't specify Azure.

microsoft/testsuites/core/azure_image_standard.py
lubaihua33: Done

lisa/base_tools/uname.py
squirrelsc: How about just reuse the enumerator class, not the lscpu itself. So it doesn't need to return str in this method.

lubaihua33: Done

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3831, opened 2025-05-30
Update xfstests version from master branch to specified version branch.

shekharsorot:
Roll back XFStest tag from master to v2024.02.09.
Will subsequently investigate stable tags that do not case tests to crash


LiliDeng (approved):

squirrelsc:
Please make title and description clear. I've updated it.

squirrelsc:
BTW, how about upgrading to a newer version?

PR #3835, opened 2025-06-02
Package dependencies to run lisa on Fedora Distribution

balakreddy:
Added dependent packages required to run lisa on fedora.
Tested by running lisa locally and on azure cloud.


squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3837, opened 2025-06-03
Modprobe Tool: Improve failure debugging

adityagesh:
If Modprobe fails to remove a module, run some
additional commands to debug why it failed

### DIFF
--- a/lisa/tools/dmesg.py
+++ b/lisa/tools/dmesg.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT license.
 
 import re
-from typing import List
+from typing import List, Optional
 
 from semver import VersionInfo
 
@@ -49,11 +49,26 @@ class Dmesg(Tool):
     def _check_exists(self) -> bool:
         return True
 
-    def get_output(self, force_run: bool = False) -> str:
-        command_output = self._run(force_run=force_run)
+    def get_output(
+        self,
+        force_run: bool = False,
+        no_error_log: bool = True,
+        no_info_log: bool = True,
+        no_debug_log: bool = False,
+        tail_lines: Optional[int] = None,
+    ) -> str:
+        command_output = self._run(
+            force_run=force_run,
+            no_error_log=no_error_log,
+            no_info_log=no_info_log,
+            no_debug_log=no_debug_log,
+        )
 
         # Remove the color code from stdout stream
         stdout = filter_ansi_escape(command_output.stdout)
+        if tail_lines is not None:
+            stdout_lines = stdout.splitlines()
+            stdout = "\n".join(stdout_lines[-tail_lines:])
         return stdout
 
     def check_kernel_errors(
@@ -104,12 +119,29 @@ class Dmesg(Tool):
                 return VersionInfo(int(major), int(minor))
         raise LisaException("No find matched vmbus version in dmesg")
 
-    def _run(self, force_run: bool = False) -> ExecutableResult:
-        # sometime it need sudo, we can retry
-        # so no_error_log for first time
-        result = self.run(force_run=force_run, no_error_log=True)
+    def _run(
+        self,
+        force_run: bool = False,
+        no_error_log: bool = True,
+        no_info_log: bool = True,
+        no_debug_log: bool = False,
+    ) -> ExecutableResult:
+        # no_error_log is set to True for the first run because
+        # it will fail for distros that require sudo to run dmesg.
+        result = self.run(
+            force_run=force_run,
+            no_error_log=True,
+            no_info_log=no_info_log,
+            no_debug_log=no_debug_log,
+        )
         if result.exit_code != 0:
             # may need sudo
-            result = self.run(sudo=True, force_run=force_run)
+            result = self.run(
+                sudo=True,
+                force_run=force_run,
+                no_error_log=no_error_log,
+                no_info_log=no_info_log,
+                no_debug_log=no_debug_log,
+            )
         self._cached_result = result
         return result
--- a/lisa/tools/journalctl.py
+++ b/lisa/tools/journalctl.py
@@ -53,3 +53,40 @@ class Journalctl(Tool):
             expected_exit_code=0,
         )
         return result.stdout
+
+    def filter_logs_by_pattern(
+        self,
+        pattern: str,
+        tail_line_count: int = 1000,
+        no_debug_log: bool = False,
+        no_error_log: bool = False,
+        no_info_log: bool = True,
+        sudo: bool = True,
+    ) -> str:
+        """
+        Filters journalctl logs by a given pattern using grep.
+
+        Args:
+            pattern (str): Passed directly to grep. Use a valid grep pattern
+                (string or regex).
+            tail_line_count (int): Optionally limits the output to the last N lines.
+        """
+        cmd = f"--no-pager | grep '{pattern}'"
+
+        if tail_line_count > 0:
+            cmd = cmd + f" | tail -n {tail_line_count}"
+
+        result = self.run(
+            cmd,
+            force_run=True,
+            shell=True,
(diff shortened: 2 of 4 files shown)


lisa/tools/modprobe.py
```diff
@@ -154,6 +162,58 @@ def load_by_file(
...
+            no_debug_log=True,
+        )
+        if not module_exists:
+            self._log.info(f"Module {mod_name} does not exist in lsmod output.")
+        else:
+            self._log.info(f"Module {mod_name} exists in lsmod output.")
+            used_by_modules = lsmod_tool.get_used_by_modules(mod_name)
+            if used_by_modules:
+                used_by_str = ", ".join(used_by_modules)
+                self._log.info(f"Module {mod_name} is used by: " f"{used_by_str}")
```
LiliDeng: ```self._log.info(f"Module {mod_name} is used by: {used_by_str}")```

lisa/tools/modprobe.py
```diff
@@ -154,6 +162,58 @@ def load_by_file(
...
+                self._log.info(f"Module {mod_name} is not used by any other modules.")
+
+    def _tail_dmesg_log(self) -> None:
+        dmesg_tool = self.node.tools[Dmesg]
+        dmesg_output = dmesg_tool.get_output(force_run=True, no_debug_log=True)
+        dmesg_lines = dmesg_output.splitlines()
+        tail_lines = 20
+        recent_dmesg_output = "\n".join(dmesg_lines[-tail_lines:])
+        self._log.info(
+            f"Last {tail_lines} dmesg lines after modprobe:\n" f"{recent_dmesg_output}"
```
LiliDeng: ditto

lisa/tools/modprobe.py
```diff
@@ -154,6 +162,58 @@ def load_by_file(
...
+            mod_name,
+            tail_line_count=tail_lines,
+            no_debug_log=True,
+            no_error_log=True,
+            no_info_log=True,
+        )
+        self._log.info(
+            f"Last {tail_lines} journalctl lines for module {mod_name}:\n"
+            f"{journalctl_out}"
+        )
```
LiliDeng, line 219: use debug instead of info?

lisa/tools/modprobe.py
```diff
@@ -154,6 +162,58 @@ def load_by_file(
...
+                used_by_str = ", ".join(used_by_modules)
+                self._log.info(f"Module {mod_name} is used by: " f"{used_by_str}")
+            else:
+                self._log.info(f"Module {mod_name} is not used by any other modules.")
+
+    def _tail_dmesg_log(self) -> None:
+        dmesg_tool = self.node.tools[Dmesg]
+        dmesg_output = dmesg_tool.get_output(force_run=True, no_debug_log=True)
+        dmesg_lines = dmesg_output.splitlines()
+        tail_lines = 20
```
LiliDeng, line 198: how about pass this as parameter of one method of dmesg? then handle logic inside of dmesg tool?

lisa/tools/modprobe.py
adityagesh: using info was intentional. The idea is to have sufficient logs for failures so that we will be able to triage failures without the need for a repro environment.
This will also help log triaging LLMs to effectively analyze the failure

lisa/tools/modprobe.py
adityagesh: updated

lisa/tools/modprobe.py
adityagesh: sorry, fixed

lisa/tools/modprobe.py
adityagesh: sorry, fixed

lisa/tools/dmesg.py
```diff
@@ -104,12 +119,29 @@ def get_vmbus_version(self) -> VersionInfo:
...
+        no_error_log: bool = True,
+        no_info_log: bool = True,
+        no_debug_log: bool = False,
+    ) -> ExecutableResult:
         # sometime it need sudo, we can retry
         # so no_error_log for first time
-        result = self.run(force_run=force_run, no_error_log=True)
+        result = self.run(
+            force_run=force_run,
+            no_error_log=no_error_log,
```
squirrelsc: The no_error_log setting is originally set to True because many distributions require root permission to access dmesg. If an error occurs, it attempts to use sudo again. Therefore, maintaining the original behavior is preferable unless there is a specific reason to change it.

lisa/tools/journalctl.py
```diff
@@ -53,3 +53,29 @@ def first_n_logs_from_boot(
             expected_exit_code=0,
         )
         return result.stdout
+
+    def filter_logs_by_pattern(
```
squirrelsc, line 57: Add comments above on the pattern format in general.

lisa/tools/journalctl.py
```diff
@@ -53,3 +53,29 @@ def first_n_logs_from_boot(
...
+
+        result = self.run(
+            cmd,
+            force_run=True,
+            shell=True,
+            sudo=sudo,
+            no_debug_log=no_debug_log,
+            no_error_log=no_error_log,
+            no_info_log=no_info_log,
+            expected_exit_code=0,
```
squirrelsc, line 87: Add error message for unexpected exit code.

lisa/tools/lsmod.py
```diff
@@ -48,3 +50,44 @@ def module_exists(
...
+        result = self.run(
+            sudo=True,
+            force_run=True,
+            no_info_log=no_info_log,
+            no_error_log=no_error_log,
+            no_debug_log=no_debug_log,
+            expected_exit_code=0,
+        )
+        used_by_modules = []
+        for match in self.__output_pattern.finditer(result.stdout):
```
squirrelsc: Use LISA built-in regexp wrappers, it makes code simpler.

lisa/tools/modprobe.py
```diff
@@ -74,15 +77,20 @@ def remove(self, mod_names: List[str], ignore_error: bool = False) -> None:
...
+                            force_run=True,
+                            sudo=True,
+                            shell=True,
+                            expected_exit_code=0,
+                            expected_exit_code_failure_message=(
+                                f"Fail to remove module {mod_name}"
+                            ),
+                        )
+                    except AssertionError as e:
+                        self._debug_modprobe_failure(mod_name)
```
squirrelsc: It's not `debug`, it's `collect`. Call it like `_collect_log`.

adityagesh: updated

lisa/tools/lsmod.py
adityagesh: updated

lisa/tools/journalctl.py
adityagesh: added

lisa/tools/journalctl.py
adityagesh: added

lisa/tools/dmesg.py
```diff
@@ -104,12 +119,29 @@ def get_vmbus_version(self) -> VersionInfo:
...
+        force_run: bool = False,
+        no_error_log: bool = True,
+        no_info_log: bool = True,
+        no_debug_log: bool = False,
+    ) -> ExecutableResult:
+        # no_error_log is set to False for the first run because
+        # it will fail for distros that require sudo to run dmesg.
+        result = self.run(
+            force_run=force_run,
+            no_error_log=False,
```
squirrelsc: Keep it to `True` as before

adityagesh: oops, sorry.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3840, opened 2025-06-03
Add hook azure_update_vm_capabilities

kamalca:
Add hook `azure_update_vm_capabilities` to allow editing of Azure VM SKU capabilities before being saved to cache.

This will allow capability information to be inserted by extensions for internal SKUs or the processing of additional SKU information such as disk and network performance. Alternatively, SKUs can be removed from the results to restrict the SKUs that LISA will deploy.

### DIFF
--- a/lisa/sut_orchestrator/azure/platform_.py
+++ b/lisa/sut_orchestrator/azure/platform_.py
@@ -1111,6 +1111,7 @@ class AzurePlatform(Platform):
                     except Exception as e:
                         log.error(f"unknown sku: {sku_obj}")
                         raise e
+            plugin_manager.hook.azure_update_vm_capabilities(capabilities=all_skus)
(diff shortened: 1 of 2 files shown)


lisa/sut_orchestrator/azure/platform_.py
```diff
@@ -1111,6 +1111,7 @@ def get_location_info(self, location: str, log: Logger) -> AzureLocation:
                     except Exception as e:
                         log.error(f"unknown sku: {sku_obj}")
                         raise e
+            plugin_manager.hook.azure_update_vm_capabilities(capabilities=all_skus)
```
squirrelsc, line 1114: How about put it in the condition `should_refresh`? it will improve performance a lot.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3845, opened 2025-06-05
Collect log using SerialConsole when SshShell fails to connect

adityagesh:
ssh may not succeed due to multiple reasons, at present if ssh fails it is either very difficult or impossible to triage the issue without reproducing it.
This change uses NonSshExecutor to run commands for log collection.


adityagesh:
@squirrelsc @LiliDeng , this PR is for early design review.

This will enhance triaging of failures for both engineers and AI

lisa/features/serial_console.py
```diff
@@ -199,3 +199,6 @@ def write(self, data: str) -> None:
     def close(self) -> None:
         # it's not required to implement close method.
         pass
+
+    def execute_command(self, commands: List[str]) -> str:
```
squirrelsc: Improve the `_write` instead of creating another similar method.

lisa/node.py
```diff
@@ -531,7 +532,26 @@ def _initialize(self, *args: Any, **kwargs: Any) -> None:
...
+                # OS is not initialized yet, so cannot use self.os.
+                commands = [
+                    "ip addr show",
+                    "ip link show",
+                    "systemctl status NetworkManager --no-pager --plain",
+                    "systemctl status network --no-pager --plain",
+                    "systemctl status systemd-networkd --no-pager --plain",
+                    "ping -c 3 -n 8.8.8.8",
+                ]
+                serial_console.execute_command(commands=commands)
```
squirrelsc: I'm not sure how much this will help, but here are a couple of points:
Most VMs appear to be stuck at the login screen, so it's not possible to input any commands directly.
We need to wait until the command produces output. If the case fails quickly, the downloaded serial log might not contain some keywords. Please try using get_console_log repeatedly until the serial log has content or a timeout occurs.

adityagesh: Thanks for the comment, Chi.
I will handle the read-retry-timeout scenario.

What do you think is the best way to determine the list of commands to execute?

> I'm not sure how much this will help 

The reason I feel this would help is because, right now when LISA fails with ssh error, we have no insight into what is happening in the VM. If we can craft a list of commands that would collection sufficient log in most scenarios, that itself would be a great reduction in the need for a repro

lisa/features/serial_console.py
adityagesh: Ack.

lisa/node.py
squirrelsc: Did you try it in real cases? Especially, the serial console is not signed in.

adityagesh: I haven't tried the stuck at login prompt scenario yet.
So far I have only tried in scenario where the VM was booted and running, but it had issues with inbound connection.

lisa/features/serial_console.py
```diff
@@ -193,7 +193,7 @@ def check_initramfs(
     def read(self) -> str:
         raise NotImplementedError
 
-    def write(self, data: str) -> None:
+    def write(self, data: str | List[str]) -> None:
```
squirrelsc: Don't change the interface, it's used in other places. If it needs to implement a list of commands, you can join them before calling `write`. It's not like a common and recommended way.

lisa/features/__init__.py
```diff
@@ -64,6 +65,7 @@
     "PasswordExtension",
     "Resize",
     "ResizeAction",
+    "RunCommand",
```
squirrelsc: It's an Azure only feature, so put it in Azure, not shared place.

lisa/features/serial_console.py
squirrelsc: Please remove it.

lisa/features/serial_console.py
adityagesh: Ack

lisa/features/__init__.py
adityagesh: I am still testing the performance of RunCommand vs SerialConsole for collecting logs.

RunCommand seems to be faster than Serial Console (including the time for reset password and login for serial console).

However, since RunCommand is only implemented in Azure platform, should we rather consider using Serial Console for the log collection?

squirrelsc: The ResetPassword feature relies on RunCommand or a similar underlying Azure capability. This means it only functions properly in Azure environments. If password authentication is disabled and access via the serial console is required, ResetPassword won’t work. However, if it offers improvements over RunCommand, we can consider using it.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -3729,3 +3751,47 @@ def _prepare_azure_file_share(
...
+        :param commands: A list of shell commands to execute.
+        :return: The output of the commands.
+        """
+        context = get_node_context(self._node)
+        platform = self._platform
+        compute_client = get_compute_client(platform)
+
+        # Prepare the RunCommandInput for Azure
+        command = RunCommandInput(
+            command_id="RunShellScript",
```
kamalca, line 3776: Is this Linux specific? Consider whether you can add support for Windows. If not possible, think about how to exclude the Feature for non-Posix systems.

lisa/node.py
```diff
@@ -733,6 +741,33 @@ def check_sudo_password_required(self) -> None:
...
+
+        if self.features.is_supported(RunCommand):
+            run_command = self.features[RunCommand]
+            commands = [
+                "echo 'Executing: ip addr show'",
+                "ip addr show",
+                "echo 'Executing: ip link show'",
+                "ip link show",
+                "echo 'Executing: systemctl status NetworkManager --no-pager --plain'",
+                "systemctl status NetworkManager --no-pager --plain",
```
adityagesh: @squirrelsc how do we determine the command(s) to be run? The OS object is not initialized at this point

lisa/features/__init__.py
squirrelsc: 1. Missed the `RunCommand` file?
2. Rename to `NonSshExecutor`, and add comments to explain it. This class is implement by serial console by default, but it can be overwritten to try both serial console and `RunCommand` on Azure.

lisa/node.py
squirrelsc: The Azure RunCommand returns the "provisioning_state", refer to `run_commandv2.py`

lisa/features/non_ssh_executor.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def execute(self, commands: list[str] = COMMANDS_TO_EXECUTE) -> list[str]:
+        """
+        Executes a list of commands on the node and returns their outputs.
+
+        :param commands: A list of shell commands to execute.
+        :return: A string containing the output of the executed commands.
+        """
+        out = []
+        serial_console = self._node.features[SerialConsole]
```
squirrelsc: Check if the feature is not available, raise an exception.

lisa/node.py
```diff
@@ -689,7 +691,20 @@ def set_connection_info(
...
+                vm_logs = self._collect_logs_using_non_ssh_executor()
+                if vm_logs:
+                    self.log.info(
+                        f"Collected information using non-ssh executor:\n{vm_logs}"
+                    )
+            except Exception as log_error:
+                self.log.debug(
+                    f"Failed to collect logs using non-ssh executor: {log_error}"
+                )
+            raise e
```
squirrelsc: `raise`

lisa/node.py
```diff
@@ -733,6 +748,24 @@ def check_sudo_password_required(self) -> None:
...
+        """
+        Collects information using the NonSshExecutor feature.
+        This is used when the connection to the node is not stable.
+        """
+        from lisa.features import NonSshExecutor
+
+        if self.features.is_supported(NonSshExecutor):
+            non_ssh_executor = self.features[NonSshExecutor]
+            out = non_ssh_executor.execute()
+            return "\n".join(out)
```
squirrelsc: Log could be output in this method directly, instead of return and output in upper level.

lisa/features/non_ssh_executor.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class NonSshExecutor(Feature):
+    """
+    NonSshExecutor is used to run commands on the node when SSH is not available.
+    Lisa by default uses SSH for connection, but this feature provides an alternative
+    execution method for scenarios where SSH connectivity is not possible or desired.
+    """
+
+    COMMANDS_TO_EXECUTE = [
```
squirrelsc: The commands are not related to `NonSshExecutor`, they should be moved to node level.

lisa/node.py
```diff
@@ -780,6 +813,26 @@ def _check_bash_prompt(self) -> None:
...
+                raise RequireUserPasswordException(
+                    "The password is not set, and generate is False."
+                )
+            self.log.debug("password is not set, generating a strong password.")
+            if not self._reset_password():
+                raise RequireUserPasswordException("Reset password failed")
+        password = self._connection_info.password
+        if not password:
+            raise RequireUserPasswordException(
+                "The password is not set, and generate is False."
```
squirrelsc: The password has not been set and generated.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -3729,3 +3732,68 @@ def _prepare_azure_file_share(
...
+        """
+        return [f"echo 'Running command: {cmd}' && {cmd}" for cmd in commands]
+
+
+class NonSshExecutor(AzureFeatureMixin, features.NonSshExecutor):
+    def execute(
+        self, commands: list[str] = features.NonSshExecutor.COMMANDS_TO_EXECUTE
+    ) -> list[str]:
+        # RunCommand is faster than SerialConsole. Hence attempt to use it first.
+        try:
```
squirrelsc, line 3810: The overrides should attempt both RunCommand and SerialConsole.

adityagesh: It is attempting to use SerialConsole if RunCommand raises an exception.

 except Exception as e:
            self._log.info(f"RunCommand failed: {e}")
            # Fallback to the default non-SSH executor behavior
            return super().execute(commands)

lisa/features/serial_console.py
```diff
@@ -190,6 +191,34 @@ def check_initramfs(
                 f"{initramfs_logs} {filesystem_exception_logs}"
             )
 
+    def login(self) -> None:
```
squirrelsc: `ensure_login`

lisa/features/non_ssh_executor.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        Executes a list of commands on the node and returns their outputs.
+
+        :param commands: A list of shell commands to execute.
+        :return: A string containing the output of the executed commands.
+        """
+        out = []
+
+        serial_console = self._node.features[SerialConsole]
+        if not serial_console.enabled():
```
squirrelsc: use `self._node.features.is_supported` to check the feature is available or not before `self._node.features[SerialConsole]`

lisa/node.py
```diff
@@ -733,6 +743,34 @@ def check_sudo_password_required(self) -> None:
...
+            non_ssh_executor = self.features[NonSshExecutor]
+            out = non_ssh_executor.execute(commands=commands)
+            out = "\n\n".join(out)
+            self.log.info(f"Collected information using NonSshExecutor:\n{out}")
+        else:
+            self.log.debug(
+                f"NonSshExecutor is not supported on {self.name}, "
+                "cannot collect logs using non-ssh executor."
+            )
+        return None
```
squirrelsc: It doesn't need to `return None` at the end.

adityagesh: My bad, I wasn't thinking straight

adityagesh:
Testing with non-Linux environments are pending.
I will test with one Windows and Freebsd image.

lisa/features/non_ssh_executor.py
```diff
@@ -0,0 +1,50 @@
...
+                "NonSshExecutor requires SerialConsole feature to be supported."
+            )
+        serial_console = self._node.features[SerialConsole]
+        serial_console.ensure_login()
+        # clear the console before executing commands
+        serial_console.write("\n")
+        _ = serial_console.read()
+        for command in commands:
+            serial_console.write(self._add_newline(command))
+            out.append(serial_console.read())
```
squirrelsc: `read()` may not get all output. How about use previous `_ = serial_console.read()` to get prompt, and check if the prompt is printed out.

squirrelsc: besides return the output, also printed them out.

adityagesh: def _collect_logs_using_non_ssh_executor (the caller) is currently printing it.

Should I add an optional argument to print it in this function itself?

adityagesh: Makes sense, added logic

        serial_console.write("\n")
        response = serial_console.read()
        if not response or "$" not in response and "#" not in response:
        
Do you recommend the above or response.strip().endswith()?

squirrelsc: After writing `\n`, it will return a full prompt. So, you can check the full prompt instead of only special chars.

squirrelsc: The method should print by itself, instead of caller. If it's used by others, they don't need to copy code to print. It's similar to how node.execute print out.

adityagesh:
@squirrelsc 
The log is currently printed to node.log and lisa.log. It does not show up in test case log.

lisa.log
image

node.log
image


Do you think we should add it to test case log as well? I think people would miss out this log if it's not in case log

squirrelsc:
> @squirrelsc The log is currently printed to node.log and lisa.log. It does not show up in test case log.
> 
> Do you think we should add it to test case log as well? I think people would miss out this log if it's not in case log

It should be, but the node.initialize happens before test case is chosen. So, it needs some changes on the flow to put the initialize log into test case log.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -3729,3 +3731,82 @@ def _prepare_azure_file_share(
...
+            if value and len(value) > 0 and value[0].get("message"):
+                message = value[0]["message"]
+            else:
+                raise LisaException(
+                    "RunCommand did not run successfully. "
+                    f"Got response: '{value}'. Expected response to contain `value[0]['message']`"
+                )
+
+        except Exception:
+            self._log.info("RunCommand failed to return expected result.")
```
squirrelsc: Because the exception is raised below, it doesn't need to print log before it.

lisa/sut_orchestrator/azure/features.py
```diff
@@ -3729,3 +3731,82 @@ def _prepare_azure_file_share(
...
+        operation = compute_client.virtual_machines.begin_run_command(
+            resource_group_name=context.resource_group_name,
+            vm_name=context.vm_name,
+            parameters=run_command_input,
+        )
+        result = wait_operation(operation=operation, failure_identity="run command")
+        try:
+            # Since wait_operation returns a dict (result.as_dict()), access as dict
+            value = result.get("value")
+            if value and len(value) > 0 and value[0].get("message"):
```
squirrelsc: The code below is enough.

`if value and value[0].get("message"):`

PR #3846, opened 2025-06-05
Azure Update VM Capabilities: Add Location

kamalca:
If the capabilities dictionary is empty for any reason, the hook implementation would have no way to know which region is being queried.


kamalca:
Sorry to change the implementation after the last PR got merged.

I was thinking about edge cases, and I think it is better to explicitly pass the location.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3848, opened 2025-06-10
Fix for XFSTest failing to build on older linux distros

shekharsorot:
This PR addresses a key issue where any XFStest based cases are failing due to build compilation errors when using the latest branch of code from repo "https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git"
Upon investigation we found major changes being done to the code base in January of 2025., which caused build failures on system with glibc-devel < 2.23 deployed.
To work around this, a table has been added to the tool to specifically look for azure supported Linux Distros that were failing XFStest build and specified a last known release tag that was known to build without any issues.

This approach also allows us to add additional distros and version ID's in the future as needed and lock them to specific XFStest release tags.

### DIFF
--- a/microsoft/testsuites/xfstests/xfstests.py
+++ b/microsoft/testsuites/xfstests/xfstests.py
@@ -41,26 +41,26 @@ class Xfstests(Tool):
     # This is the default repo and branch for xfstests.
     # Override this via _install method if needed.
     repo = "https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git"
+    # Branch default will be used when there are no match from the hash table below
     branch = "master"
-    # Dictionary mapping OS identifier (e.g., "ubuntu_18", "rhel_7")
-    # to recommended xfstests-dev git tags.
-    # This allows using specific xfstests versions for different OSes if needed.
-    # The key is typically constructed as
-    # f"{node.os.id}_{node.os.information.version.major}".
-    # The value is the git tag/branch to checkout.
-    # The 'default' key is used if a specific OS version is not found.
+    # This hash table contains recommended tags for different OS versions.
+    # Since not all distros are updated to the latest build tools
+    # XFStests may not compile or run correctly.
+    # This is a workaround to ensure that the tests run on the recommended tag
+    # for the OS version.
+    # The format for key is "<OS full_version>_< major>_<minor>_<patch>"
+    # This information is derived from node.os.information
     os_recommended_tags: Dict[str, str] = {
-        "ubuntu_18.04": "v2024.12.22",  # Ubuntu 18.04 LTS
-        "ubuntu_20.04": "v2024.12.22",  # Ubuntu 20.04 LTS
-        "centos_7": "v2024.12.22",  # CentOS 7.x (ID: centos, MajorVersion: 7)
-        "rhel_7.8": "v2024.12.22",  # RHEL 7.x (ID: rhel, MajorVersion: 7)
-        "sles_12.5": "v2024.12.22",  # SUSE Linux Enterprise Server 12
-        "debian_10": "v2024.12.22",  # Debian 10 (Buster)
-        "debian_11": "v2024.12.22",  # Debian 11 (Bullseye)
-        "cbl-mariner_1": "v2024.12.22",  # CBL-Mariner 1.0
-        "cbl-mariner_2": "v2024.12.22",  # CBL-Mariner 2.0
-        "default": "v2024.02.09",  # Default tag for other/unlisted OSes
+        "SUSE Linux Enterprise Server 15 SP5_15_5_0": "v2025.04.27",
+        "SUSE Linux Enterprise Server 12 SP5_12_5_0": "v2025.12.22",
+        "Debian GNU/Linux 11 (bullseye)_11_11_0": "v2024.12.22",
+        "Ubuntu 18.04.6 LTS_18_4_0": "v2024.12.22",
+        "Ubuntu 20.04.6 LTS_20_4_0": "v2024.12.22",
+        "Red Hat Enterprise Linux Server release 7.8 (Maipo)_7_8_0": "v2024.02.09",
+        "unknown": "v2024.02.09",  # Default tag for distros that cannot be identified
     }
+    # for all other distross not part of the above hash table,
+    # the default branch will be used from line 45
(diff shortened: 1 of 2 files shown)


microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -945,3 +968,30 @@ def create_xfstest_stack_info(
...
                 )
         return result
+
+    def get_os_id_version(self) -> str:
+        """
+        Reads /etc/os-release and extracts ID and VERSION_ID.
+        Returns a string in the format <ID>_<VERSION_ID>.
+        If /etc/os-release is not found, returns "default".
+        """
+        cat = self.node.tools[Cat]
```
squirrelsc: Please use self.node.os.information, it handles different distros and provide more comprehensive information. You can also check os version by `isinstance(self.node.os, Ubuntu)`.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -41,7 +41,26 @@ class Xfstests(Tool):
...
+        "ubuntu_18.04": "v2024.12.22",  # Ubuntu 18.04 LTS
+        "ubuntu_20.04": "v2024.12.22",  # Ubuntu 20.04 LTS
+        "centos_7": "v2024.12.22",  # CentOS 7.x (ID: centos, MajorVersion: 7)
+        "rhel_7.8": "v2024.12.22",  # RHEL 7.x (ID: rhel, MajorVersion: 7)
+        "sles_12.5": "v2024.12.22",  # SUSE Linux Enterprise Server 12
+        "debian_10": "v2024.12.22",  # Debian 10 (Buster)
+        "debian_11": "v2024.12.22",  # Debian 11 (Bullseye)
+        "cbl-mariner_1": "v2024.12.22",  # CBL-Mariner 1.0
+        "cbl-mariner_2": "v2024.12.22",  # CBL-Mariner 2.0
+        "default": "v2024.02.09",  # Default tag for other/unlisted OSes
```
squirrelsc: Should it use the master branch, if it's unknown or newer?

shekharsorot: In an event if the OS information is unknown due to any reason ( missing /etc/os-release or a custom distro build not implementing lsb-releases ) we should ideally fall back to 2024.02.09 which was the last known tag that was known to work on all releases on Azure. The lookup method is supposed to return "default" only if it can't determine the OS name and version_id.

For everything else ( which should include distro not in the hash table ) we should be using the "master" release tag

We can refine this approach bit more if you recommend.

microsoft/testsuites/xfstests/xfstests.py
shekharsorot: We used CAT tool here to account for custom distro builds that may not implement  /usr/lib/os-release ( /etc/os-release is a softlink ).

I will modify the code to use self.node.os.information, and if failure, only then fallback to CAT to do a final check before returning "default" string to force the lowest safe tag for XFStests checkout.

microsoft/testsuites/xfstests/xfstests.py
squirrelsc: How do you differentiate between "the OS information is unknown" (fallback to v2024.02.09) and "For everything else" (fallback to master)?

From my understanding, all versions use v2024.02.09, if the version is not listed here.

microsoft/testsuites/xfstests/xfstests.py
squirrelsc: When using `os.information`, it needs to check two parts: the distribution and the distribution version. If a version is not parsed, you can fix it in the corresponding distribution, so there's no need to use "cat" anymore.

microsoft/testsuites/xfstests/xfstests.py
```diff
@@ -968,30 +976,23 @@ def create_xfstest_stack_info(
...
-        cat = self.node.tools[Cat]
         try:
-            os_release_content = cat.run(
-                "/etc/os-release", sudo=True, force_run=True
-            ).stdout
-        except Exception:
-            return "default"
+            os_info = self.node.os.information
+            full_version = getattr(os_info, 'full_version', '')
+            major = getattr(os_info.version, 'major', 0) if hasattr(os_info, 'version') else 0
```
squirrelsc: @LiliDeng can you work with @shekharsorot to use the right pattern of os version comparison?

LiliDeng: @squirrelsc do you mean the way used like [here ](https://github.com/microsoft/lisa/blob/main/microsoft/testsuites/vm_extensions/linux_patch_extension.py#L64)?

microsoft/testsuites/xfstests/xfstests.py
shekharsorot: Updated the code and method. the method **get_os_id_version** will return the string **<full_name>_<major>_<minor>_<patch>** to ensure we get a unique string for a given distro.
If the **node.os.information** is unable to determine these properties, the method simply returns "unknown" which can be used to take appropriate decision


Changes to code logic:
If the returned value matches the keys in the hash table, we used the corresponding value as branch tag.
If no match from hash table, we used the default value for branch which is "master"

We will update the hash table to add/remove OS information as they are patches.  example: SLES 15 SP5 can be patched by Suse in the future to support newer XFStests code and we will then remove//update the entry accordingly. some OS like SLES 12 , RHEL 7.x may never get the required updates to build tools and will remain here.

microsoft/testsuites/xfstests/xfstests.py
squirrelsc: Yes, mostly. But the following should be sufficient. Using `<` can handle temporary test versions (like Ubuntu) and minor version upgrads.

```python
if isinstance(node.os, Redhat) and full_version < "7.1":
  pass
elif isinstance(node.os, Redhat) and full_version < "7.2":
  pass
else:
  # newer versions
  pass
```

shekharsorot: updated code to use version and release instead of full name. This simplifies lookup and ensures we are able to handle small patch versions without issues like having to add entries for each patch version to hash table.

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3851, opened 2025-06-10
test/cvm_boot: optionally add package repository before upgrading packages

trungams:
When running reboot tests, sometimes we want to pull updated packages from repositories that might not be available by default. This change adds support for 2 types of package repositories: remote and local repositories.
- Remote repository can be added using repo_url
- Local repository can be created from a tarball containing packages (only RPM support at the moment)

### DIFF
--- a/lisa/operating_system.py
+++ b/lisa/operating_system.py
@@ -19,6 +19,7 @@ from typing import (
     Type,
     Union,
 )
+from urllib.parse import urlparse
 
 from assertpy import assert_that
 from retry import retry
@@ -377,6 +378,9 @@ class Posix(OperatingSystem, BaseClassMixin):
         )
 
         return kernel_information
+
+    def add_repository(self, repo: str, **kwargs: Any) -> None:
+        raise NotImplementedError()
 
     def install_packages(
         self,
@@ -985,6 +989,7 @@ class Debian(Linux):
         no_gpgcheck: bool = True,
         repo_name: Optional[str] = None,
         keys_location: Optional[List[str]] = None,
+        **kwargs: Any,
     ) -> None:
         self._initialize_package_installation()
         if keys_location:
@@ -1537,6 +1542,7 @@ class RPMDistro(Linux):
         no_gpgcheck: bool = True,
         repo_name: Optional[str] = None,
         keys_location: Optional[List[str]] = None,
+        **kwargs: Any,
     ) -> None:
         self._node.tools[YumConfigManager].add_repository(repo, no_gpgcheck)
 
@@ -1955,6 +1961,35 @@ class CBLMariner(RPMDistro):
             sudo=True,
         )
 
+    def add_repository(
+        self,
+        repo: str,
+        no_gpgcheck: bool = True,
+        repo_name: Optional[str] = None,
+        keys_location: Optional[List[str]] = None,
+        **kwargs: Any,
+    ) -> None:
+        parsed_url = urlparse(repo)
+        if parsed_url.scheme and parsed_url.netloc:
+            self._node.tools[YumConfigManager].add_repository(repo, no_gpgcheck)
+        else:
+            self._create_local_repo(Path(repo))
+
+    def _create_local_repo(self, source_tarball: Path) -> None:
+        from lisa.tools import CreateRepo, RemoteCopy
+
+        working_path = Path(self._node.get_working_path())
+        tarball_file = source_tarball.name
+        tarball_path = working_path / tarball_file
+
+        # copy tarball to remote node
+        result = self._node.tools[RemoteCopy].copy_to_remote(
+            src=source_tarball, dest=working_path
+        )
+        self._log.debug(f"tarball copied to: {result}")
+
+        self._node.tools[CreateRepo].create_repo_from_tarball(tarball_path)
+
     # Disable KillUserProcesses to avoid test processes being terminated when
     # the SSH session is reset
     def set_kill_user_processes(self) -> None:
@@ -2054,6 +2089,7 @@ class Suse(Linux):
         no_gpgcheck: bool = True,
         repo_name: Optional[str] = None,
         keys_location: Optional[List[str]] = None,
+        **kwargs: Any,
     ) -> None:
         self._initialize_package_installation()
         cmd = "zypper ar"
@@ -2187,6 +2223,7 @@ class SlMicro(Suse):
         no_gpgcheck: bool = True,
         repo_name: Optional[str] = None,
         keys_location: Optional[List[str]] = None,
+        **kwargs: Any,
     ) -> None:
         raise SkippedException(
             UnsupportedDistroException(
--- a/microsoft/testsuites/cvm/cvm_boot.py
+++ b/microsoft/testsuites/cvm/cvm_boot.py
@@ -21,7 +21,7 @@ from lisa.features.security_profile import (
     SecurityProfile,
     SecurityProfileSettings,
 )
-from lisa.operating_system import CBLMariner, Posix
+from lisa.operating_system import CBLMariner
 from lisa.sut_orchestrator import AZURE
 from lisa.testsuite import simple_requirement
 from lisa.tools import BootCtl, Lsblk, Tpm2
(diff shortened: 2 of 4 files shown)


Camelron (approved):

Camelron (approved):
Tested the CreateRepo tool for my use case; works.

lisa/operating_system.py
```diff
@@ -378,6 +378,15 @@ def get_kernel_information(self, force_run: bool = False) -> KernelInformation:
 
         return kernel_information
 
+    def add_repository(
+        self,
+        repo: str,
+        no_gpgcheck: bool = True,
```
squirrelsc: Besides repo, other arguments shouldn't be listed in the base class, because they are not shared cross distros. Define base class like
```python
def add_repository(
        self,
        repo: str,
        **kwargs: Any
)
```

lisa/operating_system.py
```diff
@@ -378,6 +378,15 @@ def get_kernel_information(self, force_run: bool = False) -> KernelInformation:
...
+        repo_name: Optional[str] = None,
+        keys_location: Optional[List[str]] = None,
+    ) -> None:
+        raise NotImplementedError()
```
squirrelsc, line 383: Please add implementation for AzureLinux

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -110,6 +114,14 @@ def verify_boot_success_after_component_upgrade(
             pkg_version = posix_os.get_package_information(pkg, use_cached=False)
             boot_components_versions[pkg] = pkg_version.version_str
 
+        repo_url = variables.get("cvm_boot_rpm_repository_url")
```
squirrelsc: Make sure the change compatible with other distros.

trungams: This test case is only available for Azure Linux at the moment.

lisa/operating_system.py
trungams: Azure Linux/CBLMariner uses the same implementation as RPMDistro

squirrelsc: If AzureLinux uses the same implementation as RPMDistro, it should call to `RPMDistro.add_repository`, and underlying `YumConfigManager`. If it's not, please overwrite the add_repository for AzureLinux in the operating_system.py.

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -145,3 +157,23 @@ def verify_boot_success_after_component_upgrade(
...
+        tarball_path = working_path / tarball_file
+
+        # copy tarball to remote
+        result = node.tools[RemoteCopy].copy_to_remote(
+            src=source_tarball,
+            dest=working_path,
+        )
+        log.debug(f"tarball copied to: {result}")
+
+        node.tools[CreateRepo].create_repo_from_tarball(tarball_path)
```
squirrelsc: How about call install_packages of the os?

Camelron: The goal here is to make available custom binaries under test. The tarball contains optional test RPMs we are interested in verifying work for boot scenarios, which get installed in this higher priority local repo and installed instead of the standard pulling from remote repo.

So, after this create_repo_from_tarball call, the subsequent posix_os.update_packages (or install_packages) will get packages from this tarball first: https://github.com/microsoft/lisa/pull/3851/files/7fdb880ab27d2b2c116f947642b36f3eb70cbead#diff-37a00efda99b918517183389d41aaac97f4b45d64289be877c325c47729df377R126

trungams: Do you mean to install the locally available rpms? The idea of adding a local repository is that we can use the package manager's update call directly without having to check whether a newer version of the package exists in the tarball. If a new version is not available,`dnf update <pkg>` is just a no-op

lisa/operating_system.py
trungams: Just double checking, do you mean I should change `https://github.com/microsoft/lisa/pull/3851/files#diff-37a00efda99b918517183389d41aaac97f4b45d64289be877c325c47729df377R162` to RPMDistro instead of Posix? And remove this generic `add_repository` function?

microsoft/testsuites/cvm/cvm_boot.py
squirrelsc: The install_packages function supports installing local files, so there's no need to add a repository for local files. I believe install_packages applies only to newer versions (please verify). If the code is changed to call install_packages only, the logic remains the same but becomes simpler.

lisa/operating_system.py
squirrelsc: Ok, I misunderstood the code logic. It looks both `YumConfigManager` and `CreateRepo` are used. But `CreateRepo` looks not necessary. The logic can be like,
```python
if source_tarball:
  # install file by install_package
else:
  if repo_url:
    # add repo

# update package
```

trungams: Let's discuss more in the other comment

microsoft/testsuites/cvm/cvm_boot.py
trungams: I think at least for tdnf/dnf, to install a local rpm, you'd have to give the whole filename of the rpm, and that allows you to install an older version of a package as well.

trungams: Plus, by adding createrepo tool, we have another "custom" repository that is available alongside the default PMC. That will be useful when we want to install any newer version of a package, wherever it is available. 

I believe our team already have some tests that are based on this kind of local repo setup written in shell scripts. It would be really helpful to provide the capability to do the same thing for when we want to migrate our test suite to Lisa.

squirrelsc: Ok, please integrate it into add_repo of AzureLinux or RPMDistro, so the logic can be reused in other places.

trungams: Done

lisa/operating_system.py
```diff
@@ -1955,6 +1960,21 @@ def add_azure_core_repo(
             sudo=True,
         )
 
+    def create_local_repo(self, node: "RemoteNode", source_tarball: Path) -> None:
```
squirrelsc: It should be a private method in CBLMariner, but call it by `add_repository`. So, `add_repository` support both url and local tar files.

trungams: updated

squirrelsc: Don't change the arguments of `add_repository`, so it's easy to called by any place. You can detect it's a local or remote repo in the method.

trungams: Should I get those variables through `kwargs.pop()`? Or should I just reuse the `repo` parameter to specify a path to src tarball instead of repo url? I'd prefer the former to be more explicit of the usage.

squirrelsc: Please reuse the "repo" parameter. It's the local repo, similar to the remote repo.

trungams: updated

microsoft/testsuites/cvm/cvm_boot.py
```diff
@@ -107,16 +111,25 @@ def verify_boot_success_after_component_upgrade(
         boot_components = ["shim", "systemd-boot", "kernel-uki"]
         boot_components_versions: Dict[str, str] = dict()
         for pkg in boot_components:
-            pkg_version = posix_os.get_package_information(pkg, use_cached=False)
+            pkg_version = os.get_package_information(pkg, use_cached=False)
             boot_components_versions[pkg] = pkg_version.version_str
 
+        repo_url = variables.get("rpm_repository_url")
```
squirrelsc: How about just use `rpm_repository`, remove the `rpm_source_tarball`? It looks like one repo var should be enough for remote or local.

trungams: that's fair. Updated

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3852, opened 2025-06-10
Openssl sign verify

WithEnoughCoffee:
Add tests for OpenSSL sign and verify functionality

- Implement test to generate digital signatures using private keys and algorithms
- Add test to verify signatures with corresponding public keys
- Ensure signatures are correctly created and validated using OpenSSL
- Confirm integrity and authenticity of data through sign/verify workflow

### DIFF
--- a/lisa/tools/openssl.py
+++ b/lisa/tools/openssl.py
@@ -63,6 +63,9 @@ class OpenSSL(Tool):
         """
         Generate a key pair using the specified algorithm.
         Returns the private key and public key as strings.
+
+        This key generation is for testing generation of keys
+        with OpenSSL on the remote.
         """
         private_key_result = self.run(
             f"genpkey -algorithm {algorithm} -outform PEM",
--- a/lisa/tools/openssl.py
+++ b/lisa/tools/openssl.py
@@ -37,7 +37,7 @@ class OpenSSL(Tool):
         return self._run_with_piped_input(
             plaintext,
             f"enc -{algorithm} -K '{hex_key}' -iv '{hex_iv}' -base64 -A",
-            expected_exit_code_failure_message="Failed to encrypt data with OpenSSL.",
+            expected_exit_code_failure_message=("Failed to encrypt data with OpenSSL."),
         )
 
     def decrypt(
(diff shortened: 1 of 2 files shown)


!
```diff
@@ -0,0 +1,6 @@
+Merge branch 'main' of https://github.com/WithEnoughCoffee/lisa into openssl-test
```
squirrelsc: Remove this file.

squirrelsc:
Please rebase on the latest main branch changes to make the changes in this PR easier to review.

lisa/tools/openssl.py
```diff
@@ -0,0 +1,161 @@
...
+    ) -> str:
+        """
+        Sign the data using the specified private key and algorithm.
+        Returns the base64 encoded signature.
+        """
+        return self._run_with_piped_input(
+            data,
+            f"dgst -{algorithm} -sign <(echo '{private_key}') | openssl base64 -A",
+            expected_exit_code_failure_message="Failed to sign data with OpenSSL.",
+        )
```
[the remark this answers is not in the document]
tobiasb-ms: The suggested code won't work because everything is happening on the remote.

Moreover, much of the code that executes these remote processes assume bash, so I think we're safe.

lisa/tools/openssl.py
```diff
@@ -56,7 +56,66 @@ def decrypt(
         return self._run_with_piped_input(
             ciphertext,
             f"enc -d -{algorithm} -K '{hex_key}' -iv '{hex_iv}' -base64 -A",
-            expected_exit_code_failure_message="Failed to decrypt data with OpenSSL.",
+            expected_exit_code_failure_message=("Failed to decrypt data with OpenSSL."),
+        )
+
+    def create_key_pair(self, algorithm: str = "RSA") -> tuple[str, str]:
```
squirrelsc: @LiliDeng I remember we have similar code to generate ssh keys, should they be merged?

microsoft/testsuites/security/openssl.py
```diff
@@ -32,12 +32,16 @@ class OpenSSLTestSuite(TestSuite):
         priority=2,
     )
     def verify_openssl_basic(self, log: Logger, node: Node) -> None:
+        """This function tests the basic functionality of
+        OpenSSL by calling helper functions"""
         self._openssl_test_encrypt_decrypt(log, node)
+        self.openssl_test_sign_verify(log, node)
```
squirrelsc: Make the function name consistent with the one above: `_openssl_test_sign`.

lisa/tools/openssl.py
LiliDeng: [that one](https://github.com/microsoft/lisa/blob/main/lisa/util/__init__.py#L468) is to generate key pairs with python paramiko library in lisa run machine and this one uses openssl to generate key pairs on Linux machine.

WithEnoughCoffee:
All comments have been addressed and tests are passing.

lisa/tools/openssl.py
squirrelsc: @WithEnoughCoffee please refer to `get_or_generate_key_pairs` to create a util method in `util/__init__.py` to create key pairs by paramiko.

microsoft/testsuites/security/openssl.py
squirrelsc: Please change it like `openssl_test_sign`, the "verity" is redundant with "test".

lisa/tools/openssl.py
tobiasb-ms: The point of this is to test `openssl` on the remote linux machine, so we don't want to use `paramiko`. And as @LiliDeng mentioned, that would be on the lisa run machine anyway, not on the machine under test.

It's more appropriate for these tests to create key pairs using `openssl` on the remote.

microsoft/testsuites/security/openssl.py
tobiasb-ms: `verify` is the inverse of `sign` in this case, not a redundant version of `test`. To be consistent, the two helpers should either be:
```
_openssl_test_encrypt_decrypt
_openssl_test_sign_verify
```
or
```
_openssl_test_encrypt
_openssL_test_sign
```

I strongly prefer the first pattern that clearly states what it's testing.

lisa/tools/openssl.py
squirrelsc: Paramiko is used to generate keys, replacing genkey in the current code. The keys are used by OpenSSL for signing and encryption.

microsoft/testsuites/security/openssl.py
squirrelsc: Thanks, Tobias. Agree on 

```
_openssl_test_encrypt_decrypt
_openssl_test_sign_verify
```

lisa/tools/openssl.py
tobiasb-ms: yes; those keys would in fact work.

But then we wouldn't be testing `openssl`'s ability to generate keys, which is desirable to do.

squirrelsc: If the test case wants to test OpenSSL via genkey, please add a comment to prevent others from using this pattern in other test cases.

WithEnoughCoffee: Comment added

squirrelsc:
@LiliDeng LGTM

LiliDeng (approved):

PR #3853, opened 2025-06-11
Add and fix hyperv modules test

kanchansenlaskar:
This PR contains the changes:
1. Addition of other hyperv modules e.g. (hv_vmbus hv_storvsc hv_utils hv_balloon hid_hyperv hyperv_keyboard hyperv_fb)
2. Also since the test is about reloading the modules, only looking for non-built in modules is not enough, hence added a small change to look specifically for reloadable modules in the config file.
3. Running separate shell script with nohup and handling the scenarios where the ssh disconnection may be encountered


microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,133 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+        )
+
+        assert_that(new_log_level).described_as(
+            f"Expected dmesg log level to be set to {preferred_log_level}, "
+            f"but it is {new_log_level}."
+        ).is_equal_to(preferred_log_level)
 
-    def _get_not_built_in_modules(self, node: Node) -> List[str]:
+    def _get_not_built_in_modules(
+        self, node: Node, only_loadable: bool = False
```
squirrelsc: 1. I don't see this method was called with `only_loadable = True`. Why add it?
1. With this parameter `only_loadable`, the method name isn't right now. Call it like `_get_modules_by_built_in_type`, and rename the parameter from `only_loadable` to `is_built_in`

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,133 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+                "This may be due to them being built into the kernel or in use."
             )
 
-        assert_that(result.stdout.count("rmmod")).described_as(
-            f"Expected {module} to be removed {loop_count} times"
-        ).is_equal_to(loop_count)
-        assert_that(result.stdout.count("insmod")).described_as(
-            f"Expected {module} to be inserted {loop_count} times"
-        ).is_equal_to(loop_count)
+    def _set_dmesg_log_level(
```
squirrelsc: Why does it need to set log_level?

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,133 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+        for module in hv_modules:
+            try:
+                if module not in non_built_in_modules:
+                    log.info(f"{module} is not a reloadable module, ")
+                    skipped_modules.append(module)
+                    continue
+                loop_count = 100
+                log.info(f"Reloading {module} {loop_count} times")
+
+                reload_command = (
```
squirrelsc: use tool: Modprobe

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,133 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
-                "kernel and therefore can not be reloaded"
-            )
+        preferred_log_level = 4
+        log_level = int(
+            node.tools[Cat]
+            .read("/proc/sys/kernel/printk", force_run=True, sudo=True)
+            .split()[0]
+        )
+
+        # The 10-minute timeout specified in node.execute is not being honoured,
```
squirrelsc: Are you asking if the log floods cause the test to run longer, specifically up to 50 seconds? How many lines are printed when running this test case?

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,133 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+                    or result.stdout.count("insmod") != loop_count
+                ):
+                    failure_message = (
+                        f"Module {module} was not reloaded {loop_count} times. "
+                        f"rmmod count: {result.stdout.count('rmmod')}, "
+                        f"insmod count: {result.stdout.count('insmod')}"
+                    )
+                    failed_modules[module] = failure_message
+            except Exception as e:
+                if log_level > preferred_log_level:
```
squirrelsc: Move the log_level adjustment code to the finally block, so it executes regardless of whether an exception occurs or not.

kanchansenlaskar: done

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: Taking hv_netvsc as an example where we are trying to run the following command (which runs for 100 iterations as you can see below):
for i in $(seq 1 100); do modprobe -r -v hv_netvsc; modprobe -v hv_netvsc; done; sleep 1; ip link set eth0 down; ip link set eth0 up; dhclient -r eth0; dhclient eth0

I tried running locally to the VMs without LISA:
With log level 4:
The execution finishes in around 10 seconds.
200 lines

With log level 7:
The execution takes around 90 seconds.
200 + 800 lines

Coming back to LISA:
The process runs only for around 50 seconds, when the execution time goes beyond that, it shows that as not running. This is what is coming from the LocalProcess of spur lib.

Basically, in the wait_result() method of process.py file, there is a while loop condition:
while self.is_running() and timeout >= timer.elapsed(False):
In that the first condition is returning False after approx. 50 seconds and hence it comes out of the loop it stops waiting for the execution to complete even though the process is actually running. Hence the timeout of 10 mins is also not considered.

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: The previous implementation was also like this. The reason probably could be that this needs to be run at one shot in the loop with other steps, and hence Modprobe lib was not used. Do let me know if you still feel this needs to be done with Modprobe.

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: to get fewer logs as mentioned in the description of the PR, also have mentioned the reason in the response to one of your latter comments.

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: self._get_not_built_in_modules(node, only_loadable=True) is called in the test case definition of verify_reload_hyperv_modules

kanchansenlaskar: only_loadable name is used because we are trying to get only the entries in the config with "=m" (which are loadable)

squirrelsc: Ok, I found `_get_not_built_in_modules` is used in both True/False. The `only_loadable` and `built_in` are the same thing but with different names. They need to be consistent, and only one should be kept.

kanchansenlaskar: Actually, they are not same, in fact the modules which are marked with "=y" in the config file are built-in and the ones marked as "=m" are loadable modules. As you can see the conditions in line# 365 and line# 370 in the definition of _get_not_built_in_modules() method.

squirrelsc: I mean, both only_loadable and built_in describe the module build type. It doesn't need to use both of them here. It can improve readability of the code. Either change the method to `_get_modules_by_type`, or change the argument name to `is_built_in`, and update the method logic accordingly.

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,138 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+                ):
+                    failure_message = (
+                        f"Module {module} was not reloaded {loop_count} times. "
+                        f"rmmod count: {result.stdout.count('rmmod')}, "
+                        f"insmod count: {result.stdout.count('insmod')}"
+                    )
+                    failed_modules[module] = failure_message
+
+        finally:
+            log.info("Executing finally block")
```
squirrelsc: This log line is not helpful, and it shouldn't be info level.

microsoft/testsuites/core/hv_module.py
squirrelsc: I understand the reduced log level. Regarding the timeout, I need more information. Do you mean it happens only in LocalProcess, not SshProcess? If self.is_running() returns False, it means the process has exited. This is by design. How do you know the process is actually running?

microsoft/testsuites/core/hv_module.py
squirrelsc: Please extend the reload method to support multiple reloads by adding an argument. The default value should be 1.

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,138 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+        non_built_in_modules = set(
+            self._get_not_built_in_modules(node, only_loadable=True)
         )
+        # Constants
+        try:
+            log.info("Executing try block")
+            for module in hv_modules:
+                # try:
+                if module not in non_built_in_modules:
+                    log.info(f"{module} is not a reloadable module, ")
```
squirrelsc: most log lines should be debug level.

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,138 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+                        "ip link set eth0 down; ip link set eth0 up; dhclient eth0"
+                    )
 
-        assert_that(result.stdout.count("rmmod")).described_as(
-            f"Expected {module} to be removed {loop_count} times"
-        ).is_equal_to(loop_count)
-        assert_that(result.stdout.count("insmod")).described_as(
-            f"Expected {module} to be inserted {loop_count} times"
-        ).is_equal_to(loop_count)
+                log.info(f"reload command to reload {module}: {reload_command}")
```
squirrelsc: It doesn't need to output the command, it's output by LISA already.

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: done

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: done

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: _Do you mean it happens only in LocalProcess, not SshProcess?_
> I tried with SshProcess as well in LISA, and faced the same problem

_do you know the process is actually running?_
> I don't know actually, however, when I ran the same command locally on the same vm with the same log level, it ran without any issues. It only gave issue when I ran with spur lib from LISA.

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: removed

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: done

microsoft/testsuites/core/hv_module.py
kanchansenlaskar: done

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,127 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+
+        assert_that(new_log_level).described_as(
+            f"Expected dmesg log level to be set to {preferred_log_level}, "
+            f"but it is {new_log_level}."
+        ).is_equal_to(preferred_log_level)
+
+    def _get_modules_by_type(
+        self,
+        node: Node,
+        module_type: Literal["not_built_in", "loadable"] = "not_built_in",
```
squirrelsc: Define an enum

lisa/tools/modprobe.py
```diff
@@ -126,21 +111,42 @@ def module_exists(self, modules: Union[str, List[str]]) -> bool:
     def reload(
         self,
         mod_names: List[str],
+        reload_count: int = 1,
+        with_verbose: bool = False,
```
squirrelsc: `verbose`

lisa/tools/modprobe.py
```diff
@@ -126,21 +111,42 @@ def module_exists(self, modules: Union[str, List[str]]) -> bool:
     def reload(
         self,
         mod_names: List[str],
+        reload_count: int = 1,
```
squirrelsc: `times`

lisa/tools/modprobe.py
```diff
@@ -126,21 +111,42 @@ def module_exists(self, modules: Union[str, List[str]]) -> bool:
...
                 if mod_name == "hv_netvsc":
                     # handle special case
-                    self._reload_hv_netvsc()
-                else:
-                    # execute the command for regular non-network modules
-                    self.node.execute(
-                        f"modprobe -r {mod_name}; modprobe {mod_name};",
-                        sudo=True,
-                        shell=True,
+                    dh_client = self.node.tools[Dhclient]
```
squirrelsc: add a `generate_renew_command` for `dhcpclient` tool and use it.

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +206,127 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
-            raise SkippedException(
-                f"Module {module} is in use so it cannot be reloaded"
+        # The 10-minute timeout specified in node.execute is not being honoured,
+        # as the local process in the spur library stops running after approximately
+        # 50 seconds. To mitigate this, we set the log level to 4 for any VM where
+        # the log level exceeds 4. This helps reduce excessive logging during module
+        # reloads, which could otherwise cause command execution to time out due to
+        # the loop count being set to 100.
+        if log_level > preferred_log_level:
+            log.info(
```
squirrelsc: Move all log to debug level, until it tells a story.

lisa/tools/modprobe.py
```diff
@@ -22,22 +23,6 @@ def _check_exists(self) -> bool:
     def _initialize(self, *args: Any, **kwargs: Any) -> None:
         self._command = "modprobe"
 
-    # hv_netvsc needs a special case, since reloading it has the potential
-    # to leave the node without a network connection if things go wrong.
-    def _reload_hv_netvsc(self) -> None:
-        # These commands must be sent together, bundle them up as one line
```
squirrelsc, line 31: Keep the comments somewhere, so others know why compose the long command.

microsoft/testsuites/core/hv_module.py
squirrelsc: Can you check if the issue still exists, after moved to modprobe? And please share me the full log, I can take a look.

microsoft/testsuites/core/hv_module.py
```diff
@@ -18,9 +18,21 @@
...
-from lisa.tools import KernelConfig, LisDriver, Lsinitrd, Lsmod, Modinfo, Modprobe
+from lisa.tools import Cat, KernelConfig, LisDriver, Lsinitrd, Lsmod, Modinfo, Modprobe
 from lisa.util import LisaException, SkippedException
 
+ModulesType = Enum(
+    "ModulesType",
+    [
+        # Modules which dont have "=y" in the kernel config
+        # and therefore are not built into the kernel.
+        "NOT_BUILT_IN",
```
squirrelsc: It should be 
* `BUILT_IN` =y
* `MODULE` =m
* `NOT_BUILD` =n

lisa/tools/dhclient.py
```diff
@@ -42,6 +42,15 @@ def _freebsd_tool(cls) -> Optional[Type[Tool]]:
     def can_install(self) -> bool:
         return False
 
+    def generate_renew_command(self, interface: str = "eth0") -> str:
```
squirrelsc: Please also reuse it inside of the `dhclient` to reduce duplicate code.

lisa/tools/modprobe.py
```diff
@@ -125,21 +119,120 @@ def module_exists(self, modules: Union[str, List[str]]) -> bool:
...
+        )
+        device_or_resource_busy_count = int(
+            self.node.execute(
+                f"grep -o 'Device or resource busy' {nohup_output_log_file_name} | wc -l",
+                sudo=True,
+                shell=True,
+            ).stdout.strip()
+        )
+
+        # self.node.execute(
```
squirrelsc: Remove unused code, instead of comment out.

lisa/tools/modprobe.py
```diff
@@ -125,21 +119,120 @@ def module_exists(self, modules: Union[str, List[str]]) -> bool:
...
+        if verbose:
+            verbose_flag = "true"
+        else:
+            verbose_flag = "false"
+
+        modprobe_reloader_tool = CustomScriptBuilder(
+            Path(__file__).parent.joinpath("scripts"), ["modprobe_reloader.sh"]
+        )
+
+        parameters = f'{nohup_output_log_file_name} {loop_process_pid_file_name} {mod_name} {times} {verbose_flag} "{dhclient_renew_command}"'
```
squirrelsc: Add comments above to explain them.

lisa/tools/scripts/modprobe_reloader.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
+
+
+log_file="$1"
+pid_file="$2"
+module_name="$3"
+times="$4"                      # 100
+verbose="$5"                          # true/false
```
squirrelsc: Can they have a default value?

lisa/tools/scripts/modprobe_reloader.sh
```diff
@@ -0,0 +1,79 @@
...
+else
+    v=""
+fi
+
+echo "with verbose: $verbose, times: $times, module_name: $module_name, dhclient_renew_command: $dhclient_renew_command"
+echo "v: $v"
+
+
+
+if [ "$module_name" = "hv_netvsc" ]; then
```
squirrelsc, line 23: If the hv_netvsc is checked in the script, it doesn't need to pass in the `$dhclient_renew_command`, set it here, if it's needed.

lisa/tools/scripts/modprobe_reloader.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "Executing command:"
+echo "sudo nohup sh -c '(
+  for i in \$(seq 1 $times); do
+    modprobe -r $v $module_name >> $log_file 2>&1
+    modprobe $v $module_name >> $log_file 2>&1
+  done
+  sleep 1
+  ip link set eth0 down >> $log_file 2>&1
+  ip link set eth0 up >> $log_file 2>&1
+  dhcpcd -k eth0 >> $log_file 2>&1
```
squirrelsc: Should it use `dhclient_renew_command`?

lisa/tools/scripts/modprobe_reloader.sh
```diff
@@ -0,0 +1,79 @@
...
+    for i in $(seq 1 '$times'); do
+      modprobe -r '$v' '"$module_name"' >> "'"$log_file"'" 2>&1
+      modprobe '$v' '"$module_name"' >> "'"$log_file"'" 2>&1
+    done
+  ) &
+  echo $! > "'"$pid_file"'"
+'
+fi
+
+# sudo sh -c '(for i in \$(seq 1 $times); do modprobe -r $verbose $mod_name >> $nohup_output_log_file_name 2>&1; modprobe $verbose $mod_name >> $nohup_output_log_file_name 2>&1; done; sleep 1; ip link set eth0 down; ip link set eth0 up; $renew_command)' & echo \$! > $loop_process_pid_file_name"
```
squirrelsc: Remove it.

lisa/tools/scripts/modprobe_reloader.sh
```diff
@@ -0,0 +1,79 @@
...
+  ip link set eth0 up >> $log_file 2>&1
+  dhcpcd -k eth0 >> $log_file 2>&1
+  dhcpcd eth0 >> $log_file 2>&1
+  sudo service ssh status >> $log_file 2>&1
+  ip a >> $log_file 2>&1
+) &
+echo \$! > $pid_file'"
+
+
+sudo nohup sh -c '
```
squirrelsc: How about moving sudo, nohup, sh, and log file redirection outside the script? It could simplify the script by handling logging externally.

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +218,101 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+                or result["insmod_count"] != loop_count
+            ):
+                failure_message = (
+                    f"Module {module} was not reloaded {loop_count} times. "
+                    f"rmmod count: {result['rmmod_count']}, "
+                    f"insmod count: {result['insmod_count']}"
+                )
+                failed_modules[module] = failure_message
+
+        result_message = ""
```
squirrelsc: This line shouldn't be followed by the next line. Use either += without this line, or = to set the value of result_message.

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +218,101 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+        )
 
-        assert_that(result.stdout.count("rmmod")).described_as(
-            f"Expected {module} to be removed {loop_count} times"
-        ).is_equal_to(loop_count)
-        assert_that(result.stdout.count("insmod")).described_as(
-            f"Expected {module} to be inserted {loop_count} times"
-        ).is_equal_to(loop_count)
+        if failed_modules:
+            raise AssertionError(result_message)
```
squirrelsc: `raise LISAException(...`

lisa/tools/modprobe.py
```diff
@@ -125,21 +119,120 @@ def module_exists(self, modules: Union[str, List[str]]) -> bool:
...
                     )
+                    break
+            except Exception as e:
+                self._log.debug(
+                    "An exception is caught, this could be due to the VM network "
+                    f"going down during the module reload operation, {e}"
+                    "\nTrying to reconnect to the remote node in 10 sec..."
+                )
+
+                time.sleep(10)
```
squirrelsc: Make it shorter like 2 seconds.

squirrelsc:
@LiliDeng LGTM

microsoft/testsuites/core/hv_module.py
```diff
@@ -206,45 +207,100 @@ def verify_hyperv_modules(self, log: Logger, environment: Environment) -> None:
...
+        failed_modules = {}
+        hv_modules = [
+            "hv_vmbus",
+            "hv_netvsc",
+            "hv_storvsc",
+            "hv_utils",
+            "hv_balloon",
+            "hid_hyperv",
+            "hyperv_keyboard",
+            "hyperv_fb",
```
LiliDeng, line 228: Is it hyperv_fb mandatory?

kanchansenlaskar:
@LiliDeng any more comments?

LiliDeng:
@kanchansenlaskar please check the logic of verify_reload_hyperv_modules, some modules are loadable and some modules are built-in, the final result of the test case is skipped, which is unexpected.

kanchansenlaskar:
> @kanchansenlaskar please check the logic of verify_reload_hyperv_modules, some modules are loadable and some modules are built-in, the final result of the test case is skipped, which is unexpected.

@LiliDeng I think the test should be skipped for those modules, but with all the information and proper reasoning (even before this PR change, for the already existing module hv_netvsc, it was skipping wherever it was in use or built-in).

We should not mark the test pass just because the modules are in use or built in.

LiliDeng:
> > @kanchansenlaskar please check the logic of verify_reload_hyperv_modules, some modules are loadable and some modules are built-in, the final result of the test case is skipped, which is unexpected.
> 
> @LiliDeng I think the test should be skipped for those modules, but with all the information and proper reasoning (even before this PR change, for the already existing module hv_netvsc, it was skipping wherever it was in use or built-in).
> 
> We should not mark the test pass just because the modules are in use or built in.

if all modules are built-in we can mark them as skipped. but some of modules exist as module and are reloaded during testing, in this case, we should not mark the case as skippped

kanchansenlaskar:
> > > @kanchansenlaskar please check the logic of verify_reload_hyperv_modules, some modules are loadable and some modules are built-in, the final result of the test case is skipped, which is unexpected.
> > 
> > 
> > @LiliDeng I think the test should be skipped for those modules, but with all the information and proper reasoning (even before this PR change, for the already existing module hv_netvsc, it was skipping wherever it was in use or built-in).
> > We should not mark the test pass just because the modules are in use or built in.
> 
> if all modules are built-in we can mark them as skipped. but some of modules exist as module and are reloaded during testing, in this case, we should not mark the case as skippped

OK, shall we mark the test as Passed in such case?

LiliDeng:
> > > > @kanchansenlaskar please check the logic of verify_reload_hyperv_modules, some modules are loadable and some modules are built-in, the final result of the test case is skipped, which is unexpected.
> > > 
> > > 
> > > @LiliDeng I think the test should be skipped for those modules, but with all the information and proper reasoning (even before this PR change, for the already existing module hv_netvsc, it was skipping wherever it was in use or built-in).
> > > We should not mark the test pass just because the modules are in use or built in.
> > 
> > 
> > if all modules are built-in we can mark them as skipped. but some of modules exist as module and are reloaded during testing, in this case, we should not mark the case as skippped
> 
> OK, shall we mark the test as Passed in such case?

if no error happened during reload, it should be marked as pass

kanchansenlaskar:
> > > > > @kanchansenlaskar please check the logic of verify_reload_hyperv_modules, some modules are loadable and some modules are built-in, the final result of the test case is skipped, which is unexpected.
> > > > 
> > > > 
> > > > @LiliDeng I think the test should be skipped for those modules, but with all the information and proper reasoning (even before this PR change, for the already existing module hv_netvsc, it was skipping wherever it was in use or built-in).
> > > > We should not mark the test pass just because the modules are in use or built in.
> > > 
> > > 
> > > if all modules are built-in we can mark them as skipped. but some of modules exist as module and are reloaded during testing, in this case, we should not mark the case as skippped
> > 
> > 
> > OK, shall we mark the test as Passed in such case?
> 
> if no error happened during reload, it should be marked as pass

ok done

LiliDeng (approved):

PR #3857, opened 2025-06-12
Allow user passed MTUs for ntttcp network perf tests.

SRIKKANTH:
There is a need for testing network perf with different MTU sizes.
This change enables it.

### DIFF
--- a/lisa/messages.py
+++ b/lisa/messages.py
@@ -251,6 +251,8 @@ class NetworkTCPPerformanceMessage(PerfMessage):
     retransmitted_segments: Decimal = Decimal(0)
     congestion_windowsize_kb: Decimal = Decimal(0)
     protocol_type: Optional[str] = TransportProtocol.Tcp
+    client_mtu: int = -1
+    server_mtu: int = -1
 
 
 @dataclass
@@ -266,6 +268,8 @@ class NetworkUDPPerformanceMessage(PerfMessage):
     data_loss: Decimal = Decimal(0)
     packet_size_kbytes: Decimal = Decimal(0)
     protocol_type: Optional[str] = TransportProtocol.Udp
+    client_mtu: int = -1
+    server_mtu: int = -1
 
 
 @dataclass
--- a/lisa/tools/ip.py
+++ b/lisa/tools/ip.py
@@ -220,6 +220,11 @@ class Ip(Tool):
         return int(cat.read(f"/sys/class/net/{nic_name}/mtu", force_run=True))
 
     def set_mtu(self, nic_name: str, mtu: int) -> None:
+        # Check if mtu is integer
+        try:
+            mtu = int(mtu)
+        except ValueError:
+            raise LisaException(f"MTU value is not an integer: {mtu}")
         self.run(f"link set dev {nic_name} mtu {mtu}", force_run=True, sudo=True)
         new_mtu = self.get_mtu(nic_name=nic_name)
         assert_that(new_mtu).described_as("set mtu failed").is_equal_to(mtu)
--- a/lisa/tools/ntttcp.py
+++ b/lisa/tools/ntttcp.py
@@ -83,6 +83,7 @@ class NtttcpResult:
     rx_packets: Decimal = Decimal(0)
     pkts_interrupt: Decimal = Decimal(0)
     cycles_per_byte: Decimal = Decimal(0)
+    mtu: int = -1
 
 
 class Ntttcp(Tool):
@@ -389,10 +390,16 @@ class Ntttcp(Tool):
         buffer_size: int,
         test_case_name: str,
         test_result: "TestResult",
+        client_mtu: int = -1,
+        server_mtu: int = -1,
     ) -> NetworkTCPPerformanceMessage:
         other_fields: Dict[str, Any] = {}
         other_fields["tool"] = constants.NETWORK_PERFORMANCE_TOOL_NTTTCP
         other_fields["buffer_size"] = Decimal(buffer_size)
+        if client_mtu != -1:
+            other_fields["client_mtu"] = client_mtu
+        if server_mtu != -1:
+            other_fields["server_mtu"] = server_mtu
         other_fields["connections_created_time"] = int(
             client_result.connections_created_time
         )
@@ -415,6 +422,8 @@ class Ntttcp(Tool):
             buffer_size,
             test_case_name,
             test_result,
+            client_mtu,
+            server_mtu,
         )
 
         return create_perf_message(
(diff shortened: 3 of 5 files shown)


microsoft/testsuites/performance/common.py
```diff
@@ -313,6 +315,20 @@ def perf_ntttcp(  # noqa: C901
             ntttcp.setup_system(udp_mode, set_task_max)
         for lagscope in [client_lagscope, server_lagscope]:
             lagscope.set_busy_poll()
+        mtu = variables.get("network_mtu", None) if variables is not None else None
+        if not mtu:
```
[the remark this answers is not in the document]
SRIKKANTH: This condition is supposed to handle invalid mtu values such as 'None', '', "", 0.

SRIKKANTH:
@LiliDeng  please let me know if anything more needs to be done on this PR.

squirrelsc:
Please hold off on this change for now. We're currently adding a unified message for ntttcp. Once that's merged, please add the new metrics to the unified message.

#3863

squirrelsc:
> Please hold off on this change for now. We're currently adding a unified message for ntttcp. Once that's merged, please add the new metrics to the unified message.
> 
> #3863

#3863 is merged, please update the PR to enable mtu for unified ntttcp messages too.

microsoft/testsuites/performance/common.py
```diff
@@ -364,6 +366,20 @@ def perf_ntttcp(  # noqa: C901
             ntttcp.setup_system(udp_mode, set_task_max)
         for lagscope in [client_lagscope, server_lagscope]:
             lagscope.set_busy_poll()
+        mtu = variables.get("network_mtu", None) if variables is not None else None
```
squirrelsc: add prefix to differentiate it from other tests, for example `perf_ntttcp_mtu`.

squirrelsc: Verify it's an integer, if not fail the test case.

lisa/tools/ntttcp.py
```diff
@@ -389,10 +390,14 @@ def create_ntttcp_tcp_performance_message(
         buffer_size: int,
         test_case_name: str,
         test_result: "TestResult",
+        client_mtu: int = -1,
+        server_mtu: int = -1,
     ) -> NetworkTCPPerformanceMessage:
         other_fields: Dict[str, Any] = {}
         other_fields["tool"] = constants.NETWORK_PERFORMANCE_TOOL_NTTTCP
         other_fields["buffer_size"] = Decimal(buffer_size)
+        other_fields["client_mtu"] = client_mtu
```
squirrelsc: If the value is `-1`, it shouldn't set and send the message.

lisa/messages.py
```diff
@@ -251,6 +251,8 @@ class NetworkTCPPerformanceMessage(PerfMessage):
     retransmitted_segments: Decimal = Decimal(0)
     congestion_windowsize_kb: Decimal = Decimal(0)
     protocol_type: Optional[str] = TransportProtocol.Tcp
+    client_mtu: int = 0
```
squirrelsc: Don't update old messages—use the new unified message to collect data. Updating old messages requires modifying the data table, which is cumbersome.

microsoft/testsuites/performance/common.py
SRIKKANTH: Fixed

lisa/tools/ntttcp.py
SRIKKANTH: Fixed

lisa/messages.py
SRIKKANTH: Most of the powerbi reports still use old table for netperf reports, This is for backward compatibility.

squirrelsc: set it to -1 to keep consistent.

lisa/tools/ntttcp.py
```diff
@@ -566,6 +583,18 @@ def send_ntttcp_tcp_unified_perf_messages(
                 "relativity": MetricRelativity.LowerIsBetter,
                 "unit": "cycles/byte",
             },
+            {
```
squirrelsc: Not to add them, if it's -1

microsoft/testsuites/performance/common.py
squirrelsc: Where is the integer validation added? It should ideally be placed near `variable.get`.

SRIKKANTH: added "is_valid_mtu" to tool IP to validate MTU and used here.

lisa/tools/ip.py
```diff
@@ -215,6 +215,18 @@ def restart_device(
...
 
+    def is_valid_mtu(self, value: Optional[int] = None) -> bool:
+        if value is None:
+            return False
+        try:
+            mtu = int(value)
+            # typical MTU values are between 576 and 9000
+            # 576 is the minimum possible MTU for IPv4 and 1280 for IPv6
+            # 9000 is a common value for jumbo frames on Ethernet networks
+            return 576 <= mtu <= 9000
```
squirrelsc: If someone sets the value to 570 but it runs with a different value, it should raise an exception and fail the case instead of silently using the default.

SRIKKANTH: set_mtu has this logic already.

squirrelsc: Based on the logic, if the MTU is out of range, the function returns false and sets MTU to `None`, which is unexpected. If you want `set_mtu` to fail via command, remove the check here.

SRIKKANTH: Fixed as suggested. Included the mtu check into set_mtu method.

lisa/tools/ip.py
```diff
@@ -215,23 +215,16 @@ def restart_device(
...
     def get_mtu(self, nic_name: str) -> int:
         cat = self.node.tools[Cat]
         return int(cat.read(f"/sys/class/net/{nic_name}/mtu", force_run=True))
 
     def set_mtu(self, nic_name: str, mtu: int) -> None:
+        # Check if mtu is integer
+        try:
+            mtu = int(mtu)
+        except ValueError:
+            raise LisaException("MTU value is not an integer")
```
squirrelsc: Print the raw value in the string to assist with troubleshooting.

squirrelsc:
@LiliDeng LGTM

lisa/tools/ip.py
SRIKKANTH: Fixed as suggested.

LiliDeng (approved):

